PR #17, opened 2016-10-06
Ensure ${CLOUDWATCH_GROUP} lookup for single checks

lyletagawa:
- use `${CLOUDWATCH_GROUP}` as the metric-filter log-group-name


toniblyx:
Thanks @lyletagawa good catch! I missed to change those!

PR #35, opened 2016-12-29
fix for rule 2.4

davidpanofsky:
fixes rule and removes reliance on position-dependant output.


toniblyx (approved):
Good improvement, thanks!

PR #41, opened 2017-05-31
Match the entire username when running check12

AlexCline:
When a password-enabled user with a short name (e.g. "bc") is matched against another user whose username contains the first (e.g. "abcd"), check12 would erroneously display the second user "abcd" as having a password and no MFA even if the second user was not password-enabled.

This change ensures that grep matches the whole word.


toniblyx:
Nice catch @AlexCline thanks!

PR #47, opened 2017-06-14
fixes #45 - define TITLE116 variable

### LINKED ISSUE #45: Check 1.16 Missing title in output
I noticed that the title for the 1.16 check is missing. 

You can reproduce using the following command:
`./prowler -c check116`

It should output: 1.16 Ensure IAM policies are attached only to groups or roles (Scored)

MrSecure:
fixes #45 
The title variable for check 1.16 had the wrong name.

Also removes some trailing whitespace.

### DIFF
--- a/prowler
+++ b/prowler
@@ -480,7 +480,7 @@ check115(){
 }
 
 check116(){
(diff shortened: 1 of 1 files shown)


toniblyx (approved):
Thanks @MrSecure for you PR! Sorry to be late, busy days...

PR #48, opened 2017-06-14
fixes #46 - checks 1.16, 1.17, 1.18 were missing from the default lis…

### LINKED ISSUE #46: Check 1.16, 1.17, 1.18 missing from default report
Running prowler with the default options skips checks 1.16 - 1.18. These checks are executed when running prowler with the -c check1 option. The default report should execute all checks right?

MrSecure:
fixes #46 - checks 1.16, 1.17, 1.18 were missing from the default list of checks

remove some trailing whitespace.


jacobcsmith:
Thanks for the quick fixes!

PR #50, opened 2017-06-15
Add Monochrome Mode and other fixes

### LINKED ISSUE #49: Monochrome mode?
I find the escape codes can make tool output hard to diff, grep, and parse with other tools.  Any objection to me posting a PR for such a feature?

MrSecure:
Thank you for all the work you've put into this tool!  

Add a monochrome output mode which does not use any escape sequences for colors (-b for basic output).

Update the color tagging to use BAD for warnings, rather than RED

Update output text & color handling for consistency:
* Only the "OK!" text is colored on OK items
* Entire line is colored for WARNING and NOTICE items
* All WARNING and OK items include those words at the beginning of the line

Update Check 1.16 to show OK  when appropriate

Fix Check 2.4 to handle case when API region is not the bucket's HomeRegion.

FIxes #49

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,6 +93,7 @@ USAGE:
       -c <checknum>       specify a check number or group from the AWS CIS benchmark (i.e.: check11 for check 1.1 or check3 for entire section 3)
       -f <filterregion>   specify an AWS region to run checks against (i.e.: us-west-1)
       -m <maxitems>       specify the maximum number of items to return for long-running requests (default: 100)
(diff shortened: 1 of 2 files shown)


toniblyx (approved):
Approved -b option for monocrome/no colors output

toniblyx:
I liked more to have the entire OK line in green. But now looks cleaner with and without colors. Thanks for this and the other fixes.

Issue #52, opened 2017-06-26
CSV Output mode


MrSecure:
the ability to generate a CSV report would be immensely useful to compare findings across multiple runs and multiple profiles.

toniblyx:
That's a great enhancement, reviewing your PR right now. Thanks!

toniblyx:
Enhancement added in PR #53 
Thanks @MrSecure

PR #53, opened 2017-06-26
Add csv output mode

MrSecure:
Replaces "-b" switch with "-M <mode>" argument which has "text", "mono", and "csv" as valid values.

* Updates output generation in all check functions to utilize helper functions.
* Removes commas from generated output text
* When running in "csv" mode, sends the CSV output to STDOUT, and progress info to STDERR
* CSV output includes profile, region, item id, result (PASS, INFO, WARNING), is item scored, item text, generated text

Implements #52


toniblyx:
Thanks @MrSecure for this improvement and code clean up. Now it is easier to read. 

I'm testing -M csv option in OSX and I get these first lines as output (my comments below the output):

```
./prowler -p security -r us-east-1 -M csv


Generateing "," delimited report on stdout;  Diagnostics on stderr.
  Using Profile security

PROFILE,REGION,TITLE_ID,RESULT,SCORED,TITLE_TEXT,NOTES
0.0 Show report generation info
date: illegal option -- I
security,us-east-1,0.0,INFO,1,Show report generation info,ARN: arn:aws:sts::623229943871:assumed-role/oktaAdminRole/toni.delafuente@alfresco.com  TIMESTAMP:
0.1 Generating AWS IAM Credential Report...
```

Typo in "Generateing".
In OSX, date options `-Is` for Linux don't work, the right command is: `date -u +"%Y-%m-%dT%H:%M:%S%z"` we could add that here https://github.com/MrSecure/prowler/blob/8232a07437d61962f2d7a087a86d9c45f67c5167/prowler#L150 
what do you think?

MrSecure:
I like that idea ... consider it done.

prowler
```diff
@@ -349,308 +435,347 @@ check13(){
...
+        textWarn "Support Policy not applied to any Group / User / Role "
       fi
     done
   else
-    echo -e "     $BAD WARNING! No Support Policy found$NORMAL"
+    textWarn "No Support Policy found"
   fi
 }
 
 check123(){
```
MrSecure, line 749: produces errant CSV when multiple users are listed.

PR #58, opened 2017-07-07
Parallelism Tweaks

MrSecure:
* Make the temporary file name random to allow multiple instances to run in parallel
* Add a trap handler to clean up the temporary file when interrupted/killed
* Add the account id to the CSV output, in addition to the profile.

Example parallel execution - runs up to 4 simultaneously `-P 4`
```
grep -E '^\[([0-9A-Aa-z_-]+)\]'  ~/.aws/credentials | tr -d '][' | shuf |  \
xargs -n 1 -L 1 -I @ -r -P 4 ./prowler -p @ -M csv  2> /dev/null  >> all-accounts.csv
```

### DIFF
--- a/prowler
+++ b/prowler
@@ -362,7 +362,7 @@ genCredReport() {
 
 # Save report to a file, decode it, deletion at finish and after every single check, acb stands for AWS CIS Benchmark
 saveReport(){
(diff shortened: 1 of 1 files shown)


toniblyx:
Nice one! I'll add that command to the README, it is pretty helpful. Thanks!

PR #60, opened 2017-07-11
Fail early

MrSecure:
Exit the script immediately if unable to get-caller-identity - this indicates an issue with the credentials (invalid key, typo in ~/.aws/credentials, etc).

For listing topic subscribers, detect a permission failure, and fall back to showing counts of SNS topics by region, rather than subscriber details.  This works around the missing permission in the AWS Managed "SecurityAudit" policy.


toniblyx (approved):
All these changes make sense to me

PR #61, opened 2017-07-11
Policy requirements

MrSecure:
Add a policy document with just the added permissions compared to the current SecurityAudit policy, along with an example of how to create a group and user with both policies applied.


toniblyx (approved):
Very helpful for users!

PR #72, opened 2017-07-18
update check315

MrSecure:
make output more consistent across permissions and subscriptions


toniblyx (approved):
yes! That's much better 👍

PR #73, opened 2017-07-18
add extra check - look for admins w/o MFA

MrSecure:
Add a check for users in a group with the AWS managed "AdministratorAccess" policy applied, but who do not have MFA enabled.

Picked 7 as the "section" number.


toniblyx (approved):
It's time for extras!

PR #89, opened 2017-07-25
Allow Prowler to work in Win 10 Git Bash

steverigby:
Fixed tests 1.23:
sed: character class syntax is [[:space:]], not [:space:]
Fixed test 7.2:
) for parameter snapshotId is invalid. Expected: 'snap-...'.escribeSnapshotAttribute operation: Value (snap-fcc6a8dd
Fixed test 7.3:
": Bucket name must match the regex "^[a-zA-Z0-9.\-_]{1,255}$"

### DIFF
--- a/prowler
+++ b/prowler
@@ -221,8 +221,29 @@ elif [[ "$OSTYPE" == "cygwin" ]]; then
     {
       base64 -d
     }
+elif [[ "$OSTYPE" == "msys" ]]; then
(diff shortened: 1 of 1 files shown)


toniblyx:
I can't see the actual changes since I see the entire file changed. Can you please PR only the changes you have made? It will make easier to review and understand.

PR #91, opened 2017-07-28
Broken sed expression & typos

neonbunny:
A broken SED expressions prevented the check for rule 1.23 from showing accounts which have never used access key 1.
A number of typos in comments/output were also fixed.


toniblyx (approved):
Well spotted. Thanks!

PR #92, opened 2017-08-03
74 fix scored output

### LINKED ISSUE #74: CSV output has incorrect scored data
the third argument to titleText was removed from the not-scored items recently.

One hidden side-effect of the change - that was used to update the flag used to determine if a check was scored. I'll create PR shortly.

MrSecure:
fixes #74 

Adds "Level" column to CSV output
Changes call signature for titleText() to expect more self-explanatory parameters


toniblyx (approved):
Ok!

PR #97, opened 2017-08-21
removed printCurrentDate() and added current date to banner

hemi-hga:
In my opinion printCurrentDate() has two issues:

1.  It is used only once right after prowlerBanner() so it might as well be merged
2. It is using textNotice() which add the string "INFO!" at the start of the line for no apparent reason, much unlike the behavior of printColorsCode()

_Before_
image
_After_
image


toniblyx (approved):
Nice change. Makes sense. Thanks.

Issue #96, opened 2017-08-21
Removing the reference to: http://bit.ly/2g3PEf7


hemi-hga:
This reference doesn't add much information and may encourage uninformed users to click on unverified links. Whoever is working with the tool should also have the benchmark at hand to understand/handle issues.

I'll be happy to create a pull request for this issue, I'm just not sure how is best to tackle this.
Maybe remove all references from the reports aside from a link at the top of the report.

toniblyx:
Thanks @hemedga. Why do you think it may confuse users? The way to understand all these checks are looking at the CIS Benchmark guide. Did I miss something?

hemi-hga:
From a security standpoint, url shorteners may  be used to spread malware and web attacks.
Users click on these links without checking their origin.

I prefer to publish urls which are verified for the source or fully controlled by my organization.

toniblyx:
got it, you are talking about this: https://github.com/Alfresco/prowler/blob/0e0f3a334e60e57777a65f0eebe7e50900459b8e/prowler#L440

toniblyx:
You are right. You can send a PR with the way you consider is better. I would be glad to accept it.

PR #99, opened 2017-08-21
Fix issue #96 - remove bit.ly reference

hemi-hga:
Remove infoReferenceShort() from prowler.
Along with pull request #98, this totally fixed the issue in an elegant way.


toniblyx (approved):
Ok

PR #105, opened 2017-09-22
Clarify warning message

neonbunny:
A previous change replaced a comma in the message with the word "but" which is incorrect - the missing KMS key was the indicator that the trail was not encrypted.


toniblyx:
thanks @neonbunny

PR #106, opened 2017-09-26
Clarify warning message when CloudTrail is not consumed by CloudWatch

neonbunny:
This is likely another case of commas being replaced with "but" to form messages that make no sense.


toniblyx (approved):
This is a good one. Thanks.

PR #110, opened 2017-10-05
Fix typo

### LINKED ISSUE #107: Possible to not rely on the ~/.aws config?
I predominantly use API keys as environment variables by way of [apicreds](https://github.com/auraltension/apicreds) and thus haven't really ever needed to set up AWS local profiles (and would prefer not to given the local persistent plaintext creds).

I'm wondering if there's an easy way I can use prowler in this scenario? The tool seems pretty hardcoded to expect the aws config file to exist, but would love to give it a spin.

netflash:
This fixes #107

### DIFF
--- a/prowler
+++ b/prowler
@@ -247,7 +247,7 @@ if [[ $PROFILE ]]; then
     fi
   fi
 else
-  if [[ $AWS_ACESS_KEY && $AWS_SECRET_ACCESS_KEY || $AWS_SESSION_TOKEN ]];then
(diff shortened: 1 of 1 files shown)


toniblyx (approved):
Wow, good catch. It worked anyway if you declare variables properly but now it check them correctly.

netflash:
It didn't work for me without it.
I've been watching the thread and as I wanted to have this feature - tested it as soon as possible.

Nice work!

toniblyx:
Great! Thanks for your contribution @netflash glad to see it is helpful.

Issue #112, opened 2017-10-09
Check ID32="3.2,3.02" needs improvement


GaboP87:
Hi,

From the AWS_CIS_Foundations_Benchmark PDF file (https://d0.awsstatic.com/whitepapers/compliance/AWS_CIS_Foundations_Benchmark.pdf), for the check 3.2 (page 91), it is stated that it would be necessary to check for this Filter Pattern of the specified Metric:

**"filterPattern": "{ ($.eventName = "ConsoleLogin") && ($.additionalEventData.MFAUsed != "Yes") }"**

Since it is not possible to check only for this because it would involve ConsoleLogin for IAM Federated Users as well (assumed roles for the federated groups in AD), and it is not possible to configure MFA for federated users (IAM Roles) as a native AWS Solution, I created the following Filter Pattern which covers the ConsoleLogin API call with _success_ for any generic IAM User, as well as for the root user of the AWS Account:

**"filterPattern": "{ ($.userIdentity.type= IAMUser || $.userIdentity.type= Root) && ($.responseElements.ConsoleLogin= Success) && ($.additionalEventData.MFAUsed != \"Yes\") }"**

This works well from any CIS Benchmark Security Requirement. However, from a quick review of the Prowler script (Line 1058), I found the following Regex checking for:

**'userIdentity.sessionContext.attributes.mfaAuthenticated.\*true'**

This will clearly result in an error. It checks for any Metric which looks for Logs containing the sessionContext attribute. It does not appear whenever an IAM User or the root account signs successfully to the console without MFA. It will check for all the API Calls with that parameter in the Log stream. Examples of these API Calls:

- ListInstanceProfiles
- DescribeStaleSecurityGroups
- DescribeAddresses
- etc.

And neither of the API Calls containing the sessionContext parameter have ConsoleLogin as an _eventName_ parameter.

Can you help me with this please?

Cheers,
Gabriel

toniblyx:
Thanks @GaboP87, do you mean that changing L 1058 with your suggestion works fine?

GaboP87:
Hi @toniblyx, no. I mean that the Regex that is being used in the Prowler script is the wrong one in order to correctly check for a successful login without MFA. If you check the AWS_CIS_Foundations_Benchmark PDF file, in page 91 you can read the exact description of the requirement to be fulfilled in this case.

[...]

PR #117, opened 2017-10-26
Fixed issue #112

toniblyx:


morningseal:
#117 fix doesn't validate the CIS filter pattern so the false 'Warning' persists (based on code pull 2-Nov) unless one changed the filer to meet Prowler criteria

PR #122, opened 2017-11-08
Better grep pattern to find users in the IAM report

dougbyrne:
I had some false positives for check 1.23 due to user names that matched some other things in the credential report.

This change makes sure only the first column is matched when searching for the user.


toniblyx:
Good point @dougbyrne, thanks for the improvement!

PR #130, opened 2017-11-22
Avoid name clashes on word boundaries

pbugnion:
Firstly, thanks for making prowler available. It's proving very useful for us.

I think we've found an infidelity in check12 (users should only have passwort with MFA).

Let's say I have a user called `pascal` with password and MFA enabled. I also have a user called `pascal-dev` with neither password nor MFA enabled. In principle, this is a valid setup.

Unfortunately, `echo "pascal-dev" | grep -w "pascal"` matches, so it thinks that `pascal-dev` has password enabled, but no MFA. 

In this fix, we grep for `'pascal '`, including the space, to guarantee an exact match. Since IAM usernames can't include spaces, there is no room for a spurious match.


toniblyx (approved):
Thanks @pbugnion, I fixed something similar in other check few months ago. Good one!

PR #134, opened 2017-12-06
Check 1.24: fix error getting policy version when multiple policies share the same words

pbugnion:
If I have two policies called `policy` and `policy-staging`, prowler fails to get the right policy version -- the `POLICY_VERSION` string will look like `v3\nv2`, which breaks the next command.

In this PR, I fix it by replacing `grep -w $policy` with a slightly more complex awk script. An alternative (which I'm happy to implement if you think it's preferable) would to use `grep "$policy\t | awk '{ print $2 }'`.


toniblyx (approved):
Thanks

PR #135, opened 2017-12-07
Support "" (blank) values to if [[ ]] statements

sente:
This prevents errors like:

```
 ./prowler: line 672: [: ==: unary operator expected
 ./prowler: line 677: [: ==: unary operator expected
```


toniblyx (approved):
Ok, thanks

PR #136, opened 2017-12-12
Use a query to get AccountMFAEnabled on a root account rather than awk/sed

tommilata:
Parsing with awk/sed relies on the json being pretty printed (no other values on the same line), which is not always true, causing false-positive warings sometimes. Querying for SummaryMap.AccountMFAEnabled directly should be more robust.


toniblyx (approved):
Yeah, nicer this way. Thanks!

tommilata:
Thanks for your quick response!

PR #137, opened 2017-12-13
Use a query to get AccountMFAEnabled rather than awk/sed

tommilata:
This fixes the same problem as https://github.com/Alfresco/prowler/pull/136. (Sorry I didn't notice this in the previous PR).

Parsing with awk/sed relies on the json being pretty printed (no other values on the same line), which is not always true, causing false-positive warings sometimes. Querying for SummaryMap.AccountMFAEnabled directly should be more robust.


toniblyx (approved):
ok, thanks

PR #155, opened 2018-01-31
added extra79 to check for internet facing ELBs.

### LINKED ISSUE #147: Add check for internet facing ELBs
Another extra check for internet facing ELBs. Probably environment specific for people but if you generally use internal resources this could be a red flag.

sidewinder12s:
Fixes #147


toniblyx (approved):
Thanks @sidewinder12s good job!

PR #156, opened 2018-02-01
Extra check 710

### LINKED ISSUE #148: Thoughts on a check for public instances?
Thoughts on adding a check for either any instances with any public IP, any EIPs or individual checks for either?

I've kind of gone back and forth on the value of this since a public instance with a proper SG might not be an issue but one with

sidewinder12s:
I can rebase if #155 gets accepted. 

Fixes #148


toniblyx (approved):
OK

PR #160, opened 2018-02-01
Update extras 7.8-10 to display public endpoints as well as names.

### LINKED ISSUE #158: Display IPs/DNS endpoints for resources
For checks like 7.8-10, do you guys have any opposition to outputting the DNS/IP addresses of the resources it finds into the reports?

I wasn't sure if you'd just want to stick to AWS resource names but we're looking at trying to dump lists of IPs from Prowler to then point scanning tools at.

sidewinder12s:
Fixes #158


toniblyx (approved):
Ok!

PR #162, opened 2018-02-05
Update to 'check114'

subramani95:
Updated 'check114' to ensure hardware MFA is enabled for root account by:-
1) Querying for 'SerialNumber' of the Virtual MFA Devices list
2) 'SerialNumber' is ARN for Virtual MFA Device and Device Number for Hardware MFA Device; so did grep for ARN with 'root-account-mfa-device' in the expression


toniblyx (commented):
Hi @subramani95, thanks for your PR. Quick question, did you test it agains account with different configurations? Without any MFA enabled, with Only Virtual MFA enabled and with Hardware MFA enabled?

subramani95:
Hi @toniblyx , tested it with no MFA and Only Virtual MFA enabled.

toniblyx (approved):
Ok, I've checked it also to better understand your change. Thanks!

PR #174, opened 2018-02-12
Improved 'check31'

subramani95:
Get the Metric Name using Log Group and Filter names and check the alarms associated with that metric.


toniblyx (approved):

toniblyx:
Thanks @subramani95

PR #176, opened 2018-02-13
Fix printing texts when both Ok and Warn are present in 'check31'

subramani95:
We may have groups in both Ok and Warn; so we have to go through both CHECK31OK and CHECK31WARN one by one.

### DIFF
--- a/prowler
+++ b/prowler
@@ -1192,7 +1192,8 @@ check31(){
         group=${group%:*}
         textOK "CloudWatch group $group found with metric filter $metric and alarms set for Unauthorized Operation and Access Denied"
       done
-    else
+    fi
+    if [[ $CHECK31WARN ]]; then
       for group in $CHECK31WARN; do
         case $group in
            *:*) metric=${group#*:}


toniblyx:
Can you please print a sample output for this check with that case? Thanks

subramani95:
@toniblyx here is a sample output snippet:-
> 3.1  Ensure a log metric filter and alarm exist for unauthorized API calls (Scored)
>        OK!  CloudWatch group CloudTrail/AruviLogGroup found with metric filter aruvi-filter-name-401 and alarms set for Unauthorized Operation and Access Denied
>        WARNING! CloudWatch group TestUnauthorizedAPI found but no metric filters or alarms associated

toniblyx:
I don't understand how you can have groups in Ok and Warn, or didn't understand something?

subramani95:
In this configuration, I have two cloud trails where one has the log group with expected settings for rule 3.1 and another trail has the log group without metrics and alarms.

PR #180, opened 2018-02-20
Improving check41 and check42

subramani95:
Improved check41 and check42 to ensure no inbound rule exists that has:-
1. port no 22 and source of 0.0.0.0/0
2. port no in the range (i.e 0-1024) and source of 0.0.0.0/0
3. port value of all and source of 0.0.0.0/0

### DIFF
--- a/prowler
+++ b/prowler
@@ -1576,7 +1576,7 @@ check41(){
   # "Ensure no security groups allow ingress from 0.0.0.0/0 to port 22 (Scored)"
   textTitle "$ID41" "$TITLE41" "SCORED" "LEVEL1"
   for regx in $REGIONS; do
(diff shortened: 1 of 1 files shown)


toniblyx (commented):
Thanks @subramani95 what is the difference between exiting and suggested command? I see it but is it faster or better? It is just to make sure that I'm undertanding everything correctly

subramani95:
@toniblyx  Existing command handles only the first case (port no 22 and source of 0.0.0.0/0). The suggested command will handle second and third case also (port no in the range and port value of all). The checks will be better with these suggested commands.

subramani95:
Though 4.1 check summary says "Ensure no security groups allow ingress from 0.0.0.0/0 to port 22 (Scored)", in 'Audit' section it's mentioned to check all the cases for 22 (port no, range and all):-
```
3. For each security group, perform the following:
  1. Select the security group
  2. Click the Inbound Rules tab
  3. Ensure no rule exists that has a port range that includes port 22 and has a Source of 0.0.0.0/0

Note: A Port value of ALL or a port range such as 0-1024 are inclusive of port 22.```

toniblyx:
You are so right, I see it now. Thanks for clarifying!

PR #181, opened 2018-02-22
Improve check28

doshitan:
The CIS benchmarks state that only customer managed CMKs should be checked, so
exclude all AWS managed CMKs, not just the one for ACM.

Also fix up some formatting and dead code.


toniblyx (approved):
Thanks @doshitan, this is a very good one. I don't remember now but they may have changed the API response since I wrote this check.

PR #182, opened 2018-02-22
Use `#!/usr/bin/env bash` instead of hard coding `#!/bin/bash`

doshitan:
Better cross-platform support.


toniblyx:
That makes a lot of sense, thanks @doshitan

PR #187, opened 2018-03-05
Update IAM permissions needed for CloudWatch Logs

rtkjbillo:
According to the AWS documentation for the CloudWatch Logs permissions reference [1], the IAM policy to permit or deny CloudWatch Logs actions uses the `logs:` prefix rather than `cloudwatchlogs:`. This commit updates the policy additions JSON file as well as the README to reflect this change.

I confirmed this having assumed an appropriate role in an AWS account, then executing the AWS CLI command `aws logs describe-log-groups`; with the `cloudwatchlogs:` prefix an AccessDeniedException was returned to the client.

[1] https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/permissions-reference-cwl.html


toniblyx (approved):
Good one, I think I confused this with cloudwatch: . Thanks!

PR #188, opened 2018-03-06
Update prowler

ahhh:
super tiny grammar / output fix. Great tool!


toniblyx (approved):
Thanks for improving it!

PR #194, opened 2018-03-20
Fix check 3.3

mbode:
0e43a05d167d2bc569db08e668a6bf27fc82e90d introduced an issue with check 3.3 due to an unescaped variable.


toniblyx (approved):
Thanks @mbode, my bad, I missed $ when I copied the line and my tests weren't good enough either. Sorry about that.

mbode:
All good, thanks for merging so quickly!

PR #199, opened 2018-04-10
allow -c option to be specified multiple times

JTarasovic:
I'd like to be able to run a subset of checks in a single run. This request allows that just by changing CHECKNUMBER to an array and iterating. 

I tested:
- ✅ `./prowler -c check`
- ✅ `./prowler -c check -c check -c check`
- ✅ `./prowler`

If you're not interested or if there's anything you'd like done differently (especially the help text), just let me know.

Thanks for the tool, btw. v cool.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,9 @@
 # Prowler: AWS CIS Benchmark Tool
 
-## Table of Contents  
+## Table of Contents
 - [Description](#description)
-- [Features](#features)  
-- [Requirements](#requirements)  
+- [Features](#features)
+- [Requirements](#requirements)
 - [Usage](#usage)
 - [Fix](#fix)
 - [Screenshots](#screenshots)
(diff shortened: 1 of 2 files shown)


toniblyx:
Hi @JTarasovic, thanks for your PR. I'm currently working in the v2.0 of Prowler and this is very easy to perform using groups functionality. Can you please take a look to the devel branch and try creating a new group? You can create your custom-group and then run `./prowler -g custom-group` with all checks that you need. Please, let me know what do you think.

toniblyx:
I'll close this PR until review of the current implementation for 2.0-beta2

JTarasovic:
Toni,

Sorry, I haven't had a chance to check but it sounds like the 2.0 setup will work. If it doesn't, this change probably wouldn't fit w/ that work anyway.

Thanks again!

PR #202, opened 2018-04-19
Fix curl error; improve default region

ceyes:
Related to https://github.com/toniblyx/prowler/pull/200

If the ec2 instance is not attached to an IAM role, the `curl -s -m 1
http://169.254.169.254/latest/meta-data/iam/security-credentials` will
return a 404 page instead of null, INSTANCE_PROFILE will always be true
and result to curl error when trying to use default cli profile:

        curl: option -: is unknown
        curl: try 'curl --help' or 'curl --manual' for more information

### DIFF
--- a/prowler
+++ b/prowler
@@ -23,7 +23,7 @@ OPTRED="[1;31m"
 OPTNORMAL="[0;39m"
 
 # Set the defaults for these getopts variables
-REGION="us-east-1"
+REGION=""
 FILTERREGION=""
 MAXITEMS=100
 MONOCHROME=0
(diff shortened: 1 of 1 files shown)


netflash (approved):
/lgtm

toniblyx (approved):
Thanks @ceyes, I think that way to handle REGION is better. Also for the INSTANCE_PROFILE. About the wait for the buckets check, I've removed that thread approach in Prowler 2.0 beta because it may cause delays due to the amount of concurrent queries to the API. I'll also apply these changes to devel branch which is 2.0 beta.

PR #205, opened 2018-04-20
Tweaks - AWS Account ID, -V flag

### LINKED ISSUE #204: 2.0-beta - missing AWS account IDs in CSV output
when running prowler with  -c or -g, the checks are run without calling getWhoami, resulting in the ACCOUNT_NUM variable never getting set and an empty account number in the output.

MrSecure:
fixes #204 

restores AWS Account ID in csv output
adds -V flag to report prowler version

### DIFF
--- a/prowler
+++ b/prowler
@@ -256,6 +256,9 @@ show_all_titles() {
 	done
 }
 
+# Gather account data / test aws cli connectivity
+getWhoami
+
 # Execute single check if called with -c
 if [[ $CHECK_ID ]];then
   execute_check $CHECK_ID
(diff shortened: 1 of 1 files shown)


toniblyx (approved):
Thanks @MrSecure, you are always looking after Prowler, hope you like the 2.0 changes in general and  enjoyed the sticker ;)

MrSecure:
the 2.0 changes are IMPRESSIVE!!    Very excited!

toniblyx:
Thanks, btw, why capital `-V` instead of lower?

MrSecure:
seems to be somewhat normative across lots of the tools I've used ... little "v" is for "verbose",  big "V" is for "VERSION".    So, habit, I guess?

toniblyx:
makes sense, I have seen that in many ways too. Ok, it is fine then. Keep an eye, I'll PR a GDPR beta group of checks in a moment, still work in progress.

PR #207, opened 2018-04-25
Restore check levels

MrSecure:
the "CHECK_TYPE_checkN" variables were missing from all checks, resulting in a CSV output with "Unspecified or Invalid" as the level.

This restores the check levels of "Level 1", "Level 2", or "Extra" to existing checks.


toniblyx (approved):
Thanks @MrSecure, I missed that part during the refactoring.

PR #208, opened 2018-04-27
ensure credential report is available before running any checks

MrSecure:
rather than leaving check/group authors to determine if each check requires the credential report, just create the report before running checks.

running `prowler -p test_profile -c check112 -k`  returned a FAIL result, and did NOT generate the credential report.


toniblyx (approved):
Ok, makes sense

PR #216, opened 2018-05-29
Support graceful failing of buckets with corrupt/unintended permissions

chickenriggies:
There are situations where buckets are not accessible to privileged IAM users. In our accounts, we see this on buckets that have restrictive bucket policies and on buckets that have lost their "Access for your AWS account" access. Most auditing tools seem to gloss over these buckets or error-out. This little fix avoids this from happening and reports a failure on the bucket w/ a custom error message.


toniblyx (approved):
Good enhancement @hb3b, thanks!

PR #220, opened 2018-06-13
fix report

yapale:
this fix will exclude users that have  PasswordLastUsed , but Console password
 is disabled. 

users that their "Console password" is disabled  , should be ignored even if their PasswordLastUsed is populated .


toniblyx:
Thanks for the PR @yapale. Why this change? Even if "console password" is disabled I think is important to know that has been used.

yapale:
hi toni,we have alot of users that turned to be service accounts, so their
login is not relevant.
i can add flag to make it optional. wdut?

בתאריך יום ה׳, 14 ביוני 2018 ב-2:32 מאת Toni de la Fuente <
notifications@github.com>:

toniblyx:
Thanks for clarifying. You can also create an extra check with your
requirement and invoke it instead of the original one in that group.

Issue #221, opened 2018-06-13
Extra 727 and Extra 728 missing PROFILE_OPT


tmonk42:
Both of these checks are missing the profile variable on the AWS CLI call made inside of the loop.

https://github.com/toniblyx/prowler/blob/master/checks/check_extra727#L26
and
https://github.com/toniblyx/prowler/blob/master/checks/check_extra728#L26

tmonk42:
Thanks!

PR #222, opened 2018-06-13
Fix extra727 and extra728 #221

tmonk42:


toniblyx (approved):
Good one, thanks @tmonk42

PR #223, opened 2018-06-26
Check-111 Password expiration check

roo7break:
Modified logic to accept value less than or equal to 90.

Also edited print statement to show set value.


toniblyx (approved):
Good improvement, thanks!

PR #224, opened 2018-06-26
Region support for Check 27

roo7break:
Check was failing if Region was not explicitly provided. Updated script to work with or without Region (script will work regardless).

#### Improvement/Suggestion:
In my test area, just providing the following command worked. Note that in my case, only one trail and key existed. Dont know if this will work when multiple trails and keys exist.
`aws cloudtrail describe-trails --output text --profile <profile> --query 'trailList[*].KmsKeyId'`

### DIFF
--- a/checks/check114
+++ b/checks/check114
@@ -20,7 +20,7 @@ check114(){
   if [ "$COMMAND113" == "1" ]; then
     COMMAND114=$($AWSCLI iam list-virtual-mfa-devices $PROFILE_OPT --region $REGION --output text --assignment-status Assigned --query 'VirtualMFADevices[*].[SerialNumber]' | grep '^arn:aws:iam::[0-9]\{12\}:mfa/root-account-mfa-device$')
     if [[ "$COMMAND114" ]]; then
-      textFail "Only Virtual MFA is enabled for root"
+      textInfo "Only Virtual MFA is enabled for root"
     else
       textPass "Hardware MFA is enabled for root "
     fi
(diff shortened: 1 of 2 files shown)


roo7break:
Additional: Changed reporting Virtual MFA in Check 114 to Info instead of Fail. Having MFA is better than not having one at all.

toniblyx (commented):
Hi @roo7break for CIS check114 is about HW MFA, if it is not HW it is a fail. Probably it is good idea to keep it as it is now and create a new custom check with your preference and then add it to your custom group of checks. About check27 and region, if you don't want to specify region add a default one in your AWS config or here https://github.com/toniblyx/prowler/blob/master/prowler#L38
What do you think?

adamprice2:
Not sure if this is the right place to comment, but check27 seems to be failing for me.
However if I use trail ARN's instead of trail names it seems to work.
So I changed line 19 of the check27 script as follows:-


`CLOUDTRAILNAME=$($AWSCLI cloudtrail describe-trails --query 'trailList[*].Name' --output text $PROFILE_OPT --region $REGION)`

Becomes 

`CLOUDTRAILNAME=$($AWSCLI cloudtrail describe-trails --query 'trailList[*].TrailARN' --output text $PROFILE_OPT --region $REGION)`

toniblyx:
Hi @adamprice2 can you please open an issue specific for your problem? I just don't want to mix up issues. Thanks.

PR #232, opened 2018-08-10
Update Scored Value for 115 and 315. Update 13 to only check users with enabled console password.

rheak:
When I run prowler and export to json, I get a Scored value of 'Scored' instead of 'Not Scored' for 1.15 and 3.15. In their Title/Description it correctly says (Not Scored), but the outputted value needs to be updated.

Would it make sense to update check 1.3 to only check IAM users that have their password enabled? There are use cases where accounts previously used IAM account for console acess and now only use IAM accounts for Access keys. This use case still lists a last login value even though the user's console password has been disabled which results in a failed finding in the report that cannot be remediated.

### DIFF
--- a/checks/check115
+++ b/checks/check115
@@ -10,7 +10,7 @@
 
 CHECK_ID_check115="1.15"
 CHECK_TITLE_check115="[check115] Ensure security questions are registered in the AWS account (Not Scored)"
-CHECK_SCORED_check115="SCORED"
+CHECK_SCORED_check115="NOT_SCORED"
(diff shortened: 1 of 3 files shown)


toniblyx (changes requested):
Hi @rheak thanks for your fixes, however the one for check13 doesn't makes sense to me, it says `Ensure credentials unused for 90 days or greater are disabled (Scored)` and has nothing to do with console password, if you want a new check you can write it as PR as an extra check.

toniblyx (approved):

PR #240, opened 2018-08-29
Override CloudTrail region check so it passes in single region mode

echery:


toniblyx:
I hope you could use it as you needed.

PR #241, opened 2018-09-01
Support for fargate's container metadata endpoint

mattfinlayson:
If you're running prowler in a fargate container in AWS you don't get an instance profile back from `http://169.254.169.254/latest/meta-data/iam/security-credentials/` instead you get an empty string. In this case if you have an empty instance profile and the `$AWS_CONTAINER_CREDENTIALS_RELATIVE_URI` environmental variable is set you can get the same access keys and token from a slightly different URI. This change will detect this and set the required ENV vars for a clean prowler run.


toniblyx (approved):
Cool! Thanks for making this!

mattfinlayson:
👍

PR #244, opened 2018-09-12
Update Check 1.16

gpatt:
Updated Check 1.16 to look for directly attached Inline Policies per CIS v1.2 date 05-23-2018.


toniblyx (approved):
Thanks for the improvement!

PR #248, opened 2018-09-20
Take age of access key 2 from correct column

MichaelSchubertTNG:
Currently when a user has a correctly rotated first access key and an incorrectly rotated second access key, no warning is printed out.


toniblyx (approved):
Approved. Thanks!

PR #250, opened 2018-10-02
Use awscli query instead of cut for Extra 7.13

### LINKED ISSUE #249: Extra 7.13 does not work with awscli 1.16.25
… as the output of `aws guardduty get-detector` has changed.

mbode:
Fixes #249.
The text output of `aws guardduty get-detector` has changed with awscli release 1.16.25, leading to GuardDuty detectors misreported as suspended.


toniblyx (approved):
Approved, thanks!

PR #252, opened 2018-10-03
Allow AWS_PROFILE to be used from the environment

flomotlik:
AWS_PROFILE is a default AWSCLI environment variable configuring the profile to use. Prowler should accept it as well and not set the default profile.

More information on AWSCLI environment variables can be found in the docs: https://docs.aws.amazon.com/cli/latest/userguide/cli-environment.html


toniblyx (approved):
yes, it makes sense. Thanks

flomotlik:
Awsome, that was fast :)

PR #254, opened 2018-10-08
Add missing permissions for prowler policy

dan-petty:
Fixes permission denied errors from running prowler with IAM policy provided in README


toniblyx (approved):
I missed them. Thanks for this!

PR #257, opened 2018-10-09
Add missing permission to example prowler policy

dan-petty:
"support:describetrustedadvisorchecks"


toniblyx (approved):
Thanks for the change @mixmatch!

Issue #245, opened 2018-09-12
Bug with date in Ubuntu 18.04


andresriancho:
```
 1.3  [check13] Ensure credentials unused for 90 days or greater are disabled (Scored)
date: invalid date ‘+%s’
./include/os_detector: line 25: (1536721200 -  )/60/60/24: syntax error: operand expected (error token is ")/60/60/24")
./checks/check13: line 25: [: -gt: unary operator expected
```

toniblyx:
Thanks @andresriancho. I think, that's related to the operating system. What OS are you using?

andresriancho:
Ubuntu 18.04

toniblyx:
Ok, I'm gonna check that out.

toniblyx:
I'm trying to reproduce the error with the same OS version and I can't. Can you please share the output of this command? `bash -x ./prowler -c check13`, please fake aws account ID, etc.

toniblyx:
For some reason this command doesn't work in your linux `date -d $DATE_TO_COMPARE +%s` 
https://github.com/toniblyx/prowler/blob/master/include/os_detector#L24 

If you can run `bash -x ./prowler -c check13` and send me the result I can see what else is going on.

toniblyx:
This is finally fixed on PR #259

PR #259, opened 2018-10-17
fix(check13): should not pass if user never logged in

SkxNMDI5:
When a user never logged in but have password enabled (done when creating the account with a temporary password), an error in the script (date of empty string) make the check pass :

```
./include/os_detector: line 49: (1539759446 -  )/60/60/24: syntax error: operand expected (error token is ")/60/60/24")
./checks/check13: line 25: [: -gt: unary operator expected
```


toniblyx:
Great! This definitely is related to issue #245.

toniblyx (approved):
Good one!

PR #262, opened 2018-10-19
Accelerate check 122 (scope local, 1 less API call by policy)

SkxNMDI5:
Reduce the cost of list-policies by using the option "--scope Local"
Avoid a second call to list-policies to retrieve the policy version.


toniblyx (approved):
I didn't know about `--scope Local`, it is a very good improvement, thanks!

PR #266, opened 2018-11-14
Updated check29 to validate against FlowLogId which is valid for both…

onkymykiss1:
… CloudWatch and s3 destinations

The existing check fails if VPC FlowLogs are enabled and set to deliver to an s3 bucket.  The LogGroupName metadata is only populated when the destination is set to go to Cloudwatch.

All VPC Flow logs will have a valid FlowLogId so changing the filter will validate the check appropriately.


toniblyx (approved):
Good improvement. Thanks!

PR #270, opened 2018-12-05
Fixed AccessDeniedException on extra730

SDugo:
When executing Prowler using a specific profile (in my case to assume a role) , check_extra730 returns:

"An error occurred (AccessDeniedException) when calling the DescribeCertificate operation: User: [ASSUMED_ROLE_ARN] is not authorized to perform: acm:DescribeCertificate on resource: [RESOURCE_ARN]"

This is because line 28 did not contain the following parameters: "$PROFILE_OPT --region $regx" .


toniblyx (approved):
Thanks @SDugo.

PR #276, opened 2018-12-12
Check for flowlogs only in active VPCs, avoid false flag if a region …

affanhmalik:
…has no VPCs


toniblyx (approved):
Looks fine, thanks @affanhmalik

PR #283, opened 2018-12-21
New option "-E" supports exclusion of one or multiple checks

SDugo:
Added new option "-E" which will execute all tests except a list of specified checks separated by comma (i.e. check21,check31). Any invalid check name will be discarded. If just one argument is passed and this is invalid, then Prowler will execute all checks.

The option will return a list of total checks excluding the list provided by the user. Then, the functionality will overwrite CHECK_ID with the final list and the program will continue as if the user entered "-c" option and the final list of checks.

This solution complements the fix for the issue [273](https://github.com/toniblyx/prowler/issues/273)

### DIFF
--- a/prowler
+++ b/prowler
@@ -70,6 +70,7 @@ USAGE:
       -l                  list all available checks only (does not perform any check)
       -L                  list all groups (does not perform any check)
       -e                  exclude group extras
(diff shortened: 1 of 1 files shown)


toniblyx (approved):
wow @SDugo this is also awesome, thanks for your contribution!

PR #290, opened 2019-02-04
Added check for integer in response from AWS

tomcrawf90:
If no password policy is set, AWS CLI returns the value "None"

This current results in the following:
```
 1.11  [check111] Ensure IAM password policy expires passwords within 90 days or less (Scored)
       PASS! Password Policy includes expiration (Value: None)
```
Added a basic check to ensure the response returned is a integer.

Post change we are now returned:
```
1.11  [check111] Ensure IAM password policy expires passwords within 90 days or less (Scored)
       FAIL! Password expiration is not set
```


toniblyx (approved):
Good fix! Thanks @tomcrawf90

PR #292, opened 2019-02-06
Use common code for CloudWatch metric checks

rniedzial:
Common code for CloudWatch Metrics checks. This ought to be less arbitrary when looking to connect Metrics to Alarms.  Rewrote checks 31-34 as quick proofs.

Related to: https://github.com/toniblyx/prowler/issues/289

### DIFF
--- a/checks/check31
+++ b/checks/check31
@@ -16,48 +16,6 @@ CHECK_ALTERNATE_check301="check31"
 
 check31(){
   # "Ensure a log metric filter and alarm exist for unauthorized API calls (Scored)"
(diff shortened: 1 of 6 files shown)


RyanBeatty:
Cloning and running this locally via `$ ./prowler -c 31` still reports as failed. I'm a bit bandwidth constrained at work, so apologies for not offering a more helpful debug :/

toniblyx (changes requested):
Thanks for the PR @rniedzial. These CIS checks are probably the ones that make less sense to me, everybody has a different way to handle them. I like the way you simplify the `grep` part but adding additional includes make every check less independent. I think it would be great to have every check with its own code but your grep implementation to simplify it.

PR #302, opened 2019-03-07
Update cislevel names README.md

mindfulmonk:
-c is only for individual checks, need to use -g for level1 checks.


toniblyx (approved):
Looks good to me

PR #306, opened 2019-03-12
Merge RUNs. Run as non-root

nicdoye:
Merge RUN commands to reduce number of layers.
Run as non-root.


toniblyx (approved):
Beautiful, I'll add jq :)

PR #307, opened 2019-03-25
Update check22

soffensive:
Returned value is lower case "false". Thus, prowler will not detect a missing log file validation (false negative)

Probably, this behaviour should also be taken into account for other checks.
A quick grep for "False" indicated:

- check_21
- check_extra739
- check_extra737
- check_extra715
- check_extra740
- check_extra73

### DIFF
--- a/checks/check22
+++ b/checks/check22
@@ -20,7 +20,7 @@ check22(){
   if [[ $LIST_OF_TRAILS ]];then
     for trail in $LIST_OF_TRAILS;do
       LOGFILEVALIDATION_TRAIL_STATUS=$($AWSCLI cloudtrail describe-trails $PROFILE_OPT --region $REGION --query 'trailList[*].LogFileValidationEnabled' --output text --trail-name-list $trail)
-      if [[ "$LOGFILEVALIDATION_TRAIL_STATUS" == 'False' ]];then
+      if [[ "$LOGFILEVALIDATION_TRAIL_STATUS" == 'false' ]];then
         textFail "$trail trail in $REGION has not log file validation enabled"
       else
         textPass "$trail trail in $REGION has log file validation enabled"


toniblyx (commented):
I wonder if that has been a change in AWS API response... I'll take a look. Thanks!

soffensive:
Ideally, the checks should be case-insensitive for more flexibility. But in bash this is somewhat unwieldy...

toniblyx:
You are right, I'm going to see what is the best way to make those checks more consistent and fix it. I'll do it on `devel` branch first and we can test them before hand.

toniblyx:
Ok, I have been taking a deep look to all checks you mentioned and this is what I have seen:

- `check22`: awscli output is in text, so it is True or False (IF condition is correct), with json output it is lowercase true or false.
- `check21`: awscli output is in text, so it is True or False (IF condition is correct)
- `check_extra739`: no longer exist (removed in `devel` branch so far) because it is redundant with `extra717` that works correctly despite the output is lowercase.
- `check_extra737`: awscli output is in text, so it is True or False (IF condition is correct)
- `check_extra715`: awscli output is in text, so it is True or False (IF condition is correct)
- `check_extra740`: awscli output is in text, so it is True or False (IF condition is correct)
- `check_extra73`: doesn't apply this case, there is not usage of true or false conditionals

You can run prowler like `bash -x ./prowler -c check22` to see exactly the command and test with `--output text` and `--output json` to see what I'm saying. Most of the prowler checks are using `--output text`. Please, also check that you have the latest version of `awscli` installed, this is my setup: `aws-cli/1.16.133 Python/2.7.16 Darwin/18.5.0 botocore/1.12.123`

Please, let me know if it makes sense for you as well and we can close this issue.

Thanks!

soffensive:
Thanks for your investigation.
I have to apologize, the false negative in check22 was *NOT* caused by the lower/upper case, I did not thoroughly check this.

Instead, it is caused by the region. check22 will use the default region set by prowler, while trails may be residing actually in any region.

On the one hand, the **describe-trails** command will yield all trails for all regions when there is no **--trail-name-list** argument, regardless of the passed argument to the **--region** parameter.

On the other hand, when trying to get the validation enabled status of a specific trail, the **--region** argument has to match the region of the trail, otherwise an empty response is returned in CLI text mode. check22 does not take this into account, since any response that is not "False" will be considered "True".

So my proposal would be to get all output from a single describe-trails command and then parse the output. Would do you think about this?

PR #311, opened 2019-04-05
Fixed check122 to match CIS 1.22 checks requirements, instead of '=~ …

artashus:
The CIS benchmark document states: IAM policies that have a statement with "Effect": "Allow" with "Action": "\*" over "Resource": "\*" should be removed.

In the code we have `contains(Resource, '*')` and `contains (Action, '*')`. This covers cases like `s3:Get*`, `arn:aws:s3:::my-bucket/*`, etc. and might generate false alarms. 

Also the the result has been updated from **INFO** to **FAIL**.


toniblyx (approved):
Makes sense, thanks for this PR @artashus

Issue #267, opened 2018-11-19
Move main code to python calling checks as they are now allowing parallel checks and more


toniblyx:
Moving the main base code to python (leaving checks in bash so far) would open up a world of possibilities to run parallel checks, improved and new output formats, easier multi-account support, integrations, etc.

Thanks to @rmogull for the idea.

jtyers:
@toniblyx I'll take a look at this when I have time and open a PR. We've been using prowler to scan a handful of environments and it's clear that having the tool in Python would be really beneficial for speed and its overall robustness.

toniblyx:
Hi @jtyers, thanks for your offer, it would be really helpful to start at least with the basics. I can create a new branch and you can PR to it. What do you think?

jtyers:
@toniblyx yep sounds great to me :+1:

toniblyx:
ok, here you go https://github.com/toniblyx/prowler/tree/python-devel. Looking forward to it!

jtyers:
Just leaving an msg to say I am making progress. My fork has the latest code. `prowler.py` will now generate the IAM credential report and call the shell-script checks as before. One behaviourial difference I've put in is that errors caused by the checks will be sent to `stderr` rather than `stdout`, so that JSON parsing isn't disrupted in the face of errors.

As this is stlil calling the checks via shell, the Python prowler will have similar performance to the current one. The next stage after `prowler.py` is in place is to re-write the checks themselves in Python. This is where the performance gains will come.

toniblyx:
Awesome! Thanks again. Even in python, performance is tricky here due to API rate limits but from there we can think about other options. Cheers!

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

toniblyx:
Hi @jtyers I've been testing your code a bit, it looks great! Let me know when you want to PR to devel-python and do a more extensive testing. Thanks!

jtyers:
Heya, glad you like it! Apart from a few tweaks around the output, so make
it consistent with current prowler, I think this is nearly good to go. I've
tested some but not all checks against it.

toniblyx:
Hey @jtyers I would like to PR your changes to `devel-python` and continue from there. What do you think?

jtyers:
@toniblyx see above :-)

toniblyx:
Thanks @jtyers I'll play with that code and see were we can go from there!

museadmin:
Hi Guys,

[...]

PR #313, opened 2019-04-14
Added code for python checking

jtyers:
Follows up on #267, adding `prowler.py` with support for the main checking process and supporting scripts. The actual checks are still called via the existing shell scripts.


toniblyx:
Awesome, thanks!!

PR #314, opened 2019-04-17
Iterate over all regions for Cloudtrail Checks

soffensive:
This should fix the issue described in the previously mentioned pull request
https://github.com/toniblyx/prowler/pull/307


toniblyx (approved):
It looks good to me, I'll test them before MR to master. Thanks @soffensive

PR #326, opened 2019-05-13
Update README.md to clone from right repo

RyPeck:
Looks like the project was moved out of an org to your personal account.


toniblyx (approved):
Yes, that's something that I forgot to change some time ago. Thanks @RyPeck

PR #329, opened 2019-05-29
fixed check122 to check result for the value None

jhoov:
The results of a search where the policy does not match returns "None".  As a result, "if [[ $POLICY_WITH_FULL ]]" is true and the policy is listed as having *:* privileges, even though it does not.

### DIFF
--- a/checks/check122
+++ b/checks/check122
@@ -23,7 +23,7 @@ check122(){
       POLICY_ARN=$(echo $policy | awk -F ',' '{print $1}')
       POLICY_VERSION=$(echo $policy | awk -F ',' '{print $2}')
       POLICY_WITH_FULL=$($AWSCLI iam get-policy-version --output text --policy-arn $POLICY_ARN --version-id $POLICY_VERSION --query "PolicyVersion.Document.Statement[?Action!=null]|[?Effect == 'Allow' && Resource == '*' && Action == '*']" $PROFILE_OPT --region $REGION)
-      if [[ $POLICY_WITH_FULL ]]; then
+      if [[ "$POLICY_WITH_FULL" != "None" ]]; then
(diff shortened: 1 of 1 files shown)


toniblyx:
Thanks @jhoov I’ll review and test it as soon as I can, in a few days. 

toniblyx:
Hey @jhoov I'm testing your suggested change and I don't get good results, it shows policies that are not actually allowing *:* administrative privileges. Can you give me some examples of policies that Prowler doesn't detect or those that make a false positive?

Thanks in advance and sorry for the late response!

jhoov:
I'm pretty sure the policy I tested against only allowed the assume role permission.  I'll verify tomorrow and send you the details of the policy along with a screenshot of the output.  Thanks for taking a look at it.

jhoov:
Hi Toni,

The policy in question allows an auditor account in our security account to assume the Auditor role in any other account:

{
    "Version": "2012-10-17",
    "Statement": {
        "Effect": "Allow",
        "Action": "sts:AssumeRole",
        "Resource": "arn:aws:iam::*:role/Auditor"
    }
}

Can you test against this policy?  The resource has a *, but the action does not.  I ran this from the CLI to verify the return value, using the correct account number, and the output is 'None':

aws iam get-policy-version --output text --policy-arn arn:aws:iam::999999999999:policy/AssumeAuditorRole --version-id v1 --query "PolicyVersion.Document.Statement[?Action!=null]|[?Effect == 'Allow' && Resource == '*' && Action == '*']"

Please let me know if you need additional information.

Thank you!

toniblyx:
Hi @jhoov this one has been fixed in PR #362, you can check it out and let me know. I close this one but feel free to reopen it if needed. Thanks

jhoov:
Thank you

PR #331, opened 2019-06-06
new ec2 age checks

gabrielsoltz:
[extra757] Check EC2 Instances older than 6 months (Not Scored) (Not part of CIS benchmark)
[extra758] Check EC2 Instances older than 12 months (Not Scored) (Not part of CIS benchmark)


toniblyx (changes requested):
Thanks for these checks @gabrielsoltz they seem helpful for discovery. However I think you have tested them only for OSX, am I correct? if you run `date -v -6m '+%Y-%m-%d'` as indicated in `$OLDAGE` it works fine in OSX but doesn't in Linux.

Please refer to this section here:
https://github.com/toniblyx/prowler/blob/master/include/os_detector#L15
to see how to implement that particular option for multiple Operating Systems.

Thanks!

gabrielsoltz:
You totally right. 

I added a function for Mac. What do u think ? If my approach is correct, i'll add for the other systems too (need to verify them first).

toniblyx:
What is the use case for having these checks as security check? I kind of understand that having that old instances running might be an issue, but not necessarily a security issue? Can you please provide more context? (I find them useful though, just asking in terms of security)

gabrielsoltz:
Hi, the most serious security risk regarding having old instances running is regarding patch management or software and operating system vulnerabilities. The thing is that is not possible to check software status using AWS API (I'm working on a new check for trying to do this using AWS Inspector). But is not the only possible issue, regarding availability, it could be also an issue: having so old instances is opposite as the idea of [Pets vs Cattle](https://medium.com/@Joachim8675309/devops-concepts-pets-vs-cattle-2380b5aab313) or Chaos Engineering. Again, is not possible to confirm just by getting instances age, but i think it is a very useful check for getting visibility. Maybe output should be textInfo instead of textPass for doing this check more generic.

toniblyx (changes requested):
Function `get_date_previous_than_months` needs to be added also for linux and cygwin, you can use this code (I have tested it already):

```
get_date_previous_than_months()       
{         
MONTHS_TO_COMPARE=$1;         
DATE_BEFORE_MONTHS_TO_COMPARE=$(date --date="$(echo $MONTHS_TO_COMPARE) months ago" '+%Y-%m-%d');         
echo $DATE_BEFORE_MONTHS_TO_COMPARE; 
}
```

Thanks

gabrielsoltz:
@toniblyx i improved your code, coz it was failing in date with busybox implementations (like alpine). 

Already tested with Ubuntu, Amazon Linux, Mac and Alpine. Working fine.

toniblyx (approved):
Awesome, thanks @gabrielsoltz

toniblyx:
I had to add $PROFILE_OPT to the commands to make sure it takes other profile when `-p` is used. I also removed `INFO! Looking for EC2 instances in a particular region` for redundancy. Everything else is fine, thank you.

PR #333, opened 2019-06-06
review outputs

gabrielsoltz:
Output Should be Fail for never used AccessKey. right ?


toniblyx (approved):
I think this makes sense to be a FAIL instead of info, having credentials never used are not a good idea. CIS benchmark information is not very clear here.

PR #334, opened 2019-06-06
add guardduty regions

gabrielsoltz:
Guardduty Regions are not the same as Regions. 

The thing is that is not possible to get Guardduty Regions using CLI, so i raised an issue here: https://github.com/aws/aws-cli/issues/4219

Meanwhile, we can use boto3, so i created a simple function to use boto3 if python is present. 

What do u think ?


toniblyx (approved):
I like the trick, I think it probably can be used for other services as well. Not sure if all users will like to include python bits into the code though, but it is helpful.

PR #339, opened 2019-06-13
refactor check_extra734

gabrielsoltz:
Ok, i finally ended refactoring this check. 

I think now if more performant, because we don't continue if we already know encryption is disabled by getting `ServerSideEncryptionConfigurationNotFoundError`. 

Also added a check for `AccessDenied` for cases when for example Bucket have an explicit Deny on it. 

What do u think ?

### DIFF
--- a/checks/check_extra734
+++ b/checks/check_extra734
@@ -21,34 +21,33 @@ extra734(){
   if [[ $LIST_OF_BUCKETS ]]; then
     for bucket in $LIST_OF_BUCKETS;do
       # query to get if has encryption enabled or not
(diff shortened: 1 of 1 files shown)


toniblyx (approved):
LGTM thanks!

toniblyx (commented):
I had to change `textFail "Bucket does not enforce encryption!" "$bucket"` and `textFail "Access Denied Trying to Get Encryption!" "$bucket"` to `textFail "Bucket $bucket does not enforce encryption!"` and `textFail "Access Denied Trying to Get Encryption for $bucket` to make the output consistent, yours doesn't show bucket name.

PR #340, opened 2019-06-18
ignore None when user data is empty (gunzip: invalid magic)

gabrielsoltz:
When user data is empty, output for `describe-instance-attribute --attribute userData --query UserData.Value` is "None".

If not excluding None from output, you get inconsistent data, ending in code trying to gunzip something because is not base64 encoded as ASCII:

`gunzip: invalid magic`

### DIFF
--- a/checks/check_extra741
+++ b/checks/check_extra741
@@ -23,8 +23,9 @@ extra741(){
     if [[ $LIST_OF_EC2_INSTANCES ]];then
       for instance in $LIST_OF_EC2_INSTANCES; do
         USERDATA_FILE=$instance-userdata.decoded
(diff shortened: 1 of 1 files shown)


toniblyx (approved):

toniblyx:
@gabrielsoltz I had to revert it, I realized it changes more than just fixing gunzip: invalid magic. As in other PR you made, output doesn't work as expected, it is probably OSX related (if works fine in Linux). Please, feel free to PR a fix for the `gunzip:invalid magic` issue only. Also, in your changes you are leaving trace of what is found, I don't like that approach since it is not secure. Thanks!

PR #347, opened 2019-07-05
Add ports to output

gabrielsoltz:


toniblyx:
Hi @gabrielsoltz what is the point of adding ports here? The check is to show if no filters are used. What am I missing? 

Additionally, better not to add : and () in the output, it may break some json outputs/integrations. 

Thanks!

gabrielsoltz:
Hi @toniblyx, i think the check name is kind of confusing, what the check is doing is finding by security group, ports with no filter, this is ports (one or more) with 0.0.0.0/0 as source. So, because this could be one or more, my idea was to add which ports are the ones that are open/without filter, so it's easier to understand the problem, and in our case, we are doing post analyze like top ports being use open/without filter, to undercover or find potential services out of our scope.

gabrielsoltz:
json has not limitation in terms of : and (, but i can replace ithem for a space or dash, what do u think ?

gabrielsoltz:
As i see you are using `:` in most of the outputs for specifying region.

toniblyx:
hey @gabrielsoltz I'll close this PR without merging since the output might be too verbose when many ports are not filtered, better to allow the user check them when something is found. In any case, another custom check can be written if needed. Thanks anyway!

Issue #349, opened 2019-07-09
check314 capitalization disagreement


ralphrodkey:
I got an unexpected failure in check314. Upon investigation, I realized that Prowler is looking for the string "VPC", but the CIS docs specify "Vpc". The case mismatch causes the error. I made a local change to the Prowler code to be case insensitive, and that resulted in the test passing as expected.

All I really did was change the end of line https://github.com/toniblyx/prowler/blob/master/checks/check314#L27 to `/VPC/i`

toniblyx:
I made a comment in your PR

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

toniblyx:
this should be fixed already in PR #366. Please, check it out and re open it with details if needed. Thanks

PR #350, opened 2019-07-09
Made check314 less case sensitive

ralphrodkey:
Resolving issue #349


toniblyx:
Thanks @ralphrodkey. That i makes it case insensitive right? In what OS have you tested that change? 

ralphrodkey:
I ran into the problem on OSX. After some reading about the different flavors of AWK implementations, it looks like OSX uses nawk, while linux machines tend to use gawk or mawk. Most support `IGNORECASE`, but nawk apparently doesn't.

I wanted to make sure I gave you a good answer, so I tested a few combinations:
`echo "Vpc" | awk 'BEGIN {IGNORECASE=1}; /VPC/;'`
Works on Ubuntu (gawk), doesn't work on OSX (nawk)

`echo "Vpc" | awk 'BEGIN {IGNORECASE=1}; /VPC/i;'`
Works on Ubuntu (gawk), works on OSX (nawk)

`echo "Vpc" | awk '/VPC/i;'`
Works on Ubuntu (gawk), works on OSX (nawk)

toniblyx (approved):
LGTM

PR #352, opened 2019-07-10
Ability to exclude check(s) from group run

### LINKED ISSUE #351: Running a group with -E should exclude that check from the run
Right now prowler ignores the `-g` argument and continues to run every check excluding the check specified in the `-E` argument

MartiUK:
Fixes #351 whilst retaining the original functionality.

Tested with and without `-g`, with and without `-E`.


toniblyx (approved):
LGTM

PR #357, opened 2019-07-13
Ease installation - create Pipfile for dependency management

schosterbarak:
centralise Prowler dependencies under a Pipfile:
boto3
ansi2html
detect-secrets


toniblyx (approved):
Good idea and addition, thanks @schosterbarak

PR #362, opened 2019-08-07
[FIX] allow 1.22 checks on policies with only one statement block 

ghost:
We encountered an issue with false negatives on policies that only have one statement. 
PolicyVersion.Document.Statement[?Action!=null] returns 'None' when the Statement is not an array. This is not an empty string and therefore gets falsely recognised.

With this solution we force the it into an Array and deconstructing it afterwards. The result will always then always be an array even if it contains only one element.

This should also solve https://github.com/toniblyx/prowler/pull/329 and possible be contained in other checks also.


ghost:
Is https://github.com/toniblyx/prowler/issues/287 related?

toniblyx (approved):
LGTM thanks!

PR #363, opened 2019-08-08
Stop colorizing the JSON output

ghost:
If using a terminal then jq prints out JSON with color.
I suggest color should either be disabled always or with some other flag (more complicated)
jq flag: -M monochrome (don't colorize JSON);

### DIFF
--- a/include/outputs
+++ b/include/outputs
@@ -27,7 +27,7 @@ textPass(){
     else
       REPREGION=$REGION
     fi
-    jq -c \
+    jq -M -c \
     --arg PROFILE "$PROFILE" \
     --arg ACCOUNT_NUM "$ACCOUNT_NUM" \
     --arg TITLE_TEXT "$TITLE_TEXT" \


toniblyx:
Thanks @james-portman-contino, what is the use case for this? If you process the json output, what is the point is having it with or without color? 

ghost:
It is not actually possible to process it (with code/machine) with colored output, so it is only human readable at that point.
I think the default should be for it to actually output as valid JSON (not colored).

If someone wants it colored then they could pipe it into jq with color enabled, or whatever else they prefer

toniblyx:
That’s weird because I know about so many people using it that way with splunk, ELK, etc. I also integrated it with Wazuh using default json output. In any case having json output colored or not is not important so let’s merge it and if any other user complaints we can always revert it easy. 

ghost:
Thanks, the bit I might have missed is that jq colors the output depending on whether you have a terminal or not, in my case it was docker-compose making a terminal by default, which in turn made jq turn color on.
I think jq should really change its default to not color but that might take longer to fix..

PR #365, opened 2019-08-15
Allow check 1.20 to evaluate users, groups or roles

rjnienaber:
In our setup, developers belong to a specific group so it makes sense to attach the `AWSSupportAccess` policy to this group. This fails the check as it is explicitly filtering by role.

In v1.2 of `CIS Amazon Web Services Foundations` pdf, it states:

```
Check if the 'AWSSupportAccess' is attached to any IAM user, group or role:

aws iam list-entities-for-policy --policy-arn <iam_policy_arn>
```

The command given in the pdf doesn't have this filter and explicitly mentions that it can also be attached to a user or group. This change removes the filter by role and ensures that just the name of the AWS entity is returned.


toniblyx (approved):
LGTM

Issue #325, opened 2019-05-08
All entries in Monitoring, Group3 incorrectly reporting as failed


dariushazimi:
I have created all the group3 metric as per instruction in the CIS benchmark and validated, however, prowler incorrectly reports failure.

```
 3.1  [check31] Ensure a log metric filter and alarm exist for unauthorized API calls (Scored)
       FAIL! No CloudWatch group found for CloudTrail events 

 3.2  [check32] Ensure a log metric filter and alarm exist for Management Console sign-in without MFA (Scored)
       FAIL! No CloudWatch group found for CloudTrail events 

 3.3  [check33] Ensure a log metric filter and alarm exist for usage of root account (Scored)
       FAIL! No CloudWatch group found for CloudTrail events 

 3.4  [check34] Ensure a log metric filter and alarm exist for IAM policy changes (Scored)
       FAIL! No CloudWatch group found for CloudTrail events 

 3.5  [check35] Ensure a log metric filter and alarm exist for CloudTrail configuration changes (Scored)
       FAIL! No CloudWatch group found for CloudTrail events 

 3.6  [check36] Ensure a log metric filter and alarm exist for AWS Management Console authentication failures (Scored)
       FAIL! No CloudWatch group found for CloudTrail events 

 3.7  [check37] Ensure a log metric filter and alarm exist for disabling or scheduled deletion of customer created CMKs (Scored)
       FAIL! No CloudWatch group found for CloudTrail events 

 3.8  [check38] Ensure a log metric filter and alarm exist for S3 bucket policy changes (Scored)
       FAIL! No CloudWatch group found for CloudTrail events 

 3.9  [check39] Ensure a log metric filter and alarm exist for AWS Config configuration changes (Scored)
       FAIL! No CloudWatch group found for CloudTrail events 

 3.10  [check310] Ensure a log metric filter and alarm exist for security group changes (Scored)
       FAIL! No CloudWatch group found for CloudTrail events 

 3.11  [check311] Ensure a log metric filter and alarm exist for changes to Network Access Control Lists (NACL) (Scored)
       FAIL! No CloudWatch group found for CloudTrail events 

 3.12  [check312] Ensure a log metric filter and alarm exist for changes to network gateways (Scored)
       FAIL! No CloudWatch group found for CloudTrail events 

 3.13  [check313] Ensure a log metric filter and alarm exist for route table changes (Scored)
       FAIL! No CloudWatch group found for CloudTrail events 

[...]

Issue #358, opened 2019-07-19
Check34 failing to find alarm


museadmin:
Hi Toni,

I'm having an issue with check34. On line 26 you call describe alarms like so:

`HAS_ALARM_ASSOCIATED=$($AWSCLI cloudwatch describe-alarms $PROFILE_OPT --region $CLOUDWATCH_LOGGROUP_REGION --query 'MetricAlarms[].MetricName' --output text |     awk 'BEGIN {IGNORECASE=1}; /DeletePolicy/ || /DeletePolicies/ || /Policies/ || /Policy/;')`

The aws half of this line returns the names of my alarms as a single space separated string:

`aws cloudwatch describe-alarms --region eu-west-2 --query 'MetricAlarms[].MetricName' --output text`

which includes the name of my alarm for IAM policy changes - iam_policy_change

However the second half of the command uses awk to delete the string because it matches the word policy amongst others...

`awk 'BEGIN {IGNORECASE=1}; /DeletePolicy/ || /DeletePolicies/ || /Policies/ || /Policy/;`

Resulting in the false positive for no alarm being raised:

`++ aws cloudwatch describe-alarms --region eu-west-2 --query 'MetricAlarms[].MetricName' --output text
++ awk 'BEGIN {IGNORECASE=1}; /DeletePolicy/ || /DeletePolicies/ || /Policies/ || /Policy/;'
+ HAS_ALARM_ASSOCIATED=
+ [[ -n '' ]]
+ textFail 'CloudWatch group CloudTrail/DefaultLogGroup found with metric filters but no alarms associated'`

For what it's worth, this jq filter seems to do the trick:

[...]

PR #366, opened 2019-08-16
Make 3.x tests simpler and more useful

cloudlena:
This makes the 3.x tests not rely on the user naming their alarms in a certain way (which is not even provided by the CIS whitepaper so it's hard to get the names right). Furthermore, it makes the code much more maintainable and streamlined. As a third advantage, it makes the checks output more user friendly and unified across all 3.x checks.

Where the regexes for the metrics filters would have gotten too long, I just took the first two and the last two elements to keep them maintainable and prevent false positives where the check would fail.

### DIFF
--- a/checks/check31
+++ b/checks/check31
@@ -15,49 +15,5 @@ CHECK_TYPE_check31="LEVEL1"
 CHECK_ALTERNATE_check301="check31"
 
 check31(){
-  # "Ensure a log metric filter and alarm exist for unauthorized API calls (Scored)"
-  CLOUDWATCH_GROUP=$($AWSCLI cloudtrail describe-trails $PROFILE_OPT --region $REGION --query 'trailList[*].CloudWatchLogsLogGroupArn' --output text| tr '	' '
-' | awk -F: '{ print $7 }')
-  if [[ $CLOUDWATCH_GROUP ]];then
-    for group in $CLOUDWATCH_GROUP; do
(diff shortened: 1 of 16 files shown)


cloudlena:
Should fix #325, #349 and #358
Should make #350 obsolete

toniblyx (changes requested):
This is a great improvement @mastertinner, and has been a pain for long, thanks! I think it is better if we move `check3x` function file into the `includes` folder to make it more consistent with other functions and leave `checks` folder for checks only. 

it has to be added a line to `prowler` file like this:
`. $PROWLER_DIR/include/check3x`

What do you think, does it make sense? 

Thanks again.

cloudlena:
Thanks for your feedback, @toniblyx. What you wrote makes sense. I changed the files accordingly and rebased on the latest `master` branch.

toniblyx (changes requested):
Thanks @mastertinner, sorry to bother again, I didn't explain myself very good before, I meant to add `. "$PROWLER_DIR/include/check3x"` into the actual `prowler` file just once instead of adding it in every single check, it is just cleaner to add it into here fore example: 
https://github.com/toniblyx/prowler/blob/master/prowler#L168

cloudlena:
Done 😃

toniblyx (approved):
Amazing work @mastertinner, thank you very much, I hope users can take advantage of it pretty soon!

cloudlena:
Great. Thanks for all your feedback and for responding so quickly!

Issue #380, opened 2019-09-11
If prowler doesn't have permissions to run a check (e.g. in cross-account setups) it prints `FAIL`. Should print `WARN` instead.


nicdoye:
Running Check 2.6 in a setup where the CloudTrail S3 bucket is in a different account.

```
 2.6  [check26] Ensure S3 bucket access logging is enabled on the CloudTrail S3 bucket (Scored)
       FAIL! access logging is not enabled in aws-controltower-logs-XXXXXXXXXXXX-us-east-2 CloudTrail S3 bucket!
```

The bucket does have access logging enabled, but, because the bucket is in a different account, we don't have permissions to find this out.

From an Account Factory account (YYYYYYYYYYYY) where I ran prowler:
```
➜ aws s3api get-bucket-logging --bucket aws-controltower-logs-XXXXXXXXXXXX-us-east-2

An error occurred (AccessDenied) when calling the GetBucketLogging operation: Access Denied

➜ echo $?
255
```

From Control Tower Log Archive account (XXXXXXXXXXXX):
```
➜ aws s3api get-bucket-logging --bucket aws-controltower-logs-XXXXXXXXXXXX-us-east-2
{
    "LoggingEnabled": {
        "TargetBucket": "aws-controltower-s3-access-logs-XXXXXXXXXXXX-us-east-2",
        "TargetPrefix": ""
    }
}

➜ echo $?
0
```

toniblyx:
I found that check23 may have the same problem with cross accounts configurations like with Control Tower

toniblyx:
this was fixed on PR #384

PR #384, opened 2019-09-16
fixing #383 and #380

### LINKED ISSUE #383: check3x script error when getting  cloudtrail metrics from a group which is in another account
if we use aws organisations and for example create two accounts

master
dev

and if we configure dev to log cloud trails into dev account and configure master to collect logs from all accounts(with organisation cloud trail option enabled)  (i.e dev will log its logs to dev account and also to master account)

`CLOUDWATCH_GROUP=$($AWSCLI cloudtrail describe-trails $PROFILE_OPT --region "$REGION" --query 'trailList[*].CloudWatchLogsLogGroupArn' --output text| tr '\011' '\012' | awk -F: '{print $7}')`

will return 

```
arn:aws:logs:eu-west-2:xxxxxxxxxxxx:log-group:test-cwlg-org-cloudtrail:* 
arn:aws:logs:eu-west-2:yyyyyyyyyyyy:log-group:test-dev-cwlg-cloudtrail:*
```

then  

`METRICFILTER_SET=$($AWSCLI logs describe-metric-filters --log-group-name "$group" $PROFILE_OPT --region "$CLOUDWATCH_LOGGROUP_REGION" --output text | grep METRICFILTERS | grep -E "$grep_filter" | awk '{ print $3 }')`

will return `An error occurred (ResourceNotFoundException) when calling the DescribeMetricFilters operation: The specified log group does not exist.` for log group `test-cwlg-org-cloudtrail`

 because the AWS CLI with supplied credentials look for `test-cwlg-org-cloudtrail` in current account(i.e dev account with account id: xxxxxxxxxxxx) but `test-cwlg-org-cloudtrail`  is in master account with account id `yyyyyyyyyyyy` .. 

Due to this prowler will raise an error `CloudWatch group test-cwlg-org-cloudtrail found but no metric filters or alarms associated` 

To fix this issue we may need to provide a assume role to AWS CLI if it is checking another accounts but it would messy..

We need to ignore cloud watch groups that are not associated with current account....

venky999:
fixed #383 and #380

### DIFF
--- a/checks/check26
+++ b/checks/check26
@@ -16,17 +16,38 @@ CHECK_ALTERNATE_check206="check26"
 
 check26(){
   # "Ensure S3 bucket access logging is enabled on the CloudTrail S3 bucket (Scored)"
-  CLOUDTRAILBUCKET=$($AWSCLI cloudtrail describe-trails --query 'trailList[*].S3BucketName' --output text $PROFILE_OPT --region $REGION)
-    if [[ $CLOUDTRAILBUCKET ]];then
-      for bucket in $CLOUDTRAILBUCKET;do
-        CLOUDTRAILBUCKET_LOGENABLED=$($AWSCLI s3api get-bucket-logging --bucket $bucket $PROFILE_OPT --region $REGION --query 'LoggingEnabled.TargetBucket' --output text|grep -v None)
+  local CHECK_OK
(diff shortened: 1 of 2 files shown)


toniblyx (approved):
LGTM

PR #390, opened 2019-10-02
Add missing permission

roman-vynar:
The check `extra713` requires guardduty:GetDetector IAM permission.


toniblyx (approved):
Thanks, it is all right to have that fixed here but remember that SecurityAudit policy includes `"guardduty:Get*"` and `"guardduty:List*",`.

PR #391, opened 2019-10-06
add extra761 EBS default encryption

jcaffet:
As it is now possible to create EBS volumes encrypted by default (https://aws.amazon.com/fr/blogs/aws/new-opt-in-to-default-encryption-for-new-ebs-volumes/), let's check if it is activated.

### DIFF
--- a/checks/check_extra761
+++ b/checks/check_extra761
@@ -19,7 +19,7 @@ CHECK_ALTERNATE_check761="extra761"
 extra761(){
   textInfo "Looking for EBS Default Encryption activation in all regions...  "
   for regx in $REGIONS; do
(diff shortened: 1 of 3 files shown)


toniblyx (changes requested):
Thanks for this new check @jcaffet. Please, to keep consistency across all checks could you add `$PROFILE_OPT --region $regx` to your code for each AWS-CLI command?

Thanks

jcaffet:
Hi @toniblyx, I have added the parameters as requested. I did not squash my commits. Do you want me to do it ?
Regards.

toniblyx (approved):
Ok, thanks for the change. This feature is region based so much better to check all regions.

Thanks.

PR #392, opened 2019-10-12
WIP: MegaProwler Add-on

MrSecure:
Added 4 files to the "util" directory to use prowler via CodeBuild, with configuration via CloudFormation's UI, and support for auditing all accounts in multiple organizations.

Results are delivered to an S3 bucket, as CSV files, in a YYYY/MM/DD/<stamp>-<org>-<accountid>-prowler-<config>.csv structure, which should align well with usage in Athena.

* Audit_Exec_Role.yaml - CloudFormation definition of an audit execution role which can be assumed by a "master auditor" account. This can be deployed via StackSets to apply to all accounts in an organization.

* Audit_Pipeline.yaml - CloudFormation definitions for an S3 bucket to store results, a CodeBuild project to run prowler, and other resources to handle triggering the CodeBuild project and storing its configuration

* config - bash script which transforms the data present in the CodeBuild environment into the variables/formats megaprowler.sh expects.  This can be modified to allow running megaprowler.sh from an EC2 instance or any host with the AWS CLI and appropriate credentials.

* megaprowler.sh - wrapper script that visits organization master accounts & enumerates member accounts to build a list of accounts which are to be audited, then runs the set of audit jobs in parallel and collects the outputs in a structured manner.


toniblyx:
Wow! this is amazing, let me review it and let you know. Thanks!

toniblyx (changes requested):
Can you put this files in a folder called multi-account inside utils? We will need also a README.md there with a how to deploy it and all requirements, steps, etc. Code is pretty self-explanatory but takes time to read all of it to deploy it knowing what it does.

I also see some dependence checks in `megaprowler.sh` that could be implemented in prowler instead and simplify the other script a bit better. WDYT?

Thanks for putting this together @MrSecure!!

toniblyx:
btw, I don't understand well why parallel is used here.

MrSecure:
> btw, I don't understand well why parallel is used here.

Testing was done with over 100 accounts ... firing them all off in the background ``` prowler ... & ``` saturates the system.  Using parallel gives control around how many jobs are active at once. Right now, parallel will run <number of CPUs>+4 jobs in parallel, and keep that many running. As luck would have it, it's included in the standard codebuild image.

MrSecure:
> Can you put this files in a folder called multi-account inside utils? We will need also a README.md there with a how to deploy it and all requirements, steps, etc. Code is pretty self-explanatory but takes time to read all of it to deploy it knowing what it does.
> 

Certainly can do that.  And the lack of a README is the reason this was published as WIP ;-) 

> I also see some dependence checks in `megaprowler.sh` that could be implemented in prowler instead and simplify the other script a bit better. WDYT?
> 

The `megaprowler.sh` script uses `aws` & `jq` directly, so I'd prefer to keep the checks. That said, adding those to prowler itself would be worthwhile, in a different PR.

toniblyx:
Thanks for clarifying!

PR #397, opened 2019-10-15
fixing multiple exclusions overriding each other because of iteration

### LINKED ISSUE #351: Running a group with -E should exclude that check from the run

ricoli:
Fixes https://github.com/toniblyx/prowler/issues/351


toniblyx (approved):
LGTM Thanks!

ricoli:
no worries - it would be great to get https://github.com/toniblyx/prowler/pull/396 as well, as to have recent changes actually propagate to the docker images :)

PR #399, opened 2019-10-21
add check for unsupported lambda runtimes

MrSecure:
check lambda function runtimes for deprecated configurations based on the list published at https://docs.aws.amazon.com/lambda/latest/dg/runtime-support-policy.html

Include in the "extras" check group.

### DIFF
--- a/checks/check_extra762
+++ b/checks/check_extra762
@@ -32,7 +32,7 @@ extra762(){
         runtime=$(echo "$lambdafunction" | cut -d'%' -f2)
         if echo "$lambdafunction" | grep -Eq $OBSOLETE  ; then
           # obselete runtime found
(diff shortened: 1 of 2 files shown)


toniblyx (changes requested):
This check is pretty helpful, thanks @MrSecure! For `textFail` and `textPass`, shouldn´t we add the region as well? like:

`textFail "$regx: Obsolete ${runtime} runtime found in Lambda funtion ${fname}" "$regx"`

That's just to be consistent with other multi-region services checked by extras.

MrSecure:
added region to output text.  I guess I use the region column in the CSV output too much ;-)

toniblyx:
haha I knew you did it that way bc you use mostly CSV :) 

Thanks for making that change!

PR #400, opened 2019-10-22
extra 7.62 - output cleanup

MrSecure:
- remove warnings about long execution (only one list-functions call per region is needed)
- update pass/fail text to help split on ':' for CSV post-processing


toniblyx (approved):
LGTM

PR #401, opened 2019-10-23
Smarter extra73 (S3 Public Buckets)

gabrielsoltz:
I completely rethought this check, we were missing a lot of possible variables combinations. 

Improvements:
- Now check it's able to identify public-access-block configuration (ALL or by ACL and Policy). See https://docs.aws.amazon.com/cli/latest/reference/s3api/get-public-access-block.html
- Before checking Bucket Policy, let's first use `get-bucket-policy-status`
- Completely replaced awk search for *, with jq, now is more powerful, and we can analyze each statement individually
- Now we can identify conditions by statement
- Fail now is one by bucket, could be a combination of fails or just one, but it will show only one fail line, so only one count for metrics (before we were counting more than one fail by bucket if the case)
- Output with exact details of what we found

What do u think @toniblyx ?


toniblyx:
Tahnks @gabrielsoltz, can you please share differences in terms of accuracy and speed? Do you have that data to share please?

gabrielsoltz:
Hi @toniblyx, we did this because of the number of false positives we were finding, right now, false positives became 0. (Tested across more than 4k S3 buckets).
Regarding speed, it became slower, and we are doing more API calls than before: I'm working to improve that, I'll share with you code after getting an improvement.
(Same answer for the other PR)

gabrielsoltz:
I reduced 4 API calls for this check.

toniblyx:
LGTM. Thanks!

PR #403, opened 2019-10-24
Check extra727 smarter (SQS)

gabrielsoltz:
- Using jq to identify each policy statement
- Conditions are now by statement
- Output specify who can do what
- Identifying policies with and without conditions, different output

### DIFF
--- a/checks/check_extra727
+++ b/checks/check_extra727
@@ -47,36 +47,4 @@ extra727(){
       textInfo "$regx: No SQS queues found" "$regx"
     fi
   done
-
-
-
-
-
-
-
-
-
-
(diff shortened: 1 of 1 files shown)


toniblyx:
I'm not sure if it is good to add that much dependency on `jq` in Prowler but I guess it is needed for more accurate results. In your experience, how different are the results?

gabrielsoltz:
I reduced 1 API call for this check.

gabrielsoltz:
fewer api calls.

toniblyx:
LGTM, thanks @gabrielsoltz, good job!

PR #405, opened 2019-10-28
Check if S3 object-level logging is enabled globally

deppy:


deppy:
I'll have to think a bit more about "extra use cases". This seems to break if I have it globally enabled and also individual buckets because the output can also be

`arn:aws:s3:::myFakeBucketName1354/	arn:aws:s3`
`arn:aws:s3` also seems acceptable

toniblyx:
Hey @deppy, so does your code fix the issue with global and individual as well? It is what it looks like looking at your code but I want to make sure. Thanks!

deppy:
@toniblyx It does fix it, but there are edge cases that it doesn't catch and I hadn't had a chance to look at the proper regex to catch them.

This PR only catches the case when it sees `arn:aws:s3:::` which is how we have it configured, but it looks like you can also have 2 other situations:

1) `arn:aws:s3:::myFakeBucketName1354/ arn:aws:s3` (individual bucket configured + global)
2) `arn:aws:s3` configured via the web console

toniblyx:
I'm going to close this PR since it is not tested enough.

PR #407, opened 2019-11-14
Misc prowler fixes

zfLQ2qx2:
- Add GetEbsEncryptionByDefault wherever Prowler policies are mentioned
- Update Extra718 check to be aware of access denied responses
- Update Extra726 check to be more verbose for non-failure items
- Update Extra73 check to be aware of access denied responses
- Update Extra734 check to be aware of access denied responses and parse policies with jq for better accuracy
- Update Extra742 check for verbiage
- Update Extra756 check for verbiage and parameter order
- Update Extra761 check for failure scenarios (requires most recent awscli and addition to Prowler IAM policy)
- Added Extra763 check to verify that object versioning is enabled on S3 buckets
- Added Extra764 check to verify that S3 buckets enforce a secure transport policy


zfLQ2qx2:
@toniblyx I've opened a new pull request for Prowler

toniblyx:
wow! Thanks for this amazing PR @zfLQ2qx2 I´ll review it as soon as I can. I gave it a quickly review and it looks awesome, thanks!

toniblyx:
Your suggested `extra764` seems to do the same as `extra734`

zfLQ2qx2:
> Your suggested `extra764` seems to do the same as `extra734`

@toniblyx extra764 checks for a secure transport policy (denying S3 requests made over http) and extra734 is checking for server side encryption policy. What I did for both was instead of the awk stuff in the original extra734 I used jq to parse the bucket policy and do a little stronger validation. For example, in the case of check764,  its not enough to have an aws:SecureTransport - false condition, it has to be a Deny policy, for all s3 actions, applying to all principles. Its still just jq - someone could iterate all possible s3 actions individually and while technically correct the check would still report failure, but I think that person can just be unhappy. ;) The check for extra734 is similarly precise but could be toned down to the original just check for the condition if desired.

toniblyx (approved):
Awesome!

jansepke:
@zfLQ2qx2 I have one stupid question to this PR why is extra734 now enforcing an S3 policy for denying every request without a s3:x-amz-server-side-encryption:true header? I cannot find a documentation stating that you need that header if you have default encryption enabled. Wouldn't it be better to make this a separate check or checking for a policy that denies requests WITH a s3:x-amz-server-side-encryption:false header?

PR #417, opened 2019-11-21
Update extra764 and extra734, add .gitignore rules for vim

zfLQ2qx2:
Extra734 - verified with AWS support that S3 server side encryption requires only the ServerSideEncryptionConfiguration rule OR the IAM policy rule but not both.  Changed check to match.  Avoids call to get-bucket-policy if ServerSideEncryptionConfiguration is set.

Extra764 - found reference to the secure transport policy which AWS Config looks for in default s3-bucket-ssl-requests-only rule and updated the jq call to look for the same signature. Provided link to AWS resource in comments in case someone is wondering why the check is written the way it is.

.gitignore - added vim rules from https://github.com/github/gitignore/blob/master/Global/Vim.gitignore 

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,18 @@
+# Swap
+[._]*.s[a-v][a-z]
+[._]*.sw[a-p]
+[._]s[a-rt-v][a-z]
+[._]ss[a-gi-z]
+[._]sw[a-p]
+
+# Session
+Session.vim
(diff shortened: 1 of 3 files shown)


zfLQ2qx2:
@toniblyx I have opened a new pull request for Prowler

toniblyx (approved):
LGTM

PR #424, opened 2019-11-25
Fix extra764 check

willthames:
Add missing bracket to prevent:

```
jq: error: syntax error, unexpected INVALID_CHARACTER, expecting $end (Unix shell quoting issues?) at <top-level>, line 1:
.Statement[]|select(((.Principal|type == "object") and .Principal.AWS == "*") or ((.Principal|type == "string") and
.Principal == "*")) and .Action=="s3:*" and (.Resource|type == "array") and (.Resource|map({(.):0})[]|has($arn)) and
(.Resource|map({(.):0})[]|has($arn+"/*")) and .Condition.Bool."aws:SecureTransport" == "false")
```

(line breaks added to reduce commit width)

### DIFF
--- a/checks/check_extra764
+++ b/checks/check_extra764
@@ -34,7 +34,7 @@ extra764(){
       fi
 
       # https://aws.amazon.com/premiumsupport/knowledge-center/s3-bucket-policy-for-config-rule/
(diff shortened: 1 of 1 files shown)


toniblyx (approved):
LGTM

PR #425, opened 2019-11-26
Make check3x more tolerant

zfLQ2qx2:
I've made further enhancements to make Prowler more tolerant of whitespace in metrics filters (checks 31-39,310-314) as reported by @vsMeecles in https://github.com/toniblyx/prowler/issues/409. I've used jq to squeeze out whitespace in the filter patterns and replaced the tr/awk magic with jq commands. The template vsMeecles provided in that issue now passes.

I also cached responses from the cli instead of calling them repeatedly to parse out individual bits of information. 

### DIFF
--- a/include/check3x
+++ b/include/check3x
@@ -14,30 +14,35 @@ check3x(){
   local CHECK_WARN
   local CHECK_CROSS_ACCOUNT_WARN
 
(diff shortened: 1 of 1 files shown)


toniblyx (approved):
LGTM

PR #429, opened 2019-12-06
Add "access-analyzer:ListTagsForResource" to prowler-additions-policy…

dbellizzi:
….json

check extra769 (Check if IAM Access Analyzer is enabled and its findings) requires this IAM permission

### DIFF
--- a/iam/prowler-additions-policy.json
+++ b/iam/prowler-additions-policy.json
@@ -3,7 +3,7 @@
   "Statement": [
       {
           "Action": [
-              "access-analyzer:ListTagsForResource",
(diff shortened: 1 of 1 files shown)


toniblyx (changes requested):
Thanks @dbellizzi, is that the one only needed? Wouldn't be more comprehensive to add "access-analyzer:List*"?

dbellizzi:
That was the only one that I found that I needed, but I've updated the PR to include `access-analyzer:List*` instead.

toniblyx:
Thanks @dbellizzi

PR #432, opened 2019-12-12
Add trail count to check21 and fail if no trail exist

nimrodkor:
Related to https://github.com/toniblyx/prowler/issues/431

check21 will now fail if no trails were found in the account.

### DIFF
--- a/checks/check21
+++ b/checks/check21
@@ -15,19 +15,25 @@ CHECK_TYPE_check21="LEVEL1"
 CHECK_ALTERNATE_check201="check21"
 
 check21(){
+  trail_count=0
   # "Ensure CloudTrail is enabled in all regions (Scored)"
 	for regx in $REGIONS; do
 		LIST_OF_TRAILS=$($AWSCLI cloudtrail describe-trails $PROFILE_OPT --region $regx --query 'trailList[*].Name' --output text --no-include-shadow-trails)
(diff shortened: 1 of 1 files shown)


toniblyx (changes requested):
Thanks for the heads up here @nimrodkor. Whouldn't be easier to just add an `else` like this?

```
if [[ $LIST_OF_TRAILS ]];then
				for trail in $LIST_OF_TRAILS;do
				MULTIREGION_TRAIL_STATUS=$($AWSCLI cloudtrail describe-trails $PROFILE_OPT --region $regx --query 'trailList[*].IsMultiRegionTrail' --output text --trail-name-list $trail)
				if [[ "$MULTIREGION_TRAIL_STATUS" == 'False' ]];then
						textFail "$trail trail in $regx is not enabled in multi region mode"
				else
						textPass "$trail trail in $regx is enabled for all regions"
				fi
				done
		else
			textFail "No CloudTrail trails were found in the account"
		fi
```

nimrodkor:
@toniblyx I thought so too in the beginning. However, when testing it I discovered the check goes over every region, and searches for trails in that region. Implementing it as you suggested means every region which does not have a multi-region cloudtrail, throws a failure... But one multi-region trail in one region is enough for this check. So I implemented it outside the `for regx in $REGIONS; do` loop block

toniblyx (approved):
LGTM

PR #433, opened 2019-12-12
Added check_extra770, which checks for internet facing instances with an instance profile attached

kmcquade:
@toniblyx - this is what we discussed.

The purpose of this check is to identify public-facing instances that have instance profiles attached. While the presence of an instance profile on a public facing instance is not inherently a security issue, it is definitely worth discovering and triaging, given the higher impact that could come with compromise. For reference, the Capital One situation - a public facing EC2 instance with `s3:*`.

Note: We don't evaluate the actual contents of the policies attached to those instance profiles, as that kind of analysis becomes very robust to the point where it has its own subcategory of tools (see [Parliament](https://github.com/duo-labs/parliament) and [Policy Sentry](https://policy-sentry.readthedocs.io/en/latest/)).

### DIFF
--- a/checks/check_extra770
+++ b/checks/check_extra770
@@ -1,0 +1,35 @@
+#!/usr/bin/env bash
+
+# Prowler - the handy cloud security tool (copyright 2018) by Toni de la Fuente
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may not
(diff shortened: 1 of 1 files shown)


toniblyx:
This check is helpful for incident response.

PR #435, opened 2019-12-15
Fix check26 - get the account ID from sts

nimrodkor:
Related to https://github.com/toniblyx/prowler/issues/434

Added a step which gets the current account ID from STS, to compare with the trail's account ID.


toniblyx (changes requested):
You can use $CURRENT_ACCOUNT_ID or $ACCOUNT_NUM instead, they exist already, at least we reduce a call. Thanks

nimrodkor:
@toniblyx Absolutely! Done.

toniblyx (approved):
LGTM

PR #440, opened 2019-12-18
Small check fixes to extra716 & extra731

nimrodkor:
When running these checks, I noticed:
1. extra716 outputs `No Elasticsearch Service domain found` even if the scan found ES domains.
2. extra731 did not print out the name of the topic that has the bad policy.

This fixes these prints :)


toniblyx (approved):
LGTM

PR #443, opened 2019-12-26
Add error checking to checks extra77 and extra765

zfLQ2qx2:
Check extra765 - Add additional error checking, a recent awscli is required to run this check successfully.

Check extra77 - Add additional error checking, use jq for policy parsing instead of awk

### DIFF
--- a/checks/check_extra765
+++ b/checks/check_extra765
@@ -19,7 +19,6 @@
 #     --region <value> \
 #     --repository-name <value> \
 #     --image-scanning-configuration scanOnPush=true
-  
 
 CHECK_ID_extra765="7.65"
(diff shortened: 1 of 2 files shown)


zfLQ2qx2:
@toniblyx I've raised a pull request

toniblyx (changes requested):
Thanks @zfLQ2qx2, changes for `extra765` look good to me. However for `extra77` I don't agree with L39 since by default, if there is not a policy only the owner can get access so it shouldn't be a FAIL but a PASS. See first paragraph here https://docs.aws.amazon.com/AmazonECR/latest/userguide/repository-policies.html 

> only the repository owner has access to a repository

Do you agree with me that it should be PASS?

zfLQ2qx2:
@toniblyx You are absolutely correct, I was not aware of the default. I've updated the check as you suggested.

toniblyx (approved):
LGTM

PR #446, opened 2019-12-30
Try to make sure prowler cleans up its temporary files

zfLQ2qx2:
I noticed I was accumulating a lot of files in /tmp from prowler runs. This seems to resolve the issue.

### DIFF
--- a/checks/check_extra716
+++ b/checks/check_extra716
@@ -36,10 +36,10 @@ extra716(){
         else
           textPass "$regx: $domain is in a VPC" "$regx"
         fi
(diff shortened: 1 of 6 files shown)


toniblyx (changes requested):
what do you think about this solution here? https://github.com/toniblyx/prowler/blob/master/include/credentials_report#L39

zfLQ2qx2:
@toniblyx I did a quick test yesterday and it didn’t look like an explicit exit or end of script triggered the trap, so treated it as an exception handler and in general tried to clean up files at the source.

PR #447, opened 2019-12-30
Misc fixes to check extra764

zfLQ2qx2:
I double checked this with AWS support -- Principal: {AWS: *} refers to all AWS users and Principal: * refers to everybody both in and out of AWS. So the secure transport policy should only use the later. This follows what AWS does for their check in AWS Config: https://aws.amazon.com/premiumsupport/knowledge-center/s3-bucket-policy-for-config-rule/

While testing this change I found a mistake in the arg to the JQ statement, pretty sure I'm responsible, so I've fixed it.

### DIFF
--- a/checks/check_extra764
+++ b/checks/check_extra764
@@ -34,7 +34,7 @@ extra764(){
       fi
 
       # https://aws.amazon.com/premiumsupport/knowledge-center/s3-bucket-policy-for-config-rule/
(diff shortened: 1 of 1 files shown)


zfLQ2qx2:
@toniblyx I've opened a PR for Prowler

toniblyx (approved):
LGTM

PR #453, opened 2020-01-10
Add Prowler check for unused elastic IP addresses

zfLQ2qx2:
Add Prowler check for unused elastic IP addresses - AWS charges for any EIPs not associated with a network interface


zfLQ2qx2:
@toniblyx I have raised a pull request for Prowler

toniblyx:
Thanks @zfLQ2qx2, other than the cost itself, what are the security implications of having non-used Elastic IP, for example, if we add severity level to the checks (Informational, Low, Medium, High, Critical), what severity would this one be? I think Informational, wdyt?

zfLQ2qx2:
@toniblyx My thoughts are generally that any unused resources unnecessarily enlarge the attack surface and as a best practice should be removed. Since so many people do IP based access control, even though it is definitely an anti-pattern in the cloud, you would release it back into the pool to limit the opportunity that someone could co-op the formerly used address to get elevated access to another resource. Once released back into the pool there is a slim chance of reacquiring it purposefully.

PR #454, opened 2020-01-12
Add command for check119

zfLQ2qx2:
Adds a command for check 1.19 - check that ec2 instances are associated with an instance role. It is always possible something running on the instance is using an IAM key - no way to prevent that just like we can't stop someone from copying instance credentials from another instance - but if there is no instance role associated then I think it is safe to mark a fail. 

### DIFF
--- a/checks/check119
+++ b/checks/check119
@@ -9,8 +9,8 @@
 # work. If not, see <http://creativecommons.org/licenses/by-nc-sa/4.0/>.
 
 CHECK_ID_check119="1.19"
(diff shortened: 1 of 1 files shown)


zfLQ2qx2:
@toniblyx I've raised another PR for Prowler

toniblyx (changes requested):
Thanks @zfLQ2qx2, in this case to be Scored or Not Scored depends on the CIS benchmark itself so I woldn't change that part despite there new code now for that. A new check to look if an instance is using instance metadata service v2 is coming soon, that is also kind of related to this issue.

PR #457, opened 2020-01-20
Issue 163 cloud front waf

ghost:
Add **extra772** to check whether CloudFront is using WAF: https://github.com/toniblyx/prowler/issues/163


toniblyx:
Thanks!

PR #463, opened 2020-01-26
Rewrite of check extra73

zfLQ2qx2:
This is my rewrite of check extre73 to address https://github.com/toniblyx/prowler/issues/458 where I listed some places where I thought the logic was short-circuiting incorrectly. I've also added access denied checks to everything which should help address https://github.com/toniblyx/prowler/issues/459 as well.

The number of subprocesses launched by the current version of this check is pretty intensive so I've looked for opportunities to do less of that. Biggest win is if public access is blocked at the account level in which case we can skip iterating the buckets altogether. If we are using prowler to generate an inventory then maybe that isn't what people want, but seems like a huge win day to day.

### DIFF
--- a/checks/check_extra73
+++ b/checks/check_extra73
@@ -19,46 +19,65 @@ CHECK_ALTERNATE_extra703="extra73"
 CHECK_ALTERNATE_check73="extra73"
 CHECK_ALTERNATE_check703="extra73"
 
-# Improved and simplified check on Nov 18th 2018 due to a new bucket attribute
-# called PolicyStatus, not available in all regions yet.
-
-# extra73(){
-#   ALL_BUCKETS_LIST=$($AWSCLI s3api list-buckets --query 'Buckets[*].{Name:Name}' $PROFILE_OPT --region $REGION --output text)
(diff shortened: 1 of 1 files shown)


zfLQ2qx2:
@toniblyx One more PR for Prowler

toniblyx:
This one is much better now, thanks!

estemendoza:
Hi Guys.

I am experiencing a weird issue with this check. It seems that for some reason, the region value is populated with the name of the bucket, instead of the actual region, I tried today with the latest changes on master.

toniblyx:
Thanks @estemendoza, could you please share the output of this check using this command? `bash -x ./prowler -c extra73` also remember to remove/fake sensitive info please. 

estemendoza:
Ok, I executed that command and didn't want to paste the entire thing, but I think this piece of output shows that the region is taken correctly:

```
++ /usr/bin/aws s3api get-bucket-location --bucket **** --profile **** --output text
+ BUCKET_LOCATION=eu-west-1
++ echo eu-west-1
```

I was thinking that maybe the problem is when these results are saved as CSV? I forgot to say that the results are saved as CSV

PR #467, opened 2020-01-29
Fix check764

orevron:

in case that the results of the get-bucket-policy funcnction fail and returns a non json response avoiding parsing the response as json with jq


checks/check_extra764
```diff
@@ -36,11 +36,14 @@ extra764(){
...
-      CHECK_BUCKET_STP_POLICY_PRESENT=$(cat $TEMP_STP_POLICY_FILE | jq --arg arn "arn:aws:s3:::${bucket}" '.Statement[]|select(((.Principal|type == "string") and .Principal == "*") and .Action=="s3:*" and (.Resource|type == "array") and (.Resource|map({(.):0})[]|has($arn)) and (.Resource|map({(.):0})[]|has($arn+"/*")) and .Condition.Bool."aws:SecureTransport" == "false")')
-      if [[ $CHECK_BUCKET_STP_POLICY_PRESENT ]]; then
-        textPass "Bucket $bucket has S3 bucket policy to deny requests over insecure transport"
-      else
-        textFail "Bucket $bucket allows requests over insecure transport"
+      # checking if $TEMP_STP_POLICY_FILE is a valid json before converting it to json with jq
+      if jq -e . >/dev/null 2>&1 <<< "$TEMP_STP_POLICY_FILE"; then
+        CHECK_BUCKET_STP_POLICY_PRESENT=$(cat $TEMP_STP_POLICY_FILE | jq --arg arn "arn:aws:s3:::${bucket}" '.Statement[]|select(((.Principal|type == "string") and .Principal == "*") and .Action=="s3:*" and (.Resource|type == "array") and (.Resource|map({(.):0})[]|has($arn)) and (.Resource|map({(.):0})[]|has($arn+"/*")) and .Condition.Bool."aws:SecureTransport" == "false")')
```
zfLQ2qx2, line 41: @orevron AWS generally validates json before accepting it, under what condition are you receiving invalid json? Is it an error condition that is not caught above? Assuming that you've submitted invalid json and AWS has passed it back to you, don't you think it would be better to call textInfo rather the silently ignoring it?

orevron: i got it when tried to delete s3 bucket from console but the resource staied as zombie resource (the resource is not available but still exists)
in this case a string error is coming back from this request..

added else statment with text info

toniblyx:
Thanks for your contribution @orevron. @zfLQ2qx2 since you have been enhancing this check lately, could you please let me know if you are ok with it? I can't test this one properly right now. Thanks in advance!

nimrodkor:
Should move over to #474

PR #475, opened 2020-02-06
fix-check11

### LINKED ISSUE #464: Root access keys usage (check11) only shows info
I stumbled onto an account which has active access keys which are in use for the root user (very bad!).
I ran the scan, and found it only returns the following output:
`INFO! Root account last accessed (password key_1 key_2): 2019-12-05T09:34:55+00:00 2020-01-26T13:18:00+00:00 N/A`

Shouldn't it throw an error if there has been use in the last X days? maybe at least if there is use in the last day then throw a failure?

jonnyCodev:

check root account access login and fail if used in the last day
Fixes #464

### DIFF
--- a/checks/check11
+++ b/checks/check11
@@ -16,6 +16,14 @@ CHECK_ALTERNATE_check101="check11"
 
 check11(){
   # "Avoid the use of the root account (Scored)."
-  COMMAND11=$(cat $TEMP_REPORT_FILE| grep '<root_account>' | cut -d, -f5,11,16 | sed 's/,/\ /g')
-  textInfo "Root account last accessed (password key_1 key_2): $COMMAND11"
+  MAX_DAYS=-1
+  last_login_date=$(cat $TEMP_REPORT_FILE|awk -F, '{ print $1,$5 }' |grep '<root_account>' | awk '{ print $2 }')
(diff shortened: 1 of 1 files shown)


toniblyx (changes requested):
Thanks @jonnyCodev this check will make it more accurate. But it fails on OSX

```
1.1  [check11] Avoid the use of the root account (Scored)
Failed conversion of ``no_information'' using format ``%Y-%m-%d''
date: illegal time format
       PASS! Root Account arn:aws:iam::123456789012:root last accessed was more then 1 day ago
```

That error seems to come from line `   days_not_in_use=$(how_many_days_from_today ${last_login_date%T*}) `

jonnyCodev:
> Thanks @jonnyCodev this check will make it more accurate. But it fails on OSX
> 
> ```
> 1.1  [check11] Avoid the use of the root account (Scored)
> Failed conversion of ``no_information'' using format ``%Y-%m-%d''
> date: illegal time format
> usage: date [-jnRu] [-d dst] [-r seconds] [-t west] [-v[+|-]val[ymwdHMS]] ...
>             [-f fmt date | [[[mm]dd]HH]MM[[cc]yy][.ss]] [+format]
>        PASS! Root Account arn:aws:iam::123456789012:root last accessed was more then 1 day ago
> ```
> 
> That error seems to come from line ` days_not_in_use=$(how_many_days_from_today ${last_login_date%T*})`

You are right, thanks.
Fixed the bug with date validation.

PR #479, opened 2020-02-11
Remove `ses:sendemails`

nickmalcolm:
Prowler should only require permissions which are readonly. The additional policy, recommended byt the readme, adds the ability to send emails via SES. There is even a closed issue where it was described as not needed: https://github.com/toniblyx/prowler/issues/124


toniblyx:
You are right, I added it to allow a new feature that I never wrote. Thanks for the heads up @nickmalcolm.

nmalcolmlinz:
That was really fast, thanks!

nickmalcolm:
That was really fast, thanks!

PR #481, opened 2020-02-12
Check29 Associate VPCFlowLog with VPC

alphad05:
Associate VPCFlowLow with the VPC it is for to ensure accurate check. If there are multiple VPCs in a region and only some have VPC flow logs, current check will pass all VPCs even those without VPC flow logs.


toniblyx:
You are right @alphad05, the current check only shows region info and with your enhancement we can see now which VPC has or has not FlowLog enabled. Thanks!

PR #499, opened 2020-02-25
fix: check119 needs to ignore terminated instances

nexeck:
Terminated does not seem to have an instance profile (At least when they are managed by an autoscaling group). And its not
possible to start a terminated instance again.


toniblyx:
Makes sense, thanks!

PR #502, opened 2020-02-28
New check for ecr findings

nexeck:

This will check if there are any images with findings inside ecr.
At the moment it will not check the severities of those findings, i found this list of severities, maybe we could ignore INFORMATIONAL?

- "HIGH"
- "MEDIUM"
- "INFORMATIONAL"
- "LOW"
- "CRITICAL"
- "UNDEFINED"

Is there a way to configure such checks?

### DIFF
--- a/checks/check_extra765
+++ b/checks/check_extra765
@@ -44,7 +44,7 @@ extra765(){
             textFail "$region: ECR repository $repo has scan on push disabled!" "$region"
             ;;
           "None")
-            textInfo "$region: ECR repository $repo hs no scanOnPush status, newer awscli needed" "$region"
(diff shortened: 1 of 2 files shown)


toniblyx:
Thanks @nexeck, it would be better to show severity, even informational with textInfo. Could you add that please? 

nexeck:
Should we show only the highest, or all which were found? @toniblyx

nexeck:
Something like this could also be possible
image

nexeck:
Would also be possible, to show only high, critical, medium as fail, and all others as info, what do you think?
Could look like:
image

toniblyx:
I like it, I think it is pretty helpful and gives good information about what it is.

nexeck:
@toniblyx I have pushed the changes from the last screenshot :-)

PR #509, opened 2020-03-04
fix: Enable check extra776 in extra group

nexeck:

Just saw, that i forgot to add the new check to the extra group :-)


nexeck:
@toniblyx Just missed to enable the new check in the extra group :-)

PR #510, opened 2020-03-04
Improve performance of check_extra742 by limiting to one AWS CLI call per region

jonjozwiak:

In an account with 4000 CloudFormation stacks each API call takes over 1 minute.  This check takes over a day to complete.  I've modified to only call the AWS CLI once per region rather than per stack.  This results in each stack taking only 0.5 seconds and reducing overall run time to around 30 minutes at that scale.


toniblyx:
wow! that was a pretty clever enhancement @jonjozwiak. I bet there are other checks that can be improved the same way. Thanks!

PR #511, opened 2020-03-04
Find trust boundaries in VPC endpoint services

renuez:

It's my first proposal on a series of potential new checks that find trust boundaries in VPC endpoint services (whitelisted principals and connections).

I do also have one for VPC peerings and I currently work on another one for IAM roles. 
I think there are many interesting use cases to compare  a known whitelist of accounts against KMS usage/admin policies, S3 bucket policies or ACLs, ...

Maybe it's only interesting for multi-account setups with density and cohesion of AWS account trusts - what do you think of a dedicated new check group?


toniblyx (changes requested):
Thanks @renuez, let me add here some comments and changes I would suggest to make before merging your code:
- `$ACCOUNT_NUM` is a variable already set that you can use instead of `$THIS_ACCOUNT_ID`.
- Could you comment the code? Some parts are self explanatory but others are less easy to understand, at least for me. That will help others to improve it and create other checks based on these ones.
- It would be also helpful to add information about the use case in the comments as well as the syntax for the whitelist files. Btw, why not to have comma separated values in a variable inside the check file? or an array?

toniblyx (changes requested):
also, each `aws` command requries the variable `$PROFILE_OPT` that includes options if different profile is loaded.

renuez:
@toniblyx Thanks for your recent feedback and code review support.
All amendments build into feature branch now.
Ready for you final review.

toniblyx (approved):
LGTM

toniblyx:
This pull request has been already applied, I had to do a workaround with conflicts. Thanks @renuez

PR #512, opened 2020-03-09
Patch check21

lanhhuyet510:

1. Optimize check: We do not need to check all regions, run `$AWSCLI cloudtrail describe-trails --query 'trailList[*].HomeRegion' --output text` to get list of regions having trails
2. If there's a trail is compliance, the result should be `PASS`. If all trails are not compliance, the result will be `FAIL`
3. Add `ISLOGGING_STATUS`, `INCLUDEMANAGEMENTEVENTS_STATUS`, `READWRITETYPE_STATUS` to complete the check

### DIFF
--- a/checks/check21
+++ b/checks/check21
@@ -9,7 +9,7 @@
 # work. If not, see <http://creativecommons.org/licenses/by-nc-sa/4.0/>.
 
 CHECK_ID_check21="2.1,2.01"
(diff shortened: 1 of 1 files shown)


toniblyx:
Thanks @lanhhuyet510, it looks more comprehensive and faster. I have to give it a specific test since this check has been patched many times

PR #514, opened 2020-03-11
Support whitelists per check

### LINKED ISSUE #154: Support for resources whitelist (leave resources out of checks)
Would be nice to have environment definable suppressions and ignores:

An example would be to ignore `check73` if `bucket` = `my-known-problem-bucket`.

I am envisioning a file based approach, either one file for all suppressions with a config loader, or potentially a file per check so you can rely on conventions for loading your ignore list.

Example:
```text
# File Name: check73.suppressions
suppress=()
suppress+=('my-known-problem-bucket')
suppress+=('another-problem-bucket')
```

Easier examples would be line separated values or comma perhaps. Lots of options but would be nice.

urjitbhatia:

Fixes https://github.com/toniblyx/prowler/issues/154

cc @dbellizzi


toniblyx:
Thanks @urjitbhatia this is a very helpful feature, I will have to test it carefully. Can you please describe tests that you have do with it already? Thanks!

urjitbhatia:
@toniblyx we're running this in our env for a bit. I haven't seen any trouble yet but let me get back to you after we a few more checks and run it for a couple more days

toniblyx:
Cool, thanks!

toniblyx (commented):
Hi @urjitbhatia,

file `include/outputs`: I need a bit more information to understand from line 107 to 129. Why the WARN and the change on line 129.

file `prowler`: Line 174 could cause problems if CSV or json is used, am I right? I don 't think it is needed.

include/outputs
```diff
@@ -104,15 +104,29 @@ textInfo(){
...
 
 textFail(){
-  FAIL_COUNTER=$((FAIL_COUNTER+1))
-  EXITCODE=3
+  ## ignore whitelists for current check
+  level="FAIL"
+  for i in $IGNORES; do
+    ignore_value="${i#*${CHECK_NAME}:}"
+    if [[ $1 =~ ${ignore_value} ]]; then
+      level="WARNING"
```
urjitbhatia, line 98: ```file include/outputs: I need a bit more information to understand from line 107 to 129. Why the WARN and the change on line 129.```
@toniblyx ok let me describe what I was thinking here:

- Added the "WARNING" level so that even if people ignore things in a whitelist, it's still there in the output as a reminder that they might want to fix it later if it's a temporary suppression.
- The loop at line 109 is basically parsing the whitelist line-by-line, splitting the string upto the first occurrence of `$CHECK_NAME:`  so using: ```${var#*SubStr}  # will drop begin of string up to first occur of `SubStr` ```
  - and then checking if that whitelisted item matches the current item raising the failure
  - I could've made a map for this but it's only supported on bash 4 onwards so this keeps it backwards-compat (With some perf cost for the lookup)
 - It we find a match, treat it as a WARNING (And line 118-121 ensures we don't decrement FAIL_COUNTER & EXIT_CODE if we only saw warnings so far)

Line 129 is using the fact that `${level}` already holds FAIL/WARNING as the level so we can just use that to print the right warning

prowler
```diff
@@ -83,12 +84,19 @@ USAGE:
                             (i.e.: ProwlerRole)
       -T                  session durantion given to that role credentials in seconds, default 1h (3600) recommended 12h, requires -R and -T
                             (i.e.: 43200) 
+      -w                  whitelist file. (Lines starting with # are ignored as comments) Format:
```
urjitbhatia, line 89: ```file prowler: Line 174 could cause problems if CSV or json is used, am I right? I don 't think it is needed.```

@toniblyx yeah, I was hoping that this help doc clarifies the structure we're relying on here, unless I am missing something else about the CSV/JSON concern.

urjitbhatia: Overall, what do you think about migrating this project to a python/go etc to help simplify some of the abstractions? Just a thought - you might get more contribs and somewhat easier code - I noticed that the shell code in here has a lot of warnings according to shell-check.
It's def your preference as the project owner - hope this doesn't come across the wrong way :)
The checks themselves could stay in bash because that makes using AWS etc apis much easier but the scaffolding around that could be moved and then we could have easy pre-filter, post-filter hooks etc.

toniblyx:
Hey @urjitbhatia I didn't forget you, I'll work on this review soon. Thanks again!

QuantumBJump:
Hey, this is mostly a readability comment, but this implementation looks more like a blacklist/ignore list than a whitelist? It's performing all checks except those indicated in the file, if I'm reading the code correctly, whereas a whitelist would *only* perform checks if it found them in the file.

Other than that, I'm really looking forward to this feature!

toniblyx:
Thanks for your comment @QuinnStevens, I agree with you, it would be blacklist. Whatever is the case whitelist/blacklist it is a pretty helpful feature: a list to ignore resources per check, something common for buckets. I didn't merge this PR because I'm not 100% convinced on this way to implement it, first it gets rid of it on FAIL only, second it adds a WARNING... Any other ideas? 
I'd like to add this feature to 2.3 this week though.

dbellizzi:
To clarify how we're using it at ClassDojo, this is more of a "suppression list".  These are things that we know fail the check, but we don't want to disable the entire check.  For example, we have certain public facing load balancers, but we don't want people accidentally making new ones that they intended to be private.  So we suppress the failure in the cases that we know about and are fine with.

The goal for us is to get prowler to run with zero failures, so that we can keep that in our monitoring system and alert when anything unexpected does arise.

The choice to output on a suppressed match was just so that we surfaced the fact that it was suppressed.  This could be an INFO or a WARNING, but the goal was for it to not be a FAIL so that the test would pass.

QuantumBJump:
Yeah that's pretty much exactly how we're trying to use Prowler at Finbourne. I made a fork for our purposes, and it worked pretty much exactly like this, but less flexible.

toniblyx:
Good stuff, thanks @QuinnStevens @dbellizzi are you guys doing that at check level or removing those known resources once the report is generated?

urjitbhatia:
@toniblyx yeah, maybe we can rename it to a suppression list - happy to do that.

We are doing the checks continuously and once we get down to WARNINGs only, we will slowly start to improve those resource configurations as well. This lets us get our prowler run reports green with known exceptions.

We are not removing resources for now but our aim to slim down the suppression list and going forward, making sure our team has a good reason to create an exception but otherwise not growing the list.

toniblyx:
My apologies guys, I didn't forget this PR, I just have to invest some time on it and merge it. Definitely it is something that worth it to take time and include it in what is going to be a "full-of-cool-new-features" v2.3.

toniblyx:
Awesome, tested it and went all good. I will make small changes after merge it:
- reduce the text in usage
- add information about this new feature in README.md
- remove `Using Whitelist file: whitelist_sample.txt` message, if the user uses `-w whitelist_sample.txt` they know that already :) we can add that info to below the banner if needed like below Caller Identity ARN.
- remove WARNING resources from the json-asff output, we don't want to send that resource to Security Hub if we don't care its configuration.

Thanks!

PR #519, opened 2020-03-22
check26 - on failure, output info and not failure

nimrodkor:
CloudTrail bucket can be in another account, which will result if failing to get bucket logging.
Therefore, IMO, errors should be handled as `Info` and not as `Fail`.


toniblyx:
thanks @nimrodkor, this is an interesting one, if the bucket is in other account why not to add that to the INFO message. We could differentiate between a FAIL in case Prowler can't read the policy and INFO if that bucket is not in the given account, then can't read the policy. This comment applies also to PR #518 
What do you think?

PR #520, opened 2020-03-22
extra774 requires credential report to run successfully

nimrodkor:
extra774 requires a credential report... This makes it run


toniblyx:
Good catch, thanks @nimrodkor!

PR #522, opened 2020-03-27
Ignore imported ACM Certificate in check_extra724

yumminhuang:
The `UpdateCertificateOptions` command is not permitted for imported certificates, so certificate transparency logging of imported certificates should not be checked. I think it should be enabled by the issuer of imported certificates.

---


toniblyx (approved):
Good improvement, thanks!

PR #527, opened 2020-04-02
Remove --output text in CLOUDTRAILBUCKET_LOGENABLED

yumminhuang:
I am not sure is there anyone has the same issue as mine. If server access log is not enabled for one bucket, the output of following command is `None` instead of `null`.

```
s3api get-bucket-logging --bucket $bucket $PROFILE_OPT --region $REGION --query 'LoggingEnabled.TargetBucket' --output text
```

Probably caused by different version of `aws-cli` or `boto`. Here is my version

```
aws --version
aws-cli/2.0.4 Python/3.8.2 Darwin/19.4.0 botocore/2.0.0dev8
```

---

### DIFF
--- a/checks/check26
+++ b/checks/check26
@@ -27,7 +27,7 @@ check26(){
       if [[ $CLOUDTRAILBUCKET ]]; then
         bucket=$CLOUDTRAILBUCKET
         if [ "$CLOUDTRAIL_ACCOUNT_ID" == "$ACCOUNT_NUM" ]; then
-          CLOUDTRAILBUCKET_LOGENABLED=$($AWSCLI s3api get-bucket-logging --bucket $bucket $PROFILE_OPT --region $REGION --query 'LoggingEnabled.TargetBucket' --output text 2>&1)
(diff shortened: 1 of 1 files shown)


yumminhuang:
In `check_extra718`, the script also checks `None`.

toniblyx:
Thanks @yumminhuang, can you please confirm if you get the same response using `--output text`? With that option it should be `None`, without it should be `null` unless there has been a recent change on that. Can you confirm it please?

yumminhuang:
Thanks @toniblyx . You are correct. I tried aws-cli command and it returned `null` when not adding `--output text`.

Actually, there is `--output text` in `checks/check26`

https://github.com/toniblyx/prowler/blob/b5e1c9002a933c51502116b6fc6a84d475b7b74c/checks/check26#L30

I will update my PR by removing `--output text` and using `null` in if condition.

toniblyx:
Tested with 
```
aws --version
aws-cli/1.18.38 Python/3.7.6 Darwin/18.7.0 botocore/1.15.38
```

PR #529, opened 2020-04-06
Add support for externalID when AssumeRole is used

JBarnes94:

### DIFF
--- a/prowler
+++ b/prowler
@@ -32,7 +32,7 @@ OPTRED="[1;31m"
 OPTNORMAL="[0;39m"
 
 # Set the defaults variables
-PROWLER_VERSION=2.2.1
+PROWLER_VERSION=2.2.2
 PROWLER_DIR=$(dirname "$0")
 
 REGION=""
@@ -83,12 +83,13 @@ USAGE:
                             (i.e.: ProwlerRole)
       -T                  session durantion given to that role credentials in seconds, default 1h (3600) recommended 12h, requires -R and -T
                             (i.e.: 43200) 
+      -I.                 External ID to be used when assuming roles (no mandatory)
       -h                  this help
   "
   exit
 }
 
-while getopts ":hlLkqp:r:c:g:f:m:M:E:enbVsx:A:R:T:" OPTION; do
+while getopts ":hlLkqp:r:c:g:f:m:M:E:enbVsxI:A:R:T:" OPTION; do
    case $OPTION in
      h )
         usage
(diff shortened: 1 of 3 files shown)


prowler
```diff
@@ -441,10 +445,13 @@ if [[ $CHECK_ID ]];then
 fi
 
 if [[ $ACCOUNT_TO_ASSUME ]]; then
-  # unset env variables with assumed role credentials
-  unset AWS_ACCESS_KEY_ID
-  unset AWS_SECRET_ACCESS_KEY
-  unset AWS_SESSION_TOKEN
+      #This would remove permissions if all checks were run using external ID
```
toniblyx: This if `if [[ -z $ROLE_EXTERNAL_ID ]]` would make to not to clean up those variables unless external ID is created so it is not needed.

prowler
```diff
@@ -83,12 +83,13 @@ USAGE:
                             (i.e.: ProwlerRole)
       -T                  session durantion given to that role credentials in seconds, default 1h (3600) recommended 12h, requires -R and -T
                             (i.e.: 43200) 
+      -I.                 External ID to be used when assuming roles (no mandatory)
```
toniblyx: is that dot after `-I` needed?

prowler
```diff
@@ -32,7 +32,7 @@ OPTRED="[1;31m"
 OPTNORMAL="[0;39m"
 
 # Set the defaults variables
-PROWLER_VERSION=2.2.1
+PROWLER_VERSION=2.2.2
```
toniblyx: I will do that later once I add some other changes that are coming.

toniblyx:
Hi @JBarnes94, I added some comments to clarify or fix a couple of small things, thank you!

prowler
JBarnes94: In my testing it was needed if you were not proving the specific group to test against. 
Without it, the credentials that are set in assume_role, are unset before you hit Execute all. So the role assumption works, but the commands do not run as they don't have the required credentials 
This does not happen when you provide a groups of check with -g, as it is executed before the IF. 

Another option would be to leave the If as only referencing [[ $ACCOUNT_TO_ASSUME ]]
but move execute_all and scoring above this so they run with the correct credentials before clean up ?

JBarnes94:
Hey @toniblyx,Thanks for the comments and review. I've Updated and think it should address them. If not let me know.

toniblyx:
Merged already, thanks!

PR #530, opened 2020-04-07
Add 'json-asff' output mode and ability to send output to AWS Security Hub

### LINKED ISSUE #524: Output mode for submitting results directly to AWS Security Hub
Hi @toniblyx 

I came across https://aws.amazon.com/blogs/security/use-aws-fargate-prowler-send-security-configuration-findings-about-aws-services-security-hub/ which achieves my goal of submitting findings to Security Hub, but I would like to find a simpler method if I can.

Would you accept a pull request that added a new output mode, `securityhub`, then when chosen, Prowler would submit each check result directly to AWS Security Hub using the AWS CLI? I envisage the output mode generating a JSON string and then calling: `aws securityhub batch-import-findings --findings <JSON Finding>`.

If this is acceptable, I'll spend the time on a PR and add the optional IAM permission to the README etc.

Appreciate this is somewhat inefficient as the batch API can accept 100 checks at a time, so if you don't think this is a good idea, I'll look to generate a CSV and use a wrapper approach like the Wazuh integration instead.

Many thanks

marcjay:
json-asff mode outputs JSON, similar to the standard 'json' mode with one check per line, but in AWS Security Finding Format - used by AWS Security Hub
Currently uses a generic Type, Resources and ProductArn value, but sets the Id to a unique value that includes the details of the message, in order to separate out checks that run against multiple resources and output one result per resource per check. This ensures that findings can be updated, should the resource move in or out of compliance

securityhub mode generates the ASFF JSON and then passes it to an 'aws securityhub batch-import-findings' call, once per resource per check. Output to the screen is similar to the standard mode, but prints whether or not the finding was submitted successfully

Fixes #524


marcjay:
Putting this PR out early in Draft state to gather feedback. Tasks I believe need addressing:
- [x] Allow each check to set a Type, and Resources value appropriately, for `include/outputs` to pick up and use, default to their current values
- [x] Update README with securityhub mode example
- [x] Make note about IAM permission needed to insert findings, either in README or in extras policy JSON 
- [ ] Obtain a ProductArn for Prowler and drop it in

toniblyx:
awesome PR, thanks @marcjay ! I would like to have a walk through the code with you to ask few questions and probably make some changes here and there. Just sent you a DM in Linkedin

include/colors
```diff
@@ -11,15 +11,15 @@
 # CONDITIONS OF ANY KIND, either express or implied. See the License for the
 # specific language governing permissions and limitations under the License.
 
-if [[ $MODE != "mono" && $MODE != "text" && $MODE != "csv" && $MODE != "json" ]]; then
+if [[ "$MODE" != "mono" && "$MODE" != "text" && "$MODE" != "csv" && "$MODE" != "json" && "$MODE" != "json-asff" && "$MODE" != "securityhub" ]]; then
```
toniblyx: We can add a new option like `-S` (capital S) and move the securityhub integration to another function. Using `json-asff` as output and if `-S` is used then send information to SH.

include/colors
```diff
@@ -11,15 +11,15 @@
...
   echo ""
-  echo "$OPTRED ERROR!$OPTNORMAL Invalid output mode.  Choose text, mono, or csv."
+  echo "$OPTRED ERROR!$OPTNORMAL Invalid output mode.  Choose text, mono, csv, json, json-asff or securityhub."
```
toniblyx: remove securityhub here

include/outputs
```diff
@@ -18,41 +18,22 @@ textPass(){
   fi
 
   PASS_COUNTER=$((PASS_COUNTER+1))
-  if [[ "$MODE" == "csv" ]]; then
+  if [[ "$MODE" == "csv" || "$MODE" == "json" || "$MODE" == "json-asff" || "$MODE" == "securityhub" ]]; then
```
toniblyx: securityhub is not an output now

include/outputs
```diff
@@ -18,41 +18,22 @@ textPass(){
...
-      REPREGION=$2
-    else
-      REPREGION=$REGION
+    if [[ "$MODE" == "csv" ]]; then
+      echo "$PROFILE${SEP}$ACCOUNT_NUM${SEP}$REPREGION${SEP}$TITLE_ID${SEP}PASS${SEP}$ITEM_SCORED${SEP}$ITEM_LEVEL${SEP}$TITLE_TEXT${SEP}$1"
+    elif [[ "$MODE" == "json" ]]; then
+      generateJsonOutput "$1" "Pass"
+    elif [[ "$MODE" == "json-asff" ]]; then
+      generateJsonAsffOutput "$1" "PASSED" "INFORMATIONAL"
+    elif [[ "$MODE" == "securityhub" ]]; then
```
toniblyx: We no longer need this here and for the integration we can use a dedicated function in folder `prowler/include/` file called `securityhub_integration`. 

For the function it self we have to make sure next:
- Region, remember that Prowler gets region from the config file initially or whatever that you specify with `-r`.
- If SH is enabled in that region. 
- For the command we have to add next variables:
`$AWSCLI securityhub --region $REGION $PROFILEOPTS \
batch-import-findings \
--findings "$(generateJsonAsffOutput "$1" "PASSED" "INFORMATIONAL")" | jq -M -r 'if .SuccessCount == 1 then "Successfully submitted finding" else "Failed to upload finding" end'`
- I would move that if outside the `jq`

include/outputs
```diff
@@ -63,41 +44,22 @@ textInfo(){
...
-      REPREGION=$2
-    else
-      REPREGION=$REGION
+    if [[ "$MODE" == "csv" ]]; then
+      echo "$PROFILE${SEP}$ACCOUNT_NUM${SEP}$REPREGION${SEP}$TITLE_ID${SEP}INFO${SEP}$ITEM_SCORED${SEP}$ITEM_LEVEL${SEP}$TITLE_TEXT${SEP}$1"
+    elif [[ "$MODE" == "json" ]]; then
+      generateJsonOutput "$1" "Info"
+    elif [[ "$MODE" == "json-asff" ]]; then
+      generateJsonAsffOutput "$1" "NOT_AVAILABLE" "LOW"
+    elif [[ "$MODE" == "securityhub" ]]; then
```
toniblyx: same as before

include/outputs
```diff
@@ -106,41 +68,22 @@ textInfo(){
 textFail(){
   FAIL_COUNTER=$((FAIL_COUNTER+1))
   EXITCODE=3
-  if [[ "$MODE" == "csv" ]]; then
+  if [[ "$MODE" == "csv" || "$MODE" == "json" || "$MODE" == "json-asff" || "$MODE" == "securityhub" ]]; then
```
toniblyx: same as before

include/outputs
```diff
@@ -106,41 +68,22 @@ textInfo(){
...
-      REPREGION=$REGION
+    if [[ "$MODE" == "csv" ]]; then
+      echo "$PROFILE${SEP}$ACCOUNT_NUM${SEP}$REPREGION${SEP}$TITLE_ID${SEP}FAIL${SEP}$ITEM_SCORED${SEP}$ITEM_LEVEL${SEP}$TITLE_TEXT${SEP}$1"
+    elif [[ "$MODE" == "json" ]]; then
+      generateJsonOutput "$1" "Fail"
+    elif [[ "$MODE" == "json-asff" ]]; then
+      generateJsonAsffOutput "$1" "FAILED" "HIGH"
+    elif [[ "$MODE" == "securityhub" ]]; then
+      printf "      $BAD FAIL! %s... $NORMAL"  "$1"
+      aws securityhub batch-import-findings --findings "$(generateJsonAsffOutput "$1" "FAILED" "HIGH")" | jq -M -r 'if .SuccessCount == 1 then "Successfully submitted finding" else "Failed to upload finding" end'
```
toniblyx: same as before

prowler
```diff
@@ -64,7 +64,7 @@ USAGE:
       -f <filterregion>   specify an AWS region to run checks against
                             (i.e.: us-west-1)
       -m <maxitems>       specify the maximum number of items to return for long-running requests (default: 100)
-      -M <mode>           output mode: text (default), mono, json, csv (separator is ","; data is on stdout; progress on stderr)
+      -M <mode>           output mode: text (default), mono, json, json-asff, securityhub, csv (separator is ","; data is on stdout; progress on stderr)
```
toniblyx: add -S instead of securityhub

prowler
```diff
@@ -395,8 +395,8 @@ if [[ $PRINTGROUPSONLY == "1" ]]; then
   exit $EXITCODE
 fi
 
-# Check that jq is installed for JSON output
-if [[ $MODE == "json" ]]; then
+# Check that jq is installed for JSON outputs
```
toniblyx, line 412: as said before

toniblyx (changes requested):
see comments

include/colors
marcjay: Thanks for the feedback the other day, this has been addressed

include/outputs
marcjay: These points should all be addressed now

checks/check_extra780
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env bash
```
marcjay, line 1: It looks like this file was recently deleted in https://github.com/toniblyx/prowler/commit/8f83da985a7ba129317fa2716583f1181bfc4686 - if the intention is for it to be removed, I'll remove it from my branch. I just wasn't sure because I see a reference to extra780 in both `groups/group7_extras` and `groups/group14_elasticsearch`

toniblyx: It doesn't have to be removed, it was my fault.

Issue #387, opened 2019-09-25
Error evaluating Extra725


theOriginalClauBotas:
Guys have you seen that 
BUCKETS_OBJECT_LOGGING_ENABLED=$($AWSCLI cloudtrail get-event-selectors --trail-name "${TRAIL_NAME}" $PROFILE_OPT --region $TRAIL_REGION --query "EventSelectors[*].DataResources[?Type == \`AWS::S3::Object\`].Values" --output text |xargs -n1 |cut -d: -f 6|sed 's/\///g')

Has to be replaced with
BUCKETS_OBJECT_LOGGING_ENABLED=$($AWSCLI cloudtrail get-event-selectors --trail-name "${TRAIL_NAME}" $PROFILE_OPT --region $TRAIL_REGION --query "EventSelectors[*].DataResources[?Type == \`AWS::S3::Object\`].Values" --output text |xargs -n1 |cut -d: -f 6|sed 's/\///g'|sed 's/\*//g')

Because currently it is removing the / but no the * when listing the buckets

I have the next output when I run extra725
 
 7.25 [extra725] Check if S3 buckets have Object-level logging enabled in CloudTrail (Not Scored) (Not part of CIS benchmark) 
       INFO! Looking for S3 Buckets Object-level logging information in all trails...   
       PASS! us-west-2: S3 bucket BUCKET* has Object-level logging enabled in trail 
       FAIL! us-west-2: S3 bucket BUCKET has Object-level logging disabled

toniblyx:
Thanks, for the heads up. Can you explain it better please? What is your buckets name?

theOriginalClauBotas:
This is the complete output7.25 [extra725] Check if S3 buckets have Object-level logging enabled in CloudTrail (Not Scored) (Not part of CIS benchmark) 
       INFO! Looking for S3 Buckets Object-level logging information in all trails...   
       PASS! us-west-2: S3 bucket cf-templates-10286eovwqo38-us-east-1* has Object-level logging enabled in trail 
       PASS! us-west-2: S3 bucket config-bucket-200575803985* has Object-level logging enabled in trail 
       PASS! us-west-2: S3 bucket gnptrail-gnpctrails3bucket-mcv557m1ickd* has Object-level logging enabled in trail 
       PASS! us-west-2: S3 bucket gnptrail-rarchivelogsbucket-1wv9ba6thy19v* has Object-level logging enabled in trail 
[...]

PR #532, opened 2020-04-08
Extra725 - Support cross account and region cloudtrail

patdowney:
Support cross-region and cross-account object-level cloudtrail logs for S3

* Buckets that log to one or more trails are logged as `PASS!` for each trail they are associated with.
* Buckets that aren't associated with any trails are logged as `FAIL!` once.

```
...
PASS! : S3 bucket bucket-one has Object-level logging enabled in trails: arn:aws:cloudtrail:eu-west-2:123456789012:trail/central-trail
PASS! : S3 bucket bucket-two has Object-level logging enabled in trails: arn:aws:cloudtrail:eu-west-2:9876543210989:trail/trail-two
PASS! : S3 bucket bucket-two has Object-level logging enabled in trails: arn:aws:cloudtrail:eu-west-2:123456789012:trail/central-trail
FAIL! : S3 bucket bucket-three has Object-level logging disabled
...
```

The change refactors the code to look more like `extra720` and should be a little bit faster as it makes less calls to 

This change should also address #387

### DIFF
--- a/checks/check_extra725
+++ b/checks/check_extra725
@@ -33,7 +33,7 @@ extra725(){
   # now create a list with all trails available and their region
   TEMP_TRAILS_LIST_FILE=$(mktemp -t prowler.trails-list-XXXXXX)
   for regx in $REGIONS; do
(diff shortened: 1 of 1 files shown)


patdowney:
~~Please don't merge yet, it doesn't behave properly with the {{-f}} flag to {{prowler}}.~~

Never mind, {{prowler}} doesn't filter buckets to specified regions at the moment, so behaviour is consistent.

toniblyx:
Thanks @patdowney

PR #540, opened 2020-04-12
check121 - Filter out users who do not have a console password

### LINKED ISSUE #513: check121 does not first check if a user has console access
check121 (Do not setup access keys during initial user setup for all IAM users that have a console password)

When running, this check flagged 2 keys that were assigned to users that had never been used.  However, these users do not have a console password/login (Console Password == disabled) 

Probably should check that the console password is enabled as well as the keys existing.

marcjay:
According to the benchmark, only users with a console password should be considered for this check,
therefore filter out any users who do not have a console password

Fixes #513


toniblyx:
Thanks @marcjay

PR #551, opened 2020-04-14
check21 - scope variable to prevent clobbering and fix flag ordering …

scaryghosts:
This is to fix the following issue:
https://github.com/toniblyx/prowler/issues/550

Currently after check21 runs, all subsequent checks may return false negatives because they aren't checking all regions. This is a simple change to scope the variable used in this check.


toniblyx:
This change is no longer needed, I fixed it already, thanks anyway @scaryghosts!

PR #557, opened 2020-04-15
Add the ability to generate JUnit XML reports with a junit-xml output format

### LINKED ISSUE #537: Enhancement: new output format junit-xml including log duration
When running with -M json (to produce json output), we get something like this:

```
{
  "Profile": "",
  "Account Number": "XXX",
  "Control": "[extra760] Find secrets in Lambda functions code (Not Scored) (Not part of CIS benchmark)",
  "Message": "us-west-2: Potential secret found in Lambda function connect-api-serverless-stage-custom-resource-apigw-cw-role code",
  "Status": "Fail",
  "Scored": "Not Scored",
  "Level": "Extra",
  "Control ID": "7.60",
  "Region": "us-west-2",
  "Timestamp": "2020-04-11T17:54:25Z"
}
```

It would be awesome if there was an entry for "duration".

marcjay:
If the -J flag is passed, generate JUnit XML reports for each check, in-line with how Java tools generate JUnit reports.

Check section numbers equate to 'root packages', checks are second-level packages, each check equates to a testsuite (mirroring Java where each test class is a testsuite) and each pass/fail of a check equates to a testcase
Time the execution of each check and include this in the report
Include properties (Prowler version, check level etc.) in-line with standard JUnit files
XML escape all strings for safety

Detect if a user has GNU coreutils installed on Mac OS X, but not as their default, switching to using gdate for date commands if so, as it has more features, including getting dates in milliseconds
Add prowler-output, junit-reports and VSCode files to .gitignore
Update README to include JUnit info, address markdownlint warnings
Remove unused arguments to jq in generateJsonAsffOutput

Fixes #537

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,12 @@ tags
 
 # MacOs DS_Store
 *.DS_Store
+
+# Prowler output
+prowler-output-*
+
+# JUnit Reports
+junit-reports/
+
+# VSCode files
+.vscode/
--- a/README.md
+++ b/README.md
@@ -45,7 +45,6 @@ Read more about [CIS Amazon Web Services Foundations Benchmark v1.2.0 - 05-23-20
 - HIPAA [hipaa] Read more [here](#hipaa-checks)
 - Trust Boundaries [trustboundaries] Read more [here](#trustboundaries-checks)
 
-
 With Prowler you can:
 
 - get a colorful or monochrome report
@@ -68,6 +67,7 @@ This script has been written in bash using AWS-CLI and it works in Linux and OSX
     AWS-CLI can be also installed it using "brew", "apt", "yum" or manually from <https://aws.amazon.com/cli/>, but `ansi2html` and `detect-secrets` has to be installed using `pip`. You will need to install `jq` to get more accuracy in some checks.
 
 - Make sure jq is installed (example below with "apt" but use a valid package manager for your OS):
+
     ```sh
     sudo apt install jq
     ```
@@ -84,7 +84,9 @@ This script has been written in bash using AWS-CLI and it works in Linux and OSX
     ```sh
     aws configure
     ```
+
     or
+
     ```sh
     export AWS_ACCESS_KEY_ID="ASXXXXXXX"
     export AWS_SECRET_ACCESS_KEY="XXXXXXXXX"
@@ -110,7 +112,7 @@ This script has been written in bash using AWS-CLI and it works in Linux and OSX
 
     Use `-l` to list all available checks and group of checks (sections)
 
-    If you want to avoid installing dependences run it using Docker:
+    If you want to avoid installing dependencies run it using Docker:
 
     ```sh
     docker run -ti --rm --name prowler --env AWS_ACCESS_KEY_ID --env AWS_SECRET_ACCESS_KEY --env AWS_SESSION_TOKEN toniblyx/prowler:latest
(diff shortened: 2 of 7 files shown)


marcjay:
Example JUnit file - after running:
```sh
./prowler -J
```

```xml
<?xml version="1.0" encoding="UTF-8"?>
<testsuite name="1.3" timestamp="2020-04-15T00:05:58Z">
  <properties>
    <property name="prowler.version" value="2.2.1"/>
    <property name="aws.profile" value=""/>
    <property name="aws.accountNumber" value="1234567890"/>
    <property name="check.id" value="1.3"/>
    <property name="check.scored" value="Scored"/>
    <property name="check.level" value="Level 1"/>
    <property name="check.asff.type" value="Software and Configuration Checks/Industry and Regulatory Standards/CIS AWS Foundations Benchmark"/>
    <property name="check.asff.resourceType" value="AwsIamUser"/>
  </properties>
  <testcase name="[check13] Ensure credentials unused for 90 days or greater are disabled (Scored) (1)" classname="1.3" time="0.996">
    <system-out>User &quot;marcjay&quot; found with credentials used in the last 90 days</system-out>
  </testcase>
  <testcase name="[check13] Ensure credentials unused for 90 days or greater are disabled (Scored) (2)" classname="1.3" time="0.981">
    <failure message="User &quot;testuser&quot; has not logged in during the last 90 days"/>
  </testcase>
</testsuite>
```

Jenkins job post-build action:
image

Jenkins job console output (AnsiColor plugin):
image

Test result - failed tests:
image

Test result - top level (sections):
image

Failed test details:
image

Passed test details:
image

Fixed some issues, re-ran:
image

Issue shows as fixed:
image

Test trend:
image

grayaii:
@marcjay I'm getting this warning when running:

    ./prowler -c check11 -r us-west-2 -f us-west-2 -J
    ....
    ./include/junit_integration: line 81: "1586949195430" - "1586949195401": syntax error: operand 
    expected (error token is ""1586949195430" - "1586949195401"")

I got this on both a Mac and  Linux box.

I printed the values of the time here:
 
      echo "DEBUG: $time_now"
      echo "DEBUG: $JUNIT_CHECK_START_TIME"
      test_case_duration=$(printf "%.3f" "$(("$time_now" - "$JUNIT_CHECK_START_TIME"))e-3")

And they are:

    DEBUG: 1586949753520
    DEBUG: 1586949753491

I think there's something funky in that bash expansion, but I can't put my finger on it.  

I'll keep digging...

marcjay:
@grayaii thanks for testing it out - ~what shell do you use?~
EDIT: ignore that - I just pushed a fix to this branch, would be great if you could retry - thanks!

grayaii:
I'm using Bash.

I got it working by removing the quotes:
This works:

    test_case_duration=$(printf "%.3f" "$((1586949753520 - 1586949753491))e-3")

This fails:

    test_case_duration=$(printf "%.3f" "$(("1586949753520" - "1586949753491"))e-3")


I bet different shells treat arithmetic operators differently.

marcjay:
Many thanks - just pushed a fix for that - should be good to try again

grayaii:
It's working like a charm!  I just kicked off a jenkins job and so far so good!
Thanks a million!!!

toniblyx:
Thanks for this effort @marcjay, quick question, why do you used `-J` instead of adding another output to `-M`?

marcjay:
> quick question, why do you used `-J` instead of adding another output to `-M`?

I think my thinking was that JUnit results would be supplemental to an output mode, as they are written to a directory - i.e. you'd run with e.g. `-M csv -J` to write CSV to stdout and generate JUnit files at the same time. Running with just `-M junit-xml` wouldn't print very much to the user, so they might wonder if it was working ok/what's the progress etc.

Now, since I started the branch, you've added support for multiple output formats :) so perhaps a mode makes more sense? The README instruction to use this would then be `-M text,junit-xml` so you'd get sensible stdout output while it runs

EDIT: Made this change in 78f649b

PR #560, opened 2020-04-15
Only check latest version of task definition

grayaii:

This is a quick and dirty way of filtering down the latest version of an ECS Task Definition.

I'm not sure where to put that python file (maybe there is a common place to put helper scripts?), and I'm not sure where to save that temp file, ALL_TASK_DEFINITIONS.txt

This is just to let you know how to do it with python :)


checks/check_extra768
```diff
@@ -23,19 +23,22 @@ extra768(){
...
   fi
-
   textInfo "Looking for secrets in ECS task definitions' environment variables across all regions... "
   for regx in $REGIONS; do
-    LIST_OF_TASK_DEFINITIONS=$($AWSCLI ecs list-task-definitions $PROFILE_OPT --region $regx --query taskDefinitionArns[*] --output text)
-    if [[ $LIST_OF_TASK_DEFINITIONS ]]; then
-      for taskDefinition in $LIST_OF_TASK_DEFINITIONS;do
-        IFS='/' read -r -a splitArn <<< "$taskDefinition"
+    # Get a list of all families first:
+    FAMILIES=$($AWSCLI ecs list-task-definition-families $PROFILE_OPT --region $regx --status ACTIVE | jq -r .families[])
```
marcjay, line 29: Minor but I think you could use `--output text` here and drop the `| jq -r .families[]`

checks/check_extra768
```diff
@@ -23,19 +23,22 @@ extra768(){
...
+    if [[ $FAMILIES ]]; then
+      for FAMILY in $FAMILIES;do
+        # Get the full task definition arn:
+        TASK_DEFINITION_TEMP=$($AWSCLI ecs list-task-definitions $PROFILE_OPT --region $regx --family-prefix $FAMILY --sort DESC --max-items 1 | jq -r .taskDefinitionArns[0])
```
marcjay, line 33: Also minor but is jq still needed if `--max-items 1` is used along with `--output text`?

PR #562, opened 2020-04-20
using api commands to check if macie is enabled instead of looking ia…

eko0126:
…m role


eko0126:
using api commands to check if macie is enabled instead of looking iam role

toniblyx:
rolled back due to an error with endpoints that slow downs, let's hold this solution until we find a better option

PR #569, opened 2020-04-22
Check if gbase64 (GNU) is available on Mac and use it in preference to BSD base64

### LINKED ISSUE #568: os_detector does not use gbase64 if coreutils is installed on Mac
Currently, if os_detector finds that `gdate` is available it expects `base64` to be the coreutils version. It should check if gbase64 is available and use that if so

Bug introduced by #539

marcjay:
Previously it was switching to GNU versions of base64 even if base64 was the BSD version

Fixes #568

### DIFF
--- a/include/os_detector
+++ b/include/os_detector
@@ -12,6 +12,7 @@
 # specific language governing permissions and limitations under the License.
 
 DATE_CMD="date"
+BASE64_CMD="base64"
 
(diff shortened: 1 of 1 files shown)


toniblyx:
It works fine now with and without `coreutils` on Mojave. Thanks!

PR #572, opened 2020-04-24
Fix for issue 571: "json-asff" output mode

deathbywedgie:
Fix for issue 571: selecting mode "json-asff" prints both "json-asff" and regular "json"
https://github.com/toniblyx/prowler/issues/571


marcjay:
This looks to me like the right fix, but I think it should be applied to everywhere `${MODES[@]}` is referenced, and to both sides of the expression - so it doesn't leave a potential bug should a similarly named output be added in future?

deathbywedgie:
I agree, but I haven't tested it everywhere yet, just in all the ways that the immediate fix has been applied. Let's merge the current fix, and I'll work on seeking out all other places for a follow-up PR.

toniblyx:
This has been fixed already in  #576, thanks anyway @deathbywedgie!

PR #574, opened 2020-04-25
Workaround for tr error with json-asff output mode in CentOS

deathbywedgie:
Workaround for issue 573: seeing error "tr: range-endpoints of '~-\012' are in reverse collating sequence order" when there are failures in CentOS and the output mode is "json-asff"
https://github.com/toniblyx/prowler/issues/573


deathbywedgie:
Withdrawing in lieu of another PR with a cleaner fix: https://github.com/toniblyx/prowler/pull/575

PR #575, opened 2020-04-25
Ensure that hyphen is at end of tr string to prevent 'reverse collating sequence order' error in GNU tr

### LINKED ISSUE #573: Errors with json-asff in CentOS: tr: range-endpoints of '~-\012' are in reverse collating sequence order
When running in a CentOS docker image and using the json-asff output mode I am seeing errors in the output:
`tr: range-endpoints of '~-\012' are in reverse collating sequence order`

The script is completing successfully, and if I pipe stderr to a file or to /dev/null I see all the output that I would expect in stdout, so it does appear to be working properly otherwise, I just get tons of those errors in stderr that I have to sift through or filter out.

I don't see those errors when I run from MacOS.

marcjay:
Also stop echo from adding newlines using `-n`, removing the need to stop replacing new-line characters with underscores

Fixes #573

### DIFF
--- a/include/outputs
+++ b/include/outputs
@@ -201,7 +201,7 @@ generateJsonAsffOutput(){
   --arg ACCOUNT_NUM "$ACCOUNT_NUM" \
   --arg TITLE_TEXT "$TITLE_TEXT" \
   --arg MESSAGE "$(echo -e "${message}" | sed -e 's/^[[:space:]]*//')" \
-  --arg UNIQUE_ID "$(LC_ALL=C echo -e "${message}" | tr -cs '[:alnum:]._~-\n' '_')" \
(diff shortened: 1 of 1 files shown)


marcjay:
Test evidence using CentOS (though the more general issue is anywhere were tr is the GNU version)
Existing command:
```
$ LC_ALL=C echo -e "foo" | tr -cs '[:alnum:]._~-\n' '_'
tr: range-endpoints of '~-\012' are in reverse collating sequence order
```

On this branch:
```
$ LC_ALL=C echo -e -n "foo" | tr -cs '[:alnum:]._~-' '_'
foo
```

deathbywedgie:
I've confirmed that this resolves my issue and is a cleaner fix than the one I previously submitted. I have since closed my PR in favor of this one.

toniblyx:
LGTM, thanks!

PR #581, opened 2020-04-28
Exclude website hosting buckets in extra764 check

### LINKED ISSUE #580: Exclude website hosting buckets in extra764 check
According to [AWS S3 Documentation](https://docs.aws.amazon.com/AmazonS3/latest/dev/WebsiteEndpoints.html),
> Amazon S3 website endpoints do not support HTTPS.

Shall we exclude website hosting buckets when checking bucket policy in check [`extra764`](https://github.com/toniblyx/prowler/blob/master/checks/check_extra764)?

yumminhuang:
Fix #580

If bucket has website configuration, skip bucket policy check.


toniblyx:
Thanks @yumminhuang, this change makes sense and probably is something that many other users need somehow. However I'd like to address that requirement adding support for White/Blacklist of resources like suggested here #514. What do you think?

yumminhuang:
Thanks @toniblyx for your response.
#514 is a general solution. I think it can resolve my problem.

PR #582, opened 2020-04-28
Fix check12's grep to find users who really have password access

nimrodkor:
Hey!
Ran into an account with a user named `truedy`. He came up positive in this check because the of the grep... Fixed by searching for ` true` explicitly, which is what the output of `awk` is

(cherry picked from commit 4006c581a06c449b66ede8892b9ae18c735ad34c)

### DIFF
--- a/checks/check12
+++ b/checks/check12
@@ -19,7 +19,7 @@ CHECK_ALTERNATE_check102="check12"
 check12(){
   # "Ensure multi-factor authentication (MFA) is enabled for all IAM users that have a console password (Scored)"
   # List users with password enabled
-  COMMAND12_LIST_USERS_WITH_PASSWORD_ENABLED=$(cat $TEMP_REPORT_FILE|awk -F, '{ print $1,$4 }' |grep true | awk '{ print $1 }')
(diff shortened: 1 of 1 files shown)


checks/check12
```diff
@@ -19,7 +19,7 @@ CHECK_ALTERNATE_check102="check12"
 check12(){
   # "Ensure multi-factor authentication (MFA) is enabled for all IAM users that have a console password (Scored)"
   # List users with password enabled
-  COMMAND12_LIST_USERS_WITH_PASSWORD_ENABLED=$(cat $TEMP_REPORT_FILE|awk -F, '{ print $1,$4 }' |grep true | awk '{ print $1 }')
+  COMMAND12_LIST_USERS_WITH_PASSWORD_ENABLED=$(cat $TEMP_REPORT_FILE|awk -F, '{ print $1,$4 }' |grep -F ' true' | awk '{ print $1 }')
```
marcjay: The way this has been tackled in other checks is to anchor the regex, see 
https://github.com/toniblyx/prowler/blob/master/checks/check121#L25 or
https://github.com/toniblyx/prowler/blob/master/include/check_creds_last_used#L58
```
grep " true$"
```
I imagine one approach should be chosen and consistently applied in all situations - I have no preference which personally

nimrodkor: Thanks @marcjay ! I added the `$` for consistency.

toniblyx (approved):
Awesome, thanks!

PR #586, opened 2020-04-29
check21 - If no account cloudtrail trail is found, check org trail

nimrodkor:
This check started failing when we moved to an org trail...


toniblyx:
Thanks!

PR #587, opened 2020-04-29
Some Crazy Ass Python S**t

chalbersma:

* Adds a check and call objects and a python driver for them.
* Ads some persistent fs cache logic for calls (could use more tlc)
* Ads a config language to split out the checks from the logic that does the checks. Makes (really) too much use of jq as a config language to do some things.
* Adds some example checks for the CIS 1.x checks as an example of what could be possible.
* Has some unused logic around which "platform" to run checks against so in theory we could have different call logic bits for different platform.
* Could likely use a lot of changes
* Let me know what you think
* I like bullet points.


toniblyx:
What do you mean by "platform", AWS account?

toniblyx:
Thanks @chalbersma, looks like you have been having fun with it :) A list of features that mach with existing code would be helpful, at least for me to better understand what is included here.

chalbersma:
> What do you mean by "platform", AWS account?

Yes indeed. While there's no code to support it yet. The thought process is that future checks that were targeted towards azure, gcp, generic k8s etc... could be added and handled.

chalbersma:
> Thanks @chalbersma, looks like you have been having fun with it :) A list of features that mach with existing code would be helpful, at least for me to better understand what is included here.

Documentation is lacking (at best). I'll try to put something together.

chalbersma:
@toniblyx I took a first pass at documenting the check format which I think sort of defines the core of what you were looking for. Let me know what you think.

toniblyx:
I don't know what to do with this to be honest, I'm not convinced on the way to do it. If it is something totally different to Prowler it won't be Prowler, can be anything else, other project maybe? I don't have any plan to make Prowler GCP or Azure compatible. That would be really hard to maintain and evolve with a minimum of quality and speed. We should take advantage of the job done, like checks or even groups and also mapping of current features with those that can be implemented in python. To support new checks in python would be awesome but rewrite 150 checks in python is not something that I might be interested in.

chalbersma:
Fair enough. I may pursue it as a personal project. But it makes perfect sense that this may not be the way you want to move forward with the project. Feel free to close.

marcjay:
I think it's an impressive piece of engineering @chalbersma! - being able to represent checks as YAML files, and I think it's a great goal that would, in-theory, simplify adding extra checks.

However, personally I think the checks might be too complex and open-ended to be represented this way. I'm not sure that from the perspective of someone wanting to author a new check, that trying to get their head around the YAML configuration is easier than writing some python that implements the check - it's clearly non-trivial trying to represent the logic involved in a check in YAML.

The other challenge I'd see is, can you confidently say that any check that anyone might want to add in the future (or even all of the other existing checks in Prowler) can always be represented in YAML, and without changing the schema (which might be a big undertaking once checks are written). I think it would be impossible to make that statement - someone could come up with a new check tomorrow that required logic unrepresentable in the current schema - you'd need an escape-hatch of some kind.

I wonder if it might be better to focus on supporting python checks alongside bash checks, using a kind of plugin mechanism, where Prowler would call a check, passing in boto, and providing helper methods for performing common functions, e.g. iterating over all regions, fetching an IAM credential report. I'm hoping to have some time on the weekend to put together a POC along these lines.

toniblyx:
Thanks @marcjay. Also wanted to thank you @chalbersma for the code you have written, I didn't meant that it is not good or helpful what you have done, it is just to make sure we make the right step forward in this regard. I hope you may understand. Please feel free to keep contributing on anything you consider it makes Prowler better.

chalbersma:
@marcjay & @toniblyx Thanks. Ya whatever the logical engine used there's going to be some issues with it. I think in theory that new check types could be extended with new logic in the outcome. But I think the "glue" that's really needed is a way to do some sort of generic "looks like x" in a config file instead of in code.

In theory I guess we could try to support a Nagios style interface for checks where we provide a constant set of input (like a json blob) and expect a constant output (like a json blob or exit code) and that could theoretically allow checks to be written in any language that can be called by the default shell. I'll have to do more thinking on that.

Issue #498, opened 2020-02-24
false positive secretes detected (extra775)


Mexx77:
Hey,

`Find secrets in EC2 Auto Scaling Launch Configuration (Not Scored) (Not part of CIS benchmark) - Looking for secrets in EC2 Auto Scaling Launch Configuration across all regions - Potential secret found`

I am getting a false positive on extra775. Its a environment variable in EC2 user data and not a secret that ends up in plain text in user data. Prowler uses https://github.com/Yelp/detect-secrets. `detect-secret` has means to ignore certain findings. Unfortunately, prowler won't let me configure it yet (or I didn't find out how). Help appreciated :)

gliptak:
would https://github.com/toniblyx/prowler/issues/154 offer this?

toniblyx:
Hi @Mexx77, you can configure `detect-secrets` entropy from here if you want https://github.com/toniblyx/prowler/blob/master/include/secrets_detector#L29

Mexx77:
> Hi @Mexx77, you can configure `detect-secrets` entropy from here if you want https://github.com/toniblyx/prowler/blob/master/include/secrets_detector#L29

Can I? Isn't it hard coded at that line by prowler?

toniblyx:
Variables `BASE64_LIMIT` and `HEX_LIMIT` are there to be customized if needed, feel free to adapt them or the code to fit your needs. And if you find a better set up to avoid false positives, let me know to improve it.

toniblyx:
Please feel free to reopen this issue if you need further help on it.

PR #589, opened 2020-05-06
Support setting entropy limit for detect-secrets from env

yumminhuang:
Support customizing entropy limit from environment variables `BASE64_LIMIT` and `HEX_LIMIT`.
If environment variables are not set, use the default values

It can help to resolve issues like #498

---


toniblyx:
Awesome, I think it is very good enhancement. Thanks @yumminhuang

PR #592, opened 2020-05-07
Change value of FAIL to FAILED for jsonAsff output type 

wildtangent:
`FAIL` status is incompatible…with AWS Security Hub submission

https://docs.aws.amazon.com/securityhub/latest/userguide/securityhub-findings-format.html#asff-compliance

### DIFF
--- a/include/outputs
+++ b/include/outputs
@@ -222,6 +222,11 @@ generateJsonAsffOutput(){
   # Replace any successive non-conforming characters with a single underscore
   local message=$1
   local status=$2
+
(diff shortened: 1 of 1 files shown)


marcjay:
This was indeed broken as part of the whitelisting PR - #514 it seems - see https://github.com/toniblyx/prowler/pull/514/files#diff-8af903e86cda25aa5b295738c278f634R123
This fix looks good to me

toniblyx:
ooops! Thanks for clarifying.

PR #600, opened 2020-05-22
Feature/ssl checks

jonjozwiak:


toniblyx:
Thanks @jonjozwiak @nimrodkor  @jonnyCodev and @orevron pretty helpful one as usual coming from bridgecrew.io 👍 ;)

Would you mind to send separate PR per change? (one for new TLS checks and separate ones per fix in checks) It is to keep better track of the changes and make the release notes build easier (otherwise it becomes a nightmare).

Also, for the new checks, in order to better support ASFF we are now adding the resource type as for example in `extra734` with variable `CHECK_ASFF_RESOURCE_TYPE_extra734="AwsS3Bucket"`. In your case it would be  `CHECK_ASFF_RESOURCE_TYPE_extra791="AwsCloudFrontDistribution"` and `CHECK_ASFF_RESOURCE_TYPE_extra792="AwsElasticLoadBalancingV2Listener"` based on the list here https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-template-resource-type-ref.html

Thanks again!

jonjozwiak:
This is being replaced by PR 602.

toniblyx:
Thanks @nimrodkor , btw, it is a bit weird because https://github.com/toniblyx/prowler/pull/600/commits/0db690ad5fa26c0157f6f40ea651495bdd9e9715 is not actually applied in this master branch as other fixes in that list. I have noticed that one because it is not working as expected I'm afraid. Can you please confirm is that is a valid patch and if there are others we can take advantage?

PR #616, opened 2020-05-27
Split ignores by newline instead of spaces only

urjitbhatia:

I was looking at https://github.com/toniblyx/prowler/issues/598 and realized that whitelist didn't match properly on spaces.
This should help support something like `check121:User foo is not setup properly` vs just `check121:foo`


toniblyx:
awesome, thanks for taking care of this @urjitbhatia

PR #621, opened 2020-06-05
Import Security Hub finding into the same region as the related resource

### LINKED ISSUE #618: SecurityHub import fails with AccessDenied if finding is in another region
Even with policy allowing BatchImportFindings, API call can fail if we are trying to import finding from one region to Security Hub in another region, e.g. missing VPC flow logs in ap-south-1 being imported into us-east-1 Security Hub.
Region-locking prowler (-r switch) does not seem to help, because some checks are global.
This can be important if AccessDenied events are monitored by some CloudWatch alerts, and because of that we can't run prowler frequently without disturbing the team.

marcjay:
Force the batch-import-findings AWS CLI call to be directed at the region the currently reporting resource is located in, as Security Hub enforces this requirement

When checking that Security Hub is enabled, check for all regions that are in scope, e.g. all regions, unless '-f \<region>' is used

Fixes #618

### DIFF
--- a/include/outputs
+++ b/include/outputs
@@ -27,13 +27,13 @@ TIMESTAMP=$(get_iso8601_timestamp)
 PROWLER_PARAMETERS=$@
 
 # Ensure that output directory always exists when -M is used
-if [[ $MODE ]];then 
+if [[ $MODE ]];then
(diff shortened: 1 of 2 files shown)


toniblyx:
Thanks!

PR #623, opened 2020-06-10
Fixing profile and region settings for check_extra792 - ELB SSL ciphers

jonjozwiak:
A region and profile was missed on one of the API calls.  This PR corrects that.  


toniblyx:
Good catch, thanks @jonjozwiak

PR #628, opened 2020-06-26
Fix listing configurations if default output format is not JSON

Anthirian:

This prevents errors like:
```
parse error: Invalid numeric literal at line 1, column 2 
```

### DIFF
--- a/checks/check119
+++ b/checks/check119
@@ -18,7 +18,7 @@ CHECK_ALTERNATE_check119="check119"
 
 check119(){
   for regx in $REGIONS; do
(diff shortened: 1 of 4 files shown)


toniblyx:
Thanks @Anthirian, I guess you need that change because your AWSCLI is not configured to use json by default, right?

Anthirian:
Yes correct, I have no default output format configured.

toniblyx:
That's a bit weird because cli output is json by default, what output do you get instead? I don't think your suggested changes will break anything so it will be merged after a test. Thanks.

Anthirian:
Maybe it had to do with the fact that I originally set it to a different format, but modified the config files to remove that part. In any case I don't think it will be harmful to explicitly state the desired output. In fact it may prove more robust in the future.

PR #632, opened 2020-07-11
pipenv: add lockfile

chenrui333:
Add lockfile for reproducible build

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,0 +1,138 @@
+{
+    "_meta": {
+        "hash": {
+            "sha256": "962c73a3636be996770ae2686c85f357df06cd504dd2c169617cb55072b81873"
+        },
+        "pipfile-spec": 6,
+        "requires": {
+            "python_version": "3.7"
+        },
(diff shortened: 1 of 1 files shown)


toniblyx:
Thanks @chenrui333, can you explain a bit the reason of this change? I see you are maintatiner of Homebrew, is that somehow related?

chenrui333:
> Thanks @chenrui333, can you explain a bit the reason of this change? I see you are maintatiner of Homebrew, is that somehow related?

Yeah, totally

Ideally, it would be great to leverage the lockfile to build from the source as it is (which I intended to bring it to the homebrew side), but I run into some issue (I will post it later on this).

But in general, I think it is preferred to have lockfile version controlled along with the source code.

toniblyx:
Could you please point me to some documentation to better understand what you are willing to do? Thanks.

toniblyx:
Since I haven't had more information about this I'll close it. Please feel free to reopen it and provide more information about packaging Prowler for Homebrew.

PR #633, opened 2020-07-16
extra792 - skip check if no HTTPS/SSL Listener plus add NLB Support

jonjozwiak:
Updated extra792 to return INFO if no HTTPS or SSL listener is setup rather than returning PASS. 
I also extended this to check NLBs if TLS is enabled.  


toniblyx:
LGTM, thanks

PR #634, opened 2020-07-17
Adding check for SSL on load balancers

jonjozwiak:

This check looks for load balancers without SSL listeners.  For classic ELBs it will fail if no HTTPS/SSL listener or return info if both HTTP and HTTPS (as redirect would need to be handled on the backend target).  

For ALBs, this check will pass if HTTPS only.  It will return info if there is HTTP redirecting to HTTPS.  Otherwise it will fail if HTTP only or HTTP without redirect.


toniblyx:
LGTM

PR #638, opened 2020-07-24
Don't fail check extra737 for keys scheduled for deletion

QuantumBJump:

If a KMS key is scheduled for deletion, it can't have rotation enabled, but it still shouldn't cause the test to fail. This fixes that.


toniblyx (commented):
Thanks @QuinnStevens! Could you please add --output json to the command? That is in case somebody has set up other than json default output already configured.

PR #642, opened 2020-07-31
Check org Cloudtrail first and pass if enabled

tkallenberg-tw:
Hi @toniblyx 

We changed the Cloudtrail check implementation to first check if an organizational trail is enabled and if so we stop and report PASS. Without this change the check would fail at the first single region trail, even if there is an organizational / other trail with multi region enabled.

### DIFF
--- a/checks/check21
+++ b/checks/check21
@@ -17,8 +17,13 @@ CHECK_ASFF_RESOURCE_TYPE_check21="AwsCloudTrailTrail"
 CHECK_ALTERNATE_check201="check21"
 
 check21(){
-  trail_count=0
-  # "Ensure CloudTrail is enabled in all regions (Scored)"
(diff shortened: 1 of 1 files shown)


toniblyx:
Thanks @tkallenberg-tw I'd like to replicate your first issue to understand if this change is really needed since some other changes have been applied here to make it more accurate. Could you please share your configuration or regions your are using and enabling cloudtrail and how you do run prowler? Are you forcing any region with -f or -r? Thanks!

toniblyx:
So far I will leave this check21 behavior as it is, I understand that to be strict with what it asking it should be pass if just one trail is global, however is good to see if there are other trails and its configuration. Even if there are more than one global could be suspicious. Thanks.

PR #647, opened 2020-08-21
[Check12] Bugfix: Remove $ from grep

xeroxnir:
Check is failing to detect users without MFA, solved by removing `$` sign.

### DIFF
--- a/checks/check12
+++ b/checks/check12
@@ -19,7 +19,7 @@ CHECK_ALTERNATE_check102="check12"
 check12(){
   # "Ensure multi-factor authentication (MFA) is enabled for all IAM users that have a console password (Scored)"
   # List users with password enabled
-  COMMAND12_LIST_USERS_WITH_PASSWORD_ENABLED=$(cat $TEMP_REPORT_FILE|awk -F, '{ print $1,$4 }' |grep -F ' true$' | awk '{ print $1 }')
+  COMMAND12_LIST_USERS_WITH_PASSWORD_ENABLED=$(cat $TEMP_REPORT_FILE|awk -F, '{ print $1,$4 }' |grep -F ' true' | awk '{ print $1 }')
   COMMAND12=$(
     for i in $COMMAND12_LIST_USERS_WITH_PASSWORD_ENABLED; do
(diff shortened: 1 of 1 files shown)


checks/check12
```diff
@@ -19,7 +19,7 @@ CHECK_ALTERNATE_check102="check12"
 check12(){
   # "Ensure multi-factor authentication (MFA) is enabled for all IAM users that have a console password (Scored)"
   # List users with password enabled
-  COMMAND12_LIST_USERS_WITH_PASSWORD_ENABLED=$(cat $TEMP_REPORT_FILE|awk -F, '{ print $1,$4 }' |grep -F ' true$' | awk '{ print $1 }')
+  COMMAND12_LIST_USERS_WITH_PASSWORD_ENABLED=$(cat $TEMP_REPORT_FILE|awk -F, '{ print $1,$4 }' |grep -F ' true' | awk '{ print $1 }')
```
toniblyx: hey @nimrodkor, you made this change 4 months ago [here](https://github.com/toniblyx/prowler/commit/dbca70ef2e47ddfe0851756f83ee1883b75ed67b). For a list of users like this sample below, with MFA enabled, it fails, but works without `-F` in that grep.

```
username,arn:aws:iam::123456789012:user/username,2020-08-20T15:28:56+00:00,true,no_information,2020-08-20T15:28:58+00:00,N/A,false,false,N/A,N/A,N/A,N/A,false,N/A,N/A,N/A,N/A,false,N/A,false,N/A
```

Could you clarify please?

nimrodkor: Hey @toniblyx , sure thing!
I ran into a user which had the word true inside the username. I.e. `true-admin`
A user like that would pass the regex `true`, but not `true$`.
If it works without the $ - it works, but it is faulty. I did not check without the -F, but if it works without the -F thats fine by me

toniblyx: Thanks for clarifying, I have tested it again using `true-admin` with `awk -F, '{ print $1,$4 }' |grep 'true$'` and works fine, definitely `-F` and `space` is not needed.

PR #652, opened 2020-09-04
Whitelist improvements

QuantumBJump:

This allows you to use regexes in the whitelist file - my company was in a situation where we wanted to ignore all buckets called <something>-logs, and didn't want to write out every one individually.


toniblyx:
Thanks @QuinnStevens, I'll give it a look as soon as possible. Seems pretty helpful for others!

PR #667, opened 2020-10-02
Fix PROFILE_OPT arg

QuantumBJump:

It looks like Prowler makes lots of references to the $PROFILE_OPT arg throughout the codebase, but the actual argument is never set. Unless I'm missing something, I think this means that the `-p` flag hasn't been working at all for quite a while.


toniblyx:
That is not needed, that logic is actually in `include/aws_profile_loader`. It works when -p is used and a profile other than the environment variables or instance profile are used.

QuantumBJump:
Ah, I see! Thanks for the clarification!

PR #671, opened 2020-10-21
Adding fix to generate test summary so reports display graphs correctly

stevecjones:

When using the junit outputs tests get recorded correctly however there is a slight issue with the following line.

<testsuite name="7.31" timestamp="2020-10-21T15:11:59Z">

When passing the output in tools like Azure Devops or AWS CodeBuild the engine expects a summary of all tests within the junit file like the following:-

<testsuite name="7.31" tests="4" failures="0" skipped="0" errors="0" timestamp="2020-10-21T15:11:59Z">

This PR addresses the creation of this line after enumerating the test cases within. So in the case of AWS CodeBuild the following graph should then populate.

image

Thanks!

### DIFF
--- a/include/junit_integration
+++ b/include/junit_integration
@@ -14,6 +14,11 @@
 # Generates JUnit XML reports which can be read by Jenkins or other CI tools
 
 JUNIT_OUTPUT_DIRECTORY="junit-reports"
+JUNIT_TESTS_COUNT="0"
+JUNIT_SUCCESS_COUNT="0"
+JUNIT_FAILURES_COUNT="0"
+JUNIT_SKIPPED_COUNT="0"
+JUNIT_ERRORS_COUNT="0"
 
 is_junit_output_enabled() {
   if [[ " ${MODES[@]} " =~ " junit-xml " ]]; then
@@ -44,7 +49,7 @@ prepare_junit_check_output() {
   JUNIT_OUTPUT_FILE="$JUNIT_OUTPUT_DIRECTORY/TEST-$1.xml"
   printf '%s\n' \
     "<?xml version=\"1.0\" encoding=\"UTF-8\"?>" \
-    "<testsuite name=\"$(xml_escape "$(get_junit_classname)")\" timestamp=\"$(get_iso8601_timestamp)\">" \
+    "<testsuite name=\"$(xml_escape "$(get_junit_classname)")\" tests=\"_TESTS_COUNT_\" failures=\"_FAILURES_COUNT_\" skipped=\"_SKIPPED_COUNT_\" errors=\"_ERRORS_COUNT_\" timestamp=\"$(get_iso8601_timestamp)\">" \
     "  <properties>" \
     "    <property name=\"prowler.version\" value=\"$(xml_escape "$PROWLER_VERSION")\"/>" \
     "    <property name=\"aws.profile\" value=\"$(xml_escape "$PROFILE")\"/>" \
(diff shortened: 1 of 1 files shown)


toniblyx:
Thanks @stevecjones, I would like to be able to test this fix can you provide briefly how to reproduce it with AWS CodeBuild? I'm not familiar with that type of usage for Prowler and would be helpful to test it that way. @marcjay what do you think about this change?

include/junit_integration
```diff
@@ -61,6 +61,20 @@ prepare_junit_check_output() {
...
+  TEST_COUNT="0"
+  TEST_SUCCESS_COUNT="0"
+  TEST_FAILURE_COUNT="0"
+  TEST_SKIPPED_COUNT="0"
+  TEST_ERROR_COUNT="0"
+  TEST_COUNT=$(grep -c "testcase name=" "$JUNIT_OUTPUT_FILE")
+  TEST_SUCCESS_COUNT=$(grep -A1 "testcase name=" "$JUNIT_OUTPUT_FILE" | grep -c "<system-out>")
+  TEST_FAILURE_COUNT=$(grep -A1 "testcase name=" "$JUNIT_OUTPUT_FILE" | grep -c "<failure message")
+  TEST_SKIPPED_COUNT=$(grep -A1 "testcase name=" "$JUNIT_OUTPUT_FILE" | grep -c "<skipped message")
+  TEST_ERROR_COUNT=$(grep -A1 "testcase name=" "$JUNIT_OUTPUT_FILE" | grep -c "<error message")
```
marcjay: Would it be better to maintain counters that increment on every pass/fail/skip? It's something that other formats could make use of, and avoids the need to use `grep` to count for XML tags that are somewhat side-effects of the pass/fail themselves (e.g. `<system-out>` for successes)

include/junit_integration
```diff
@@ -61,6 +61,20 @@ prepare_junit_check_output() {
...
+  EXTRACT_TESTSUITE_NAME=$(awk '/<testsuite name=/ { print $1,$2}' "$JUNIT_OUTPUT_FILE")
+  EXTRACT_TESTSUITE_TIME=$(awk '/<testsuite name=/ { print $3}' "$JUNIT_OUTPUT_FILE");
+  NEW_TESTSUITE_NAME=$(echo "$EXTRACT_TESTSUITE_NAME tests=\"$TEST_COUNT\" failures=\"$TEST_FAILURE_COUNT\" skipped=\"$TEST_SKIPPED_COUNT\" errors=\"$TEST_ERROR_COUNT\" $EXTRACT_TESTSUITE_TIME")
+  sed "s/<testsuite name=.*/$NEW_TESTSUITE_NAME/g" "$JUNIT_OUTPUT_FILE" > "$JUNIT_OUTPUT_FILE.$$" && mv "$JUNIT_OUTPUT_FILE.$$" "$JUNIT_OUTPUT_FILE"
```
marcjay: This is a bit gnarly. What about using placeholders in the [initial line](https://github.com/toniblyx/prowler/blob/master/include/junit_integration#L47) that's printed, that get replaced here at the end, e.g.
```
<testsuite name=\"$(xml_escape "$(get_junit_classname)")\" tests=\"_TESTS_COUNT_\"  failures=\"_FAILURES_COUNT_\" skipped=\"_SKIPPED_COUNT_\" errors=\"_ERRORS_COUNT_\" timestamp=\"$(get_iso8601_timestamp)\">"
```
and then your `sed` calls could target those placeholders, something like (not tested):
```
sed "s/_TESTS_COUNT_/${TEST_COUNT}/g;s/_FAILURES_COUNT_/${TEST_FAILURE_COUNT}/g;s/_SKIPPED_COUNT_/${TEST_SKIPPED_COUNT}/g;s/_ERRORS_COUNT_/${TEST_ERROR_COUNT}/g" "${JUNIT_OUTPUT_FILE}"
```

include/junit_integration
stevecjones: Totally agree, have refined first cut with counters however this is limited to the junit output include.

include/junit_integration
stevecjones: Refined sed with counter vars and reset counter post junit output write.

stevecjones:
Hey @marcjay please see latest commit on this PR for consideration.

Note I have included JUNIT_ERRORS_COUNT however we don't seem to catch on errored tests. Happy to strip this out if required.

include/junit_integration
```diff
@@ -73,10 +89,12 @@ output_junit_info() {
 }
 
 output_junit_failure() {
+  JUNIT_FAILURES_COUNT=$(($JUNIT_FAILURES_COUNT+1))
   output_junit_test_case "$1" "<failure message=\"$(xml_escape "$1")\"/>"
 }
 
 output_junit_skipped() {
+  JUNIT_SKIPPED_COUNT=$(($UNIT_SKIPPED_COUNT+1))
```
marcjay: This can be simplified (also there's a typo on the right-hand side)
```suggestion
  ((JUNIT_SKIPPED_COUNT++))
```

include/junit_integration
```diff
@@ -73,10 +89,12 @@ output_junit_info() {
 }
 
 output_junit_failure() {
+  JUNIT_FAILURES_COUNT=$(($JUNIT_FAILURES_COUNT+1))
```
marcjay: This can be simplified
```suggestion
  ((JUNIT_FAILURES_COUNT++))
```

include/junit_integration
```diff
@@ -60,10 +65,21 @@ prepare_junit_check_output() {
...
+  # Reset global counters as test output closed
+  JUNIT_TESTS_COUNT="0"
+  JUNIT_SUCCESS_COUNT="0"
+  JUNIT_FAILURES_COUNT="0"
+  JUNIT_SKIPPED_COUNT="0"
+  JUNIT_ERRORS_COUNT="0"
 }
 
 output_junit_success() {
+  JUNIT_SUCCESS_COUNT=$(($JUNIT_SUCCESS_COUNT+1))
```
marcjay: This can be simplified
```suggestion
  ((JUNIT_SUCCESS_COUNT++))
```

marcjay:
Cool I think it's looking good to me (apart from that `UNIT` typo) (though I'm not a maintainer). I'll try to give it a test with Jenkins later today to make sure that still works (I can't see why it wouldn't)

> Note I have included JUNIT_ERRORS_COUNT however we don't seem to catch on errored tests. Happy to strip this out if required.

Yeah a runtime error is pretty fatal to Prowler, I suppose a future enhancement could be to increment that count and always finalise the JUnit output on error, [like is done with HTML](https://github.com/toniblyx/prowler/blob/master/prowler#L199-L202)

toniblyx (approved):

stevecjones:
Let me know if this needs any more work from me to get merged to master. Thanks!

marcjay:
> Let me know if this needs any more work from me to get merged to master. Thanks!

As it stands, the skipped count won't work, due to a typo: https://github.com/toniblyx/prowler/pull/671#discussion_r515173755

stevecjones:
> > Let me know if this needs any more work from me to get merged to master. Thanks!
> 
> As it stands, the skipped count won't work, due to a typo: [#671 (comment)](https://github.com/toniblyx/prowler/pull/671#discussion_r515173755)

Sorry my mistake, corrected and pushed.

toniblyx (approved):

toniblyx:
Thanks @stevecjones for the fix and @marcjay for your support on the review.

@stevecjones, would you mind if we add that template to the project? It may help others. It requires some small changes that I've done already. WDYT?

toniblyx:
Here is the template to add to the project https://gist.github.com/toniblyx/d7a30ff7b7cfc3969c1fac30bfc2e1da

stevecjones:
> Thanks @stevecjones for the fix and @marcjay for your support on the review.
> 
> @stevecjones, would you mind if we add that template to the project? It may help others. It requires some small changes that I've done already. WDYT?

Go for it, happy to contribute. Prowler is awesome

PR #674, opened 2020-10-26
Fix quotes in check for public RDS instances

gsoltis:

While running prowler, we noticed it was failing to find some public RDS instances. Digging into it, it appears that the quoting in the `query` was causing the condition to always evaluate to false.

I was able to verify that public RDS instances show up in the results now with this change.

### DIFF
--- a/checks/check_extra78
+++ b/checks/check_extra78
@@ -24,7 +24,7 @@ extra78(){
   # "Ensure there are no Public Accessible RDS instances (Not Scored) (Not part of CIS benchmark)"
   textInfo "Looking for RDS instances in all regions...  "
   for regx in $REGIONS; do
-    LIST_OF_RDS_PUBLIC_INSTANCES=$($AWSCLI rds describe-db-instances $PROFILE_OPT --region $regx --query 'DBInstances[?PubliclyAccessible==`true` && DBInstanceStatus=="available"].[DBInstanceIdentifier,Endpoint.Address]' --output text)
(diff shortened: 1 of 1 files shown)


toniblyx:
Thanks @gsoltis could you please let me know what version of AWS CLI, operating system are you using? and also how to reproduce this error that needs this change. Thanks.

gsoltis:
Using a docker image built from the provided `Dockerfile`, I was able to demonstrate the difference:
```
docker run -it --entrypoint /bin/bash prowler:latest
bash-4.4$ python3.6
Python 3.6.9 (default, Jul 19 2020, 03:46:11) 
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import jmespath
>>> orig_search = 'DBInstances[?PubliclyAccessible==`true` && DBInstanceStatus=="available"].[DBInstanceIdentifier,Endpoint.Address]'
>>> quote_search = 'DBInstances[?PubliclyAccessible==`true` && DBInstanceStatus==`"available"`].[DBInstanceIdentifier,Endpoint.Address]'
>>> payload =  { 'DBInstances': [{'PubliclyAccessible': True, 'DBInstanceStatus': 'available', 'DBInstanceIdentifier': 'MyDB', 'Endpoint': {'Address': '1.2.3.4'}}] }
>>> 
>>> 
>>> jmespath.search(orig_search, payload)
[]
>>> jmespath.search(quote_search, payload)
[['MyDB', '1.2.3.4']]
>>>
```

This uses `jmespath==0.10.0`, installed by `aws-cli==1.18.129`

toniblyx:
Thanks for clarifying and the fix @gsoltis

PR #676, opened 2020-10-27
Run assume_role before aws commands

yumminhuang:
I used such command to run Prowler

```
/opt/prowler/prowler -A **** -R 'prowler-runner-role' -T 10800 -r ap-southeast-2
```

but it failed to get `REGIONS` and got such error

```shell
++ /bin/aws ec2 describe-regions --query 'Regions[].RegionName' --output text --region ap-southeast-2 --region-names

An error occurred (UnauthorizedOperation) when calling the DescribeRegions operation: You are not authorized to perform this operation.
+ REGIONS=
```

`ec2:DescribeRegions` is defined in the IAM role `prowler-runner-role`, so `assume-role` function should be run before any other `aws` commands.

---

### DIFF
--- a/prowler
+++ b/prowler
@@ -232,13 +232,6 @@ trap handle_ctrl_c INT
 . $PROWLER_DIR/include/securityhub_integration
 . $PROWLER_DIR/include/junit_integration
 
-# Get a list of all available AWS Regions
-REGIONS=$($AWSCLI ec2 describe-regions --query 'Regions[].RegionName' \
-  --output text \
-  $PROFILE_OPT \
-  --region $REGION \
-  --region-names $FILTERREGION)
-
 # Pre-process whitelist file if supplied
 if [[ -n "$WHITELIST_FILE" ]]; then
   # ignore lines starting with # (comments)
@@ -319,7 +312,7 @@ execute_check() {
   local asff_type_var=CHECK_ASFF_TYPE_$1
 
   local severity_var=CHECK_SEVERITY_$1
-  
+
   CHECK_SEVERITY="${!severity_var}"
 
   CHECK_ID="$1"
@@ -416,7 +409,7 @@ execute_group() {
 
 # Function to execute group by name
 execute_group_by_id() {
-  
+
   	for i in "${!GROUP_ID[@]}"; do
 		if [ "${GROUP_ID[$i]}" == "$1" ]; then
 			execute_group ${i} $2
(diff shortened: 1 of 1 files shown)


toniblyx:
hey @yumminhuang what version are you using, I have just tested the latest code in master and I don't get that error. With your command I get this error instead `An error occurred (ValidationError) when calling the AssumeRole operation: The requested DurationSeconds exceeds the 1 hour session limit for roles assumed by role chaining.`. I run `./prowler -A ******* -R assumedrole -r ap-southeast-2` and works fine.

yumminhuang:
I guess `MaxSessionDuration` of your IAM role is `3600`, so there was the `ValidationError`.

In my scenario, I run Prowler on an EC2 instance, which can be launched weekly by an ASG with scheduled policy. While the instance profile only has the permission to assume `prowler-runner-role`. Here is the CloudFormation snippet.

```yaml
  InstanceRole:
    Type: AWS::IAM::Role
    Properties:
      Path: /application/
      AssumeRolePolicyDocument:
        Version: 2012-10-17
        Statement:
          - Effect: Allow
            Principal:
              Service:
                - ec2.amazonaws.com
            Action:
              - sts:AssumeRole
      ManagedPolicyArns:
        - 'arn:aws:iam::aws:policy/AmazonSSMManagedInstanceCore'
      Policies:
        - PolicyName: ProwlerPolicy
          PolicyDocument:
            Version: "2012-10-17"
            Statement:
              - Sid: AllowAssumeProwlerRole
                Effect: Allow
                Action:
                  - "sts:AssumeRole"
                Resource:
                  - "arn:aws:iam::****:role/application/prowler-runner-role"
```

Maybe the easiest solution for me is adding `ec2:DescribeRegions` into the EC2 instance profile.

toniblyx:
Yes, I get that. I have only 1h that is the reason but I don't see the other error that needs this patch. Can you please try with the latest code in master branch to see if that is the case?

toniblyx:
With some last changes I made to the way Prowler handles STS this should be fixed. Can you test latest code in master please?

yumminhuang:
Hello @toniblyx , I still got the error in my case.

```shell
++ /usr/bin/aws ec2 describe-regions --query 'Regions[].RegionName' --output text --region ap-southeast-2 --region-names

An error occurred (UnauthorizedOperation) when calling the DescribeRegions operation: You are not authorized to perform this operation.
+ REGIONS=
```

That's fine. I can fix it by adding such IAM permission to the EC2 instance profile. Then EC2 can get regions before running `assume-role` function.

```yaml
              - Sid: AllowDescribeRegions
                Effect: Allow
                Action:
                  - "ec2:DescribeRegions"
                Resource: "*"
```

yumminhuang:
I will close this pull request, as the error can be avoided by adding `ec2:DescribeRegions` permission to the EC2 instance profile.

PR #684, opened 2020-11-13
Fix: Security Hub eventual consistency + PREFIX query bug + Archive PASSED

xeroxnir:
* **Feature:** resolveSecurityHubPreviousFails -  Also archive PASSED findings (in case -q is not used).
* **Bug:** resolveSecurityHubPreviousFails - Security Hub is eventual consistent. GetFindings filtered by UpdatedAt can return old finding versions resulting in findings being ARCHIVED incorrectly. Refactored integration to use "Read-Your-Own-Writes". 
* **Bug:** resolveSecurityHubPreviousFails - GeneratorId matched using EQUALS, not PREFIX (extra7100 matches extra71)

### DIFF
--- a/include/os_detector
+++ b/include/os_detector
@@ -104,24 +104,8 @@ gnu_get_iso8601_timestamp() {
   "$DATE_CMD" -u +"%Y-%m-%dT%H:%M:%SZ"
 }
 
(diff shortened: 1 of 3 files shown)


toniblyx:
Awesome, thanks for keep improving it!

PR #691, opened 2020-11-18
adapt check119 to exclude instances shutting down... this time with correct check...

stku1985:
brain fart: used logical 'or' instead of correct '&&'. Sorry about that.
this really fixes issue https://github.com/toniblyx/prowler/issues/689

### DIFF
--- a/checks/check119
+++ b/checks/check119
@@ -25,7 +25,7 @@ check119(){
     if [[ $INSTANCE_LIST ]]; then
       for instance in $INSTANCE_LIST; do
         STATE_NAME=$(echo $EC2_DATA | jq -r --arg i "$instance" 'select(.InstanceId==$i)|.StateName')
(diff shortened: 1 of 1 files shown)


toniblyx:
Are you sure? Can an instance be in two states at the same time?

toniblyx:
based on https://docs.aws.amazon.com/cli/latest/reference/ec2/describe-instances.html instance-state-name - The state of the instance (pending | running | shutting-down | terminated | stopping | stopped ), hence OR should work and AND shouldn't.

stku1985:
I am pretty sure that we only want to check if the ec2 instance is both not "terminated" _and_ not "shutting-down", that is if it is either pending, running, stopping or stopped. I do not claim that an instance can be in two states at once ^^.

toniblyx (approved):

toniblyx:
I get it, tested and ok.

Issue #696, opened 2020-11-20
[bug] Error with AWS CLI V2 with a profile's default output being changed


Kirizan:
I discovered when I was running prowler with AWS CLI V2, I was getting the following error three times at the start of the script: `parse error: Invalid numeric literal at line 1, column 8`

Since the scans still ran, I didn't think much of it, until I got to extra72, then I started getting some very odd behavior. The scan extra72 was giving a lot of unable to DescribeSnapshotAttributes errors. I took the following steps to troubleshoot:
1. When I looked the snapshots that were being checked, they were snapshots attached to AMI's, and didn't belong to the account.
2. I added a few lines of output to the extra72 scan and discovered the environment variable `ACCOUNT_NUM` was never being set.
3. I ran the command found in the `include/whoami` script, the one on line 16, and discovered the output was not in JSON, which caused the next few lines to fail.

To fix this issue, I did the following: 
* I added the flag `--output json` to the command on line 16, and when I reran extra72, everything worked as expected.

I adjusted the default output because when I'm running the commands myself, I prefer yaml, so I don't want to change the profile defaults back to yaml. I don't know where else in the project JSON output is relied on, but it seems those commands need the `--output json` flag added to them to ensure you are getting the output you expect.

Steps to reproduce this error:
1. Add `output = yaml` (or any other valid output) to your profile in the `.aws/config` file
2. Run Prowler, you should see the error `parse error: Invalid numeric literal at line 1, column 8` 3 times right away.
3. Any check that relies on the `ACCOUNT_NUM` environment variable will not run as expected

Addition potential issues:
Anywhere JSON output is expected and needed for something to work would be effected by this, and would require the `--output json` flag added to the command. Additionally, the environment variable AWS_DEFAULT_OUTPUT could be set to json and that should override any settings from the config file.

toniblyx:
Ok, thanks @Kirizan, how is this issue related to #695 ? Good catch on the sts get-caller.... without output set. Can you please send a PR with that fix? Thanks again!

Kirizan:
@toniblyx What branch would you like me to push the PR to? Also, where else is JSON expected in an aws command? The change to the whoami file might not be the only place the fix is needed. Once my current scan is finished I'll see if that error pops up anywhere else and test it. I'll also see if I can get the environment variable working, that might be the best fix so that it applies everywhere, not just on a few commands.

It's only related to #695 if Casey has changed the default output. Otherwise that issues is seperate.

[...]

PR #697, opened 2020-11-20
Force default AWS CLI output issue #696

Kirizan:
Added two chunks of code to the main prowler file, the first one will set the `AWS_DEFAULT_OUTPUT` environment variable to `json`, the second chunk resets it back to the value when the script began. This will ensure any awscli command that is not overriding output with the `--output` parameter, will have a JSON output.


toniblyx:
Hi @Kirizan, sorry it took more time than expected to review this one. I think adding it to https://github.com/toniblyx/prowler/blob/master/include/whoami#L32 would be suffice for now, otherwise to make that global change might be a bit too broad change and requires more testing. Wdyt?

Kirizan:
@toniblyx That sounds fine, however you want to fix it will work. This way just ensures any future checks are not broken by a profile changing the output.

PR #710, opened 2020-12-16
Refresh assumed role credentials to avoid role chaining limitations @…

michael-dickinson-sainsburys:
…michael-dickinson-sainsburys

Refresh assumed role credentials to avoid role chaining limitations @michael-dickinson-sainsburys


toniblyx:
I thought that epoch time in linux is different than bsd, doesn't make sesn and it is a bit weird. Linux ` date -d 2020-12-17T11:59:42 +%s` results `1608206382` and bsd/mac/darwin using the same date and time `date -j -f %Y-%m-%dT%H:%M:%S 2020-12-17T11:59:42 +%s` results `1608202782`, a difference of 3600 seconds, 1h, between linux and bsd. So makes `$AWS_SESSION_EXPIRATION < $SESSION_CUTOFF` be different depending on OS. I'm testing the bsd one with something like

```
bsd_convert_date_to_timestamp() {
  echo $(( $(date -j -f %Y-%m-%dT%H:%M:%S "$1" +%s) + 3600 ))
}
```

toniblyx (approved):

Issue #711, opened 2020-12-17
Permission Issues Resolving Previous Findings in Security Hub


gyrospectre:
I'm running Prowler via an ECS task in Fargate, using an assumed task role with [all the right permissions](https://github.com/toniblyx/prowler/blob/master/iam/prowler-security-hub.json). New findings are getting added to Security Hub fine, via the native SH support (i.e. -S), but existing findings that have since been resolved are failing to update with a permissions error:

```An error occurred (AccessDeniedException) when calling the BatchImportFindings operation: User: <redacted> is not authorized to perform: securityhub:BatchImportFindings on resource: arn:aws:securityhub:ap-southeast-2::product/prowler/prowler```

Even though that role definitely has that permission. After some debug, it looks like the `resolveSecurityHubPreviousFails()` function in `include/securityhub_integration` is using `batch-import-findings`, rather than the `batch-update-findings` I would expect. The permissions error seems to be a red herring, and the issue is you cannot update existing findings in this way, using the batch import. I'm not sure if this recently changed with AWS?

Has anyone else experienced this issue? I'm playing with a patch to move this function over to `batch-update-findings`, so if this is not just something silly I'm doing and is actually a bug, happy to cut a PR for the fix.

toniblyx:
Not sure if it is related to this same issue here https://github.com/toniblyx/prowler/issues/705 that I'm working on it. But also, since a month or so, Prowler is an official integrated product to Security Hub so you have to enable it first on each account region you need. See updated documentation here https://github.com/toniblyx/prowler#security-hub-integration. Do you think your issue is related to that?

gyrospectre:
Thanks for the response! I did have a look at #705 , but my issue is with a single account, so don't believe it's related. Security Hub appears to be setup properly, I can get findings into SH from Prowler without issue, it just won't update (Archive) existing issues if they are fixed after initially being raised. After patching the code per my original post, everything is working fine (using `batch-update-findings` instead) - I'm out of the woods now, but if others are also experiencing the same issues I wanted to share my fixes in case it helps others.

toniblyx:
Do you mean changing call here https://github.com/toniblyx/prowler/blob/master/include/securityhub_integration#L60 from import to update?

toniblyx:
For some reason seems like the fargate execution role is not being assigned properly and it can't do `BatchImportFindings`. Can you debug that in CloudTrail to see what is it trying to do? it is suspicious to get that `AccessDeniedException`.

[...]

Issue #705, opened 2020-12-06
[help wanted] Is it possible to push finding to SecurityHub by assuming role from central account ?


bkdinoop:
#### Description

We are trying to run security assessment using **_prowler_** on multiple accounts from a central account and push it as finding to security hub of the respective account. We are running prowler in a Fargate from central account and sends the Target account id, IAM role and other parameters as arguments of the docker.

Unfortunately, Prowler is trying to sent the finding to central account's Security hub where the fargate runs. But what i was expecting is to be sent to the securityhub of the target account as it has already assumed to that account.

#### To Reproduce

Docker file I have used is refered from [prowler/utils](https://github.com/toniblyx/prowler/blob/master/util/Dockerfile)

```dockerfile
FROM alpine:latest

ARG USERNAME=prowler
ARG USERID=34000

RUN addgroup -g ${USERID} ${USERNAME} && \
    adduser -s /bin/sh -G ${USERNAME} -D -u ${USERID} ${USERNAME} && \
    apk --update --no-cache add python3 py3-pip bash curl jq file coreutils git && \
    pip3 install --upgrade pip && \
    pip3 install awscli boto3 detect-secrets && \
    git clone https://github.com/toniblyx/prowler

WORKDIR /prowler

RUN chown -R prowler . 

ENTRYPOINT ["./prowler"]
```

Invoking fargate task by sending prowler argument as docker's argument.

```python
ecsClient = boto3.client("ecs")
resp = ecsClient.run_task(
        cluster=FARGATE_CLUSTER,
        launchType="FARGATE",
        taskDefinition=FARGATE_TASK_DEF_NAME,
        platformVersion='LATEST',
        overrides={
            'containerOverrides': [
                {
                    "command": ["-A", "111111111", "-R", "SECAUDITROLE", "-S", "-M","json-asff","-b", "-q","-g", "group4", "-f", AWS_REGION_NAME],
                    "name": "prowler-sechub-task",                    
                }
            ]
        },
        networkConfiguration={
            'awsvpcConfiguration': {
                'subnets': SUBNETS,
                'assignPublicIp': "DISABLED",
                'securityGroups': [SG_ID],
            },
        }
    )
```

**_SECUDITROLE_** has following permission to push to security hub
```
- securityhub:GetFindings
- securityhub:BatchImportFindings
```

And the FARGATE Task role has required permission to assume secudit role.

##### OUTPUT:

[...]

PR #712, opened 2020-12-18
Use batch-update-findings for issue updates

gyrospectre:

Update the way previously raised, now fixed, findings are updated by Prowler. Rather than re-import, which seems to cause problems when using ECS/Fargate, use the `batch-update-findings` action to update the finding.

### DIFF
--- a/include/securityhub_integration
+++ b/include/securityhub_integration
@@ -57,8 +57,10 @@ resolveSecurityHubPreviousFails(){
       for i in `seq 0 100 $FINDINGS_COUNT`; 
       do 
         BATCH_FINDINGS=$(echo $SECURITY_HUB_PREVIOUS_FINDINGS | jq -c '.['"$i:$i+100"']')
-        BATCH_IMPORT_RESULT=$($AWSCLI securityhub --region "$regx" $PROFILE_OPT batch-import-findings --findings "${BATCH_FINDINGS}")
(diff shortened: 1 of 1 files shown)


toniblyx:
if you look at https://docs.aws.amazon.com/securityhub/latest/partnerguide/partnerguide.pdf:
```
Does the partner update findings by calling the BatchImportFindings operation with the existing
findings' finding ID?
To update a finding, you must overwrite the existing finding by submitting the same finding ID.
The BatchUpdateFindings operation should only be used by customers.
Does the partner update findings using the BatchUpdateFindings API?
If you take action on findings, you can use the BatchUpdateFindings operation to update specific
fields
```

So the error should come from other place. We are testing and releasing soon a template that will allow to use fargate for multiaccount, look at this one as an example https://github.com/cdemko-mwr/prowler-multiaccount/

toniblyx:
I said this already but can you also review this issue comments? Not related? https://github.com/toniblyx/prowler/issues/705

toniblyx:
Can you also make sure you have Prowler integration enabled in both master and child accounts?

gyrospectre:
Per my comments in #711, this is an issue with a single account, not multiaccount - I don't believe this issue is related to #705. Prowler integration is indeed enabled on the account in question. Happy for you to reject this PR if it's not correct and a better fix is on the way?

gyrospectre:
Closing this pull request per concerns above.

PR #714, opened 2020-12-24
Fix - EKS California

njgibbon:
Hello,

As of September 2020 EKS is available in the California region and so there no longer needs to be a condition to check region within the EKS related checks. 

AWS Announcement: https://aws.amazon.com/about-aws/whats-new/2020/09/amazon-eks-now-available-n-california-region/

us-west-1 eks region ref: https://docs.aws.amazon.com/general/latest/gr/eks.html

**License**


toniblyx:
Thanks @njgibbon

PR #718, opened 2021-01-12
Fix for busybox date command

zfLQ2qx2:

@toniblyx I noticed when using the -A -R -T parameters with Prowler running on Alpine Linux 3.12.3 (which uses busybox) there is an reoccurring error in the Prowler output because the date command doesn't understand RFC3339 format dates by default - but date can parse them if you combine the -D parameter with the -d parameter. Really odd that the date command is one of the most non-standard between different OS, you think that would be a solved problem after sixty years.

### DIFF
--- a/include/os_detector
+++ b/include/os_detector
@@ -109,7 +109,11 @@ bsd_get_iso8601_timestamp() {
 }
 
 gnu_convert_date_to_timestamp() {
-  date -d "$1" +%s
+  if [ "$OSTYPE" == "linux-musl" ]; then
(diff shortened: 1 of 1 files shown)


toniblyx:
Thanks! 

I got stuck also with something similar here https://github.com/toniblyx/prowler/blob/master/include/os_detector#L121 timestamp differs between linux and bsd !?

jspiewak:
Could this be reverted for now to remove the errors?

toniblyx:
Yes, working on it. Hold a moment please.

toniblyx:
done @jspiewak, get latest from master.

PR #719, opened 2021-01-12
Additional check for location of awscli

zfLQ2qx2:

@toniblyx I've encountered the most bizarre situation on MacOS. The aws client is in my path, works when I run it, but the which command doesn't report its location. It is not an alias, hash bang is valid, python interpreter is valid. The type command does report its location, so this patch adds it as a fallback in case someone else has the same issue. Both type and which are bash builtins, so this should be very low overhead.

### DIFF
--- a/include/awscli_detector
+++ b/include/awscli_detector
@@ -12,8 +12,11 @@
 # specific language governing permissions and limitations under the License.
 
 # AWS-CLI detector variable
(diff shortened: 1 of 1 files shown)


toniblyx:
Awesome, I didn't know about that `type -p` trick. Thanks!

PR #721, opened 2021-01-12
Handle shadow cloudtrails more gracefully

zfLQ2qx2:

@toniblyx This changes checks 21, 22, 24, and 27 so that checks for CloudTrail attributes are done in all regions, but are only reported in the "home" region of the trail. This keeps multi-region and organization CloudTrails from appearing multiple times in the check output but doesn't miss regional trails. I've also changed the checks to use TrailARN instead of Name when referencing CloudTrails so that the checks work for organization level CloudTrails.

### DIFF
--- a/checks/check21
+++ b/checks/check21
@@ -21,27 +21,28 @@ CHECK_ASFF_COMPLIANCE_TYPE_check21="ens-op.acc.7.aws.iam.1,ens-op.mon.1.aws.trai
 check21(){
   trail_count=0
   # "Ensure CloudTrail is enabled in all regions (Scored)"
-	for regx in $REGIONS; do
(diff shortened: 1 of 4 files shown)


zfLQ2qx2:
@toniblyx Hold off on this one a second - works from cli, works from docker, but fails in an ECS environment. I need to figure out what is happening in that case.

zfLQ2qx2:
@toniblyx nm, ecs policy didn't allow ec2:describeregions which is an edge case that impacted a lot of things, code works fine once that is fixed.

toniblyx:
your change made in check27 didn't work and I have fixed it here https://github.com/toniblyx/prowler/commit/b23f9b3b5d9fcb7641e3cefd93812c50bbecbc14

PR #726, opened 2021-01-14
Check for errors generating credential report, limit loop iterations

zfLQ2qx2:

@toniblyx This is to address the endless loop in genCredReport reported in issue 617

### DIFF
--- a/include/credentials_report
+++ b/include/credentials_report
@@ -14,9 +14,19 @@
 # Generate Credential Report
 genCredReport() {
   textTitle "0.1" "Generating AWS IAM Credential Report..." "NOT_SCORED" "SUPPORT"
-  until $( $AWSCLI iam generate-credential-report --output text --query 'State' $PROFILE_OPT --region $REGION |grep -q -m 1 "COMPLETE") ; do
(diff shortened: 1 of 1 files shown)


toniblyx:
hey @zfLQ2qx2 please reach out to me [here](https://blyx.com/contacto/) there are some issues with recent changes for busybox and date command that are impacting other functions from your recent changes and I'd like to clarify. Thanks in advance!

zfLQ2qx2:
@toniblyx Oops! Feel free to revert those changes in the mean time if they are causing more harm than good, they are working well for me locally and I can patch them in locally until I work out a version that works for everyone.

While I've not seen any issues I can guess that something was depending on date autodetecting one of the non-RFC3339 formats that date supports. I'm not a Python guy (and I love that so much of Prowler is in bash because I can tell junior people to go look at the code and they can understand it) but since Python is already installed for Prowler maybe the most universal solution is to have a python script to parse dates into epoch format rather then jumping through hoops with all the differing OS implementations.

toniblyx:
I agree with you on the date handling pain with bash. It is a pain with python as well but less for sure :D

PR #727, opened 2021-01-14
fix regex in check43

ilyas28:
Let's say you have a rule like 10.0.0.0/8 in a default security group, the check fails when it should not.

The output from aws ec2 command is like this:

```
IPPERMISSIONS   2181    tcp     2181
IPRANGES        10.0.0.0/8
IPPERMISSIONS   9092    tcp     9092
IPRANGES        10.0.0.0/8
```

Simply changing the expression by adding a space fixes this problem
```
egrep ' 0.0.0.0|\:\:\/0'
```


toniblyx:
Good catch @ilyas28. Thanks!

PR #729, opened 2021-01-14
Add access checks for several checks

zfLQ2qx2:

@toniblyx The other issue raised in issue 617 is that the reporter was using a policy that didn't allow cloudtrail describe-trails and that led to unexpected results. That seems to be a common theme in the issues section and I think the only real fix is to be defensive and look for access denied errors whenever an aws cli command is run. This adds a number of AccessDenied checks in the check_2x checks and a couple extra_7xx checks that happen to call describe-trails.

### DIFF
--- a/checks/check21
+++ b/checks/check21
@@ -23,7 +23,11 @@ check21(){
   trail_count=0
   # "Ensure CloudTrail is enabled in all regions (Scored)"
   for regx in $REGIONS; do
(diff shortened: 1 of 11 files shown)


toniblyx (changes requested):
@zfLQ2qx2 thanks for adding this better way of handling AccessDenied. I added CHECK_SERVICENAME variable to each check, could you add it to the checks before I merge it? See any check in master for reference.

zfLQ2qx2:
> @zfLQ2qx2 thanks for adding this better way of handling AccessDenied. I added CHECK_SERVICENAME variable to each check, could you add it to the checks before I merge it? See any check in master for reference.

@toniblyx More to come! I pulled your latest changes and rebased, I spot checked three of them and the CHECK_SERVICENAME_ tags are now there, so I think this is ready for merging.

PR #754, opened 2021-03-11
Include missing AWS function policy to check AWS Lambda

jfagoagas:
Hi!

There is a missing function in `prowler-additions-policy.json`. 

To check AWS Lambda related controls included in Prowler: 

- extra720
- extra759
- extra760
- extra762
- extra798

it is necessary to allow in `prowler-additions-policy.json` the AWS function `lambda:GetFunction` otherwise it will fail with the following error:
```
An error occurred (AccessDeniedException) when calling the GetFunction operation: User: arn:aws:sts::<account-id>:assumed-role/<role>/botocore-session-<ID> is not authorized to perform: lambda:GetFunction on resource: arn:aws:lambda:<region>:<account-id>:function:<lambda-function-name>
``` 

### DIFF
--- a/iam/prowler-additions-policy.json
+++ b/iam/prowler-additions-policy.json
@@ -9,7 +9,8 @@
                 "ec2:GetEbsEncryptionByDefault",
                 "ecr:Describe*",
                 "support:Describe*",
-                "tag:GetTagKeys"
+                "tag:GetTagKeys",
+		"lambda:GetFunction"
             ],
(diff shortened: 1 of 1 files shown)


toniblyx:
You are totally right @jfagoagas. lambda:GetFunction is not either included in AWS managed policies SecurityAudit and ViewOnlyAccess. Thanks for the heads up!

PR #756, opened 2021-03-12
Add python2.7 to obsolete Lambda runtimes

mpratsch:
Since python version 2.7 is already deprecated we should also cover that within the obsolete Lambda runtimes check. 

### DIFF
--- a/checks/check_extra762
+++ b/checks/check_extra762
@@ -23,7 +23,7 @@ extra762(){
 
   # regex to match OBSOLETE runtimes in string functionName%runtime
   # https://docs.aws.amazon.com/lambda/latest/dg/runtime-support-policy.html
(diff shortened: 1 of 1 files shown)


toniblyx:
Thanks for the suggestion @mpratsch, however despite Python 2.7 reached end of life on 2020, it will remain actually supported, see note here https://docs.aws.amazon.com/lambda/latest/dg/runtime-support-policy.html.

mpratsch:
> 
> 
> Thanks for the suggestion @mpratsch, however despite Python 2.7 reached end of life on 2020, it will remain actually supported, see note here https://docs.aws.amazon.com/lambda/latest/dg/runtime-support-policy.html.

Yes you are right. Last year the announcement was EOL by end of 2020 for the lambda runtime as well but I didn't see they have extended to 1st of June.
https://aws.amazon.com/blogs/compute/continued-support-for-python-2-7-on-aws-lambda/

I'm going to close this PR.

PR #772, opened 2021-04-07
Score in html

HG00:

This adds a really simple score output to html. Since we dont know the score until we're done, it adds a placeholder which is then replaced on completion.


HG00:
Sorry @toniblyx , looks like this is a bit further ahead than i thought - is it better to a different branch than master?

toniblyx:
Thanks, that sed -i command only works for linux and not on bsd sed (osx) I created a couple of functions to make it easier and also had to make other changes in that sed logic. I will push the changes to 2.4 in a moment. Now the html report will show scoring even if you don't use `-s` however you will still need `-s` to see scoring in the default output.

PR #784, opened 2021-04-21
Fix for Security Hub integration with custom region

reka193:
If the findings are required to be sent to Security Hub, the resolveSecurityHubPreviousFails function is iterating through all regions, even if a region is set (with -r), and it throws the following error for all the regions where Security Hub is not enabled:
```
An error occurred (InvalidAccessException) when calling the GetFindings operation: Account xxxxxxxxxx is not subscribed to AWS Security Hub

Error parsing parameter '--findings': Expected: '=', received: 'EOF' for input:

^

 ERROR! Failed to send check output to AWS Security Hub
```

This pull request solves this issue by using only the selected region.

### DIFF
--- a/include/securityhub_integration
+++ b/include/securityhub_integration
@@ -42,7 +42,14 @@ checkSecurityHubCompatibility(){
 
 resolveSecurityHubPreviousFails(){
   # Move previous check findings RecordState to ARCHIVED (as prowler didn't re-detect them)
-  for regx in $REGIONS; do
+  if [ -z "$REGION_OPT" ]
+  then
+    REGIONS_SECHUB=$REGIONS
+  else
+    REGIONS_SECHUB=$REGION_OPT
+  fi
+
+  for regx in $REGIONS_SECHUB; do
 
     local check="$1"
     NEW_TIMESTAMP=$(get_iso8601_timestamp)
@@ -54,8 +61,8 @@ resolveSecurityHubPreviousFails(){
 
     if [[ $SECURITY_HUB_PREVIOUS_FINDINGS != "[]" ]]; then
       FINDINGS_COUNT=$(echo $SECURITY_HUB_PREVIOUS_FINDINGS | jq '. | length')
-      for i in `seq 0 100 $FINDINGS_COUNT`; 
-      do 
+      for i in `seq 0 100 $FINDINGS_COUNT`;
+      do
         BATCH_FINDINGS=$(echo $SECURITY_HUB_PREVIOUS_FINDINGS | jq -c '.['"$i:$i+100"']')
         BATCH_IMPORT_RESULT=$($AWSCLI securityhub --region "$regx" $PROFILE_OPT batch-import-findings --findings "${BATCH_FINDINGS}")
         if [[ -z "${BATCH_IMPORT_RESULT}" ]] || jq -e '.FailedCount >= 1' <<< "${BATCH_IMPORT_RESULT}" > /dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


toniblyx:
Hi @reka193, thanks for the heads up, however to filter by region and send findings to the proper region Security Hub you have to use `-f region` as mentioned in the documentation: https://github.com/toniblyx/prowler#security-hub-integration

reka193:
Hi @toniblyx, thanks for the quick reply.
In my case, I would like to use prowler for cross-account checks in different regions and send the results to Security Hub.
If I use filtering for e.g. eu-centrail-1, I'm receiving an error for us-east-1 (I believe the first region in the list), unless I set up the region in the .aws/config file, which is not the best idea in this case.
Do you have any recommendations how it could be done?

reka193:
Hi @toniblyx, thanks for the quick reply, my bad for using the wrong switch.
In my case, I would like to use prowler for cross-account checks in different regions and send the results to Security Hub.
But if I use filtering for e.g. eu-central-1, I'm receiving an error for us-east-1, unless I set up the region in the .aws/config file, which is not the best idea in this case.

```
An error occurred (AccessDeniedException) when calling the BatchImportFindings operation: User: arn:aws:sts::account:role/role/ProwlerAssessmentSession is not authorized to perform: securityhub:BatchImportFindings on resource: arn:aws:securityhub:us-east-1::product/prowler/prowler

 ERROR! Failed to send check output to AWS Security Hub
```

Do you have any recommendations how it could be done?

toniblyx:
If I understand correctly you want to check many accounts and many regions and send all of them to the "same" Security Hub? In Security Hub (regardless Prowler), you have to send findings per region and per account to a master account where there will be everything centralized. Therefore, if you use Prowler integrated to Security Hub you have to scan each region per account and send findings to each region and account Security Hub, then Security Hub will send them to the master account. I hope that is more clear now.

reka193:
Thanks, that I understand, I am trying to do it this way. In general, the problem I am facing now is independent from the cross-account usage.

As far as I got with the investigation, the problem is with the specific check that I was trying to run. In check15, the region is not passed to the textPass/textFail functions, which then runs sendToSecurityHub without the specified region.

Modifying check15 to the following seems to solve this problem, but correct me if I misunderstood something.

```
check15(){
  # "Ensure IAM password policy requires at least one uppercase letter (Scored)"
  COMMAND15=$($AWSCLI iam get-account-password-policy $PROFILE_OPT --region $REGION --output json --query 'PasswordPolicy.RequireUppercaseCharacters' 2> /dev/null) # must be true
  if [[ "$COMMAND15" == "true" ]];then
    textPass "Password Policy requires upper case" "$FILTERREGION"
  else
    textFail "Password Policy missing upper-case requirement" "$FILTERREGION"
  fi
}
```

toniblyx:
check15 is part of the CIS AWS benchmark and that is included in Security Hub (you don´t need Prowler for that).

reka193:
Thank you, I am aware of that. We wanted to run a selected number of checks using prowler, and e.g. extra7100 produces the same error which is not part of the CIS benchmark. (But in general any similar check which doesn't pass on the filtered region.)

It can also be fixed similarly by passing on the $FILTERREGION variable, but it is not part of this pull request anymore.

reka193:
Since we are still planning to perform these checks using prowler, not the CIS benchmark (mainly to skip Config costs), would you have any recommendations to the above mentioned problem? Thank you!

toniblyx:
It is planned to consolidate all checks using the same format as the ones in Extras, to make sure any check works the same way regardless if it comes from CIS or not, that is planned but not sure when it will be implemented. We can try for 2.5 but not sure. The idea is to include also the resource name or ID in the textFail and textPass as variable and leave it available for json, csv, json-asff, etc. that will help for auto-remediation triggers, etc.

PR #800, opened 2021-05-20
ALB Header Check

Outrun207:
ALB header check written and tested. ALBs without attribute to drop invalid headers will be marked as 'fail'.


toniblyx (changes requested):
Thanks @Outrun207, this looks great. The format for textPass, textFail and textInfo is like this example taken from extra7174:
`textFail "$regx: Potential secret found SSM Document $ssmdoc" "$regx" "$ssmdoc"`
see regx at the end and also resource id.
Thanks!

Outrun207 (commented):
Added $regex and resource to text outputs

PR #822, opened 2021-06-17
Add ResourceID to all checks output for ASFF and other output formats

singergs:
Here are updates to prowler checks that support passing the resource name in with the third parameter.   For some resources that dont have regions hardcoded us-east-1 as the second parameter.


toniblyx (changes requested):
Thanks @singergs, I see you have hardcoded region `us-east-1` in IAM related checks. That will make GovCloud or China to fail, it is better to leave $REGION that will take us-east-1 if no region is used or the one related to GovCloud or China if specified with `-r` that is required for those partitions.

PR #836, opened 2021-07-29
Add glue permissions

joshes:
Adding necessary IAM permissions to run OOTB WRT glue checks

```
glue:GetConnections
glue:SearchTables
```

Without these an OOTB run of `prowler` does not succeed due to `AccessDeniedException` due to these permissions.


joshes:
I was only confirming this via the util/codebuild CloudFormation template, but noticed the bootstrap IAM was doing something similar so added the permissions there as well.

Feedback welcome on that. 

Thanks!

toniblyx:
Thanks @joshes, I did that (partially) in branch 2.5 (current devel one) but I just noticed that I missed glue:SearchTables. I will add it. Thank you!

PR #849, opened 2021-08-18
Fix extra737 remove false positives due to policies with condition @rinaudjaws

rinaudjaws:
CDK for example implements callerAccount as a condition for the KMS policy resulting in too many false positives.


toniblyx:
Thank you @rinaudjaws

PR #851, opened 2021-08-19
Ignore archived findings in GuardDuty for check extra7139 @chbiel

chbiel:
As the check should only look at active findings, not those that were moved to the archive of guardduty.
It does not make sense to include the archived findings as this is the guardduty way of saying "I accept this finding".


w0rmr1d3r (approved):
Looks good to me, let's wait for @toniblyx and his approval

PR #856, opened 2021-08-26
Fix bug in extra784 @tayivan-sg

tayivan-sg:
The evaluation of the if statement in extra 784 should be flopped around. 
Should fail if there are internal users instead.


toniblyx:
Thanks @tayivan-sg, I'll review this next week.

w0rmr1d3r (approved):
Looks good to me!

toniblyx:
Thanks for the fix @tayivan-sg

PR #860, opened 2021-09-01
Added feature to allow role ARN while using -R parameter @mmuller88 #859

### LINKED ISSUE #859: R can take role arn as well
As I develop cdk-prowler https://github.com/mmuller88/cdk-prowler and currently work on multi accounts. For the prowler master role you would need to set it up in every account with like using stack set which is currently not supported by CDK. It would be massively conviently if we could instead just pass the cross account role per the R parameter as well eg: 

./prowler -A 111222333 -R arn:aws:iam::999999:role/ProwlerAudit . 

I will open a PR soon to show you the necessary small changes. 

Thanks :)

mmuller88:
fixes: #859

Tested it on my workbranch and it works like a charm https://github.com/mmuller88/cdk-prowler/blob/feature/7_multi/src/index.ts#L168 .

I don't need to bother with stack sets and only give the role arn for R which allows cross account access

### DIFF
--- a/include/assume_role
+++ b/include/assume_role
@@ -26,22 +26,29 @@ assume_role(){
     # temporary file where to store credentials
     TEMP_STS_ASSUMED_FILE=$(mktemp -t prowler.sts_assumed-XXXXXX)
 
+    # check if role arn or role name
+    if [[ $ROLE_TO_ASSUME == arn:* ]]; then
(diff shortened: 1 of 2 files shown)


toniblyx:
Thanks  @mmuller88, I'll review this PR next week.

toniblyx:
thanks @mmuller88

PR #861, opened 2021-09-01
Terraform Kickstarter @singergs

singergs:

The following supports the creation of a Prowler resources using terraform.  The modules support Prowler running inside an AWS Codebuild project with the results being sent to AWS SecurityHub.

The terraform will activate SecurityHub along with the integration to Prowler.  See readme.md for all resources created when deploying.


toniblyx:
Thank you @singergs, I'll review this next week.

terraform-kickstarter/prowler_build_spec.yml
```diff
@@ -0,0 +1,25 @@
...
+      - curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip"
+      - unzip awscliv2.zip
+      - ./aws/install
+      - git clone https://github.com/toniblyx/prowler
+      
+  build:
+    commands:
+      - echo "Running Prowler as ./prowler $PROWLER_OPTIONS"
+      - cd prowler
+      - git checkout 2.5
```
w0rmr1d3r: I'd move these lines, together with the `git clone` from above to the pre-build phase instead.
At the same time, why are we forcing 2.5, what will happen when a new version appears?

terraform-kickstarter/main.tf
```diff
@@ -0,0 +1,535 @@
...
+  default     =  "-q -M json-asff -S -f us-east-1"
+}
+variable "prowler_schedule"{
+  description = "Run Prowler based on cron schedule"
+  default="cron(0 0 ? * * *)"
+  type=string
+
+}
+
+variable "codebuild_timeout" {
```
w0rmr1d3r: should we separate the variables and data to other files?

terraform-kickstarter/prowler_build_spec.yml
singergs: Thanks for the comment and made the changes as suggested.

terraform-kickstarter/main.tf
singergs: I have this as a TODO.  Thanks for the comment!

singergs: Ok i was able to break out the various components.

terraform-kickstarter/prowler_build_spec.yml
w0rmr1d3r: perfect!

terraform-kickstarter/main.tf
w0rmr1d3r: oh! nice! I saw the changes, much better now!

toniblyx:
Awesome work @singergs, thank you!

PR #862, opened 2021-09-03
Fix support policy arn in check120 @hersh86

hersh86:

For situations when the account has copied the default AWS policy but kept the name this would fail as the for loop variable wasn't in the right place. Working as expected on ubuntu 20.04.


w0rmr1d3r (approved):
looks good to me

toniblyx:
Thanks @hersh86

PR #866, opened 2021-09-08
New check 7148 EFS File systems have backup enabled @georgie969

georgie969:
By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license. This Pull request is called [extra7148] to Check if EFS has backup enabled


w0rmr1d3r (commented):
I do have a question, what if AWS Backup is doing backups for an EFS instead of the EFS having backups enabled, can that be possible?
What would happen to that check if it is possible?

georgie969:
_"I do have a question, what if AWS Backup is doing backups for an EFS instead of the EFS having backups enabled, can that be possible? What would happen to that check if it is possible?"_

----------

**If AWS Backup is doing backups for EFS, The check still reports the same information. I have tested this scenario by creating a backup via AWS Backup, using EFS as the resource. After the Backup was created for that specific File system using AWS Backup, its recognized that the specific File system has a backup policy attached and the same principles apply.**

w0rmr1d3r (approved):
Careful, modify the PR title, since it's refering to check 7148, but it's actually doing 7149 !

georgie969:
Yes, Thank you. I made an error and submitted a different Custom Check 7149 which is for a different service "Redshift" under the same pull request. I will remove that. Thank you

w0rmr1d3r (approved):
looks good to me :) nice work!

georgie969:
Hi @w0rmr1d3r I made an error and submitted two different Prowler checks on the same Pull request. Can you kindly roll back the approvals so I can make two separate branches and two new pull requests? One for the EFS Prowler Check and another for the Redshift Prowler Check? Thank you

georgie969:
@w0rmr1d3r  Okay. I have deleted the 7149 Redshift Check and This Prowler check 7148 is good to go! Thank you!

toniblyx:
Good job! Thanks!

georgie969:
> Good job! Thanks!

You are welcome! Much appreciation!

PR #869, opened 2021-09-14
New check 7150 ELB deletion protection enabled @ManuelUgarte

ManuelUgarte:

This Pull request is called [extra7150] to Check if ELB has deletion protection attribute is enabled.


checks/check_extra7150
```diff
@@ -0,0 +1,44 @@
...
+CHECK_CAF_EPIC_extra7150='Data Protection'
+
+extra7150(){
+  # "Check if Elastic Load Balancers have delete protection enabled."
+  for regx in $REGIONS; do
+    LIST_OF_ELBSV2=$($AWSCLI elbv2 describe-load-balancers $PROFILE_OPT --region $regx --query 'LoadBalancers[*].LoadBalancerArn' --output text|xargs -n1)
+    if [[ $LIST_OF_ELBSV2 ]]; then
+      for elbarn in $LIST_OF_ELBSV2; do
+        CHECK_DELETION_PROTECTION_ENABLED=$($AWSCLI elbv2 describe-load-balancer-attributes $PROFILE_OPT --region $regx --load-balancer-arn $elbarn --query Attributes[*] --output text|grep -B 1 "deletion_protection.enabled" | grep true)
+        ELBV2_NAME=$(echo $elbarn|cut -d\/ -f3)
```
w0rmr1d3r: Doing this inside the for-loop might take a lot of time, do we have a workaround to do a for-loop but with the data already obtained?

checks/check_extra7150
```diff
@@ -0,0 +1,44 @@
...
+          textPass "$regx: $ELBV2_NAME has the attribute deletion protection enabled" "$regx" "$elb"
+        else
+          textFail "$regx: $ELBV2_NAME does not have deletion protection enabled." "$regx" "$elb"
+        fi
+      done
+    else
+      textInfo "$regx: No ELBs found" "$regx"
+    fi
+  done
+}
```
w0rmr1d3r: EOL

checks/check_extra7150
ManuelUgarte: Did you mean for the data already obtained via `describe-load-balancers`? The attributes do not show there unfortunately. For line 32, since there are multiple versions of ELBv2 (NLB, ALB and Gateway LB) the response/output varies depending on which one it is. The [response/output](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/elbv2/describe-load-balancer-attributes.html) is a list of attributes but the order of those varies in the response depending on the type of ELB and therefore I can't extract the boolean directly. That is why I decided to use grep there to determine the boolean expression. The line 33 is just to make it easier to read for the user instead of returning the full Arn but also because the `describe-load-balancer-attributes` needs the full arn and can't use only the name of the ELB to run.

ManuelUgarte: For reference, it took a second or so to show data from my 3 deployed ELBs in us-east-1. Please let me know if you have an idea of a workaround I could use.

ManuelUgarte: Also, I did notice I can remove` -B 1` from `grep -B 1 "deletion_protection.enabled"` so I will make that update too before the next update. I didnt say it before but thank you for the feedback!

checks/check_extra7150
ManuelUgarte: Will add. Thanks!

w0rmr1d3r: thank you!

checks/check_extra7150
w0rmr1d3r: Sorry for the delay!
Thank you for your answer! It looks good to me :)

w0rmr1d3r (approved):
Looks good to me!

toniblyx (approved):

PR #877, opened 2021-09-21
New check 7154 CloudFormation stack termination protection enabled @ShubhamShah11

ShubhamShah11:
Adding a check to verify if CloudFormation stacks have termination protection enabled.
 

checks/check_extra7154
```diff
@@ -0,0 +1,56 @@
...
+  for regx in $REGIONS; do
+    CFN_STACKS=$($AWSCLI cloudformation describe-stacks $PROFILE_OPT --region $regx --output json)
+    LIST_OF_CFN_STACKS=$(echo $CFN_STACKS | jq -r '.Stacks[].StackName')
+    if [[ $LIST_OF_CFN_STACKS ]];then
+      for stack in $LIST_OF_CFN_STACKS; do
+        CFN_STACK_DETAILS=$($AWSCLI cloudformation describe-stacks $PROFILE_OPT --region $regx --stack-name $stack --output json)
+        TERMINATION_ENABLED=$(echo $CFN_STACK_DETAILS | jq -r '.Stacks[].EnableTerminationProtection')
+        ROOT_ID=$(echo $CFN_STACK_DETAILS | jq -r '.Stacks[].RootId')
+        if [[ $ROOT_ID != null && $TERMINATION_ENABLED == "false" ]]; then
+          textFail "$regx: $stack is a nested stack, enable termination protection on the root stack $ROOT_ID" "$regx" "$stack" "$ROOT_ID"
```
w0rmr1d3r: Question here, for a nested stack, this check will output 2 failures, 1 for the root stack and another for the nested one.
By fixing the root stack, it will fix both failures at the same time.
What if this was an info for nested stacks? Since for the root stack will fail as well, by fixing the root, this will transform from an info to a pass, instead of from a failure to a pass, since only a change on the root stack is needed.
@toniblyx any thoughts on this?

ShubhamShah11: Yes, I agree ! Info can be displayed for nested stack. Will update as per final confirmation.

ShubhamShah11: Updated the PR with Info message for Nested Stack. Thanks

w0rmr1d3r (approved):
Looks good to me!
@toniblyx the group7 is being deleted here, is that alright?

toniblyx:
extras group shouldn´t be deleted however considering that are a bunch of checks coming up, there are conflicts so I will update it later

PR #892, opened 2021-10-07
Revised integration with Wazuh 4.2.x

ns2switch:

- Revised integration with wazuh 4.2.x
- Support multiaccount for wazuh.
- Just a zone for command in wazuh. It can be modified in the future adding a zone.lst

### DIFF
--- a/account.lst
+++ b/account.lst
@@ -1,0 +1,2 @@
+#Add accounts numbers you wanna monitor.
+XXXXXXXXXXXX
--- a/include/colors
+++ b/include/colors
@@ -14,12 +14,12 @@
 
 IFS=',' read -ra MODES <<< "${MODE}"
 for MODE in "${MODES[@]}"; do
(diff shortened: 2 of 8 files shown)


toniblyx:
Thanks for the PR @ns2switch however I don't understand why you do need to change files include/colors, include/outputs and include/whoami, you are actually using the default JSON output, why not to get from the outputs dir and then process it with the wrapper, etc? There is not a real need of creating a new output since the json output is already in place, isn't it?

ns2switch:
Hi, the wrapper takes the information directly from stdout , and if there is text in the stdout the json processor in wrapper fails.
I have change the output file to eliminate the text output in the prowler command.
I added the wazuh options ( in the include/whoami) to not to altered the json output, or if its need to be changed in the future.
Also , I can drop the integration module , and do everything in output , so prowler can talk directly with wazuh

toniblyx:
@ns2switch As per our confersations offline I´ll close this PR and waiting for the other one using syslog integration. Thanks!

PR #896, opened 2021-10-15
Fix check extra734 about S3 buckets default encryption with StringNotEquals @rustic

rustic:
A fix to check for StringNotEquals rather than StringEquals to prevent false positives.


toniblyx:
Thanks @rustic!

PR #907, opened 2021-11-03
Option to send prowler results to syslog. Improving wazuh support with decoder and new rules

ns2switch:


toniblyx:
Thanks @ns2switch, I think there is some confusion here, I'll discuss it with you offline anyway but wanted leave here a comment. This is an integration an doesn't require any new output format so we could treat it as it is and can make it even easier.

PR #911, opened 2021-11-05
New check 7158 ELBV2 has listeners underneath @kbgoll05

kbgoll05:
Adding a check to verify if ELBV2 load balancers have listerners underneath.

### DIFF
--- a/checks/check_extra7158
+++ b/checks/check_extra7158
@@ -1,0 +1,43 @@
+#!/usr/bin/env bash
+
+# Prowler - the handy cloud security tool (copyright 2018) by Toni de la Fuente
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may not
+# use this file except in compliance with the License. You may obtain a copy
+# of the License at http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software distributed
+# under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+# CONDITIONS OF ANY KIND, either express or implied. See the License for the
+# specific language governing permissions and limitations under the License.
+CHECK_ID_extra7158="7.158"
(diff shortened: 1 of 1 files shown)


toniblyx:
Hey @kbgoll05 thanks for these new checks. Why to have them in separate checks instead of having both in the same one?

kbgoll05:
> Hey @kbgoll05 thanks for these new checks. Why to have them in separate checks instead of having both in the same one?

Hey @toniblyx, Since the AWS CLI separates Classic and other load balancers as **_elb_** and **_elbv2_** respectively, I felt it would be better and easier to manage the security checks if they are separated as two different checks.

Also, fetching the listeners for classic load balancers(**_elb_**) is quite different when compared with other load balancers(**_elbv2_**). The former uses **_describe-load-balancers_** API call and then query down to listeners section while the latter uses **_describe-listeners_** API call. 

Moreover, classic load balancer's(**_elb_**)  **_describe-load-balancers_** API call can take the load balancer name directly through **_--load-balancer-name_** flag. Whereas **_describe-listeners_** API call of **_elbv2_** take ARN as an input. Additionally we have to strip the ARN to get the name out of it for elbv2 load balancers.

So, to reduce complexity in the prowler checks and to keep it simple, I felt making it as two different checks would be a better approach.

toniblyx:
Thanks for the comprehensive response @kbgoll05, I'm going to go ahead and merge them.

toniblyx:
My apologies @kbgoll05 I merged them to master by mistake and that's why I reverted them. Would you please mind to send the PR to the branch 2.5.1 instead of master? I totally forgot to change it by myself and screwed it up. Thanks! Same for both PRs.

kbgoll05:
> My apologies @kbgoll05 I merged them to master by mistake and that's why I reverted them. Would you please mind to send the PR to the branch 2.5.1 instead of master? I totally forgot to change it by myself and screwed it up. Thanks! Same for both PRs.

Yeah Sure. I apologize for creating a pull request to the wrong branch.

PR #921, opened 2021-11-10
New check 7159 ELB Classic Load balancer has listeners underneath @kbgoll05

kbgoll05:
Adding a check to verify if ELB Classic Load Balancers have listerners underneath.


toniblyx:
Awesome, thank you @kbgoll05!!

PR #932, opened 2021-11-16
New check 7160 Redshift cluster AutomaticVersionUpgrade enabled @jonloza

jonloza:
Check if RedShift has automatic upgrades enabled


toniblyx:
Thanks @jonloza!! would you mind to change the case to `Redshift` instead of `RedShift`? 
Also you missed `$PROFILE_OPT` on each cli command, look at this one for reference: https://github.com/toniblyx/prowler/blob/master/checks/check_extra7149#L29

jonloza:
@toniblyx sorry for late reply .. updated thank you!

PR #935, opened 2021-11-19
New check 7162 CloudWatch log groups have 365 days retention @Obiakara

Obiakara:

### DIFF
--- a/checks/check_extra7162
+++ b/checks/check_extra7162
@@ -1,0 +1,42 @@
+#!/usr/bin/env bash
+
+# Prowler - the handy cloud security tool (copyright 2018) by Toni de la Fuente
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may not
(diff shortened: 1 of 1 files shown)


toniblyx:
Thanks @Obiakara, let me add some comments and change suggestions here:
- For the title probably "Check if CloudWatch Log Groups have a retention policy of 365 days" sounds more comprehensive.
- Resource type is: AwsLogsLogGroup
- Service name is: cloudwatch (I know it is logs but we are consolidating this type of checks in cloudwatch so far).
- Please, for future optional changes put 365 in a variable inside the function. Like LOG_GROUP_RETENTION_PERIOD_DAYS= in case somebody else needs to adapt it is will be easier. Then use that variable in the text messages as well.

Thanks!

Obiakara:
Thanks @toniblyx  for your feedback. I made the changes and commited again

PR #952, opened 2021-12-01
New check 7164 Check if EFS has backup policy enabled @maisenhe

maisenhe:
Check if EFS has backup policy enabled

### DIFF
--- a/checks/check_extra7164
+++ b/checks/check_extra7164
@@ -1,0 +1,54 @@
+#!/usr/bin/env bash
+
+# Prowler - the handy cloud security tool (copyright 2019) by Toni de la Fuente
+#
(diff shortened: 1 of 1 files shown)


toniblyx:
Thanks @maisenhe, that checks is already written, it is actually  `extra7148`.

toniblyx:
We are working on a public list of checks to write to make that easier for the community. Do you want ideas? What about a check to look for `Cloudwatch log group encrypted` or `CloudFront distribution pointing to non existing buckets to prevent subdomain hijacking`?

maisenhe:
@toniblyx when you are talking about log group encryption are you talking about having AWS Key Management Service for this encryption? As Log group data is always encrypted in CloudWatch Logs. Thanks

toniblyx:
exactly, with kms.

PR #953, opened 2021-12-01
Fix issue with Security Hub integration when resolving closed findings are either a lot of new findings, or a lot of resolved findings @Kirizan

### LINKED ISSUE #950: Issues importing to Security Hub when a finding has a lot of results
While running a Prowler Scan in an account with 469 Lambda functions and two Cloud Trails, generating 938 findings for [extra720](https://github.com/toniblyx/prowler/blob/42e54c42cf878eb3928465c981f838765062eb01/checks/check_extra720).

The error message given is:

    prowler/include/securityhub_integration: line 53: /usr/bin/jq: Argument list too long

This line in [securityhub_integration](https://github.com/toniblyx/prowler/blob/42e54c42cf878eb3928465c981f838765062eb01/include/securityhub_integration) is trying to get a list of the prior findings. I believe the error is happening at `--argjson ids "$NEW_FINDING_IDS"` part of the command.

I modified the line to replace the double quotes with single quotes for that part so it looks like `--argjson ids '$NEW_FINDING_IDS'`. I did it based on [this](https://stackoverflow.com/questions/11475221/why-do-i-get-bin-sh-argument-list-too-long-when-passing-quoted-arguments) Stack Overflow question.

I'm guessing this issue will also appear whenever there are a lot of findings. I'm not sure if there is a good way to test the limits of how many findings can be sent at a time, and break them down so Prowler can send in batches based on how many can be sent at once.

Kirizan:
Fixes #950

### DIFF
--- a/include/securityhub_integration
+++ b/include/securityhub_integration
@@ -83,7 +83,7 @@ resolveSecurityHubPreviousFails(){
       done
     fi
   done
-
+  rm -rf "$SH_TEMP_FOLDER"
 }
 
 sendToSecurityHub(){
--- a/include/securityhub_integration
+++ b/include/securityhub_integration
@@ -70,12 +70,15 @@ resolveSecurityHubPreviousFails(){
 
     if [[ $SECURITY_HUB_PREVIOUS_FINDINGS != "[]" ]]; then
       FINDINGS_COUNT=$(echo $SECURITY_HUB_PREVIOUS_FINDINGS | jq '. | length')
-      for i in `seq 0 100 $FINDINGS_COUNT`; 
+      for i in $(seq 0 50 $FINDINGS_COUNT);
       do 
-        BATCH_FINDINGS=$(echo $SECURITY_HUB_PREVIOUS_FINDINGS | jq -c '.['"$i:$i+100"']')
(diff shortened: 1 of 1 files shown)


include/securityhub_integration
```diff
@@ -42,24 +42,43 @@ checkSecurityHubCompatibility(){
 
 resolveSecurityHubPreviousFails(){
   # Move previous check findings RecordState to ARCHIVED (as prowler didn't re-detect them)
+    SH_TEMP_FOLDER="$PROWLER_DIR/SH-$ACCOUNT_NUM"
+    if [[ ! -d $SH_TEMP_FOLDER ]]; then
+      # this folder is deleted once the security hub update is completed
```
toniblyx, line 47: where is the folder deleted? I don't see a `rm` here.

toniblyx (changes requested):

include/securityhub_integration
Kirizan: That was my mistake, I forgot to add that line since I was doing everything in a container, there was nothing left to remind me. I've added an additional commit that adds `rm -rf "$SH_TEMP_FOLDER"` to the end of the `resolveSecurityHubPreviousFails(){` function.

toniblyx (approved):

toniblyx:
Thanks @Kirizan this is a great improvement!

PR #969, opened 2021-12-15
updated alpine, used aws cli v2 instead of v1, and added directory cl…

sectoramen:
Updated alpine, use aws cli v2 instead of v1, and added directory cleanup to reduce image size


sectoramen:
pulling this back as per comments on Discord.

PR #973, opened 2021-12-16
Docker Image

sectoramen:
Updated docker image, awscli v2, cleaner docker image workflow, additional cleanup of the pip cache which further reduces the image size.


sectoramen:
Already fixed

PR #974, opened 2021-12-19
Added -D option to copy to S3 with the initial AWS credentials instead of the assumed as with -B option @sectoramen

sectoramen:
Toni, 

Does it make sense to save the original AWS credentials and then restore them before doing  a copy to S3? I don't see a reason to keep the assumed credentials before the copy. The credentials of the individual/process that launched prowler are most likely to be the one used to perform other tasks, such as copy to S3. Optionally, we could add a new option, i.e., 
-Bn  Custom output bucket used with original prowler role, requires -M <mode> and it can work also with -o flag.
                            (i.e.: -M csv -B my-bucket or -M csv -B my-bucket/folder/)


j2clerck (commented):
I believe there are more cases to cover here, depending on how Prowler is initially getting credentials:
- Instance Profile
- Named Profile
- Environment Variables
- AWS Config / Credentials file
It might be worth reviewing the existing options and test how it behave. Setting environment variable to '' does not seem to bother aws cli. But you need to be able to revert the profile too at least. Happy to do some testing.

toniblyx:
I always have tried to keep Prowler consistent with the way that the cli works in terms of [configuration settings and precedence](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-quickstart.html#cli-configure-quickstart-precedence)

1. Command line options – Overrides settings in any other location. You can specify --region, --output, and --profile as parameters on the command line.
2. Environment variables – You can store values in your system's environment variables.
3. CLI credentials file – ~/.aws/credentials on Linux or macOS, or at C:\Users\USERNAME\.aws\credentials on Windows
4. CLI configuration file – The credentials and config file are updated when you run the command aws configure. The config file is located at ~/.aws/config on Linux or macOS, or at C:\Users\USERNAME\.aws\config on Windows. 
5. Container credentials
6. Instance profile credentials

Additionally, Prowler handles:

7. With credentials gotten from any of the previous 6 steps, Prowler can assume a role that the previous user or role has permission to assume in the same account or another one.

sectoramen:
This is ready to go. It was tested and it appears to work fine.

toniblyx:
Awesome, thanks @sectoramen!

PR #987, opened 2022-01-11
fix(extra761): Match insensitive values

jfagoagas:
### Motivation

A new bug was discovered at https://github.com/toniblyx/prowler/issues/986 because this check performs a case sensitive comparison.

### Description

Match insensitive values using `shopt -s nocasematch`

### License


toniblyx (changes requested):
I would add `--output text` to the main variable and then `True` in the if to avoid adding more commands.

PR #991, opened 2022-01-14
switching from comparing strings  to the integers in assume role

IhorKravchuk:
**Problem:** 
Currently  [[ $MINIMUM_REMAINING_TIME_ALLOWED > $SESSION_TIME_REMAINING ]] is not evaluated as an comparing integers, but as a comparing  string. 
_Reasons:_ 
http://mywiki.wooledge.org/BashFAQ/031 
**why it needs to be fixed:**
let's say "3599" becoming smaller than "600",  causing often calls to the assumed role
These often calls cause the corporate proxy (and potentially AWS API endpoints) to perform throttling and break prowler execution.

**Proposed fix**
we have 2 options:

1. switch to the  ArithmeticExpression using (( instead of  [[
2. use -gt  or -lt with the same [[

In PR I use -gt to keep a uniform approach ([[ ]]]) to all TEST expressions.


toniblyx:
Thanks @IhorKravchuk have you tested the latest devel branch 2.7? That should be fixed there if I´m not mistaken.

IhorKravchuk:
Yep, you are absolutely correct. It's fixed there using ArithmeticExpression . Thank you

Issue #1022, opened 2022-02-02
Add checks for S3 bucket ACLs


jeffmaley:
**Is your feature request related to a problem? Please describe.**
AWS has deprecated S3 bucket ACLs in favor of IAM and bucket policies.
https://docs.aws.amazon.com/AmazonS3/latest/userguide/about-object-ownership.html

**Describe the solution you'd like**
There should be a check for buckets and whether they have ACLs enabled

**Describe alternatives you've considered**
None

**Additional context**
I've submitted a PR for this.

jfagoagas:
@jeffmaley can we close this issue?

jeffmaley:
Yep.

PR #1023, opened 2022-02-02
feat(check): New check7172 for S3 Bucket ACLs

jeffmaley:
### Context 

https://github.com/prowler-cloud/prowler/issues/1022

### Description

This pr adds a check (7172) to look for bucket ACLs

### License

### DIFF
--- a/checks/check_extra7172
+++ b/checks/check_extra7172
@@ -25,12 +25,11 @@ CHECK_CAF_EPIC_extra7172='Logging and Monitoring'
 
 extra7172(){
   # "Check if S3 buckets have server access logging enabled "
-  ACCESS_CHECK=$($AWSCLI s3api list-buckets $PROFILE_OPT --output text 2>&1)
-  if [[ $(echo "$ACCESS_CHECK" | grep -E 'AccessDenied|UnauthorizedOperation|AuthorizationError') ]]; then
+  LIST_OF_BUCKETS=$($AWSCLI s3api list-buckets $PROFILE_OPT --query Buckets[*].Name --output text 2>&1)
+  if [[ $(echo "$LIST_OF_BUCKETS" | grep -E 'AccessDenied|UnauthorizedOperation|AuthorizationError') ]]; then
     textInfo "$REGION: Access Denied Trying to list buckets" "$REGION"
     exit
   fi
-  LIST_OF_BUCKETS=$($AWSCLI s3api list-buckets $PROFILE_OPT --query Buckets[*].Name --output text|xargs -n1 2>&1)
   if [[ $LIST_OF_BUCKETS ]]; then
     for bucket in $LIST_OF_BUCKETS;do
       BUCKET_ACLS_DISABLED=$($AWSCLI s3api get-bucket-ownership-controls --bucket $bucket $PROFILE_OPT --output text 2>&1)
--- a/checks/check_extra7172
+++ b/checks/check_extra7172
@@ -25,11 +25,16 @@ CHECK_CAF_EPIC_extra7172='Logging and Monitoring'
 
 extra7172(){
   # "Check if S3 buckets have server access logging enabled "
-  LIST_OF_BUCKETS=$($AWSCLI s3api list-buckets $PROFILE_OPT --query Buckets[*].Name --output text|xargs -n1)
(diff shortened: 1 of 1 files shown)


checks/check_extra7172
```diff
@@ -0,0 +1,49 @@
...
+      elif [[ $(echo "$BUCKET_ACLS_DISABLED" | grep "BucketOwnerPreferred") ]]; then
+        textFail "$REGION: Bucket $bucket has bucket ACLs enabled!" "$REGION" "$bucket"
+      elif [[ $(echo "$BUCKET_ACLS_DISABLED" | grep "OwnershipControlsNotFoundError") ]]; then
+        textFail "$REGION: Bucket $bucket has bucket ACLs enabled!" "$REGION" "$bucket"
+      else
+        textFail "$REGION: Bucket $bucket returned unknown error" $BUCKET_ACLS_DISABLED
+      fi
+    done
+  else
+    textInfo "$REGION: No S3 Buckets found" "$REGION" "$bucket"
```
lazize: Variable `$bucket` doesn't exist inside this `else` statement.

checks/check_extra7172
```diff
@@ -0,0 +1,49 @@
...
+CHECK_ALTERNATE_check7172="extra7172"
+CHECK_SERVICENAME_extra7172="s3"
+CHECK_RISK_extra7172='S3 ACLs are a legacy access control mechanism that predates IAM. IAM and bucket policies are currently the preferred methods.'
+CHECK_REMEDIATION_extra7172='Ensure that S3 ACLs are disabled (BucketOwnerEnforced). Use IAM policies and bucket policies to manage access.'
+CHECK_DOC_extra7172='https://docs.aws.amazon.com/AmazonS3/latest/userguide/about-object-ownership.html'
+CHECK_CAF_EPIC_extra7172='Logging and Monitoring'
+
+extra7172(){
+  # "Check if S3 buckets have server access logging enabled "
+  LIST_OF_BUCKETS=$($AWSCLI s3api list-buckets $PROFILE_OPT --query Buckets[*].Name --output text|xargs -n1)
```
lazize: Consider add AccessDenied verification here as well

lazize:
```shell
  LIST_OF_BUCKETS=$($AWSCLI s3api list-buckets $PROFILE_OPT --region $REGION --query "sort_by(Buckets, &Name)[].Name" --output text 2>&1)
  if [[ $(echo "$LIST_OF_BUCKETS" | grep -E 'AccessDenied|UnauthorizedOperation|AuthorizationError') ]]; then
      textInfo "$REGION: Access Denied trying to list buckets" "$REGION"
      return
  fi
```

checks/check_extra7172
jeffmaley: I've removed it.

checks/check_extra7172
jeffmaley: I've added those extra errors to the handling and added an extra check for list-buckets access.

lazize:
Why execute two CLI commands if one is enough?

Variable `LIST_OF_BUCKETS` can be used for both cases. Access check and loop iteration.  
You just need to remove `| xargs -n1`.

`| xargs -n1` will give a better view of the output, but as it will be used by checking error and to perform `for` loop iteration, it doesn't need a better view. Both cases work fine with a "not so beautiful" output.
Variable `LIST_OF_BUCKETS` is not used to output details to user.

jeffmaley:
My thought was to keep the two bits separate. If we need to change the access check or bucket controls check, we don't have to refactor the other piece. If you'd rather they be combined, I will make the change.

lazize:
There is a throttling that impacts Prowler when it is running inside AWS, so I believe fewer request is better.
https://docs.aws.amazon.com/AWSEC2/latest/APIReference/throttling.html

But let's wait maintainer people to comment.

toniblyx:
I agree with @lazize the less the better regardless of cli and boto3 underneath.

toniblyx:
and thank you @jeffmaley for the check!

jeffmaley:
I've condensed the access check into the s3 check itself and removed the extra api call.

toniblyx (approved):
LGTM

PR #1036, opened 2022-02-08
fix(extra736): Recover only Customer Managed KMS keys

jfagoagas:
### Context 

Previously, check `extra736` only checked if KMS Keys **with an alias configured** were publicly exposed.

### Description

Right now, this control recovers every KMS Key at the account and checks the `KeyManager` property to extract only the policy for the ones not managed by AWS. So, with that KMS Keys without an alias configured are also audited.

### License


toniblyx (approved):
LGTM

n4ch04 (approved):

PR #1045, opened 2022-02-15
fix(extra765/776): Add right region to CSV if access is denied

roman-mueller:
### Context 

If access is denied on checks extra765 or extra776 currently no region is specified.
In the CSV output those checks then default to a different/wrong region. 

### Description

Add region parameter to `textInfo` call. 

### License


toniblyx (approved):

toniblyx (approved):

toniblyx:
thanks @roman-mueller

PR #1053, opened 2022-03-03
fix: extra7167 Advanced Shield and CloudFront bug parsing None output without distributions

NMuee:
### Context 

My AWS account is encountering false positive with extra7167
We do not have any cloudfront distribution, yet they are being flagged as 
"Cloudfront distribution None is not protected by AWS Shield Advanced"
due to `aws cloudfront list-distributions --profile $profile --query 'DistributionList.Items[*].[Id,ARN]' --output text` is outputting as 'None' even thought there is no cloudfront distribution.

### Description

In the if statement, Instead of detecting if there is any value in the variable $LIST_OF_CLOUDFRONT_DISTRIBUTIONS, I updated the code with `$LIST_OF_CLOUDFRONT_DISTRIBUTIONS != 'None'`

### License

### DIFF
--- a/checks/check_extra7167
+++ b/checks/check_extra7167
@@ -27,7 +27,7 @@ CHECK_CAF_EPIC_extra7167='Infrastructure security'
 extra7167() {
   if [[ "$($AWSCLI $PROFILE_OPT shield get-subscription-state --output text)" == "ACTIVE" ]]; then
     LIST_OF_CLOUDFRONT_DISTRIBUTIONS=$($AWSCLI cloudfront list-distributions $PROFILE_OPT --query 'DistributionList.Items[*].[Id,ARN]' --output text)
-    if [[ $LIST_OF_CLOUDFRONT_DISTRIBUTIONS ]]; then
+    if [[ $LIST_OF_CLOUDFRONT_DISTRIBUTIONS != "None"]]; then
       while read -r distribution; do
         DISTRIBUTION_ID=$(echo $distribution | awk '{ print $1; }')
         DISTRIBUTION_ARN=$(echo $distribution | awk '{ print $2; }')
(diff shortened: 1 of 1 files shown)


checks/check_extra7167
```diff
@@ -27,7 +27,7 @@ CHECK_CAF_EPIC_extra7167='Infrastructure security'
 extra7167() {
   if [[ "$($AWSCLI $PROFILE_OPT shield get-subscription-state --output text)" == "ACTIVE" ]]; then
     LIST_OF_CLOUDFRONT_DISTRIBUTIONS=$($AWSCLI cloudfront list-distributions $PROFILE_OPT --query 'DistributionList.Items[*].[Id,ARN]' --output text)
-    if [[ $LIST_OF_CLOUDFRONT_DISTRIBUTIONS ]]; then
+    if [[ $LIST_OF_CLOUDFRONT_DISTRIBUTIONS != "None"]]; then
```
toniblyx: And in that case, if the variable is empty we can just check the if like:
```suggestion
    if [[ $LIST_OF_CLOUDFRONT_DISTRIBUTIONS ]]; then
```

checks/check_extra7167
```diff
@@ -27,7 +27,7 @@ CHECK_CAF_EPIC_extra7167='Infrastructure security'
 extra7167() {
   if [[ "$($AWSCLI $PROFILE_OPT shield get-subscription-state --output text)" == "ACTIVE" ]]; then
     LIST_OF_CLOUDFRONT_DISTRIBUTIONS=$($AWSCLI cloudfront list-distributions $PROFILE_OPT --query 'DistributionList.Items[*].[Id,ARN]' --output text)
```
toniblyx: have you test it when the variable is empty? if empty it will go thru the if and shouldn't. So better  to add at the end of line 29 something like: 
```suggestion
    LIST_OF_CLOUDFRONT_DISTRIBUTIONS=$($AWSCLI cloudfront list-distributions $PROFILE_OPT --query 'DistributionList.Items[*].[Id,ARN]' --output text | grep -v None)
```

toniblyx (changes requested):

checks/check_extra7167
NMuee: Have updated the code and tested on my side, works good, thankyouuu

toniblyx:
Awesome, thanks @NMuee!

toniblyx (approved):

PR #1061, opened 2022-03-08
fix(whitelist): Whitelist logic reformulated again

n4ch04:
### Context 

 Current whitelisting logic did not detect whitelisted element if  includes POSIX character classes 

### Description

Reformulate whitelisting logic to detect POSIX character classes  into whitelisted element

### License

### DIFF
--- a/include/outputs
+++ b/include/outputs
@@ -169,16 +169,17 @@ textFail(){
   colorcode="$BAD"
   while read -r excluded_item; do
     ignore_check_name="${excluded_item%%:*}" # Check name is everything up to the first :
-    # Resource value is the second field of line included in whitelist divided by :
-    resource_value=$(awk -F ":" '{print $2}' <<< $excluded_item)
+    # Resource value is what it comes after CHECK_NAME: :
+    resource_value=$(awk -F "$CHECK_NAME:" '{print $2}' <<< "${excluded_item}")
     # Checked value is the whole log message that comes as argument
     checked_value=$1
     if [[ "${ignore_check_name}" != "${CHECK_NAME}" ]]; then
       # not for this check
       continue
     fi
-    # To set WARNING flag both values must be exactly the same
-    if [[ "${checked_value}" == *"${resource_value}"* ]]; then
+    # To set WARNING flag checked_value have to include value of resource_value
+    # If it is treated as only expanse (*[]*) will not detect regex like [:alpha:]
+    if [[ "${checked_value}" =~ ${resource_value} ]]; then
       level="WARNING"
       colorcode="${WARNING}"
       break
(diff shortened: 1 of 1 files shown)


include/outputs
```diff
@@ -169,16 +169,17 @@ textFail(){
   colorcode="$BAD"
   while read -r excluded_item; do
     ignore_check_name="${excluded_item%%:*}" # Check name is everything up to the first :
-    # Resource value is the second field of line included in whitelist divided by :
-    resource_value=$(awk -F ":" '{print $2}' <<< $excluded_item)
+    # Resource value is what it comes after CHECK_NAME: :
+    resource_value=$(awk -F "$CHECK_NAME:" '{print $2}' <<< "${excluded_item}")
```
ghost, line 174: Just a suggestion, use the same method as "ignore_check_name" above? ie `resource_value=${excluded_item#*:}`.

include/outputs
```diff
@@ -169,16 +169,17 @@ textFail(){
...
     checked_value=$1
     if [[ "${ignore_check_name}" != "${CHECK_NAME}" ]]; then
       # not for this check
       continue
     fi
-    # To set WARNING flag both values must be exactly the same
-    if [[ "${checked_value}" == *"${resource_value}"* ]]; then
+    # To set WARNING flag checked_value have to include value of resource_value
+    # If it is treated as only expanse (*[]*) will not detect regex like [:alpha:]
+    if [[ "${checked_value}" =~ ${resource_value} ]]; then
```
ghost, line 183: This look good! Thanks for sorting it out :)

ghost (dismissed):

MrCloudSec (approved):
Thank you @wobblesprout

include/outputs
n4ch04: Thanks for your suggestion ! It's the same coded in a different way
We have scheduled the reformulation of whitelisting logic, including this, we just have changed ignore_check_name to follow the same syntax as resource_value, please check it again and let us know if works like it should

include/outputs
n4ch04: You're welcome !!

include/outputs
wobblesprout: Setting the delimiter to the name of the check in the awk command feel iffy to me - but if it works ... meh.

Personally I prefer using native bash when you can and not relying on external commands if you get the same result. While no one writes bash scripts with performance and resource usage being their main focus, it should at least be a consideration. Extra fork/exec/etc can start to add up, especially when we are running on a CI/CD pipeline that shares resources with other jobs. But again, this is only a suggestion - feel free to ignore.

jfagoagas (approved):
Good job @n4ch04 

🔝

PR #1063, opened 2022-03-08
fix(extra776): Handle image tag commas and json output

jfagoagas:
### Context 

We need to fix `extra776` in order not to break `csv` output format.

### Description

- Handle ECR images with multiple tag separated by commas only recovering last occurrence with `head -n 1`
- Use `--output json` when needed otherwise it fails if your default output is `text`
- Included a more granular handling for authentication errors.
- More comments

### License


toniblyx (approved):
LGTM

PR #1067, opened 2022-03-14
Fix(check122): Error when policy name contains commas

plarso:
### Context 

AWS IAM policy names allow commas (','), but check 122 does not support the case where there is a comma in the policy name.

### Description

* Change the awk command for the POLICY_VERSION variable in the script to get the version number from the last column rather than the second column.
* Change the awk command for the POLICY_ARN variable in the script to get the policy ARN from all columns except the last instead of just the first column.

### License

### DIFF
--- a/checks/check122
+++ b/checks/check122
@@ -30,8 +30,8 @@ check122(){
   LIST_CUSTOM_POLICIES=$($AWSCLI iam list-policies --output text $PROFILE_OPT --region $REGION --scope Local --query 'Policies[*].[Arn,DefaultVersionId]' | grep -v -e '^None$' | awk -F '\t' '{print $1","$2"\n"}')
   if [[ $LIST_CUSTOM_POLICIES ]]; then
     for policy in $LIST_CUSTOM_POLICIES; do
-      POLICY_ARN=$(echo $policy | awk 'BEGIN{FS=OFS=","}{NF--; print}')
-      POLICY_VERSION=$(echo $policy | awk -F ',' '{print $(NF)}')
+      POLICY_ARN=$(awk 'BEGIN{FS=OFS=","}{NF--; print}' <<< "${policy}")
+      POLICY_VERSION=$(awk -F ',' '{print $(NF)}' <<< "${policy}")
       POLICY_WITH_FULL=$($AWSCLI iam get-policy-version --output text --policy-arn $POLICY_ARN --version-id $POLICY_VERSION --query "[PolicyVersion.Document.Statement] | [] | [?Action!=null] | [?Effect == 'Allow' && Resource == '*' && Action == '*']" $PROFILE_OPT --region $REGION)
       if [[ $POLICY_WITH_FULL ]]; then
         POLICIES_ALLOW_LIST="$POLICIES_ALLOW_LIST $POLICY_ARN"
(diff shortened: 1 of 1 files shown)


checks/check122
```diff
@@ -30,8 +30,8 @@ check122(){
   LIST_CUSTOM_POLICIES=$($AWSCLI iam list-policies --output text $PROFILE_OPT --region $REGION --scope Local --query 'Policies[*].[Arn,DefaultVersionId]' | grep -v -e '^None$' | awk -F '\t' '{print $1","$2"\n"}')
   if [[ $LIST_CUSTOM_POLICIES ]]; then
     for policy in $LIST_CUSTOM_POLICIES; do
-      POLICY_ARN=$(echo $policy | awk -F ',' '{print $1}')
-      POLICY_VERSION=$(echo $policy | awk -F ',' '{print $2}')
+      POLICY_ARN=$(echo $policy | awk 'BEGIN{FS=OFS=","}{NF--; print}')
```
jfagoagas:
```suggestion
      POLICY_ARN=$(awk 'BEGIN{FS=OFS=","}{NF--; print}' <<< "${policy}")
```

checks/check122
```diff
@@ -30,8 +30,8 @@ check122(){
...
+      POLICY_VERSION=$(echo $policy | awk -F ',' '{print $(NF)}')
```
jfagoagas:
```suggestion
      POLICY_VERSION=$(awk -F ',' '{print $(NF)}' <<< "${policy}")
```

jfagoagas (changes requested):
Thanks @plarso, please review comments above to merge this PR!

lazize:
If it has comma, it will break csv output.
If it is allowed by AWS, we should replace it by something else.

jfagoagas:
> If it has comma, it will break csv output. If it is allowed by AWS, we should replace it by something else.

Good point! I'm going to test a initial approach to solve this.

jfagoagas:
Hi @plarso could you include the following as a replace in line 42?
```
textFail "$REGION: Policy ${policy//,/[comma]} allows \"*:*\"" "$REGION" "$policy"
```

This is a valid approach but we are modifying role name in the output.

1. AWS Test Roles
Screenshot 2022-03-15 at 17 55 30
2. Prowler Result
Screenshot 2022-03-15 at 18 08 43


Also @lazize please check this.

Thanks!

checks/check122
plarso: Suggestion implemented in new commit.

checks/check122
plarso: Suggestion implemented in new commit.

plarso:
I've added the requested changes.

lazize:
Based on my recent experience, I believe this kind of variable expansion will not be supported by old version of bash, like the one original from macOS.
I will test it tomorrow.

jfagoagas:
> I've added the requested changes.

Cool, thank you @plarso !

jfagoagas:
> Based on my recent experience, I believe this kind of variable expansion will not be supported by old version of bash, like the one original from macOS.
> 
> I will test it tomorrow.

All of this changes has been tested with `Bash 3.15`, this kind of variable expansion is supported in old versions of Bash.

jfagoagas (approved):
Good job @plarso !!

Thanks 👏

Issue #887, opened 2021-09-30
Getting error while running prowler for multiple regions in a account


anuj1366:
I am running Prolwer through AWS Code Build for multiple regions in a single AWS account and below command is getting executed:
**-r eu-central-1 -f 'eu-central-1 eu-west-1' -M junit-xml,json-asff -S** 

**Now I am getting below error:**
Access Denied trying to describe regions! Review permissions as described here: https://github.com/toniblyx/prowler/#requirements-and-installation
974 |  
975 | [Container] 2021/09/30 08:48:50 Command did not exit successfully ./prowler $PROWLER_OPTIONS exit status 1
976 | [Container] 2021/09/30 08:48:50 Phase complete: BUILD State: FAILED
977 | [Container] 2021/09/30 08:48:50 Phase context status code: COMMAND_EXECUTION_ERROR Message: Error while executing command: ./prowler $PROWLER_OPTIONS. Reason: exit status 1

By analyzing the AWS cloudtrail, I have figure out like Prowler is sending filter regions with single quote attached.
![image](https://user-images.githubusercontent.com/91662053/135421267-5040c1d9-8c04-465d-9ca9-782a22703cbc.png)

Could you please suggest how to run Prowler for multiple regions.
Note: I have tried to run the prowler for multiple regions without giving single quote for filter region like below:
**-r eu-central-1 -f eu-central-1 eu-west-1 -M junit-xml,json-asff -S** 
But its considering only eu-central-1.

toniblyx:
Hi @anuj1366, this is a matter of how to escape those quotes in bash. Are you using the CFN template provided in the repo or your own?

Kirizan:
@anuj1366 For codebuild, put the region list in double quotes, not single quotes. Try this command: `-r eu-central-1 -f "eu-central-1 eu-west-1" -M junix-xml,json-asff -S`

anuj1366:
> @anuj1366 For codebuild, put the region list in double quotes, not single quotes. Try this command: `-r eu-central-1 -f "eu-central-1 eu-west-1" -M junix-xml,json-asff -S`

@Kirizan I have already tried to give multiple regions in double quotes but it didn't worked and getting below error.
![image](https://user-images.githubusercontent.com/91662053/135870929-2090fcd7-0d77-4dc1-9ee7-d3b5e5b141e8.png)

anuj1366:
[...]

PR #1071, opened 2022-03-16
fix(filter-region): Support comma separated regions

thetemplateblog:
### Context 

When filtering on regions, the results are not as expected. None of the workarounds from https://github.com/prowler-cloud/prowler/issues/887 worked in my case. I have found that if regions are comma delimited, the filter works as intended. The problem with using single quotes is that the single quotes are passed into $FILTERREGION and the aws command line won't work like that.

### Description

Using a simple bash replace character on line 344 of the prowler bash script allows the -f value to be delimited using a comma. 

### License

### DIFF
--- a/README.md
+++ b/README.md
@@ -230,7 +230,7 @@ Prowler has been written in bash using AWS-CLI underneath and it works in Linux,
 
 By default, Prowler scans all opt-in regions available, that might take a long execution time depending on the number of resources and regions used. Same applies for GovCloud or China regions. See below Advance usage for examples.
 
-Prowler has two parameters related to regions: `-r` that is used query AWS services API endpoints (it uses `us-east-1` by default and required for GovCloud or China) and the option `-f` that is to filter those regions you only want to scan. For example if you want to scan Dublin only use `-f eu-west-1` and if you want to scan Dublin and Ohio `-f eu-west-1,us-east-s`, note the regions are separated by a comma deliminator.
+Prowler has two parameters related to regions: `-r` that is used query AWS services API endpoints (it uses `us-east-1` by default and required for GovCloud or China) and the option `-f` that is to filter those regions you only want to scan. For example if you want to scan Dublin only use `-f eu-west-1` and if you want to scan Dublin and Ohio `-f eu-west-1,us-east-s`, note the regions are separated by a comma deliminator (it can be used as before with `-f 'eu-west-1,us-east-1'`).
 
 ## Screenshots
 
--- a/README.md
+++ b/README.md
@@ -230,7 +230,7 @@ Prowler has been written in bash using AWS-CLI underneath and it works in Linux,
 
 By default, Prowler scans all opt-in regions available, that might take a long execution time depending on the number of resources and regions used. Same applies for GovCloud or China regions. See below Advance usage for examples.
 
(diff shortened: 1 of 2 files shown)


toniblyx:
Hi @thetemplateblog, this is a change that will break the existing users that have it automated with `-f 'region1 region2'` format, we will add that to v3. Could you please explain with more detail how are you running the command to understand why that format is not valid for you?

README.md
```diff
@@ -230,7 +230,7 @@ Prowler has been written in bash using AWS-CLI underneath and it works in Linux,
 
 By default, Prowler scans all opt-in regions available, that might take a long execution time depending on the number of resources and regions used. Same applies for GovCloud or China regions. See below Advance usage for examples.
 
-Prowler has two parameters related to regions: `-r` that is used query AWS services API endpoints (it uses `us-east-1` by default and required for GovCloud or China) and the option `-f` that is to filter those regions you only want to scan. For example if you want to scan Dublin only use `-f eu-west-1` and if you want to scan Dublin and Ohio `-f 'eu-west-1 us-east-s'`, note the single quotes and space between regions.
+Prowler has two parameters related to regions: `-r` that is used query AWS services API endpoints (it uses `us-east-1` by default and required for GovCloud or China) and the option `-f` that is to filter those regions you only want to scan. For example if you want to scan Dublin only use `-f eu-west-1` and if you want to scan Dublin and Ohio `-f eu-west-1,us-east-s`, note the regions are separated by a comma deliminator.
```
toniblyx:
```suggestion
Prowler has two parameters related to regions: `-r` that is used query AWS services API endpoints (it uses `us-east-1` by default and required for GovCloud or China) and the option `-f` that is to filter those regions you only want to scan. For example if you want to scan Dublin only use `-f eu-west-1` and if you want to scan Dublin and Ohio `-f eu-west-1,us-east-s`, note the regions are separated by a comma deliminator (it can be used as before with `-f 'eu-west-1,us-east-1'`).
```

toniblyx (changes requested):

README.md
```diff
@@ -230,7 +230,7 @@ Prowler has been written in bash using AWS-CLI underneath and it works in Linux,
 
 By default, Prowler scans all opt-in regions available, that might take a long execution time depending on the number of resources and regions used. Same applies for GovCloud or China regions. See below Advance usage for examples.
 
-Prowler has two parameters related to regions: `-r` that is used query AWS services API endpoints (it uses `us-east-1` by default and required for GovCloud or China) and the option `-f` that is to filter those regions you only want to scan. For example if you want to scan Dublin only use `-f eu-west-1` and if you want to scan Dublin and Ohio `-f 'eu-west-1 us-east-s'`, note the single quotes and space between regions.
+Prowler has two parameters related to regions: `-r` that is used query AWS services API endpoints (it uses `us-east-1` by default and required for GovCloud or China) and the option `-f` that is to filter those regions you only want to scan. For example if you want to scan Dublin only use `-f eu-west-1` and if you want to scan Dublin and Ohio `-f eu-west-1,us-east-s`, note the regions are separated by a comma deliminator (it can be used as before with `-f 'eu-west-1,us-east-1'`).
```
jfagoagas:
```suggestion
Prowler has two parameters related to regions: `-r` that is used query AWS services API endpoints (it uses `us-east-1` by default and required for GovCloud or China) and the option `-f` that is to filter those regions you only want to scan. For example if you want to scan Dublin only use `-f eu-west-1` and if you want to scan Dublin and Ohio `-f eu-west-1,us-east-1`, note the regions are separated by a comma deliminator (it can be used as before with `-f 'eu-west-1,us-east-1'`).
```

toniblyx (approved):

jfagoagas:
Thank you @thetemplateblog, good catch!! 👏

PR #1080, opened 2022-03-24
fix(secrets_library): Verify if detect-secrets library is missing

MrCloudSec:
### Context 

At this moment, Prowler doesn't detect that detect-secrets is missing when running the checks extra741, extra742, extra759, extra760, extra768, extra775 and extra7141.

It handles the missing dependency as a FAIL within the check.

### Description

The exit code when the library is not installed was changed to 241, so the checks can handle it apart by printing an INFO when these checks are run.

### License

### DIFF
--- a/checks/check_extra7141
+++ b/checks/check_extra7141
@@ -31,32 +31,37 @@ extra7141(){
   fi
 
   for regx in ${REGIONS}; do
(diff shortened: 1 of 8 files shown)


jfagoagas (approved):
Great!!

PR #1085, opened 2022-03-29
New Custom Check - extra9999

sectoramen:
### Context 

This new check extra9999 can be used to build a custom check on the fly. This will help with edge and special cases and when a check does not exist in prowler. It will also help those that are familiar with the AWS CLI but do not have the time or do not know how to create new prowler checks. 

You can build your own on-the-fly custom check by specifying the AWS CLI command to execute. The prowler commands requires  "-c extra9999 -a <aws_cli_cmd>"

Note: Omit the actual "aws" command and only use its parameters within quotes. Nesting quotes in the aws parameter may results in errors, therefore do not use quotes inside the parameter. Note:  --output text is already included in the check and can be omitted from the "aws" command parameter list

Here is an example of a check to find SGs with inbound port 80:

./prowler -c extra9999 -a 'ec2 describe-security-groups --filters Name=ip-permission.to-port,Values=80 --query SecurityGroups[*].GroupId[]]'     

### Description

### License


jfagoagas:
Hi @sectoramen, please review this PR commit history because it contains some other commits related to another security check https://github.com/prowler-cloud/prowler/pull/1095. I think we only need the ones related to check9999.

Thank you.

jfagoagas:
Hi @sectoramen, I have just seen that you merged master branch into this PR. You need to remove the commits that do not belong to this PR because some of it contains a duplicate check.

sectoramen:
I submitted a new pull request https://github.com/prowler-cloud/prowler/pull/1103

PR #1099, opened 2022-04-07
check empty array in SECURITYGROUPS object

nealalan:
DESCRIPTION
Logic was checking an object to see if it is null. This should be checking for the array in the object to see if it is empty.

BEFORE:
7.5 [extra75] Ensure there are no Security Groups not being used - ec2 [Informational]
jq: error (at :1): null (null) has no keys
jq: error (at :1): null (null) has no keys
jq: error (at :1): null (null) has no keys
jq: error (at :1): null (null) has no keys
jq: error (at :1): null (null) has no keys
PASS! eu-west-1: sg-00002cf0000000000 is being used

AFTER:
7.5 [extra75] Ensure there are no Security Groups not being used - ec2 [Informational]
INFO! af-south-1: No Security Groups found in af-south-1
INFO! eu-north-1: No Security Groups found in eu-north-1
INFO! ap-south-1: No Security Groups found in ap-south-1
INFO! eu-west-3: No Security Groups found in eu-west-3
INFO! eu-west-2: No Security Groups found in eu-west-2
PASS! eu-west-1: sg-00002cf0000000000 is being used

Context
I was using this check and was confused as to why only 100 results were returned. I noticed this error was also an issue.

License

### DIFF
--- a/checks/check_extra75
+++ b/checks/check_extra75
@@ -34,9 +34,6 @@ extra75(){
         textInfo "$regx: Access Denied trying to describe security groups" "$regx"
         continue
     fi
-    if [[ $SECURITYGROUPS == "null" ]]; then
-      continue
-    fi
     if [[ $(echo "$SECURITYGROUPS" | jq '."SecurityGroups" | length') -eq 0 ]]; then
       textInfo "$regx: No Security Groups found in $regx"
       continue
--- a/checks/check_extra75
+++ b/checks/check_extra75
@@ -37,6 +37,10 @@ extra75(){
     if [[ $SECURITYGROUPS == "null" ]]; then
       continue
     fi
+    if [[ $(echo "$SECURITYGROUPS" | jq '."SecurityGroups" | length') -eq 0 ]]; then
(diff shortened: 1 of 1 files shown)


checks/check_extra75
```diff
@@ -37,6 +37,10 @@ extra75(){
     if [[ $SECURITYGROUPS == "null" ]]; then
       continue
     fi
```
jfagoagas: I think this comparison could be removed because the next is the replacement for this, am I right?

We no longer need to compare if `SECURITYGROUPS` object is a null object.

MrCloudSec: I would remove it too.

MrCloudSec (approved):
LGTM, thank you @nealalan, it was a good catch!

checks/check_extra75
nealalan: @jfaroagas I went back and forth on removing it and decided to let it be reviewed.

jfagoagas: Thank you!!

PR #1103, opened 2022-04-11
feat(new): New custom check extra9999 to build a custom check on the fly

sectoramen:
### Context 

This new check extra9999 can be used to build a custom check on the fly. This will help with edge and special cases and when a check does not exist in prowler. It will also help those that are familiar with the AWS CLI but do not have the time or do not know how to create new prowler checks.

You can build your own on-the-fly custom check by specifying the AWS CLI command to execute. The prowler commands requires "-c extra9999 -a <aws_cli_cmd>"

Note: Omit the actual "aws" command and only use its parameters within quotes. Nesting quotes in the aws parameter may results in errors, therefore do not use quotes inside the parameter. Note: --output text is already included in the check and can be omitted from the "aws" command parameter list

Here is an example of a check to find SGs with inbound port 80:

./prowler -c extra9999 -a 'ec2 describe-security-groups --filters Name=ip-permission.to-port,Values=80 --query SecurityGroups[*].GroupId[]]'

### Description

### License


toniblyx (dismissed):
LGTM, we have to add that usage part to the README.md though

MrCloudSec:
Added Readme

jfagoagas (approved):
Great feature!!

Thanks @sectoramen

PR #1124, opened 2022-05-02
check_extra7113: Fix wrong listing of RDS instances in regions without databases

Sinnohd:
### Context 

Remove listing of databases in wrong regions.

### Description

This simple fix, removes the listing of a database from one region (as example database exists in eu-central-1), in another region (as example in region eu-west-1, which is enabled, but didn't host a single database).

### License

### DIFF
--- a/checks/check_extra7113
+++ b/checks/check_extra7113
@@ -52,7 +52,7 @@ extra7113(){
         fi
       done
     else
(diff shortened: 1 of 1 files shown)


jfagoagas (approved):
Thank you @Sinnohd, really good catch here!

PR #1128, opened 2022-05-05
feat(assume-role): Properly handle External ID variable

chrisdlangton:
### Context 

3.0-dev branch has an alternative patch

for master branch, the external id is simply broken. this quick change of 2 characters keeps all existing code as-is and ensure the master branch functions as was always intended

### Description

The fix is that the `--external-id` just needs to be an treated as argument, not a string literal as it currently is. The only way a literal can work is if the string is run through `eval`, which it isn't, and using `eval` for something easily avoided (as this PR shows) is not advisable

### License


jfagoagas (approved):
Good catch @chrisdlangton. Thank you very much for your contribution.

Issue #1133, opened 2022-05-12
[Bug]: Fails with "ERROR! - Invalid output format copying to S3."


cburgmer:
**What happened?**  
Prowler fails with `ERROR! - Invalid output format copying to S3.` after finishing all checks. This failure seems to start with 210f44f66f2b8ff5d6145866f5284b3ca6f5a19c, so seems to be related to https://github.com/prowler-cloud/prowler/pull/1129.

**How to reproduce it**  
1. `./prowler -r MY-REGION -f MY-REGION -s -E check42 -w ./ignorelist`
3. Error:

```
[...]
7.178 [extra7178] EMR Account Public Access Block enabled - emr [High]
       PASS! MY-REGION: EMR Account has Block Public Access enabled
 ERROR! - Invalid output format copying to S3. Use -h for help.
```

**Expected behavior**  
Print report and export normally

**From where are you running Prowler?**  
Please, complete the following information:
 - AWS-CLI Version [`aws --version`]: aws-cli/1.23.12 Python/3.9.5 Linux/4.14.268-205.500.amzn2.x86_64 botocore/1.25.12
 - Prowler Version [`./prowler -V`]: Prowler 2.9.0-13April2022

**Additional context**
It seems that the code now failing was never executed before. The option checked was removed in https://github.com/prowler-cloud/prowler/commit/210f44f66f2b8ff5d6145866f5284b3ca6f5a19c#diff-bfa61c2655ca74f20f41d49db363540c8aa5945cffe8cf4310386a9218aae559L39.

MrCloudSec:
Thank you @cburgmer for letting us now this bug! We have marked it down as critical and working on it.

MrCloudSec:
@cburgmer [PR #1134 ](https://github.com/prowler-cloud/prowler/pull/1134) was created to solve this issue, let us know if that works by your side to close this issue. Please, remember that the latest version is for beta, if you want a stable version use the release.

PR #1134, opened 2022-05-12
fix(copyToS3): Upload to S3 only when indicated.

MrCloudSec:
### Context 

Prowler fails with ERROR! - Invalid output format copying to S3 when a custom output bucket is not introduced. This PR is related to [Issue #1133](https://github.com/prowler-cloud/prowler/issues/1133).

### Description

To solve this error, copyToS3 function is executed only when an output bucket is introduced (with flags -B or -D)

### License

### DIFF
--- a/include/outputs_bucket
+++ b/include/outputs_bucket
@@ -13,23 +13,14 @@
 
 if [[ $OUTPUT_BUCKET ]]; then
   # output mode has to be set to other than text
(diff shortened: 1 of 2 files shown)


jfagoagas (approved):
Fastest fixer in town!!!! 🚀 🚀🚀🚀🚀

cburgmer:
Thanks, this works for us!

PR #1154, opened 2022-05-20
Timestamp to date casting issues solved

n4ch04:
### Context 

It seems that some times returned by AWS were in epoch format, and now are returned are timestamp-like (YYMMDD HHMMSS)
In prowler gnu date handling functions  date command needs an option to handle epoch inputs, which crashes when the input is not epoch.

### Description

Use the epoch option in prowler date functions only when the input is in epoch format.

### License


jfagoagas (approved):
Awesome!!!!! 

Please, modify PR name to use the convention.

PR #1155, opened 2022-05-23
fix(IllegalLocationConstraintException): Recover bucket policy using the right region endpoint

jfagoagas:
### Context 

`extra771` was failing due to the following error:

`An error occurred (IllegalLocationConstraintException) when calling the GetBucketPolicy operation: The af-south-1 location constraint is incompatible for the region specific endpoint this request was sent to.'`

### Description

The solution to this problem is to recover the bucket location using `s3api get-bucket-location` and then use to query the right AWS region endpoint.

Also, I've improved the error handling logic.

### License


MrCloudSec (approved):
Great job, it works!!

PR #1180, opened 2022-06-02
Consolidate prowler output functions

n4ch04:
### Context 

Currently Prowler handles the outputs through three functions (textPass/Info/Fail) that duplicate code

### Description

Refactor the code to create a general output function bearing in mind the general change of the checks output

### License


include/colors
```diff
@@ -12,58 +12,25 @@
...
+BLUE="[0;34m"
+BROWN="[0;33m"
+DARKGRAY="[0;30m"
+GRAY="[0;37m"
+GREEN="[1;32m"
+MAGENTA="[1;35m"
+PURPLE="[0;35m"
+RED="[1;31m"
+YELLOW="[1;33m"
+WHITE="[1;37m"
```
jfagoagas: Are all of these color variables in use somewhere in the code? 

If not we need to remove the unused ones. What do you think?

include/outputs
```diff
@@ -71,249 +38,226 @@ PROWLER_PARAMETERS=$@
...
 
-# Ensure that output directory always exists when -M is used
-if [[ $MODE ]];then
-  mkdir -p "${OUTPUT_DIR}"
-  if [[ "${MODES[@]}" =~ "html" ]]; then
-    addHtmlHeader > ${OUTPUT_FILE_NAME}.$EXTENSION_HTML
-    HTML_REPORT_INIT="1"
+# Set exitcode for codebuild/jenkins-like integrations
+set_exitcode () {
+  if [ "$FAILED_CHECK_FAILED_SCAN" == 1 ] && [ -z "$FAILED_CHECK_FAILED_SCAN_LIST" ]
```
jfagoagas:
```suggestion
  if [ "${FAILED_CHECK_FAILED_SCAN}" == 1 ] && [ -z "${FAILED_CHECK_FAILED_SCAN_LIST}" ]
```

include/outputs
```diff
@@ -71,249 +38,226 @@ PROWLER_PARAMETERS=$@
...
+  then
+      EXITCODE=3
   fi
-  if [[ "${MODES[@]}" =~ "csv" ]]; then
-    printCsvHeader
+  if [[ "${FAILED_CHECK_FAILED_SCAN_LIST[@]}" =~ "$CHECK_NAME" ]]
+  then
+      EXITCODE=3
   fi
```
jfagoagas: Please, take a look at this using shellcheck. This is a strange way to find something inside an array using Bash, that could lead to unexpected events.

include/outputs
```diff
@@ -71,249 +38,226 @@ PROWLER_PARAMETERS=$@
...
+  for MODE in "${MODES[@]}"; 
+  do
+    if  [ "${MODE}" == 'html' ] 
+    then
+        addHtmlHeader > "${OUTPUT_FILE_NAME}"."${EXTENSION_HTML}"
+        HTML_REPORT_INIT="1"
+    fi
+    if [ "${MODE}" == 'csv' ]
+    then
+        printCsvHeader
```
jfagoagas: I think here we need to be consistent with the things we do. So if `printCsvHeader` inserts the CSV header using a different function, `addHtmlHeader` should do the same.

include/outputs
```diff
@@ -71,249 +38,226 @@ PROWLER_PARAMETERS=$@
...
+  # Checked value is the whole log message that comes as argument
+  CHECKED_VALUE=${1}
+  # Default value of result is Fail
+  CHECK_RESULT='FAIL'
+  ## ignore allowlists for current check
+  while read -r excluded_item; do
+    # ignore_check_name is the check with resources allowlisted
+    IGNORE_CHECK_NAME=$(awk -F ":" '{print $1}' <<< "${excluded_item}")
+    # Resource value is what it comes after CHECK_NAME: :
+    RESOURCE_VALUE=$(awk -F "$CHECK_NAME:" '{print $2}' <<< "${excluded_item}")
```
jfagoagas: Here we can use something like this to avoid the use of external binaries like `awk`.

```suggestion
    # IGNORE_CHECK_NAME is the check with resources allowlisted
    # RESOURCE_VALUE is what it comes after 'CHECK_NAME:'
    IFS=':' read -r IGNORE_CHECK_NAME RESOURCE_VALUE <<< "${excluded_item}"
```

include/outputs
```diff
@@ -71,249 +38,226 @@ PROWLER_PARAMETERS=$@
...
+    if [[ "${IGNORE_CHECK_NAME}" != "${CHECK_NAME}" ]]; then
+      # not for this check
+      continue
+    fi
+    # To set WARNING flag checked_value have to include value of resource_value
+    # If it is treated as only expanse (*[]*) will not detect regex like [:alpha:]
+    if [[ "${CHECKED_VALUE}" =~ ${RESOURCE_VALUE} ]]; then
+      CHECK_RESULT="WARNING"
+      break
+    fi
```
jfagoagas:
```suggestion
    if [[ "${IGNORE_CHECK_NAME}" == "${CHECK_NAME}" ]]; then
      # To set WARNING flag checked_value have to include value of resource_value
      # If it is treated as only expanse (*[]*) will not detect regex like [:alpha:]
      if [[ "${CHECKED_VALUE}" =~ ${RESOURCE_VALUE} ]]; then
        CHECK_RESULT="WARNING"
        break
    fi
```

include/outputs
```diff
@@ -71,249 +38,226 @@ PROWLER_PARAMETERS=$@
...
-    REPREGION=$REGION
+      return 1
   fi
+}
 
+general_output() {
+  CHECK_RESULT="${1}"
+  CHECK_RESULT_EXTENDED="${2}"
+  CHECK_RESOURCE_ID="${3}"
+  REPREGION="${4}"
```
jfagoagas: What is `REPREGION` ? I always forget, a better name would be great.

include/outputs
```diff
@@ -71,249 +38,226 @@ PROWLER_PARAMETERS=$@
...
 
+EXTENSION_CSV="csv"
+EXTENSION_JSON="json"
+EXTENSION_ASFF="asff.json"
+EXTENSION_TEXT="txt"
+EXTENSION_HTML="html"
+HTML_LOGO_URL="https://github.com/prowler-cloud/prowler/"
+HTML_LOGO_IMG="https://github.com/prowler-cloud/prowler/raw/master/util/html/prowler-logo-new.png"
+TIMESTAMP=$(get_iso8601_timestamp)
+PROWLER_PARAMETERS=$@
```
jfagoagas, line 49: This 4 variables are unused. I think we need to delete it, what do you think?

include/outputs
```diff
@@ -71,249 +38,226 @@ PROWLER_PARAMETERS=$@
...
-  CHECK_RESULT_EXTENDED="$1"
-  CHECK_RESOURCE_ID="$3"
+# Add header to certain output files readed from mode
+output_files_init() {
+  for MODE in "${MODES[@]}"; 
+  do
+    if  [ "${MODE}" == 'html' ] 
+    then
+        addHtmlHeader > "${OUTPUT_FILE_NAME}"."${EXTENSION_HTML}"
+        HTML_REPORT_INIT="1"
```
jfagoagas, line 73: The same for this variables, is unused.

include/outputs
```diff
@@ -71,249 +38,226 @@ PROWLER_PARAMETERS=$@
...
+  CHECK_ASFF_COMP_TYPE="${6}" 
+
   if [[ $NUMERAL ]]; then
     # Left-pad the check ID with zeros to simplify sorting, e.g. 1.1 -> 1.01
     TITLE_ID=$(awk -F'.' '{ printf "%d.%02d", $1, $2 }' <<< "$TITLE_ID")
   fi
 
-  TITLE_TEXT=$2
   local CHECK_SERVICENAME="$MAGENTA$3$NORMAL"
   local CHECK_SEVERITY="$BROWN[$4]$NORMAL"
```
jfagoagas, line 251: Take a look at this, we are using several variables without using braces.

jfagoagas (changes requested):
I need to do some testing, this review is only for the code.

include/colors
toniblyx: I agree, as far as I remember, we use only NORMAL, WARNING, NOTICE, OK and BAD. But we should test it

include/outputs
toniblyx: which ones?

include/outputs
toniblyx: agreed

toniblyx (dismissed):
Please review the comments

include/outputs
jfagoagas: I said that this variables seems unused in this lib. Maybe we need to move it to the right file.

include/colors
n4ch04: Deleted unused colors, please check it out here https://github.com/prowler-cloud/prowler/pull/1180/commits/880d378b161eff7c47e12ed8155fbeebd31befb0

include/outputs
n4ch04: Solved, check it out here https://github.com/prowler-cloud/prowler/pull/1180/commits/4ee6a150d9ff7f311cd168c359e8cffb612b5521

include/outputs
n4ch04: Solved, check it out here https://github.com/prowler-cloud/prowler/pull/1180/commits/4ee6a150d9ff7f311cd168c359e8cffb612b5521

include/outputs
n4ch04: Solved, check it out here https://github.com/prowler-cloud/prowler/pull/1180/commits/4ee6a150d9ff7f311cd168c359e8cffb612b5521

include/outputs
n4ch04: Solved here https://github.com/prowler-cloud/prowler/pull/1180/commits/bed5466a2b0ff87e120b725d79dca9ef3e834cc2

n4ch04: Real solving here https://github.com/prowler-cloud/prowler/pull/1180/commits/d1a96640213db773920fdc0015a32bbfe3d2d19b IFS here not works because we need to split by `"$CHECK_NAME:"` in `RESOURCE_VALUE `

include/outputs
n4ch04: Real solving here https://github.com/prowler-cloud/prowler/commit/d1a96640213db773920fdc0015a32bbfe3d2d19b

include/outputs
n4ch04: changed, solved here https://github.com/prowler-cloud/prowler/pull/1180/commits/9593b7df55713ff3c8fbbbef53c8c6fb6bd74d75

include/outputs
n4ch04: All are used

include/outputs
n4ch04: Dark magic

jfagoagas (approved):
Thank god!!!! 

Superb @n4ch04 🚀 🚀🚀🚀

PR #1181, opened 2022-06-03
fix(assume_role): Use date instead of jq

### LINKED ISSUE #1165: The security token included in the request is expired
Is there a way to refresh the token while prowler is running so no checks are missed or a way to pause prowler before re-generating new token? I can refresh token at any point, but if prowler runs over the max session I receive the above error.

jfagoagas:
### Context 

Right now Prowler is not handling correctly the `AWS_SESSION_EXPIRATION` when you use an IAM Role to scan an AWS account if your timezone is different than `UTC`

Fixes https://github.com/prowler-cloud/prowler/issues/1165
### Description

Right now we transform the `AWS_SESSION_EXPIRATION` to `epoch` using the following `jq` code:
https://github.com/prowler-cloud/prowler/blob/432416d09ee40913732c0fc2ec8a21d7b85342ce/include/assume_role#L79

As you can see in the above `jq` issue, the dates are handled in the local timezone and not in `UTC`
https://github.com/stedolan/jq/issues/2001#issuecomment-1145864809

So, the issue here occurs when someone is using Prowler with a timezone different than `UTC`. In my case, using `CEST` the following occurs:
```
➜  prowler git:(master) ✗ date                                                            
Fri  3 Jun 2022 13:26:59 CEST
➜  prowler git:(master) ✗ echo '"2022-06-03T11:06:11+00:00"' |  jq '. | sub("\\+00:00";"Z") | strptime("%Y-%m-%dT%H:%M:%SZ")|mktime | todate' 
"2022-06-03T12:06:11Z"
```

Screenshot 2022-06-03 at 13 27 24
 
As you can see there is 1 hour delta between the right one and the bad one, just the difference of the summer time.

### License


toniblyx (approved):
LGTM

Issue #1184, opened 2022-06-07
[Bug]: Check119 incorrect remediation


bpinerou:
**What happened?**  
Check119 (CIS1.19) provide an remediation that is actually NOT true in AWS. It states "IAM roles can only be associated at the launch of an instance. To remediate an instance to add it to a role you must create or re-launch a new instance"
Actually there is no need to re-launch the instance and the role can be modified from the console.

**How to reproduce it**  
1. What command are you running?
Any
3. Environment you have, like single account, multi-account, organizations, etc.
Any environment with an EC2 instance without associated role that fails the CIS 1.19 control
5. See error

**Expected behavior**  

**Screenshots or Logs**  
Also, you can add logs (anonymize them first!). Here a command that may help to share a log
`bash -x ./prowler -options > debug.log 2>&1` then attach here `debug.log`

**From where are you running Prowler?**  
Please, complete the following information:
 - Resource: [e.g. EC2 instance, Fargate task, Docker container manually, EKS, Cloud9, CodeBuild, workstation, etc.)  
 - OS: [e.g. Amazon Linux 2, Mac, Alpine, Windows, etc. ]
 - AWS-CLI Version [`aws --version`]:
 - Prowler Version [`./prowler -V`]:
 - Shell and version:
 - Others:

**Additional context**

MrCloudSec:
Hi @bpinerou , thank you for reaching out!
You are right, we have to update the remediation of this check, we'll get back to you when we fix it.

bpinerou:
OK, dale una colleja a Tony de parte de Blas

Un saludo

MrCloudSec:
He hecho la [PR](https://github.com/prowler-cloud/prowler/pull/1185) que resuelve este problema, échala un vistazo y dime si te parece bien. Gracias y un saludo!

PR #1185, opened 2022-06-07
fix(check119_remediation): Update check remediation text.

MrCloudSec:
### Context 

Regarding to [Issue 1884](https://github.com/prowler-cloud/prowler/issues/1184), Check119 (CIS1.19) provide a remediation that is actually NOT true in AWS since, right now, there is no need to re-launch the instance and the role can be modified from the console.

### Description

Update check remediation text for an actual remediation.

### License


jfagoagas (approved):
Good change!! Thanks @sergargar

PR #1195, opened 2022-06-14
fix(session_duration): Use jq with TZ=UTC

### LINKED ISSUE #1186: [Bug]: prowler/prowler: line 525: ((: 600 > : syntax error: operand expected (error token is "> ")
I'm getting the following error on Alpine Linux v3.16 - prowler/prowler: line 525: ((: 600 > : syntax error: operand expected (error token is "> ")

I'm not sure when this started, I just happened to be looking at the logs from my prowler run for an unrelated reason.

jfagoagas:
### Context 

We have been experiencing some issues with `jq` and different timezones due to how `jq` handles UTC and local time zones.

### Description

This PR aims to fix this behaviour in every system using `jq` with the timezone environment variable set to UTC `TZ=UTC`

Fix https://github.com/prowler-cloud/prowler/issues/1186

### License

### DIFF
--- a/include/assume_role
+++ b/include/assume_role
@@ -18,7 +18,7 @@ assume_role(){
         # If profile is not defined, restore original credentials from environment variables, if they exists!
         restoreInitialAWSCredentials
     fi
-    
+
     # Both variables are mandatory to be set together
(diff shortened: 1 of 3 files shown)


MrCloudSec (approved):
Good job!!!

PR #1198, opened 2022-06-14
Remove false positive for check_extra767

zsecducna:
Exclude distributions that does not support `POST` requests
From AWS documentation: 
"Field-level encryption configurations help you protect specific data that viewers upload in `POST` requests."

And we cannot enable field-level encryption if the distribution does not support `POST` method


jfagoagas:
Thank you for your contribution @zsecducna, good catch!!

jfagoagas (approved):
Hi @zsecducna, I made some changes to adapt this check to the Prowler's structure. 

Thanks!

PR #1214, opened 2022-06-21
fix(instance metadata): missing raw flag in jq parser

n4ch04:
### Context 

Since last reformat prowler did not work when launched in a EC2 instance, it is needed to inspect why

### Description

When using jq to extract credential fields from the instance metadata answer there were some missing raw flags, it is needed to add it

### License


jfagoagas (approved):
Really good catch!! 🔝

PR #1215, opened 2022-06-21
feat(validate-metadata): Validate Check's metadata and list checks

jfagoagas:
### Context 

We need to validate Check's metadata in order to comply with the defined structure.

### Description

- Use `pydantic` to validate Check's metadata model.
- Parse all metadata files at once to recover its properties.
- List checks using this metadata information.

Also, created some properties to recover every value from the Check class, without the ability of modify any of it.

### License


n4ch04 (approved):
Nice !!!

PR #1219, opened 2022-06-22
fix(shub_fails): Treat failed findings as failed in SHub.

### LINKED ISSUE #1218: [Bug]: Failed findings in Security Hub treated like Passed
Hello guys, I just want to ask if the new version of prowler works correctly. Today I implemented Prowler into AWS account, but all findings that are in SecurityHub have Compliance status “PASSED” instead of “FAILED”.  I am using  “-q” flag for only failed findings. Previous version worked correctly with the same parameters which I used today. 
Prowler

MrCloudSec:
### Context 

Failed findings in Security Hub treated like Passed always, according to https://github.com/prowler-cloud/prowler/issues/1218.

### Description

Treat findings as their original status in Security Hub.

### License


jfagoagas (approved):
Good catch 👯

PR #1220, opened 2022-06-22
fix(quit): Replace with `sys.exit()`

jfagoagas:
### Context 

We must not use `quit()` in a production environment. 

Rationale: https://stackoverflow.com/questions/19747371/python-exit-commands-why-so-many-and-when-should-each-be-used

### Description

- Replace `quit()` with `sys.exit()`

### License


MrCloudSec (approved):
LGTM

PR #1228, opened 2022-06-23
feat(default_regions): Set profile region as default for global regions.

MrCloudSec:
### Context 

For global services such as IAM, set the region to the default one within the user's profile, instead of hard coding us-east-1.

### Description

- Set region of global services to profile region.
- Add region names to logging.

### License


jfagoagas (approved):
awesome!! good job @sergargar

PR #1230, opened 2022-06-24
feat(outputS): Output generation format CSV

n4ch04:
### Context 

Prowler needs to be able to export results to csv

### Description

A new folder into lib has been created storing the output logic, the Check_Output type stores all the check metadata coming from the `.json` model , general info like account or profile and organizations info, and has methods to send all to a csv file

### License


lib/check/check.py
```diff
@@ -145,22 +149,30 @@ def import_check(check_path: str) -> ModuleType:
...
     logger.debug(f"Executing check: {check.checkName}")
     findings = check.execute()
-    report(findings, output_options)
+    organizations_info = Organizations_Info(
+        account_details_email="test@verica.io",
+        account_details_name="washemen",
+        account_details_arn="arn:aws:iam::account:test/test",
+        account_details_org="ou-washemen",
+        account_details_tags="tag1:likisomawei",
+    )
```
jfagoagas: Please, remove these hardcoded values.

Pipfile
```diff
@@ -9,6 +9,7 @@ boto3 = "1.24.8"
 arnparse = "0.0.2"
 botocore = "1.27.8"
 pydantic = "1.9.1"
+pytest = "*"
```
jfagoagas: Please, include the specific version here.

config/config.py
```diff
@@ -11,3 +11,43 @@
...
+csv_header = [
+    "PROVIDER",
+    "PROFILE",
+    "ACCOUNT_ID",
+    "REGION",
+    "CHECKID",
+    "CHECKNAME",
+    "CHECKTITLE",
+    "STATUS",
+    "RESULT_EXTENDED",
```
jfagoagas: Why some fields has an `_` and other fields don't?

We need to set the same pattern for every header, what do you think?

prowler
```diff
@@ -91,6 +91,14 @@ if __name__ == "__main__":
         nargs="+",
         help="AWS region names to run Prowler against",
     )
+    parser.add_argument(
+        "-M",
+        "--output-modes",
+        nargs="+",
+        help="Output mode, by default csv",
+        choices=["csv"],  # , "json", "html", "json-asff", "junit-xml"]
+        default=["csv"],
```
jfagoagas: Is this only for testing purposes?

lib/check/check.py
```diff
@@ -145,22 +149,30 @@ def import_check(check_path: str) -> ModuleType:
     return lib
 
 
-def set_output_options(quiet):
+def set_output_options(quiet: bool, output_modes: list):
```
jfagoagas: If this function does not have test, please create a new story to address it in the future.

lib/check/check.py
```diff
@@ -145,22 +149,30 @@ def import_check(check_path: str) -> ModuleType:
...
     findings = check.execute()
-    report(findings, output_options)
+    organizations_info = Organizations_Info(
+        account_details_email="test@verica.io",
+        account_details_name="washemen",
+        account_details_arn="arn:aws:iam::account:test/test",
+        account_details_org="ou-washemen",
+        account_details_tags="tag1:likisomawei",
+    )
+    report(findings, output_options, audit_info, organizations_info)
```
jfagoagas: I think it would be better to split the `run_check` function in two separate functions, one to run the check and another to report the execution findings.

We need to be able to test this logic separately.

What do you think?

lib/check/check.py
n4ch04: Done

config/config.py
```diff
@@ -11,3 +11,43 @@
...
 )
 aws_services_json_file = "providers/aws/aws_regions_services.json"
+
+
+csv_header = [
+    "PROVIDER",
+    "PROFILE",
+    "ACCOUNT_ID",
+    "REGION",
+    "CHECKID",
```
jfagoagas: Is there any way to generate this headers dynamically using the `Check_Output` dataclass?

lib/outputs/models.py
```diff
@@ -0,0 +1,170 @@
...
+@dataclass
+class Compliance_Framework:
+    Framework: str
+    Version: str
+    Group: list
+    Control: list
+
+
+@dataclass
+class Check_Output:
```
jfagoagas: I think this `Check_Output` dataclass is only for the `csv` mode, right?

lib/outputs/models.py
```diff
@@ -0,0 +1,170 @@
...
+    STATUS: str
+    REGION: str
+    RESULT_EXTENDED: str
+    PROFILE: str
+    ACCOUNT_ID: int
+    PROVIDER: str
+    CHECKID: str
+    CHECKNAME: str
+    CHECKTITLE: str
+    CHECKTYPE: str
```
jfagoagas: Why all of these fields are declared in uppercase?

lib/outputs/models.py
```diff
@@ -0,0 +1,170 @@
...
+        unrolled_items = ""
+        separator = "|"
+        for item in listed_items:
+            if not unrolled_items:
+                unrolled_items = f"{item}"
+            else:
+                unrolled_items = f"{unrolled_items}{separator}{item}"
+
+        return unrolled_items
+
```
jfagoagas, line 135: What is the difference to unroll a list and unroll compliance list? I think a generic version to unroll values would be great!

lib/outputs/outputs.py
```diff
@@ -0,0 +1,78 @@
...
+                    "a",
+                )
+                # csv_writer = writer(file_descriptor, delimiter=';')
+                # csv_writer.writerow(csv_header)
+                csv_writer = DictWriter(
+                    file_descriptor, fieldnames=csv_header, delimiter=";"
+                )
+                csv_writer.writeheader()
+
+            file_descriptors.update({output_mode: file_descriptor})
```
jfagoagas: Why don't you include this code in a function outside report? Because it is only intended to extract a new `fd` for each file. What do you think?

lib/outputs/outputs.py
```diff
@@ -0,0 +1,78 @@
...
+                f"prowler-output-{audit_info.audited_account}-{format_timestamp}.csv"
+            )
+            if file_exists(filename):
+                file_descriptor = open_file(
+                    f"prowler-output-{audit_info.audited_account}-{format_timestamp}.csv",
+                    "a",
+                )
+            else:
+                file_descriptor = open_file(
+                    f"prowler-output-{audit_info.audited_account}-{format_timestamp}.csv",
```
jfagoagas: Include the file suffix name and extension in variables, we are going to need it in other report formats.

lib/utils/utils.py
```diff
@@ -18,10 +18,22 @@ def open_file(input_file: str) -> TextIOWrapper:
 
 # Parse checks from file
 def parse_json_file(input_file: TextIOWrapper) -> Any:
+    # First recover the available groups in groups.json
```
jfagoagas: Please remove this old comment

jfagoagas (changes requested):
Good job @n4ch04 !!

As an overall comment, we need to split functions trying to avoid multiple responsibilities for each one.

lib/outputs/models.py
n4ch04: Right now yes

lib/outputs/models.py
n4ch04: At the beginning it was due to I thought it was necessary, but now it is only to help match with CSV Headers

lib/outputs/models.py
n4ch04: Since for compliance is neeeded to make a type cast for `Compliance_Framework` the code would look like this:
```
if item is compliance:
#All logic of __unroll_compliance__
else:
# All logic of __unroll_list__
```
I think is better splitted

config/config.py
n4ch04: Matched with output doc

config/config.py
jfagoagas: You can use the following https://docs.python.org/3/library/dataclasses.html#dataclasses.asdict

prowler
n4ch04: Deleted

n4ch04:
> Is this only for testing purposes? 
Deleted

n4ch04:
> You can use the following https://docs.python.org/3/library/dataclasses.html#dataclasses.asdict 

Done, now the headers are created dynamically

n4ch04:
> Why don't you include this code in a function outside report? Because it is only intended to extract a new `fd` for each file. What do you think? 

Done

lib/utils/utils.py
n4ch04: Done

Pipfile
n4ch04: DOne

config/config.py
n4ch04: Already done

lib/outputs/models.py
n4ch04: Declared in lower case right now

lib/outputs/outputs.py
n4ch04: It is coded in a different function right now

lib/outputs/outputs.py
n4ch04: Included

lib/check/check.py
n4ch04: It is divided now, a call for filling orgs info is made, right now that function only instantiates an empty object to be passed to `report` function call

lib/outputs/models.py
n4ch04: Changing its name to clarify it is only for `csv`

MrCloudSec (approved):
Great job! 🔝

config/config.py
```diff
@@ -1,10 +1,19 @@
...
 prowler_version = "3.0-alfa"
 
 # Groups
 groups_file = "groups.json"
 
 # AWS services-regions matrix json
-aws_services_json_file = "providers/aws/aws_regions_by_service.json"
+aws_services_json_url = (
+    "https://api.regional-table.region-services.aws.a2z.com/index.json"
+)
```
jfagoagas:
```suggestion
```

lib/outputs/models.py
jfagoagas: Perfect!

lib/outputs/outputs.py
```diff
@@ -0,0 +1,108 @@
...
+            print(
+                f"{color}{finding.status}{Style.RESET_ALL} {finding.region}: {finding.status_extended}"
+            )
+        # sending the finding to input options
+        if "csv" in file_descriptors:
+
+            csv_writer = DictWriter(
+                file_descriptors["csv"], fieldnames=csv_fields, delimiter=";"
+            )
+            # csv_line = [audit_info.profile,audit_info.audited_account,finding.region,finding.check_metadata.CheckID,finding.status,finding.check_metadata.CheckTitle,finding.result_extended]
```
jfagoagas:
```suggestion
```

jfagoagas (approved):
Tests :D

PR #1233, opened 2022-06-24
feat(check): Ensure default internet access from Amazon AppStream fleet should be disabled.

rajarshidas:
### Context 

Default Internet Access from your fleet streaming instances should be controlled using a NAT gateway in the VPC.

### Description

Check to ensure default internet access from Amazon AppStream fleet streaming instances should remain unchecked.

### License


MrCloudSec (approved):
Thank you @rajarshidas for your continuing and amazing contribution.

PR #1240, opened 2022-06-27
feat(check7164): 365 days or more in a Cloudwatch log retention should be consider PASS

bcarranza:
### Context 

Prowler fails when log retention period is more 365 days, b/c the validation is `==` instead of `>= `

### Description

Many customer are using modules like gruntworks in vpc So this is using a default value for log retention which is 731 days
Screen Shot 2022-06-27 at 14 50 39

I think that we can considered that a cloud watch log rentention Greater than 365 days is passing. 
What do you think comunity?

Also, with this fix Log Groups that never expires are included as PASS.

### License

### DIFF
--- a/checks/check_extra7162
+++ b/checks/check_extra7162
@@ -48,6 +48,7 @@ extra7162() {
                 textPass "${regx}: ${log} Log Group retention period never expires!" "${regx}" "${log}"
             done
         fi
+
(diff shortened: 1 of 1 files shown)


jfagoagas (approved):
Thank you for your contribution @bcarranza, it is a really good fix for this check.

I've also included minor changes to texts and another test to review never expire log groups.

PR #1244, opened 2022-06-30
fix(postgres): Fix postgres connector issues.

MrCloudSec:
### Context 

Fix connection when the database is not named postgres.

### Description

With this fix, even if the user is postgres you can insert a different database name.

### License


jfagoagas (approved):
Thanks! 🚀

PR #1248, opened 2022-07-04
feat(organizations):  Extract Metadata from Management Account ID (-O)

MrCloudSec:
### Context 

 Extract Metadata from Management Account ID (-O) by inputting a role arn.

### Description

Adding -O function to extract organizations metadata with a role ARN.

### License

### DIFF
--- a/lib/outputs/models.py
+++ b/lib/outputs/models.py
@@ -1,7 +1,8 @@
 from dataclasses import asdict, dataclass
 
 from config.config import timestamp
-from lib.check.models import Check_Report, Organizations_Info
+from lib.check.models import Check_Report
+from providers.aws.models import Organizations_Info
 
 
 @dataclass
@@ -72,11 +73,12 @@ class Check_Output_CSV:
         self.provider = report.check_metadata.Provider
         self.profile = profile
         self.account_id = account
-        self.account_name = organizations.account_details_name
-        self.account_email = organizations.account_details_email
-        self.account_arn = organizations.account_details_arn
-        self.account_org = organizations.account_details_org
-        self.account_tags = organizations.account_details_tags
+        if organizations:
+            self.account_name = organizations.account_details_name
+            self.account_email = organizations.account_details_email
+            self.account_arn = organizations.account_details_arn
+            self.account_org = organizations.account_details_org
+            self.account_tags = organizations.account_details_tags
         self.region = report.region
         self.check_id = report.check_metadata.CheckID
         self.check_name = report.check_metadata.CheckName
@@ -181,9 +183,3 @@ class Check_Output_CSV:
             groups = ""
 
         return unrolled_compliance
-
-    def get_csv_header(self):
-        csv_header = []
-        for key in asdict(self):
-            csv_header = csv_header.append(key)
-        return csv_header
--- a/lib/outputs/models.py
+++ b/lib/outputs/models.py
@@ -2,7 +2,7 @@ from dataclasses import asdict, dataclass
 
 from config.config import timestamp
 from lib.check.models import Check_Report
-from providers.aws.models import Organizations_Info
+from providers.aws.models import AWS_Organizations_Info
 
 
 @dataclass
@@ -66,19 +66,18 @@ class Check_Output_CSV:
         account: str,
         profile: str,
         report: Check_Report,
-        organizations: Organizations_Info,
+        organizations: AWS_Organizations_Info,
     ):
         self.assessment_start_time = timestamp.isoformat()
         self.finding_unique_id = ""
         self.provider = report.check_metadata.Provider
         self.profile = profile
         self.account_id = account
-        if organizations:
-            self.account_name = organizations.account_details_name
-            self.account_email = organizations.account_details_email
-            self.account_arn = organizations.account_details_arn
-            self.account_org = organizations.account_details_org
-            self.account_tags = organizations.account_details_tags
+        self.account_name = organizations.account_details_name
+        self.account_email = organizations.account_details_email
+        self.account_arn = organizations.account_details_arn
+        self.account_org = organizations.account_details_org
+        self.account_tags = organizations.account_details_tags
         self.region = report.region
         self.check_id = report.check_metadata.CheckID
         self.check_name = report.check_metadata.CheckName
--- a/lib/outputs/outputs.py
+++ b/lib/outputs/outputs.py
@@ -15,7 +15,7 @@ def report(check_findings, output_options, audit_info):
     file_descriptors = {}
     if output_options.output_modes:
         if "csv" in output_options.output_modes:
-            csv_fields = generate_csv_fields(audit_info.organizations_metadata)
+            csv_fields = generate_csv_fields()
 
         file_descriptors = fill_file_descriptors(
             output_options.output_modes,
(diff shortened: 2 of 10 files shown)


prowler
```diff
@@ -133,6 +133,12 @@ if __name__ == "__main__":
         help="Custom output directory, by default the folder where Prowler is stored",
         default=default_output_directory,
     )
+    parser.add_argument(
+        "-O",
+        "--organizations-role",
+        nargs="?",
+        help="Specify AWS Organizations management role ARN to be assumed.",
```
jfagoagas: Specify why the role assumption is needed to.

providers/aws/aws_provider.py
```diff
@@ -153,15 +193,14 @@ def provider_set_session(
         logger.info("Audit session is the original one")
         current_audit_info.audit_session = current_audit_info.original_session
 
-
+    logger.info("Checking if organizations metadata is needed ...")
```
jfagoagas: Are we going to use the three dots at the end of the logs?

providers/aws/aws_provider.py
```diff
@@ -200,3 +239,37 @@ def assume_role(audit_info: AWS_Audit_Info) -> dict:
 
     else:
         return assumed_credentials
+
+
+def get_organizations_metadata(assumed_credentials):
```
jfagoagas: Please, include type hints for this function.

providers/aws/aws_provider.py
```diff
@@ -200,3 +239,37 @@ def assume_role(audit_info: AWS_Audit_Info) -> dict:
...
+        )
+        list_tags_for_resource = organizations_client.list_tags_for_resource(
+            ResourceId=current_audit_info.audited_account
+        )
+        # Convert Tags dictionary to String
+        account_details_tags = ""
+        for tag in list_tags_for_resource["Tags"]:
+            account_details_tags = (
+                account_details_tags + tag["Key"] + ":" + tag["Value"] + ","
+            )
```
jfagoagas: You can do the following to concatenate strings:
```suggestion
            account_details_tags +=  tag["Key"] + ":" + tag["Value"] + ","
```

providers/aws/aws_provider.py
```diff
@@ -200,3 +239,37 @@ def assume_role(audit_info: AWS_Audit_Info) -> dict:
...
+            aws_secret_access_key=assumed_credentials["Credentials"]["SecretAccessKey"],
+            aws_session_token=assumed_credentials["Credentials"]["SessionToken"],
+        )
+        organizations_metadata = organizations_client.describe_account(
+            AccountId=current_audit_info.audited_account
+        )
+        list_tags_for_resource = organizations_client.list_tags_for_resource(
+            ResourceId=current_audit_info.audited_account
+        )
+        # Convert Tags dictionary to String
```
jfagoagas, line 262: I think every line from this point should be under the `else` clause.

providers/aws/models.py
```diff
@@ -19,6 +19,15 @@ class AWS_Assume_Role:
     external_id: str
 
 
+@dataclass
+class Organizations_Info:
```
jfagoagas: I prefer to name it as the other classes with the `AWS` prefix. What do you think?

lib/outputs/models.py
```diff
@@ -181,9 +183,3 @@ def __unroll_compliance__(self, compliance: list):
             groups = ""
 
         return unrolled_compliance
-
-    def get_csv_header(self):
-        csv_header = []
-        for key in asdict(self):
-            csv_header = csv_header.append(key)
-        return csv_header
```
jfagoagas, line 189: Why do you remove this? Is interesting to have this logic inside that class.

lib/outputs/outputs.py
```diff
@@ -3,20 +3,19 @@
 from colorama import Fore, Style
 
 from config.config import csv_file_suffix
-from lib.check.models import Organizations_Info
 from lib.outputs.models import Check_Output_CSV
 from lib.utils.utils import file_exists, open_file
 
 
-def report(check_findings, output_options, audit_info, organizations_info):
```
jfagoagas, line 11: Why don't you include `organizations_info` within `audit_info`? Organizations are a feature of AWS and we don't want to couple our common logic with a cloud vendor. Maybe we can include it there, what's your opinion?

jfagoagas (changes requested):
Good job @sergargar, the only thing I'm missing is testing for this code. Please check this library to code some test for the organizations metadata function. 

https://github.com/spulec/moto/blob/master/tests/test_organizations/test_organizations_boto3.py

lib/outputs/models.py
```diff
@@ -72,11 +73,12 @@ def __init__(
         self.provider = report.check_metadata.Provider
         self.profile = profile
         self.account_id = account
-        self.account_name = organizations.account_details_name
-        self.account_email = organizations.account_details_email
-        self.account_arn = organizations.account_details_arn
-        self.account_org = organizations.account_details_org
-        self.account_tags = organizations.account_details_tags
+        if organizations:
```
n4ch04: Why include here the conditional? You are enforcing in the constructor to have a `Organizations_Info` object as argument, empty or not, so the attributes related must be filled in any case, what do you think?

lib/outputs/models.py
n4ch04: I think is no longer used, but yes, please, check it out

lib/outputs/outputs.py
```diff
@@ -3,20 +3,19 @@
...
+def report(check_findings, output_options, audit_info):
     check_findings.sort(key=lambda x: x.region)
 
     csv_fields = []
     # check output options
     file_descriptors = {}
     if output_options.output_modes:
         if "csv" in output_options.output_modes:
-            csv_fields = generate_csv_fields()
+            csv_fields = generate_csv_fields(audit_info.organizations_metadata)
```
n4ch04: Please, I think it is not a good practice to have variable fields in the csv file. 
If we do not fill the fields that fields must be empty

lib/outputs/outputs.py
```diff
@@ -101,19 +99,15 @@ def set_report_color(status):
...
-def generate_csv_fields():
+def generate_csv_fields(organizations_metadata):
     csv_fields = []
+    organizations_fields = [
+        "account_email",
+        "account_arn",
+        "account_org",
+        "account_tags",
+    ]
     for field in Check_Output_CSV.__dict__["__annotations__"].keys():
```
n4ch04, line 104: What I said before, It is preferable to have static csv fields

providers/aws/aws_provider.py
```diff
@@ -115,11 +126,40 @@ def provider_set_session(
...
-    if current_audit_info.assumed_role_info.role_arn:
+    if input_role:
+        current_audit_info.assumed_role_info.role_arn = input_role
+        current_audit_info.assumed_role_info.session_duration = input_session_duration
+        current_audit_info.assumed_role_info.external_id = input_external_id
+
         # Check if role arn is valid
         try:
             # this returns the arn already parsed, calls arnparse, into a dict to be used when it is needed to access its fields
+            print(current_audit_info.assumed_role_info)
```
n4ch04: Do we need this `print` here? Maybe little debugging code forgotten?
Check it out

providers/aws/aws_provider.py
n4ch04: This logger line is not the same as line 129? I think it should not be here 
What do you think?

providers/aws/aws_provider.py
n4ch04: Yes, into the `try` statement only the calls to the apis, all the data processing in else clause

n4ch04 (commented):
Nice job! please check it out the few suggestions

lib/outputs/models.py
MrCloudSec: I removed it because the function was duplicated, it is also in line 58.

lib/outputs/outputs.py
MrCloudSec: organizations_info is already within audit_info

n4ch04: Agree

jfagoagas (approved):
🔝

n4ch04 (approved):

PR #1258, opened 2022-07-08
feat(inventory): Prowler quick inventory including IAM resources

toniblyx:
### Context 

### Description

### License


jfagoagas (changes requested):
Please, fulfil the PR body prior to ask for review. Thanks!

n4ch04 (commented):
Are we including inventory from services like IAM? 
What if we include that as part of the results to extract added value from that information?

n4ch04 (commented):
Do we really need to use temp files? 
I'am going to give it a try changing it to store info in memory

jfagoagas (changes requested):
Please, include error handling for each AWS CLI call.

jfagoagas (approved):
Let's go!

PR #1261, opened 2022-07-11
fix(apigatewayv2): handle BadRequestException

MrCloudSec:
### Context 

In checks extra7156 and extra7157 there is a BadRequestException for the regions that does not have apigatewayv2.

### Description

Handle BadRequestException correctly to solve the issue.

### License


jfagoagas (approved):
Good catch!

Issue #1259, opened 2022-07-09
[Bug]: CodeBuild jobs running > 60 minutes use expired AWS credentials


jameswarren:
**What happened?**  
After deploying Prowler to run from AWS CodeBuild, jobs running for > 60 minutes use expired credentials and cannot further execute checks.

**How to reproduce it**  
1. What command are you running? `prowler -r us-west-2 -g iso27001 -M html -z`
2. Environment you have, like single account, multi-account, organizations, etc. `single account`
3. See error: `screenshot attached below`

**Expected behavior**  
Prowler runs using CodeBuild taking > 60 minutes should successfully execute all checks using valid AWS credentials.

**Screenshots or Logs**  
At the 60-minute mark (maximum session time used by CodeBuild), checks start using expired credentials
![Screen Shot 2022-07-08 at 5 05 43 PM](https://user-images.githubusercontent.com/260910/178084494-291402af-6022-454d-af46-ebfff2e1b5ed.png)

**From where are you running Prowler?**  
 - Resource: CodeBuild
 - OS: Amazon Linux 2
 - AWS-CLI Version [`aws --version`]: aws-cli/2.7.14
 - Prowler Version [`./prowler -V`]: v2.10.0-25May2022
 - Shell and version: bash

**Additional context**
I've identified the issue that [credentials are extracted from the container metadata at the beginning of the run and exported as environment variables](https://github.com/prowler-cloud/prowler/blob/ff33f426e5314ea24e248dcd4826d6b17679eac8/include/aws_profile_loader#L35-L40). While CodeBuild will refresh the automatically refresh the container credentials, they are not made accessible to prowler.

For our immediate needs, commenting out the `EXPORT` statements ensures prowler uses credentials from the container and executions > 60 minutes can run successfully. It is unclear to me how the specific code is used in other contexts (e.g. for organization-wide checks), but given additional guidance, I'm happy to work towards submitting a patch.

MrCloudSec:
Hi @jameswarren, thank you for reaching out!
We'll take a look at it and get back to you.

sergargar:
Please, try it again with the latest version. We have fix the error with the above PR.

jfagoagas:
Hi @jameswarren, please test it again using branch `expired-token-codebuild` from this PR https://github.com/prowler-cloud/prowler/pull/1265. 

Let us know if everything is fixed.

jameswarren:
[...]

Issue #1182, opened 2022-06-06
[Bug]: Codebuild COMMAND_EXECUTION_ERROR


SecOp187:
**What happened?**  
I followed exactly the instructions to run Prowler in a CodeBuild environment https://github.com/prowler-cloud/prowler#codebuild-deployment by deploying the CloudFormation template. After running for 33 minutes, the build fails with the error message: "Command did not exit successfully ./prowler $PROWLER_OPTIONS exit status 3". These are the environment variables:
![image](https://user-images.githubusercontent.com/82619023/172215638-b4c672ea-784d-4634-9067-5c913366dc4b.png)

**How to reproduce it**  
1. What command are you running? 
![image](https://user-images.githubusercontent.com/82619023/172215916-1c25dea6-2a58-4f90-8783-52a6958d879c.png)
2. Environment you have, like single account, multi-account, organizations, etc. Single AWS account
3. See error
![image](https://user-images.githubusercontent.com/82619023/172216167-29a7b3b6-2f7f-44d7-a28a-0bd1017c156a.png)

**Screenshots or Logs**  
See above

**From where are you running Prowler?**  
 - Resource: [e.g. EC2 instance, Fargate task, Docker container manually, EKS, Cloud9, CodeBuild, workstation, etc.)  CodeBuild
 - OS: [e.g. Amazon Linux 2, Mac, Alpine, Windows, etc. ] aws/codebuild/amazonlinux2-x86_64-standard:3.0
 - AWS-CLI Version [`aws --version`]: v2
 - Prowler Version [`./prowler -V`]: 2.5.0

MrCloudSec:
Hi @SecOp187, thank you for reaching out!
We'll take a look at it and get back to you.

MrCloudSec:
How did you include the PROWLER_BRANCH variable?

MrCloudSec:
We have solve some issues in this [PR](https://github.com/prowler-cloud/prowler/pull/1183) , please try it again and let us know if that fix your errors.

SecOp187:
> How did you include the PROWLER_BRANCH variable?

I modified the PROWLER_BRANCH attribute (default value) by changing 2.5 to 2.5.0 in the yaml file (https://static.us-east-1.prod.workshops.aws/public/ba7ded10-50ba-4afc-9932-4e168003e650/static/templates/3-Deploy-Run-Prowler-CodeBuild.yaml). Because I was getting error due to the non-existent branch 2.5.
[...]

PR #1262, opened 2022-07-11
fix(codebuild): expired token error

MrCloudSec:
### Context 

According to issues https://github.com/prowler-cloud/prowler/issues/1259 and https://github.com/prowler-cloud/prowler/issues/1182, Codebuild jobs that run more than 60 minutes use expired AWS credentials.

### Description

Refresh credentials when the Codebuild session is expired.

### License


jfagoagas (approved):
🚀 🚀 🚀 🚀

PR #1266, opened 2022-07-14
feat(s3_checks): add several checks for s3 

MrCloudSec:
### Context 

There were added 2 checks of S3 as well as the client.

### Description

Add checks extra718 and extra 763.

### License


jfagoagas (approved):
Good job @sergargar !!! 🪣🪣🪣🪣

providers/aws/services/s3/s3_service.py
```diff
@@ -0,0 +1,100 @@
...
+            threads.append(threading.Thread(target=call, args=(regional_client,)))
+        for t in threads:
+            t.start()
+        for t in threads:
+            t.join()
+
+    def __list_buckets__(self, regional_client):
+        logger.info("S3 - Listing buckets...")
+        try:
+            list_buckets = regional_client.list_buckets()
```
jfagoagas, line 34: Why if you query AWS only once to retrieve the bucket list?

jfagoagas (changes requested):
Check my comment please

providers/aws/services/s3/s3_service.py
MrCloudSec: At the moment, the only solution to speed up this algorithm is to have regional resources, which is going to be done in the future.

jfagoagas (approved):
🚀  awesome!

PR #1281, opened 2022-07-26
fix(check12): Improve remediation

jfagoagas:
### Description

Improve remediation for `check12` because  this check is not related to root users.

### License


MrCloudSec (approved):
Good catch!

PR #1285, opened 2022-07-27
feat(regions): add regions to resources

MrCloudSec:
### Context 

We realized we have to add the region to each resource in order to speed up the information retrieval from AWS and to reduce API calls.

### Description

Regional clients are now only used in the provider so in the check you only have to iterate through the resources (independently from the regions)
Now, it does not show if there are no resources per region.

### License


jfagoagas (approved):
🚀 🚀 🚀 🚀

PR #1286, opened 2022-07-27
fix(extra712): changed Macie service detection

williambrady:
### Context 

Reference issue 1279 in regards to a better way of determining Macie status.

### Description

Instead of looking for an IAM role and guessing that Macie may be enabled, directly querying get-macie-session status will determine the current state.

### License


MrCloudSec:
Hi @williambrady, good catch!
I have added some changes to this PR such as adding regional context and handling of errors.
Thank you for the contribution 😄

MrCloudSec (approved):

PR #1291, opened 2022-07-28
feat(extra780): Check for Cognito or SAML authentication on OpenSearch

kagahd:
### Context 

check_extra780 fails if Amazon OpenSearch has Amazon Cognito authentication for Kibana **not** enabled. However, besides Cognito, AWS recommends to use **SAML** authentication for Kibana which means that Cognito must be disabled. Since the check_extra780 does not test whether **either** Cognito **or** SAML is enabled, it will fail when SAML is enabled.

### Description

I extended check_extra780 in a way that it tests whether **either** Cognito **or** SAML is enabled for Kibana on AWS OpenSearch. If either is enabled, check_extra780 will pass.

### License


jfagoagas (approved):
Thank you for your contribution @kagahd, this is a really good enhancement for the check `extra780` 👏 👏 👏 

I have added some changes to this PR such as adding regional context and handling of errors.

Issue #1293, opened 2022-07-29
extra7190, 7191, 7192, 7193 appstream regions


williambrady:
**Is your feature request related to a problem? Please describe.**

Currently appstream is not available in all regions, so when the scans happen for `sa-east-1` for example, they generate errors which result in a Fail state. It's also not available in `eu-north-1` but AWS returns a different error, making the check overall unreliable.

**Describe the solution you'd like**

Unless you already have a service-specific region list, I would like to generate one by adding a check-specific REGIONS value:

`curl -s https://api.regional-table.region-services.aws.a2z.com/index.json | jq -r '.prices[].attributes|select(."aws:serviceName" == "Amazon AppStream 2.0") | ."aws:region"'`

This will return a current list of regions for the specific service. This could also be done at a higher level of code logic (like globally generate lists for all the services and have the checks call back) but embedding this in the affected modules doesn't seem to add excessive overhead.

**Describe alternatives you've considered**

Another option would be to populate a map of services to regions that  is populated at run-time, and allow each module to call their respective REGION list. This could be better long term, but would take more planning since I don't know all the inner workings of the scripts yet. Additionally, we'd likely load up all the regions for all services even if we are only checking 1-2 services, so there could be unexpected latency in execution.

williambrady:
Submitted PR for potential improvement to these scans.

jfagoagas:
Hi @williambrady, thank you for your contribution, we are going to include this map of regions and services for Prowler v3. 

Regarding your issue and the related PR, we are not being able to reproduce your issue because these checks correctly handle errors in the event that the audited region does not support the AppStream service. Please, take a look here:
- `extra7190`: https://github.com/prowler-cloud/prowler/blob/c1d60054e996ed50bb33024077e1344fc1fb9643/checks/check_extra7190#L31
- `extra7191`: https://github.com/prowler-cloud/prowler/blob/c1d60054e996ed50bb33024077e1344fc1fb9643/checks/check_extra7191#L31
- `extra7192`: https://github.com/prowler-cloud/prowler/blob/c1d60054e996ed50bb33024077e1344fc1fb9643/checks/check_extra7192#L31
- `extra7193`: https://github.com/prowler-cloud/prowler/blob/c1d60054e996ed50bb33024077e1344fc1fb9643/checks/check_extra7193#L31

Thanks!

jfagoagas:
Hi @williambrady, I've seen the problem.

[...]

PR #1294, opened 2022-07-29
Issue1293: Updated region detection for appstream service scans

williambrady:
### Context 

Address concerns in #1293 

The service does not exist in all AWS regions, so results were unreliable.

### Description

Added logic to the individual scans to compare the AWS Regions available to the account with the AWS Regions the service is available in to ensure only valid regions are scanned.

### License


jfagoagas:
I'm closing this PR because the problem here is that these checks are not handling correctly the timeout errors returned by AWS AppStream service.

PR #1328, opened 2022-08-08
fix(check120): correct AWS support policy name

JArmandoG:
According to the official documentation, which the check refers to (URL provided by the variable` $CHECK_DOC_check120`), the policy is not "`AWSSupportAccess`" but "`AWSSupportServiceRolePolicy`". The policy `AWSSupportAccess` check returns FAIL because it is not attached to any entity (The response in line 31 is empty), and it seems correct for this check to call the `AWSSupportServiceRolePolicy` instead

### Context 

I was getting FAIL for check120, even though I had the correct configuration according to this check's official AWS documentation so I decided to debug the code and found that AWSSupportServiceRolePolicy is the correct role to extract information from in order to verify this check

**NOTE: I followed only the provided URL for this check to debug and correct this** [https://docs.aws.amazon.com/awssupport/latest/user/using-service-linked-roles-sup.html](https://docs.aws.amazon.com/awssupport/latest/user/using-service-linked-roles-sup.html)

### License

### DIFF
--- a/checks/check120
+++ b/checks/check120
@@ -28,7 +28,7 @@ CHECK_CAF_EPIC_check120='IAM'
 
 check120(){
   # "Ensure a support role has been created to manage incidents with AWS Support (Scored)"
-  SUPPORTPOLICYARN=$($AWSCLI iam list-policies --query "Policies[?PolicyName == 'AWSSupportAccess'].Arn" $PROFILE_OPT --region $REGION --output text)
+  SUPPORTPOLICYARN=$($AWSCLI iam list-policies --query "Policies[?PolicyName == 'AWSSupportServiceRolePolicy'].Arn" $PROFILE_OPT --region $REGION --output text)
   if [[ $SUPPORTPOLICYARN ]];then
     for policyarn in $SUPPORTPOLICYARN;do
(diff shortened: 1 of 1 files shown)


JArmandoG:
I just checked the documentation ([https://docs.aws.amazon.com/awssupport/latest/user/accessing-support.html](https://docs.aws.amazon.com/awssupport/latest/user/accessing-support.html)) and it seems that the policy called `AWSSupportAccess` (The original one Prowler checks for), which "_grants an IAM user access to AWS Support. An IAM user with this policy can access all AWS Support operations and resources_." is what Prowler originally checked for. 

If this situation is what we're trying to check for in check120, and I'm wrong about this PR, then maybe we should change the URL in the variable `$CHECK_DOC_check120`

MrCloudSec:
Hi @JArmandoG, as you mentioned above, this check is for verifying if any IAM identity has access to  manage incidents with AWS Support. However, `AWSSupportServiceRolePolicy` is for allowing AWS Support to manage incidents in your account. I have modify the related URL for the correct one that you indicated.
Thank you for four contribution!

MrCloudSec (approved):

PR #1332, opened 2022-08-11
fix(csv_separator): Change CSV separator to pipe

joeskop:
News seperator "PIPE"

### Context 

Exporting on CSV has many problems due some descriptions that includes the comma "," and break the separators count.

### Description

Just, simple change the variable "SEP" from "," to "|".

### License

### DIFF
--- a/include/default_variables
+++ b/include/default_variables
@@ -46,7 +46,7 @@ export OUTPUT_DATE
 # Default Prowler Options
 QUIET=0
 export QUIET
-SEP=','
+SEP='|'
 export SEP
 KEEPCREDREPORT=0
 export KEEPCREDREPORT


MrCloudSec:
Hi @joeskop, thank you for your contribution! We will review it shortly and let you know if this change can fit in all Prowler's CSV output functionalities.

StylusFrost:
Hi @joeskop, If you make this change it no longer fulfils the [CSV standard ](https://www.ietf.org/rfc/rfc4180.txt ).The standard indicates that you must include between the character " when there is the possibility of a comma in the data. This may cause problems with importing data in the future.
image

MrCloudSec:
Hi @joeskop, as @StylusFrost mentioned above, these changes do not comply the CSV standard. But this problem will be solved in version 3 of Prowler. Thank you anyway for your suggestion!

toniblyx:
We will use `;` in v3 for CSV as default separator.

PR #1338, opened 2022-08-22
fix(s3_regions): verify if there are filter regions

MrCloudSec:
### Context 

When running S3 all buckets where listed despite the audited regions.

### Description

Only list the buckets from the audited regions, if any.

### License


jfagoagas (approved):
🔝

PR #1339, opened 2022-08-22
fix(mkdir_shub): mkdir when using security-hub

MrCloudSec:
### Context 

Make output directory when using security hub feature.

### Description

Make output directory when using security hub feature.

### License


jfagoagas (approved):
Good catch!

PR #1341, opened 2022-08-24
feat(shub_compatibility): send finding to filter regions and change checkType to list

MrCloudSec:
### Context 

When checking global services we need to tell Prowler where to send the findings.
Also, SecurityHub change the RelatedRequirements attribute to a list.

### Description

- When using -f, send SecurityHub finding of global services to the first filter region.
- Change checkType attribute of metadata to list.

### License

### DIFF
--- a/lib/check/models.py
+++ b/lib/check/models.py
@@ -57,7 +57,7 @@ class Check_Metadata_Model(BaseModel):
     # CheckName: str
     CheckTitle: str
     # CheckAlias: str
-    CheckType: str
+    CheckType: List[str]
     ServiceName: str
     SubServiceName: str
     ResourceIdTemplate: str
--- a/providers/aws/aws_provider.py
+++ b/providers/aws/aws_provider.py
@@ -322,3 +322,10 @@ def generate_regional_clients(service, audit_info):
         regional_clients[region] = regional_client
         # regional_clients.append(regional_client)
     return regional_clients
+
+
+def get_region_global_service(audit_info):
+    # Check if global service to send the finding to first audited region
+    if audit_info.audited_regions:
+        return audit_info.audited_regions[0]
+    return audit_info.profile_region
--- a/providers/aws/aws_provider.py
+++ b/providers/aws/aws_provider.py
@@ -304,7 +304,7 @@ def get_organizations_metadata(
         return organizations_info
 
 
-def generate_regional_clients(service, audit_info):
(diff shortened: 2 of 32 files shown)


lib/check/models.py
```diff
@@ -57,7 +57,7 @@ class Check_Metadata_Model(BaseModel):
     # CheckName: str
     CheckTitle: str
     # CheckAlias: str
-    CheckType: str
```
jfagoagas, line 60: 🔝

providers/aws/aws_provider.py
```diff
@@ -322,3 +322,10 @@ def generate_regional_clients(service, audit_info):
         regional_clients[region] = regional_client
         # regional_clients.append(regional_client)
     return regional_clients
+
+
+def get_region_global_service(audit_info):
+    # Check if global service to send the finding to first audited region
+    if audit_info.audited_regions:
+        return audit_info.audited_regions[0]
+    return audit_info.profile_region
```
jfagoagas: Please, include a test for this function

providers/aws/aws_provider.py
```diff
@@ -322,3 +322,10 @@ def generate_regional_clients(service, audit_info):
         regional_clients[region] = regional_client
         # regional_clients.append(regional_client)
     return regional_clients
+
+
+def get_region_global_service(audit_info):
```
jfagoagas: Also, include the return type `-> <type>`

jfagoagas (changes requested):
Good job @sergargar, take a look at my comments above, thanks!

providers/aws/aws_provider.py
MrCloudSec: Done

providers/aws/aws_provider.py
MrCloudSec: Why is it necessary for this simple function?

providers/aws/aws_provider_test.py
```diff
@@ -166,3 +167,25 @@ def test_organizations(self):
...
+            credentials=None,
+            assumed_role_info=None,
+            audited_regions=["eu-west-2", "eu-west-1"],
+            organizations_metadata=None,
+        )
+
+        assert (
+            get_region_global_service(input_audit_info)
+            == input_audit_info.audited_regions[0]
+        )
```
jfagoagas, line 191: Wel done!! 👏

jfagoagas (approved):
Good job @sergargar !

PR #1342, opened 2022-08-24
feat(api_banner): remove API region from banner

MrCloudSec:
### Context 

Remove API region from banner since there is no -r option.

### Description

Remove API region from banner since there is no -r option.

### License


jfagoagas (approved):
🔝 Thanks @sergargar

PR #1343, opened 2022-08-24
feat(s3_output): send outputs to S3 bucket

MrCloudSec:
### Context 

Add -B flag for sending outputs to a custom S3 bucket, for example: `s3::bucket/output_folder/extension/filename`

### Description

- Add -B flag for sending outputs to a custom S3 bucket.
- Mkdir when there is a custom directory

### License


prowler
```diff
@@ -247,12 +254,8 @@ if __name__ == "__main__":
...
     if output_directory:
         if not isdir(output_directory):
-            if output_directory == default_output_directory:
-                if output_modes:
-                    mkdir(default_output_directory)
-            else:
-                logger.critical("Output directory does not exist")
-                sys.exit()
+            if output_modes:
+                mkdir(output_directory)
```
jfagoagas: You need to check if that directory exists, if not it will fail
Screenshot 2022-08-25 at 07 58 39

lib/outputs/outputs.py
```diff
@@ -252,3 +254,29 @@ def close_json(output_directory, audited_account, mode):
...
+        elif output_mode == "json":
+            filename = f"prowler-output-{audit_info.audited_account}-{json_file_suffix}"
+        elif output_mode == "json-asff":
+            filename = (
+                f"prowler-output-{audit_info.audited_account}-{json_asff_file_suffix}"
+            )
+        logger.info(f"Sending outputs to S3 bucket {output_bucket}")
+        # Check if security hub is enabled in current region
+        s3_client = audit_info.audit_session.client("s3")
+        s3_client.upload_file(
```
jfagoagas: Here you need to test if that file exists in the output location directory.

lib/outputs/outputs.py
```diff
@@ -252,3 +254,29 @@ def close_json(output_directory, audited_account, mode):
     file_descriptor.truncate()
     file_descriptor.write("]")
     file_descriptor.close()
+
+
+def send_to_s3_bucket(output_directory, output_mode, output_bucket, audit_info):
```
jfagoagas: Please, include tests for this function.

lib/outputs/outputs.py
```diff
@@ -252,3 +254,29 @@ def close_json(output_directory, audited_account, mode):
...
+            filename = (
+                f"prowler-output-{audit_info.audited_account}-{json_asff_file_suffix}"
+            )
+        logger.info(f"Sending outputs to S3 bucket {output_bucket}")
+        # Check if security hub is enabled in current region
+        s3_client = audit_info.audit_session.client("s3")
+        s3_client.upload_file(
+            output_directory + "/" + filename,
+            output_bucket,
+            output_directory + "/" + output_mode + "/" + filename,
```
jfagoagas: Also, for a better readability:
```
file_name =  output_directory + "/" + filename
bucket_name = output_bucket,
object_name = output_directory + "/" + output_mode + "/" + filename

s3_client.upload_file(file_name, bucket_name, object_name)
```

jfagoagas (changes requested):
Good job @sergargar, please check my comments above!

prowler
MrCloudSec: I'm checking it with the conditional `if not isdir(output_directory):`

lib/outputs/outputs.py
MrCloudSec: Sure

lib/outputs/outputs.py
MrCloudSec: Why would not it exist?

lib/outputs/outputs.py
MrCloudSec: Nice idea!

lib/outputs/outputs.py
jfagoagas: I don't know but it's better to check it

prowler
jfagoagas: You're totally right 🔝

lib/outputs/outputs_test.py
```diff
@@ -254,10 +257,68 @@ def test_fill_json_asff(self):
...
+            client.get_object(
+                Bucket=bucket_name,
+                Key=output_directory
+                + "/"
+                + output_mode
+                + "/"
+                + filename
+                + csv_file_suffix,
+            )["ContentType"]
+            == "binary/octet-stream"
```
jfagoagas, line 322: Why don't we check if the uploaded file match with the generated file?

I prefer to check the file name.

jfagoagas (changes requested):

jfagoagas (approved):
Well done @sergargar 🥇

PR #1345, opened 2022-08-25
feat(custom_filename): custom output filename

MrCloudSec:
### Context 

Add flag `-F` to put a custom filename for Prowler output reports.

### Description

Add flag `-F` to put a custom filename for Prowler output reports.

### License

### DIFF
--- a/lib/outputs/outputs.py
+++ b/lib/outputs/outputs.py
@@ -9,6 +9,7 @@ from config.config import (
     csv_file_suffix,
     json_asff_file_suffix,
     json_file_suffix,
+    output_file_timestamp,
     prowler_version,
     timestamp_iso,
     timestamp_utc,
@@ -42,6 +43,7 @@ def report(check_findings, output_options, audit_info):
             audit_info.audited_account,
             output_options.output_directory,
             csv_fields,
+            output_options.output_filename,
         )
 
     if check_findings:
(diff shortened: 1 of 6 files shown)


lib/outputs/outputs.py
```diff
@@ -101,13 +105,16 @@ def report(check_findings, output_options, audit_info):
...
 
 
-def fill_file_descriptors(output_modes, audited_account, output_directory, csv_fields):
+def fill_file_descriptors(
+    output_modes, audited_account, output_directory, csv_fields, output_filename
+):
     file_descriptors = {}
+    # If no input custom output filename, set default:
+    if not output_filename:
+        output_filename = f"prowler-output-{audited_account}-{output_file_timestamp}"
```
jfagoagas: You are doing the same at lines 250 and 251, why don't you include this at a higher level so both functions always receive a non-empty `output_filename`. What do you think?

prowler
```diff
@@ -132,9 +132,16 @@ if __name__ == "__main__":
         help="Output mode, by default csv",
         choices=["csv", "json", "json-asff"],
     )
+    parser.add_argument(
+        "-F",
+        "--output-filename",
+        nargs="?",
+        default=None,
+        help="Custom output report name, if not specified will use default output/prowler-output-ACCOUNT_NUM-OUTPUT_DATE.format.",
+    )
```
jfagoagas, line 141: 👏 👏 👏 👏

jfagoagas (changes requested):

lib/outputs/outputs.py
MrCloudSec: Great idea! I put it at a high level now.

jfagoagas (approved):
Good job @sergargar it looks great!!!!

PR #1350, opened 2022-08-28
fix(check_network_acl): check with all rules together

StylusFrost:
### Context 

Checking the behaviour of the network ACL rule checker, it seems that it only checks one rule at a time and in order. This is incorrect from my point of view. The result of a network ACL check should use all your rules together, also check the priority of these rules and don't forget the deny rules.

#### 1st Topic : Priority Rules and Explicit Deny Rules

As you can see in the AWS documentation [Network ACL Priority](https://docs.aws.amazon.com/vpc/latest/userguide/vpc-network-acls.html) priority is very important when applying Network ACL rules. 

image

The Boto3 API and the AWS client are returning for the same ACL network the rules in ascending order. Therefore, the code is checking the rules in order but it would be nice if the code ordered them. Even if this order is respected, the explicit deny rules are not checked. Note that users sometimes play with activating and deactivating rules temporarily, so it is normal to find explicit deny rules. 

Several examples with false positives:

- Explicit deny rule with high priority:

image

image

- All traffic denied with a higher priority than the allowed rule:

image

image

- First deny and later allowed:

image

image

#### 2nd Topic : Explicit deny rule with IPv4 only or IPv4 and IPv6 together

Rules for IPv6 can be added without the Network ACL being applied to an IPv6 enabled subnet. As soon as there is a rule with the public address ::/0, it is assumed that IPv6 access is also checked, even if it is not actually activated.
It is necessary to check IPv6 rules separately from IPv4 rules. For example, in cases where there are explicit deny for IPv6 and not for IPv4 and vice versa. If for IPv6 there is a rule that allows public access ::/0, you don't need to very any further. If everything is correct for IPv6, you will have to check IPv4 rules.

I have implemented a new check_network_acl which I think is more complete and fix these problems. I changed the 2 checks that use it too.

I have attached a test file to verify various situations described above.

*** I think same problem exists in the master branch but I have not verified it in depth.

### License

### DIFF
--- a/providers/aws/services/ec2/lib/network_acls_test.py
+++ b/providers/aws/services/ec2/lib/network_acls_test.py
@@ -1,0 +1,1106 @@
+from providers.aws.services.ec2.lib.network_acls import check_network_acl
+
+default_deny_entry_ingress_IPv4 = {
+    "CidrBlock": '0.0.0.0/0',
+    "Egress": False,
+    "NetworkAclId": "acl-072d520d07e1c1471",
+    "Protocol": "-1",
+    "RuleAction": 'deny',
+    "RuleNumber": 32767}
+
+default_deny_entry_ingress_IPv6 = {
+    "Ipv6CidrBlock": '::/0',
+    "Egress": False,
+    "NetworkAclId": "acl-072d520d07e1c1471",
+    "Protocol": "-1",
+    "RuleAction": 'deny',
+    "RuleNumber": 32768}
+
+default_deny_entry_egress_IPv4 = {
+    "CidrBlock": '0.0.0.0/0',
+    "Egress": True,
+    "NetworkAclId": "acl-072d520d07e1c1471",
+    "Protocol": "-1",
+    "RuleAction": 'deny',
+    "RuleNumber": 32767}
+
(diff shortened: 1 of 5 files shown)


providers/aws/services/ec2/lib/network_acls_test.py
```diff
@@ -0,0 +1,1106 @@
...
+    "RuleAction": 'allow',
+    "RuleNumber": 32766}
+
+allow_all_entry_ingress_IPv6 = {
+    "Ipv6CidrBlock": '::/0',
+    "Egress": False,
+    "NetworkAclId": "acl-072d520d07e1c1471",
+    "Protocol": "-1",
+    "RuleAction": 'allow',
+    "RuleNumber": 32766}
```
jfagoagas, line 49: Impressive!!!! 🥇

jfagoagas (approved):
Awesome! This test battery looks amazing 🔥 🔥 🔥 

Thank you very much for your contribution @StylusFrost !!

MrCloudSec (approved):
This is awesome, @StylusFrost 👏🏼 Thank you for the contribution!

PR #1356, opened 2022-09-01
feat(shodan_integration): add ec2_elastic_ip_shodan check and config yaml

MrCloudSec:
### Context 

Include shodan integration as well as the ec2 check.

### Description

- Add ec2_elastic_ip_shodan check 
- Add shodan integration (-N flag)
- Add config yaml feature

### License


config/config.py
```diff
@@ -18,3 +20,21 @@
...
+
+
+def change_config_var(variable, value):
+    with open(config_yaml) as f:
+        doc = yaml.safe_load(f)
+
+    doc[variable] = value
+
+    with open(config_yaml, "w") as f:
+        yaml.dump(doc, f)
```
jfagoagas, line 33: Why do we need this? I think this config file must be read only.

providers/aws/config.yaml
```diff
@@ -0,0 +1 @@
+shodan_api_key: null
```
jfagoagas, line 1: Why `null` ?

I think for the checks we need to use the config file instead of command line arguments. It's better to only have on path for that kind of operation.

providers/aws/services/ec2/ec2_elastic_ip_shodan/ec2_elastic_ip_shodan.py
```diff
@@ -0,0 +1,46 @@
...
+
+from config.config import get_config_var
+from lib.check.models import Check, Check_Report
+from lib.logger import logger
+from providers.aws.services.ec2.ec2_client import ec2_client
+
+
+class ec2_elastic_ip_shodan(Check):
+    def execute(self):
+        findings = []
```
jfagoagas, line 11: Please, include the `try-catch` for all the check code.

providers/aws/services/ec2/ec2_elastic_ip_shodan/ec2_elastic_ip_shodan.py
```diff
@@ -0,0 +1,46 @@
...
+                        if "No information available for that IP" in error.value:
+                            report.status = "PASS"
+                            report.status_extended = (
+                                f"Elastic IP {eip.public_ip} is not listed in Shodan."
+                            )
+                            report.resource_id = eip.public_ip
+                            findings.append(report)
+                            continue
+                        else:
+                            logger.error(f"{error.value}")
```
jfagoagas: Please, add more information in the error message.

providers/aws/services/ec2/ec2_service.py
```diff
@@ -228,9 +254,26 @@ def __init__(self, name, region, id, ingress_rules, egress_rules):
...
+    vpc: str
+    subnet: str
+    region: str
+
+    def __init__(self, public_ip, public_dns, vpc, subnet, region):
+        self.public_ip = public_ip
+        self.public_dns = public_dns
+        self.vpc = vpc
+        self.subnet = subnet
+        self.region = region
```
jfagoagas: awesome!!

prowler
```diff
@@ -175,6 +179,13 @@ if __name__ == "__main__":
         default=None,
         help="Same as -B but do not use the assumed role credentials to put objects to the bucket, instead uses the initial credentials.",
     )
+    parser.add_argument(
+        "-N",
+        "--shodan",
+        nargs="?",
+        default=None,
+        help="Shodan API key used by check ec2_elastic_ip_shodan.",
+    )
```
jfagoagas, line 188: As I said above, for the check's input parameters we should only use the new configuration file.

prowler
```diff
@@ -218,6 +229,9 @@ if __name__ == "__main__":
         print_services(list_services(provider))
         sys.exit()
 
+    if args.shodan:
+        change_config_var("shodan_api_key", args.shodan)
+
```
jfagoagas, line 234: Checks comments above for the configuration file.

jfagoagas (changes requested):
Good job @sergargar, please check my comments in order to merge this PR 🚀

providers/aws/config.yaml
jfagoagas: If not it would be a little mess trying to find what value is the right one.

jfagoagas:
Also, I'm missing the tests. Please include them.

providers/aws/config.yaml
MrCloudSec: I think that we have to have the same functionality as version 2, at least at the beginning. That's why the best option for keeping both feature is to do it this way.

prowler
MrCloudSec: I think that we have to have the same functionality as version 2, at least at the beginning. That's why the best option for keeping both feature is to do it this way.

config/config.py
MrCloudSec: I think that we have to have the same functionality as version 2, at least at the beginning. That's why the best option for keeping both feature is to do it this way.

providers/aws/services/ec2/ec2_elastic_ip_shodan/ec2_elastic_ip_shodan.py
MrCloudSec: Sure!

providers/aws/services/ec2/ec2_elastic_ip_shodan/ec2_elastic_ip_shodan.py
MrCloudSec: Don't think is needed.

jfagoagas (approved):

PR #1370, opened 2022-09-15
Dockerfile build instructions

walkerab:
Build instructions referenced Dockerfile that has moved.

Could also just drop the `-f` option.

### License


Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Build command
-# docker build --platform=linux/amd64  --no-cache -t prowler:latest -f util/Dockerfile .
+# docker build --platform=linux/amd64  --no-cache -t prowler:latest -f ./Dockerfile .
```
walkerab, line 2: Could also just do:

```suggestion
# docker build --platform=linux/amd64  --no-cache -t prowler:latest .
```

MrCloudSec (approved):
Good catch! Thank you @walkerab

PR #1383, opened 2022-10-05
fix(checks): Handle checks not returning result

n4ch04:
### Context 

There exists some checks that under certain conditions don't return any results

### Description

Modify those checks to return results when there are not resources checked:
- extra72
- extra74
- extra729
- extra7164

### License


jfagoagas (approved):
Thanks @n4ch04, good job!!

PR #1397, opened 2022-10-10
fix(inventory): quick inventory input fixed

MrCloudSec:
### Context 

Quick inventory option `-i` is not working.

### Description

`-i` option is asking for an argument but it is not needed, so remove the extra `:`

### License


jfagoagas (approved):
Good catch!!

PR #1402, opened 2022-10-12
fix(check_extra77): Add missing check_resource_id to the report

kagahd:
### Context 

check_extra77 was missing to print out the resource id when the check failed

### Description

I just added the repo name to be printed out if the check fails as it's done in several places already if the check passes.

### License


MrCloudSec (approved):
Hi @kagahd, good catch! Thank you for your contribution 🥇

PR #1409, opened 2022-10-17
fix(extra77): Deleted resource id from exception results

n4ch04:
### Context 

Check extra77 includes resource id field when exceptions are raised

### Description

Not include resource id in results when an exception is raised

### License


jfagoagas (approved):
Awesome @n4ch04 🍪

PR #1410, opened 2022-10-17
fix(extra7183): Exception handling error UnsupportedOperationException

n4ch04:
### Context 

Check extra7183 does not handle UnsupportedOperationException response from `list-certificates` api calls

### Description

Include exception handling for that api call, exit the loop and label the result as INFO

### License


jfagoagas (approved):
Good job @n4ch04 !!

PR #1420, opened 2022-10-19
delete(extra7145): It is redundant to check_extra798

kagahd:
### Context 

check_extra7145 does the same as check_extra798 and can therefore be deleted

### Description

Even the implementations of both checks are almost identical. The only differences are:
1. slightly different wording of `textFail`, `textPass` and `textInfo`
2. output format of `lambda get-policy` is `text` in extra798 but `json` in extra7145
3. since the output formats are different, extra7145 uses additionally `jq '. | fromjson'` 
4. extra7142 tests explicitly on a `ResourceNotFoundException` while extra798 just tests on existence of a policy

### License


MrCloudSec:
Hi @kagahd, good catch, thank you for your contribution!

But, why did you remove the cron from the [refresh_aws_services_regions.yml](https://github.com/prowler-cloud/prowler/pull/1420/commits/e15d07c8bf6cea3a50521f42d7826146ffe75386)?

kagahd:
> But, why did you remove the cron from the [refresh_aws_services_regions.yml](https://github.com/prowler-cloud/prowler/pull/1420/commits/e15d07c8bf6cea3a50521f42d7826146ffe75386)?

I removed it from my forked repo because the workflow [run failed](https://github.com/kagahd/prowler/actions/runs/3288360800) and alerted me daily by e-mail. However, I [restored](https://github.com/prowler-cloud/prowler/pull/1420/commits/e15d07c8bf6cea3a50521f42d7826146ffe75386) the file but to avoid the daily alerts I commented out the cron. I guess I shouldn't have done this, right? But how to avoid the failed runs respectively the daily alerts?

kagahd:
I may have found out by myself why the workflow failed: it uses the branch `prowler-3.0-dev` which my forked repo hadn't had. To fix it, I just added that branch to my forked repo. Since then, the [workflow is running without problems](https://github.com/kagahd/prowler/actions/runs/3340992112), so I activated the cron again as it's the case in the upstream prowler repo.
Now you should be able to merge without conflicts.

kagahd:
I see that the status is still `awaiting response` even though I answered your question already @sergargar. Do you need more input from me?

MrCloudSec:
Hi @kagahd, sorry for the delayed response.
The check is duplicated and we included it in the roadmap of the new version 3 that we will release soon.
Stay tuned for the release, thank you again for letting us know!

PR #1422, opened 2022-10-20
feat(line_no): Add line number to errors

MrCloudSec:
### Context 

To enhance debugging we need to add line numbers in error logs.

### Description

Add line numbers to the logs.

### License

### DIFF
--- a/lib/check/check.py
+++ b/lib/check/check.py
@@ -205,7 +205,9 @@ def run_check(check, audit_info, output_options):
         findings = check.execute()
     except Exception as error:
         print(f"Something went wrong in {check.checkID}, please use --log-level ERROR")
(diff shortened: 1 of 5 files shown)


jfagoagas (changes requested):
This is great but I think we need to encapsulate this behaviour in this way:
```
from lib import logger

def handle_logger(level, exception, region):
    logger.error(
                f"{self.region} -- {error.__class__.__name__}[{error.__traceback__.tb_lineno}]: {error}"
    )
```

then we you need to handle an exception you can do:
```
handle_logger(LEVEL_ERROR, error, region)
```

What do you think?

jfagoagas (approved):
Good job 👏

PR #1425, opened 2022-10-24
fix(allowlist): allowlist file default value

n4ch04:
### Context 

The allowlist feature needs to have a default value in case of a non-set flag

### Description

Deletes a debug print into allowlist parsing process and sets default value for allowlist file

### License


MrCloudSec (approved):
good catch 🔝

PR #1446, opened 2022-11-02
feat(account): add account service and manual checks

MrCloudSec:
### Context 

Add checks and service for account since we need to migrate them to the new version 3.

### Description

Add checks and service for account.

### License


jfagoagas (approved):
Let's go 🚀

PR #1505, opened 2022-11-20
fix(): Fix CloudTrail trail S3 logging public bucket false positive result when trail bucket doesn't exist

acknosyn:
### Context 

I noticed that when a trail specifies an S3 bucket and it was deleted (but still configured in the trail), the check would incorrectly mark the bucket as publicly accessible.

Running the `aws s3api get-bucket-location` command for a trail configured with a deleted bucket returns a `NoSuchBucket` error.

The bucket skips all other checks and ends up at the end of the while loop which contains the following condition:

CleanShot 2022-11-21 at 12 38 18

The condition evaluates to `false` (because the `aws s3api get-bucket-acl` command also returns the same error) and ends up assigning the trail with a publicly accessible bucket.

### Description

I added a check after the `get-bucket-location` command following the `AccessDenied` check to check for the `NoSuchBucket` error and marked the trail as having a logging bucket that does not exist.

### License

### DIFF
--- a/checks/check23
+++ b/checks/check23
@@ -57,6 +57,11 @@ check23(){
             textInfo "$regx: Trail ${TRAIL_NAME} with home region ${TRAIL_HOME_REGION} Access Denied getting bucket location for bucket $TRAIL_BUCKET" "$regx" "$TRAIL_NAME"
             continue
           fi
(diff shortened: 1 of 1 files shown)


MrCloudSec (approved):
Hi @acknosyn, good catch! Thank you for this contribution 😄

PR #1507, opened 2022-11-21
fix(README): include more details about db connector

n4ch04:
### Context 

Includes more details about the db connector.

### Description

To work, the db connector needs the `-u` flag set, and in the docs it is detailed as optional. Changing that.

### License


jfagoagas (approved):
🔝

PR #1528, opened 2022-12-01
feat(azure): multiple azure auths

n4ch04:
### Context 

Prowler only allowed to launch the azure scan using service principal credentials, we need to include the other auth methods supported by the sdk

### Description

Adds support to Prowler azure for the following auth methods:

- az cli
- browser credentials
- managed identity

### License

### DIFF
--- a/lib/outputs/outputs.py
+++ b/lib/outputs/outputs.py
@@ -49,13 +49,14 @@ def report(check_findings, output_options, audit_info):
         # csv_fields = []
         file_descriptors = {}
         if output_options.output_modes:
-            # We have to create the required output files
-            file_descriptors = fill_file_descriptors(
-                output_options.output_modes,
-                output_options.output_directory,
-                output_options.output_filename,
-                audit_info,
-            )
+            if not hasattr(audit_info, "is_azure"):
+                # We have to create the required output files
+                file_descriptors = fill_file_descriptors(
+                    output_options.output_modes,
+                    output_options.output_directory,
+                    output_options.output_filename,
+                    audit_info,
+                )
 
         if check_findings:
             for finding in check_findings:
@@ -325,6 +326,7 @@ def generate_csv_fields(format: Any) -> list[str]:
 
 
 def fill_json(finding_output, audit_info, finding):
+
     finding_output.AssessmentStartTime = timestamp_iso
     finding_output.FindingUniqueId = ""
     finding_output.Profile = audit_info.profile
@@ -457,7 +459,7 @@ def send_to_s3_bucket(
 
 def display_summary_table(
     findings: list,
-    audit_info: AWS_Audit_Info,
+    audit_info,
     output_options: Output_From_Options,
     provider: str,
 ):
@@ -466,8 +468,14 @@ def display_summary_table(
     try:
         if provider == "aws":
             entity_type = "Account"
+            audited_entities = audit_info.audited_account
         elif provider == "azure":
-            entity_type = "Tenant Domain"
+            if audit_info.identity.domain:
+                entity_type = "Tenant Domain"
+                audited_entities = audit_info.identity.domain
+            else:
+                entity_type = "Tenant ID/s"
+                audited_entities = " ".join(audit_info.identity.tenant_ids)
 
         if findings:
             current = {
@@ -531,12 +539,13 @@ def display_summary_table(
                 ]
             ]
             print(tabulate(overview_table, tablefmt="rounded_grid"))
+
             print(
-                f"\n{entity_type} {Fore.YELLOW}{audit_info.audited_account}{Style.RESET_ALL} Scan Results (severity columns are for fails only):"
+                f"\n{entity_type} {Fore.YELLOW}{audited_entities}{Style.RESET_ALL} Scan Results (severity columns are for fails only):"
             )
             if provider == "azure":
                 print(
-                    f"\nSubscriptions scanned: {Fore.YELLOW}{' '.join(audit_info.subscriptions.keys())}{Style.RESET_ALL}"
+                    f"\nSubscriptions scanned: {Fore.YELLOW}{' '.join(audit_info.identity.subscriptions.keys())}{Style.RESET_ALL}"
                 )
             print(tabulate(findings_table, headers="keys", tablefmt="rounded_grid"))
             print(
@@ -552,7 +561,7 @@ def display_summary_table(
 
         else:
             print(
-                f"\n {Style.BRIGHT}There are no findings in {entity_type} {Fore.YELLOW}{audit_info.audited_account}{Style.RESET_ALL}\n"
+                f"\n {Style.BRIGHT}There are no findings in {entity_type} {Fore.YELLOW}{audited_entities}{Style.RESET_ALL}\n"
             )
 
     except Exception as error:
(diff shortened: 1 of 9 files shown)


providers/azure/azure_provider.py
```diff
@@ -11,82 +11,186 @@
...
+                    exclude_environment_credential=not sp_env_auth,
+                    exclude_cli_credential=not az_cli_auth,
+                    exclude_managed_identity_credential=not managed_entity_auth,
+                    exclude_visual_studio_code_credential=True,
+                    exclude_shared_token_cache_credential=True,
+                    exclude_powershell_credential=True,
+                )
+            except Exception as error:
+                logger.critical("Failed to retrieve azure credentials")
+                logger.critical(
```
jfagoagas, line 58: Include the `sys.exit()` here as we are logging some critical event.

providers/azure/azure_provider.py
```diff
@@ -11,82 +11,186 @@
...
+                )
+        else:
+            credentials = InteractiveBrowserCredential()
 
+        return credentials
 
-def check_credential_env_vars() -> Azure_Identity_Info:
-    for env_var in ["AZURE_CLIENT_ID", "AZURE_TENANT_ID", "AZURE_CLIENT_SECRET"]:
-        if not getenv(env_var):
+    def __check_sp_creds_env_vars__(self):
```
jfagoagas, line 67: I prefer this name since the only way to check the service principal credentials is using environment variables.

```suggestion
    def __check_service_principal_credentials(self):
```

providers/azure/azure_provider.py
```diff
@@ -11,82 +11,186 @@
...
+        credentials,
+        az_cli_auth,
+        sp_env_auth,
+        browser_auth,
+        managed_entity_auth,
+        subscription_ids,
+    ):
+        identity = Azure_Identity_Info()
+
+        # If credentials comes from service principal or browser, it the needed permissions are assigned
```
jfagoagas:
```suggestion
        # If credentials comes from service principal or browser, if the required permissions are assigned
```

providers/azure/azure_provider.py
```diff
@@ -11,82 +11,186 @@
...
+                )
+            # since that exception is not considered as critical, we keep filling another identity fields
+            if sp_env_auth:
+                # The id of the sp can be retrieved from environment variables
+                identity.identity_id = getenv("AZURE_CLIENT_ID")
+                identity.identity_type = "Service Principal"
+            # Same here, if user can access AAD, some fields are retrieved if not, default value, for az cli
+            # should work but it doesn't, pending issue
+            else:
+                identity.identity_id = "unkown user id (NO AAD permissions)"
```
jfagoagas:
```suggestion
                identity.identity_id = "Unknown user id (NO AAD permissions)"
```

providers/azure/azure_provider.py
```diff
@@ -11,82 +11,186 @@
...
+                except Exception as error:
+                    logger.error(
+                        "Provided identity does not have permissions to access AAD to retrieve user's metadata"
+                    )
+                    logger.error(
+                        f"{error.__class__.__name__}[{error.__traceback__.tb_lineno}] -- {error}"
+                    )
+        # Managed identities only can be assigned resource, resource group and subscription scope permissions
+        elif managed_entity_auth:
+            identity.identity_id = "Default managed identity id"
```
jfagoagas:
```suggestion
            identity.identity_id = "Default Managed Identity ID"
```

providers/azure/lib/audit_info/models.py
```diff
@@ -5,20 +5,20 @@
...
+    subscriptions: dict = {}
 
 
 @dataclass
 class Azure_Audit_Info:
     credentials: DefaultAzureCredential
     identity: Azure_Identity_Info
-    subscriptions: dict
-    audited_account: str
+    is_azure: bool
```
jfagoagas: Why do you need this?

prowler
```diff
@@ -370,13 +407,21 @@ if __name__ == "__main__":
...
-        )
+        if provider == "aws":
+            output_filename = (
+                f"prowler-output-{audit_info.audited_account}-{output_file_timestamp}"
+            )
+        elif provider == "azure":
+            if audit_info.identity.domain:
+                output_filename = f"prowler-output-{audit_info.identity.domain}-{output_file_timestamp}"
+            else:
+                output_filename = f"prowler-output-{'-'.join(audit_info.identity.tenant_ids)}-{output_file_timestamp}"
```
jfagoagas, line 424: Can we move this logic to outputs?

lib/outputs/outputs.py
```diff
@@ -49,13 +49,14 @@ def report(check_findings, output_options, audit_info):
...
         file_descriptors = {}
         if output_options.output_modes:
-            # We have to create the required output files
-            file_descriptors = fill_file_descriptors(
-                output_options.output_modes,
-                output_options.output_directory,
-                output_options.output_filename,
-                audit_info,
-            )
+            if not hasattr(audit_info, "is_azure"):
```
jfagoagas: If this is only for AWS, please include a comment to mark it because we will need to address it in the future. Also you can check if the `audit_info` is a class of `AWS_Audit_Info` with `isinstance(audit_info, AWS_Audit_Info)`, you don't need to include more fields.

providers/azure/azure_provider.py
```diff
@@ -11,82 +11,186 @@
...
+            try:
+                # Since the input vars come as True when it is wanted to be used, we need to inverse it since
+                # DefaultAzureCredential sets the auth method excluding the others
+                credentials = DefaultAzureCredential(
+                    exclude_environment_credential=not sp_env_auth,
+                    exclude_cli_credential=not az_cli_auth,
+                    exclude_managed_identity_credential=not managed_entity_auth,
+                    exclude_visual_studio_code_credential=True,
+                    exclude_shared_token_cache_credential=True,
+                    exclude_powershell_credential=True,
```
jfagoagas:
```suggestion
                    # Azure Auth using Visual Studio is not supported
                    exclude_visual_studio_code_credential=True,
                    # Azure Auth using Shared Token Cache is not supported
                    exclude_shared_token_cache_credential=True,
                    # Azure Auth using PowerShell is not supported
                    exclude_powershell_credential=True,
```

jfagoagas (changes requested):
Awesome!

providers/azure/azure_provider.py
n4ch04: We are actually getting the env variables that includes service principal information, we are not checking credentials

providers/azure/azure_provider.py
n4ch04: Included

providers/azure/azure_provider.py
n4ch04: Changed

providers/azure/azure_provider.py
n4ch04: Included

providers/azure/azure_provider.py
n4ch04: Included

prowler
n4ch04: Right now we need to keep it here since we are parsing here input options

providers/azure/lib/audit_info/models.py
n4ch04: Deleted

lib/outputs/outputs.py
n4ch04: Changed

jfagoagas (approved):

Issue #1573, opened 2022-12-21
[Bug]: Unable to complete scans for Chinese AWS accounts


daftkid:
**What happened?**
I've tried to run all-checks scan using Prowler v3, however, prowler suddenly stops execution and is only able to show me the report for the 3 services:

```
Date: 2022-12-21 16:47:49


This report is being generated using credentials below:

AWS-CLI Profile: [default] AWS Filter Region: [all]
AWS Account: [xxxxxxxxxx] UserId: [XXXXXXXXXXX:xxxxxx]
Caller Identity ARN: [arn:aws-cn:sts::xxxxxxxxxx:assumed-role/xxx/xxx]

Executing 244 checks, please wait...

-> Scan completed! |▉▉▉▉▉▉▉⚠︎                                | (!) 43/244 [18%] in 30.5s

Overview Results:
╭───────────────────┬───────────────────╮
│ 26.0% (13) Failed │ 68.0% (34) Passed │
╰───────────────────┴───────────────────╯

Account xxxxxxxxxx Scan Results (severity columns are for fails only):
╭────────────┬────────────┬───────────┬────────────┬────────┬──────────┬───────╮
│ Provider   │ Service    │ Status    │   Critical │   High │   Medium │   Low │
├────────────┼────────────┼───────────┼────────────┼────────┼──────────┼───────┤
│ aws        │ account    │ PASS (3)  │          0 │      0 │        0 │     0 │
├────────────┼────────────┼───────────┼────────────┼────────┼──────────┼───────┤
│ aws        │ cloudwatch │ PASS (1)  │          0 │      0 │        0 │     0 │
├────────────┼────────────┼───────────┼────────────┼────────┼──────────┼───────┤
│ aws        │ iam        │ FAIL (13) │          2 │      3 │        6 │     2 │
╰────────────┴────────────┴───────────┴────────────┴────────┴──────────┴───────╯
* You only see here those services that contains resources.
```

**How to reproduce it**
1. What command are you running?
`prowler aws -M HTML`
3. Environment you have, like single account, multi-account, organizations, etc.
Single AWS account in the Chinese partition
5. See error
After I enabled log-level=DEBUG I can see the following errors for each set of checks:
```
2022-12-21 16:52:24,271 [File: check.py:343]    [Module: check]  ERROR: KeyError[325]: 'aws-cn'
2022-12-21 16:52:24,274 [File: check.py:343]    [Module: check]  ERROR: KeyError[325]: 'aws-cn'
[...]

PR #1594, opened 2022-12-22
fix(global_services): handle global regions correctly

MrCloudSec:
### Description

- Fix Trusted Advisor service
- Set profile region as region for Account and CloudTrail services
- Fix the following global services: Route53, S3Control, CloudFront, IAM and Shield
https://github.com/prowler-cloud/prowler/issues/1573
### License


prowler/providers/aws/aws_provider.py
```diff
@@ -103,58 +103,80 @@ def assume_role(audit_info: AWS_Audit_Info) -> dict:
...
+                set(json_regions).intersection(audit_info.audited_regions)
+            )  # Get common regions between input and json
+        else:  # Get all regions from json of the service and partition
+            regions = json_regions
+        # Check if it is global service to gather only one region
+        if global_service:
+            if regions:
+                if audit_info.profile_region in regions:
+                    regions = [audit_info.profile_region]
+                regions = regions[:1]
```
jfagoagas, line 130: I don't know if you can do this, for example the `route53domains` service is not present in the regions JSON file. I think this logic needs to be at the top of the function because the general behaviour is going to fail for the non-comercial partitions.

prowler/providers/aws/aws_provider.py
```diff
@@ -103,58 +103,80 @@ def assume_role(audit_info: AWS_Audit_Info) -> dict:
...
+            json_regions = data["services"]["elb"]["regions"][
+                audit_info.audited_partition
+            ]
+        elif service == "s3control":
+            json_regions = data["services"]["s3"]["regions"][
+                audit_info.audited_partition
+            ]
+        elif service == "wafv2" or service == "waf-regional":
+            json_regions = data["services"]["waf"]["regions"][
+                audit_info.audited_partition
```
jfagoagas: We need to include this in the JSON generator.

prowler/providers/aws/services/trustedadvisor/trustedadvisor_service.py
```diff
@@ -12,54 +12,55 @@ class TrustedAdvisor:
     def __init__(self, audit_info):
         self.service = "support"
         self.session = audit_info.audit_session
-        self.regional_clients = generate_regional_clients(self.service, audit_info)
+        global_client = generate_regional_clients(
+            self.service, audit_info, global_service=True
+        )
```
jfagoagas: Good catch though

jfagoagas (changes requested):

prowler/providers/aws/aws_provider.py
MrCloudSec: Done

jfagoagas (approved):

n4ch04 (approved):

PR #1599, opened 2022-12-23
fix(send to s3): fixed send to s3 feature

n4ch04:
### Context

There was an issue when sending the output to a remote s3 bucket, due to the html default format.

### Description

Now the results are sent to the bucket in the three formats, by default in a directory called `output`

### License


jfagoagas (approved):
Well done @n4ch04 !!

PR #1617, opened 2022-12-26
Remove additional `apk update` in Dockerfile

PeterDaveHello:
### Context

I just happened to see this can be improved as I was passing by.

### Description

`apt upgrade` with `--no-cache` will do the cache update thing before the upgrade, just without leaving cache. Another `apt update` will only spend more time and network bandwidth ;)

### License

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@ FROM python:3.9-alpine
 LABEL maintainer="https://github.com/prowler-cloud/prowler"
 
 # Update system dependencies
-RUN apk --no-cache update && apk --no-cache upgrade
+RUN apk --no-cache upgrade
 
 # Create nonroot user
 RUN mkdir -p /home/prowler && \


jfagoagas:
Hi @PeterDaveHello, I think the `apk update` is required to get the latest index list of available packages but if you want to get rid of that command to only have the `apk upgrade` an additional `-U` flag is required, so the final command will be `apk --no-cache upgrade -U `

The official documentation states the above: https://wiki.alpinelinux.org/wiki/Upgrading_Alpine#Updating_package_lists

PeterDaveHello:
@jfagoagas `--no-cache` will do the index thing for you ;)

jfagoagas:
I've just saw it and it's include since Alpine Linux 3.3.

Thanks for the tip and the PR!

jfagoagas (approved):
🔝

PR #1632, opened 2023-01-02
fix(output_filename): Use custom output filename when set

### LINKED ISSUE #1616: [Bug]: Wrong default file naming: None.format
 - Resource: Docker container manually
 - OS: Mac
 - Prowler Version: 3.0.1
 - Parameters: -M html -B my-s3-bucket -o cis-benchmark

**What happened?**
As showed below, documentation says if custom output report name is not specified, filename will use default output/prowler-output-ACCOUNT_NUM-OUTPUT_DATE.format (-F flag), but instead of that, default name is None.format

image

**How to reproduce it**
As I'm using docker
`docker run --rm --name prowler --env AWS_ACCESS_KEY_ID --env AWS_SECRET_ACCESS_KEY --env AWS_SESSION_TOKEN toniblyx/prowler:3.0.1 -M html -B my-s3-bucket -o cis-benchmark`

image
image

Note: I also tried without -o and -M flag and same behaviour.

**Expected behavior**
Get a file with expected file naming: output/prowler-output-ACCOUNT_NUM-OUTPUT_DATE.format
In my case: html/prowler-output-xxxxx-20221226133115.html

jfagoagas:
### Context

Fix https://github.com/prowler-cloud/prowler/issues/1616

### Description

- Use custom output filename when set
- Set the `profile` field in the `Aws_Check_Output_CSV` model as `Optional[str]`

### License

### DIFF
--- a/prowler/lib/cli/parser.py
+++ b/prowler/lib/cli/parser.py
@@ -115,7 +115,6 @@ Detailed documentation at https://docs.prowler.cloud
             "-F",
             "--output-filename",
             nargs="?",
-            default=None,
             help="Custom output report name without the file extension, if not specified will use default output/prowler-output-ACCOUNT_NUM-OUTPUT_DATE.format",
         )
         common_outputs_parser.add_argument(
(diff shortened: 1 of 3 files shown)


MrCloudSec (approved):
👏🏼 👏🏼 👏🏼 👏🏼 👏🏼

PR #1634, opened 2023-01-02
fix(): password enabled issues in iam_user_mfa_enabled_console_access

### LINKED ISSUE #1605: [Bug]: Users with console password disabled are failing iam_user_mfa_enabled_console_access check
**What happened?**
I have a user with the console password disabled and this user is failing a check with "User prowler has Console Password enabled but MFA disabled".

**How to reproduce it**

1. Create an IAM user with the console password disabled and without MFA enabled.
2. Run prowler 
3. Filter the results to `iam` service and `iam_user_mfa_enabled_console_access` check.

**Expected behavior**
A user with the console password disabled should not show as failing for this check.

**Screenshots or Logs**
IAM settings for user in AWS console:
image

Result for this check for this user:
image

Please, complete the following information:
 - Resource: Docker container manually
 - OS: Ubuntu 20.04
 - Prowler Version [`./prowler --version`]: 3.0.1
 - Shell and version: bash 5.0.17(1)-release

**Additional context**
I see that in [iam_user_mfa_enabled_console_access.py](https://github.com/prowler-cloud/prowler/blob/f05cb2859e2f6555304b4dea825b81e81f8c7f81/prowler/providers/aws/services/iam/iam_user_mfa_enabled_console_access/iam_user_mfa_enabled_console_access.py#L14) the code is checking that `password_enabled` does not equal `not_supported` to determine if the console password is enabled. I think what it should be doing instead is checking is `user["password_enabled"] != "false"`. According to https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_getting-report.html, for the `password_enabled` field, "When the user has a password, this value is TRUE. Otherwise it is FALSE.The value for the AWS account root user is always not_supported."

n4ch04:
### Context

Comes from https://github.com/prowler-cloud/prowler/issues/1605.

### Description

The issue reported is that a regular user with no console enabled is generating a fail. It comes from the check logic, which only filters if the user is root, and then assume that all the other users has console enabled. That logic has been modified

### License

### DIFF
--- a/prowler/providers/aws/services/iam/iam_user_mfa_enabled_console_access/iam_user_mfa_enabled_console_access.py
+++ b/prowler/providers/aws/services/iam/iam_user_mfa_enabled_console_access/iam_user_mfa_enabled_console_access.py
@@ -11,13 +11,22 @@ class iam_user_mfa_enabled_console_access(Check):
             report.resource_id = user["user"]
             report.resource_arn = user["arn"]
             report.region = iam_client.region
+            # all the users but root (which by default does not support console password)
             if user["password_enabled"] != "not_supported":
-                if user["mfa_active"] == "false":
-                    report.status = "FAIL"
-                    report.status_extended = f"User {user['user']} has Console Password enabled but MFA disabled."
+                # check if the user has password enabled
+                if user["password_enabled"] == "true":
+                    if user["mfa_active"] == "false":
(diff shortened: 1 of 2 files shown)


MrCloudSec (approved):
👏🏼 👏🏼 👏🏼 👏🏼 👏🏼

PR #1637, opened 2023-01-02
fix(shub): apply -q to security hub

### LINKED ISSUE #1624: [Bug]: Prowler v3 integration to AWS Security Hub and usage of -q flag to send only FAILs
### Discussed in https://github.com/prowler-cloud/prowler/discussions/1623

<div type='discussions-op-text'>

<sup>Originally posted by **am-avios** December 28, 2022</sup>
Hello All

With previous version of prowler i was able to send only failed checks to security hub with flag -q, however it seems v3 sends both passed and failed checks to security hub even with -q flag. Is this intentional?  </div>

MrCloudSec:
### Description

Fixing https://github.com/prowler-cloud/prowler/issues/1624, apply -q option to Security Hub so only fails are sent.

### License

### DIFF
--- a/prowler/providers/aws/lib/security_hub/security_hub.py
+++ b/prowler/providers/aws/lib/security_hub/security_hub.py
@@ -15,33 +15,39 @@ from prowler.providers.aws.lib.audit_info.models import AWS_Audit_Info
 
 
 def send_to_security_hub(
-    region: str, finding_output: Check_Output_JSON_ASFF, session: session.Session
+    is_quiet: bool,
+    finding_status: str,
+    region: str,
+    finding_output: Check_Output_JSON_ASFF,
+    session: session.Session,
 ) -> int:
+    success_count = 0
     try:
-        logger.info("Sending findings to Security Hub.")
-        success_count = 0
-        # Check if security hub is enabled in current region
-        security_hub_client = session.client("securityhub", region_name=region)
-        security_hub_client.describe_hub()
+        # Check if -q option is set
+        if not is_quiet or (is_quiet and finding_status == "FAIL"):
+            logger.info("Sending findings to Security Hub.")
+            # Check if security hub is enabled in current region
+            security_hub_client = session.client("securityhub", region_name=region)
+            security_hub_client.describe_hub()
 
-        # Check if Prowler integration is enabled in Security Hub
-        if "prowler/prowler" not in str(
-            security_hub_client.list_enabled_products_for_import()
-        ):
-            logger.error(
-                f"Security Hub is enabled in {region} but Prowler integration does not accept findings. More info: https://github.com/prowler-cloud/prowler/#security-hub-integration"
-            )
+            # Check if Prowler integration is enabled in Security Hub
+            if "prowler/prowler" not in str(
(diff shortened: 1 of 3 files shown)


prowler/providers/aws/lib/security_hub/security_hub.py
```diff
@@ -15,34 +15,42 @@
 
 
 def send_to_security_hub(
```
jfagoagas, line 17: Would you mind to include a brief description using the docstring format?

```python
def send_to_security_hub(....):
"""
send_to_security_hub does the following and return X in the case of and Y in the case of...
"""
```

prowler/providers/aws/lib/security_hub/security_hub.py
```diff
@@ -15,34 +15,42 @@
...
-            logger.error(
-                f"Security Hub is enabled in {region} but Prowler integration does not accept findings. More info: https://github.com/prowler-cloud/prowler/#security-hub-integration"
-            )
+            # Check if Prowler integration is enabled in Security Hub
+            if "prowler/prowler" not in str(
+                security_hub_client.list_enabled_products_for_import()
+            ):
+                logger.error(
+                    f"Security Hub is enabled in {region} but Prowler integration does not accept findings. More info: https://github.com/prowler-cloud/prowler/#security-hub-integration"
+                )
```
jfagoagas, line 42: Why this if we are always sending findings to Security Hub?

prowler/providers/aws/lib/security_hub/security_hub.py
```diff
@@ -15,34 +15,42 @@
...
             )
-
+            if batch_import["FailedCount"] > 0:
+                failed_import = batch_import["FailedFindings"][0]
+                logger.error(
+                    f"Failed to send archived findings to AWS Security Hub -- {failed_import['ErrorCode']} -- {failed_import['ErrorMessage']}"
+                )
+            success_count = batch_import["SuccessCount"]
     except Exception as error:
         logger.error(f"{error.__class__.__name__} -- {error} in region {region}")
```
jfagoagas: Can you include the `traceback` to get the error line number? Thanks!

jfagoagas (changes requested):

prowler/providers/aws/lib/security_hub/security_hub.py
MrCloudSec: We need to check if Prowler Integration is enabled.

prowler/providers/aws/lib/security_hub/security_hub.py
MrCloudSec: Sure!

prowler/providers/aws/lib/security_hub/security_hub.py
jfagoagas: Yes but if it is not enabled we are sending the findings anyway, is that correct?

MrCloudSec: You are right! I corrected it.

jfagoagas (approved):

PR #1648, opened 2023-01-03
fix(codeartifact): set Namespace attribute as optional

MrCloudSec:
### Description

To avoid Key Errors with packages without namespaces, we need to set the attribute to optional.

### License


jfagoagas (dismissed):
Good catch @sergargar

jfagoagas (approved):

PR #1657, opened 2023-01-05
feat(): add ECS task revision number

Fennerr:
### Context

The current output doesn't specify which revision of the task definition potentially contains plaintext secrets

This results in multiple findings for the same task definition being created that look identical, as this check will flag on multiple different versions. It also requires you to manually go and check which revisions actually may contain plaintext secrets.

Adding the revision number to the report output will help clarify which exact definitions may contain plaintext secrets for the user to further investigate.

### Description

I have used the arn to get the revision number for the task definition 

A sample revision is: 'arn:aws:ecs:eu-west-1:ACCOUNTNUMBER:task-definition/task-definition-name:4'

`task_definition.arn.split(':')[-1]` will therefore result in the '4' being obtained, and added to the output

### License

### DIFF
--- a/prowler/providers/aws/services/ecs/ecs_service.py
+++ b/prowler/providers/aws/services/ecs/ecs_service.py
@@ -41,6 +41,7 @@ class ECS:
                             # we want the family name without the revision
                             name=sub(":.*", "", task_definition.split("/")[1]),
                             arn=task_definition,
+                            revision=task_definition.split(":")[-1],
                             region=regional_client.region,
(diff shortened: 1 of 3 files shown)


MrCloudSec:
Thank you for the PR @Fennerr, good idea!
We will review it shortly.

MrCloudSec:
@Fennerr I add the revision variable to the ECS service, let me know if it still works for you.

MrCloudSec (approved):

PR #1658, opened 2023-01-05
feat(): update recommendation of ecs_task_definitions_no_environment_secrets

Fennerr:
### Context

Added additional recommendation on how to remediate this issue, so that people don't just move the credentials to Secrets Manager or Parameter Store and think that the issue is addressed - whilst the credentials will still exist in the previous task definitions.

### Description

As seen in [this open issue](https://github.com/aws/containers-roadmap/issues/685), the AWS team have finally recognized that they should implement a feature to allow you to delete previous task definition revisions.

If a user has previously stored plaintext secrets in their task definition, and have now decided to address it by moving the credentials to Secrets Manager or Paramater Store, the plaintext secrets will still be accessible in the previous revisions of the task definition.

I therefore wanted to clarify that it should be recommended that these credentials are also rotated, until AWS implements this feature to let you delete previous task definition revisions that contain plaintext secrets.

### License

### DIFF
--- a/prowler/providers/aws/services/ecs/ecs_task_definitions_no_environment_secrets/ecs_task_definitions_no_environment_secrets.metadata.json
+++ b/prowler/providers/aws/services/ecs/ecs_task_definitions_no_environment_secrets/ecs_task_definitions_no_environment_secrets.metadata.json
@@ -23,7 +23,7 @@
       "Terraform": ""
     },
     "Recommendation": {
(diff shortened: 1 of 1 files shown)


MrCloudSec:
@Fennerr thank you for the information!
Just to understand it correctly, what about deregistering task definitions? That will delete the revision that you want, correct me if I'm wrong.

MrCloudSec:
@Fennerr I see what you mean now, you can deregister the tasks but they remain as inactive. Good catch!

MrCloudSec (approved):

PR #1665, opened 2023-01-06
feat(ecs_task_definitions_no_environment_secrets): Update resource_id

Fennerr:
Following on from yesterday's changes to include the revision number in the status_extended message, I would like to update the resource ID to also include the task definition revision number. Im not sure if the resource ID is allowed to have spaces in it

### License


n4ch04:
Hi @Fennerr we are going to check this and let you know

MrCloudSec (approved):
@Fennerr We've updated the check to include the family revision as AWS states it like `family:revision` and update the check's test.

Thanks for your contribution!

PR #1671, opened 2023-01-09
fix(db connector): fixing db connector validations

n4ch04:
### Context

The db connector in prowler v2 needs to limit the results retrieved from the testing query

### Description

The db connector in prowler v2 needs to limit the results retrieved from the testing query

### License


jfagoagas (approved):
🥷

PR #1672, opened 2023-01-09
fix(shub): associate resource_arn as resourceId in Security Hub

### LINKED ISSUE #1661: [Bug]: ASFF Resources['Id'] using resource_id instead of resource_arn
Hi Prowler Team! 

**What happened?**

After reviewing the integration between Security Hub and Prowler, I found that ASFF output for field Resources['Id'] it's being filled by resource_id instead of resource_arn as it should be. As I see it, this is a big deal. If you need to correlate findings from different scanners/sources based on affected resource, you can't because the Resources['Id'] being reported by Prowler is not the ARN as it is supposed to be (when correlating findings is not the same `i-0123456789` than `arn:ec2:..../i-0123456789`)

Here is what I found:

- ASFF Reosource['Id'] field must be the ARN of the affected resource ([ASFF](https://docs.aws.amazon.com/securityhub/latest/userguide/asff-resources.html))
- Prowler is using the resource_id instead of the resource_arn ([Id](https://github.com/prowler-cloud/prowler/blob/master/prowler/lib/outputs/json.py#L35))
- Looks like AWS is not validating this field, as it's working this way... 
- I found that Prowler is not generating the ResourceArn for all the scanned resource types. This could explain why it is sending the resource_id (available for every resource).
- I already started [here](https://github.com/prowler-cloud/prowler/pull/1649) adding the missing ResourceArn. Having the ResourceArn for each finding is key for any good integration, as it is the unique identifier for an AWS resource (not the resource_id).
- What I wanted to propose is modifying the ASFF generator to start using resource_arn only when it's available, and when not, use the resource_id like we are doing now, so something like this [here:](https://github.com/prowler-cloud/prowler/blob/master/prowler/lib/outputs/json.py#L15):

```
    if finding.resource_id == "":
        finding.resource_id = "NONE_PROVIDED"
    if finding.resource_arn == "":
        finding.resource_arn = finding.resource_id
    ...
    ...
    finding_output.Resources = [
        Resource(
            Id=finding.resource_arn,
            Type=finding.check_metadata.ResourceType,
            Partition=audit_info.audited_partition,
            Region=finding.region,
        )
    ]
```

- Last but not least important: the Finding Id for each finding is generated based on the [resource_id](https://github.com/prowler-cloud/prowler/blob/master/prowler/lib/outputs/json.py#L19), so we can keep that to avoid making this a breaking change in terms of integration as far as I understand based on AWS Documentation. Findings are unique based on this Id, so when this change, it's deployed. The id will continue being the same; the only field that will be updated is the Resources['Id']

**Expected behavior**

Having ResourceArn for each affected resource and generating the Resoucers['Id'] field for ASFF output based on that field.

`Id=finding.resource_arn`

MrCloudSec:
### Description

 Associate resource_arn as the resourceId in Security Hub regarding https://github.com/prowler-cloud/prowler/issues/1661

### License

### DIFF
--- a/prowler/lib/outputs/json.py
+++ b/prowler/lib/outputs/json.py
@@ -14,12 +14,14 @@ from prowler.lib.utils.utils import hash_sha512, open_file
 
 def fill_json_asff(finding_output, audit_info, finding):
     # Check if there are no resources in the finding
-    if finding.resource_id == "":
-        finding.resource_id = "NONE_PROVIDED"
+    if finding.resource_arn == "":
+        if finding.resource_id == "":
+            finding.resource_id = "NONE_PROVIDED"
+        finding.resource_arn = finding.resource_id
     finding_output.Id = f"prowler-{finding.check_metadata.CheckID}-{audit_info.audited_account}-{finding.region}-{hash_sha512(finding.resource_id)}"
     finding_output.ProductArn = f"arn:{audit_info.audited_partition}:securityhub:{finding.region}::product/prowler/prowler"
     finding_output.ProductFields = ProductFields(
-        ProviderVersion=prowler_version, ProwlerResourceName=finding.resource_id
+        ProviderVersion=prowler_version, ProwlerResourceName=finding.resource_arn
     )
     finding_output.GeneratorId = "prowler-" + finding.check_metadata.CheckID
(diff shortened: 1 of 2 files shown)


jfagoagas (approved):
🔝

PR #1708, opened 2023-01-16
feat(only_logs): New logging flag to only show execution logs

jfagoagas:
### Description

New flag `--only-logs` to output nothing but the execution logs. This flag sets the `--no-banner` to true.

Also, logging level can be set with the `LOG_LEVEL` environment variable.

Also the default `ERROR` value set in the  `set_logging_config` at `lib/logger/logger.py` has been removed since it was not used.

### License


MrCloudSec (dismissed):
Awesome 👏🏼

n4ch04 (approved):

PR #1714, opened 2023-01-16
fix(extra7131): exclude DocumentDB since AutoMinorVersionUpgrade is only for relational databases

kagahd:
### Context

Prowler `check_extra7131` returns false positives for DocumentDB instances.

### Description

As written in AWS's [documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-docdb-dbinstance.html#cfn-docdb-dbinstance-autominorversionupgrade), the parameter `AutoMinorVersionUpgrade` does not apply to Amazon DocumentDB.  Neither [CDK](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_docdb.DatabaseCluster.html) or AWS Web-GUI allow to set `AutoMinorVersionUpgrade` for DocumentDB's.
The AWS CLI command that `check_extra7131` is using returns `"AutoMinorVersionUpgrade": false` for DocumentDB instances which let's the check fail.
Therefore, prowler `check_extra7131` needs to exclude DocumentDB instances to avoid false positives. 

Btw. the same has been done correctly already in:

- check_extra7113
- check_extra7132
- check_extra7133

### License

### DIFF
--- a/checks/check_extra7131
+++ b/checks/check_extra7131
@@ -18,15 +18,15 @@ CHECK_SEVERITY_extra7131="Low"
 CHECK_ASFF_RESOURCE_TYPE_extra7131="AwsRdsDbInstance"
 CHECK_ALTERNATE_check7131="extra7131"
 CHECK_SERVICENAME_extra7131="rds"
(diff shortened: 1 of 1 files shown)


MrCloudSec (approved):
Good catch @kagahd , thank you!

PR #1723, opened 2023-01-17
fix(arguments): improve quiet option

MrCloudSec:
### Description

- Remove verbose from quiet option.
- Add only fails to outputs.

### License


jfagoagas (changes requested):
I'd remove the PASS findings from the summary table if the `--quiet` is supplied.

jfagoagas (changes requested):
Can you include that new behaviour in the documentation and a quick hint in the parser lib?

Also include that the summary table shows also the PASS findings.

jfagoagas (approved):
🔝

PR #1734, opened 2023-01-18
fix(allowlist): remove re.escape

MrCloudSec:
### Description

Remove re.escape to allow Regex expressions with special characters such as `.*`.

### License

### DIFF
--- a/tests/providers/aws/lib/allowlist/allowlist_test.py
+++ b/tests/providers/aws/lib/allowlist/allowlist_test.py
@@ -109,7 +109,45 @@ class Test_Allowlist:
                     "Checks": {
                         "check_test": {
                             "Regions": ["us-east-1", "eu-west-1"],
-                            "Resources": ["prowler", "^test"],
+                            "Resources": ["prowler", "^test", "prowler-pro"],
+                        }
+                    }
+                }
+            }
+        }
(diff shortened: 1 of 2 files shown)


tests/providers/aws/lib/allowlist/allowlist_test.py
```diff
@@ -109,7 +109,45 @@ def test_is_allowlisted(self):
...
+    def test_is_allowlisted_wildcard(self):
+
+        # Allowlist example
+        allowlist = {
+            "Accounts": {
+                "*": {
+                    "Checks": {
+                        "check_test": {
+                            "Regions": ["us-east-1", "eu-west-1"],
+                            "Resources": [".*"],
```
jfagoagas, line 150: I'd change this to an `*` and the internally include the `.` if the only argument is an `*`. What do you think?

jfagoagas (changes requested):

tests/providers/aws/lib/allowlist/allowlist_test.py
MrCloudSec: Done, great idea!

jfagoagas (approved):
Awesome! Good catch!

PR #1790, opened 2023-01-28
fix(metadata): use docs.aws.amazon.com like other aws checks, not docs.amazonaws.cn

ifduyue:
### Context

Use docs.aws.amazon.com URLs (like in other AWS checks) instead of docs.amazonaws.cn URLs

### Description

.cn is country-code top-level domain, besides .cn doc url is only used in rds_instance_no_public_access check.

### License


MrCloudSec (approved):
Thank you for your contribution @ifduyue, good catch!

PR #1792, opened 2023-01-28
fix(allowlist): validate allowlist for any database format (file, dynamo, s3, etc)

pplu:
### Context

While trying out allowlist functionality, I observed that the (very useful) allowlist validation was only happening when the allowlist is a file.

### Description

This changes puts the allowlist validation at the level of all types of allowlist types

### License


MrCloudSec (approved):
Thank you for your contribution @pplu, good catch!

PR #1793, opened 2023-01-28
feat(allowlist): AWS Lambda function support

pplu:
### Context

While studying how to implement Prowler for our use case, I found some challenges with the current Allowlist options.

Some examples:
 - For the `ec2_instance_public_ip` check, we have instances in AutoScalingGroups that cannot be listed in Resources. Allowlisting all instances with `Resources: [ '' ]` in the Allowlist will make potentially unallowed instances to be Allowed
 - We want to allow some distrubtions for the `cloudfront_distributions_using_waf` check, but due to the dynamic nature of resources in our accounts (stacks get created and destroyed with CloudFormation) we cannot whitelist a set of CloudFront distributions that are allowed to not have WAF.
 - Some of our stacks create a "dormant" Security Group which is seldomly used, but very convenient to have pre-configured. To whitelist those resources in `ec2_securitygroup_not_used`  we cannot know beforehand the security group Ids that are going to be allowed.

### Description

I've added an option for the allowlist parameter to be an AWS Lambda ARN which will be invoked to obtain the Allowlist. The Lambda can then query AWS APIs and generate Allowlist checks for resources following the organizations needs.

I have added some documentation with example code also.

I would like to know if this feature is good for being accepted before continuing to implement prowler depending on this feature (I would prefer to depend on a non-forked version going forward)

Note that this PR builds on #1792, which can be merged separately if this feature is not acceptable.

### License

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license

### DIFF
--- a/docs/tutorials/allowlist.md
+++ b/docs/tutorials/allowlist.md
@@ -74,3 +74,36 @@ prowler aws -w arn:aws:dynamodb:<region_name>:<account_id>:table/<table_name>
 <img src="../img/allowlist-row.png"/>
 
 > Make sure that the used AWS credentials have `dynamodb:PartiQLSelect` permissions in the table.
+
+### AWS Lambda ARN
+
+Passing an AWS Lambda ARN:
+
+```
+prowler aws -w arn:aws:lambda:REGION:ACCOUNT_ID:function:FUNCTION_NAME
+```
+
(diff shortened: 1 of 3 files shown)


MrCloudSec (approved):
This is an awesome feature, thank you for your contribution @pplu !!

PR #1833, opened 2023-02-03
fix(awslambda_function_no_secrets_in_code): Retrieve Code if set

jfagoagas:
### Context

Actually we are retrieving the AWS Lambda Code even if the `awslambda_function_no_secrets_in_code` check is excluded.

### Description

Only call AWS Lambda Service `__get_function__` if the `awslambda_function_no_secrets_in_code` is set.

### License


MrCloudSec (dismissed):
🔝

MrCloudSec (approved):

PR #1837, opened 2023-02-03
fix(lambda-runtime): Init value must be empty string

jfagoagas:
### Description

Init value for Lambda Runtime must be an empty string.

### License


MrCloudSec (approved):
🔝

PR #1852, opened 2023-02-07
fix(iam): change prowler additional policy json due errors in creation

theist:
### Context

I found a problem adding this policy to a dedicated user in IAM

### Description

Hi!

I was test driving prowler and added this policy to its dedicated IAM user. When I create the policy in the editor It complained about `s3:GetPublicAccessBlock` not existing. I changed it for the one I think is required, but not sure about it.

https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetPublicAccessBlock.html

I think the problem is that while the API call is `GetPublicAccessBlock` the permission required according to the doc is `s3:GetBucketPublicAccessBlock`

### License

### DIFF
--- a/permissions/prowler-additions-policy.json
+++ b/permissions/prowler-additions-policy.json
@@ -18,7 +18,7 @@
         "lambda:GetFunction*",
         "macie2:GetMacieSession",
         "s3:GetAccountPublicAccessBlock",
-        "s3:GetPublicAccessBlock",
+        "s3:GetBucketPublicAccessBlock",
         "shield:DescribeProtection",
         "shield:GetSubscriptionState",
         "ssm:GetDocument",
--- a/permissions/prowler-additions-policy.json
+++ b/permissions/prowler-additions-policy.json
@@ -18,7 +18,6 @@
         "lambda:GetFunction*",
         "macie2:GetMacieSession",
         "s3:GetAccountPublicAccessBlock",
-        "s3:GetBucketPublicAccessBlock",
         "shield:DescribeProtection",
         "shield:GetSubscriptionState",
         "ssm:GetDocument",


MrCloudSec:
Hi @theist, thank you for your contribution, but this policy was made to extend the permissions of the `SecurityAudit` and `ViewOnlyAccess` AWS Managed Policies. To use Prowler with a dedicated IAM user please, add the two mentioned policies which already have the `"s3:GetBucket*"` permissions.

theist:
Thanks @sergargar 

I was following documentation in https://docs.prowler.cloud/en/latest/getting-started/requirements/

And it states:

> Those credentials must be associated to a user or role with proper permissions to do all checks. To make sure, add the following AWS managed policies to the user or role being used:
> 
> `arn:aws:iam::aws:policy/SecurityAudit`
> `arn:aws:iam::aws:policy/job-function/ViewOnlyAccess`
> 
> > Moreover, some read-only additional permissions are needed for several checks, make sure you attach also the custom policy `prowler-additions-policy.json` to the role you are using.
> >
> > If you want Prowler to send findings to AWS Security Hub, make sure you also attach the custom policy prowler-security-hub.json.

So I understood that I have to add the four policies to `prowler` user (thru a group in my case)

`prowler-additions-policy.json`  includes an error for IAM and the editor complains

image

It is true that the SecurityAudit policy does include the `s3:GetBucket*` which covers the policy I corrected, but still, the line represents an error in the UI console (or tools like terraform that try to set that json) so either it should be corrected or removed because it causes an error to people following the documentation. 

My contribution corrected it, but it is true that is better to remove it, as it is already covered by the other recommended policies.  I leave it to you, but keeping the file with unknown permission can mislead users.

Cheers.

MrCloudSec:
Thank you for the info @theist !
You're right, the `s3:GetPublicAccessBlock` permission is incorrect. I reopen the PR so we can remove this permission from the policy.

permissions/prowler-additions-policy.json
```diff
@@ -18,7 +18,7 @@
         "lambda:GetFunction*",
         "macie2:GetMacieSession",
         "s3:GetAccountPublicAccessBlock",
-        "s3:GetPublicAccessBlock",
+        "s3:GetBucketPublicAccessBlock",
```
MrCloudSec:
```suggestion
```

MrCloudSec (changes requested):

MrCloudSec (approved):

PR #1854, opened 2023-02-07
fix(output_bucket): Use full path for -o option with output to S3 bucket

### LINKED ISSUE #1850: [Bug]: Unable to push report into complex S3 object key
**What happened?**
I tried to upload Prowler reports into S3 using `-B` option together with `-o` option specifying full path to the report file.
When it comes to file upload to S3, prowler takes only the last directory in the `-o` option value and uses it as a key for the S3 object.
I'd want to store reports in S3 in the same way AWS do for logs (to have a directory structure in a next way: `<account_id>/<year>/<month>/<day>/csv/report_file.csv`.

For now, even if I specify `-o <account_id>/<year>/<month>/<day>` it will store reports in S3 within the `<day>/csv/report_ile.csv>`, so it just drops all previous directories in the path and taking only the last one. I suppose that's because of this line: https://github.com/prowler-cloud/prowler/blob/master/prowler/lib/outputs/outputs.py#L200. This code has been intentionally done for some reason so it might not be an actual bug (the app is doing exactly what it's being told to do) but it's kinda strange to have this in place.

**How to reproduce it**
1. What command are you running? `mkdir -p ${TARGET_AWS_ACCOUNT_ID}/$(date '+%Y')/$(date '+%m')/$(date '+%d'); prowler aws -M csv html -D ${S3_BUCKET_NAME} --log-level ERROR -R arn:aws-cn:iam::${TARGET_AWS_ACCOUNT_ID}:role/${TARGET_AWS_ACCOUNT_ROLE_NAME} -o ${TARGET_AWS_ACCOUNT_ID}/$(date '+%Y')/$(date '+%m')/$(date '+%d')`
2. Cloud provider you are launching: `aws`
3. Environment you have like single account, multi-account, organizations, multi or single subsctiption, etc.: `multi-account within the org`
4. See error: No errors

**Expected behavior**
I'd want to have S3 objects to be uploaded in the same directory structure as they are written locally using `-o` option.

**Screenshots or Logs**
Also, you can add logs (anonymize them first!). Here a command that may help to share a log
`prowler <your arguments> --log-level DEBUG --log-file $(date +%F)_debug.log` then attach here the log file.

Please, complete the following information:
 - Resource: (e.g. EC2 instance, Fargate task, Docker container manually, EKS, Cloud9, CodeBuild, workstation, etc.): workstation, fargate task, docker container
 - OS: [e.g. Amazon Linux 2, Mac, Alpine, Windows, etc. ] Mac
 - Prowler Version [`prowler --version`]: Prowler 3.1.2
 - Python version [`python --version`]: Python 3.9.12
 - Pip version [`pip --version`]:
 - Installation method (Are you running it from pip package or cloning the github repo?): both
 - Others:

**Additional context**

MrCloudSec:
### Description

Regarding https://github.com/prowler-cloud/prowler/issues/1850, when it comes to file upload to S3, prowler takes only the last directory in the -o option value and uses it as a key for the S3 object.

### License

### DIFF
--- a/tests/lib/outputs/outputs_test.py
+++ b/tests/lib/outputs/outputs_test.py
@@ -1,5 +1,4 @@
 import os
-import pathlib
 from os import path, remove
 from unittest import mock
 
@@ -337,9 +336,7 @@ class Test_Outputs:
         client.create_bucket(Bucket=bucket_name)
         # Create mock csv output file
         fixtures_dir = "fixtures"
-        output_directory = (
-            f"{pathlib.Path().absolute()}/tests/lib/outputs/{fixtures_dir}"
-        )
+        output_directory = f"tests/lib/outputs/{fixtures_dir}"
         output_mode = "csv"
         filename = f"prowler-output-{input_audit_info.audited_account}"
         # Send mock csv file to mock S3 Bucket
@@ -354,7 +351,12 @@ class Test_Outputs:
         assert (
             client.get_object(
                 Bucket=bucket_name,
-                Key=fixtures_dir + "/" + output_mode + "/" + filename + csv_file_suffix,
+                Key=output_directory
+                + "/"
+                + output_mode
+                + "/"
+                + filename
+                + csv_file_suffix,
             )["ContentType"]
             == "binary/octet-stream"
         )
--- a/tests/lib/outputs/outputs_test.py
+++ b/tests/lib/outputs/outputs_test.py
@@ -1,5 +1,5 @@
 import os
-from os import path, remove
+from os import getcwd, path, remove
 from unittest import mock
 
 import boto3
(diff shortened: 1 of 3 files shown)


tests/lib/outputs/outputs_test.py
```diff
@@ -337,9 +336,7 @@ def test_send_to_s3_bucket(self):
         client.create_bucket(Bucket=bucket_name)
         # Create mock csv output file
         fixtures_dir = "fixtures"
-        output_directory = (
-            f"{pathlib.Path().absolute()}/tests/lib/outputs/{fixtures_dir}"
-        )
```
jfagoagas, line 342: it would be interesting to test both cases to check if we are covering the old and the new paths. What do you think?

jfagoagas (changes requested):

tests/lib/outputs/outputs_test.py
MrCloudSec: Sure, done!

tests/lib/outputs/outputs_test.py
```diff
@@ -336,10 +335,9 @@ def test_send_to_s3_bucket(self):
         client = boto3.client("s3")
         client.create_bucket(Bucket=bucket_name)
         # Create mock csv output file
-        fixtures_dir = "fixtures"
-        output_directory = (
-            f"{pathlib.Path().absolute()}/tests/lib/outputs/{fixtures_dir}"
-        )
+        fixtures_dir = "tests/lib/outputs/fixtures"
+        output_directory = getcwd() + "/" + fixtures_dir
+        print(output_directory)
```
jfagoagas:
```suggestion
```

jfagoagas (approved):

PR #1859, opened 2023-02-07
fix(checks): added validation for non-existing VPC endpoint policy

daftkid:
### Context

I have just noticed the following error for one of the VPC endpoint-related checks:
```
2023-02-07 18:36:25,881 [File: check.py:307] 	[Module: check]	 ERROR: vpc_endpoint_connections_trust_boundaries -- TypeError[13]: 'NoneType' object is not subscriptable
```
I added some debug info (please see `Policy Document is equal to None, skipping the test` in the output below) and found out that some VPC endpoints do not have any policies attached to them, e.g. (some data is masked):

```
VpcEndpoint(id='vpce-034bxxxxxxx', vpc_id='vpc-xxxxxxxx', state='available', policy_document=None, owner_id='xxxxxxxxxx', region='cn-northwest-1')
Policy Document is equal to None, skipping the test
VpcEndpoint(id='vpce-0a12exxxxxx', vpc_id='vpc-xxxxxxxx', state='available', policy_document=None, owner_id='xxxxxxxxxx', region='cn-northwest-1')
Policy Document is equal to None, skipping the test
VpcEndpoint(id='vpce-0290xxxxxxxx', vpc_id='vpc-xxxxxxxx', state='available', policy_document=None, owner_id='xxxxxxxxxx', region='cn-northwest-1')
Policy Document is equal to None, skipping the test
VpcEndpoint(id='vpce-00cxxxxxxx', vpc_id='vpc-xxxxxxxx', state='available', policy_document={'Version': '2008-10-17', 'Statement': [{'Effect': 'Allow', 'Principal': '*', 'Action': '*', 'Resource': '*'}]}, owner_id='xxxxxxxxxx', region='cn-northwest-1')
```
As you can see most of the endpoints have `policy_document` set to `None` and there are no conditions for filtering out these resources.

After adding this condition, I'm able to run `./prowler.py aws -M html -c vpc_endpoint_connections_trust_boundaries --log-level ERROR` and check is able to pass for all endpoints that have policies attached and skips all others which do not have them not failing the full check.

P.S. I do not know if it's true for Non-Chinese regions and unfortunately I'm not able to check it now (I mean the structure of the object that is returned by AWS API for VPC endpoint), but I believe other regions will have the same problem so it should fix all of the regions at once.

### Description

Added condition to check if VPC Endpoint resource has any policy attached to it.

### License

### DIFF
--- a/prowler/providers/aws/services/vpc/vpc_endpoint_connections_trust_boundaries/vpc_endpoint_connections_trust_boundaries.py
+++ b/prowler/providers/aws/services/vpc/vpc_endpoint_connections_trust_boundaries/vpc_endpoint_connections_trust_boundaries.py
@@ -10,7 +10,11 @@ class vpc_endpoint_connections_trust_boundaries(Check):
         trusted_account_ids = get_config_var("trusted_account_ids")
         for endpoint in vpc_client.vpc_endpoints:
             # Check VPC endpoint policy
-            for statement in endpoint.policy_document["Statement"]:
+            policy_statements = []
+            if endpoint.policy_document is not None:
+                policy_statements = endpoint.policy_document["Statement"]
+
+            for statement in policy_statements:
                 if "*" == statement["Principal"]:
                     report = Check_Report_AWS(self.metadata())
                     report.region = endpoint.region
--- a/prowler/providers/aws/services/vpc/vpc_endpoint_connections_trust_boundaries/vpc_endpoint_connections_trust_boundaries.py
+++ b/prowler/providers/aws/services/vpc/vpc_endpoint_connections_trust_boundaries/vpc_endpoint_connections_trust_boundaries.py
@@ -10,45 +10,42 @@ class vpc_endpoint_connections_trust_boundaries(Check):
         trusted_account_ids = get_config_var("trusted_account_ids")
         for endpoint in vpc_client.vpc_endpoints:
             # Check VPC endpoint policy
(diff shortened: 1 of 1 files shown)


MrCloudSec:
Hi @daftkid ! 
Thank you for your contribution, I have made some changes in order to follow the pattern that we have in the checks, let me know if that works for you.
I think it is specific for Chinese regions, but, if a VPC Endpoint has no policy, the endpoint is not accessible?

daftkid:
Hi @sergargar these VPC endpoints are of type `GatewayLoadBalancer` and they do not have a policy, access is managed via `AllowPrincipals` for the EndpointService resource. Anyway, I'll try to check non-Chinese region later today and will update you

daftkid:
@sergargar Please refer to this awscli doc for the VPC endpoint - https://docs.aws.amazon.com/cli/latest/reference/ec2/create-vpc-endpoint.html (`Example 3: To create a Gateway Load Balancer endpoint` section), as you can see `PolicyDocument` field is absent in the output.

daftkid:
@sergargar I've checked your commit, and it works as expected! Thanks. Do you have a contribution guide somewhere? I'd like to know what are your rules, styles etc. 😀

MrCloudSec (approved):
Awesome @daftkid ! We're working on a Development Guide, we'll let you know when it's available. Thanks for the contribution 😄

PR #1872, opened 2023-02-09
chore(compliance): add manual checks to compliance CSV

MrCloudSec:
### Description

Add manual checks to compliances' CSV.

### License

### DIFF
--- a/prowler/lib/check/compliance.py
+++ b/prowler/lib/check/compliance.py
@@ -63,7 +63,7 @@ def update_checks_metadata_with_compliance(
                     check_compliance.append(compliance)
             # Create metadata for Manual Control
             manual_check_metadata = """{
-                "Provider" : "AWS",
+                "Provider" : "aws",
                 "CheckID" :  "manual_check",
                 "CheckTitle" : "Manual Check",
                 "CheckType" : [],
--- a/prowler/lib/check/compliance.py
+++ b/prowler/lib/check/compliance.py
@@ -4,6 +4,7 @@ from prowler.lib.check.compliance_models import (
     Compliance_Base_Model,
     Compliance_Requirement,
 )
+from prowler.lib.check.models import Check_Report_AWS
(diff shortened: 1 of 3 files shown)


prowler/lib/check/compliance.py
```diff
@@ -41,6 +43,64 @@ def update_checks_metadata_with_compliance(
                         check_compliance.append(compliance)
             # Save it into the check's metadata
             bulk_checks_metadata[check].Compliance = check_compliance
+
+        # Add requirements of Manual Controls
+        for framework in bulk_compliance_frameworks.values():
```
jfagoagas, line 48: why are you doing the same for again here? 

You can use the one above.

jfagoagas (changes requested):

prowler/lib/check/compliance.py
MrCloudSec: Because the one above is iterating for each check and it has to be only done once.

prowler/lib/check/compliance.py
```diff
@@ -41,6 +43,64 @@ def update_checks_metadata_with_compliance(
...
+                    )
+                    # Include the compliance framework for the check
+                    check_compliance.append(compliance)
+            # Create metadata for Manual Control
+            manual_check_metadata = """{
+                "Provider" : "AWS",
+                "CheckID" :  "manual_check",
+                "CheckTitle" : "Manual Check",
+                "CheckType" : [],
+                "ServiceName" : "",
```
jfagoagas:
```suggestion
            manual_check_metadata = f"""{
                "Provider" : {framework.Provider},
                "CheckID" :  "manual_check",
                "CheckTitle" : "Manual Check",
                "CheckType" : [],
                "ServiceName" : "",
```

jfagoagas (dismissed):

jfagoagas (approved):

PR #1899, opened 2023-02-14
fix(compliance): Set Version as optional and fix list

jfagoagas:
### Context

Version was not set as optional and also the `--list-compliance-requirements` is broken.

### Description

- Set Version as optional
- Include Framework and Provider validator
- Remove code to fix the `--list-compliance-requirements`

### License

### DIFF
--- a/prowler/__main__.py
+++ b/prowler/__main__.py
@@ -216,14 +216,15 @@ def prowler():
         )
 
         if compliance_framework and findings:
-            # Display compliance table
-            display_compliance_table(
-                findings,
-                bulk_checks_metadata,
-                compliance_framework,
-                audit_output_options.output_filename,
-                audit_output_options.output_directory,
-            )
+            for compliance in compliance_framework:
+                # Display compliance table
+                display_compliance_table(
+                    findings,
+                    bulk_checks_metadata,
+                    compliance,
+                    audit_output_options.output_filename,
+                    audit_output_options.output_directory,
+                )
 
     # If there are failed findings exit code 3, except if -z is input
     if not args.ignore_exit_code_3 and stats["total_fail"] > 0:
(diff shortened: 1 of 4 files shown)


toniblyx (changes requested):
Despite listing seems to work fine now. However when executing more than one framework like in:
```
./prowler.py aws --compliance aws_audit_manager_control_tower_guardrails_aws ffiec_aws gdpr_aws aws_foundational_security_best_practices_aws
```
Only the first one is shown, in the example above only `Detailed results of AWS_AUDIT_MANAGER_CONTROL_TOWER_GUARDRAILS_AWS are in:` however all files are generated properly. So the results  of file locations are not properly shown.

jfagoagas:
> Despite listing seems to work fine now. However when executing more than one framework like in:
> 
> ```
> ./prowler.py aws --compliance aws_audit_manager_control_tower_guardrails_aws ffiec_aws gdpr_aws aws_foundational_security_best_practices_aws
> ```
> 
> Only the first one is shown, in the example above only `Detailed results of AWS_AUDIT_MANAGER_CONTROL_TOWER_GUARDRAILS_AWS are in:` however all files are generated properly. So the results of file locations are not properly shown.


@sergargar could you please take a look at this lines? https://github.com/prowler-cloud/prowler/blob/master/prowler/lib/outputs/compliance.py#L361-L369

We have to loop through every framework to print its output location.

MrCloudSec:
> Despite listing seems to work fine now. However when executing more than one framework like in:
> 
> ```
> ./prowler.py aws --compliance aws_audit_manager_control_tower_guardrails_aws ffiec_aws gdpr_aws aws_foundational_security_best_practices_aws
> ```
> 
> Only the first one is shown, in the example above only `Detailed results of AWS_AUDIT_MANAGER_CONTROL_TOWER_GUARDRAILS_AWS are in:` however all files are generated properly. So the results of file locations are not properly shown.

Good catch! This is fix now.

toniblyx (approved):

PR #1903, opened 2023-02-14
fix(iam-credentials-expiration): IAM password policy expires passwords fix

congon4tor:
### Context

Currently if a password policy is set to expire after 90 days it will fail this check even though the recommendation Says to set it to 90 or less. `Ensure Password expiration period (in days): is set to 90 or less.` With this change 90 days will pass the check

### Description

Made the check `<= 90` instead of `< 90`

### License


jfagoagas:
Hi @congon4tor, could you please check the other credentials expiration check for IAM like the one for the 45 days expiration and check if it has the same issue? 

Thanks!

congon4tor:
@jfagoagas All the other checks appear to be correct, I checked `iam_disable_30_days_credentials`, `iam_disable_45_days_credentials`, `am_disable_90_days_credentials` and `iam_rotate_access_key_90_days`

jfagoagas:
> @jfagoagas All the other checks appear to be correct, I checked `iam_disable_30_days_credentials`, `iam_disable_45_days_credentials`, `am_disable_90_days_credentials` and `iam_rotate_access_key_90_days`

Great! Thanks you!

jfagoagas (approved):
Thanks for your contribution @congon4tor !

I've included a new test case just to check the new behaviour.

PR #1911, opened 2023-02-15
chore(security hub): add --skip-sh-update

### LINKED ISSUE #1867: Skip SecurityHub updates when scanning failures occur
Based on my understanding of the code in https://github.com/prowler-cloud/prowler/blob/master/prowler/providers/aws/lib/security_hub/security_hub.py, Prowler uses the following logic to archive SecurityHub findings:

1. Create a list of finding IDs for findings generated by the scan performed (i.e. current findings) from local ASFF data.
2. Call `securityhub:GetFindings` to retrieve all ACTIVE Prowler findings for the AWS account and region that was scanned.
3. Archive any ACTIVE findings whose IDs are not in the list of current findings.

The issue with this logic is that if a scanning error occurs, valid findings can be missing from the generated list of "current" findings. These missing findings would then result in SecurityHub findings being archived even though the underlying issues associated with these findings are still present, essentially creating a false negative in SecurityHub.

I see two potential solutions to this issue:

1. (Easier) Add a command line option such as `--skip-sh-update-on-error` which will cause Prowler to skip the SecurityHub update process (or at least the call to `resolve_security_hub_previous_findings()`) when any scanning errors occur.
2. (Harder) Add logic to track which checks or services failed, and have Prowler only avoid resolving the impacted finding types.

**Additional context**
This is primarily an issue in environments where Prowler is used for continuous scanning, since stale data with potential false positives is preferable to data containing false negatives.

MrCloudSec:
### Description

Regarding https://github.com/prowler-cloud/prowler/issues/1867, allow users to skip updates of previous Security Hub findings of Prowler.

### License

### DIFF
--- a/docs/tutorials/aws/securityhub.md
+++ b/docs/tutorials/aws/securityhub.md
@@ -36,3 +36,12 @@ or for only one filtered region like eu-west-1:
 Once you run findings for first time you will be able to see Prowler findings in Findings section:
 
 ![Screenshot 2020-10-29 at 10 29 05 PM](https://user-images.githubusercontent.com/3985464/97634676-66c9f600-1a36-11eb-9341-70feb06f6331.png)
+
+## Skip sending updates of findings to Security Hub
+
+By default, Prowler archives all its findings in Security Hub that have not appeared in the last scan.
+You can skip this logic by using the option `--skip-sh-update` so Prowler will not archive older findings:
+
+```sh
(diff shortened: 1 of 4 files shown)


jfagoagas (changes requested):
Good one @sergargar ! 

Could you please include the documentation of this new argument?

Thanks!

jfagoagas (approved):
Awesome!

PR #1922, opened 2023-02-16
fix(ecs_task_definitions_no_environment_secrets): dump_env_vars is reintialised

Fennerr:
Fixed a bug on how the environment variables were being stored in dump_env_vars before the they written to disk

### Context

Fixed a bug in the ecs_task_definitions_no_environment_secrets check

### Description

Confirmed through debugging that the dump_env_vars dict was being reintialised each time it went through the for loop, so the check would only run detect_secrets against the last env_var

### License


MrCloudSec (approved):
Good catch @Fennerr, thank you!!

PR #1929, opened 2023-02-17
fix(iam_rotate_access_key_90_days): check only active access keys

Fennerr:
Updated the check to only report on access keys that haven't been rotated and are active

### Context

I was confused as to why the check for if the root user has access keys had passed, but the check for rotating the access key had failed for the root user. Turns out the rotation check doesn't check that the key is active before reporting on it. Not sure if the check should or shouldn't report on rotating inactive keys, but to minimize false-positives I thought I would make this update.

The other checks for rotations after different amounts of days should also change depending on if this is accepted.

### Description

Included an and statement to check if the key is active
Example values for `access_key_1_active`:
- `'access_key_1_active': 'true'` 
- `'access_key_1_active': 'false'`

### License

### DIFF
--- a/prowler/providers/aws/services/iam/iam_rotate_access_key_90_days/iam_rotate_access_key_90_days.py
+++ b/prowler/providers/aws/services/iam/iam_rotate_access_key_90_days/iam_rotate_access_key_90_days.py
@@ -26,7 +26,7 @@ class iam_rotate_access_key_90_days(Check):
                 )
             else:
                 old_access_keys = False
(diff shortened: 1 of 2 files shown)


MrCloudSec (dismissed):
Nice catch @Fennerr , thank you!
I'd recommend for your next contribution to install our pre-commit so we can speed up our review, you can do this with `pre-commit install`.

MrCloudSec (approved):

PR #1937, opened 2023-02-20
chore(secrets): Improve the status_extended with more information

Fennerr:
### Context

Updated the ECS, Lambda code and environment variables, and SSM documents secrets check output to be more informative so that the user can more easily investigate the finding. It now outputs the types of secrets found and the line number it was found on.

### Description

Created a secrets_string that contains the type of secret and line number for each secret detected and adds it to the status_extended message. 

I have indented the json.dumps so that the json output in the file that is scanned will span multiple lines, instead of being on one line. This let's me determine the line number which detect-secrets flagged on.

### License

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -66,10 +66,10 @@ repos:
         entry: bash -c 'pylint --disable=W,C,R,E -j 0 -rn -sn prowler/'
         language: system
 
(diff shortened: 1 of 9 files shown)


MrCloudSec (approved):
This is awesome, thank you @Fennerr 🤟🏼

PR #1942, opened 2023-02-21
fix(errors): handle errors when S3 buckets or EC2 instances are deleted

### LINKED ISSUE #1907: Treat missing resources as warnings instead of errors
I've noticed an edge condition on several Prowler AWS scans where the following happens:

1. Prowler makes a list or describe call to retrieve a list of all resources for a particular AWS service in a region (such as `s3:ListBuckets` or `ec2:DescribeInstances`).
2. Some of the resources in this list are either already been deleted (but are still shown in the list/describe response due to eventual consistency), or the resources are deleted shortly after the list/describe call.
3. Prowler makes a get or describe call (such as `ec2:DescribeInstanceAttribute` or `s3:GetBucketLocation`) on one of the deleted resources to pull more information about it.
4. This second API call fails.

Some example logs of this are shown below, but this issue is not necessarily limited to the S3 and EC2 services:
- `[File: ec2_service.py:243] 	[Module: ec2_service]	 ERROR: ClientError[237]: An error occurred (InvalidInstanceID.NotFound) when calling the DescribeInstanceAttribute operation: The instance ID '[REDACTED]' does not exist`
- `[File: s3_service.py:69] 	[Module: s3_service]	 ERROR: {'Name': '[REDACTED]', 'CreationDate': [REDACTED]} -- NoSuchBucket[48]: An error occurred (NoSuchBucket) when calling the GetBucketLocation operation: The specified bucket does not exist`

Currently this results in Prowler producing `ERROR` logs. This is problematic for us because we use the presence of an `ERROR` log during a scan to indicate that a partial scanning failure occurred (therefore resulting in a visibility gap). However, the case of a missing resource does not actually represent such a gap.

If Prowler could handle errors due to missing resources by producing a `WARNING` log and then skipping all checks for the missing resource, this would make it much easier to identify real visibility gaps in `ERROR` logs.

**Additional context**
Examples of issues that can cause "real visibility gaps" include but are not limited to the following:
- Missing permissions on the IAM user or role used to perform a Prowler scan
- Under-permissive resource policies configured on a resource being scanned
- Rate limiting

MrCloudSec:
### Description

Regarding https://github.com/prowler-cloud/prowler/issues/1907, handle errors when S3 Bukcets or EC2 Instances are deleted while Prowler is scanning.

### License

### DIFF
--- a/prowler/providers/aws/services/ec2/ec2_service.py
+++ b/prowler/providers/aws/services/ec2/ec2_service.py
@@ -1,6 +1,8 @@
 import threading
 from dataclasses import dataclass
 
+from botocore.client import ClientError
+
 from prowler.lib.logger import logger
 from prowler.lib.scan_filters.scan_filters import is_resource_filtered
 from prowler.providers.aws.aws_provider import generate_regional_clients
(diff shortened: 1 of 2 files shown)


jfagoagas (approved):
🔝

PR #1953, opened 2023-02-23
feat(s3_bucket_level_public_access_block): new check

### LINKED ISSUE #1916: Add s3_bucket_level_public_access_block check
Prowler already has a `s3_account_level_public_access_block` check. In the majority of cases, it is preferable to rely on AWS account level BPA settings and use this check to validate them, when attempting to enforce framework compliance or simply prevent over-exposed S3 buckets. 

However, in some cases, and AWS account may have a fixed number of buckets which are intentionally public for legitimate reasons (for example, hosting static assets or publishing CloudFormation templates). In these cases, it is not feasible to use either account level BPA settings or the `s3_account_level_public_access_block` check. Instead, bucket level BPA settings must be used with the buckets in the account which are not meant to be public.

Add a check named `s3_bucket_level_public_access_block`, which can be used to validate that bucket level BPA settings in accounts which cannot rely on account level BPA settings.

**Describe alternatives you've considered**
Suppressing the check entirely for accounts which cannot use account level BPAs. This is a sub-optimal solution as it results in visibility gaps for the non-public buckets in the account.

**Additional context**
The CIS AWS Foundations 1.5 benchmark `2.1.5` does not mandate account level BPAs, it only requires that some sort of BPA is applied to all buckets in an AWS account.

MrCloudSec:
### Description

Regarding https://github.com/prowler-cloud/prowler/issues/1916, add new check that complies with CIS control 2.1.5.

### License

### DIFF
--- a/prowler/compliance/aws/cis_1.5_aws.json
+++ b/prowler/compliance/aws/cis_1.5_aws.json
@@ -533,7 +533,7 @@
       "Id": "2.1.5",
       "Description": "Ensure that S3 Buckets are configured with 'Block public access (bucket settings)'",
       "Checks": [
-        "s3_account_level_public_access_blocks"
+        "s3_bucket_level_public_access_block"
       ],
       "Attributes": [
         {
--- a/prowler/compliance/aws/cis_1.4_aws.json
+++ b/prowler/compliance/aws/cis_1.4_aws.json
@@ -533,7 +533,7 @@
       "Id": "2.1.5",
       "Description": "Ensure that S3 Buckets are configured with 'Block public access (bucket settings)'",
       "Checks": [
-        "s3_account_level_public_access_blocks"
+        "s3_bucket_level_public_access_block"
       ],
       "Attributes": [
         {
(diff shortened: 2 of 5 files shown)


n4ch04 (approved):

prowler/compliance/aws/cis_1.5_aws.json
```diff
@@ -533,7 +533,7 @@
       "Id": "2.1.5",
       "Description": "Ensure that S3 Buckets are configured with 'Block public access (bucket settings)'",
       "Checks": [
-        "s3_account_level_public_access_blocks"
+        "s3_bucket_level_public_access_block"
```
addefisher, line 536: While I requested this check, I do not think it should be added to the CIS compliance file. Generally account level BPA settings are a better way to achieve CIS compliance. The `s3_bucket_level_public_access_block` check is extremely useful for edge cases (where an account has a legitimate need for a public bucket and a risk exception has been made), but the account level checks are a preferable approach for most environments / AWS accounts.

PR #1961, opened 2023-02-23
fix(kms): handle if describe_keys returns no value

n4ch04:
### Context

Fixes kms service issue when `DescribeKey` api call fails

### Description

Add `key` object attributes test to ensure that those attributes are not `None` after calling `DescribeKey`

### License


jfagoagas (approved):
Good catch!

PR #1970, opened 2023-02-24
fix(acm): Fix issues with list-certificates

jfagoagas:
### Context

There was an issue with the ACM service since we were retrieving the certificate data from the `describe_certificate` instead of the `list_certificates` and in the case of a Bad Request error from AWS the above checks were failing.

### Description

- Fix the following errors:
 - `acm_certificates_expiration_check -- AttributeError[13]: 'Certificate' object has no attribute 'expiration_days'`
 - `acm_certificates_transparency_logs_enabled -- AttributeError[11]: 'Certificate' object has no attribute 'type'`
- Improve the test cases for the ACM service and the above checks.

Also I've fixed a flaky test from TrustedAdvisor since in some cases the AWS Region was not present in the client.

### License

### DIFF
--- a/prowler/providers/aws/services/acm/acm_service.py
+++ b/prowler/providers/aws/services/acm/acm_service.py
@@ -1,7 +1,9 @@
 import threading
-from dataclasses import dataclass
+from datetime import datetime
+from typing import Optional
 
-from prowler.config.config import timestamp_utc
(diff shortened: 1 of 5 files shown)


MrCloudSec (approved):
Awesome!!!

PR #1977, opened 2023-02-24
chore(compliance): implements dynamic handling of available compliance frameworks

pedrooot:
### Context

Implements a method to handle compliance names in a dynamic way

### Description

Replaces the compliance list with a loop to take the names from compliance dir

### License


prowler/config/config.py
```diff
@@ -18,28 +18,17 @@
...
-]
+compliance_aws_dir = "./prowler/compliance/aws"
+available_compliance_frameworks = []
+with os.scandir(compliance_aws_dir) as files:
+    files = [
+        file.name
+        for file in files
+        if file.is_file()
+        and file.name.endswith(".json")
+        and available_compliance_frameworks.append(file.name.split(".")[0])
```
MrCloudSec: Have you though about the case that a framework name contains a `.` (for example cis_1.4_aws)?

MrCloudSec (changes requested):
Awesome job!!
Can you please remove the Pipfile files?
Also, check my comment.

prowler/config/config.py
pedrooot: True!! I fixed it, thanks Sergio

prowler/config/config.py
```diff
@@ -18,28 +18,17 @@
...
-]
+compliance_aws_dir = "./prowler/compliance/aws"
+available_compliance_frameworks = []
+with os.scandir(compliance_aws_dir) as files:
+    files = [
+        file.name
+        for file in files
+        if file.is_file()
+        and file.name.endswith(".json")
+        and available_compliance_frameworks.append(file.name.split(".j")[0])
```
JeffreySouza:
```suggestion
        and available_compliance_frameworks.append(file.name.removesuffix('.json')
```

pedrooot: Better! Thanks @JeffreySouza

MrCloudSec (approved):

PR #2001, opened 2023-02-28
Update AWS provider's default retry mode

Fennerr:
### Context

If the aws_retries_max_attempts argument is not specified, the retry mode would be set to legacy.

The standard retry mode has some improvements to the legacy mode, including handling a wider range of exceptions that should result in a retry being performed.
Further info: https://boto3.amazonaws.com/v1/documentation/api/latest/guide/retries.html

This change defaults the retry mode to the standard retry mode, and set the max retries to 5.

### Description

Added an else clause to specify that the standard retry mode with a max retries of 5 is used if the aws_retries_max_attempts argument is not specified.

### License

### DIFF
--- a/prowler/providers/common/audit_info.py
+++ b/prowler/providers/common/audit_info.py
@@ -137,6 +137,17 @@ Caller Identity ARN: {Fore.YELLOW}[{audit_info.audited_identity_arn}]{Style.RESE
             # Merge the new configuration
             new_boto3_config = current_audit_info.session_config.merge(config)
             current_audit_info.session_config = new_boto3_config
+        else:
+            # Use the standard retry mode, and increase max retries from 3 to 5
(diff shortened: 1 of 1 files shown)


jfagoagas:
Hi @Fennerr, within Prowler the Boto3 retryer is configured at the `AWS_Audit_Info` class here https://github.com/prowler-cloud/prowler/blob/27df481967cc2a746b0e7a765e933461715f441c/prowler/providers/aws/lib/audit_info/audit_info.py#L15 so even if the `aws_retries_max_attempts` is not set each client is configured with the standard retry mode with 3 max attempts.

Fennerr:
I see, thanks for pointing that out. I just didn't see it in this file and assumed it wasn't being done. My bad!

PR #2010, opened 2023-03-01
fix(iam): pydantic migration and reformat logic service

n4ch04:
### Context

When there were no permissions to get IAM info IAM checks were not being executed properly.

### Description

Reformat the IAM service to avoid this behaviour including migration from `dataclass` to `pydantic` `BaseModel` classes

### License

### DIFF
--- a/prowler/providers/aws/services/iam/iam_check_saml_providers_sts/iam_check_saml_providers_sts.py
+++ b/prowler/providers/aws/services/iam/iam_check_saml_providers_sts/iam_check_saml_providers_sts.py
@@ -15,13 +15,5 @@ class iam_check_saml_providers_sts(Check):
                 report.status = "PASS"
                 report.status_extended = f"SAML Provider {provider_name} has been found"
                 findings.append(report)
-        else:
-            report = Check_Report_AWS(self.metadata())
-            report.resource_id = ""
-            report.resource_arn = ""
-            report.region = iam_client.region
-            report.status = "FAIL"
-            report.status_extended = "No SAML Providers found"
-            findings.append(report)
 
         return findings
--- a/prowler/providers/aws/services/iam/iam_check_saml_providers_sts/iam_check_saml_providers_sts.py
+++ b/prowler/providers/aws/services/iam/iam_check_saml_providers_sts/iam_check_saml_providers_sts.py
@@ -5,14 +5,23 @@ from prowler.providers.aws.services.iam.iam_client import iam_client
 class iam_check_saml_providers_sts(Check):
     def execute(self) -> Check_Report_AWS:
         findings = []
-        for provider in iam_client.saml_providers:
+        if iam_client.saml_providers:
+            for provider in iam_client.saml_providers:
+                report = Check_Report_AWS(self.metadata())
+                provider_name = provider["Arn"].split("/")[1]
+                report.resource_id = provider_name
+                report.resource_arn = provider["Arn"]
+                report.region = iam_client.region
+                report.status = "PASS"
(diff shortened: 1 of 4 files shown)


sergargar:
Good job! Could you change the Any types to a particular one?

MrCloudSec (changes requested):
Good job! Could you change the Any types to a particular one?

prowler/providers/aws/services/iam/iam_check_saml_providers_sts/iam_check_saml_providers_sts.py
```diff
@@ -5,14 +5,23 @@
...
-            provider_name = provider["Arn"].split("/")[1]
-            report.resource_id = provider_name
-            report.resource_arn = provider["Arn"]
+            report.resource_id = ""
+            report.resource_arn = ""
             report.region = iam_client.region
-            report.status = "PASS"
-            report.status_extended = f"SAML Provider {provider_name} has been found"
+            report.status = "FAIL"
+            report.status_extended = "No SAML Providers found"
```
jfagoagas: Are we sure of this change? 

Also we cannot set empty both `resource_id` and `resource_arn`.

prowler/providers/aws/services/iam/iam_service.py
```diff
@@ -99,44 +103,43 @@ def __get_credential_report__(self):
...
+            logger.error(
+                f"{self.region} -- {error.__class__.__name__}[{error.__traceback__.tb_lineno}]: {error}"
+            )
+        finally:
             return groups
 
     def __get_account_summary__(self):
         try:
             account_summary = self.client.get_account_summary()
+            return account_summary
```
jfagoagas: What happens if the `self.client.get_account_summary()` raises an Exception?

prowler/providers/aws/services/iam/iam_service.py
```diff
@@ -150,70 +153,70 @@ def __get_password_policy__(self):
...
+        except Exception as error:
+            logger.error(
+                f"{self.region} -- {error.__class__.__name__}[{error.__traceback__.tb_lineno}]: {error}"
+            )
+        finally:
             return users
 
     def __list_virtual_mfa_devices__(self):
         try:
             list_virtual_mfa_devices_paginator = self.client.get_paginator(
```
jfagoagas, line 211: If the `"list_virtual_mfa_devices"` raises an exception we never return the `mfa_devices` because is referenced before assigned.

jfagoagas (changes requested):
Please check the line `363` in the `__list_policies_version__` function, why do we have a `pass` there?

prowler/providers/aws/services/iam/iam_check_saml_providers_sts/iam_check_saml_providers_sts.py
MrCloudSec: I agree, if there are no resources, we don't create a finding.

n4ch04: Changed

prowler/providers/aws/services/iam/iam_service.py
n4ch04: Changed

prowler/providers/aws/services/iam/iam_service.py
n4ch04: Changed

MrCloudSec (dismissed):

jfagoagas (approved):

PR #2023, opened 2023-03-02
fix(actions): fixed dispatch commit message

n4ch04:
### Context

Action was not sending the latest commit to remote dispatch

### Description

Change `LATEST_COMMIT_HASH` to use `github.event.after` instead of `github.event.head`

### License


jfagoagas (approved):
Good catch!

PR #2025, opened 2023-03-02
fix(cloudtrail): list tags only in owned trails

MrCloudSec:
### Description

Check the account of the trail in order to fix the following error when listing trail tags:
`ERROR: CloudTrailARNInvalidException[145]: An error occurred (CloudTrailARNInvalidException) when calling the ListTags operation: AccountID in ARN is wrong, you can only access resources with AccountID <redacted>.`

### License


jfagoagas (approved):
Nice one!

PR #2028, opened 2023-03-02
fix(windows-path):  --list-services bad split

garym-krrv:
### Context

When executing --list-services when running prowler on a windows machine, the split fails because the file structure on windows uses back slashes as apposed to forward slashes on linux.
On a windows box an error :-
File "C:\Source\3rd-party\prowler\prowler\lib\check\check.py", line 125
    print checks_tuple

This fix resolves this issue when executing on a Windows machine.

### Description

Added os.name if statement to check what os the prowler is being executed on when executing the --list-services command. This is so that the code correctly parses the file path with back slashes as apposed to forward slashes.

### License


MrCloudSec (approved):
Good catch! Thank you @garym-krrv

PR #2063, opened 2023-03-08
fix(ec2): avoid terminated instances

### LINKED ISSUE #2059: [Bug]: ec2_instance_managed_by_ssm triggers on terminated instances
**What happened?**

We are getting FAILs for terminated instances on rule ec2_instance_managed_by_ssm, although the terminated instances were managed by SSM

**How to reproduce it**

Terminate an instance managed by SSM, and in short time, run prowler.

**Expected behavior**

FAIL is not reported on terminated instances

**Additional context**

https://github.com/prowler-cloud/prowler/blob/eb9d6fa25c0978136b1fb084bca2cf46f0a3332c/prowler/providers/aws/services/ec2/ec2_instance_managed_by_ssm/ec2_instance_managed_by_ssm.py#L9 seems to return terminated instances for a while after they are terminated (this is standard AWS DescribeInstances behaviour). The call to SSM fails when passed an instance id that is not managed anymore by SSM.

MrCloudSec:
### Description

Make sure not to run ec2 checks on terminated instances.

### License


jfagoagas (approved):
Good job @sergargar 👏

PR #2092, opened 2023-03-15
feat(lambda_cloudtrail check): improved logic and status extended

n4ch04:
### Context

Improves check `awslambda_function_invoke_api_operations_cloudtrail_logging_enabled` logic and wording in `status extended`

### Description

Looks for the entry `AWS::Lambda::Function` in the array instead of checking only the first element and adds the `trail` word to `status_extended`

### License


jfagoagas (approved):
Good catch!

PR #2094, opened 2023-03-16
fix(ulimit check): test only when platform is not windows

n4ch04:
### Context

Prowler checks the ulimit setting to warn if it is less than 4096

### Description

Don't import the library `resource` to check it unless the platform is not Windows

### License


jfagoagas (approved):
Good one!

PR #2096, opened 2023-03-16
fix(ulimit check): try except when checking ulimit

n4ch04:
### Context

Fixes `resource` module import when dealing with `ulimit`

### Description

Adds `try ... except` logic to the ulimit check

### License

### DIFF
--- a/prowler/lib/check/check.py
+++ b/prowler/lib/check/check.py
@@ -356,7 +356,7 @@ def execute_checks(
         audit_progress=0,
     )
 
-    if not sys.platform.startswith("win32") or not sys.platform.startswith("cygwin"):
+    if os.name != "nt":
         try:
             from resource import RLIMIT_NOFILE, getrlimit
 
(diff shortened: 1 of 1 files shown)


prowler/lib/check/check.py
```diff
@@ -357,12 +356,23 @@ def execute_checks(
         audit_progress=0,
     )
 
-    # Check ulimit for the maximum system open files
-    soft, _ = getrlimit(RLIMIT_NOFILE)
-    if soft < 4096:
-        logger.warning(
-            f"Your session file descriptors limit ({soft} open files) is below 4096. We recommend to increase it to avoid errors. Solve it running this command `ulimit -n 4096`. For more info visit https://docs.prowler.cloud/en/latest/troubleshooting/"
-        )
+    if not sys.platform.startswith("win32") or not sys.platform.startswith("cygwin"):
```
MrCloudSec:
```suggestion
    if os.name != "nt":
```

MrCloudSec: This is what we are using to detect a Windows system

MrCloudSec (approved):

PR #2106, opened 2023-03-17
chore(version): check latest version

MrCloudSec:
### Description

Check latest version of Prowler in `--version` flag:
`Prowler 3.3.0 (it is the latest version, yay!)`
`Prowler 3.2.0 (latest is 3.3.0, upgrade for the latest features)`

### License

### DIFF
--- a/prowler/__main__.py
+++ b/prowler/__main__.py
@@ -3,6 +3,7 @@
 
 import sys
 
+from prowler.config.config import check_current_version
 from prowler.lib.banner import print_banner
 from prowler.lib.check.check import (
     bulk_load_checks_metadata,
@@ -54,6 +55,9 @@ def prowler():
     checks_file = args.checks_file
     severities = args.severity
     compliance_framework = args.compliance
+
+    if args.version:
+        check_current_version()
 
     # We treat the compliance framework as another output format
     if compliance_framework:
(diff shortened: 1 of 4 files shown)


prowler/__main__.py
```diff
@@ -3,6 +3,7 @@
 
 import sys
 
+from prowler.config.config import check_current_version
```
MrCloudSec:
```suggestion
```

prowler/__main__.py
```diff
@@ -55,6 +56,9 @@ def prowler():
     severities = args.severity
     compliance_framework = args.compliance
 
+    if args.version:
+        check_current_version()
+
```
MrCloudSec:
```suggestion
```

prowler/config/config.py
```diff
@@ -43,6 +44,19 @@
...
+    try:
+        latest_version = requests.get(
+            "https://api.github.com/repos/prowler-cloud/prowler/tags"
+        ).json()[0]["name"]
+        if latest_version != prowler_version:
+            return f"(latest is {latest_version}, upgrade for the latest features)"
+        else:
+            return "(it is the latest version, yay!)"
+    except Exception:
+        return ""
```
jfagoagas, line 57: Maybe we need to log some error here and then return, like:
```suggestion
       log.error()
   finally:
       return
```

jfagoagas (changes requested):

jfagoagas (approved):

PR #2108, opened 2023-03-19
feat(threading): Improve threading with some general functions

Fennerr:
### Context

I have come up with a new strategy for threading. There are 2 options, which have been implemented as decorators.
1. thread_per_region This will use a pool of threads (ThreadPoolExecutor), and initialize each thread to handle 1 region. This is intended to replace the previous __threading_call__ implementation
2. thread_per_item This is a separate pool of threads which is used to thread per item. You specify which list of items you would like to use as the iterator by passing the name of the object's attribute to the decorator. I did this because I could not seem to pass self.some_attribute as a variable to the decorator.

Thread_per_item is the main improvement here. As an example, this used to be the implementation of the `__list_tags_for_resource__` function for the `CloudWatch` class.

```
def __list_tags_for_resource__(self):
        logger.info("CloudWatch - List Tags...")
        try:
            for metric_alarm in self.metric_alarms:
                regional_client = self.regional_clients[metric_alarm.region]
                response = regional_client.list_tags_for_resource(
                    ResourceARN=metric_alarm.arn
                )["Tags"]
                metric_alarm.tags = response
        except Exception as error:
            logger.error(
            )
```

Instead of handling all of the metric_alarms (`for metric_alarm in self.metric_alarms:`), the function now takes in a `metric_alarm` parameter (shown below). `@thread_per_item("metric_alarms")` indicates to the decorator that a thread should be spawned for each item in `self.metric_alarms`, which it will pass in as a parameter to the decorated function. This allows for improved parallel performance, as each API call (`regional_client.list_tags_for_resource`) is now handled in a separate thread in the `ThreadPoolExecutor` pool, rather than 1 thread making all the calls sequentially. I assume this improvement will be more apparent when it comes to other services, such as retrieving the lambda code for each function.
```
@thread_per_item("metric_alarms")
    def __list_tags_for_resource__(self, metric_alarm):
        logger.info(f"CloudWatch - Listing Tags for metric alarm {metric_alarm.name}")
        try:
            regional_client = self.regional_clients[metric_alarm.region]
            response = regional_client.list_tags_for_resource(
                ResourceARN=metric_alarm.arn
            )["Tags"]
            metric_alarm.tags = response
        except Exception as error:
            logger.error(
            )
```

### Description

I have changed the implementation for the cloudwatch_service.py and ec2_service.py files.
To change from the `__threading_call__` to `@thread_per_region`:
- Make the service subclass the `from prowler.providers.aws.lib.classes import Service` class
- Remove the regional_client input parameter, and change instances of `regional_client` within the function to rather be `self.regional_client`

To use `@thread_per_item`:
- Make the service subclass the `from prowler.providers.aws.lib.classes import Service` class
- Remove the handling of the list of items from within the function
- Create the decorator with the name of the object's attribute that will be used to iterate through, and pass an item to the function (ie `self.log_groups`)
- Create another input paramter for the function for the item to be passed in (ie `log_group`)

### License

### DIFF
--- a/prowler/providers/aws/aws_provider.py
+++ b/prowler/providers/aws/aws_provider.py
@@ -10,14 +10,26 @@ from prowler.config.config import aws_services_json_file
 from prowler.lib.logger import logger
 from prowler.lib.utils.utils import open_file, parse_json_file
 from prowler.providers.aws.lib.audit_info.models import AWS_Assume_Role, AWS_Audit_Info
-
+import logging
 
 ################## AWS PROVIDER
 class AWS_Provider:
-    def __init__(self, audit_info):
+    def __init__(self, audit_info, thread_provider=False):
+        if thread_provider:
+            logger.disabled = True
+            original_boto3_level = logging.getLogger('boto3').getEffectiveLevel()
+            original_botocore_level = logging.getLogger('botocore').getEffectiveLevel()
+            logging.getLogger('boto3').setLevel(logging.WARNING)
+            logging.getLogger('botocore').setLevel(logging.WARNING)
+        
         logger.info("Instantiating aws provider ...")
         self.aws_session = self.set_session(audit_info)
         self.role_info = audit_info.assumed_role_info
+        
+        if thread_provider:
+            logger.disabled = False
+            logging.getLogger('boto3').setLevel(original_boto3_level)
+            logging.getLogger('botocore').setLevel(original_botocore_level)
 
     def get_session(self):
         return self.aws_session
--- a/prowler/providers/aws/aws_provider.py
+++ b/prowler/providers/aws/aws_provider.py
@@ -15,21 +15,9 @@ import logging
 ################## AWS PROVIDER
 class AWS_Provider:
     def __init__(self, audit_info, thread_provider=False):
-        if thread_provider:
-            logger.disabled = True
-            original_boto3_level = logging.getLogger('boto3').getEffectiveLevel()
-            original_botocore_level = logging.getLogger('botocore').getEffectiveLevel()
-            logging.getLogger('boto3').setLevel(logging.WARNING)
-            logging.getLogger('botocore').setLevel(logging.WARNING)
-        
         logger.info("Instantiating aws provider ...")
         self.aws_session = self.set_session(audit_info)
         self.role_info = audit_info.assumed_role_info
-        
-        if thread_provider:
-            logger.disabled = False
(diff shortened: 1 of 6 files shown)


n4ch04:
Hi @Fennerr, we are going to review this and test if we are going to integrate it in our current schema.
As always, thank you so much for investing your time in helping us to improve Prowler.

We are very thankful of the fantastic community Prowler has.

prowler/providers/aws/aws_provider.py
```diff
@@ -141,6 +140,36 @@ def generate_regional_clients(
         )
 
 
+def generate_regions_for_service(
```
jfagoagas, line 143: @Fennerr just a question before starting to review all the code in depth, why do we need to have both `generate_regional_clients` and `generate_regions_for_service`. Since the logic seems to be pretty much the same, could we merge this two functions into one?

Fennerr: Hey, you are correct - this can be merged into 1

At first I wanted to generate a list of regions for a service without generating the regional clients, and then created boto3 sessions per region/thread, and then created the regional client from the session. But I did some more looking around online, and the boto clients are generally threadsafe (there are some caveats mentioned here: https://boto3.amazonaws.com/v1/documentation/api/1.19.0/guide/clients.html#multithreading-or-multiprocessing-with-clients), and I didn't like the long initialization time that it took each thread to setup it's own boto session, so I changed the logic.

Something else I am also concerned about is the thread_per_region always creating 1 thread per region. I am doing this off the assumption that the threads will be called ThreadPoolExecutorX_Y (X = the pool, Y = the threads, and it will increase from 0 to len(regions)). It seems to always work as intended so far, but I would welcome it if you have a more succinct way of ensuring that 1 thread per region is spawned.

toniblyx:
hey @Fennerr we didn´t forget this, we just need more time for testing.

jfagoagas:
Hi @Fennerr since you created https://github.com/prowler-cloud/prowler/pull/3175 I think that PR overrides this one, am I right? If so, please close this PR pointing it with a comment to the new one.

Thanks!

Fennerr:
@jfagoagas that is correct, the newer implementation is more straightforward to use compared to these decorators, but I still like the decorator concept.

1 thing I would still like to explore is keeping the thread pool per service - instead of creating a new one each time the __threading_call__ is used. Im not sure if this might help with the open file descriptors or what implications it could have. But this PR can be closed for now

PR #2133, opened 2023-03-26
feat(new_checks): New AWS Organizations related checks

gabrielsoltz:
### Context

New AWS Organizations Service and Checks:

- `organizations_account_part_of_organizations`
- `organizations_delegated_administrators`
- `organizations_scp_check_deny_regions`

### Description

### License


prowler/providers/aws/services/organizations/organizations_account_part_of_organizations/organizations_account_part_of_organizations.metadata.json
```diff
@@ -0,0 +1,32 @@
...
+      "CLI": "",
+      "NativeIaC": "",
+      "Other": "",
+      "Terraform": ""
+    },
+    "Recommendation": {
+      "Text": "",
+      "Url": ""
+    }
+  },
```
jfagoagas: Why all these fields are empty? At least I think we can fill the risk and the remediation.

Please refer to https://www.trendmicro.com/cloudoneconformity/knowledge-base/aws/Organizations/ and https://docs.bridgecrew.io/docs to find more about the remediation.

prowler/providers/aws/services/organizations/organizations_service.py
```diff
@@ -0,0 +1,231 @@
...
+    def __describe_organization__(self):
+        logger.info("Organizations - Describe Organization...")
+
+        try:
+            # Check if Organizations is in-use
+            try:
+                organization_desc = self.client.describe_organization()["Organization"]
+                organization_arn = organization_desc["Arn"]
+                organization_id = organization_desc["Id"]
+                organization_master_id = organization_desc["MasterAccountId"]
```
jfagoagas: For accessing a key from a dictionary is better to use `dict.get()` like `organization_desc.get("Arn")`. This way if the value is not present it not raises a `KeyError` and returns `None`. Also you can set a default value if `None` is raised using ``organization_desc.get("Arn", default_value)`.

jfagoagas: Please, change it for every dict access. Thanks!

prowler/providers/aws/services/organizations/organizations_service.py
```diff
@@ -0,0 +1,231 @@
...
+                            delegated_administrators=organization_d_administrator,
+                        )
+                    )
+                else:
+                    # is filtered
+                    self.organizations.append(
+                        Organization(
+                            arn="", id="", status="NOT_AVAILABLE", master_id=""
+                        )
+                    )
```
jfagoagas: Why this here? We are not storing the resource if we don't want to audit it.

prowler/providers/aws/services/organizations/organizations_service.py
```diff
@@ -0,0 +1,231 @@
...
+    #                             aws_managed=policy["AwsManaged"],
+    #                             content=policy_content,
+    #                         )
+    #                     )
+    #     except Exception as error:
+    #         logger.error(
+    #             f"{self.region} -- {error.__class__.__name__}[{error.__traceback__.tb_lineno}]: {error}"
+    #         )
+    #     finally:
+    #         return policies
```
jfagoagas: Do we need this? We try to remove the commented code if we are not going to use it.

prowler/providers/aws/services/organizations/organizations_service.py
```diff
@@ -0,0 +1,231 @@
...
+                organization_master_id = organization_desc["MasterAccountId"]
+                organization_available_policy_types = organization_desc[
+                    "AvailablePolicyTypes"
+                ]
+                # Fetch policies for organization:
+                organization_policies = self.__list_policies__(
+                    organization_available_policy_types
+                )
+                # Fetch delegated administrators for organization:
+                organization_d_administrator = self.__list_delegated_administrators__()
```
jfagoagas: Could you change `organization_d_administrator` for `organization_delegated_administrator`?

prowler/providers/aws/services/organizations/organizations_account_part_of_organizations/organizations_account_part_of_organizations.metadata.json
jfagoagas: This comment applies for all the metadata files of this PR. Thanks!

prowler/providers/aws/services/organizations/organizations_delegated_administrators/organizations_delegated_administrators.metadata.json
```diff
@@ -0,0 +1,32 @@
+{
+  "Provider": "aws",
+  "CheckID": "organizations_delegated_administrators",
+  "CheckTitle": "Check if trusted delegated administratos",
```
jfagoagas: Could you include a more descriptive title like "Check if the AWS Organization for this account has trusted delegated administrators". What do you think?

jfagoagas:
```suggestion
  "CheckTitle": "Check if trusted delegated administrators",
```

prowler/providers/aws/services/organizations/organizations_delegated_administrators/organizations_delegated_administrators.metadata.json
```diff
@@ -0,0 +1,32 @@
...
+  "CheckType": [
+    "Logging and Monitoring"
+  ],
+  "ServiceName": "organizations",
+  "SubServiceName": "",
+  "ResourceIdTemplate": "arn:partition:service::account-id:organization/organization-id",
+  "Severity": "medium",
+  "ResourceType": "Other",
+  "Description": "Check if trusted delegated administratos",
```
jfagoagas:
```suggestion
  "Description": "Check if trusted delegated administrators",
```

prowler/providers/aws/services/organizations/organizations_delegated_administrators/organizations_delegated_administrators.py
```diff
@@ -0,0 +1,46 @@
...
+            report.resource_arn = org.arn
+            if org.status == "ACTIVE":
+                if org.delegated_administrators:
+                    for delegated_administrator in org.delegated_administrators:
+                        if (
+                            delegated_administrator.id
+                            not in organizations_trusted_delegated_administrators
+                        ):
+                            report.status = "FAIL"
+                            report.status_extended = f"Untrusted Deledated Administrators: {delegated_administrator.id}"
```
jfagoagas:
```suggestion
                            report.status_extended = f"Untrusted Delegated Administrators: {delegated_administrator.id}"
```

prowler/providers/aws/services/organizations/organizations_delegated_administrators/organizations_delegated_administrators.py
```diff
@@ -0,0 +1,46 @@
...
+                else:
+                    report.status = "PASS"
+                    report.status_extended = f"No Delegated Administrators: {org.id}"
+
+                    findings.append(report)
+            else:
+                report.status = "PASS"
+                report.status_extended = (
+                    "AWS Organizations is not in-use for this AWS Account"
+                )
```
jfagoagas: I think this is out of the scope of the check, we try to keep it as simple as we can. So if there is a check that checks if the AWS Organization is in use we should not include that here.

prowler/providers/aws/services/organizations/organizations_scp_check_deny_regions/organizations_scp_check_deny_regions.py
```diff
@@ -0,0 +1,102 @@
...
+                    report.status = "FAIL"
+                    report.status_extended = (
+                        f"No SCP restricting by regions found for org: {org.id}"
+                    )
+
+            else:
+                report.status = "FAIL"
+                report.status_extended = (
+                    "AWS Organizations is not in-use for this AWS Account"
+                )
```
jfagoagas, line 95: The same as before, this is is out of context here, so we only check if the organization is filtering regions.

tests/providers/aws/services/organizations/organizations_account_part_of_organizations/organizations_account_part_of_organizations.py
```diff
@@ -0,0 +1,60 @@
...
+from boto3 import client
+from moto import mock_organizations
+
+AWS_REGION = "us-east-1"
+
+
+class Test_organizations_account_part_of_organizations:
+    @mock_organizations
+    def test_no_organization(self):
+        from prowler.providers.aws.lib.audit_info.audit_info import current_audit_info
```
jfagoagas: We are changing this because is better to mock the `current_audit_info` object every time we use it in tests so use the `mock_set_audit_info` instead.

tests/providers/aws/services/organizations/organizations_account_part_of_organizations/organizations_account_part_of_organizations.py
```diff
@@ -0,0 +1,60 @@
...
+            # Test Check
+            from prowler.providers.aws.services.organizations.organizations_account_part_of_organizations.organizations_account_part_of_organizations import (
+                organizations_account_part_of_organizations,
+            )
+
+            check = organizations_account_part_of_organizations()
+            result = check.execute()
+
+            assert len(result) == 1
+            assert result[0].status == "FAIL"
```
jfagoagas: Please, assert all the values that are present in the `result[0]`.

tests/providers/aws/services/organizations/organizations_account_part_of_organizations/organizations_account_part_of_organizations.py
```diff
@@ -0,0 +1,60 @@
...
+        ):
+            # Test Check
+            from prowler.providers.aws.services.organizations.organizations_account_part_of_organizations.organizations_account_part_of_organizations import (
+                organizations_account_part_of_organizations,
+            )
+
+            check = organizations_account_part_of_organizations()
+            result = check.execute()
+
+            assert len(result) == 1
```
jfagoagas: The same here, we need to assert every value.

tests/providers/aws/services/organizations/organizations_account_part_of_organizations/organizations_account_part_of_organizations.py
jfagoagas, line 1: Please, include another test to check the scenario where the account is not present, so in this tests we need to cover:
- No organizations
- Organization == "Active"
- Organization != "Active"

tests/providers/aws/services/organizations/organizations_delegated_administrators/organizations_delegated_administrators.py
```diff
@@ -0,0 +1,145 @@
...
+from boto3 import client
+from moto import mock_organizations
+
+AWS_REGION = "us-east-1"
+
+
+class Test_organizations_delegated_administrators:
+    @mock_organizations
+    def test_no_organization(self):
+        from prowler.providers.aws.lib.audit_info.audit_info import current_audit_info
```
jfagoagas: The same as before, better to use the `set_mock_audit_info` here.

tests/providers/aws/services/organizations/organizations_delegated_administrators/organizations_delegated_administrators.py
```diff
@@ -0,0 +1,145 @@
...
+        # Delegate Administrator
+        conn.register_delegated_administrator(
+            AccountId=response["CreateAccountStatus"]["AccountId"],
+            ServicePrincipal="config-multiaccountsetup.amazonaws.com",
+        )
+
+        def mock_get_config_var(config_var):
+            if config_var == "organizations_trusted_delegated_administrators":
+                return [response["CreateAccountStatus"]["AccountId"]]
+            return []
```
jfagoagas: Could yo move this function to the top of the file? Is better for other test cases to use it.

tests/providers/aws/services/organizations/organizations_scp_check_deny_regions/organizations_scp_check_deny_regions.py
jfagoagas, line 1: The same as before, I leave it just as a reminder to:
- Use the `set_mock_audit_info`
- Assert all the values from the result

tests/providers/aws/services/organizations/organizations_scp_check_deny_regions/organizations_scp_check_deny_regions.py
```diff
@@ -0,0 +1,152 @@
...
+            Content=scp_restrict_regions_with_deny(),
+            Description="Test",
+            Name="Test",
+            Type="SERVICE_CONTROL_POLICY",
+        )
+
+        def mock_get_config_var(config_var):
+            if config_var == "organizations_enabled_regions":
+                return ["us-east-1"]
+            return []
```
jfagoagas: Please, move this to the top of the function to be able to reuse it.

tests/providers/aws/services/organizations/organizations_service_test.py
jfagoagas, line 1: I only see two test for the service's functions, are we testing all?

jfagoagas (changes requested):
Good job @gabrielsoltz !! Please check my comments and if you have any question please feel free to ping me. Thanks!

prowler/providers/aws/services/organizations/organizations_account_part_of_organizations/organizations_account_part_of_organizations.metadata.json
gabrielsoltz: This was pending to define; I added basic remediation.

prowler/providers/aws/services/organizations/organizations_service.py
gabrielsoltz: True. Good Catch.

prowler/providers/aws/services/organizations/organizations_service.py
gabrielsoltz: Agree that this could be confusing, but this was on purpose: What we are defining here is that organization is not in use. `self.organizations` is a list, but it will always have one element (there is only a maximum one organization by account), what I'm doing here is adding a dummy organization in case the service is disabled. The reason why I'm using a list here is to be consistent with all the other checks, like here: https://github.com/prowler-cloud/prowler/blob/master/prowler/providers/aws/services/securityhub/securityhub_service.py#L88

prowler/providers/aws/services/organizations/organizations_service.py
gabrielsoltz: We should be using this API Call (`list_policies_for_target`), but for some AWS Magic reason, it is not answering what it is supposed to answer; that's why I left it for reference in case someone wants to replace `list_policies` with this one.

prowler/providers/aws/services/organizations/organizations_service.py
gabrielsoltz: 🚀

prowler/providers/aws/services/organizations/organizations_scp_check_deny_regions/organizations_scp_check_deny_regions.py
gabrielsoltz: If AWS Organizations is not enabled, wich means SCP policies are not enabled, which means regions are not restricted, which menas FAIL as I see it. 🤔

tests/providers/aws/services/organizations/organizations_account_part_of_organizations/organizations_account_part_of_organizations.py
gabrielsoltz: I don't understand what you mean here; all the tests are defined like this for checks, can u show me how should it look like? do I need to change all tests?

tests/providers/aws/services/organizations/organizations_account_part_of_organizations/organizations_account_part_of_organizations.py
gabrielsoltz: 🚀

tests/providers/aws/services/organizations/organizations_account_part_of_organizations/organizations_account_part_of_organizations.py
gabrielsoltz: 🚀

tests/providers/aws/services/organizations/organizations_service_test.py
gabrielsoltz: Yes, the list_policies function then calls the other functions, so we are testing all of them.

tests/providers/aws/services/organizations/organizations_delegated_administrators/organizations_delegated_administrators.py
gabrielsoltz: I can't move it to the top as it depends on `conn.create_account()`

tests/providers/aws/services/organizations/organizations_scp_check_deny_regions/organizations_scp_check_deny_regions.py
gabrielsoltz: But it is different for each test, on purpose.

prowler/providers/aws/services/organizations/organizations_service.py
jfagoagas: You're right, thanks for the explanation!

prowler/providers/aws/services/organizations/organizations_service.py
jfagoagas: Ok, we can leave this there for the future.

tests/providers/aws/services/organizations/organizations_account_part_of_organizations/organizations_account_part_of_organizations.py
jfagoagas: At the beginning we started importing the  `current_audit_info` object into the check's test but we realised that it was a bad practice since if shared among the code and for the tests is better to be in an isolated scenario. What I said is to use the `set_mock_audit_info`:
```python
# Mocked Audit Info
    def set_mocked_audit_info(self):
        audit_info = AWS_Audit_Info(
            session_config=None,
            original_session=None,
            audit_session=session.Session(
                profile_name=None,
                botocore_session=None,
            ),
            audited_account=None,
            audited_user_id=None,
            audited_partition="aws",
            audited_identity_arn=None,
            profile=None,
            profile_region=None,
            credentials=None,
            assumed_role_info=None,
            audited_regions=None,
            organizations_metadata=None,
            audit_resources=None,
        )
        return audit_info
```

So, if you can change the test you did and instead of importing the shared object create a mocked one.

tests/providers/aws/services/organizations/organizations_delegated_administrators/organizations_delegated_administrators.py
jfagoagas: Yes, but you should mock that return value there.

tests/providers/aws/services/organizations/organizations_scp_check_deny_regions/organizations_scp_check_deny_regions.py
jfagoagas: Ok, we can leave it there for the moment but I think you can pass custom arguments when mocking.

tests/providers/aws/services/organizations/organizations_service_test.py
jfagoagas: Awesome!

tests/providers/aws/services/organizations/organizations_delegated_administrators/organizations_delegated_administrators.py
jfagoagas, line 1: You have to rename this file appending `_test.py`.

tests/providers/aws/services/organizations/organizations_account_part_of_organizations/organizations_account_part_of_organizations.py
jfagoagas, line 1: You have to rename this file appending `_test.py`.

tests/providers/aws/services/organizations/organizations_scp_check_deny_regions/organizations_scp_check_deny_regions.py
jfagoagas, line 1: You have to rename this file appending `_test.py`.

prowler/providers/aws/services/organizations/organizations_account_part_of_organizations/organizations_account_part_of_organizations.metadata.json
```diff
@@ -0,0 +1,32 @@
...
+  "CheckType": [
+    "Logging and Monitoring"
+  ],
+  "ServiceName": "organizations",
+  "SubServiceName": "",
+  "ResourceIdTemplate": "arn:partition:service::account-id:organization/organization-id",
+  "Severity": "medium",
+  "ResourceType": "Other",
+  "Description": "Ensure that AWS Organizations service is currently in use.",
+  "Risk": "",
```
jfagoagas: Could it be possible to fill the risk at least?

prowler/providers/aws/services/organizations/organizations_delegated_administrators/organizations_delegated_administrators.metadata.json
```diff
@@ -0,0 +1,32 @@
...
+  "CheckType": [
+    "Logging and Monitoring"
+  ],
+  "ServiceName": "organizations",
+  "SubServiceName": "",
+  "ResourceIdTemplate": "arn:partition:service::account-id:organization/organization-id",
+  "Severity": "medium",
+  "ResourceType": "Other",
+  "Description": "This check verify if there are AWS Organizations delegated adminsitrators and if they are trusted (you can define your trusted delegated adminsitrator in Prowler configuration)",
+  "Risk": "",
```
jfagoagas: Could it be possible to fill the risk at least?

prowler/providers/aws/services/organizations/organizations_delegated_administrators/organizations_delegated_administrators.py
```diff
@@ -0,0 +1,46 @@
...
+                            delegated_administrator.id
+                            not in organizations_trusted_delegated_administrators
+                        ):
+                            report.status = "FAIL"
+                            report.status_extended = f"Untrusted Delegated Administrators: {delegated_administrator.id}"
+                        else:
+                            report.status = "PASS"
+                            report.status_extended = f"Trusted Delegated Administrator: {delegated_administrator.id}"
+
+                        findings.append(report)
```
jfagoagas: I think you can call this once at the end of the function instead of having three right?

prowler/providers/aws/services/organizations/organizations_delegated_administrators/organizations_delegated_administrators.py
jfagoagas: I need to give this a thought again.

prowler/providers/aws/services/organizations/organizations_scp_check_deny_regions/organizations_scp_check_deny_regions.metadata.json
```diff
@@ -0,0 +1,32 @@
...
+  "CheckType": [
+    "Logging and Monitoring"
+  ],
+  "ServiceName": "organizations",
+  "SubServiceName": "",
+  "ResourceIdTemplate": "arn:partition:service::account-id:organization/organization-id",
+  "Severity": "low",
+  "ResourceType": "Other",
+  "Description": "As best practice, AWS Regions should be restricted and only allow the ones that are needed.",
+  "Risk": "",
```
jfagoagas: Could it be possible to fill the risk at least?

prowler/providers/aws/services/organizations/organizations_scp_check_deny_regions/organizations_scp_check_deny_regions.py
jfagoagas: I need to give this a thought again.

prowler/providers/aws/services/organizations/organizations_scp_check_deny_regions/organizations_scp_check_deny_regions.py
```diff
@@ -0,0 +1,102 @@
...
+                                else:
+                                    # We found a statement restricting some regions, we continue to check if there is another one more restrictive
+                                    scp_check_deny_some_regions = True
+
+                        if scp_check_deny_all_regions:
+                            report.status = "PASS"
+                            report.status_extended = (
+                                f"SCP policy restricting regions found: {policy.id}"
+                            )
+                            findings.append(report)
```
jfagoagas: I think you can call this once at the end of the function instead of having some right?

prowler/providers/aws/services/organizations/organizations_service.py
```diff
@@ -0,0 +1,233 @@
...
+        try:
+            # Check if Organizations is in-use
+            try:
+                organization_desc = self.client.describe_organization()["Organization"]
+                organization_arn = organization_desc.get("Arn")
+                organization_id = organization_desc.get("Id")
+                organization_master_id = organization_desc.get("MasterAccountId")
+                organization_available_policy_types = organization_desc[
+                    "AvailablePolicyTypes"
+                ]
```
jfagoagas:
```suggestion
                organization_available_policy_types = organization_desc.get("AvailablePolicyTypes")
```

prowler/providers/aws/services/organizations/organizations_service.py
```diff
@@ -0,0 +1,233 @@
...
+                ):
+                    for policy in page["Policies"]:
+                        policy_content = self.__describe_policy__(policy["Id"])
+                        policy_targets = self.__list_targets_for_policy__(policy["Id"])
+                        self.policies.append(
+                            Policy(
+                                arn=policy["Arn"],
+                                id=policy["Id"],
+                                type=policy["Type"],
+                                aws_managed=policy["AwsManaged"],
```
jfagoagas: Could you do the same here using `dict.get()`? Thanks!

prowler/providers/aws/services/organizations/organizations_service.py
```diff
@@ -0,0 +1,233 @@
...
+            for page in list_delegated_administrators_paginator.paginate():
+                for delegated_administrator in page["DelegatedAdministrators"]:
+                    self.delegated_administrators.append(
+                        DelegatedAdministrator(
+                            arn=delegated_administrator["Arn"],
+                            id=delegated_administrator["Id"],
+                            name=delegated_administrator["Name"],
+                            email=delegated_administrator["Email"],
+                            status=delegated_administrator["Status"],
+                            joinedmethod=delegated_administrator["JoinedMethod"],
```
jfagoagas: Could you do the same here using dict.get()? Thanks!

prowler/providers/aws/services/organizations/organizations_service.py
```diff
@@ -0,0 +1,233 @@
...
+                            master_id=organization_master_id,
+                            policies=organization_policies,
+                            delegated_administrators=organization_delegated_administrator,
+                        )
+                    )
+                else:
+                    # is filtered
+                    self.organizations.append(
+                        Organization(
+                            arn="", id="", status="NOT_AVAILABLE", master_id=""
```
jfagoagas: If we are creating this dummy object here we cannot leave these fields as empty strings since all check's findings should have at least the ID or the ARN to be filled.

Check the dummy SecurityHub Hub:
```python
SecurityHubHub(
                            arn="",
                            id="Security Hub",
                            status="NOT_AVAILABLE",
                            standards="",
                            integrations="",
                        )
```

tests/providers/aws/services/organizations/organizations_account_part_of_organizations/organizations_account_part_of_organizations.py
```diff
@@ -0,0 +1,73 @@
...
+            assert result[0].status == "FAIL"
+            assert search(
+                "AWS Organizations is not in-use for this AWS Account",
+                result[0].status_extended,
+            )
+            assert result[0].resource_id == ""
```
jfagoagas: This can't be empty, please refer to the service comment I posted.

jfagoagas (changes requested):

tests/providers/aws/services/organizations/organizations_delegated_administrators/organizations_delegated_administrators.py
gabrielsoltz: renamed.

tests/providers/aws/services/organizations/organizations_account_part_of_organizations/organizations_account_part_of_organizations.py
gabrielsoltz: renamed.

tests/providers/aws/services/organizations/organizations_scp_check_deny_regions/organizations_scp_check_deny_regions.py
gabrielsoltz: renamed.

prowler/providers/aws/services/organizations/organizations_delegated_administrators/organizations_delegated_administrators.py
gabrielsoltz: I don't know what I was thinking. Fixed!

prowler/providers/aws/services/organizations/organizations_account_part_of_organizations/organizations_account_part_of_organizations.metadata.json
gabrielsoltz: added

prowler/providers/aws/services/organizations/organizations_delegated_administrators/organizations_delegated_administrators.metadata.json
gabrielsoltz: risk added

prowler/providers/aws/services/organizations/organizations_scp_check_deny_regions/organizations_scp_check_deny_regions.metadata.json
gabrielsoltz: risk added

prowler/providers/aws/services/organizations/organizations_scp_check_deny_regions/organizations_scp_check_deny_regions.py
gabrielsoltz: i improved this. thanks.

prowler/providers/aws/services/organizations/organizations_service.py
gabrielsoltz: i fixed this.

prowler/providers/aws/services/organizations/organizations_service.py
gabrielsoltz: fixed.

prowler/providers/aws/services/organizations/organizations_service.py
gabrielsoltz: fixed.

tests/providers/aws/services/organizations/organizations_account_part_of_organizations/organizations_account_part_of_organizations.py
gabrielsoltz: fixed.

tests/providers/aws/services/organizations/organizations_account_part_of_organizations/organizations_account_part_of_organizations.py
gabrielsoltz: fixed 🚀

tests/providers/aws/services/organizations/organizations_scp_check_deny_regions/organizations_scp_check_deny_regions.py
gabrielsoltz: fixed 🚀

tests/providers/aws/services/organizations/organizations_delegated_administrators/organizations_delegated_administrators.py
gabrielsoltz: fixed.

tests/providers/aws/services/organizations/organizations_account_part_of_organizations/organizations_account_part_of_organizations.py
gabrielsoltz: `No organizations` and `Organization != "Active"` is the same test. If there is an organization (AWS organization is active), the account is part of it, always: This test is executed from the AWS account "point of view", so you are describing organizations the account is part of (https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/organizations/client/describe_organization.html). You either get the organization you are part of, or no organizations (the service is disabled).

prowler/config/config.yaml
```diff
@@ -41,3 +41,13 @@ obsolete_lambda_runtimes:
     "dotnetcore2.1",
     "ruby2.5",
   ]
+
+# AWS Enabled Regions
+# organizations_scp_check_deny_regions
+organizations_enabled_regions: [
+  'eu-central-1',
+  'eu-west-1',
+  "us-east-1"
```
jfagoagas: I think we should leave this empty since it is used by the check.

prowler/providers/aws/services/organizations/organizations_delegated_administrators/organizations_delegated_administrators.metadata.json
```diff
@@ -0,0 +1,32 @@
...
+  "CheckTitle": "Check if AWS Organizations delegated administrators are trusted",
+  "CheckType": [
+    "Logging and Monitoring"
+  ],
+  "ServiceName": "organizations",
+  "SubServiceName": "",
+  "ResourceIdTemplate": "arn:partition:service::account-id:organization/organization-id",
+  "Severity": "high",
+  "ResourceType": "Other",
+  "Description": "This check verify if there are AWS Organizations delegated adminsitrators and if they are trusted (you can define your trusted delegated adminsitrator in Prowler configuration)",
```
jfagoagas:
```suggestion
  "Description": "This check verify if there are AWS Organizations delegated administrators and if they are trusted (you can define your trusted delegated administrator in Prowler configuration)",
```

prowler/providers/aws/services/organizations/organizations_delegated_administrators/organizations_delegated_administrators.metadata.json
```diff
@@ -0,0 +1,32 @@
...
+  "Risk": "The risk associated with having untrusted AWS Organizations delegated administrators is that they may have the ability to access and make changes to sensitive data and resources within an organization's AWS accounts. This can result in unauthorized access or data breaches, which can lead to financial losses, damage to reputation, and legal liabilities. It's important to carefully vet and monitor AWS Organizations delegated administrators to ensure that they are trustworthy and have a legitimate need for access to the organization's resources.",
```
jfagoagas:
```suggestion
  "Risk": "The risk associated with having untrusted delegated administrators within an AWS Organizations is that they may have the ability to access and make changes to sensitive data and resources within an organization's AWS accounts. This can result in unauthorized access or data breaches, which can lead to financial losses, damage to reputation, and legal liabilities. It's important to carefully vet and monitor AWS Organizations delegated administrators to ensure that they are trustworthy and have a legitimate need for access to the organization's resources.",
```

prowler/providers/aws/services/organizations/organizations_scp_check_deny_regions/organizations_scp_check_deny_regions.py
jfagoagas, line 1: I'm testing the check and it is raising a fail for one of our accounts that has SCPs, we can review it later.

prowler/providers/aws/services/organizations/organizations_delegated_administrators/organizations_delegated_administrators.py
jfagoagas: In this case I'd leave the check without the last part, so if there is no organizations we could not evaluate the deletegated administrators since it's a feature of the Orgs.

prowler/providers/aws/services/organizations/organizations_scp_check_deny_regions/organizations_scp_check_deny_regions.py
jfagoagas: In this case, is right since filtering regions affects the account, so this check is correct.

jfagoagas (changes requested):

prowler/providers/aws/services/organizations/organizations_scp_check_deny_regions/organizations_scp_check_deny_regions.py
jfagoagas: I get the following errors:
```
2023-03-30 09:01:22,587 [File: organizations_service.py:120]    [Module: organizations_service]  ERROR: us-east-1 -- AccessDeniedException[101]: An error occurred (AccessDeniedException) when calling the ListPolicies operation: You don't have permissions to access this resource.
2023-03-30 09:01:23,218 [File: organizations_service.py:177]    [Module: organizations_service]  ERROR: us-east-1 -- AccessDeniedException[164]: An error occurred (AccessDeniedException) when calling the ListDelegatedAdministrators operation: You don't have permissions to access this resource.
```

So we need to handle that because if we don't have permissions to check it we can't raise a FAIL.

tests/providers/aws/services/organizations/organizations_delegated_administrators/organizations_delegated_administrators.py
gabrielsoltz: function removed.

prowler/providers/aws/services/organizations/organizations_delegated_administrators/organizations_delegated_administrators.py
gabrielsoltz: changed

prowler/config/config.yaml
gabrielsoltz: ok

prowler/providers/aws/services/organizations/organizations_scp_check_deny_regions/organizations_scp_check_deny_regions.py
gabrielsoltz: fixed

jfagoagas (approved):
Awesome 👏 

Thank your very much @gabrielsoltz 🚀

PR #2134, opened 2023-03-26
feat(python-option): Make it possible to run Prowler from Python as a library

meowmeowxw:
The --python option enable prowler to be called by a python program to manipulate findings directly

### Context

Hello, It would be nice to be able to use prowler from python. In this way it's possible to have the type hint of the checks and manipulate them on the fly. Maybe with an additional PR we could also change the code to return the findings with a `yield` keyword and get them in real time.

To run prowler from python:

```
# inside the root directory of the project
pip install .
```

And then with a script:

```py
from prowler import __main__
from prowler.lib.check.models import Check_Report_AWS
from typing import List

arguments = ["aws", "--region", "eu-west-1", "--python", "-c", "account_maintain_current_contact_details"]
findings: List[Check_Report_AWS] = __main__.prowler(args=arguments)
for finding in findings:
    print(finding.status)
    print(finding.check_metadata)
```

### Description

No dependencies are required for this PR. The PR make it possible to run prowler directly from python. Let me know if it's something that you think is useful for your project, or I'll close the PR. A redesign of the __main__.py would be needed to make it a bit cleaner :D 

### License

### DIFF
--- a/prowler/__main__.py
+++ b/prowler/__main__.py
@@ -39,10 +39,10 @@ from prowler.providers.common.outputs import set_provider_output_options
 from prowler.providers.common.quick_inventory import run_provider_quick_inventory
 
 
-def prowler():
+def prowler(args=None):
     # Parse Arguments
     parser = ProwlerArgumentParser()
-    args = parser.parse()
+    args = parser.parse(args=args)
 
     # Save Arguments
     provider = args.provider
@@ -50,6 +50,7 @@ def prowler():
     excluded_checks = args.excluded_checks
     excluded_services = args.excluded_services
     services = args.services
+    use_python = args.python
     categories = args.categories
     checks_file = args.checks_file
     severities = args.severity
@@ -66,7 +67,10 @@ def prowler():
         print_banner(args)
 
     if args.list_services:
-        print_services(list_services(provider))
+        service_list = list_services(provider)
+        if use_python:
+            return service_list
+        print_services(service_list)
         sys.exit()
 
     # Load checks metadata
@@ -74,7 +78,10 @@ def prowler():
     bulk_checks_metadata = bulk_load_checks_metadata(provider)
 
     if args.list_categories:
-        print_categories(list_categories(provider, bulk_checks_metadata))
+        category_list = list_categories(provider, bulk_checks_metadata)
+        if use_python:
+            return category_list
+        print_categories(category_list)
         sys.exit()
 
     bulk_compliance_frameworks = {}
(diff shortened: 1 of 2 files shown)


jfagoagas:
Hi @meowmeowxw, this is a great addition to Prowler but I'm concerned about all the places where if we catch a critical error Prowler exits the execution, since it will cause to exit the father Python process usign Prowler as a library, causing some abrupt errors instead of the ability to catch the exception.

What do you think?

Also, as I can see this option does not provide a full compatibility with all Prowler's features right?

Thanks for using Prowler.

meowmeowxw:
Hello @jfagoagas ,
Thank you for the reply.

> Also, as I can see this option does not provide a full compatibility with all Prowler's features right?

Yes just some of them, but I think the most important one is the ability to collect findings. The other `flags` of prowler are mostly just to get a list of checks/services/options right?

> I'm concerned about all the places where if we catch a critical error Prowler exits the execution, since it will cause to exit the father Python process usign Prowler as a library, causing some abrupt errors instead of the ability to catch the exception.

I'll check better the codebase and see what I can do.

What do you think if I focus only on the findings gathering part? Would it suits you or would you like to enable python for everything? (It would require a major refactoring for most of the functions to not print results directly)

jfagoagas:
> Also, as I can see this option does not provide a full compatibility with all Prowler's features right?
> 
>> Yes just some of them, but I think the most important one is the ability to collect findings. The other `flags` of prowler are mostly just to get a list of checks/services/options right?
> 

There are flags to export findings to Security Hub, to select the output format you want, etc. You can check it running `prowler aws/azure/gcp --help`.


> What do you think if I focus only on the findings gathering part? Would it suits you or would you like to enable python for everything? (It would require a major refactoring for most of the functions to not print results directly)

I think focusing only on the findings gathering part would be great as a first approach to this new feature, I think the `yield` part can be really useful here at least to give it a try.

Thanks!

toniblyx:
hi @meowmeowxw, if I may ask, what is the use case for this feature? Is it by any chance to run it as a Lambda function?

meowmeowxw:
Hi @toniblyx ,
Yes it would enable an easier integration into lambda instead of using subprocesses.
It would also be possible to manipulate the findings in real time, and add type hinting (in this way you don't have to load the json and convert it back to Check_Report_AWS)

jfagoagas:
Hi @meowmeowxw, I apologize for not having answered sooner. Actually we are working towards having a way to call Prowler as a library, this is in our roadmap. Do you want to leave this PR opened until that or we can close it?

Thanks for your time 🙌

meowmeowxw:
Hi,
I will close this PR since I don't have time.
Thanks for the ping :smile:

PR #2135, opened 2023-03-27
fix(brew): move brew formula action to the bottom

MrCloudSec:
### Description

Fix action for updating prowler brew formula.

### License


.github/workflows/pypi-release.yml
jfagoagas, line 1: Why do we need to move it?

MrCloudSec: In case this action fails, so the rest of them can be completed first.

jfagoagas (approved):
Awesome, good catch!

PR #2144, opened 2023-03-28
fix(azure output): change default values of audit identity metadata

n4ch04:
### Context

Azure output files are failing when the tenant domain info can't be retrieved

### Description

Change the default values of the `Azure_Identity_Info` attributes to be compatible with the definition of `Azure_Check_Output_CSV`

### License


jfagoagas (approved):
Good catch!

Issue #2107, opened 2023-03-19
[Bug]: Security groups: Ensure there are no Security Groups without ingress filtering being used no correct description


JairoBlanco:
### Steps to Reproduce

I'm running Prowler on AWS and I'm getting what I understand is not a correct output as it points that some Security groups "has no ingress filtering and it is being used". It's pointing specially to LoadBalancer Security groups that have ingress rules that look like this: 

![Example](https://user-images.githubusercontent.com/111416013/226187085-b5519c6f-2bae-446f-8dec-68f1405f2af2.png)

I've being hardly trying to get rid of it but it's imposible and the explanation it's not clear enought. Can you at least tell me what's going on and how I supposed to filter internet connections through HTTPS to a ALB?

### Expected behavior

No error should be sawn or at least a clear explanation of what's wrong.

### Actual Result with Screenshots or Logs

![image](https://user-images.githubusercontent.com/111416013/226187246-6958b8b5-17af-478d-b19b-0ba175fe1b93.png)

### How did you install Prowler?

Using pip

### Environment Resource

EC2

### OS used

Windows 10

### Prowler version

Prowler 3.2.4

### Pip version

pip 22.2.2

n4ch04:
Hi @JairoBlanco, we are going to review this and let you know asap.

Thanks for using Prowler !!

JairoBlanco:
@n4ch04  I had a similar issue with EFS efs_not_publicly_accessible. Then I realize it was a permissions issue. Maybe this is the same, dunno. I will create a PR in the following weeks with the credentials that are missing.

JairoBlanco:
I tried to give full permissions and the flag was still failing. Is there any update from your side? 😄

JairoBlanco:
Hello, any update on this?

n4ch04:
Hi @JairoBlanco,  we are working on this right now, when we have an update we'll let you know.

n4ch04:
Hi @JairoBlanco,
After reviewing the check and how it works we have decided to delete the check `ec2_securitygroup_in_use_without_ingress_filtering` from Prowler, since the check `ec2_securitygroup_allow_ingress_from_internet_to_any_port` looks for sgs with `0.0.0.0/0` CIDR ranges in ingress rules.
You can check the changes here #2148

JairoBlanco:
Hello @n4ch04, thanks you very much for you help! I'll close this issue 😄

PR #2148, opened 2023-03-28
fix(delete check): delete check ec2.._in_use_without_ingrgess_filtering

n4ch04:
### Context

Coming from https://github.com/prowler-cloud/prowler/issues/2107

The check `ec2_securitygroup_in_use_without_ingress_filtering` was redundant and was not adding added value since it was labeling as FAIL any ingress rule in sgs with any public ip (not only `0.0.0.0/0`) in it.

### Description

Delete the check and all its ocurrences, replacing it by `ec2_securitygroup_allow_ingress_from_internet_to_any_port` when necessary  

### License


jfagoagas (approved):
Good one!

PR #2156, opened 2023-03-31
fix(secretsmanager_automatic_rotation_enabled): Improve description for Secrets Manager secret rotation

visit1985:
"key rotation" is misleading in this check, because secrets use kms keys.

Not sure if this is the only place.


jfagoagas (approved):
Good one! Thanks for using Prowler @visit1985 🚀

PR #2176, opened 2023-04-05
fix(iam): handle no display name error in service account

MrCloudSec:
### Description

Handle no display name error in service account

### License


jfagoagas (approved):
Good catch!

PR #2181, opened 2023-04-10
feat(html): add html to Azure and GCP

MrCloudSec:
### Description

Add HTML output to Azure and GCP providers.
image
image

### License

### DIFF
--- a/prowler/lib/outputs/html.py
+++ b/prowler/lib/outputs/html.py
@@ -1,3 +1,4 @@
+import importlib
 import sys
 from os import path
 
@@ -334,7 +335,7 @@ def add_html_footer(output_filename, output_directory):
         sys.exit(1)
 
 
-def get_assessment_summary(audit_info):
+def get_aws_html_assessment_summary(audit_info):
     try:
         if isinstance(audit_info, AWS_Audit_Info):
             if not audit_info.profile:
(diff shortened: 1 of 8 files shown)


toniblyx (changes requested):
Can you attach sample screenshots of both GCP and Azure html reports?

MrCloudSec:
> Can you attach sample screenshots of both GCP and Azure html reports?

Done

jfagoagas (changes requested):
Please, rebase this branch with `master` to get the tests fixes.

prowler/lib/outputs/html.py
```diff
@@ -205,9 +159,10 @@ def add_html_header(file_descriptor, audit_info):
     """
         )
     except Exception as error:
-        logger.error(
-            f"{error.__class__.__name__}[{error.__traceback__.tb_lineno}]: {error}"
+        logger.critical(
+            f"{error.__class__.__name__}[{error.__traceback__.tb_lineno}] -- {error}"
         )
+        sys.exit(1)
```
jfagoagas, line 166: I'm not sure about this change, what are we doing for the other output formats?

prowler/providers/common/audit_info.py
```diff
@@ -71,32 +71,32 @@ def print_gcp_credentials(self, audit_info: GCP_Audit_Info):
         # Beautify audited profile, set "default" if there is no profile set
         try:
             getattr(audit_info.credentials, "_service_account_email")
-            profile = (
+            audit_info.profile = (
```
jfagoagas: Why are you changing the input `audit_info` instead of using a new variable? Just curious.

prowler/providers/common/outputs.py
jfagoagas, line 1: I like this one!

toniblyx (dismissed):

jfagoagas (changes requested):

prowler/providers/common/audit_info.py
MrCloudSec: to store it and use it in the future

jfagoagas: But you are not returning it in this function, I'm not happy with the idea of modifying that object to use it in the future.

prowler/lib/outputs/html.py
MrCloudSec: We are using the critical error too, if a file cannot be created is a sys exit.

prowler/providers/common/audit_info.py
MrCloudSec: How we should do it?

jfagoagas: What is the problem leaving it as it was?

MrCloudSec: Because the profile is need it in the HTML

MrCloudSec: I can do the same condition when doing the html

jfagoagas: I still don't understand it, the `profile` used within `print_gcp_credentials` is never returned so the `profile` used in the HTML should be another object with the same name. Am I missing something?

MrCloudSec: Done!

toniblyx (approved):

jfagoagas (approved):
🚀

PR #2184, opened 2023-04-10
feat(check): New CloudTrail check cloudtrail_insights_exist

gabrielsoltz:
### Context

Adding new `cloudtrail` check `cloudtrail_insights_exist`.

CloudTrail Insights provides a powerful way to search and analyze CloudTrail log data using pre-built queries and machine learning algorithms. This can help you to identify potential security threats and suspicious activity in near real-time, such as unauthorized access attempts, policy changes, or resource modifications.

### License

### DIFF
--- a/prowler/providers/aws/services/cloudtrail/cloudtrail_service.py
+++ b/prowler/providers/aws/services/cloudtrail/cloudtrail_service.py
@@ -2,6 +2,7 @@ import threading
 from datetime import datetime
 from typing import Optional
 
+from botocore.client import ClientError
 from pydantic import BaseModel
 
 from prowler.lib.logger import logger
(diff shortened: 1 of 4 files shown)


prowler/providers/aws/services/cloudtrail/cloudtrail_service.py
```diff
@@ -134,6 +136,29 @@ def __get_event_selectors__(self):
...
+                        if insight_selectors:
+                            trail.has_insight_selectors = insight_selectors[0].get(
+                                "InsightType"
+                            )
+
+        except Exception as error:
+            logger.error(
+                f"{client.region} -- {error.__class__.__name__}[{error.__traceback__.tb_lineno}]: {error}"
+            )
+
```
jfagoagas: What do you think will happen if some exception is raised within the try? I mean that maybe we need to duplicate the `try/except` logic or to move it within the `for trail in self.trails:` not to exit that loop if something unexpected happens. I know that we have the same logic for all the services function but we need to give it a thought.

Thanks!

jfagoagas (changes requested):
Awesome @gabrielsoltz 🚀 🚀 🚀 🚀 🚀 

Please check my comments and include the `findings.region` assert in the test. Thanks!

prowler/providers/aws/services/cloudtrail/cloudtrail_service.py
gabrielsoltz: Yes, we could lose the whole function. Duplicating the try/except makes more sense, as I see it, so we follow all the other functions' logic and try/except anything unexpected... I think this will need to be reviewed across the whole code, as I saw it like this everywhere. 

Check this option:

```
    def __get_insight_selectors__(self):
        logger.info("Cloudtrail - Getting trail insihgt selectors...")

        try:
            for trail in self.trails:
                for region, client in self.regional_clients.items():
                    if trail.region == region and trail.name:
                        insight_selectors = None
                        trail.has_insight_selectors = None
                        try:
                            client_insight_selectors = client.get_insight_selectors(
                                TrailName=trail.arn
                            )
                            insight_selectors = client_insight_selectors.get(
                                "InsightSelectors"
                            )
                        except Exception as error:
                            logger.error(
                                f"{client.region} -- {error.__class__.__name__}[{error.__traceback__.tb_lineno}]: {error}"
                            )
                        if insight_selectors:
                            trail.has_insight_selectors = insight_selectors[0].get(
                                "InsightType"
                            )

        except Exception as error:
            logger.error(
            )
```

jfagoagas: I think we would need to include a `continue` within the second `except` to continue the `for trail in self.trails:` loop, right?

jfagoagas: not needed 🚀

jfagoagas (approved):
🔝

PR #2193, opened 2023-04-11
fix(security hub): remove region filter when archiving findings

MrCloudSec:
### Description

Remove region filter when archiving Security Hub findings so when using an aggregated hub, the findings of the different regions are also archived if they are not present in the execution.

### License

### DIFF
--- a/prowler/providers/aws/lib/security_hub/security_hub.py
+++ b/prowler/providers/aws/lib/security_hub/security_hub.py
@@ -99,7 +99,7 @@ def resolve_security_hub_previous_findings(
                 "AwsAccountId": [
                     {"Value": audit_info.audited_account, "Comparison": "EQUALS"}
                 ],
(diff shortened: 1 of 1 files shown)


prowler/providers/aws/lib/security_hub/security_hub.py
```diff
@@ -99,7 +99,6 @@ def resolve_security_hub_previous_findings(
                 "AwsAccountId": [
                     {"Value": audit_info.audited_account, "Comparison": "EQUALS"}
                 ],
-                "Region": [{"Value": region, "Comparison": "EQUALS"}],
```
toniblyx, line 102: Where is then the region?

toniblyx (changes requested):

prowler/providers/aws/lib/security_hub/security_hub.py
MrCloudSec: when defining the boto3 client

PR #2207, opened 2023-04-14
fix(elb-test): Use a mocked current audit info

jfagoagas:
### Description

Fix ELB and ELBv2 test to use a mocked `current_audit_info` object instead of the global one which is causing to some test to fail due to race conditions.

### License


MrCloudSec (approved):
🔝 🔝 🔝 🔝

PR #2218, opened 2023-04-16
feat(check): New VPC checks

gabrielsoltz:
### Context

Adding 3 VPCs Checks:
- `vpc_different_regions`: Ensure there are vpcs in more than one region
- `vpc_subnet_different_az`: Ensure all vpc has subnets in more than one availability zone
- `vpc_subnet_separate_private_public`: Ensure all vpc has public and private subnets defined

I improved the `VPCs` object with the property `subnets` as a list of the new object `VpcSubnet`, and I'm also keeping `self.vpc_subnets` as a part of the class in the case in the future we want to add checks that loop across subnets without checking the VPC they are attached to.

I modified this function name `__describe_route_tables__` to `__describe_peering_route_tables__` because I found it confusing; it is not describing the route tables, but only a subset of them, the ones that are related to peering.  I'm also describing route tables, in my case, for each subnet: I thought maybe I could add another function like that one, something like `__describe_subnet_route_tables__`, but then I found this approach (everything in the same describe_subnets function) more clear, as I'm only describing them here to understand if they are public or not.

### License

### DIFF
--- a/prowler/providers/aws/services/vpc/vpc_different_regions/vpc_different_regions.metadata.json
+++ b/prowler/providers/aws/services/vpc/vpc_different_regions/vpc_different_regions.metadata.json
@@ -15,7 +15,7 @@
   "RelatedUrl": "https://docs.aws.amazon.com/vpc/latest/userguide/VPC_Scenario2.html",
   "Remediation": {
     "Code": {
-      "CLI": "aws ec2 create-subnet",
+      "CLI": "aws ec2 create-vpc",
       "NativeIaC": "",
       "Other": "",
       "Terraform": ""
--- a/prowler/providers/aws/services/vpc/vpc_different_regions/vpc_different_regions.metadata.json
+++ b/prowler/providers/aws/services/vpc/vpc_different_regions/vpc_different_regions.metadata.json
@@ -1,0 +1,32 @@
+{
+  "Provider": "aws",
+  "CheckID": "vpc_different_regions",
+  "CheckTitle": "Ensure there are vpcs in more than one region",
+  "CheckType": [
+    "Infrastructure Security"
+  ],
+  "ServiceName": "vpc",
+  "SubServiceName": "subnet",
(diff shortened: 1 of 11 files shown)


prowler/providers/aws/services/vpc/vpc_different_regions/vpc_different_regions.metadata.json
```diff
@@ -0,0 +1,32 @@
...
+  "SubServiceName": "subnet",
+  "ResourceIdTemplate": "arn:partition:service:region:account-id:resource-id",
+  "Severity": "medium",
+  "ResourceType": "AwsEc2Vpc",
+  "Description": "Ensure there are vpcs in more than one region",
+  "Risk": "",
+  "RelatedUrl": "https://docs.aws.amazon.com/vpc/latest/userguide/VPC_Scenario2.html",
+  "Remediation": {
+    "Code": {
+      "CLI": "aws ec2 create-subnet",
```
MrCloudSec: Would not it be create-vpc?

prowler/providers/aws/services/vpc/vpc_subnet_different_az/vpc_subnet_different_az.py
```diff
@@ -0,0 +1,32 @@
...
+            report.resource_id = vpc.id
+            if vpc.subnets:
+                availability_zone = None
+                for subnet in vpc.subnets:
+                    if (
+                        availability_zone
+                        and subnet.availability_zone != availability_zone
+                    ):
+                        report.status = "PASS"
+                        report.status_extended = f"VPC {vpc.id} has subnets in more than one availbility zone."
```
MrCloudSec:
```suggestion
                        report.status_extended = f"VPC {vpc.id} has subnets in more than one availability zone."
```

MrCloudSec (approved):
Great job!!

PR #2219, opened 2023-04-16
feat(ssmincidents): New Service and Checks

gabrielsoltz:
### Context

New Service `ssmincidents` and check `ssmincidents_enabled_with_plans`:  Ensure SSM Incidents is enabled with response plans.

### License


prowler/providers/aws/services/ssmincidents/ssmincidents_enabled_with_plans/ssmincidents_enabled_with_plans.py
jfagoagas, line 1: You are only checking the first element of the `replication_sets`, what happens if the second element is active?

prowler/providers/aws/services/ssmincidents/ssmincidents_service.py
```diff
@@ -0,0 +1,144 @@
...
+                    if not self.audit_resources or (
+                        is_resource_filtered(replication_set, self.audit_resources)
+                    ):
+                        # The ARN is listed in all regions, but it's only one uniq replication set
+                        self.replication_sets = [
+                            ReplicationSet(
+                                arn=replication_set,
+                            )
+                        ]
+                        break
```
jfagoagas: Why this `break` here?

prowler/providers/aws/services/ssmincidents/ssmincidents_service.py
```diff
@@ -0,0 +1,144 @@
...
+        try:
+            list_replication_sets_paginator = regional_client.get_paginator(
+                "list_replication_sets"
+            )
+            for page in list_replication_sets_paginator.paginate():
+                for replication_set in page["replicationSetArns"]:
+                    if not self.audit_resources or (
+                        is_resource_filtered(replication_set, self.audit_resources)
+                    ):
+                        # The ARN is listed in all regions, but it's only one uniq replication set
```
jfagoagas: Are we calling this function for each region?

prowler/providers/aws/services/ssmincidents/ssmincidents_enabled_with_plans/ssmincidents_enabled_with_plans.py
```diff
@@ -0,0 +1,27 @@
+from prowler.lib.check.models import Check, Check_Report_AWS
+from prowler.providers.aws.services.ssmincidents.ssmincidents_client import (
+    ssmincidents_client,
+)
+
+
+class ssmincidents_enabled_with_plans(Check):
```
jfagoagas, line 7: What is the relation between`ssmincidents_enabled_with_plans` enabled and the SSM replication sets? Just curious.

tests/providers/aws/services/ssmincidents/ssmincidents_service_test.py
```diff
@@ -0,0 +1,135 @@
...
+
+def mock_make_api_call(self, operation_name, kwargs):
+    """We have to mock every AWS API call using Boto3"""
+    if operation_name == "ListReplicationSets":
+        return {"replicationSetArns": [REPLICATION_SET_ARN]}
+    if operation_name == "GetReplicationSet":
+        return {
+            "replicationSet": {
+                "arn": REPLICATION_SET_ARN,
+                "createdBy": "thegaby",
```
jfagoagas:
```suggestion
                "createdBy": "Prowler",
```

jfagoagas (changes requested):
Pretty good 👍 

Please check my comments to merge this PR, thanks!

prowler/providers/aws/services/ssmincidents/ssmincidents_enabled_with_plans/ssmincidents_enabled_with_plans.py
gabrielsoltz: Answering here for both comments:

This is a weird service: I found that `ReplicationSets` are always deployed across all regions (no matter which region you enable the service). It's kinda Global service, but not, it's regional but it replicate to all other regions ALWAYS. The ARN is the same for every region, that's why i only save one in the service: https://github.com/gabrielsoltz/prowler/blob/ssmincidents/prowler/providers/aws/services/ssmincidents/ssmincidents_service.py#L57

Now after seeing your comment, maybe could be better to change https://github.com/gabrielsoltz/prowler/blob/ssmincidents/prowler/providers/aws/services/ssmincidents/ssmincidents_service.py#L57 this and don't loop across them and only save the element 0, but i wanted to be consistent with other checks. That's why I'm breaking the loop if i find one, that already means all the other ones are present (that's how a replication set work, still, very confusing.)

prowler/providers/aws/services/ssmincidents/ssmincidents_service.py
gabrielsoltz: Answered in the other comment.

prowler/providers/aws/services/ssmincidents/ssmincidents_service.py
gabrielsoltz: Yes, this is not a global service, it's regional, the difference is this `ReplicationSets` that is always replicated across regions and share the same ARN.

prowler/providers/aws/services/ssmincidents/ssmincidents_enabled_with_plans/ssmincidents_enabled_with_plans.py
gabrielsoltz: For enabling ssm incidents, the first step is creating replication sets. No replications sets means service is disabled. Not sure if i answered your questions. 
Plans is on top of that, you can enable ssm incidents with or without plans. Here we are checking boths. 
Without plans the service is doing basically nothing.

prowler/providers/aws/services/ssmincidents/ssmincidents_enabled_with_plans/ssmincidents_enabled_with_plans.py
jfagoagas: It's weird also another issue I saw is that you are executing the `break` but there are more for loops. Maybe it's better to talk about this later today, are you available?

prowler/providers/aws/services/ssmincidents/ssmincidents_enabled_with_plans/ssmincidents_enabled_with_plans.py
jfagoagas: Cool, understood!

jfagoagas (approved):

PR #2246, opened 2023-04-20
fix(redshift): correct description in redshift_cluster_automatic_upgrades

rubtoa:
### Context

The description seems to be copied and is not relevant

### Description

The description was "Check for Publicly Accessible Redshift Clusters" which was probably copied from that check

### License


MrCloudSec (approved):
Good catch!

PR #2263, opened 2023-04-24
fix(version): execute check current version function only when `-v`

MrCloudSec:
### Description

Execute check current version function only when `-v` is passed, and handle error in the logger.

### License


jfagoagas (approved):
🚀

PR #2289, opened 2023-04-27
fix(s3): handle NoSuchBucket error

MrCloudSec:
### Description

Handle NoSuchBucket errors in the following functions:

- GetBucketOwnershipControls
- GetBucketTagging
- GetBucketEncryption
- GetPublicAccessBlock

### License

### DIFF
--- a/prowler/providers/aws/services/s3/s3_service.py
+++ b/prowler/providers/aws/services/s3/s3_service.py
@@ -123,6 +123,15 @@ class S3:
             ][
                 "SSEAlgorithm"
             ]
+        except ClientError as error:
+            if error.response["Error"]["Code"] == "NoSuchBucket":
+                logger.warning(
+                    f"{regional_client.region} -- {error.__class__.__name__}[{error.__traceback__.tb_lineno}]: {error}"
+                )
+            else:
+                logger.error(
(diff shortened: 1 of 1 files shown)


prowler/providers/aws/services/s3/s3_service.py
```diff
@@ -285,9 +312,14 @@ def __get_bucket_tagging__(self, bucket):
         except ClientError as error:
             bucket.tags = []
             if error.response["Error"]["Code"] != "NoSuchTagSet":
-                logger.error(
-                    f"{regional_client.region} -- {error.__class__.__name__}[{error.__traceback__.tb_lineno}]: {error}"
-                )
+                if error.response["Error"]["Code"] == "NoSuchBucket":
```
jfagoagas, line 315: should it be at the same level?

jfagoagas (changes requested):

prowler/providers/aws/services/s3/s3_service.py
MrCloudSec: Nope, we don't want to log the exception when the error is `NoSuchTagSet`

jfagoagas: Good one, thanks for the explanation!

jfagoagas (approved):
🚀

PR #2105, opened 2023-03-17
feat(allowlist): add tags filter to allowlist

MrCloudSec:
### Description

Add the Tags filter to allowlist, examples below:

```
Allowlist:
  Accounts:
    "123456789012":
      Checks:
        "*":
          Regions:
            - "*"
          Resources:
            - "test"             
          Tags:
            - "test=test"      
            - "project=test"
# Will ignore every resource containing the string "test" and the tags 'test=test' and 'project=test' in account 123456789012 and every region
    "*":
      Checks:
        "*":
          Regions:
            - "*"
          Resources:
            - "*"
          Tags:
            - "environment=dev" 
# Will ignore every resource containing the tag 'environment=dev' in every account and region
```
### License

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2300, opened 2023-05-02
feat(allowlist): Support regexes in Tags to allow "or"-like conditional matching

kppullin:
### Context

The current allowlisting support for `tags` does not allow for "or" conditions.  Quoting from the original PR for tag support in the allowlist (#2105), it mentions and describes that multiple tags combine into an "and" conditional:

```
Allowlist:
  Accounts:
    "123456789012":
      Checks:
        "*":
          Regions:
            - "*"
          Resources:
            - "test"             
          Tags:
            - "test=test"      
            - "project=test"
# Will ignore every resource containing the string "test" and the tags 'test=test' and 'project=test' in account 123456789012 and every region
```

### Description

This change adds support for defining regular expressions in the `tags` list, which will allow for "or" like conditional matching of tags.  For example, we have multiple security groups in one account that are controlled by external components, with tags such as `GuardDutyManaged=true` in one group and `kubernetes.io/cluster/<CLUSTER_NAME>=owned` in another group.  In the existing code, we can only allow one or the other, but not both.  With this change, we can define the following allow rule to match either set of tags and allow if any of the tags are matched:

```
Allowlist:
  Accounts:
    "123456789012":
      Checks:
        "*":
          Regions:
            - "*"
          Resources:
            - "test"             
          Tags:
            - "(GuardDutyManaged=true|kubernetes.io/cluster/<CLUSTER_NAME>=owned)"
```

If this change looks generally promising, I can follow up with documentation updates.

### License

### DIFF
--- a/docs/tutorials/allowlist.md
+++ b/docs/tutorials/allowlist.md
@@ -28,10 +28,8 @@ You can use `-w`/`--allowlist-file` with the path of your allowlist yaml file, b
               Resources:
                 - "test"
               Tags:
-                # Will ignore every resource containing the string "test" and the tags 'test=test' and
-                # _either_ of ('project=test' OR project=stage) in account 123456789012 and every region
-                - "test=test"
-                - "project=test|project=stage"
+                - "test=test"         # Will ignore every resource containing the string "test" and the tags 'test=test' and
+                - "project=test|project=stage" # either of ('project=test' OR project=stage) in account 123456789012 and every region
 
         "*":
           Checks:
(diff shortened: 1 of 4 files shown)


MrCloudSec:
This is awesome @kppullin !! 
Can you please document this changes in `docs/tutorials/allowlist.md` and add this use case to `prowler/config/allowlist.yaml`?
Thank you!

kppullin-nt:
@sergargar I've made an attempt to improve the docs.  Let me know if this is clearer or not.  Thanks!

MrCloudSec (approved):
Awesome 🚀

PR #2306, opened 2023-05-03
feat(ssmincidents): Use regional_client region instead of audit_profile region

gabrielsoltz:
### Context

```
[File: ssmincidents_service.py:70]      [Module: ssmincidents_service]   ERROR: KeyError:55 -- 'us-east-1'
[File: ssmincidents_service.py:107]     [Module: ssmincidents_service]   ERROR: IndexError:77 -- list index out of range
```

Trying to fix the above error, I think the problem happens when `audit_info.profile_region` is not the same as `list(self.regional_clients.keys())[0]`. 

So i changed it and i made the `__get_replication_set__` only if `__list_replication_sets__` found the replication set. 

### License


MrCloudSec (approved):
Great catch!!

PR #2307, opened 2023-05-03
fix(typo): remove redundant lines

kagahd:
### Context

These two lines of code are redundant and as such should be removed.

### Description

I just removed both redundant code lines.

### License


MrCloudSec (approved):
Good catch!!

PR #2351, opened 2023-05-12
fix(route53): handle empty Records in Zones

MrCloudSec:
### Description

Handle empty Records in Route53 Zones
`KeyError[77]: 'ResourceRecords'`

### License


jfagoagas (approved):
Good catch!

PR #2372, opened 2023-05-18
fix(ssm incidents): check if service available in aws partition

MrCloudSec:
### Description

Check if `ssm-incidentes` service is available in the AWS partition.

### License


jfagoagas (approved):
Good catch!

PR #2375, opened 2023-05-18
fix(allowlist) - `tags` parameter is a string, not a list

kppullin:
### Context

As part of #2300 I had attempted to interpret the `tags` parameter as a list and not a string.  I now realize that the `tags` parameter is a pipe separated string generated via `unroll_tags`.  This reverts the logic back to treating the `tags` as a string.

### License

### DIFF
--- a/prowler/providers/aws/lib/allowlist/allowlist.py
+++ b/prowler/providers/aws/lib/allowlist/allowlist.py
@@ -212,13 +212,12 @@ def is_allowlisted_in_tags(check_allowlist, elem, resource, tags):
             all_allowed_tags_in_resource_tags = True
             for allowed_tag in check_allowlist["Tags"]:
                 found_allowed_tag = False
-                for resource_tag in tags:
-                    if re.search(allowed_tag, resource_tag):
-                        found_allowed_tag = True
-                        break
+                if re.search(allowed_tag, tags):
+                    found_allowed_tag = True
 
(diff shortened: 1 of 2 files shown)


MrCloudSec:
Thanks @kppullin !
Is this interpretation of tags as lists generating any issues? Or is just for having it as before your PR?

kppullin:
@sergargar while I think it'd be cleaner to have a list of tags than a specially formatted string, it is not causing any issues aside from my confusion due to "opaqueness" of the parameter (I had started adding type annotations in this PR, but it touched a lot of code and I didn't know when to stop nor if it should be mixed in with this fix, so I yanked them out).  So this PR just reverts back to how things were prior to my update in #2300.

MrCloudSec:
I like more the way you did it with lists, so if you are agree, we can close this PR.

kppullin:
If you'd prefer using lists, we'd need to update `unroll_tags` to return a `list` instead of a string.  Here's a commit doing just that: https://github.com/prowler-cloud/prowler/commit/9a5c787b777c61d4e6a11d9afe95595f5ebd1b10 .

And I think I was unclear in my comments above - the current code is broken by my prior PR and the allowllist checking for tags fails.  So either this PR or something similar to linked change above will be needed to unbreak the tag allowlisting.

MrCloudSec (approved):
I understand it now, merging this PR to solve the issue.

PR #2393, opened 2023-05-23
feat(gcp): add `--project-ids` flag and scan all projects by default

### LINKED ISSUE #2380: GCP - run checks on all projects in organization
### New feature motivation

Seems like the only way to run checks on all projects inside an organization is to either create or modify the credentials file for each project.
I first thought I would be able to list projects, then for each project to use the "gcloud config set project" command (the core project).

Would appreciate if you can elaborate on this and explain what are the best practices in this scenario.
 

### Solution Proposed

I would suggest a solution similar to the -c flag that takes an array of checks, to have --project flag with an array of project ids, if the user / service account doesn't have the right permissions error would pop up.
Another suggestion is to have another environment variable for project_id. 

Currently, seems like multiple credentials file or modification each iteration is the only solution, correct me if I am wrong.

### Additional context

_No response_

MrCloudSec:
### Description

Add `--project-ids` flag and scan all GCP projects by default.

### License


prowler/providers/gcp/services/compute/compute_default_service_account_in_use/compute_default_service_account_in_use.py
```diff
@@ -7,19 +7,17 @@ def execute(self) -> Check_Report_GCP:
...
         for instance in compute_client.instances:
             report = Check_Report_GCP(self.metadata())
-            report.project_id = compute_client.project_id
+            report.project_id = instance.project_id
             report.resource_id = instance.id
             report.resource_name = instance.name
             report.location = instance.zone
             report.status = "PASS"
             report.status_extended = f"The default service account is not configured to be used with VM Instance {instance.name}"
+            print(instance.service_accounts)
```
jfagoagas:
```suggestion
```

jfagoagas (approved):
🚀

PR #2398, opened 2023-05-23
fix(OSError): handle different OSErrors

kij:
Not all oserrors would be "too many files open". Met "File name too long" today.

### Context

Spent way to much time today trying to figure out why my ulimits weren't registering. 
When in reality,  my path just got too long.

### Description

  [Module: utils]  CRITICAL: Ooops! You reached your user session maximum open files. To solve this issue, increase the shell session limit by running this command `ulimit -n 4096`. For more info visit https://docs.prowler.cloud/en/latest/troubleshooting/

Is not correct in the case of lets say,  a very long file name is generated - which is easily fixed with the `-F` option.

No bug associated, just a quick fix for clearity. The linked issue could of cause be kept.  And then extended with hints to `-F` option on OSError:[36] File name too long.    - But stateing  max open files on every OSError is wrong.
  
### License

### DIFF
--- a/prowler/lib/utils/utils.py
+++ b/prowler/lib/utils/utils.py
@@ -17,9 +17,12 @@ def open_file(input_file: str, mode: str = "r") -> TextIOWrapper:
     try:
         f = open(input_file, mode)
     except OSError as ose:
-        logger.critical(
-            f"{input_file}: OSError:{ose.errno}, {ose.strerror}"
-        )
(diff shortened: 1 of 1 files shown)


MrCloudSec:
Hi @tmonk42, good catch!!
I have added the message only to that type of OSError.

MrCloudSec (approved):

PR #2410, opened 2023-05-25
fix(codebuild): handle FAIL in codebuild_project_user_controlled_buildspec

MrCloudSec:
### Description

Using an user controlled buildspec in a CodeBuild project has to be flagged as FAIL.

### License


jfagoagas (approved):
Good catch!!

PR #2471, opened 2023-06-09
fix(arn validator): include `:` in regex

n4ch04:
### Context

Arn parser was failing when input was a lambda arn like `arn:aws:lambda:eu-west-1:123456789012:function:lambda-function`

### Description

Fix the parser including `:` char in the last capturing group

### License


MrCloudSec (approved):
🔝

PR #2478, opened 2023-06-12
feat(aws): Add MFA flag if it is required by AWS IAM Entity

senyberg:
### Context

Currently you cannot assume roles in AWS if the roles are using MFA

### Description

Add "--mfa" flag which asks user for MFA ARN + Code to authenticate

### Acceptance Criteria

1. Run Prowler with user that requires MFA without --mfa -> Access Denied
2. Run Prowler with user that requires MFA with --mfa-> No errors
3. Run Prowler assuming a role that requires MFA without --mfa -> Access Denied
4. Run Prowler assuming a role that requires MFA with --mfa -> No errors

### License


prowler/providers/aws/aws_provider.py
```diff
@@ -91,11 +91,22 @@ def assume_role(session: session.Session, assumed_role_info: AWS_Assume_Role) ->
...
         # else assume the role without the external id
         else:
-            assumed_credentials = sts_client.assume_role(
-                RoleArn=assumed_role_info.role_arn,
-                RoleSessionName="ProwlerProAsessmentSession",
-                DurationSeconds=assumed_role_info.session_duration,
-            )
+            if assumed_role_info.mfa_enabled:
+                mfa_ARN = input("Enter ARN of MFA: ")
+                mfa_TOTP = input("Enter MFA code: ")
```
jfagoagas: Does it makes sense to use `string.strip()` to remove leading and trailing spaces?

jfagoagas (changes requested):
Hi @senyberg, this is an awesome PR!!

Some comments/questions about this in order to merge it:
- Could you update the Prowler documentation [here](https://docs.prowler.cloud/en/latest/tutorials/aws/role-assumption/) to include this new flag?
- Could you include a new test for the `parser.py` like these https://github.com/prowler-cloud/prowler/blob/master/tests/lib/cli/parser_test.py#L673?

Thanks again for using Prowler 🚀

senyberg:
I added the test and strip() to the input. I have a couple of questions:

- Where do you update the documentation?
- Do you want the flag to be --mfa, or would you like it to be more specific to the role assumption? For example --role-mfa?

tests/lib/cli/parser_test.py
```diff
@@ -3,6 +3,7 @@
 import pytest
 
 from prowler.lib.cli.parser import ProwlerArgumentParser
+from prowler.lib.logger import logger
```
jfagoagas: The linter is raising that this value is unused, please remove it if so.

jfagoagas (changes requested):

jfagoagas:
> I added the test and strip() to the input. I have a couple of questions:

Awesome!

> * Where do you update the documentation?

The tutorial for the Role assumption in Boto3 is here https://github.com/prowler-cloud/prowler/blob/master/docs/tutorials/aws/role-assumption.md

> * Do you want the flag to be --mfa, or would you like it to be more specific to the role assumption? For example --role-mfa?

Good question, we could use `--role-mfa` to be more specific. Please change the references in the code and tests.

Thank you for your contribution!

tests/lib/cli/parser_test.py
senyberg: Yes, forgot it there. I will remove it. Just need to fix a bunch of other tests

jfagoagas: If you need help with the tests tell me. Thanks!

MrCloudSec (approved):

jfagoagas (approved):
🚀 🚀 🚀 🚀 🚀 🚀

PR #2484, opened 2023-06-13
fix(dataevents checks): add trails home region

n4ch04:
### Context

In cloudtrail checks that look for data events the home region of the trail is not included in `status_extended` field

### Description

Include home region in `status_extended` field for those checks to clarify where the trail comes from in case of multiregion trail

### License


jfagoagas (approved):
🚀

PR #2489, opened 2023-06-14
fix(rds checks): test if key exists prior checking it

n4ch04:
### Context

In checks `rds_instance_deletion_protection`& `rds_instance_multi_az` a key from a dict was being referenced without checking its existence

### Description

### License


jfagoagas (approved):
Awesome, good catch!

PR #2495, opened 2023-06-15
chore(ec2): add SG name to resource_details

### LINKED ISSUE #2493: Ability to match security groups by name
### New feature motivation

Security groups are identified in prowler primarily by their sg-xxxxxxxxxx identifier - the only place I see the name being used is in the extended status. The sg identifiers by themselves don't mean anything and requires someone to look them up by name when maintaining allowlists or other mechanisms.

### Solution Proposed

My proposed solution would be to make the security group name available as an additional resource identifier or maybe a virtual Name tag so that we can match against it with regex.

In the mean time I have a script I can run which generates a list of sg identifiers for groups matching a list of regex patterns.

### Additional context

_No response_

MrCloudSec:
### Description

The sg identifiers by themselves don't mean anything and requires someone to look them up by name when maintaining allowlists or other mechanisms.

That is why we need to add the Security Group Resource Details to its name.

### License


jfagoagas (changes requested):
I'm not sure about start using the `report.resource_details` field because is not being used in Prowler.

MrCloudSec:
> I'm not sure about start using the `report.resource_details` field because is not being used in Prowler.

The `resource_details` field appears in every output mode so I do not see any problem.

jfagoagas:
> > I'm not sure about start using the `report.resource_details` field because is not being used in Prowler.
> 
> The `resource_details` field appears in every output mode so I do not see any problem.

Yes, but we are not using it in any place. Let's discuss it later.

jfagoagas (approved):
🚀

PR #2501, opened 2023-06-19
fix(vpc): handle ephemeral VPC endpoint services

n4ch04:
### Context

Handle ephemeral devices in call `describe_vpc_endpoint_service_permissions`

### Description

Raise a warning when the `VpcEndpointServiceId`  already stored does not exists

### License


MrCloudSec (approved):
🔝 🔝 🔝 🔝

PR #2514, opened 2023-06-20
fix(compliance): remove unnecessary Optional attributes

MrCloudSec:
### Description

Remove unnecessary Optional attributes in Compliance models for a correct validation.

### License


jfagoagas (approved):
What's the purpose of this change?

MrCloudSec:
> What's the purpose of this change?

So Pydantic can know which Compliance Model have to choose, if all attributes are optional, Pydantic will choose randomly a model since all of them match.

PR #2527, opened 2023-06-26
feat(allowlist): add exceptions to allowlist

### LINKED ISSUE #2412: Improve Allowlist exceptions
### New feature motivation

One of the long time issues I've had with Prowler is that we use AWS Organizations in our environment so some services, particularly Cloudtrail, are actually enabled at the organization level so I want to ignore a lot of the per-account AWS-CIS checks because all of the cloudtrails, metric filters, etc. are configured and received in our organization master account. However I do want the Prowler checks to run on the organization master account in case they are ever disabled or misconfigured.

The Allowlist functionality allows me to ignore the checks in all of the accounts, however there is no ability to omit the master account from that allowlist entry short of creating an entry for every account that isn't the master - which doesn't scale well when you have > 100 AWS accounts.

### Solution Proposed

I wrote my own report generator that filters the Json output from Prowler. I duplicated the Allowlist functionality in Prowler and then I added exceptions - the internal structure (in Golang) looks like this:

```
type Check struct {
        Regions          []string           `yaml:"Regions"`
        Resources        []*regexp.Regexp   `yaml:"-"`
        Tags             [][]*regexp.Regexp `yaml:"-"`
        ResourcePatterns []string           `yaml:"Resources"`
        TagPatterns      []string           `yaml:"Tags,omitempty"`
        Exceptions       struct {
                AccountIDs       []string           `yaml:"AccountIDs,omitempty"`
                Regions          []string           `yaml:"Regions,omitempty"`
                Resources        []*regexp.Regexp   `yaml:"-"`
                Tags             [][]*regexp.Regexp `yaml:"-"`
                ResourcePatterns []string           `yaml:"Resources"`
                TagPatterns      []string           `yaml:"Tags,omitempty"`
        } `yaml:"Exceptions,omitempty"`
}
```
The corresponding yaml looks like this:

```
        'cloudtrail_s3_dataevents_write_enabled':
          Regions:
            - '*'
          Resources:
            - '*'
          Exceptions:
            AccountIDs:
              - 'ORG_MASTER_ACCOUT_ID'
            Regions:
              - '*'
            Resources:
              - '*'
```

If I find a result with a matching allowlist entry, then I look to see if any of the exceptions negate the match. The logic of the exceptions works exactly like the logic of the Allowlist except I allow a list of account IDs (including the wildcard). For future proofing I might go back and allow an Array of Exceptions.

I wanted to keep to the spirit and feel of what Prowler has today so I based my solution on that concept. If I hire someone who is already familiar with Prowler the allowlist should seem familiar except for the additional exception logic.

### Additional context

Feel free to use the idea! Always been a big Prowler fan, I used to contribute code to v2 and agreed to all the terms, but I don't know Python well enough to contribute code to v3.

MrCloudSec:
### Description

Now, you can add exceptions to your allowlist at the check level, for example:
```
Allowlist:
  Accounts:
    "*":
      Checks:
        "ecs_task_definitions_no_environment_secrets":
          Regions:
            - "*"
          Resources:
            - "*"
          Exceptions:
            Accounts:
              - "0123456789012"
            Regions:
              - "eu-west-1"
              - "eu-south-2"        # Will ignore every resource in check ecs_task_definitions_no_environment_secrets except the ones in account 0123456789012 located in eu-south-2 or eu-west-1

    "123456789012":
      Checks:
        "*":
          Regions:
            - "*"
          Resources:
            - "*"
          Exceptions:
            Resources:
              - "test"
            Tags:
              - "environment=prod"   # Will ignore every resource except in account 123456789012 except the ones containing the string "test" and tag environment=prod
```
### License

### DIFF
--- a/prowler/config/allowlist.yaml
+++ b/prowler/config/allowlist.yaml
@@ -45,6 +45,34 @@ Allowlist:
           Tags:
             - "environment=dev"    # Will ignore every resource containing the tag 'environment=dev' in every account and region
 
+    "*":
+      Checks:
+        "ecs_task_definitions_no_environment_secrets":
+          Regions:
+            - "*"
+          Resources:
+            - "*"
+          Exceptions:
+            Accounts:
+              - "0123456789012"
+            Regions:
+              - "eu-west-1"
+              - "eu-south-2"        # Will ignore every resource in check ecs_task_definitions_no_environment_secrets except the ones in account 0123456789012 located in eu-south-2 or eu-west-1
+
+    "123456789012":
+      Checks:
+        "*":
+          Regions:
+            - "*"
+          Resources:
+            - "*"
+          Exceptions:
+            Resources:
+              - "test"
+            Tags:
+              - "environment=prod"   # Will ignore every resource except in account 123456789012 except the ones containing the string "test" and tag environment=prod
+
+
 
 # EXAMPLE: CONTROL TOWER (to migrate)
 # When using Control Tower, guardrails prevent access to certain protected resources. The allowlist
--- a/prowler/config/allowlist.yaml
+++ b/prowler/config/allowlist.yaml
@@ -2,6 +2,7 @@
 ### Resources and tags are lists that can have either Regex or Keywords.
 ### Tags is an optional list that matches on tuples of 'key=value' and are "ANDed" together.
 ### Use an alternation Regex to match one of multiple tags with "ORed" logic.
+### For each check you can except Accounts, Regions, Resources and/or Tags.
 ###########################  ALLOWLIST EXAMPLE  ###########################
 Allowlist:
   Accounts:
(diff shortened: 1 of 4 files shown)


prowler/config/allowlist.yaml
```diff
@@ -45,6 +45,34 @@ Allowlist:
           Tags:
             - "environment=dev"    # Will ignore every resource containing the tag 'environment=dev' in every account and region
 
+    "*":
```
jfagoagas, line 49: Please update the documentation with this feature.

jfagoagas (changes requested):
Awesome!

prowler/config/allowlist.yaml
MrCloudSec: Done

jfagoagas (approved):
🔝

PR #2538, opened 2023-07-03
feat(lambda service): mapping lambda service to awslambda

n4ch04:
### Context

To launch `lambda` service in Prowler is needed to enter `awslambda` since `lambda` is a reserved word in Python

### Description

Map `lambda` to `awslambda` when passing it as a service in cli 

### License


jfagoagas (approved):
👍

Issue #2576, opened 2023-07-12
[Bug]: Allow list is ignore * if account number exists


capps1994:
### Steps to Reproduce

While running prowler 3.7.0 we have come across an issue that when using an allowlist it is ignoring the * account if you have both an account number and * in the allowlist.yaml.

This is the command that we are running `prowler.py  -w allowlist.yml --compliance cis_1.4_aws cis_1.5_aws`

Below is an example allowlist.yml file that worked in the version 3.6.1 but now does not work in 3.7.0
``` yaml
Allowlist:
  Accounts:
    "123456789": 
      Checks:
        "ec2_securitygroup_allow_ingress_from_internet_to_any_port":
          Regions:
            - "eu-west-1"
          Resources:
            - "sg-test" 
            - "sg-test" 

        "cloudtrail_logs_s3_bucket_access_logging_enabled":
           Regions:
             - "*"
           Resources:
             - "*"

    "*":
      Checks:
        "cloudwatch_log_metric_filter_unauthorized_api_calls":
          Regions:
            - "*"
          Resources:
            - "*"
        "cloudwatch_log_metric_filter_sign_in_without_mfa":
          Regions:
            - "*"
          Resources:
            - "*"
```

### Expected behavior

The expected behavior is that prowler will accept both the account number that it is running against as well as * to indicate all accounts.

### Actual Result with Screenshots or Logs

Here are the results of yaml fail previous stated in prowler version 3.6.1
![image](https://github.com/prowler-cloud/prowler/assets/1837800/ead2ae95-8fd9-4639-bb87-ba10a9101a52)

using the same yaml file in prowler version 3.7.0

![image](https://github.com/prowler-cloud/prowler/assets/1837800/bec4f399-2655-4acc-8d30-07b20ca533cc)

### How did you install Prowler?

Docker (docker pull toniblyx/prowler)

### Environment Resource

1. EC2 Instance

### OS used

Ubuntu

### Prowler version

3.7.0

### Pip version

pip 23.1.2

n4ch04:
Hi @capps1994,

Good catch !! We have solved this here #2577 
Please test it and let us know if everything works as it should

capps1994:
Hi @n4ch04,

Thanks for the quick response, i have just tested your branch and this is now working as expected.

PR #2577, opened 2023-07-12
fix(allowlist): handle wildcard in account field

n4ch04:
### Context

From https://github.com/prowler-cloud/prowler/issues/2576

### Description

It modifies how the allowlisted checks list is formed including those coming from the `"*"`  account entry if present in the allowlist file

### License


jfagoagas (approved):
🚀

PR #2587, opened 2023-07-14
fix(assume_role): Set the AWS STS endpoint region

### LINKED ISSUE #2583: Adding docs regarding Region compatibility of session tokens for the global endpoint (AWS)
### New feature motivation

Prowler is using the global endpoint of STS for assuming roles, and this works fine when the working with regions that are enabled by default. But as stated in AWS [documentation](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_enable-regions.html) You need to change the compatibility of global endpoint to work with all regions and this option is not enabled by default in AWS.

So when using Prowler in regions that are not enabled by default you will get the invalid token error. At least this was the case for me. After changing the settings in IAM, the tokens got valid and I have no errors. 

### Solution Proposed

The easiest option would be to add this fact to the docs. So users can solve the issue by changing the AWS IAM settings.

 Another workaround (IMHO) would be to use specific region endpoint for STS when initiating the client. This solves the problem and there is no need to change the AWS IAM settings.

`sts_client = session.client("sts", aws_region)`

### Additional context

_No response_

jfagoagas:
### Context

Fix #2583 

### Description

Create a new AWS Provider flag `--sts-endpoint-region` to specify the AWS STS region when calling `assume_role` and `get_caller_identity`.

### License


prowler/lib/cli/parser.py
```diff
@@ -8,14 +8,7 @@
...
 )
 from prowler.providers.aws.aws_provider import get_aws_available_regions
-from prowler.providers.aws.lib.arn.arn import is_valid_arn
-
-
-def arn_type(arn: str) -> bool:
-    """arn_type returns a string ARN if it is valid and raises an argparse.ArgumentError if not."""
-    if not is_valid_arn(arn):
-        raise argparse.ArgumentError("Invalid ARN")
-    return arn
```
jfagoagas, line 18: Moved to `prowler/providers/aws/lib/arn/arn.py`

MrCloudSec (approved):
Awesome 🚀

PR #2608, opened 2023-07-24
fix(s3): `__get_object_lock_configuration__` warning logs

jfagoagas:
### Description

Change `error` logs for `warning` in the `__get_object_lock_configuration__` if the `NoSuchBucket` exception is raised.

### License

### DIFF
--- a/prowler/providers/aws/services/s3/s3_service.py
+++ b/prowler/providers/aws/services/s3/s3_service.py
@@ -291,8 +291,19 @@ class S3:
             regional_client.get_object_lock_configuration(Bucket=bucket.name)
             bucket.object_lock = True
         except Exception as error:
-            if "ObjectLockConfigurationNotFoundError" in str(error):
+            if (
+                "ObjectLockConfigurationNotFoundError" in str(error)
+                or error.response["Error"]["Code"] == "NoSuchBucket"
+            ):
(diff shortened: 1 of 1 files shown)


prowler/providers/aws/services/s3/s3_service.py
```diff
@@ -291,8 +291,19 @@ def __get_object_lock_configuration__(self, bucket):
...
+            ):
                 bucket.object_lock = False
+                if regional_client:
+                    logger.warning(
+                        f"{regional_client.region} -- {error.__class__.__name__}[{error.__traceback__.tb_lineno}]: {error}"
+                    )
+                else:
+                    logger.warning(
+                        f"{error.__class__.__name__}[{error.__traceback__.tb_lineno}]: {error}"
+                    )
```
MrCloudSec, line 306: You can avoid repeating this lines by doing it outside the if.

jfagoagas: Yes but we want to raise `error` logs if a different exception is raised. The ones handled are set with `warning`.

MrCloudSec: You're right 👍🏼

MrCloudSec (approved):
🔝

PR #2616, opened 2023-07-25
fix(guardduty): handle disabled detectors in `guardduty_is_enabled`

### LINKED ISSUE #2612: [Bug]: GuardDuty List Detectors Only Gets Enabled Detectors
### Steps to Reproduce

1. Running prowler aws -c guardduty_is_enabled
2. This check is only outputting enabled detectors and is not listing when regions fail with GuardDuty disabled
3. This is because line 45 in guardduty_service.py is looking for DetectorID. However, this field is blank when the detector is not enabled. Therefore the detectors array is being populated with enabled detectors only.

### Expected behavior

This check should output both enabled and disabled GuardDuty detectors

The check is only outputting enabled GuardDuty detectors

### Environment Resource

Workstation

### OS used

MacOS

### Prowler version

3.7

### Pip version

23.1.2

### Context

_No response_

MrCloudSec:
### Description

The check `guardduty_is_enabled` was only outputting enabled detectors and is not listing when regions fail with GuardDuty disabled.

### License

### DIFF
--- a/prowler/providers/aws/services/guardduty/guardduty_service.py
+++ b/prowler/providers/aws/services/guardduty/guardduty_service.py
@@ -40,9 +40,11 @@ class GuardDuty:
     def __list_detectors__(self, regional_client):
         logger.info("GuardDuty - listing detectors...")
         try:
+            detectors = False
             list_detectors_paginator = regional_client.get_paginator("list_detectors")
             for page in list_detectors_paginator.paginate():
                 for detector in page["DetectorIds"]:
+                    detectors = True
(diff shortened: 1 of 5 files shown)


prowler/providers/aws/services/guardduty/guardduty_service.py
```diff
@@ -40,9 +40,11 @@ def __threading_call__(self, call):
     def __list_detectors__(self, regional_client):
         logger.info("GuardDuty - listing detectors...")
         try:
+            detectors = 0
```
jfagoagas: Why a counter here? Use a boolean if you are not using the value.

jfagoagas (changes requested):

jfagoagas (approved):
🚀

PR #2638, opened 2023-07-31
feat(aws): New AWSService class as parent

jfagoagas:
### Context

We needed a parent class for each AWS Service not to repeat code and to have a centralised point of initialising an AWS service class.

Thanks @Fennerr for the idea based on https://github.com/prowler-cloud/prowler/pull/2108. We are working on to test and integrate that code into Prowler but we need to break it down into a series of steps to achieve it.

### Description

New `AWSService` class as parent for each AWS service class.
- `__threading_call__` and `get_session` methods has been moved to the `AWSService` class.
- Fix some typos
- Include the `Audit_Metadata` in each `set_mocked_audit_info`.

### License

### DIFF
--- a/prowler/providers/aws/lib/service/service.py
+++ b/prowler/providers/aws/lib/service/service.py
@@ -1,3 +1,5 @@
+import threading
+
 from prowler.providers.aws.aws_provider import (
     generate_regional_clients,
     get_default_region,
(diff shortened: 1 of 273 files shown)


MrCloudSec (approved):
🔝 🔝 🔝 🔝 🔝

PR #2641, opened 2023-07-31
feat(gcp_service): Parent class

jfagoagas:
### Context

We needed a parent class for each GCP Service not to repeat code and to have a centralised point of initialising an GCP service class.

### Description

New `GCPService` class as parent for each GCP service class.

### License


MrCloudSec (approved):
Awesome!

Issue #2438, opened 2023-06-04
[Bug]: Problem with AWS Organizations Cloudtrail


zfLQ2qx2:
### Steps to Reproduce

I had this same issue with Prowler2 years ago but it looks like its back in Prowler3. My employer uses AWS Organizations which is an API to manage groups of accounts, and in addition to consolidated billing you get the ability to manage some services at the organization level (i.e. cross-account and cross-region). One of those services is Cloudtrail. When you create an organization level cloudtrail it will appear in every account and region so you know it exists but most of the CIS checks, for example, will fail because the Cloudwatch logs groups and/or S3 buckets being used by the organization trail usually exist in another account. So now I am running checks in the organization master account and they are all failing. In particular these checks:

```
cloudtrail_s3_dataevents_read_enabled
cloudtrail_s3_dataevents_write_enabled
cloudwatch_changes_to_network_acls_alarm_configured
cloudwatch_changes_to_network_gateways_alarm_configured
cloudwatch_changes_to_network_route_tables_alarm_configured
cloudwatch_changes_to_vpcs_alarm_configured
cloudwatch_log_metric_filter_and_alarm_for_aws_config_configuration_changes_enabled
cloudwatch_log_metric_filter_and_alarm_for_cloudtrail_configuration_changes_enabled
cloudwatch_log_metric_filter_authentication_failures
cloudwatch_log_metric_filter_aws_organizations_changes
cloudwatch_log_metric_filter_disable_or_scheduled_deletion_of_kms_cmk
cloudwatch_log_metric_filter_for_s3_bucket_policy_changes
cloudwatch_log_metric_filter_policy_changes
cloudwatch_log_metric_filter_root_usage
cloudwatch_log_metric_filter_security_group_changes
cloudwatch_log_metric_filter_sign_in_without_mfa
cloudwatch_log_metric_filter_unauthorized_api_calls
```

And because it is every single check I suspect there is a common source like someone discarding the organization (aka shadow) trails because the prowler results indicate no trails at all when in fact there is the one organizational trail.

For example: 

```
prowler aws --role ${ROLEARN} --verbose --output-modes json --checks cloudtrail_s3_dataevents_read_enabled --log-level INFO --session-duration 3600
```

Gives the output:

```
Check ID: cloudtrail_s3_dataevents_read_enabled - cloudtrail [low]
	FAIL us-east-1: No CloudTrail trails have a data event to record all S3 object-level API operations.
```

However:

```
[...]

PR #2643, opened 2023-07-31
fix(cloudtrail): Set status to INFO when trail is outside the audited account

jfagoagas:
### Description

Testing https://github.com/prowler-cloud/prowler/issues/2438 I realized that we are raising `PASS` or `FAIL` when the bucket trail is a cross-account bucket in another account out of Prowler's permissions scope, so we cannot check anything against it.

Due to the fact that we state `please check it manually` in the `status_extended`, I think we need to raise an `INFO` finding.

### License


MrCloudSec (approved):
Good catch!

PR #2656, opened 2023-08-02
Improve GuardDuty is Enabled Check

scrapbird:
Adding check for region with no detectors

### Context

The guard duty is enabled check will not fail when a region has no detector at all.

### Description

This adds a simple check at the end of the guard duty check which goes through every region and checks if it didn't appear in the list of findings. If not it will create a new finding to indicate this region has no detector.

### License

### DIFF
--- a/prowler/providers/aws/services/guardduty/guardduty_is_enabled/guardduty_is_enabled.py
+++ b/prowler/providers/aws/services/guardduty/guardduty_is_enabled/guardduty_is_enabled.py
@@ -1,5 +1,6 @@
 from prowler.lib.check.models import Check, Check_Report_AWS
 from prowler.providers.aws.services.guardduty.guardduty_client import guardduty_client
+from prowler.providers.aws import aws_provider
 
 
 class guardduty_is_enabled(Check):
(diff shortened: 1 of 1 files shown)


jfagoagas (changes requested):
Hi @scrapbird thanks for your contribution but I don't see the point of putting that code within the check. I think we fixed that issue in this PR that was merged last week https://github.com/prowler-cloud/prowler/pull/2616.

Please check that PR since now the check reports four different states:
- Detector enabled
- Detector not enabled --> I think this is your use case
- Detector enabled but not configured
- Detector enabled but suspended

Thanks!

MrCloudSec:
Hi @scrapbird, this check will fail when a region has no detector if you have permissions to list the detectors in that region. In the regions that you have an Access Denied will not appear.

scrapbird:
> Hi @scrapbird thanks for your contribution but I don't see the point of putting that code within the check. I think we fixed that issue in this PR that was merged last week #2616.
> 
> Please check that PR since now the check reports four different states:
> 
> * Detector enabled
> * Detector not enabled --> I think this is your use case
> * Detector enabled but not configured
> * Detector enabled but suspended
> 
> Thanks!

Oh cool, I didn't see that PR I've been using this change in my local copy for a while and figured I should push it upstream :) Glad to know there is a better fix for it, I will close this PR.

PR #2645, opened 2023-07-31
feat(aws): Add 6 new AWS checks

jit-contrib:
### Context
Adding 6 new AWS rules

### Description
This PR adds 6 new AWS rules alongside with respective unit tests.

List of new added rules:

* cloudtrail_management_exist_with_multi_region_enabled
* dlm_ebs_snapshot_policy_exists
* documentdb_instance_storage_encrypted
* ec2_ebs_snapshots_exists
* secretsmanager_check_if_service_is_in_use
* ssm_check_if_secure_strings_is_in_use

### License
By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2675, opened 2023-08-03
feat(aws): New CloudTrail, DLM, DocumentDB, EC2, Account and Support checks

jit-contrib:
### Context

Adding 6 new AWS rules

### Description

This PR implement the fixes suggested in https://github.com/prowler-cloud/prowler/pull/2645.

List of new added rules:

- `account_maintain_different_contact_details_to_security_billing_and_operations`
- `cloudtrail_multi_region_enabled_logging_management_events`
- `dlm_ebs_snapshot_lifecycle_policy_exists`
- `ec2_ebs_volume_snapshots_exists`
- `documentdb_instance_storage_encrypted`
- `trustedadvisor_premium_support_plan_subscribed`

### License


prowler/providers/aws/services/ec2/ec2_ebs_snapshots_exists/ec2_ebs_snapshots_exists.py
```diff
@@ -0,0 +1,21 @@
...
+        volumes_with_snapshots = [snapshot.volume for snapshot in ec2_client.snapshots]
+        for volume in ec2_client.volumes:
+            report = Check_Report_AWS(self.metadata())
+            report.status = "FAIL"
+            report.region = volume.region
+            report.resource_id = volume.id
+            report.resource_arn = volume.arn
+            report.resource_tags = volume.tags
+            report.status_extended = f"Snapshots not found to EBS volume {volume.id}"
+            if volume.id in volumes_with_snapshots:
```
jfagoagas: In large environments this could be an expensive operation. I think that at the service level you can:
A. Link the Volume with their snapshots so just checking `if len(volume.snapshots) > 0`
B. Create a map of Volumes and Snapshots like `volume_snapshots = {volume.id : [Snapshot1, Snapshot2]`

With the above approach a lookup will be `O(1)` because you have the key to look for.

jfagoagas (changes requested):

prowler/providers/aws/services/account/account_maintain_different_contact_details_to_security_billing_and_operations/account_maintain_different_contact_details_to_security_billing_and_operations.metadata.json
```diff
@@ -0,0 +1,32 @@
...
+  "SubServiceName": "",
+  "ResourceIdTemplate": "arn:partition:access-recorder:region:account-id:recorder/resource-id",
+  "Severity": "medium",
+  "ResourceType": "Other",
+  "Description": "Maintain current contact details.",
+  "Risk": "Ensure contact email and telephone details for AWS accounts are current and map to more than one individual in your organization. An AWS account supports a number of contact details; and AWS will use these to contact the account owner if activity judged to be in breach of Acceptable Use Policy. If an AWS account is observed to be behaving in a prohibited or suspicious manner; AWS will attempt to contact the account owner by email and phone using the contact details listed. If this is unsuccessful and the account behavior needs urgent mitigation; proactive measures may be taken; including throttling of traffic between the account exhibiting suspicious behavior and the AWS API endpoints and the Internet. This will result in impaired service to and from the account in question.",
+  "RelatedUrl": "",
+  "Remediation": {
+    "Code": {
+      "CLI": "No command available.",
```
jfagoagas:
```suggestion
      "CLI": "",
```

prowler/providers/aws/services/account/account_maintain_different_contact_details_to_security_billing_and_operations/account_maintain_different_contact_details_to_security_billing_and_operations.metadata.json
```diff
@@ -0,0 +1,32 @@
+{
+  "Provider": "aws",
+  "CheckID": "account_maintain_different_contact_details_to_security_billing_and_operations",
+  "CheckTitle": "Maintain current contact details.",
```
jfagoagas:
```suggestion
  "CheckTitle": "Maintain different contact details to security, billing and operations.",
```

prowler/providers/aws/services/account/account_maintain_different_contact_details_to_security_billing_and_operations/account_maintain_different_contact_details_to_security_billing_and_operations.metadata.json
```diff
@@ -0,0 +1,32 @@
...
+  "CheckType": [
+    "IAM"
+  ],
+  "ServiceName": "account",
+  "SubServiceName": "",
+  "ResourceIdTemplate": "arn:partition:access-recorder:region:account-id:recorder/resource-id",
+  "Severity": "medium",
+  "ResourceType": "Other",
+  "Description": "Maintain current contact details.",
```
jfagoagas:
```suggestion
  "Description": "Maintain different contact details to security, billing and operations.",
```

poetry.lock
jfagoagas, line 1: Please, make sure that you don't push changes for the poetry.lock since it's managed automatically by our Github Actions.

prowler/providers/aws/services/account/account_maintain_different_contact_details_to_security_billing_and_operations/account_maintain_different_contact_details_to_security_billing_and_operations.py
```diff
@@ -0,0 +1,48 @@
+from prowler.lib.check.models import Check, Check_Report_AWS
+from prowler.providers.aws.services.account.account_client import account_client
+
+# This check has no findings since it is manual
```
jfagoagas:
```suggestion
```

prowler/providers/aws/services/account/account_maintain_different_contact_details_to_security_billing_and_operations/account_maintain_different_contact_details_to_security_billing_and_operations.py
```diff
@@ -0,0 +1,48 @@
...
+            security_contact = account_client.client.get_alternate_contact(
+                AlternateContactType="SECURITY"
+            )
+            operations_contact = account_client.client.get_alternate_contact(
+                AlternateContactType="OPERATIONS"
+            )
+        except Exception:
+            billing_contact = None
+            security_contact = None
+            operations_contact = None
```
jfagoagas: This business logic cannot be in the check's code, please move it to the `account` service.

prowler/providers/aws/services/cloudtrail/cloudtrail_management_exist_with_multi_region_enabled/cloudtrail_management_exist_with_multi_region_enabled.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+class cloudtrail_management_exist_with_multi_region_enabled(Check):
+    def execute(self):
+        findings = []
+        report = Check_Report_AWS(self.metadata())
+        report.status = "FAIL"
+        report.status_extended = (
+            "No trail found with multi-region enabled and logging management events"
+        )
```
jfagoagas: In this report block the following fields are missing:
```
report.region = trail.region
report.resource_id = trail.name
report.resource_arn = trail.arn
report.resource_tags = trail.tags
```

prowler/providers/aws/services/cloudtrail/cloudtrail_management_exist_with_multi_region_enabled/cloudtrail_management_exist_with_multi_region_enabled.py
```diff
@@ -0,0 +1,44 @@
...
+                            report.resource_tags = trail.tags
+                            report.status = "PASS"
+                            report.status_extended = f"Trail {trail.name} is multi-region, is logging and have management events logs enabled"
+                            if report.resource_id not in [
+                                f.resource_id for f in findings
+                            ]:
+                                findings.append(report)
+                                break
+        if len(findings) == 0:
+            findings.append(report)
```
jfagoagas: I think this block can be simplified with a boolean variable to break the loop.

prowler/providers/aws/services/dlm/dlm_ebs_snapshot_policy_exists/dlm_ebs_snapshot_policy_exists.metadata.json
jfagoagas, line 1: I'd change policies here to `lifecycle policies`, WDYT?

prowler/providers/aws/services/cloudtrail/cloudtrail_management_exist_with_multi_region_enabled/cloudtrail_management_exist_with_multi_region_enabled.metadata.json
```diff
@@ -0,0 +1,34 @@
...
+  "ResourceType": "AwsCloudTrailTrail",
+  "Description": "Ensure CloudTrail logging management events in All Regions",
+  "Risk": "AWS CloudTrail enables governance, compliance, operational auditing, and risk auditing of your AWS account. To meet FTR requirements, you must have management events enabled for all AWS accounts and in all regions and aggregate these logs into an Amazon Simple Storage Service (Amazon S3) bucket owned by a separate AWS account.",
+  "RelatedUrl": "",
+  "Remediation": {
+    "Code": {
+      "CLI": "",
+      "NativeIaC": "",
+      "Other": "",
+      "Terraform": ""
```
jfagoagas: We have to fill this fields, please check the other cloudtrail trails checks.

prowler/providers/aws/services/dlm/dlm_client.py
```diff
@@ -0,0 +1,4 @@
+from prowler.providers.aws.lib.audit_info.audit_info import current_audit_info
+from prowler.providers.aws.services.dlm.dlm_service import Dlm
+
+dlm_client = Dlm(current_audit_info)
```
jfagoagas:
```suggestion
dlm_client = DLM(current_audit_info)
```

prowler/providers/aws/services/dlm/dlm_ebs_snapshot_policy_exists/dlm_ebs_snapshot_policy_exists.metadata.json
```diff
@@ -0,0 +1,34 @@
...
+    "Code": {
+      "CLI": "",
+      "NativeIaC": "",
+      "Other": "",
+      "Terraform": ""
+    },
+    "Recommendation": {
+      "Text": "",
+      "Url": ""
+    }
```
jfagoagas: Did you check these fields?

https://www.trendmicro.com/cloudoneconformity/knowledge-base/aws/DLM/ebs-snapshot-automation.html

prowler/providers/aws/services/dlm/dlm_ebs_snapshot_policy_exists/dlm_ebs_snapshot_policy_exists.py
jfagoagas, line 1: I'd change `policies` here to `lifecycle policies`, WDYT?

prowler/providers/aws/services/dlm/dlm_service.py
```diff
@@ -0,0 +1,40 @@
+import threading
+
+from prowler.lib.logger import logger
+from prowler.providers.aws.aws_provider import generate_regional_clients
+
+
+################## Config
```
jfagoagas:
```suggestion
################## Data Lifecycle Manager
```

prowler/providers/aws/services/dlm/dlm_service.py
```diff
@@ -0,0 +1,40 @@
...
+    def __init__(self, audit_info):
+        self.service = "dlm"
+        self.session = audit_info.audit_session
+        self.audited_account = audit_info.audited_account
+        self.audited_partition = audit_info.audited_partition
+        self.audited_account_arn = audit_info.audited_account_arn
+        self.audit_resources = audit_info.audit_resources
+        self.regional_clients = generate_regional_clients(self.service, audit_info)
+        self.lifecycle_policies = []
+        self.__threading_call__(self.__get_lifecycle_policies__)
```
jfagoagas: All this fields have been moved to a new `AWSService` class here https://github.com/prowler-cloud/prowler/pull/2638

```suggestion
class DLM(AWSService):
    def __init__(self, audit_info):
        # Call AWSService's __init__
        super().__init__(__class__.__name__, audit_info)
        self.lifecycle_policies = []
        self.__threading_call__(self.__get_lifecycle_policies__)
```

prowler/providers/aws/services/dlm/dlm_service.py
```diff
@@ -0,0 +1,40 @@
...
+        return self.session
+
+    def __threading_call__(self, call):
+        threads = []
+        for regional_client in self.regional_clients.values():
+            threads.append(threading.Thread(target=call, args=(regional_client,)))
+        for t in threads:
+            t.start()
+        for t in threads:
+            t.join()
```
jfagoagas: Moved to the `AWSService` class too.

```suggestion
```

prowler/providers/aws/services/documentdb/documentdb_service.py
```diff
@@ -0,0 +1,89 @@
...
+class DocumentDB:
+    def __init__(self, audit_info):
+        self.service = "docdb"
+        self.session = audit_info.audit_session
+        self.audited_account = audit_info.audited_account
+        self.audit_resources = audit_info.audit_resources
+        self.audited_partition = audit_info.audited_partition
+        self.regional_clients = generate_regional_clients(self.service, audit_info)
+        self.db_instances = []
+        self.__threading_call__(self.__describe_db_instances__)
```
jfagoagas: All this fields have been moved to a new `AWSService` class here https://github.com/prowler-cloud/prowler/pull/2638

```suggestion
class DocumentDB(AWSService):
    def __init__(self, audit_info):
        # Call AWSService's __init__
        super().__init__("docdb", audit_info)
        self.db_instances = []
        self.__threading_call__(self.__describe_db_instances__)
```

prowler/providers/aws/services/documentdb/documentdb_service.py
```diff
@@ -0,0 +1,89 @@
...
+        return self.session
+
+    def __threading_call__(self, call):
+        threads = []
+        for regional_client in self.regional_clients.values():
+            threads.append(threading.Thread(target=call, args=(regional_client,)))
+        for t in threads:
+            t.start()
+        for t in threads:
+            t.join()
```
jfagoagas:
```suggestion
```

prowler/providers/aws/services/documentdb/documentdb_service.py
```diff
@@ -0,0 +1,89 @@
...
+from typing import Optional
+
+from pydantic import BaseModel
+
+from prowler.lib.logger import logger
+from prowler.providers.aws.aws_provider import generate_regional_clients
+
+
+################## DocumentDB
+class DocumentDB:
```
jfagoagas: I think you can use the RDS service since it returns the same information and then import it in the check. So there is no need for a new `DocumentDB` service.

prowler/providers/aws/services/documentdb/documentdb_instance_storage_encrypted/documentdb_instance_storage_encrypted.metadata.json
```diff
@@ -0,0 +1,30 @@
+{
+  "Provider": "aws",
+  "CheckID": "documentdb_instance_storage_encrypted",
+  "CheckTitle": "Check if DocumentDB instances storage is encrypted.",
+  "CheckType": [],
+  "ServiceName": "documentdb",
+  "SubServiceName": "",
+  "ResourceIdTemplate": "",
```
jfagoagas: Please fill the ARN model for this resource.

prowler/providers/aws/services/documentdb/documentdb_instance_storage_encrypted/documentdb_instance_storage_encrypted.py
```diff
@@ -0,0 +1,28 @@
...
+)
+
+
+class documentdb_instance_storage_encrypted(Check):
+    def execute(self):
+        findings = []
+        for db_instance in documentdb_client.db_instances:
+            report = Check_Report_AWS(self.metadata())
+            report.region = db_instance.region
+            report.resource_id = db_instance.id
```
jfagoagas, line 13: Include the `report.resource_arn` please.

prowler/providers/aws/services/ec2/ec2_ebs_snapshots_exists/ec2_ebs_snapshots_exists.metadata.json
```diff
@@ -0,0 +1,34 @@
...
+    "Data Protection"
+  ],
+  "ServiceName": "ec2",
+  "SubServiceName": "snapshot",
+  "ResourceIdTemplate": "arn:partition:service:region:account-id:resource-id",
+  "Severity": "medium",
+  "ResourceType": "AwsEc2Snapshot",
+  "Description": "Check if EBS snapshots exists.",
+  "Risk": "",
+  "RelatedUrl": "",
```
jfagoagas: Please fill these fields.

prowler/providers/aws/services/ec2/ec2_ebs_snapshots_exists/ec2_ebs_snapshots_exists.metadata.json
```diff
@@ -0,0 +1,34 @@
...
+  "Remediation": {
+    "Code": {
+      "CLI": "",
+      "NativeIaC": "",
+      "Other": "https://www.trendmicro.com/cloudoneconformity-staging/knowledge-base/aws/EBS/ebs-volumes-recent-snapshots.html",
+      "Terraform": ""
+    },
+    "Recommendation": {
+      "Text": "Creating point-in-time EBS snapshots periodically will allow you to handle efficiently your data recovery process in the event of a failure, to save your data before shutting down an EC2 instance, to back up data for geographical expansion and to maintain your disaster recovery stack up to date.",
+      "Url": ""
```
jfagoagas: Please fill this field.

prowler/providers/aws/services/documentdb/documentdb_instance_storage_encrypted/documentdb_instance_storage_encrypted.metadata.json
```diff
@@ -0,0 +1,30 @@
...
+  "Severity": "medium",
+  "ResourceType": "AwsDocDbInstance",
+  "Description": "Check if DocumentDB instances storage is encrypted.",
+  "Risk": "If not enabled sensitive information at rest is not protected.",
+  "RelatedUrl": "",
+  "Remediation": {
+    "Code": {
+      "CLI": "aws docdb create-db-cluster --db-cluster-identifier <db_cluster_id> --port 27017 --engine docdb --master-username <yourMasterUsername> --master-user-password <yourMasterPassword> --storage-encrypted",
+      "NativeIaC": "",
+      "Other": "",
```
jfagoagas:
```suggestion
      "Other": "https://www.trendmicro.com/cloudoneconformity-staging/knowledge-base/aws/DocumentDB/encryption-enabled.html",
```

prowler/providers/aws/services/documentdb/documentdb_instance_storage_encrypted/documentdb_instance_storage_encrypted.metadata.json
```diff
@@ -0,0 +1,30 @@
...
+      "Terraform": "https://docs.bridgecrew.io/docs/bc_aws_general_28#terraform"
+    },
+    "Recommendation": {
+      "Text": "Enable Encryption. Use a CMK where possible. It will provide additional management and privacy benefits.",
+      "Url": ""
```
jfagoagas: Include an URI from the AWS documentation here if possible.

prowler/providers/aws/services/documentdb/documentdb_instance_storage_encrypted/documentdb_instance_storage_encrypted.metadata.json
```diff
@@ -0,0 +1,30 @@
...
+  "CheckTitle": "Check if DocumentDB instances storage is encrypted.",
+  "CheckType": [],
+  "ServiceName": "documentdb",
+  "SubServiceName": "",
+  "ResourceIdTemplate": "",
+  "Severity": "medium",
+  "ResourceType": "AwsDocDbInstance",
+  "Description": "Check if DocumentDB instances storage is encrypted.",
+  "Risk": "If not enabled sensitive information at rest is not protected.",
+  "RelatedUrl": "",
```
jfagoagas: Include an URI from the AWS documentation here if possible.

prowler/providers/aws/services/documentdb/documentdb_service.py
```diff
@@ -0,0 +1,89 @@
...
+                        )
+                    )
+        except Exception as error:
+            logger.error(
+                f"{regional_client.region} -- {error.__class__.__name__}[{error.__traceback__.tb_lineno}]: {error}"
+            )
+
+
+class DBInstance(BaseModel):
+    id: str
```
jfagoagas, line 76: Please include the ARN here using https://docs.aws.amazon.com/documentdb/latest/developerguide/documentdb-arns.html

prowler/providers/aws/services/ec2/ec2_service.py
```diff
@@ -185,8 +186,11 @@ def __describe_snapshots__(self, regional_client):
                                 region=regional_client.region,
                                 encrypted=encrypted,
                                 tags=snapshot.get("Tags"),
+                                volume=snapshot["VolumeId"],
                             )
                         )
+                        if self.volumes_with_snapshots.get(snapshot["VolumeId"], True):
+                            self.volumes_with_snapshots[snapshot["VolumeId"]] = True
```
jfagoagas: What is happening here? I don't get it, since you are setting to true something that is already true.

prowler/providers/aws/services/ec2/ec2_service.py
```diff
@@ -418,6 +422,7 @@ class Snapshot(BaseModel):
     encrypted: bool
     public: bool = False
     tags: Optional[list] = []
+    volume: str
```
jfagoagas: Is this field always present?

prowler/providers/aws/services/support/support_plan_subscribed/support_plan_subscribed.metadata.json
jfagoagas, line 1: Please fill all the missing fields in the metadata file.

prowler/providers/aws/services/support/support_service.py
jfagoagas, line 1: I think you can merge this in the TrustedAdvisor service which is using the `support` service underneath.

prowler/providers/aws/services/support/support_plan_subscribed/support_plan_subscribed.py
```diff
@@ -0,0 +1,18 @@
+from prowler.lib.check.models import Check, Check_Report_AWS
+from prowler.providers.aws.services.support.support_client import support_client
+
+
+class support_plan_subscribed(Check):
+    def execute(self):
+        findings = []
+        report = Check_Report_AWS(self.metadata())
```
jfagoagas: It is a must to include the following report fields:
- resource_id
- resource_arn
- region could not be `global` since it's retrieved from the service init for the global services.

prowler/providers/aws/services/support/support_plan_subscribed/support_plan_subscribed.py
jfagoagas, line 1: I think this check needs to use a configuration variable to set a list of AWS accounts since it is only recommended for the production accounts and we cannot raise a FAIL for each account scanned if it is not a production account.

tests/providers/aws/services/documentdb/documentdb_instance_storage_encrypted/documentdb_instance_storage_encrypted_test.py
jfagoagas, line 1: Why are you using magic mock instead of the `mock_rds`?

prowler/providers/aws/services/dlm/dlm_ebs_snapshot_policy_exists/dlm_ebs_snapshot_policy_exists.py
```diff
@@ -0,0 +1,15 @@
+from prowler.lib.check.models import Check, Check_Report_AWS
+from prowler.providers.aws.services.dlm.dlm_client import dlm_client
+
+
+class dlm_ebs_snapshot_policy_exists(Check):
+    def execute(self):
+        findings = []
+        report = Check_Report_AWS(self.metadata())
+        report.status = "FAIL"
+        report.status_extended = "No EBS snapshot policies found."
```
jfagoagas: `resource_arn`, `resource_id`, `region` and `tags` fields are missing.

prowler/providers/aws/services/cloudtrail/cloudtrail_management_exist_with_multi_region_enabled/cloudtrail_management_exist_with_multi_region_enabled.py
jfagoagas, line 1: This check is not working as expected, in our demo environments it is raising a `FAIL` finding while having a multi-region CloudTrail logging all the management events. Please check this out.

prowler/providers/aws/services/support/support_service.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+################## Support
+class Support(AWSService):
+    def __init__(self, audit_info):
+        # Call AWSService's __init__
+        super().__init__(__class__.__name__, audit_info)
+        self.support_services = []
+        self.__threading_call__(self.__describe_services__)
+        print(self.support_services)
```
jfagoagas:
```suggestion
```

prowler/providers/aws/services/support/support_service.py
```diff
@@ -0,0 +1,47 @@
...
+                    error.response.get("Error", {}).get("Code")
+                    == "SubscriptionRequiredException"
+                ):
+                    self.support_services.append(
+                        SupportServices(
+                            region=regional_client.region, premium_support=False
+                        )
+                    )
+                else:
+                    raise error
```
jfagoagas:
```suggestion
                    logger.error(
                 f"{regional_client.region} --"
                 f" {error.__class__.__name__}[{error.__traceback__.tb_lineno}]:"
                 f" {error}"
             )
```

jfagoagas (changes requested):
I've left a lot of comments / changes since some parts of this PR are not coded following our guidelines:
- Some services are duplicated since `documentdb` (like `neptune`) API points to the `rds` API underneath.
- The `support` service is created within the `trustedadvisor` service so there is no need to have another service.
- Regarding the `support` check we have to include a list of accounts using the configuration file because we cannot raise a FAIL in each Prowler scan if the account has not AWS Premium Support configured.

I recommend you to follow the Prowler Developer Guide here https://docs.prowler.cloud/en/latest/developer-guide/introduction/ and we can talk more about this using the Slack.

prowler/providers/aws/services/support/support_service.py
jit-contrib: In this case I'm catching an error inside a try...catch block already, the raise is to raise other errors that can happen other than the one I want to catch, and be catched and logged by the outer block. I think that the behavior will be the same, the difference is just duplicating the code to log the error instead of reuse this part of the code of the outer try...catch block. Do you prefer me to apply your change? Feel free to do it

prowler/providers/aws/services/cloudtrail/cloudtrail_management_exist_with_multi_region_enabled/cloudtrail_management_exist_with_multi_region_enabled.py
jit-contrib: Can you give more details on your idea here? I didn't get what you are suggesting

jfagoagas: I think you want to mimic the logic like the `cloudtrail_multi_region_enabled` check don't? 

What is the idea behind this check? Do you want to raise a `PASS/FAIL` per region? Or a global `PASS/FAIL`?

jit-contrib: The idea here is to raise a global one.

prowler/providers/aws/services/cloudtrail/cloudtrail_management_exist_with_multi_region_enabled/cloudtrail_management_exist_with_multi_region_enabled.py
jit-contrib: Regarding this one, do make sense to include the trail.region on the report considering the check is a global one?

jfagoagas: Even making the finding global you have to set all the attributes above, you can use the `cloudtrail_client.region` which holds the default region for the customer based on the following logic: https://github.com/prowler-cloud/prowler/blob/081a7ead4c072e4692367c45f0eaee270af5fc55/prowler/providers/aws/aws_provider.py#L285

prowler/providers/aws/services/cloudtrail/cloudtrail_management_exist_with_multi_region_enabled/cloudtrail_management_exist_with_multi_region_enabled.py
jfagoagas: I think it should be great to talk about it in the Slack to see if this is the best way to do it, I don't want to make that decision by my own.

prowler/providers/aws/services/documentdb/documentdb_service.py
jit-contrib: As suggested before, I removed the documentdb_service and imported the RDS service directly in check. Due to that change, this comment don't make sense anymore.

prowler/providers/aws/services/ec2/ec2_service.py
jit-contrib: The value can be None, in that case I'm defaulting it do "True" in the dict.get() method, and change the value to True in the case it is None

prowler/providers/aws/services/dlm/dlm_ebs_snapshot_policy_exists/dlm_ebs_snapshot_policy_exists.py
jit-contrib: This rule is verifying if at least one EBS snapshots lifecycle exists into the AWS account, is not checking if every volume are at least covered by one policy. Considering the check is doing makes no sense to include the data you requested in the report. Maybe can make sense to change the rule to link every volume to the snapshot lifecycle polices by it's tags, and in that case include the volume arn, id, region, and tags in the results.

jfagoagas: Hi @jit-contrib, regardless the purpose of the check, all checks must fill the `resource_arn`, `resource_id`, `region` and `tags` (if possible). Please refer to the Developer Guide here https://docs.prowler.cloud/en/latest/developer-guide/checks/#resource-id-name-and-arn because if the resource audited is the AWS account these fields will be as follows:
- `resource_arn` --> AWS Account Root ARN, like `arn:aws:iam::123456789012:root`
- `resource_id` --> AWS Account number, like `123456789012`
- `region` --> The `service.region` object.


Please, review the developer guide and give us feedback is something is not present or clear.

Thanks!

prowler/providers/aws/services/account/account_maintain_different_contact_details_to_security_billing_and_operations/account_maintain_different_contact_details_to_security_billing_and_operations.py
jfagoagas: It is this comment addressed yet?

prowler/providers/aws/services/ec2/ec2_ebs_snapshots_exists/ec2_ebs_snapshots_exists.py
jit-contrib: Already updated

prowler/providers/aws/services/support/support_plan_subscribed/support_plan_subscribed.py
jit-contrib: Can you please give an example on that? Do you mean pass a flag when you call prowler command to include or not this specific check?

prowler/providers/aws/services/ec2/ec2_service.py
jit-contrib: Only in the new I created, I'll change it to Optional, nice spot!

tests/providers/aws/services/documentdb/documentdb_instance_storage_encrypted/documentdb_instance_storage_encrypted_test.py
jit-contrib: At time I built the rule I wasn't aware that documentdb service is using rds API, I used magic mock since I didn't find documentdb on moto. Should I make the change to use mock_rds instead?

poetry.lock
jfagoagas, line 1: Please, remove this file from the PR. The `poetry.lock` is automatically managed by the github actions.

jfagoagas: I think you have to remove it since it's already there.

prowler/providers/aws/services/cloudtrail/cloudtrail_management_exist_with_multi_region_enabled/cloudtrail_management_exist_with_multi_region_enabled.py
jfagoagas: This comment is still pending.

prowler/providers/aws/services/dlm/dlm_ebs_snapshot_lifecycle_policy_exists/dlm_ebs_snapshot_lifecycle_policy_exists.metadata.json
```diff
@@ -0,0 +1,34 @@
...
+  "SubServiceName": "ebs",
+  "ResourceIdTemplate": "",
+  "Severity": "medium",
+  "ResourceType": "Other",
+  "Description": "Ensure EBS Snapshot lifecycle policies are defined.",
+  "Risk": "",
+  "RelatedUrl": "",
+  "Remediation": {
+    "Code": {
+      "CLI": "aws dlm create-lifecycle-policy --region <region> --execution-role-arn <execution-role-arn> --description <decription> --state ENABLED --policy-details file://<policy filenanem>.json",
```
jfagoagas:
```suggestion
      "CLI": "aws dlm create-lifecycle-policy --region <region> --execution-role-arn <execution-role-arn> --description <description> --state ENABLED --policy-details file://<policy filename>.json",
```

prowler/providers/aws/services/dlm/dlm_ebs_snapshot_lifecycle_policy_exists/dlm_ebs_snapshot_lifecycle_policy_exists.metadata.json
```diff
@@ -0,0 +1,34 @@
...
+      "NativeIaC": "",
+      "Other": "",
+      "Terraform": ""
+    },
+    "Recommendation": {
+      "Text": "",
```
jfagoagas: The `Recommendation.Text` cannot be empty, please fill it with the risk.

prowler/providers/aws/services/dlm/dlm_ebs_snapshot_lifecycle_policy_exists/dlm_ebs_snapshot_lifecycle_policy_exists.metadata.json
```diff
@@ -0,0 +1,34 @@
...
+    "Data Protection"
+  ],
+  "ServiceName": "dlm",
+  "SubServiceName": "ebs",
+  "ResourceIdTemplate": "",
+  "Severity": "medium",
+  "ResourceType": "Other",
+  "Description": "Ensure EBS Snapshot lifecycle policies are defined.",
+  "Risk": "",
+  "RelatedUrl": "",
```
jfagoagas: This two fields cannot be empty.

prowler/providers/aws/services/dlm/dlm_ebs_snapshot_lifecycle_policy_exists/dlm_ebs_snapshot_lifecycle_policy_exists.metadata.json
```diff
@@ -0,0 +1,34 @@
+{
+  "Provider": "aws",
+  "CheckID": "dlm_ebs_snapshot_lifecycle_policy_exists",
+  "CheckTitle": "Ensure EBS Snapshot lifecycle policies are defined.",
+  "CheckType": [
+    "Data Protection"
+  ],
+  "ServiceName": "dlm",
+  "SubServiceName": "ebs",
+  "ResourceIdTemplate": "",
```
jfagoagas: Can you include here the format of the resource ARN?

prowler/providers/aws/services/dlm/dlm_ebs_snapshot_lifecycle_policy_exists/dlm_ebs_snapshot_lifecycle_policy_exists.py
```diff
@@ -0,0 +1,18 @@
...
+
+class dlm_ebs_snapshot_lifecycle_policy_exists(Check):
+    def execute(self):
+        findings = []
+        report = Check_Report_AWS(self.metadata())
+        report.status = "FAIL"
+        report.status_extended = "No EBS snapshot lifecycle policies found."
+        report.resource_id = dlm_client.audited_account
+        report.resource_arn = dlm_client.audited_account_arn
+        report.region = "global"
```
jfagoagas: This value is not valid, you should use the resource region or the client region instead.

prowler/providers/aws/services/documentdb/documentdb_service.py
```diff
@@ -0,0 +1,72 @@
...
+        # Call AWSService's __init__
+        super().__init__("docdb", audit_info)
+        self.db_instances = []
+        self.__threading_call__(self.__describe_db_instances__)
+
+    def __describe_db_instances__(self, regional_client):
+        logger.info("RDS - Describe Instances...")
+        try:
+            describe_db_instances_paginator = regional_client.get_paginator(
+                "describe_db_instances"
```
jfagoagas, line 23: Here you have to filter only for the `docdb` engine instances.
```
response = client.describe_db_instances(
    Filters=[
        {
            'Name': 'engine',
            'Values': [
                'docdb',
            ]
        },
    ],
)
```

prowler/providers/aws/services/documentdb/documentdb_instance_storage_encrypted/documentdb_instance_storage_encrypted.py
```diff
@@ -0,0 +1,29 @@
...
+            report.resource_tags = db_instance.tags
+            report.resource_arn = documentdb_client.audited_account_arn
```
jfagoagas: Why are you setting here the AWS account ARN instead of the resource ARN?

prowler/providers/aws/services/documentdb/documentdb_service.py
jfagoagas: Since the `documentdb` service is present here include the ARN for the instance.

jfagoagas (changes requested):
I think the following tests are missing:
- `documentdb` service
- `account` check
- `support` check

Please review the above comments from the past revision since some of them are not addressed yet. Thanks!

tests/providers/aws/services/documentdb/documentdb_instance_storage_encrypted/documentdb_instance_storage_encrypted_test.py
jfagoagas: You can try but I'm not sure if this will work, if not you can use the `MagicMock`.

poetry.lock
jit-contrib: I just updated the file with the one present in master branch. Do you think there is a better strategy to fix this one?

jfagoagas: I think it's better to remove the file from the commit.

jit-contrib: I noticed you already answered this question in another comment.

prowler/providers/aws/services/dlm/dlm_ebs_snapshot_lifecycle_policy_exists/dlm_ebs_snapshot_lifecycle_policy_exists.py
jit-contrib: The rule is verifying is there is lifecycle policies defined in all regions, if one of the regions have these kind of policies defined the rule is return PASS result, in that case makes no sense to include one region in the report. What maybe can make sense is to divide the results per region, instead ok making a global verification. Let me check if it's OK to go that way. In the case the rule requirement doesn't allow me to do that, what should I assign to region in global verifications?

prowler/providers/aws/services/dlm/dlm_ebs_snapshot_lifecycle_policy_exists/dlm_ebs_snapshot_lifecycle_policy_exists.metadata.json
jit-contrib: This API is not returning any resource ARN

prowler/providers/aws/services/cloudtrail/cloudtrail_management_exist_with_multi_region_enabled/cloudtrail_management_exist_with_multi_region_enabled.py
jit-contrib: I already added this information to the results in the last push.

prowler/providers/aws/services/documentdb/documentdb_instance_storage_encrypted/documentdb_instance_storage_encrypted.py
jit-contrib: Due to the results being a global one, not based on each resource, check this comment I made in other one related with the same rule https://github.com/prowler-cloud/prowler/pull/2675#discussion_r1350463465

prowler/providers/aws/services/dlm/dlm_ebs_snapshot_lifecycle_policy_exists/dlm_ebs_snapshot_lifecycle_policy_exists.py
jfagoagas: In this case, having just one global finding, you must use `dlm_client.region`. `global` is not a valid region within the AWS provider.

prowler/providers/aws/services/account/account_service.py
jit-contrib, line 1: Create a contact object with all the contacts instead of using a list of Nones in the case the contacts are not set

jfagoagas (approved):

jfagoagas (dismissed):
May thanks @jit-contrib for this huge contribution to keep improving and growing Prowler 🚀 🚀 🚀 🚀

jfagoagas (approved):

PR #2679, opened 2023-08-07
fix(config): Pass a configuration file using `--config-file config.yaml`

### LINKED ISSUE #2602: CLI option to set the path to config file
### New feature motivation

Configuration file `config.yaml` can be read by the tool only if it's placed in the `prowler/config/config.yaml`. But if the prowler is installed by pip which is installed by the pyenv (e.g) it's difficult to find the place to put a custom config file because it will be like this `.pyenv/versions/3.9.13/lib/python3.9/site-packages/prowler/config/`

### Solution Proposed

I proposed 3 options:
- to add a cli argument which will tell the prowler from where to take the config file like `--config ./config.yaml` 
- try to find config file in the current working directory with fallback to default location
- both of 2 above

just hardcode the path to destination folder in the pipeline

### Additional context

_No response_

jfagoagas:
### Context

Currently Prowler only uses the configuration file from the default location. We need to have the ability of pass it from a custom location.

### Description

- Create the `--config-file` flag
- Update documentation
- New way to handle the configuration file
- Update tests

### License

### DIFF
--- a/docs/tutorials/configuration_file.md
+++ b/docs/tutorials/configuration_file.md
@@ -37,6 +37,7 @@ The following list includes all the checks with configurable variables that can 
 ## GCP
 
 ## Config YAML File Structure
+> This is the new Prowler configuration file format. The old one without provider keys is still compatible just for the AWS provider.
 ```yaml
 # AWS Configuration
 aws:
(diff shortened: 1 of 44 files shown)


MrCloudSec (dismissed):
This is awesome!!!

MrCloudSec (approved):

PR #2683, opened 2023-08-07
feat(checks): dump all checks as a json file

jchrisfarris:
### Context

This PR adds a CLI option to dump all the available checks into a json file that can be consumed by the `--checks-file` flag. 

### Description

```bash
prowler aws -jb > my_checks.json
# user edits my_checks.json to remove the ones they don't want
prowler aws --checks-file my_checks.json
```

### License

### DIFF
--- a/prowler/__main__.py
+++ b/prowler/__main__.py
@@ -113,6 +113,12 @@ def prowler():
         provider,
     )
 
+    if args.list_checks_json:
+        import json
+        output = {provider: sorted(checks_to_execute)}
+        print(json.dumps(output, indent=2, default=str))
+        sys.exit()
+
     # If -l/--list-checks passed as argument, print checks to execute and quit
     if args.list_checks:
         print_checks(provider, sorted(checks_to_execute), bulk_checks_metadata)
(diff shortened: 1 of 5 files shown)


prowler/__main__.py
```diff
@@ -113,6 +113,12 @@ def prowler():
         provider,
     )
 
+    if args.list_checks_json:
```
jfagoagas, line 118: Could you include this within a function in the `check` lib?

I think you can reutilize the code there.

prowler/lib/cli/parser.py
```diff
@@ -242,6 +242,9 @@ def __init_list_checks_parser__(self):
         list_group.add_argument(
             "-l", "--list-checks", action="store_true", help="List checks"
         )
+        list_group.add_argument(
```
jfagoagas, line 247: For new flags and arguments we are not including short options for the flags, but we can leave it if you find it useful.

We have to include a test for this new option.

jfagoagas (changes requested):
Thanks @jchrisfarris, this is great but please check the comments.

prowler/lib/cli/parser.py
jchrisfarris: Short code isn't needed. We can remove. I'll look at the other request when I circle back to this.

jfagoagas (approved):
Awesome, thanks @jchrisfarris !

PR #2690, opened 2023-08-08
fix(ec2-securitygroups): Handle IPv6 public

jfagoagas:
### Description

We are not properly checking IPv6 CIDRs when auditing security groups and a public IPv6 is marked as public even with the `any_address` is set to `True`.

### License


MrCloudSec (dismissed):
Good catch!!

MrCloudSec (approved):

PR #2694, opened 2023-08-10
fix(iam): password policy expiration

jfagoagas:
### Description

Fix IAM password policy expiration since it wasn't read from the right value.

### License


MrCloudSec (approved):
🔝

PR #2715, opened 2023-08-11
fix(s3_bucket_policy_public_write_access): look at account and bucket-level public access block settings

jchrisfarris:
…to reduce false positivies

### Context

The s3_bucket_policy_public_write_access check will report a false positive if the bucket policy has statements that allow public write, but the policy is neutered by a Block Public Access setting. 

### Description

If the Account or Bucket has `restrict_public_buckets` set, then the bucket automatically passes this check

### License


MrCloudSec (approved):
Great catch!!

PR #2718, opened 2023-08-14
test(python): 3.9, 3.10, 3.11

jfagoagas:
### Description

We have to check if Prowler runs with Python 3.9, 3.10 and 3.11, so we are running the CI with these 3 versions in parallel.

### License


toniblyx (approved):
LGTM

PR #2757, opened 2023-08-21
feat(s3): Add S3 KMS encryption check

singergs:
### Context

Adding a new encryption check for KMS

### Description

The following is a check to determine if KMS  encryption is enabled on an S3 bucket. 

Starting January 5, 2023, all new object uploads to Amazon S3 are automatically encrypted at no additional cost and with no impact on performance.  The following check s3_bucket_default_encryption should always return a pass for newly created S3 buckets.

Another check was created to test if S3 buckets have additional encryption mechanism enabled.
### License


jfagoagas (approved):
That's great @singergs !!

I've included some fixes and tests that were missing, for the next time you can use the https://docs.prowler.cloud/en/latest/developer-guide/introduction/ to help you creating new checks!

Thanks for using Prowler 🚀

PR #2764, opened 2023-08-23
fix(get_checks_from_input_arn): fix logic and add tests

n4ch04:
### Context

`get_checks_from_input_arn` function was not fully covered by tests and was not working properly, since it did not add checks if there were not include the subservice of the resource arn

### Description

Add tests for all the services and sub-services included in the logic and fix the general case

### License


jfagoagas (approved):
🔝

PR #2766, opened 2023-08-23
fix(get_regions_from_audit_resources): fix logic and add tests

n4ch04:
### Context

Function `get_regions_from_audit_resources` was not being fully covered by test, and also logic was not correct 

### Description

Fix logic  when no regions are returned and add the remaining tests

### License


jfagoagas (approved):
👍

PR #2786, opened 2023-08-29
feat(json-asff): adds AWS resource tags in json-asff and SecurityHub findings

sbldevnet:
### Context

This PR includes AWS Resource Tags, if any, in the JSON-ASFF file and Security Hub findings.

### Description

ASFF format includes _Tags_ attribute under _Resources_, but this is not passed to Security Hub finding.
Tags is an optional attribute but Security Hub requires at least 1 property, so passing an empty dict will cause an error:
```
{"timestamp": "2023-08-28 18:36:20,074", "filename": "security_hub.py:46", "level": "ERROR", "module": "security_hub", "message": "Failed to send archived findings to AWS Security Hub -- InvalidInput -- Finding does not adhere to Amazon Finding Format. data.Resources[0].Tags should NOT have fewer than 1 properties."}
```
To avoid this, if the tags array is empty, None will be returned.

On the Pydantic validation side, if the _finding_output_ includes `Tags: null` it will cause an error even if the model field is optional:
```
{"timestamp": "2023-08-28 18:43:10,510", "filename": "security_hub.py:51", "level": "ERROR", "module": "security_hub", "message": "ParamValidationError -- [41]:Parameter validation failed:
Invalid type for parameter Findings[0].Resources[0].Tags, value: None, type: <class 'NoneType'>, valid types: <class 'dict'> in region us-east-1"}
```
Adding `exclude_none=True` into Pydatic dict will exclude that value and fixes the issue.

The if/else condition is due to the difference in the _finding.resource_tags_ content. The values can be:
- None (e.g from ec2 ami or ec2 ebs)
- [] (e.g ec2 security group without tags)
- [{}] (e.g access analyzer)
- [{'my_tag_key1':'my_tag_value1', 'my_tag_key2':'my_tag_value2'}] (e.g awslambda)
- [{'Key':'my_tag_key1', 'Value':'my_tag_value1'}, {'Key':'my_tag_key2', 'Value':'my_tag_value2'}] (e.g ec2 securitygroup)

This is an example on how the _Resources_ attribute looks with _Tags_:
```json
"Resources": [
    {
        "Type": "AwsEc2SecurityGroup",
        "Id": "arn:aws:ec2:us-east-1:123456789012:security-group/sg-000000000001aaaa1",
        "Partition": "aws",
        "Region": "us-east-1",
        "Tags": {
            "team": "security",
            "provider": "terraform"
        }
    }
],
```

### License


toniblyx:
Hi Samuel! Good to see your code here :D 
We will give it a look and get back to you soon, thank you!

prowler/lib/outputs/json.py
```diff
@@ -63,12 +63,25 @@ def fill_json_asff(finding_output, audit_info, finding, output_options):
...
+        if finding.resource_tags:
+            for tag in finding.resource_tags:
+                if "Key" in tag and "Value" in tag:
+                    resource_tags[tag["Key"]] = tag["Value"]
+                else:
+                    resource_tags.update(tag)
+            if len(resource_tags) == 0:
+                resource_tags = None
+        else:
+            resource_tags = None
```
jfagoagas: Could you please include this piece of code in a separate function? It'd be great if you could include test for this new function.

prowler/lib/outputs/models.py
```diff
@@ -680,6 +680,7 @@ class Resource(BaseModel):
     Id: str
     Partition: str
     Region: str
+    Tags: Optional[dict]
```
jfagoagas, line 686: 🏅

jfagoagas (changes requested):
That's great @sbldevnet !!

Please review my comments, thanks!!

prowler/lib/outputs/json.py
sbldevnet: It should be ready. Let me know if we can improve it

```
❯ poetry run pytest tests/lib/outputs/outputs_test.py::Test_Outputs
============================= test session starts =============================
platform darwin -- Python 3.11.5, pytest-7.4.2, pluggy-1.0.0
Using --randomly-seed=912372200
rootdir: /prowler
configfile: pyproject.toml
plugins: randomly-3.15.0, cov-4.1.0, xdist-3.3.1
collected 26 items

tests/lib/outputs/outputs_test.py ..........................                        [100%]

============================= 26 passed in 0.42s ==============================
```

jfagoagas: Awesome!! 🚀

jfagoagas (dismissed):
Awesome @sbldevnet!!

Thanks for rebasing your changes with the latest SecurityHub refactor and for making tests 🥇 

🚀

jfagoagas (approved):
🚀 🚀 🚀 🚀

PR #2789, opened 2023-08-30
fix(is_valid_arn): include . into resource name

n4ch04:
### Context

When using the flag `--resource-arn` to filter resources that has to be scanned, if the resource name included a `.`
it was returning the error `Invalid ARN ....`

### Description

Include the `.` character into the arn regexp in `is_valid_arn` function

### License


toniblyx (approved):
LGTM

PR #2790, opened 2023-08-30
fix(html): unroll regions set prior concat

n4ch04:
### Context

When launching Prowler with the flag `--resource-arn` the `html` file was not being generated because it was trying to concat a `set` of regions to `str` fields

### Description

Unroll the `set` of regions in `get_aws_html_assessment_summary` and transform it into `str`

### License


toniblyx (approved):
LGTM

PR #2813, opened 2023-09-05
feat(unix timestamp): add unix timestamp to outputs

### LINKED ISSUE #2805: Ability to write pure unix timestamp to output files
### New feature motivation

Currently, the file output (for example, a json file) contains the "AssessmentStartTime" field, which has already been converted taking into account the clock time of the running environment.
In a large organization with different time zones and different SIEMs, we want to have a universal metric.

### Solution Proposed

Add the key `--unixtimestamp`, which will write unix timestamp instead of the converted time

This is a simple addition, no alternatives

### Additional context

In real life, we encountered the fact that SIEM from one region does not correctly parse the “AssessmentStartTime” field and thus events fall into a future period, which for SIEM has not yet arrived

n4ch04:
### Context

Fixes #2805 

### Description

Allows the usage of unix timestamps in outputs instead of iso format.
For example:

```
./prowler.py  --unix-timestamp
```
### License

### DIFF
--- a/prowler/lib/cli/parser.py
+++ b/prowler/lib/cli/parser.py
@@ -159,6 +159,12 @@ Detailed documentation at https://docs.prowler.cloud
             action="store_true",
             help="Send a summary of the execution with a Slack APP in your channel. Environment variables SLACK_API_TOKEN and SLACK_CHANNEL_ID are required (see more in https://docs.prowler.cloud/en/latest/tutorials/integrations/#slack).",
         )
+        common_outputs_parser.add_argument(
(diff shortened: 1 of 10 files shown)


ramses999:
@n4ch04 I tested it on a real account and uploaded it to SIEM, everything works correctly, thanks for doing it quickly!

I will be looking forward to the merge into the main branch and the new version of https://hub.docker.com/r/toniblyx/prowler to use in the Kubernetes environment )

MrCloudSec (approved):
🔝 🔝 🔝

PR #2820, opened 2023-09-07
fix(iam_inline_policy_no_administrative_privileges): set resource id as the entity name

MrCloudSec:
### Description

Set resource id as the entity name.

### License

### DIFF
--- a/prowler/providers/aws/services/iam/iam_inline_policy_no_administrative_privileges/iam_inline_policy_no_administrative_privileges.py
+++ b/prowler/providers/aws/services/iam/iam_inline_policy_no_administrative_privileges/iam_inline_policy_no_administrative_privileges.py
@@ -10,7 +10,7 @@ class iam_inline_policy_no_administrative_privileges(Check):
                 report = Check_Report_AWS(self.metadata())
                 report.region = iam_client.region
                 report.resource_arn = policy.arn
-                report.resource_id = policy.name
+                report.resource_id = policy.entity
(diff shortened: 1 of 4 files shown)


toniblyx (dismissed):
LGTM

prowler/providers/aws/services/iam/iam_inline_policy_no_administrative_privileges/iam_inline_policy_no_administrative_privileges.py
```diff
@@ -10,7 +10,7 @@ def execute(self) -> Check_Report_AWS:
                 report = Check_Report_AWS(self.metadata())
                 report.region = iam_client.region
                 report.resource_arn = policy.arn
-                report.resource_id = policy.name
+                report.resource_id = policy.entity
```
jfagoagas, line 13: Why this change? Please include an explanation in the PR body.

Thanks!

MrCloudSec: Because the inline policy it is not an unique resource and it may lead to duplicate finding IDs.

jfagoagas: I didn't see the "inline" part, totally agree then. Good catch!!

jfagoagas (approved):
🔝

PR #2822, opened 2023-09-11
feat(iam): add new check iam_role_administratoraccess_policy

kagahd:
### Context

The existing check `iam_aws_attached_policy_no_administrative_privileges` was not appropriate for us because it only lists the resource that the check was running on, in this case all `"*:*"` open policies such as the AWS managed policy  `AdministratorAccess`, and not for every role where such a policy is attached to.
However, we have some company-wide rules in place, to which the `AdministratorAccess` policy is attached for good reasons. In order to be able to allowlist these roles, I wrote the new check `iam_role_administratoraccess_policy`. 

### Description

The new check `iam_role_administratoraccess_policy` fails for roles where the `AdministratorAccess` policy is attached to. The check  reports the role name so you may allowlist specific roles.

### License


MrCloudSec (approved):
Awesome PR! love the check and tests 🚀

PR #2825, opened 2023-09-11
fix(eks_endpoints_not_publicly_accessible): handle endpoint private access

Fennerr:
### Context
This check would create false-negatives when the EKS API endpoint was set to "Public and private" and the Public access source allowlist was set to "0.0.0.0/0(open to all traffic)"

This change fixed this issue

### Description

An EKS cluster could have cluster.endpoint_public_access and cluster.endpoint_private_access set to True. This would result in the AND bool operator failing, and the check would pass, when it should fail.

### License


jfagoagas (approved):
Thanks @Fennerr for all the contributions!!

I've included some missing asserts in the tests.

PR #2847, opened 2023-09-18
fix(iam): findings of some checks may have been lost

kagahd:
### Context

The following checks did overwrite the first finding (access key 1) by the second finding (access key 2) in case that both access keys were not used, respectively not rotated, within the defined number of days.

- `iam_disable_30_days_credentials`
- `iam_disable_45_days_credentials`
- `iam_disable_90_days_credentials`
- `iam_rotate_access_key_90_days`

### Description

I fixed it so that **both** findings will be reported if **both** access keys were not used, respectively not rotated. That's also the way the corresponding prowler v2 checks and the similar prowler v3 check `iam_user_no_setup_initial_access_key` are working. 
I added unit tests for these modified checks as well.

Last but not least, I added a similar unit test for check `iam_user_no_setup_initial_access_key`. The check itself did already correctly report both findings if both access keys have never been used. However, it was not covered by  unit tests until now.

### License


MrCloudSec (approved):
Awesome!! Great catch @kagahd, thanks for the tests!

PR #2848, opened 2023-09-18
fix(sqs_queues_not_publicly_accessible): Improve status extended

Fennerr:
### Context

Came across some SQS queues that failed the check, but had conditions that used the `ArnEquals` and `aws:SourceArn` keys, but the resources were a list of SNS topics in another account.

This change is to make it more clear that there is a condition that may be limiting access, but the check has still failed because it does not restrict access to resources within the same account

### License


jfagoagas (approved):
Thanks @Fennerr, I've made a little change and included a new test. 🚀

Issue #2841, opened 2023-09-13
Change where Prowler integration check for AWS Security Hub is called


scottbward:
### New feature motivation

When a customer is importing findings into AWS Security Hub a call is made to confirm that the prowler integration is enabled in Security Hub.    The current implementation of the security hub integration is making this confirmation call every time a finding is sent to Security Hub.   This means that if during one prowler report run a customer has 1000 findings, this call to confirm the integration is enabled is also called 1000 times when only needs to be called once before sending the first finding as part of a report run.  

### Solution Proposed

The current check to see if the integration is enabled is called in this library: https://github.com/prowler-cloud/prowler/blob/bc11537350e1e919eff4a02d4d3fcd3e109c6d11/prowler/providers/aws/lib/security_hub/security_hub.py#L32C1-L33C1.    I think that the right place to put this check is right before the call to send a finding to security hub: https://github.com/prowler-cloud/prowler/blob/bc11537350e1e919eff4a02d4d3fcd3e109c6d11/prowler/lib/outputs/outputs.py#L111.  Logic here should have a variable to track that the integration is confirmed as enabled and then use the state of that variable to only check that the integration is enabled for the first finding and then skip the check for all additional findings.  

### Describe alternatives you've considered

No alternatives considered.

jfagoagas:
Thanks for reporting it @scottbward, I'll take a look this week. Thanks!

jfagoagas:
Hi @scottbward, the above PR improves the behaviour of the Security Hub integration, checking just one time per region audited if the service is enabled and the integration configured.

Thanks for using Prowler 🚀

PR #2868, opened 2023-09-26
refactor(security_hub): Send findings in batches

### LINKED ISSUE #2840: Add batching for sending AWS Security Hub findings
### New feature motivation

When customers with a large volume of resources run prowler to report on their resources and import the results into Security Hub they are running into API throttling rate limits which is impacting the ability for customers to get all their Prowler findings into Security Hub in a timely manner.  The main cause of this is that findings are sent to Security Hub one by one instead of batches.   The BatchImportFindings API for Security Hub allows up to 100 findings to be batched together in a single request.  

The feature ask is that the integration with AWS Security Hub be updated so that findings are collected and sent to Security Hub in batches.  Batch sizes of 100 are the max that can be sent but even an non-optimal batching strategy could help in significantly reducing the number of calls that are made to BatchImportFindings.  

Here is the API limits that are published for Security Hub: https://docs.aws.amazon.com/securityhub/1.0/APIReference/Welcome.html

### Solution Proposed

From what I can tell the sending of findings to Security Hub is initiated here: https://github.com/prowler-cloud/prowler/blob/bc11537350e1e919eff4a02d4d3fcd3e109c6d11/prowler/lib/outputs/outputs.py#L116.   

What I believe would need to happen is that some logic would need to be added before the send_to_security_hub call is made where findings are collected into a batch and then when a batch limit is reached, or there are no more findings to report on, the current batch of findings is sent to Security Hub.  

No other alternatives considered.  

### Additional context

_No response_

jfagoagas:
### Context

Fixes #2840 and #2841

Right now the AWS Security Hub integration was strongly tied to the `report` method in such a way that each finding is sent to security hub which causes tons of calls to that API instead of using batches.

Also, we are checking if the service is enabled and configured each time we send a finding.

### Description

This PR improves de AWS Security Hub integration sending the findings in batches of 100 findings (https://docs.aws.amazon.com/securityhub/1.0/APIReference/Welcome.html) the same way we do when archiving findings. We've included a new function to group findings per region to be sent to Security Hub if the service is enabled and configured in that region.

### License

### DIFF
--- a/prowler/__main__.py
+++ b/prowler/__main__.py
@@ -4,6 +4,8 @@
 import os
 import sys
 
+from colorama import Style
+
 from prowler.lib.banner import print_banner
 from prowler.lib.check.check import (
     bulk_load_checks_metadata,
@@ -230,6 +232,9 @@ def prowler():
 
     # AWS Security Hub Integration
     if provider == "aws" and args.security_hub:
+        print(
+            f"{Style.BRIGHT}\nSending findings to AWS Security Hub...{Style.RESET_ALL}"
+        )
         # Verify where AWS Security Hub is enabled
         aws_security_enabled_regions = []
         for region in audit_info.audited_regions:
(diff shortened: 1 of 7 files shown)


MrCloudSec (dismissed):
🚀 🚀 🚀

prowler/__main__.py
```diff
@@ -225,14 +231,52 @@ def prowler():
...
-            audit_output_options.output_filename,
-            audit_info,
+    # AWS Security Hub Integration
+    if provider == "aws" and args.security_hub:
+        print(
+            f"{Style.BRIGHT}\nSending findings to AWS Security Hub, please wait...{Style.RESET_ALL}"
+        )
+        # Verify where AWS Security Hub is enabled
+        aws_security_enabled_regions = []
+        shub_regions = (
```
jfagoagas:
```suggestion
        security_hub_regions = (
```

prowler/__main__.py
```diff
@@ -225,14 +231,52 @@ def prowler():
...
+            f"{Style.BRIGHT}\nSending findings to AWS Security Hub, please wait...{Style.RESET_ALL}"
+        )
+        # Verify where AWS Security Hub is enabled
+        aws_security_enabled_regions = []
+        security_hub_regions = (
+            get_available_aws_service_regions("securityhub", audit_info)
+            if not audit_info.audited_regions
+            else audit_info.audited_regions
+        )
+        for region in shub_regions:
```
jfagoagas:
```suggestion
        for region in security_hub_regions:
```

MrCloudSec (approved):

PR #2881, opened 2023-10-02
fix(azure): Typo in SQL check

JackStuart:
### Context

Simple fix to update a typo noticed when running Prowler against an Azure SQL Server

### Description

Simple fix to update a typo noticed when running Prowler against an Azure SQL Server

### License


jfagoagas (approved):
Thanks for the fix @JackStuart !

PR #2899, opened 2023-10-03
fix(iam creds tests): dont use search and negative indexes

n4ch04:
### Context

In iam credentials checks there were negative indexes and `search` function was being used

### Description

Unify logic using positive indexes and direct assertion when comparing `str` fields

### License


jfagoagas (approved):
💪 Testing!

PR #2901, opened 2023-10-03
fix(custom checks): fix import from s3

n4ch04:
### Context

When importing custom checks from a s3 bucket there are problems when the bucket folders are created from the web console.

https://aws.amazon.com/premiumsupport/knowledge-center/s3-prefix-nested-folders-difference/

### Description

Process the files inside the s3 bucket when they are not directories

### License


MrCloudSec (approved):
🏅

PR #2909, opened 2023-10-05
feat(iam): improve disable credentials checks

### LINKED ISSUE #2377: Pull apart status_extended from iam_disable_30_days_credentials and make more check_ids
### New feature motivation

There's so many status_extended in iam_disable_30_days_credentials and in some cases, the check doesn't make sense to be iam_disable_30_days_credentials but with status_extended='User john.doe has not used access key [...]' because that can be other check (like iam_unused_30_days_credentials).

In my case, I have an account with two findings with the same finding_unique_id (ex: prowler-aws-iam_disable_30_days_credentials-12345-us-east-2-john.doe but one with status_extended='User john.doe has not logged in to the console in the past 30 days.' and the other with status_extended='User john.doe does not have unused access keys for 30 days.'

### Solution Proposed

I propose to separate the check iam_disable_30_days_credentials actually with many status_extended to be other individual checks like iam_disable_30_days_credentials, iam_unused_30_days_credentials, iam_not_exist_credentials or something like that

Now in the project where I using it I discard the duplicated checks because it was causing an error in our side, because we considered finding_unique_id unique but there not.

### Additional context

This is the check https://github.com/prowler-cloud/prowler/tree/8fbee7737bec4869c9ba96e3a25494d966aaaa91/prowler/providers/aws/services/iam/iam_disable_30_days_credentials

I can take the feature and make a PR but in like two weeks if no one can take it.

MrCloudSec:
### Description

Create two configurable checks for IAM credentials:

- iam_user_accesskey_unused
- iam_user_console_access_unused

### License


prowler/providers/aws/services/iam/iam_user_accesskey_unused/iam_user_accesskey_unused.py
```diff
@@ -3,40 +3,12 @@
 from prowler.lib.check.models import Check, Check_Report_AWS
 from prowler.providers.aws.services.iam.iam_client import iam_client
 
-maximum_expiration_days = 45
+maximum_expiration_days = iam_client.audit_config.get("max_unused_access_keys_days")
```
jfagoagas: Putting this line there makes the tests to fail, you have to move it under line 10.

And set a default value with the same value in `config.yaml`. 

https://docs.prowler.cloud/en/latest/developer-guide/checks/#using-the-audit-config

prowler/providers/aws/services/iam/iam_user_console_access_unused/iam_user_console_access_unused.py
```diff
@@ -0,0 +1,39 @@
+import datetime
+
+from prowler.lib.check.models import Check, Check_Report_AWS
+from prowler.providers.aws.services.iam.iam_client import iam_client
+
+maximum_expiration_days = iam_client.audit_config.get("max_console_access_days")
```
jfagoagas: Putting this line there makes the tests to fail, you have to move it under line 10.

And set a default value with the same value in `config.yaml`. 

https://docs.prowler.cloud/en/latest/developer-guide/checks/#using-the-audit-config

jfagoagas (changes requested):

jfagoagas (dismissed):
🚀

jfagoagas (dismissed):

jfagoagas (approved):

PR #2911, opened 2023-10-06
fix(pipeline): launch linters with file changes

n4ch04:
### Context

We were launching tests and linters in the PRs and `main` pushes even if no python/Dockerfile files were involved

### Description

Launch linters only when source code files are involved

### License

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -4,9 +4,15 @@ on:
   push:
     branches:
       - "master"
+    paths:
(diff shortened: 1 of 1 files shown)


MrCloudSec:
how about json files?

jfagoagas (approved):
Great addition @n4ch04 !!

n4ch04:
@sergargar we have changed the approach, we do not launch the linters and tests when changes are from `/docs`

PR #2941, opened 2023-10-18
docs(config): add missing configurable variables

kagahd:
### Context

Some configurable variables were missing in the documentation.

### Description

I added the following configurable variables:

- `max_unused_access_keys_days`
- `max_console_access_days`
- `verify_premium_support_plans`

The first two variables were introduced in the latest prowler version (3.10.0) and did not yet figure in the documentation.
The last variable seems to be introduced only a few hours ago.

The order of the variables in the table corresponds to the order of the variables in the [config.yaml file](https://github.com/prowler-cloud/prowler/blob/master/prowler/config/config.yaml).
I also reformatted the table for better readability.

### License


jfagoagas (approved):
Great catch and improvement @kagahd !! Thanks 🚀

PR #2944, opened 2023-10-19
fix(security group): check if security groups are used by Lambda

MrCloudSec:
### Description

In check `ec2_securitygroup_not_used` verify if security groups are used by any Lambda functions.

### License

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2947, opened 2023-10-19
fix(ec2_securitygroup_not_used): Mock Lambda service

jfagoagas:
### Description

The check `ec2_securitygroup_not_used` now uses a `Lambda` service client (https://github.com/prowler-cloud/prowler/pull/2944) so we need to mock that service client in the check's tests.

I fixed all the `Lambda` mocks using the same way with the check `awslambda_client` and created a `utils` lib with some code that is always present in the tests.

### License


MrCloudSec (approved):
machine 🤖

PR #2948, opened 2023-10-19
feat(report interface): add reporting interface call after report

n4ch04:
### Context

We need to have a more flexible custom report interface to suit our needs 

### Description

Import the custom report interface after `report` native function has been performed

### License

### DIFF
--- a/prowler/lib/check/check.py
+++ b/prowler/lib/check/check.py
@@ -562,9 +562,9 @@ def execute(
             logger.info("Using custom report interface ...")
             lib = os.environ["PROWLER_REPORT_LIB_PATH"]
             outputs_module = importlib.import_module(lib)
-            curstom_report_interface = getattr(outputs_module, "report")
+            custom_report_interface = getattr(outputs_module, "report")
 
-            curstom_report_interface(check_findings, audit_output_options, audit_info)
+            custom_report_interface(check_findings, audit_output_options, audit_info)
         except Exception:
             sys.exit(1)
 
--- a/prowler/lib/check/check.py
+++ b/prowler/lib/check/check.py
@@ -13,21 +13,12 @@ from typing import Any
 from alive_progress import alive_bar
 from colorama import Fore, Style
 
+import prowler
 from prowler.config.config import orange_color
 from prowler.lib.check.compliance_models import load_compliance_framework
 from prowler.lib.check.models import Check, load_check_metadata
 from prowler.lib.logger import logger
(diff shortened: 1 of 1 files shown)


prowler/lib/check/check.py
```diff
@@ -566,6 +557,17 @@ def execute(
...
     report(check_findings, audit_output_options, audit_info)
 
+    if os.environ.get("PROWLER_REPORT_LIB_PATH"):
+        try:
+            logger.info("Using custom report interface ...")
+            lib = os.environ["PROWLER_REPORT_LIB_PATH"]
+            outputs_module = importlib.import_module(lib)
+            curstom_report_interface = getattr(outputs_module, "report")
+
+            curstom_report_interface(check_findings, audit_output_options, audit_info)
```
jfagoagas:
```suggestion
            custom_report_interface(check_findings, audit_output_options, audit_info)
```

prowler/lib/check/check.py
```diff
@@ -566,6 +557,17 @@ def execute(
     # Report the check's findings
     report(check_findings, audit_output_options, audit_info)
 
+    if os.environ.get("PROWLER_REPORT_LIB_PATH"):
+        try:
+            logger.info("Using custom report interface ...")
+            lib = os.environ["PROWLER_REPORT_LIB_PATH"]
+            outputs_module = importlib.import_module(lib)
+            curstom_report_interface = getattr(outputs_module, "report")
```
jfagoagas:
```suggestion
            custom_report_interface = getattr(outputs_module, "report")
```

jfagoagas: I think you have to handle more than a generic `Exception` here not to abort the execution if the `"PROWLER_REPORT_LIB_PATH"` is not present or if the `import` or `getattr` fails.

jfagoagas (changes requested):

prowler/lib/check/check.py
n4ch04: Since the  general `report` function is imported always we actually want Prowler to exit the execution if something wrong happens there

prowler/lib/check/check.py
n4ch04: Done

prowler/lib/check/check.py
n4ch04: Done

jfagoagas (approved):
Smart move!

PR #2956, opened 2023-10-23
fix(sts): force v2 STS tokens

MrCloudSec:
### Description

Set a regional STS endpoint URLs in STS sessions since these always return v2 tokens so they always work with any region.

### License


jfagoagas (approved):
Really good catch!!  Thanks @sergargar for the code and tests 💯

PR #2971, opened 2023-10-26
feat(alias): add check alias functionality

MrCloudSec:
### Description

Create a check name alias for the cases when we rename a check not to break the execution since now if the check is not present we raise an exception aborting the execution.

Change CRITICAL log to ERROR log if the check is not present, e.g.: deleted, not to break the execution too.

### License

### DIFF
--- a/prowler/lib/check/checks_loader.py
+++ b/prowler/lib/check/checks_loader.py
@@ -82,7 +82,7 @@ def load_checks_to_execute(
     # Get Check Aliases mapping
     check_aliases = {}
     for check, metadata in bulk_checks_metadata.items():
-        for alias in metadata.CheckAlias:
+        for alias in metadata.CheckAliases:
             check_aliases[alias] = check
 
     # Verify if any input check is an alias of another check
--- a/prowler/lib/check/checks_loader.py
+++ b/prowler/lib/check/checks_loader.py
@@ -79,15 +79,23 @@ def load_checks_to_execute(
                 check_name = check_info[0]
                 checks_to_execute.add(check_name)
 
+    # Get Check Aliases mapping
+    check_aliases = {}
+    for check, metadata in bulk_checks_metadata.items():
+        for alias in metadata.CheckAlias:
+            check_aliases[alias] = check
+
     # Verify if any input check is an alias of another check
     for input_check in checks_to_execute:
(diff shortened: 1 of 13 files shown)


prowler/lib/check/models.py
```diff
@@ -38,6 +38,7 @@ class Check_Metadata_Model(BaseModel):
     CheckID: str
     CheckTitle: str
     CheckType: list[str]
+    CheckAlias: list[str] = []
```
jfagoagas: - Could you leave it as an `Optional` instead of a empty list?
- Is this field included in the outputs?

prowler/lib/check/checks_loader.py
```diff
@@ -77,4 +79,15 @@ def load_checks_to_execute(
                 check_name = check_info[0]
                 checks_to_execute.add(check_name)
 
+    # Verify if any input check is an alias of another check
+    for input_check in checks_to_execute:
+        for check, metadata in bulk_checks_metadata.items():
```
jfagoagas: Is this the only way to get it? If we have more than 300 checks for AWS, this could result in 9000 iterations. Give this a thought

jfagoagas (changes requested):
Great addition @sergargar 🚀 

Please check my comments and if we renamed some other checks in the past to include it.

prowler/lib/check/models.py
```diff
@@ -38,7 +39,7 @@ class Check_Metadata_Model(BaseModel):
     CheckID: str
     CheckTitle: str
     CheckType: list[str]
-    CheckAlias: list[str] = []
+    CheckAlias: Optional[list[str]] = []
```
jfagoagas: You don't need the `= []` when the type is `Optional` since underneath it will default to `None`

jfagoagas (changes requested):

prowler/lib/check/models.py
MrCloudSec: If I don't add it, it fails since I need it to be an empty list by default, not None.

jfagoagas: In that case remove the optional, since it won't be `None`.

jfagoagas (approved):
Well done @sergargar !! 🚀

PR #2978, opened 2023-10-30
fix(ec2_instance_imdsv2_enabled ): verify if metadata service is disabled

therealtoastycat:
### Context

Hello - I believe the check '**ec2_instance_imdsv2_enabled**' provides inaccurate results when the metadata service is disabled. In my opinion, the result should always be 'PASS' since the HTTP endpoint cannot be queried, regardless of whether IMDSv2 is enabled or not.

Furthermore, the current code will always return 'FAIL' if `http_endpoint == "disabled"`, which leads to a false positive.

Thank you

### Description

I have added an `elif` condition that address this scenario.

### License


MrCloudSec (approved):
Great catch @therealtoastycat , thank you!

PR #2997, opened 2023-11-03
fix(clean local output dirs): clean dirs when output to s3

n4ch04:
### Context

When sending output to s3 bucket with a custom output directory we were not deleting that local output directory

### Description

When options `output_bucket` or `output_bucket_no_assume` are set with `output_directory` and that directory is not the default one -> delete it

### License


prowler/__main__.py
```diff
@@ -301,6 +302,9 @@ def prowler():
     if checks_folder:
         remove_custom_checks_module(checks_folder, provider)
 
+    # clean local directories
+    clean_provider_local_output_directories(args)
+
```
MrCloudSec, line 307: Make this function compatible with all providers or do the condition `if args.output_bucket or args.output_bucket_no_assume` here

MrCloudSec (approved):

PR #3003, opened 2023-11-06
feat(compliance): execute all compliance by default

MrCloudSec:
### Description

Without `--compliance`, prowler will execute all compliance frameworks by default.
All compliance reports will be in output/compliance directory.
image

### License


prowler/lib/outputs/compliance.py
```diff
@@ -45,20 +51,40 @@ def add_manual_controls(output_options, audit_info, file_descriptors):
...
+                    + compliance.Version.lower()
+                    + "_"
+                    + compliance.Provider.lower()
+                )
+            else:
+                compliance_name = (
+                    compliance.Framework.lower() + "_" + compliance.Provider.lower()
+                )
+            if compliance_name.replace("-", "_") in input_compliance_frameworks:
+                check_compliances.append(compliance)
```
jfagoagas: Can you move this to a separate function to handle compliance framework names?

prowler/lib/outputs/compliance.py
jfagoagas, line 1: Can you move each compliance output generator to a separate function? It'll help us to split code, do better tests and improve readability. Thanks!

For the following functions:
- `fill_compliance`
- `display_compliance_table`

prowler/lib/outputs/file_descriptors.py
jfagoagas, line 1: The same here, create a function to generate the file descriptor for each compliance mode not to have the same over and over.

jfagoagas (changes requested):
That's great !! Please check the comments, thanks!

jfagoagas (approved):
👍

PR #3010, opened 2023-11-08
fix(iam-sqs): Handle exceptions for non-existent resources

jfagoagas:
### Description

- SQS: Handle `AWS.SimpleQueueService.NonExistentQueue` errors
- IAM: Handle `NoSuchEntity` and warnings. I've changed `NoSuchEntityException` to `NoSuchEntity` since the `error.response["Error"]["Code"]` contains just the `NoSuchEntity`.

The IAM change can be tested with this code:
```python
from boto3 import client
from botocore.client import ClientError

iam_client = client("iam", "eu-west-1")
arn = "arn:aws:iam::123456789012:policy/test-policy"

try:
    iam_client.list_policy_tags(PolicyArn=arn)
except ClientError as error:
    print(error.response["Error"]["Code"])
```

### License

### DIFF
--- a/prowler/providers/aws/services/iam/iam_service.py
+++ b/prowler/providers/aws/services/iam/iam_service.py
@@ -371,7 +371,7 @@ class IAM(AWSService):
 
                     role.attached_policies = attached_role_policies
                 except ClientError as error:
-                    if error.response["Error"]["Code"] == "NoSuchEntityException":
+                    if error.response["Error"]["Code"] == "NoSuchEntity":
                         logger.warning(
                             f"{self.region} -- {error.__class__.__name__}[{error.__traceback__.tb_lineno}]: {error}"
                         )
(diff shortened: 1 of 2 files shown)


prowler/providers/aws/services/iam/iam_service.py
```diff
@@ -697,7 +697,7 @@ def __get_last_accessed_services__(self):
                     ]
 
                 except ClientError as error:
-                    logger.error(
+                    logger.warning(
```
MrCloudSec: why did you change this one?

prowler/providers/aws/services/iam/iam_service.py
```diff
@@ -717,7 +717,7 @@ def __get_access_keys_metadata__(self):
                             "AccessKeyMetadata"
                         ]
                 except ClientError as error:
-                    logger.error(
+                    logger.warning(
```
MrCloudSec: and this one?

prowler/providers/aws/services/iam/iam_service.py
jfagoagas: Good catch, I've included both exceptions.

prowler/providers/aws/services/iam/iam_service.py
jfagoagas: Good catch, I've included both exceptions.

MrCloudSec (approved):

PR #3012, opened 2023-11-08
fix(rds): check if engines exist in region

MrCloudSec:
### Description

Check if RDS engines exist in region in check `rds_instance_deprecated_engine_version`

### License


jfagoagas (approved):
Good catch!

PR #3014, opened 2023-11-08
fix(iam): do not list tags for inline policies

MrCloudSec:
### Description

Do not list tags for inline policies in IAM Service.

### License


jfagoagas (approved):
Good catch!

PR #3016, opened 2023-11-09
fix(SQS): fix invalid SQS ARNs

mtronrd:
### Context

Prowler is currently generating invalid ARNs for SQS queues.  Instead of parsing the queue name from the queue URL to be the resource id of the ARN, it is using the entire queue URL.  So for instance a queue named `test-queue` should have an ARN like `arn:aws:sqs:eu-west-1:123456789012:test-queue` but prowler is instead generating the invalid ARN `arn:aws:sqs:eu-west-1:123456789012:https://sqs.eu-west-1.amazonaws.com/123456789012/test-queue` and this is what is being reported to Security Hub.

### Description

The sqs_service has been fixed to parse the last path segment of the queue as the queue name, and use that instead of the queue URL when generating the ARN.  The tests for the sqs_service and existing checks have been strengthened to more accurately simulate AWS API responses and to check that the SQS ARNs generated by Prowler are correct.  All tests are passing.

I left the queue id as the full URL, as the URL is guaranteed to be unique and the queue name can collide between regions and accounts.

### License

### DIFF
--- a/prowler/providers/aws/services/sqs/sqs_service.py
+++ b/prowler/providers/aws/services/sqs/sqs_service.py
@@ -23,7 +23,11 @@ class SQS(AWSService):
         logger.info("SQS - describing queues...")
         try:
             list_queues_paginator = regional_client.get_paginator("list_queues")
-            for page in list_queues_paginator.paginate():
+            # The SQS API uses nonstandard pagination
+            # you must specify a PageSize if there are more than 1000 queues
+            for page in list_queues_paginator.paginate(
+                PaginationConfig={"PageSize": 1000}
+            ):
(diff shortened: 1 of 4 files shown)


mtronrd:
Caught and fixed another SQS bug testing this with an account that has >1000 queues.  The SQS API handles pagination differently from most other AWS APIs and you must always specify `"PageSize": 1000` when paginating or it will truncate at 1000.  I wrote a test for this but left it commented, as moto does not currently mock this correctly and is hardcoded to truncate at 1000 regardless of parameters.

MrCloudSec (approved):
Great catch @mtronrd, thank you for your contribution! Love the tests.

PR #3028, opened 2023-11-13
fix(GuardDuty): only execute checks if GuardDuty enabled

MrCloudSec:
### Description

Only execute checks `guardduty_centrally_managed` and `guardduty_no_high_severity_findings` if GuardDuty is enabled.

### License


jfagoagas (changes requested):
That's great to reduce findings if the service is not enabled. Could you include two tests to verify this new behaviour? 

Thanks!

jfagoagas (approved):

PR #3038, opened 2023-11-14
feat(custom_checks_metadata): Add checks metadata overide for severity

### LINKED ISSUE #2834: Severity level overrides
### New feature motivation

Currently, “Severity” is strictly specified in every check, but our company does not agree with some of them and we want to redefine them.
For example, for the "acm_certificates_expiration_check" check the level is high, but in accordance with internal regulations we have this is the “medium” level (the monitoring team’s fixation metrics depend on this).
Now, as a temporary solution, all prowler outputs are added up “as is”, and only then they are redefined on the SIEM (a simple example in the screenshot can be seen in the MySeverity column).
image

### Solution Proposed

Add an argument (in addition to "--checks-file") that parses "severity" (if present) and overrides the default value. As a result, it will be possible to specify in one file the list of checks of interest and their new level of criticality.

-

### Additional context

In mature companies, response time greatly depends on Severity; each organization must determine the level of criticality for itself depending on the assets. The ability to use one input file (in addition to "--checks-file") to specify a list of checks of interest and their new level of criticality will be a great addition.

venkateswarluvajrala:
Fixes: #2834

### Context:
Currently there is no way to override checks metadata like severity. 
Ex:  `s3_bucket_level_public_access_block` is by default has severity `medium` but it may not be true for all the organisations. In mature companies, response time greatly depends on Severity; each organization must determine the level of criticality for itself depending on the assets. The ability to use one input file (in addition to "--checks-file") to specify a list of checks of interest and their new level of criticality will be a great addition.

While running only high-level severities as well, we may wish to get high level severities as per our own organisation standards. There is also an option to run just `high` level severity or `low` using `--severity` flag. If we introduce this new flag `--custom-checks-metadata-file`, it will be beneficial to run severities based on their use case as well.
 
### Solution:
This PR solves the problem by introducing a new argument called `--custom-checks-metadata-file `.
It will take custom checks metadata file as an argument and overrides default severity values for the checks with the provided values.

### License


jfagoagas:
Hi @venkyvajrala, thanks for the PR, that's awesome! We will review it as soon as we can 🚀

jchrisfarris:
This is awesome. I was just thinking last night I needed to implement something similar!

jfagoagas:
Hi @venkyvajrala we've made some changes in the PR not to make provider-specific logic since the metadata model is shared between the providers. Please check it out and give us your thoughts.

I've pending to include the tests for the new business logic included.

prowler/lib/check/custom_checks_metadata.py
```diff
@@ -0,0 +1,53 @@
...
+        )
+        sys.exit(1)
+
+
+def update_checks_metadata(bulk_checks_metadata, custom_checks_metadata):
+    try:
+        # Update checks metadata from CustomChecksMetadata file
+        for check, custom_metadata in custom_checks_metadata["Checks"].items():
+            check_metadata = bulk_checks_metadata.get(check)
+            if check:
```
venkyvajrala, line 38: Can we check if check_metadata is present here for given check . 
User can give some `check = "some_random_check_name"` which aren't existed then 
`check_metadata = bulk_checks_metadata.get(check)`  becomes  `None` as check is not existed.
It will then try updating a check which isn't existed in next line.

What's your thoughts on this?

prowler/lib/check/custom_checks_metadata.py
```diff
@@ -0,0 +1,53 @@
...
+                bulk_checks_metadata[check] = update_check_metadata(
+                    check_metadata, custom_metadata
+                )
```
venkateswarluvajrala: Can we check if check_metadata is present here for given check .
User can give some `check = "some_random_check_name"` which aren't existed then
`check_metadata = bulk_checks_metadata.get(check)` becomes `None` as check is not existed.
It will then try updating a check which isn't existed in next line and error out

What's your thoughts on this @jfagoagas ?

jfagoagas: I understand your point but we are looping over the `custom_checks_metadata` so we only modify the checks present in that file regardless of user input. If you set `--checks non_existent_check` nothing happens there since that function will only update the checks with the modified metadata.

prowler/lib/cli/parser.py
```diff
@@ -286,3 +287,15 @@ def __init_config_parser__(self):
...
+    def __init_custom_checks_metadata_parser__(self):
+        # CustomChecksMetadata
+        custom_checks_metadata_subparser = (
+            self.common_providers_parser.add_argument_group("Custom Checks Metadata")
+        )
+        custom_checks_metadata_subparser.add_argument(
+            "--custom-checks-metadata-file",
+            nargs="?",
+            default=None,
+            help="Path for the custom checks metadata YAML file. See example prowler/config/custom_checks_metadata.yaml for reference and format. See more in https://docs.prowler.cloud/en/latest/tutorials/custom-checks-metadata/",
```
MrCloudSec: Can you change the filename to `prowler/config/example_custom_checks_metadata.yaml` and add the necessary documentation? Thanks.

prowler/lib/check/custom_checks_metadata.py
venkateswarluvajrala: I just tried in my local as well 
```bash
prowler aws --custom-checks-metadata-file prowler/config/custom_checks_metadata.yaml -s s3 --log-level DEBUG
```
```yaml
CustomChecksMetadata:
  aws:
    Checks:
      "some_randome_check":
        Severity: "high"

```
 It gave below error. 
`[File: custom_checks_metadata.py:48]    [Module: custom_checks_metadata]         CRITICAL: AttributeError -- 'NoneType' object has no attribute 'Severity'[42]`

When I updated the condition from `if check:` to `if check_metadata:` it worked.

jfagoagas: What is the content of the `custom_checks_metadata.yaml`?

venkateswarluvajrala:
```yaml
CustomChecksMetadata:
  aws:
    Checks:
      "some_randome_check":
        Severity: "high"
      
```

jfagoagas: Now I get it, this case it's not handled, also I think tests would arise that, I'll fix that tomorrow. Any other thought about the refactor? I think that storing everything within the same file for all the providers is simple enough, what do you think?

venkateswarluvajrala: Yes @jfagoagas I also found this approach more simpler.

MrCloudSec (dismissed):
Awesome feature, thank you! 🚀

jfagoagas (approved):
Ready to merge!! Thanks again @venkyvajrala, this feature is awesome 🚀

PR #3040, opened 2023-11-16
fix(securityhub): findings not being imported or archived in non-aws partitions

johnny2lu:
### Context

Prowler is incorrectly reporting no items being imported and archived to non-aws partitions when Prowler is executed with the -S flag to send findings to securityhub.

### Description

The prowler integration arn is incorrectly formatted in non-aws partitions, causing the verify_security_hub_integration_enabled_per_region to return false in regions where securityhub is actually enabled. 

### License


MrCloudSec (dismissed):

MrCloudSec (approved):

jfagoagas:
@johnny2lu Thank you very much for your contribution, good catch!

PR #3046, opened 2023-11-20
chore(sts-endpoint): deprecate `--sts-endpoint-region`

MrCloudSec:
### Description

Deprecate `--sts-endpoint-region` since with v3.11.0, Prowler uses a regional token in STS sessions so it can scan all AWS regions without needing the `--sts-endpoint-region` argument.
### License


jfagoagas (approved):
👍

PR #3063, opened 2023-11-23
fix(set_azure_audit_info): assign correct logging when no auth

n4ch04:
### Context

In azure audit info, when we are not getting required env variables we are exiting using
```
raise Exception
```
Instead of 
```
logger.critical()
sys.exit(1)
```
That must be unified

### Description

Replace `raise Exception` for the common logging critical path in `prowler/providers/common/audit_info.py`

### License

### DIFF
--- a/prowler/providers/azure/lib/exception/exception.py
+++ b/prowler/providers/azure/lib/exception/exception.py
@@ -1,0 +1,11 @@
+class AzureException(Exception):
+    """
+    Exception raised when dealing with Azure Provider/Azure audit info instance
+
+    Attributes:
+        message -- message to be displayed
+    """
+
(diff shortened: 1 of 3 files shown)


jfagoagas (changes requested):
I prefer to let the caller (in this case `set_provider_audit_info`) to decide what to do (current code) and return `Exception` from the `set_azure_audit_info` but I'd create a new `class AzureException(Exception)` with the error message. What do you think?

n4ch04:
I would prefer to have all unified prior to start with that, but we can start with this to implement custom `Exception` classes

jfagoagas (approved):
Awesome!!

More tests 🙌🙌

PR #3064, opened 2023-11-23
fix(gcp provider): move generate_client for consistency

n4ch04:
### Context

In `prowler/providers/gcp/gcp_provider.py` there is a function  (`generate_client`) that is not a class method  and should be in `prowler/providers/gcp/lib/service/service.py`   for consistency

### Description

Relocate `generate_client` function from `prowler/providers/gcp/gcp_provider.py` to `prowler/providers/gcp/lib/service/service.py`

### License

### DIFF
--- a/prowler/providers/gcp/gcp_provider.py
+++ b/prowler/providers/gcp/gcp_provider.py
@@ -3,10 +3,8 @@ import sys
 
 from google import auth
 from googleapiclient import discovery
-from googleapiclient.discovery import Resource
 
 from prowler.lib.logger import logger
-from prowler.providers.gcp.lib.audit_info.models import GCP_Audit_Info
 
 
 class GCP_Provider:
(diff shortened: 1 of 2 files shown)


jfagoagas (changes requested):
Totally agree with this change but my two cents:
1. Why don't we move this method to the `GCPService` class?
2. Do we need a function just to execute `discovery.build(service, api_version, credentials=audit_info.credentials)`?
3. Can we pass just the `credentials` object instead of passing the whole `audit_info`?

And please, include tests for that function.

Thanks!

n4ch04:
Added as method, makes sense since that function is not being used outside that context.
Since the error treatment is handled per method and not in `__init__` I think we need it yes.
Passed only the credentials and not the whole `audit_info` object

Regarding the tests, a new task will be generated to create test for the whole service class

jfagoagas (approved):
Great one, thanks for addressing the comments @n4ch04  🚀

PR #3066, opened 2023-11-27
refactor(load_checks_to_execute): Refactor function and add tests

jfagoagas:
### Context

### Description

### License

### DIFF
--- a/prowler/lib/check/check.py
+++ b/prowler/lib/check/check.py
@@ -107,14 +107,19 @@ def exclude_services_to_run(
 
 # Load checks from checklist.json
 def parse_checks_from_file(input_file: str, provider: str) -> set:
-    checks_to_execute = set()
-    with open_file(input_file) as f:
-        json_file = parse_json_file(f)
+    try:
+        checks_to_execute = set()
+        with open_file(input_file) as f:
+            json_file = parse_json_file(f)
 
-    for check_name in json_file[provider]:
-        checks_to_execute.add(check_name)
+        for check_name in json_file[provider]:
+            checks_to_execute.add(check_name)
 
-    return checks_to_execute
+        return checks_to_execute
+    except Exception as error:
+        logger.error(
+            f"{error.__class__.__name__}[{error.__traceback__.tb_lineno}] -- {error}"
+        )
 
 
 # Load checks from custom folder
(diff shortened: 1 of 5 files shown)


prowler/lib/check/check.py
```diff
@@ -607,22 +613,32 @@ def update_audit_metadata(
...
-    return checks
+    Returns a set of checks from the given services
+    """
+    try:
+        checks = set()
+        service_list = [
+            "awslambda" if service == "lambda" else service for service in service_list
+        ]
+        for service in service_list:
+            modules = recover_checks_from_provider(provider, service)
```
n4ch04: Here `module` is not descriptive, why not `service_checks` or something like that?
It seems that you are handling a different entity

prowler/lib/check/check.py
```diff
@@ -607,22 +613,32 @@ def update_audit_metadata(
...
+            if not modules:
+                logger.error(f"Service '{service}' does not have checks.")
+
+            else:
+                for check_module in modules:
```
n4ch04: Then :
```
for check in service_checks:
...
```

prowler/lib/check/checks_loader.py
```diff
@@ -22,73 +22,99 @@ def load_checks_to_execute(
...
-        if service_list:
-            checks_to_execute = (
-                recover_checks_from_service(service_list, provider) & checks_to_execute
-            )
+    """Generate the list of checks to execute based on the cloud provider and the input arguments given"""
+    try:
+        # Local subsets
+        checks_to_execute = set()
+        check_aliases = {}
+        check_severities = {
```
n4ch04, line 30: I know that severities are not gonna change in the short term, but having this  here hardcoded can create a point of failure.
What do you think?

n4ch04 (changes requested):

prowler/lib/check/check.py
jfagoagas: Totally agree with you!

prowler/lib/check/check.py
jfagoagas: Donete!

prowler/lib/check/checks_loader.py
jfagoagas: This is a good point but I'd say to leave it that way for now and came back after some time, what do you think?

n4ch04: Yes, it does not have an easy solution, and since severity is not gonna change for now we can leave it here

n4ch04 (approved):

PR #3069, opened 2023-11-27
feat(CloudProvider): introduce global provider Azure&GCP

n4ch04:
### Context

In order to ease creation of new providers we want Prowler to consolidate all the providers code in a single module

### Description

Create a new `CloudProvider` class that will be inherited by `GcpProvider` and `AzureProvider`.
Those new provider classes will replace `audit_info` objects in the execution path (outputs  path will be addressed in the futere)

### License


prowler/providers/azure/azure_provider_new.py
```diff
@@ -0,0 +1,270 @@
...
+from colorama import Fore, Style
+from msgraph.core import GraphClient
+from pydantic import BaseModel
+
+from prowler.lib.logger import logger
+from prowler.providers.azure.lib.regions.regions import get_regions_config
+from prowler.providers.common.provider import CloudProvider
+
+
+class Azure_Identity_Info(BaseModel):
```
jfagoagas: I think we should remove `_` from the class names since is not the convention in Python.

prowler/providers/azure/azure_provider_new.py
jfagoagas, line 1: I don't know if makes sense to use `BaseModel` in the classes if we are going to use any validation at all.

prowler/providers/azure/azure_provider_new.py
```diff
@@ -0,0 +1,270 @@
...
+            except Exception as error:
+                logger.critical("Failed to retrieve azure credentials")
+                logger.critical(
+                    f"{error.__class__.__name__}[{error.__traceback__.tb_lineno}] -- {error}"
+                )
+                sys.exit(1)
+
+        return credentials
+
+    def __check_sp_creds_env_vars__(self):
```
jfagoagas: Can you rename this function to something more readable? 

Also we need to decide whether to use the `__function_name__` notation or not since the rest is not using it.

prowler/providers/azure/lib/service/service.py
```diff
@@ -1,30 +1,30 @@
...
         self,
         service: str,
-        audit_info: Azure_Audit_Info,
+        provider: AzureProvider,
     ):
         self.clients = self.__set_clients__(
-            audit_info.identity.subscriptions,
-            audit_info.credentials,
+            provider.identity.subscriptions,
+            provider.session,
```
jfagoagas, line 13: Why is session now if the object is called Credentials?

prowler/providers/azure/services/defender/defender_client.py
```diff
@@ -1,4 +1,4 @@
-from prowler.providers.azure.lib.audit_info.audit_info import azure_audit_info
 from prowler.providers.azure.services.defender.defender_service import Defender
+from prowler.providers.common.common import global_provider
 
-defender_client = Defender(azure_audit_info)
+defender_client = Defender(global_provider)
```
jfagoagas, line 4: Why this is receiving something called `global_provider` instead of some `Azure....`?

prowler/providers/common/audit_info.py
```diff
@@ -322,8 +322,9 @@ def set_azure_audit_info(self, arguments) -> Azure_Audit_Info:
             credential_scopes=region_config["credential_scopes"],
         )
 
-        if not arguments.get("only_logs"):
-            self.print_azure_credentials(azure_audit_info)
+        # this logic is being processed in general provider
+        # if not arguments.get("only_logs"):
+        #     self.print_azure_credentials(azure_audit_info)
```
jfagoagas, line 329: If this is not needed please remove it or include a `TODO` to remove it.

prowler/providers/common/audit_info.py
```diff
@@ -348,8 +349,9 @@ def set_gcp_audit_info(self, arguments) -> GCP_Audit_Info:
             gcp_audit_info.project_ids,
         ) = gcp_provider.get_credentials()
 
-        if not arguments.get("only_logs"):
-            self.print_gcp_credentials(gcp_audit_info)
+        # this logic is being processed in general provider
+        # if not arguments.get("only_logs"):
+        #     self.print_gcp_credentials(gcp_audit_info)
```
jfagoagas, line 357: The same as above.

prowler/providers/common/common.py
```diff
@@ -1,7 +1,20 @@
...
 
 providers_prowler_lib_path = "prowler.providers"
 
+# SHARED PROVIDER OBJECT ACROSS ALL PROWLER CODE
+global_provider = None
+
+
+def set_provider(provider, arguments) -> Any:
+    provider_class_name = f"{provider.capitalize()}Provider"
+    import_module_path = f"prowler.providers.{provider}.azure_provider_testing"
```
jfagoagas, line 12: Watch out, the module path says `azure_provider_testing`, is it expected?

prowler/providers/common/provider.py
```diff
@@ -0,0 +1,29 @@
...
+    audit_resources: list = None
+    is_quiet: bool
+    output_modes: list
+    output_directory: str
+    allowlist_file: str
+    bulk_checks_metadata: dict
+    verbose: str
+    output_filename: str
+    only_logs: bool
+    unix_timestamp: bool
```
jfagoagas: Why there are output options stored in the `CloudProvider`?

prowler/providers/gcp/gcp_provider_new.py
jfagoagas, line 1: Is this file name expected?

prowler/providers/gcp/gcp_provider_new.py
```diff
@@ -0,0 +1,119 @@
...
+from colorama import Fore, Style
+from google import auth
+from google.oauth2.credentials import Credentials
+from googleapiclient import discovery
+
+from prowler.lib.logger import logger
+from prowler.providers.common.provider import CloudProvider
+
+
+class GcpProvider(CloudProvider):
```
jfagoagas, line 14: We should do something like `GCPProvider`, `AzureProvider`, `AWSProvider`, what do you think?

prowler/providers/gcp/lib/service/service.py
```diff
@@ -66,3 +67,16 @@ def __is_api_active__(self, audited_project_ids):
                     f"{error.__class__.__name__}[{error.__traceback__.tb_lineno}]: {error}"
                 )
         return project_ids
+
+
+def generate_client(
```
jfagoagas: I think this should be moved to the class, right?

jfagoagas (changes requested):
That's a great 1st approach to the Providers class, thanks @n4ch04 !!

prowler/providers/azure/azure_provider_new.py
n4ch04: Classes names now match camelcase structure

prowler/providers/azure/azure_provider_new.py
n4ch04: We will address this in the future

prowler/providers/azure/azure_provider_new.py
n4ch04: It is a private function, now renamed to be more explicit

prowler/providers/azure/lib/service/service.py
n4ch04: The goal is to have a general name for the provider session, we were using session in AWS and credentials in Azure and GCP, but only because the internal classes for auth and session are called `Credentials`

I think session is more general and suits for all the existing and future providers

prowler/providers/common/common.py
n4ch04: Yes, it is a way to make the new implementation cohabit with the old one until we know everything works

prowler/providers/common/audit_info.py
n4ch04: Added

prowler/providers/common/audit_info.py
n4ch04: Added

prowler/providers/common/provider.py
n4ch04: Deleted, at the beginning they were included thinking about including the outputs changes in this PR but maybe is better to include it in the future

prowler/providers/gcp/gcp_provider_new.py
n4ch04: Yes, same as azure

prowler/providers/gcp/lib/service/service.py
n4ch04: Moved during merge

prowler/providers/azure/lib/audit_info/audit_info.py
```diff
@@ -1,14 +1,14 @@
...
 
 azure_audit_info = Azure_Audit_Info(
     credentials=None,
-    identity=Azure_Identity_Info(),
+    identity=AzureIdentityInfo(),
     audit_resources=None,
     audit_metadata=None,
     audit_config=None,
-    azure_region_config=Azure_Region_Config(),
+    AzureRegionConfig=AzureRegionConfig(),
```
jfagoagas:
```suggestion
    azure_region_config=AzureRegionConfig(),
```

prowler/providers/azure/lib/audit_info/models.py
```diff
@@ -23,11 +23,11 @@ class Azure_Region_Config(BaseModel):
 @dataclass
 class Azure_Audit_Info:
     credentials: DefaultAzureCredential
-    identity: Azure_Identity_Info
+    identity: AzureIdentityInfo
     audit_resources: Optional[Any]
     audit_metadata: Optional[Any]
     audit_config: dict
-    azure_region_config: Azure_Region_Config
+    AzureRegionConfig: AzureRegionConfig
```
jfagoagas:
```suggestion
    azure_region_config: AzureRegionConfig
```

prowler/providers/azure/lib/audit_info/models.py
```diff
@@ -36,11 +36,11 @@ def __init__(
         audit_metadata,
         audit_resources,
         audit_config,
-        azure_region_config,
+        AzureRegionConfig,
```
jfagoagas:
```suggestion
        azure_region_config,
```

prowler/providers/azure/lib/audit_info/models.py
```diff
@@ -36,11 +36,11 @@ def __init__(
...
     ):
         self.credentials = credentials
         self.identity = identity
         self.audit_metadata = audit_metadata
         self.audit_resources = audit_resources
         self.audit_config = audit_config
-        self.azure_region_config = azure_region_config
+        self.AzureRegionConfig = AzureRegionConfig
```
jfagoagas:
```suggestion
        self.azure_region_config = AzureRegionConfig
```

jfagoagas (changes requested):

jfagoagas (dismissed):
One step closer 👏

jfagoagas (approved):
👍

PR #3075, opened 2023-11-28
fix(trustedadvisor): handle missing dict key

n4ch04:
### Context

In `trustedadvisor` service we were not handling if there were no checks

### Description

Include a `.get()` method when looking for `checks`

### License


jfagoagas (approved):
Great catch @n4ch04 👏

PR #3083, opened 2023-11-30
chore(clean output dirs): clean local output dirs with flag

n4ch04:
### Context

Cleaning output local directories when sending outputs to external entities could not be always useful, since it deletes original files

### Description

Only delete that information when including `--clean-local-output-directories`

### License

### DIFF
--- a/prowler/__main__.py
+++ b/prowler/__main__.py
@@ -325,7 +325,8 @@ def prowler():
         remove_custom_checks_module(checks_folder, provider)
 
     # clean local directories
-    clean_provider_local_output_directories(args)
+    if args.clean_local_output_directories:
+        clean_provider_local_output_directories(args)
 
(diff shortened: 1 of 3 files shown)


jfagoagas (changes requested):
I think this is better than before not to break current integrations but there is something that needs to be addressed because the Prowler output banner says:
```
Detailed results are in:
 - HTML: ./test-output/prowler-output--20231201102237.html
 - JSON-OCSF: ./test-output/prowler-output--20231201102237.ocsf.json
 - CSV: ./test-output/prowler-output--20231201102237.csv
 - JSON: ./test-output/prowler-output--20231201102237.json
 ```

n4ch04:
After reviewing this, it implies also changing output message depending on remote output options.
Maybe the solution generates more issues than the original problem. 
For now it is gonna be better if we simply do not delete the local output dirs

PR #3088, opened 2023-12-01
fix(send_to_s3_bucket): don't kill exec when fail

n4ch04:
### Context

When something was going wrong sending outputs to an s3 bucket Prowler was exiting the execution.
Since there are some actions that have are taken after calling `send_to_s3_bucket` we do not want to exit the exec

### Description

- Change `CRITICAL` log to `ERROR` in `send_to_s3_bucket`

- Remove the `sys.exit(1)`

### License


jfagoagas (approved):
Great catch @n4ch04 !!

PR #3095, opened 2023-12-04
 fix(aws_regions): Get enabled regions

### LINKED ISSUE #3089: Suggestions: Check for enabled regions
### New feature motivation

Slightly reduce the number of open file descriptors and number of calls due to connections made to endpoints in regions not in use

### Solution Proposed

Use the describe_regions method for the ec2 service to determine enabled regions. Im just not sure which region to make the first call to.

A list of regions can be supplied to prowler, but it would be nice to do this as the default behavior 
```
def get_enabled_regions():
    ec2 = boto3.client('ec2', region_name='eu-west-1')
    response = ec2.describe_regions(AllRegions=False)
    regions = [region['RegionName'] for region in response['Regions']]
    return regions
```

None

### Additional context

_No response_

jfagoagas:
### Context

Fixes #3089 Thanks for the idea @Fennerr !

RIght now Prowler audits all the regions by a given service not taking into account if the region is enabled within AWS.

### Description

For each AWS service supported in Prowler check if the regions, where the service is available, are enabled in the audited AWS account.

### License


MrCloudSec (approved):
Great PR !! 🚀

PR #3175, opened 2023-12-11
fix(threading): Improved threading for the AWS Service

Fennerr:
### Context

Improved the threading implementation for the `AWSService` base class to allow for an optional `iterator` parameter to be passed to it. It will then create a thread per item. This can be used in cases when we are not just listing resources in a region - such as collecting user data for each EC2 instance or collecting details for each EBS snapshot.

Also changed the implementation to use a thread pool.

### Description

Added an optional iterator for when threading needs to be implemented against a list of items, not just the regional clients

Also changed from using the `threading` lib to the `concurrent.futures.ThreadPoolExecutor` thread pool. It manages a pool of reusable threads and efficiently handles a large number of tasks. This is to reduce the overhead of creating a thread once-per-iterator-item.

The behavior is nearly identical to the previous implementation when an `iterator` parameter isnt passed to the __threading_call__ function (except for the max_workers set to 10, which may be less than the number of concurrent threads before when 1 was created per region. This could be increased to ~20 (this is approximately the max regions I think?) to replicate the concurrency achieved before). The back-off strategy implemented by boto for retries should avoid the API rate-limiting being an issue. We could also increase the max retries to 5 by default for added safety. 

But now an additional `iterator` parameter can be passed. I have implemented this for the functions that collect information for the snapshots and ec2 instance user data.

### License

### DIFF
--- a/prowler/providers/aws/lib/service/service.py
+++ b/prowler/providers/aws/lib/service/service.py
@@ -1,5 +1,5 @@
 import threading
-
+from concurrent.futures import ThreadPoolExecutor, as_completed
 from prowler.providers.aws.aws_provider import (
     generate_regional_clients,
     get_default_region,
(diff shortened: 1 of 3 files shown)


Fennerr:
The threaded functions were logging a message, which became a lot, especially when threading per item (when collecting user data/snapshot data).

I modified the `__threading_call__` function to print a message based on the call that was being threaded - so that this message is only logged once per call

Also added the thread pool

Fennerr:
Can also use this to replace the `__threading_call__` implementation in the [s3 service](https://github.com/prowler-cloud/prowler/blob/master/prowler/providers/aws/services/s3/s3_service.py)

jfagoagas:
> Can also use this to replace the `__threading_call__` implementation in the [s3 service](https://github.com/prowler-cloud/prowler/blob/master/prowler/providers/aws/services/s3/s3_service.py)

Can you include a `# TODO` comment there?

jfagoagas (approved):
This is great @Fennerr !! I've included a constant for the max_workers and a `TODO` for the S3 service.

🚀

PR #3183, opened 2023-12-12
fix(apigw_restapi_auth check): add method auth testing

### LINKED ISSUE #3082: [Bug]: apigateway_restapi_authorizers_enabled.py does not consider authorizers configured at the method level
### Steps to Reproduce
```
prowler -s apigateway
```
### Expected behavior

Prowler produces a false positive (kind of) when there isnt a authorizer enabled at at the rest api level - but all the methods in the rest api had an authorizer configured

Screenshot from 2023-11-30 17-05-22

### Environment Resource

Local instance (kali)

### OS used

Debian-based (kali)

### Prowler version

Prowler 3.11.3 (You are running the latest version, yay!)

### Pip version

pip 23.3

### Context

_No response_

n4ch04:
### Context

Solves #3082 

### Description

When there is no authorized configured at API level, test if all the methods are authorized

### License

### DIFF
--- a/prowler/providers/aws/services/apigateway/apigateway_restapi_authorizers_enabled/apigateway_restapi_authorizers_enabled.py
+++ b/prowler/providers/aws/services/apigateway/apigateway_restapi_authorizers_enabled/apigateway_restapi_authorizers_enabled.py
@@ -13,27 +13,40 @@ class apigateway_restapi_authorizers_enabled(Check):
             report.resource_id = rest_api.name
             report.resource_arn = rest_api.arn
             report.resource_tags = rest_api.tags
+            # it there are not authorizers at api level and resources without methods (default case) ->
             report.status = "FAIL"
-            report.status_extended = f"API Gateway {rest_api.name} ID {rest_api.id} does not have an authorizer configured at api or methods level."
+            report.status_extended = f"API Gateway {rest_api.name} ID {rest_api.id} does not have an authorizer configured at api level."
             if rest_api.authorizer:
                 report.status = "PASS"
-                report.status_extended = f"API Gateway {rest_api.name} ID {rest_api.id} has an authorizer configured at api or methods scope."
+                report.status_extended = f"API Gateway {rest_api.name} ID {rest_api.id} has an authorizer configured at api level"
             else:
                 # we want to know if api has not authorizers and all the resources don't have methods configured
-                api_resources_with_all_methods_authorized = False
+                resources_have_methods = False
+                all_methods_authorized = True
+                resource_paths_with_unathorized_methods = []
                 for resource in rest_api.resources:
                     # if the resource has methods test if they have all configured authorizer
                     if resource.resource_methods:
-                        if "NONE" not in resource.resource_methods.values():
-                            api_resources_with_all_methods_authorized = True
-                        else:
-                            # with only one unauthorized method -> return FAIL
-                            api_resources_with_all_methods_authorized = False
-                            break
-
-                if api_resources_with_all_methods_authorized:
+                        resources_have_methods = True
+                        for (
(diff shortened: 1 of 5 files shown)


prowler/providers/aws/services/apigateway/apigateway_restapi_authorizers_enabled/apigateway_restapi_authorizers_enabled.py
```diff
@@ -13,12 +13,28 @@ def execute(self):
...
+                api_resources_with_all_methods_authorized = False
+                for resource in rest_api.resources:
+                    # if the resource has methods test if they have all configured authorizer
+                    if resource.resource_methods:
+                        if "NONE" not in resource.resource_methods.values():
+                            api_resources_with_all_methods_authorized = True
+                        else:
+                            # with only one unauthorized method -> return FAIL
+                            api_resources_with_all_methods_authorized = False
+                            break
```
jfagoagas: Can you move this logic to the APIGateway service?

prowler/providers/aws/services/apigateway/apigateway_service.py
```diff
@@ -95,7 +100,46 @@ def __get_stages__(self):
...
+                get_resources_paginator = regional_client.get_paginator("get_resources")
+                for page in get_resources_paginator.paginate(restApiId=rest_api.id):
+                    for resource in page["items"]:
+                        id = resource["id"]
+                        resource_methods = []
+                        methods_auth = {}
+                        for resource_method in resource.get(
+                            "resourceMethods", {}
+                        ).keys():
+                            # rest_api.resources.resource_methods.update({resource_method:resource_config.get("authorizationType", None)})
```
jfagoagas:
```suggestion
```

prowler/providers/aws/services/apigateway/apigateway_service.py
```diff
@@ -107,6 +151,11 @@ class Stage(BaseModel):
     tags: Optional[list] = []
 
 
+class PathResourceMethods(BaseModel):
+    path: str
+    resource_methods: dict
```
jfagoagas, line 156: Can we create a type for this attribute?

jfagoagas (changes requested):

prowler/providers/aws/services/apigateway/apigateway_restapi_authorizers_enabled/apigateway_restapi_authorizers_enabled.py
n4ch04: In this case applies to the check

prowler/providers/aws/services/apigateway/apigateway_service.py
n4ch04: In this case since we are only getting that info adding a type for that info complicates more the check's logic.
In the future if we need to retrieve more info from the resources we will include the model

prowler/providers/aws/services/apigateway/apigateway_service.py
n4ch04: Done

Fennerr:
I have added a [pull request](https://github.com/prowler-cloud/prowler/pull/3188) with some modifications.

I ran this against an account and the logic is working as intended, but the status message could have more information for the end user

n4ch04:
I've added more detail to the `status_extended` field and handled `OPTIONS` case in the service, also added more tests.
Please @Fennerr  check it out.  It was easier modifying the original than working from your PR but thanks for all the help and your work!!

MrCloudSec (approved):

jfagoagas (approved):
Good one 🚀

PR #3185, opened 2023-12-12
fix(gcp): improve logging messages

MrCloudSec:
### Context

If an user does not have the `cloudresourcemanager` API enabled, it will throw an ERROR log and abort the execution of Prowler and he may not see it.

### Description

Adding a print showing that the `cloudresourcemanager` API is not enabled and also solves a typo in the other print.

### License


jfagoagas (approved):
🔝

PR #3188, opened 2023-12-12
More verbose status message for failed checks

Fennerr:
### Context
Made changes to [3183](https://github.com/prowler-cloud/prowler/pull/3183), but do not have permissions to commit to this branch.

### Description

Added more verbose messaging, so that it is easier for the user to identify why the check failed, and perform remediation.

### License

### DIFF
--- a/prowler/providers/aws/services/apigateway/apigateway_restapi_authorizers_enabled/apigateway_restapi_authorizers_enabled.py
+++ b/prowler/providers/aws/services/apigateway/apigateway_restapi_authorizers_enabled/apigateway_restapi_authorizers_enabled.py
@@ -38,7 +38,7 @@ class apigateway_restapi_authorizers_enabled(Check):
                     report.status = "FAIL"
                     failed_status_parts = []
                     for path, methods in paths_without_authorizers.items():
-                        methods_str = self.array_to_string(methods)
+                        methods_str = ", ".join(methods)
                         if len(methods) == 1:
                             failed_status_parts.append(
                                 f"{path} without authorizer for the {methods_str} method"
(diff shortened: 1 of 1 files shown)


Fennerr:
I also filtered out the OPTION methods. It is common for pre-flight requests to be sent without authentication material - so the OPTION method usually doesn't have an authorizer. But we can remove the filtering out of the OPTION methods if you would like

prowler/providers/aws/services/apigateway/apigateway_restapi_authorizers_enabled/apigateway_restapi_authorizers_enabled.py
```diff
@@ -14,27 +14,60 @@ def execute(self):
...
                     report.status = "PASS"
-                    report.status_extended = f"API Gateway {rest_api.name} ID {rest_api.id} has an authorizer configured at api or methods level."
+                    report.status_extended = f"API Gateway {rest_api.name} ({rest_api.id}) has all methods authorized."
 
             findings.append(report)
 
         return findings
+
+    @staticmethod
+    def array_to_string(arr):
```
jfagoagas: I think there is no need for this function, it is enough if you show the values comma-separated, like:
```python
>>> ", ".join(['A', 'B', 'C'])
'A, B, C'
```

jfagoagas (changes requested):

prowler/providers/aws/services/apigateway/apigateway_restapi_authorizers_enabled/apigateway_restapi_authorizers_enabled.py
Fennerr: I have removed it

n4ch04:
Please @Fennerr  refer to https://github.com/prowler-cloud/prowler/pull/3183#issuecomment-1853881493

PR #3195, opened 2023-12-15
feat(securityhub): Send only FAILs but storing all in the output files

jfagoagas:
### Description

Send only FAILs to Security Hub but storing all in the output files using a new `--send-sh-only-fails` flag.

### License


MrCloudSec (approved):
🔝

PR #3216, opened 2023-12-22
fix(trustedadvisor): solve trustedadvisor check metadata

MrCloudSec:
### Context

An empty ResourceType is giving the following Security Hub error: `data.Resources[0].Type should NOT be shorter than 1 characters`

### Description

Add Other as Resource Type for not leaving it empty.

### License


jfagoagas (approved):
Good catch!

PR #3239, opened 2024-01-08
fix(organizations_scp_check_deny_regions): enhance check logic

MrCloudSec:
### Context

Enhance `organizations_scp_check_deny_regions` logic so not all the restricted regions in the SCP has to be in the config variable.

### Description

Only check that the regions in the config variable are in the SCP restrictions.

### License


jfagoagas (approved):
Nice one! Thanks @sergargar 🚀

PR #3259, opened 2024-01-11
fix(fms): handle list compliance status error

n4ch04:
### Context

In `fms` service we were not handling correctly response parsing of `list_compliance_status` api call

### Description

Ensure `EvaluationResults` key exists in api call answer and contains a non-empty list prior trying to access its items

### License

### DIFF
--- a/prowler/providers/aws/services/fms/fms_service.py
+++ b/prowler/providers/aws/services/fms/fms_service.py
@@ -69,7 +69,7 @@ class FMS(AWSService):
                     for fms_compliance_status in page.get(
                         "PolicyComplianceStatusList", []
                     ):
-                        compliance_status = ""
+                        compliance_status = None
                         if fms_compliance_status.get("EvaluationResults"):
                             compliance_status = fms_compliance_status.get(
                                 "EvaluationResults"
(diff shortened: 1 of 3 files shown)


prowler/providers/aws/services/fms/fms_service.py
```diff
@@ -69,13 +69,16 @@ def __list_compliance_status__(self):
                     for fms_compliance_status in page.get(
                         "PolicyComplianceStatusList", []
                     ):
+                        compliance_status = ""
+                        if fms_compliance_status.get("EvaluationResults"):
+                            compliance_status = fms_compliance_status.get(
+                                "EvaluationResults"
+                            )[0].get("ComplianceStatus")
```
MrCloudSec: are we handling the case that the compliance_status is None when the key ComplianceStatus does not exist?

n4ch04: We are setting its default value just before the `if` block as `""`, we can set it to `None` (since in python `str` are falsy values)

n4ch04: I'll do it and push it

MrCloudSec: I mean in the check

n4ch04: I'll write a test to see it

MrCloudSec (approved):
good catch!

PR #3262, opened 2024-01-11
chore(precommit): set trufflehog as command

n4ch04:
### Context

Pre-commit was launching a trufflehog container, using the command is faster and cleaner

### Description

Use trufflehog command instead of container

### License


jfagoagas (approved):
Good one!

PR #3263, opened 2024-01-11
fix(apigatewayv2_api_access_logging_enabled): Finding ID should be unique

jfagoagas:
### Description

Finding ID should be unique, changed from `{api.name}` to `{api.name}-{stage.name}`.

### License


MrCloudSec (approved):
good catch!

PR #3293, opened 2024-01-16
fix(acm): adding more details on remaining expiration days

estemendoza:
### Context

The `ACM Certificates Expiration` check produces evidence that is not very clear when the certificate is actually going to expire.

### Description

The evidence could be more clear regarding the certificates expiration days.

The current evidence looks like this

```
ACM Certificate {certificate.id} for {certificate.name} is about to expire in {DAYS_TO_EXPIRE_THRESHOLD} days.
```

The problem with this is that `DAYS_TO_EXPIRE_THRESHOLD` equals always to `7` and then the user couldn't really know from the evidence when exactly the certificate is going to expire or how many days has the certificate been expired.

My proposal is to have 2 types of evidence when the check fails:

- when the certificate is about to expire
  - `ACM Certificate blebleble for cert_name is about to expire in 6 days`
- when the certificate has expired
  - `ACM Certificate blebleble for cert_name has expired (30 days ago).`

### License

### DIFF
--- a/prowler/providers/aws/services/acm/acm_certificates_expiration_check/acm_certificates_expiration_check.py
+++ b/prowler/providers/aws/services/acm/acm_certificates_expiration_check/acm_certificates_expiration_check.py
@@ -19,7 +19,12 @@ class acm_certificates_expiration_check(Check):
                 report.resource_tags = certificate.tags
             else:
                 report.status = "FAIL"
-                report.status_extended = f"ACM Certificate {certificate.id} for {certificate.name} is about to expire in {DAYS_TO_EXPIRE_THRESHOLD} days."
+
+                if certificate.expiration_days < 0:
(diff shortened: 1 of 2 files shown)


prowler/providers/aws/services/acm/acm_certificates_expiration_check/acm_certificates_expiration_check.py
```diff
@@ -19,7 +19,12 @@ def execute(self):
                 report.resource_tags = certificate.tags
             else:
                 report.status = "FAIL"
-                report.status_extended = f"ACM Certificate {certificate.id} for {certificate.name} is about to expire in {DAYS_TO_EXPIRE_THRESHOLD} days."
+
```
jfagoagas:
```suggestion
```

jfagoagas (approved):
Totally agree @estemendoza, makes much more sense to include that level of detail in the check.

Thanks for creating the PR following our style guide and including new test cases, it's a pleasure to do reviews in this way.

🚀 🚀🚀

PR #3319, opened 2024-01-26
fix(organizations): Handle non existent policy

jfagoagas:
### Description

Handle JSON policy returned by `describe_policy` when an exception is caught.

### License

### DIFF
--- a/prowler/providers/aws/services/organizations/organizations_service.py
+++ b/prowler/providers/aws/services/organizations/organizations_service.py
@@ -151,9 +151,11 @@ class Organizations(AWSService):
         logger.info("Organizations - List Targets for policy: %s ...", policy_id)
 
         try:
-            targets_for_policy = self.client.list_targets_for_policy(
-                PolicyId=policy_id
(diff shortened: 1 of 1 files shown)


prowler/providers/aws/services/organizations/organizations_service.py
```diff
@@ -134,15 +133,19 @@ def __describe_policy__(self, policy_id):
...
+                )
         except Exception as error:
             logger.error(
                 f"{self.region} -- {error.__class__.__name__}[{error.__traceback__.tb_lineno}]: {error}"
             )
         finally:
-            return policy_content_json
+            return json.loads(policy_content)
 
     def __list_targets_for_policy__(self, policy_id):
```
MrCloudSec, line 150: You have to check in this function if policy_id is not None too.

jfagoagas: Done!

MrCloudSec (approved):
Good catch!

PR #3373, opened 2024-02-07
fix(defender): Manage 404 exception for "default" security contacts

puchy22:
### Context

The Defender service does not contemplate error 404 for "default" security contacts

`(404) Could not find security contact with name default for subscriptionId: XXXXX`

### Description

Manage exception 404 and add fail values for security contact

### License

### DIFF
--- a/prowler/providers/azure/services/defender/defender_service.py
+++ b/prowler/providers/azure/services/defender/defender_service.py
@@ -121,9 +121,9 @@ class Defender(AzureService):
         security_contacts = {}
         for subscription_name, client in self.clients.items():
             try:
+                security_contacts.update({subscription_name: {}})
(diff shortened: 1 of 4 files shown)


prowler/providers/azure/services/defender/defender_service.py
```diff
@@ -136,6 +137,25 @@ def __get_security_contacts__(self):
                         )
                     }
                 )
+            except HttpResponseError as error:
+                if error.status_code == 404:
+                    security_contacts[subscription_name].update(
+                        {
```
MrCloudSec, line 143: Please, add a test for this use case in the check `defender_additional_email_configured_with_a_security_contact`.

MrCloudSec (approved):
Looks great 😄

PR #3378, opened 2024-02-08
fix(alias): allow multiple check aliases

MrCloudSec:
### Context

When a Check Alias applies to multiple checks, Prowler only executes the first one.

### Description

Execute all the applicable check to the specific alias.

### License

### DIFF
--- a/prowler/lib/check/checks_loader.py
+++ b/prowler/lib/check/checks_loader.py
@@ -34,7 +34,9 @@ def load_checks_to_execute(
         for check, metadata in bulk_checks_metadata.items():
             # Aliases
             for alias in metadata.CheckAliases:
-                check_aliases[alias] = check
+                if alias not in check_aliases:
+                    check_aliases[alias] = []
+                check_aliases[alias].append(check)
 
             # Severities
             if metadata.Severity:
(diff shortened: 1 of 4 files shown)


prowler/lib/check/checks_loader.py
```diff
@@ -110,15 +112,20 @@ def update_checks_to_execute_with_aliases(
...
-        ):
-            # Remove input check name and add the real one
-            checks_to_execute.remove(input_check)
-            checks_to_execute.add(check_aliases[input_check])
-            print(
-                f"\nUsing alias {Fore.YELLOW}{input_check}{Style.RESET_ALL} for check {Fore.YELLOW}{check_aliases[input_check]}{Style.RESET_ALL}...\n"
-            )
-    return checks_to_execute
+    try:
+        new_checks_to_execute = checks_to_execute.copy()
```
jfagoagas, line 116: Why do you need this copy? If you use a `set` you are not going to have duplicates in that object.

jfagoagas (changes requested):

prowler/lib/check/checks_loader.py
MrCloudSec: Because I only want to modify the copied set, not the original one since it is being iterated.

MrCloudSec: If the original set is modified while iterating through it, it will throw an error.

jfagoagas: Why would that happen? We were removing checks from the `checks_to_execute`.

jfagoagas (approved):
Great one! Thanks @sergargar 🚀

PR #3404, opened 2024-02-16
test(aws): Add default Boto3 credentials

jfagoagas:
### Description

- Add default environment variables for testing and mocking with AWS Boto3 and [moto](https://github.com/getmoto/moto). Rationale: https://docs.getmoto.org/en/latest/docs/getting_started.html#recommended-usage
- Add [`pytest-env`](https://pypi.org/project/pytest-env/) plugin to set those environment variables

### License


MrCloudSec (approved):
🔝

PR #3431, opened 2024-02-22
feat(namespace): add `--namespaces` argument and solve bugs

MrCloudSec:
### Description

Add `--namespaces` argument to allow scanning one or more specific namespaces and solve the following bugs:

1. Extracted credentials and prints from the first context of the configuration.
2. Implemented the functionality to extract the correct context.
3. Implemented a timeout verification mechanism.
4. Resolved the issue of erroneous namespaces in credentials.
5. Addressed the duplication of services in the summary table.
6. Updated filename handling to support special characters such as ":" and "/" from the ARN.

### License


prowler/providers/kubernetes/kubernetes_provider_new.py
```diff
@@ -14,32 +14,59 @@ class KubernetesProvider(Provider):
     # TODO change class name from Provider to Provider
```
jfagoagas:
```suggestion
    # TODO change class name from Provider to Provider
```

jfagoagas (approved):

PR #3436, opened 2024-02-26
fix(docs): Add docs group to install

jfagoagas:
### Description

Add docs group to allow to install only that group.

### License


MrCloudSec (approved):
Good catch!

PR #3459, opened 2024-02-28
fix(gcp): remove Default Project ID requirement

MrCloudSec:
### Description

Since Prowler scans all accesible projects by default, there is no need now to have set a Default Project ID in your GCP credentials.

### License

### DIFF
--- a/prowler/providers/common/audit_info.py
+++ b/prowler/providers/common/audit_info.py
@@ -46,10 +46,8 @@ class Audit_Info:
         # Beautify audited profile, set "default" if there is no profile set
         try:
             getattr(audit_info.credentials, "_service_account_email")
-            profile = (
-                audit_info.credentials._service_account_email
-                if audit_info.credentials._service_account_email is not None
-                else "default"
+            profile = getattr(
+                audit_info.credentials, "_service_account_email", "default"
             )
         except AttributeError:
             profile = "default"
--- a/prowler/providers/common/audit_info.py
+++ b/prowler/providers/common/audit_info.py
@@ -44,13 +44,7 @@ class Audit_Info:
 
     def print_gcp_credentials(self, audit_info: GCP_Audit_Info):
         # Beautify audited profile, set "default" if there is no profile set
-        try:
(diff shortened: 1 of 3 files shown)


prowler/providers/common/audit_info.py
```diff
@@ -46,10 +46,8 @@ def print_gcp_credentials(self, audit_info: GCP_Audit_Info):
         # Beautify audited profile, set "default" if there is no profile set
         try:
             getattr(audit_info.credentials, "_service_account_email")
```
jfagoagas: Why this? I think now is no longer needed since the `getattr` is done in the next line.

prowler/providers/common/audit_info.py
```diff
@@ -46,10 +46,8 @@ def print_gcp_credentials(self, audit_info: GCP_Audit_Info):
...
-            profile = (
-                audit_info.credentials._service_account_email
-                if audit_info.credentials._service_account_email is not None
-                else "default"
+            profile = getattr(
+                audit_info.credentials, "_service_account_email", "default"
             )
         except AttributeError:
             profile = "default"
```
jfagoagas: Neither this line nor the `try`.

prowler/providers/gcp/gcp_provider.py
```diff
@@ -41,6 +38,10 @@ def __init__(
             # If not projects were input, all accessible projects are scanned by default
             self.project_ids = accessible_projects
 
+        # Set Default Project ID if not set in credentials
+        if not self.default_project_id:
+            self.default_project_id = self.project_ids[0]
```
jfagoagas, line 43: Why are you getting the first project as the default? Can we skip that part?

jfagoagas (changes requested):

prowler/providers/gcp/gcp_provider.py
MrCloudSec: Because the output file name needs the default_project_id.

jfagoagas (approved):
Great catch @sergargar !!

Thanks for keeping the level so high!!

PR #3484, opened 2024-03-04
fix(compliance): add default severity to Manual Mocked Metadata

MrCloudSec:
### Description

Since we add a Severity validator, we need to add a default severity to the Manual Mocked Metadata.

### License


jfagoagas (approved):
⭐

PR #3502, opened 2024-03-05
fix(compliance): set the provider dynamically in Manual checks

MrCloudSec:
### Description

Set the provider dynamically in Manual checks since it was set AWS by default.
Thanks @pedrooot for the catch!

### License


jfagoagas (approved):
Great catch!

PR #3529, opened 2024-03-12
feat(Azure): Support version `beta` of MSGraph API

puchy22:
### Context

In most of checks Entra service needs to use the version beta of the MSGraph API.

### Description

The service support a list of clients, specifications of which version on Entra service, and change some checks

### License


pedrooot (approved):
Nice job @puchy22 🤯

prowler/providers/azure/lib/service/service.py
```diff
@@ -5,13 +5,16 @@
 class AzureService:
     def __init__(
         self,
-        service: str,
+        services: list,
```
jfagoagas: Could we set here the following? To let callers know that services could be a list or a string.
```suggestion
        services: Union[list[object], object],
```

jfagoagas: Also it'd be great to add a comment to the init function to give more details about that.
Like:
```
def __init__():
"""
details line 1
details line 2
"""
```

prowler/providers/azure/lib/service/service.py
```diff
@@ -20,23 +23,44 @@ def __init__(
 
         self.audit_config = audit_info.audit_config
 
-    def __set_clients__(self, identity, credentials, service, region_config):
+    def __set_clients__(self, identity, credentials, services, region_config):
         clients = {}
         try:
-            if "GraphServiceClient" in str(service):
-                clients.update({identity.domain: service(credentials=credentials)})
+            if "GraphServiceClient" in str(services):
```
jfagoagas, line 36: For me it's a little strange to keep two different versions of the `clients` object whether it requires the `GraphServiceClient` or not. Could we have a common version for all? Like:
```json
"clients" : {
      "tenant_domain":{
         "client_name_1":"client_object_1",
         "client_name_2":"client_object_2"
      }
}
```

jfagoagas (changes requested):
That's great @puchy22 👏 

It's a good solution to have different versions of the same client.

prowler/providers/azure/lib/service/service.py
puchy22: It should not be a string, the most common type of variable is something like this `<class 'azure.mgmt.containerservice._container_service_client.ContainerServiceClient'>`

jfagoagas: Good point, in that case it will be better to set something like `Union[list[object], object]` and explain that in the comment.

jfagoagas: btw `from typing import Union`

prowler/providers/azure/lib/service/service.py
```diff
@@ -5,13 +5,21 @@
...
     def __init__(
         self,
-        service: str,
+        services: list,
         audit_info: Azure_Audit_Info,
     ):
+        """
+        services must be a list of Azure clients to use in service (for now only GraphServiceClient support multiple client for use "v1" and "beta" versions)
+        If the service is not in a list, it will be converted to a list of one element
+        If a list is passed, and it not contains GraphServiceClient, it will be used only the last client of the list for all subscriptions
```
jfagoagas:
```suggestion
        services argument must be a list of Azure client's class to be set for the current service (for now only GraphServiceClient supports multiple clients to use "v1" and "beta" versions)
        If the service not a list, thus a class object, it will be converted to a list of one element with that class object.
        If a list is passed, and it not contains GraphServiceClient, it will be used only the last client of the list for all subscriptions
```

jfagoagas: For the last sentence `If a list is passed, and it not contains GraphServiceClient, it will be used only the last client of the list for all subscriptions`. I think we should set up all the clients passed in the list instead of just setting the first or the last, what do you think?

prowler/providers/azure/lib/service/service.py
```diff
@@ -20,23 +28,32 @@ def __init__(
...
+                if "msgraph_beta." in str(client_service):
+                    clients[identity.domain].update(
+                        {"beta": client_service(credentials=credentials)}
+                    )
+                elif "msgraph." in str(client_service):
+                    clients[identity.domain].update(
+                        {"v1": client_service(credentials=credentials)}
                     )
+                else:
+                    for display_name, id in identity.subscriptions.items():
```
jfagoagas: If I go through this else what would be the clients? One key per service and another with the tenant?

jfagoagas (changes requested):

prowler/providers/azure/lib/service/service.py
puchy22: The first time that I was developing this I setup to store in a subscription multiple clients service if it is passed, the problem is it breaks all Azure services done by now. This is because all services assume that there is only one client per subscription.

For example this is app_insights service when it search for components the for loop is in subscriptions but it doesn't iterate over clients again because it assume that can only be one.
```
def __get_components__(self):
        logger.info("AppInsights - Getting components...")
        components = {}

        for subscription_name, client in self.clients.items():
            try:
                components_list = client.components.list()
                components.update({subscription_name: {}})

                for component in components_list:
                    components[subscription_name].update(
                        {
                            component.app_id: Component(
                                resource_id=component.id, resource_name=component.name
                            )
                        }
                    )
            except Exception as error:
                logger.error(
                    f"Subscription name: {subscription_name} -- {error.__class__.__name__}[{error.__traceback__.tb_lineno}]: {error}"
                )

        return components
```

prowler/providers/azure/lib/service/service.py
puchy22: Yes, there would be one key is a tenant with MSGraphs clients and the rest of keys would be the subscriptions of that tenant with its respective client. The idea is the else is for any other possible client, the problem is if there are more than one that are not MSGraph. I do in this way to be compatible with all other services that are done yet and if for some reason anybody makes a new service and use for example [ContainerClient, MSGraph, MSGraphBeta] it doesn't break anything and it still working.

The problems could come when in a new service you use for example: [ContainerClient, MSGraph, MSGraphBeta, VMsClient], but that is a very strange situation in my opinion, the most normal thing is to only need one client per service.

puchy22:
This PR will be keep in draft until a "beta" endpoint must be necessary for any check

PR #3537, opened 2024-03-14
fix(iam): ignore Root User in iam_user_mfa_enabled_console_access

### LINKED ISSUE #3519: [Bug]: iam_user_mfa_enabled_console_access - includes root user
### Steps to Reproduce

In the AWS Check iam_user_mfa_enabled_console_access, the results include a finding for the root user (since it is included in the credential report), the `Status Extended` field states: "User <root_account> does not have Console Password enabled." which is factually incorrect, as the root user can always login to the console.

### Expected behavior

Exclude the root user from this check. There is a specific check for root MFA. 

I don't have the logs. 

### Environment Resource

ECS

### OS used

Amazon Linux I think

### Prowler version

3.13.0

### Pip version

pip 23.2.1 

### Context

_No response_

MrCloudSec:
### Context

In the AWS Check iam_user_mfa_enabled_console_access, the results include a finding for the root user (since it is included in the credential report), the Status Extended field states: "User <root_account> does not have Console Password enabled." which is factually incorrect, as the root user can always login to the console.

### Description

Exclude the root user from this check. There is a specific check for root MFA.
Thanks @jchrisfarris for the catch!

### License

### DIFF
--- a/prowler/providers/aws/services/iam/iam_user_mfa_enabled_console_access/iam_user_mfa_enabled_console_access.py
+++ b/prowler/providers/aws/services/iam/iam_user_mfa_enabled_console_access/iam_user_mfa_enabled_console_access.py
@@ -7,12 +7,12 @@ class iam_user_mfa_enabled_console_access(Check):
         findings = []
         response = iam_client.credential_report
         for user in response:
-            report = Check_Report_AWS(self.metadata())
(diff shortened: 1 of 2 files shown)


jfagoagas (approved):
🤩

PR #3570, opened 2024-03-19
chore(muted): handle new Muted status

MrCloudSec:
### Description

Handle new Muted status in check with `mute_non_default_regions` variable.

### License

### DIFF
--- a/prowler/lib/outputs/json_asff/json_asff.py
+++ b/prowler/lib/outputs/json_asff/json_asff.py
@@ -85,7 +85,7 @@ def fill_json_asff(provider, finding):
                 compliance_summary.append(item)
 
         # Ensures finding_status matches allowed values in ASFF
-        finding_status = generate_json_asff_status(finding.status)
+        finding_status = generate_json_asff_status(finding.status, finding.muted)
 
         json_asff_output = Check_Output_JSON_ASFF(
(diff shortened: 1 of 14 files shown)


prowler/lib/outputs/json_asff/json_asff.py
```diff
@@ -11,16 +11,17 @@
 from prowler.lib.utils.utils import hash_sha512
 
 
-def generate_json_asff_status(status: str) -> str:
+def generate_json_asff_status(status: str, muted: bool = False) -> str:
```
jfagoagas, line 14: - Where are you calling this function with the new muted field?
- Please add a test to check the new behaviour?

jfagoagas, line 14: Where are you calling this function with the new muted field?
Please add a test to check the new behaviour?

jfagoagas (changes requested):

prowler/lib/outputs/json_asff/json_asff.py
MrCloudSec: I added where it is needed, good catch. There is already a test for this.

jfagoagas (dismissed):

jfagoagas (approved):

PR #3574, opened 2024-03-20
fix(compliance): fix csv output for framework Mitre Attack

pedrooot:
### Description

The output format of the csv for Mitre Attack framework had some line breaks that were breaking the output

### License

### DIFF
--- a/prowler/lib/outputs/compliance/mitre_attack_aws.py
+++ b/prowler/lib/outputs/compliance/mitre_attack_aws.py
@@ -36,10 +36,10 @@ def write_compliance_row_mitre_attack_aws(
             attributes_categories += attribute.Category + ", "
             attributes_values += attribute.Value + ", "
             attributes_comments += attribute.Comment + ", "
-        attributes_aws_services = attributes_aws_services[:-2]
-        attributes_categories = attributes_categories[:-2]
-        attributes_values = attributes_values[:-2]
-        attributes_comments = attributes_comments[:-2]
+        attributes_aws_services = attributes_aws_services.rstrip(", ")
+        attributes_categories = attributes_categories.rstrip(", ")
+        attributes_values = attributes_values.rstrip(", ")
+        attributes_comments = attributes_comments.rstrip(", ")
         compliance_row = Check_Output_MITRE_ATTACK(
             Provider=finding.check_metadata.Provider,
             Description=compliance.Description,
--- a/prowler/lib/outputs/compliance/mitre_attack_aws.py
+++ b/prowler/lib/outputs/compliance/mitre_attack_aws.py
@@ -27,19 +27,18 @@ def write_compliance_row_mitre_attack_aws(
         requirement_description = requirement.Description
         requirement_id = requirement.Id
         requirement_name = requirement.Name
-        attributes_aws_services = ""
-        attributes_categories = ""
-        attributes_values = ""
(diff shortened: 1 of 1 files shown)


MrCloudSec (dismissed):
Great catch!

prowler/lib/outputs/compliance/mitre_attack_aws.py
```diff
@@ -32,10 +32,14 @@ def write_compliance_row_mitre_attack_aws(
...
-            attributes_values += attribute.Value + "\n"
-            attributes_comments += attribute.Comment + "\n"
+            attributes_aws_services += attribute.AWSService + ", "
+            attributes_categories += attribute.Category + ", "
+            attributes_values += attribute.Value + ", "
+            attributes_comments += attribute.Comment + ", "
+        attributes_aws_services = attributes_aws_services[:-2]
+        attributes_categories = attributes_categories[:-2]
+        attributes_values = attributes_values[:-2]
+        attributes_comments = attributes_comments[:-2]
```
jfagoagas: Why are you doing this? Could you add a comment to explain it?

jfagoagas (changes requested):

prowler/lib/outputs/compliance/mitre_attack_aws.py
pedrooot: Hi! This part of the code deletes the ", " from the end of the string. I explain it with an example:

When the output has 4 attributes in aws_services it will create an string like this one: `service1, service2, service3, service4, `. This line has 2 extra characters that are: ", " and if the code from the line 39 is applied, these extra characters will be removed.

jfagoagas: For that you can use `.rstrip(',')` , it's more readable than playing with static indexes.

jfagoagas:
Is this PR just for Prowler 4? Should we fix that in v3?

prowler/lib/outputs/compliance/mitre_attack_aws.py
pedrooot: Great! thanks for your review

pedrooot:
I think it should be fixed in both

prowler/lib/outputs/compliance/mitre_attack_aws.py
```diff
@@ -36,10 +36,10 @@ def write_compliance_row_mitre_attack_aws(
...
             attributes_values += attribute.Value + ", "
             attributes_comments += attribute.Comment + ", "
-        attributes_aws_services = attributes_aws_services[:-2]
-        attributes_categories = attributes_categories[:-2]
-        attributes_values = attributes_values[:-2]
-        attributes_comments = attributes_comments[:-2]
+        attributes_aws_services = attributes_aws_services.rstrip(", ")
+        attributes_categories = attributes_categories.rstrip(", ")
+        attributes_values = attributes_values.rstrip(", ")
+        attributes_comments = attributes_comments.rstrip(", ")
```
jfagoagas: What if you do a `", ".join(<attribute>`, maybe it's easier within the loop and with that you won't need the `rstrip`. Up to you, just an improvement.

Please, open up another PR to fix this in V3, thanks!

jfagoagas (changes requested):

jfagoagas (approved):
👍

PR #3590, opened 2024-03-22
fix(securityhub): Add validation and handle errors

jfagoagas:
### Description

- Add validation for the `Description` and `Remediation.Recommendation.Text`
- Create the models for `Remediation` and `Recommendation`
- Handle errors when calling the Security Hub API
- Improve tests

### License

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #3593, opened 2024-03-22
fix(securityhub): Remove region from exception match

jfagoagas:
### Description

Remove `in region {region}` since that part may not be present in the API error.

This PR doesn't need backport since it's covered here https://github.com/prowler-cloud/prowler/pull/3590

### License


MrCloudSec (approved):
🎖️

PR #3599, opened 2024-03-23
fix(metadata): change ResourceType Type for AWS Inline Policy Check

gabrielsoltz:
### Context

Hello Heroes! 

The check `iam_inline_policy_no_administrative_privileges` is incorrectly defined as ASFF Type `AwsIamPolicy`... 
Inline Policies are not an AWS Resource itself. Instead, they are part of an IAM Role.
The correct Type for this check is `AwsIamRole`; we are evaluating IAM Inline policies in a Role. 
This became obvious when comparing with the ID that we have as the output of the same check, it's the IAM Role ARN (`arn:aws:iam::XXXXXXXX:role/ROLE_NAME`).

Thanks! 

### License


MrCloudSec (approved):
Good catch! Thanks @gabrielsoltz

PR #3618, opened 2024-03-27
fix(aws): break loop after FAIL in SQS and SNS checks

kagahd:
### Context

Both checks `sns_topics_not_publicly_accessible` and `sqs_queues_not_publicly_accessible` may report "FAIL" even though their report message states that the resource is **not** public.
This happens because Prowler iterates over **all** statements in the policy and does not break the loop in case of "FAIL". Therefore, if Prowler encounters thereafter a policy statement without security findings, the report message changes to "is not public" although the status remains "FAIL".

### Description

The bugfix breaks the loop in case of "FAIL" so that succeeding statements will not change the report message anymore.
To summarize, one found "FAILed" statement within the policy is sufficient to "FAIL" the checked resource. 

### License


MrCloudSec (approved):
Really great catch @kagahd, thanks 🚀

PR #3635, opened 2024-04-02
chore(compliance): only execute all compliances in normal execution

MrCloudSec:
### Description
Only execute all compliances in a normal execution of Prowler.

### License

### DIFF
--- a/prowler/__main__.py
+++ b/prowler/__main__.py
@@ -66,6 +66,16 @@ def prowler():
     severities = args.severity
     compliance_framework = args.compliance
     custom_checks_metadata_file = args.custom_checks_metadata_file
+    default_execution = (
+        not checks
+        and not services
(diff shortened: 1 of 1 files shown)


prowler/__main__.py
```diff
@@ -305,7 +311,14 @@ def prowler():
             global_provider.output_options,
         )
 
-        if findings:
+        if (
```
jfagoagas: Please include a comment here to know the reason behind that or create a new attribute based on that conditional not to repeat it, up to you.

jfagoagas (changes requested):

prowler/__main__.py
MrCloudSec: Done!

jfagoagas (approved):
😍

PR #3689, opened 2024-04-08
fix(k8s): sanitize context syntax only for output file names

MrCloudSec:
### Description

Sanitize context syntax only for output file names, since we want the full EKS ARN for the rest of the locations.

### License


jfagoagas (approved):
🔝

PR #3706, opened 2024-04-08
chore(Azure): Optimize Entra service to use async funcs

puchy22:
### Context

The implementation of Entra use `await` sentencies but execute all functions in order, this PR change to execute it concurrently.

### Description

Change the Entra service to use the `gather` function to run all functions concurrently.

### License

### DIFF
--- a/prowler/providers/azure/services/entra/entra_service.py
+++ b/prowler/providers/azure/services/entra/entra_service.py
@@ -22,6 +22,7 @@ class Entra(AzureService):
 
         loop = get_event_loop()
 
+        # Get users first alone because it is a dependency for other attributes
         self.users = loop.run_until_complete(self.__get_users__())
 
         attributes = loop.run_until_complete(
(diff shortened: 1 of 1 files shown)


prowler/providers/azure/services/entra/entra_service.py
```diff
@@ -19,26 +19,30 @@
...
+                self.__get_authorization_policy__(),
+                self.__get_group_settings__(),
+                self.__get_security_default__(),
+                self.__get_named_locations__(),
+                self.__get_directory_roles__(),
+                self.__get_conditional_access_policy__(),
+            )
         )
 
+        self.authorization_policy = attributes[0]
```
jfagoagas, line 39: Is there any way to get this as a map? Fixed items in list are dangerous sometimes

puchy22: I know that is not the best way to assign, but the only function that is valid in this case is `run_until_complete` because it returns the actual values of attributes and no coroutines like other functions of asyncio. It gives it back in this format, an array ordered with the return values functions. At the moment I can't think of a better way to assign these values.

jfagoagas: That's fine, if this is know and controlled, thanks @puchy22

jfagoagas (approved):
Great improvement! 👏

PR #3741, opened 2024-04-09
feat(gcp): improve Google Projects scan customization

MrCloudSec:
### Description

Improve Google Projects scan customization by:

- Allowing asterisk  `*` in `--project-ids` flag.
image

- Adding `--list-project-ids` flag to allow the user to copy and paste values and know the accessible projects to be scanned.
image

- Adding `--excluded-project-ids` flag to allow the user to exclude projects to be scanned (it also accepts asterisk `*`).
image

### License

### DIFF
--- a/prowler/providers/gcp/gcp_provider.py
+++ b/prowler/providers/gcp/gcp_provider.py
@@ -1,4 +1,5 @@
 import os
+import re
 import sys
 
 from colorama import Fore, Style
@@ -25,6 +26,7 @@ class GcpProvider(Provider):
     _type: str = "gcp"
     _session: Credentials
     _project_ids: list
+    _excluded_project_ids: list
     _identity: GCPIdentityInfo
     _audit_config: dict
     _output_options: GCPOutputOptions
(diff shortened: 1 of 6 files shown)


prowler/providers/gcp/gcp_provider.py
```diff
@@ -35,33 +37,58 @@ class GcpProvider(Provider):
...
             for project_id, project in accessible_projects.items():
                 self._projects[project_id] = project
                 self._project_ids.append(project_id)
 
+        # Remove excluded projects if any input
+        if excluded_project_ids:
+            for excluded_project in excluded_project_ids:
+                for project_id in self._project_ids:
+                    if self.is_project_matching(excluded_project, project_id):
+                        self._excluded_project_ids.append(project_id)
```
jfagoagas, line 75: Why don't you do the following three lines at line 75? I mean if you know that the project wants to be excluded you don't need to iterate again through all the `self._excluded_project_ids`, you can just do the `pop` and `remove` right?

prowler/providers/gcp/gcp_provider.py
```diff
@@ -95,6 +122,10 @@ def projects(self):
     def project_ids(self):
         return self._project_ids
 
+    @property
+    def excluded_project_ids(self):
```
jfagoagas, line 126: Are you going to use this value outside the provider? If not you don't need to expose it as a property.

jfagoagas (changes requested):
That's a great improvement @sergargar 👏 

Please review my comments, thanks!

prowler/providers/gcp/gcp_provider.py
MrCloudSec: Yes, I am using it when printing the credentials.

prowler/providers/gcp/gcp_provider.py
MrCloudSec: No, because the excluded_project could have an asterisk and apply to one or more project ids.

jfagoagas: Right, thanks for clarifying this.

jfagoagas (approved):
🤩

jchrisfarris:
Thank you! you all rock!

PR #3751, opened 2024-04-10
fix(compliance): Add muted info to compliance outputs

pedrooot:
### Context

Muted info about findings is added to compliance outputs

### Description

Now, muted will appear in compliance-results: 
Screenshot 2024-04-10 at 17 13 43

Also, csv outputs for compliance will have the column "MUTED"

*If an finding is muted, it won´t add value in fail/pass column*

### License


jfagoagas (approved):
Thanks for fixing this and taking care of the code, keep raising the bar 👏

PR #3805, opened 2024-04-17
fix(utils): import libraries when needed

MrCloudSec:
### Description

Import libraries when needed since they may be not supported by the OS.

### License

### DIFF
--- a/prowler/lib/utils/utils.py
+++ b/prowler/lib/utils/utils.py
@@ -1,5 +1,12 @@
 import json
 import os
+
+try:
+    import grp
+    import pwd
+except ImportError:
+    pass
+
 import re
 import sys
 import tempfile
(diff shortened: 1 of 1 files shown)


pedrooot (dismissed):
Thanks for the fix! @sergargar

prowler/lib/utils/utils.py
```diff
@@ -144,6 +142,9 @@ def is_owned_by_root(file_path: str) -> bool:
         bool: True if owned by root, False otherwise or None if file does not exist.
     """
     try:
+        import grp
```
jfagoagas: I prefer to use the `try` except logic at the beginning of the file, including imports in the code could introduce problems.


```
try:
    import grp
    import pwd
except ImportError:
    pass
```

jfagoagas (changes requested):

jfagoagas (approved):

PR #3806, opened 2024-04-17
fix(dashboard): fix error in windows for csvreader

pedrooot:
### Context

Fixes #3766

### Description

Windows was having some problems with the encoding from csv files, adding `encoding="utf-8"` when a csv file is opened fixes it.

### License


dashboard/pages/compliance.py
```diff
@@ -37,7 +37,7 @@
 
 csv_files = []
 for file in glob.glob(os.path.join(folder_path_compliance, "*.csv")):
-    with open(file, "r", newline="") as csvfile:
+    with open(file, "r", newline="", encoding="utf-8") as csvfile:
```
jfagoagas: Can you add the encoding in the config and the use it here? The same for the other file.

jfagoagas (changes requested):

dashboard/pages/compliance.py
pedrooot: Good catch, I´ve already done it

jfagoagas (approved):
Great catch!

PR #3811, opened 2024-04-18
fix(mutelist): Be called whatever the provider

jfagoagas:
### Context

Currently the mutelist is not being loaded if no provider is set, which means `aws` is executed. No trouble if a provider is set as the first argument.

### Description

- Move the mutelist logic to the provider class since this is common for all the providers.
- Create a new attribute to get the mutelist file path.
- Handle if the allowlist file is set for whatever the provider.

### License

### DIFF
--- a/prowler/__main__.py
+++ b/prowler/__main__.py
@@ -215,7 +215,7 @@ def prowler():
             checks_to_execute,
             global_provider,
             custom_checks_metadata,
-            getattr(args, "mutelist_file", None),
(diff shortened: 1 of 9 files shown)


n4ch04 (dismissed):

jfagoagas:
@sergargar I need you to test this PR when you get a chance. I put the `no-merge` until then.

Thanks!

MrCloudSec (approved):
Love this! Tested an reviewed!

PR #3813, opened 2024-04-18
chore(ec2): rename EC2 any port check

MrCloudSec:
### Description

Rename check `ec2_securitygroup_allow_ingress_from_internet_to_any_port` to `ec2_securitygroup_allow_ingress_from_internet_to_all_ports` for a more accurate naming.

Thanks @jchrisfarris for the idea!

### License


jfagoagas:
I'm going to put the no merge tag since we have to discuss what to do if we change a check ID since we have to include something different in the output, if not we are going to introduce breaking changes for the outputs.

MrCloudSec:
> I'm going to put the no merge tag since we have to discuss what to do if we change a check ID since we have to include something different in the output, if not we are going to introduce breaking changes for the outputs.

I think it would not be a breaking change since this is not the first alias and Prowler will work with the older check ID.

jfagoagas:
> > I'm going to put the no merge tag since we have to discuss what to do if we change a check ID since we have to include something different in the output, if not we are going to introduce breaking changes for the outputs.
> 
> I think it would not be a breaking change since this is not the first alias and Prowler will work with the older check ID.

You can break customer's automation if we change the `check_id` to a new one. The alias should store the new name but the original name needs to be left there.

MrCloudSec:
Closing this PR since it is not longer needed

PR #3821, opened 2024-04-19
fix(aws): Include record names for dangling IPs

### LINKED ISSUE #3820: [AWS] Include record name in Route 53 dangling IP Status Extended output
### New feature motivation

Prowler's `route53_dangling_ip_subdomain_takeover` check that identifies dangling IPs in Route 53 currently produces the following extended status for each output:

```
Route53 record <ip> in Hosted Zone <zone-information>. is a dangling IP which can lead to a subdomain takeover attack.
```

**The output does not mention which record set the finding is in relation to, despite being record-set specific**. Each Hosted Zone can contain multiple record sets.

If a hosted zone contains multiple record sets with the same IP, i.e.
- `A` record maps `foo.domain.com.` to a dangling IP (`123.123.123.123`), and
- `A` record maps `bar.domain.com.` to *the same dangling IP* (`123.123.123.123`)

⚠️ **Prowler will produce two *identical* findings**:

```
Route53 record 123.123.123.123 in Hosted Zone <zone-information>. is a dangling IP which can lead to a subdomain takeover attack.
Route53 record 123.123.123.123 in Hosted Zone <zone-information>. is a dangling IP which can lead to a subdomain takeover attack.
```

Though this is only the Status Extended attribute, the JSON information is also identical.

### Solution Proposed

Include the record set's `name` attribute in the `Status Extended` output.

For the example above, this would produce the following `Status Extended` outputs:

```
Route53 record 123.123.123.123 (name: foo.domain.com.) in Hosted Zone <zone-information>. is a dangling IP which can lead to a subdomain takeover attack.
Route53 record 123.123.123.123 (name: bar.domain.com.) in Hosted Zone <zone-information>. is a dangling IP which can lead to a subdomain takeover attack.
```

The current output is lacking. This is the best idea I could come up with.

### Additional context

_No response_

rieck-srlabs:
### Context

Fixes #3820

Improves the "Status Extended" output field for the `route53_dangling_ip_subdomain_takeover` check.

### Description

This PR adds record set names to the "Status Extended" output of the check that identifies dangling IPs in AWS Route 53. It fixes an issue where currently identical findings are produced if one dangling IP is mapped to multiple record names in Route 53 and improves the rule's output in all cases. See #3820 for more information.

### License

### DIFF
--- a/prowler/providers/aws/services/route53/route53_dangling_ip_subdomain_takeover/route53_dangling_ip_subdomain_takeover.py
+++ b/prowler/providers/aws/services/route53/route53_dangling_ip_subdomain_takeover/route53_dangling_ip_subdomain_takeover.py
@@ -38,18 +38,18 @@ class route53_dangling_ip_subdomain_takeover(Check):
                         ].tags
                         report.region = record_set.region
                         report.status = "PASS"
-                        report.status_extended = f"Route53 record {record} in Hosted Zone {route53_client.hosted_zones[record_set.hosted_zone_id].name} is not a dangling IP."
+                        report.status_extended = f"Route53 record {record} (name: {record_set.name}) in Hosted Zone {route53_client.hosted_zones[record_set.hosted_zone_id].name} is not a dangling IP."
                         # If Public IP check if it is in the AWS Account
                         if (
                             not ip_address(record).is_private
                             and record not in public_ips
                         ):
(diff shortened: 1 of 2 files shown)


jfagoagas (approved):
That's a great addition @rieck-srlabs 👏 Thanks for this contribution.

I've made some changes in the tests since I expected them to fail and the I realised the asserts wasn't good enough.

PR #3830, opened 2024-04-22
fix(metadata): remove semicolons from metadata texts

### LINKED ISSUE #3829: Semicolons used in descriptions break CSV import process - remediation required
### New feature motivation

Semicolons are used in sentences as per below example:

```
"remediation": 
{ 
"desc": "Ensure that S3 buckets have Logging enabled. CloudTrail data events can be used in place of S3 bucket logging. If that is the case; this finding can be considered a false positive.",
...
```

This is not a problem if JSON format is in use, but unfortunately, it also exists in the CSV report

```
;;"Ensure that S3 buckets have Logging enabled. CloudTrail data events can be used in place of S3 bucket logging. If that is the case; this finding can be considered a false positive.";
```
This means, in the import process (to a spreadsheet of some kind, or - in my case - to AWS Athena) those additional semicolons are interpreted as field dividers, breaking formatting/data.

### Solution Proposed

I don't have a universal solution that would suit everyone.
The first that comes to mind is, to encode semicolons to Unicode or HTML entity. But that might disrupt a lot of people's workflows.

Second approach: allow for a custom character to be defined as a field divider in CSV report. For example, I don't see `%` in CSV reports - if I could pass a flag to prowler to use `%` rather than `;` in CSV, I can modify my import functions / Athena schemas easily.

### Additional context

_No response_

MrCloudSec:
### Description

Remove semicolons from metadata texts to avoid conflicts with the CSV separator.

### License


jfagoagas (changes requested):
I think this is not needed. The `;` within double quotes is a valid way to include that character without breaking the CSV output.

jfagoagas (approved):
🔝

PR #3853, opened 2024-04-24
feat(json-ocsf): Add new fields for py-ocsf 0.1.0

pedrooot:
### Description

New py-ocsf-models version is released and new fields must be added:
* cloud.account.labels
* risk_details

### License


prowler/lib/outputs/common_models.py
```diff
@@ -40,7 +40,7 @@ class FindingOutput(BaseModel):
     # Optional since depends on permissions
     account_organization_name: Optional[str]
     # Optional since depends on permissions
-    account_tags: Optional[str]
+    account_tags: Optional[list[str]]
```
jfagoagas, line 43: I'm not sure about this change since in the CSV this is being shown as a comma separated string like `"key1:value1,key2:value2"`. Could you please review that for the CSV? Also check the tests there.

jfagoagas (changes requested):

prowler/lib/outputs/common_models.py
jfagoagas: To clarify, this change is fine but we need to verify that we are not breaking anything.

pedrooot: I totally agree with you, the best approach is to get a list of tags for json and tags separated by "," for csv outputs

jfagoagas (dismissed):
Great improvement @pedrooot 👏 🚀

jfagoagas (approved):

PR #3856, opened 2024-04-25
fix(KeyError): handle CacheSubnetGroupName keyError

pedrooot:
### Description

Handle case when when `CacheSubnetGroupName` don't exists in ElasticCache service

### License

### DIFF
--- a/prowler/providers/aws/services/elasticache/elasticache_service.py
+++ b/prowler/providers/aws/services/elasticache/elasticache_service.py
@@ -100,6 +100,6 @@ class Cluster(BaseModel):
     id: str
     arn: str
     region: str
-    cache_subnet_group_id: str
+    cache_subnet_group_id: Optional[str]
     subnets: Optional[list]
     tags: Optional[list]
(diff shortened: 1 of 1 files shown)


prowler/providers/aws/services/elasticache/elasticache_service.py
```diff
@@ -32,7 +32,9 @@ def __describe_cache_clusters__(self, regional_client):
                         id=cache_cluster["CacheClusterId"],
                         arn=cluster_arn,
                         region=regional_client.region,
-                        cache_subnet_group_id=cache_cluster["CacheSubnetGroupName"],
+                        cache_subnet_group_id=cache_cluster.get(
```
jfagoagas, line 35: You should set `cache_subnet_group_id: Optional[str]` in the model then.

jfagoagas (changes requested):

prowler/providers/aws/services/elasticache/elasticache_service.py
pedrooot: Done

jfagoagas (approved):
Great catch! 👏

PR #3859, opened 2024-04-25
fix(s3-integration): Store compliance outputs in their folder

pedrooot:
### Description

Since Prowler v4 is adding compliance outputs inside the `/output/compliance` path it is needed to make some changes to add the outputs in the correct path inside the s3 bucket

### License


jfagoagas (approved):
🤩

PR #3863, opened 2024-04-26
fix(html): Produce valid HTML output in Prowler v3

rieck-srlabs:
### Description

Prowler (v3) failed to produce _valid_ HTML outputs. There were
- tags not closed
- attributes not placed in double quotes
- text elements not properly HTML-escaped
- superfluous `</div>` elements

Browsers don't seem to mind, but I was experimenting with modifying the generated HTML and ran into issues with XML parsers not accepting the output.

### License

### DIFF
--- a/prowler/lib/outputs/html.py
+++ b/prowler/lib/outputs/html.py
@@ -1,3 +1,4 @@
+import html
 import importlib
 import sys
 from os import path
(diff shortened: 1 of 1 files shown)


pedrooot (approved):
Thanks for the changes! @rieck-srlabs  🔝 🚀

PR #3903, opened 2024-04-30
fix(vpc): solve AWS principal key error

MrCloudSec:
### Description

Handle the AWS KeyError when the VPC Endpoint policy does not contain an AWS principal.
If the principal is not an AWS principal, we don't need to check it since it could be a service or a federated principal.

### License

### DIFF
--- a/prowler/providers/aws/services/vpc/vpc_endpoint_connections_trust_boundaries/vpc_endpoint_connections_trust_boundaries.py
+++ b/prowler/providers/aws/services/vpc/vpc_endpoint_connections_trust_boundaries/vpc_endpoint_connections_trust_boundaries.py
@@ -55,10 +55,14 @@ class vpc_endpoint_connections_trust_boundaries(Check):
                             break
 
                     else:
-                        if isinstance(statement["Principal"]["AWS"], str):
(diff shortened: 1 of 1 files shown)


prowler/providers/aws/services/vpc/vpc_endpoint_connections_trust_boundaries/vpc_endpoint_connections_trust_boundaries.py
```diff
@@ -55,10 +55,14 @@ def execute(self):
...
-                            principals = [statement["Principal"]["AWS"]]
+                        if "AWS" in statement["Principal"]:
+                            if isinstance(statement["Principal"]["AWS"], str):
+                                principals = [statement["Principal"]["AWS"]]
+                            else:
+                                principals = statement["Principal"]["AWS"]
                         else:
-                            principals = statement["Principal"]["AWS"]
+                            # If the principal is not an AWS principal, we don't need to check it since it could be a service or a federated principal
+                            principals = []
```
jfagoagas, line 65: Can this be moved to another place? I think some checks are using the same, please give it a look.

jfagoagas (approved):

PR #3906, opened 2024-04-30
fix(ec2): handle non-existing private ip

MrCloudSec:
### Description

Handle non-existing Private IPs in Instances to solve `[KeyError[258]: 'PrivateIpAddress']`

### License


jfagoagas (approved):
Great catch!

PR #3917, opened 2024-05-03
fix(efs): change public EFS check metadata

### LINKED ISSUE #3913: [Bug]: efs_not_publicly_accessible check based on misunderstanding?
### Steps to Reproduce

This issue concerns the `efs_not_publicly_accessible` check. The bug report concerns the _validity / usefulness of the check itself, **not** its implementation_.

#### The (potential) problem

I don't understand the issue flagged by the `efs_not_publicly_accessible` check.

The check flags file systems that either
1. have no file system policy, or
2. that have a public file system policy with a principle as `*`.

It flags non-conforming resources with **critical** severity and suggests such resources are publicly accessible.

> [!IMPORTANT]
> I believe the **severity rating and associated risk is incorrect and results from a misunderstanding.**

#### Accessing EFS resources

My understanding of EFS is as follows: file systems are only accessible through dedicated **mount points** or **via AWS Transfer Family**:

- **Mount points**: Mount points are dedicated resources that need to be created and assigned a VPC. Mount points can [never have a public IP](https://docs.aws.amazon.com/efs/latest/ug/sg-information.html) and are therefore **never available over the Internet**. Clients can only connect to and mount a file system if they have network access to the mount point, which is restricted to VPCs. With a _public_ file system policy (or without a file system policy at all), EFS relies on UID / GID supplied by NFS clients. This is an inherent issue in NFS and not AWS specific. **It has little to do with file system policies in AWS.**

- **AWS Transfer Family**: For _public_ EFS resources, **[no access via AWS Transfer Family is possible](https://docs.aws.amazon.com/efs/latest/ug/access-control-block-public-access.html#block-efs-public-access-with-transferfamily)**.

Searching online for more information regarding public EFS systems, I found
- [this additional resource](https://plerion.com/resources/cloud-knowledge-base/aws/efs-file-systems-should-not-be-publicly-accessible). It looks like it is based directly on Prowler's rule and provides no additional information.
- The related [AWS documentation](https://docs.aws.amazon.com/efs/latest/ug/access-control-block-public-access.html), which does not indicate direct security risks associated with "public" EFS resources and highlights primarily usability issues to keep in mind.
- [This SO post](https://stackoverflow.com/questions/71402926/when-is-the-efs-policy-condition-elasticfilesystemaccessedviamounttarget-false) supports my understanding from above.

#### So what?

Reading the rule's **critical** risk (`EFS accessible to everyone could expose sensitive data to bad actors`) suggests to me that public EFS resources are akin to publicly accessible S3 buckets. However, **this is not the case**, as access over the Internet is never possible.

Based on this,
1. the severity rating of critical seems too high to me, and
2. the risk statement is misleading and confusing.

From my understanding, the only _bad actors_ that sensitive data could be exposed to are _**internal** actors_ with preexisting access that abuse NFS protocol flaws, similar to users with overprivileged policy assignments.

I'd love to know more about this check and know what I might have gotten wrong above. Please let me know if your understanding differs here and what your reasoning is as to how public EFS file systems are accessible to everyone.

not relevant

### Environment Resource

not relevant

### OS used

not relevant

### Prowler version

Prowler 4.1.0 (You are running the latest version, yay!)

### Pip version

pip 24.0

### Context

_No response_

MrCloudSec:
### Description

Speaking with AWS Support, they confirmed that the EFS would only be accessed to any principal within the VPC not to the public internet. That is why, we have to change the metadata as well as the severity. 

Thanks @rieck-srlabs for the catch! 

### License

### DIFF
--- a/prowler/providers/aws/services/efs/efs_not_publicly_accessible/efs_not_publicly_accessible.metadata.json
+++ b/prowler/providers/aws/services/efs/efs_not_publicly_accessible/efs_not_publicly_accessible.metadata.json
@@ -1,7 +1,7 @@
 {
   "Provider": "aws",
   "CheckID": "efs_not_publicly_accessible",
-  "CheckTitle": "Check if EFS have policies which allow access to everyone",
+  "CheckTitle": "Check if EFS have policies which allow access to any client within the VPC",
   "CheckType": [
     "Protect",
     "Data protection"
@@ -9,10 +9,10 @@
   "ServiceName": "efs",
   "SubServiceName": "",
   "ResourceIdTemplate": "arn:partition:service:region:account-id:resource-id",
-  "Severity": "critical",
+  "Severity": "medium",
   "ResourceType": "AwsEFSFileSystem",
-  "Description": "Check if EFS have policies which allow access to everyone",
-  "Risk": "EFS accessible to everyone could expose sensitive data to bad actors",
+  "Description": "Check if EFS have policies which allow access to any client within the VPC",
+  "Risk": "Restricting access to EFS file systems is a security best practice. Allowing access to any client within the VPC can lead to unauthorized access to the file system.",
   "RelatedUrl": "",
   "Remediation": {
     "Code": {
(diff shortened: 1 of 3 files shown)


jfagoagas (approved):
🤩

PR #3959, opened 2024-05-08
fix(s3): Handle if regional client is present

jfagoagas:
### Context

Something can happen while creating regional clients, so we need to handle that.

### Description

Pass the regional client, if present, based on the bucket region to each function.

### License


MrCloudSec (approved):
great catch!

PR #3961, opened 2024-05-08
fix(encoding): set utf-8 encoding for csv

pedrooot:
### Description

An encoding format is set to make outputs use"utf-8" without depending on the os that is running Prowler.

Thanks @jfagoagas for the help debugging the issue.
### License

### DIFF
--- a/dashboard/config.py
+++ b/dashboard/config.py
@@ -1,5 +1,4 @@
 import os
-import sys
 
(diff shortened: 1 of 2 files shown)


jfagoagas (approved):
Finally we've fixed that 🎉 🎉🎉🎉🎉

PR #3962, opened 2024-05-08
feat(ec2): add EC2 Security group check to verify if at least one port is opened

MrCloudSec:
### Context

The check `ec2_securitygroup_allow_ingress_from_internet_to_any_port` is failing only if the security has all ports opened… We needed to rename that check to `ec2_securitygroup_allow_ingress_from_internet_to_all_ports`, and create an accurate `ec2_securitygroup_allow_ingress_from_internet_to_any_port`.

### Description

- Rename `ec2_securitygroup_allow_ingress_from_internet_to_any_port` to `ec2_securitygroup_allow_ingress_from_internet_to_all_ports`.
- Create `ec2_securitygroup_allow_ingress_from_internet_to_any_port` to check if a security group has at least one port opened.
- Improve `check_security_group` by adding the case of inputting None ports to check if any port is opened.

### License

### DIFF
--- a/prowler/providers/aws/services/ec2/lib/security_groups.py
+++ b/prowler/providers/aws/services/ec2/lib/security_groups.py
@@ -28,12 +28,14 @@ def check_security_group(
         'ToPort': 123,
     }
 
-    @param procotol: Protocol to check.
+    @param protocol: Protocol to check. If -1, all protocols will be checked.
 
 
-    @param ports: List of ports to check. (Default: [])
+    @param ports: List of ports to check. If empty, any port will be checked. If None, any port will be checked. (Default: [])
 
     @param any_address: If True, only 0.0.0.0/0 or "::/0" will be public and do not search for public addresses. (Default: False)
+
+    @return: True if the security group has public access to the check_ports using the protocol
     """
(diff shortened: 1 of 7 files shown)


prowler/providers/aws/services/ec2/lib/security_groups.py
```diff
@@ -85,9 +90,12 @@ def check_security_group(
                             and ingress_rule["IpProtocol"] == protocol
                         ):
                             return True
-                # If no input ports check if all ports are open
+                # If empty input ports check if all ports are open
                 if len(set(ingress_port_range)) == 65536:
                     return True
+                # If None input ports check if any port is open
+                if ports is None:
```
jfagoagas, line 97: Can you explain this a little bit, I tried but I don't get it.

MrCloudSec: If None, any port will be checked. Which means that if there is at least one port open, it returns true. If empty ([]), all the ports must be open to return True.

jfagoagas (approved):
Great improvement @sergargar !! Thanks for taking care of this 🚀

PR #3976, opened 2024-05-10
feat(custom-checks-metadata): add new fields

### LINKED ISSUE #3923: Support for custom metadata fields
### New feature motivation

I'd love to be able to override the various elements of a check's metadata. Right now we can override Severity, but I'd also like to override:
1. CheckTitle so I can align specific CheckIDs to my policy statements
2. Description & Risk so I can also reference internal documentation
3. RelatedUrl so I can point to internal wikis and KBs
4. Remedation.Code.Terraform / NativeIAC to point to my internal golden modules
5. Same for Recommendation

I'm not sure what the "Notes" element is for, so perhaps that can be left off the list

### Solution Proposed

Extend support in the metadata.yaml file for the above elements

I could create my own lookup table and decorate/replace going into reports, but that wouldn't allow me to adjust what is sent to SecurityHub. 

### Additional context

IIRC we discussed this awhile back when the custom severities were implemented. Not sure if the Issue made it over into the v4 list of issues.

pedrooot:
### Context
Fixes #3923 
Custom checks metadata are useful to assing the needed values for metadata.

### Description

In this pr new fields are added for custom checks metadata:
* CheckTitle
* Description
* Risk
* RelatedUrl
* Remediation
   * Code
     * CLI
     * NativeIaC
     * Other
     * Terraform
   * Recommendation
     * Text
     * Url

Thanks @jchrisfarris for the recommendation

### License

### DIFF
--- a/prowler/lib/check/custom_checks_metadata.py
+++ b/prowler/lib/check/custom_checks_metadata.py
@@ -117,34 +117,12 @@ def update_check_metadata(check_metadata, custom_metadata):
             for attribute in custom_metadata:
                 if attribute == "Remediation":
                     for remediation_attribute in custom_metadata[attribute]:
-                        if remediation_attribute == "Code":
-                            for code_attribute in custom_metadata[attribute][
-                                remediation_attribute
-                            ]:
-                                try:
-                                    setattr(
-                                        check_metadata.Remediation.Code,
-                                        code_attribute,
-                                        custom_metadata[attribute][
-                                            remediation_attribute
-                                        ][code_attribute],
-                                    )
-                                except ValueError:
-                                    pass
-                        elif remediation_attribute == "Recommendation":
-                            for recommendation_attribute in custom_metadata[attribute][
-                                remediation_attribute
-                            ]:
-                                try:
-                                    setattr(
-                                        check_metadata.Remediation.Recommendation,
-                                        recommendation_attribute,
-                                        custom_metadata[attribute][
-                                            remediation_attribute
-                                        ][recommendation_attribute],
-                                    )
-                                except ValueError:
-                                    pass
+                        update_check_metadata_remediation(
+                            check_metadata,
+                            custom_metadata,
+                            attribute,
+                            remediation_attribute,
+                        )
                 else:
                     try:
                         setattr(check_metadata, attribute, custom_metadata[attribute])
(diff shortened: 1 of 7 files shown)


prowler/lib/check/custom_checks_metadata.py
```diff
@@ -18,9 +18,71 @@
...
                     },
-                    "required": ["Severity"],
+                    "required": [
+                        "Severity",
+                        "CheckTitle",
+                        "Description",
+                        "Risk",
+                        "RelatedUrl",
+                        "Remediation",
+                    ],
```
jfagoagas: Do we want to mark all as required? I think it should be optional. What do you think?

jchrisfarris: I don't think they should be required. If missing, then the values from the check's json should be used

pedrooot: Yes, I totally agree with that.

prowler/lib/check/custom_checks_metadata.py
```diff
@@ -18,9 +18,63 @@
...
+                                        "Terraform": {
+                                            "type": "string",
+                                        },
+                                    },
+                                    "required": [
+                                        "CLI",
+                                        "NativeIaC",
+                                        "Other",
+                                        "Terraform",
+                                    ],
```
jfagoagas: Are we sure of this? I think we want just to override the fields present in the custom metadata file.

prowler/lib/check/custom_checks_metadata.py
```diff
@@ -18,9 +18,63 @@
...
+                                    "type": "object",
+                                    "properties": {
+                                        "Text": {
+                                            "type": "string",
+                                        },
+                                        "Url": {
+                                            "type": "string",
+                                        },
+                                    },
+                                    "required": ["Text", "Url"],
```
jfagoagas: Are we sure of this? I think we want just to override the fields present in the custom metadata file.

prowler/lib/check/custom_checks_metadata.py
```diff
@@ -18,9 +18,63 @@
...
+                        "RelatedUrl": {
+                            "type": "string",
+                        },
+                        "Remediation": {
+                            "type": "object",
+                            "properties": {
+                                "Code": {
+                                    "type": "object",
+                                    "properties": {
+                                        "CLI": {
```
jfagoagas, line 40: Can you add some tests to modify for example the `Remediation.Code.CLI` and check that the rest of the original check's metadata is kept.

jfagoagas (changes requested):
Great job @pedrooot 👏 I left some comments to verify the logic.

prowler/lib/check/custom_checks_metadata.py
pedrooot: You are true, I think no field should be required.

prowler/lib/check/custom_checks_metadata.py
pedrooot: You are true, I think no field should be required.

prowler/lib/check/custom_checks_metadata.py
pedrooot: Done, also I've modified the `update_check_metadata` method since It wasn't working fine for sub-fields.

jfagoagas (approved):
🚀

PR #3977, opened 2024-05-10
fix PR 3962 ec2_securitygroup_allow_ingress_from_internet_to_any_port

kagahd:
### Context

As  mentioned on [prowlers slack channel](https://prowler-workspace.slack.com/archives/C0451NDLC4X/p1715349265915979?thread_ts=1715115373.241189&cid=C0451NDLC4X), I think the PR #3962 contained some semantic glitches and redundant code that this PR fixes.

### Description

I think the newly introduced check `ec2_securitygroup_allow_ingress_from_internet_to_any_port` does not need to call again for each security group, respectively for each of their rules, the method `check_security_group` because this had been done already beforehand when collecting all data about security groups and their rules.
However, I had to sligthly modify the method `check_security_group` in `security_groups.py` to return not only `True` if `ports` is `None` but also if `ports` is an empty list.
Doing so, `security_group.public_ports` becomes `True` on security groups without ingress filtering (0.0.0.0/0) and as such the check `ec2_securitygroup_allow_ingress_from_internet_to_any_port` does not need to iterate all security groups, respectively their rules again, to let FAIL the check in case of open ports.
Last but not least, the method `__describe_security_groups__` in `ec2_service.py` needs to test whether check `ec2_securitygroup_allow_ingress_from_internet_to_any_port` is being executed (and not `ec2_securitygroup_allow_ingress_from_internet_to_all_ports`) because `ec2_securitygroup_allow_ingress_from_internet_to_any_port` may set `all_public_ports` to `True`.

All 60 unit tests from within `aws_provider_test.py` were executed successfully.

### License

### DIFF
--- a/prowler/providers/aws/services/ec2/ec2_securitygroup_allow_ingress_from_internet_to_any_port/ec2_securitygroup_allow_ingress_from_internet_to_any_port.py
+++ b/prowler/providers/aws/services/ec2/ec2_securitygroup_allow_ingress_from_internet_to_any_port/ec2_securitygroup_allow_ingress_from_internet_to_any_port.py
@@ -22,14 +22,9 @@ class ec2_securitygroup_allow_ingress_from_internet_to_any_port(Check):
                 report.resource_id = security_group.id
                 report.resource_arn = security_group.arn
                 report.resource_tags = security_group.tags
-                if not security_group.public_ports:
-                    # Loop through every security group's ingress rule and check it
-                    for ingress_rule in security_group.ingress_rules:
-                        if check_security_group(
-                            ingress_rule, "-1", ports=None, any_address=True
-                        ):
-                            report.status = "FAIL"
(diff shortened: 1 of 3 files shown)


kagahd:
I withdraw my PR because it seems that I did not execute all related unit tests just by running `aws_provider_test.py`  because I thought this test will include all the unit tests located in its subfolders. So, if really **all** unit tests are run, some are failing because of my code changes.
Unfortunately I don't have yet an idea how to optimize  @sergargar's PR #3962  to avoid the code redundancies and to save some CPU cycles.
Also the "semantic glitches" I mentioned earlier are due to my mix of prowler v3 and prowler v4. In pure prowler v4 I don't see these "semantic glitches" anymore, so @sergargar's PR #3962 is fine for me.

kagahd:
I opened PR #3979 to address the code redundancies and to save some CPU cycles.

PR #3981, opened 2024-05-13
chore(parser): add --subservices tag

pedrooot:
### Context

Prowler allows to execute checks from a specific service using --service / --services tag

### Description

This pr adds the --subservice / --subservices tag in Prowler. This feature allows you to run checks for a specific subservice(s)

### License


MrCloudSec:
Good job @pedrooot ! But we will have to check all the subservices within the check's metadata. For example, in Kubernetes there are no subservices.

MrCloudSec:
And we need to differentiate the service for each subservice.

jfagoagas:
Are we going to allow the use of `--service` and `--subservice` in the same execution? I think it's important just to check an specific subservice instead of all.

jfagoagas:
We are closing this issue. We are not sure about the `subservice` concept since it was crafted by us and maybe it should be replaced by the `ResourceType`. We will circle around this again in the future.

Thanks for your effort @pedrooot 👏

PR #4007, opened 2024-05-15
chore(safety): ignore pip vulnerability

MrCloudSec:
### Context

Ignore Safety vulnerability https://data.safetycli.com/v/67599/97c/ since it does not apply as we do not use -extra-index-url flag with pip.

### Description

Add `--ignore 67599` flag to safety check.

### License


pedrooot (approved):
Great!!

PR #4018, opened 2024-05-16
chore(aws): Add failed_checks to track

kagahd:
### Context

This PR makes the AWS ec2 service and around around 15 prowler checks, that verify whether ports are open to the Internet, more concise. This PR relates to PR #3977 and PR #3979. 

### Description

I find it more concise to introduce a `state_manager` instead to share with a dozen of other checks a specific property `public_ports` of a specific check `ec2_securitygroup_allow_ingress_from_internet_to_all_ports` within the `ec2_service`.
I find the the code now clearer, cleaner, easier to extend and easier to maintain. It also saves some CPU cycles because  the method `check_security_group` in the `ec2_service` does not need to be called for each check anymore.

### License


jfagoagas:
Hi @kagahd I like your idea but we are not sure about including it at this time since we want to create an execution manager in top of Prowler's scan. Maybe it'll be easier for now to save something within the EC2 service, what do you think?

kagahd:
Hi @jfagoagas,

> Maybe it'll be easier for now to save something within the EC2 service, what do you think?

I preferred a separation of concerns by using the `state_manager` but we can put its functionality also in the ec2 service, if you prefer.
I just adapted the PR accordingly.

prowler/providers/aws/services/ec2/ec2_service.py
```diff
@@ -11,6 +11,17 @@
 
 ################## EC2
 class EC2(AWSService):
+    check_ids = set()
```
jfagoagas: What about putting this into the `AWSService` class? 

I'm concerned about the logic to set failed checks:
- I'd rename `check_ids` to `failed_check_ids`
- Don't we need to store the resource id and region for it?

kagahd: > What about putting this into the `AWSService` class?

Yes, it could be used then also by other services as well, not only the ec2 service.

> I'm concerned about the logic to set failed checks:
> 
> * Don't we need to store the resource id and region for it?

You are right, thanks for the reminder. I adapted the code accordingly.

> * I'd rename `check_ids` to `failed_check_ids`

It's now renamed to `failed_checks` because it contains more than a `check_id` since I added the resource id and region as well.

jfagoagas: This makes a ton of sense!

prowler/providers/aws/lib/service/service.py
```diff
@@ -21,6 +21,17 @@ class AWSService:
...
+    failed_checks = set()
+
+    @classmethod
+    def set_failed_check(cls, check_id=None, ressource_id=None, region=None):
+        if check_id is not None and ressource_id is not None and region is not None:
+            cls.failed_checks.add((check_id, ressource_id, region))
+
+    @classmethod
+    def is_failed_check(cls, check_id, ressource_id, region):
+        return (check_id, ressource_id, region) in cls.failed_checks
```
jfagoagas:
```suggestion
    @classmethod
    def set_failed_check(cls, check_id=None, resource_id=None, region=None):
        if check_id is not None and resource_id is not None and region is not None:
            cls.failed_checks.add((check_id, resource_id, region))

    @classmethod
    def is_failed_check(cls, check_id, resource_id, region):
        return (check_id, resource_id, region) in cls.failed_checks
```

jfagoagas (changes requested):
Could you add a test for each check while checking that `ec2_client.is_failed_check()` returns true, thus no findings will be generated?

The work is this PR is neat 💯

kagahd:
> Could you add a test for each check while checking that `ec2_client.is_failed_check()` returns true, thus no findings will be generated?

Sure, I will add them later today.

kagahd:
> Could you add a test for each check while checking that `ec2_client.is_failed_check()`

It's done.

prowler/providers/aws/lib/service/service.py
```diff
@@ -21,6 +21,17 @@ class AWSService:
     - Also handles if the AWS Service is Global
     """
 
+    failed_checks = set()
+
+    @classmethod
+    def set_failed_check(cls, check_id=None, resource_id=None, region=None):
+        if check_id is not None and resource_id is not None and region is not None:
+            cls.failed_checks.add((check_id.split(".")[-1], resource_id, region))
```
jfagoagas: Why do you need the `split(".")`?

prowler/providers/aws/lib/service/service.py
```diff
@@ -21,6 +21,17 @@ class AWSService:
...
+
+    @classmethod
+    def is_failed_check(cls, check_id, resource_id, region):
+        return (check_id.split(".")[-1], resource_id, region) in cls.failed_checks
```
jfagoagas: Why do you need the `split(".")`?

tests/providers/aws/services/ec2/ec2_securitygroup_allow_ingress_from_internet_to_any_port/ec2_securitygroup_allow_ingress_from_internet_to_any_port_test.py
```diff
@@ -359,3 +359,144 @@ def test_ec2_default_sgs_with_all_ports_check(self):
...
+                )
+                result_specific_port = check_specific_port.execute()
+
+                # One default sg per region
+                assert len(result_specific_port) == 3
+                # Search changed sg
+                for sg in result_specific_port:
+                    if sg.resource_id == default_sg_id:
+                        assert sg.status == "PASS"
+                        assert sg.region == AWS_REGION_US_EAST_1
```
jfagoagas, line 447: Please assert also the tags and details.

prowler/providers/aws/lib/service/service.py
jfagoagas: We need to add tests to this functions in the `TestAWSService` class.

prowler/providers/aws/lib/service/service.py
jfagoagas: We need to add tests to this functions in the `TestAWSService` class.

prowler/providers/aws/services/ec2/ec2_securitygroup_allow_ingress_from_internet_to_any_port/ec2_securitygroup_allow_ingress_from_internet_to_any_port.py
```diff
@@ -22,14 +25,22 @@ def execute(self):
                 report.resource_id = security_group.id
                 report.resource_arn = security_group.arn
                 report.resource_tags = security_group.tags
-                if not security_group.public_ports:
+                # only proceed if check "..._to_all_ports" did not run or did not FAIL to avoid to report open ports twice
+                if not ec2_client.is_failed_check(
+                    ec2_securitygroup_allow_ingress_from_internet_to_all_ports.__name__,
+                    security_group.id,
+                    security_group.region,
```
jfagoagas: What if we use just the ARN?

tests/providers/aws/services/ec2/ec2_securitygroup_allow_ingress_from_internet_to_all_ports/ec2_securitygroup_allow_ingress_from_internet_to_all_ports_test.py
jfagoagas, line 1: I think in this file we should test the scenario where the `EC2.set_failed_check` is called within the check and verify that the right values are set in the service.

tests/providers/aws/services/ec2/ec2_securitygroup_allow_ingress_from_internet_to_any_port/ec2_securitygroup_allow_ingress_from_internet_to_any_port_test.py
```diff
@@ -359,3 +359,144 @@ def test_ec2_default_sgs_with_all_ports_check(self):
...
+                security_group_id,
+                region,
+            )
+
+            # Ensure that is_failed_check returns True
+            assert EC2.is_failed_check(
+                ec2_securitygroup_allow_ingress_from_internet_to_any_port.__name__,
+                security_group_id,
+                region,
+            )
```
jfagoagas: I think this shouldn't be done here, since this is testing the `AWSService` logic.

tests/providers/aws/services/ec2/ec2_securitygroup_allow_ingress_from_internet_to_any_port/ec2_securitygroup_allow_ingress_from_internet_to_any_port_test.py
```diff
@@ -359,3 +359,144 @@ def test_ec2_default_sgs_with_all_ports_check(self):
...
+            check_all_ports = (
+                ec2_securitygroup_allow_ingress_from_internet_to_all_ports()
+            )
+            result_all_ports = check_all_ports.execute()
+
+            # Verify that the all ports check has detected the issue
+            assert any(
+                sg.status == "FAIL" and sg.resource_id == default_sg_id
+                for sg in result_all_ports
+            )
```
jfagoagas, line 421: Maybe is easier just to call `AWSService.set_failed_check` and then run the `ec2_securitygroup_allow_ingress_from_internet_to_any_port` and verify that the finding's status extended is the new one.

tests/providers/aws/services/ec2/ec2_securitygroup_allow_ingress_from_internet_to_any_port/ec2_securitygroup_allow_ingress_from_internet_to_any_port_test.py
jfagoagas: This comment applies to all the tests.

tests/providers/aws/services/ec2/ec2_securitygroup_allow_ingress_from_internet_to_any_port/ec2_securitygroup_allow_ingress_from_internet_to_any_port_test.py
jfagoagas: This comment applies to all the tests.

jfagoagas (changes requested):
@kagahd I left some comments about the tests and a tiny detail in the service logic, please let me know if you need help. Thanks!

prowler/providers/aws/lib/service/service.py
kagahd: > Why do you need the `split(".")`?

In order to split the class name from the path. 
Even though both Getter and Setter method pass the class name (instead of the full qualified name which contains the path), the Get method passes the qualified name while the Set method passes the class name.
This is how the value is set:
```
ec2_client.set_failed_check(
    self.__class__.__name__,
    security_group.id,
    security_group.region,
)
```
While this is how the value is retrieved:

```
ec2_client.is_failed_check(
    ec2_securitygroup_allow_ingress_from_internet_to_all_ports.__name__,
    security_group.id,
    security_group.region,
)
```
As you can see, both are using `__name__` and not `__qualname__`. However, The Get method seems to pass the full qualified class name which contains the path. 
To resolve it, it's cut by the `split(".")` method, so we are sure to have the class name without path in all cases. If you have a better solution, let me know please.

kagahd: > We need to add tests to this functions in the `TestAWSService` class.

Ok, no problem.

prowler/providers/aws/lib/service/service.py
kagahd: see above

tests/providers/aws/services/ec2/ec2_securitygroup_allow_ingress_from_internet_to_any_port/ec2_securitygroup_allow_ingress_from_internet_to_any_port_test.py
kagahd: > Please assert also the tags and details.

ok, no problem

kagahd: > This comment applies to all the tests.

sure

prowler/providers/aws/services/ec2/ec2_securitygroup_allow_ingress_from_internet_to_any_port/ec2_securitygroup_allow_ingress_from_internet_to_any_port.py
kagahd: > What if we use just the ARN?

I thought the same but when you suggested to use `resource_id` and `region` I thought you have valid reasons to do so, so I followed your advice. The ARN should be well enough since it fully qualifies the resource. It hat all the infos, and even more, compared to just `resource_id` and `region`.
So do you agree that I replace `resource_id` and `region` by the ARN?

jfagoagas: For sure, I said `region` since I just saw the `resource_id` but makes much more sense to place the `ARN`.

tests/providers/aws/services/ec2/ec2_securitygroup_allow_ingress_from_internet_to_all_ports/ec2_securitygroup_allow_ingress_from_internet_to_all_ports_test.py
kagahd: > I think in this file we should test the scenario where the `EC2.set_failed_check` is called within the check and verify that the right values are set in the service.

Ok, I'll add this.

tests/providers/aws/services/ec2/ec2_securitygroup_allow_ingress_from_internet_to_any_port/ec2_securitygroup_allow_ingress_from_internet_to_any_port_test.py
kagahd: Ok, if we move it to the `AWSService` test , I think we can remove all the `test_ec2_is_failed_check_when_no_findings` from all the specific port checks unit tests. What do you think?

jfagoagas: Right, I think that makes much more sense.

tests/providers/aws/services/ec2/ec2_securitygroup_allow_ingress_from_internet_to_any_port/ec2_securitygroup_allow_ingress_from_internet_to_any_port_test.py
kagahd: > Maybe is easier just to call `AWSService.set_failed_check` and then run the `ec2_securitygroup_allow_ingress_from_internet_to_any_port` and verify that the finding's status extended is the new one.

Yes, sure, it would be **easier** but it's cheating because it would not reflect the exact way how the values are set/get in real life by the app. 
To be consistent, we need to first execute the check `ec2_securitygroup_allow_ingress_from_internet_to_all_ports` by checking an open security group to all ports. And after that, we have to verify that a **specific port** check such as `ec2_securitygroup_allow_ingress_from_internet_to_any_port` is in status `PASS`, only because the previous run check `ec2_securitygroup_allow_ingress_from_internet_to_all_ports` is in status `FAIL`, in order to avoid to report the specific open ports twice. 
You need to do it this way to respect and verify by unit tests the call chain of the app.

kagahd:
Hi @jfagoagas 
could you please advise, why the `pr-lint-test` [tests](https://github.com/prowler-cloud/prowler/actions/runs/9174232628/job/25224637564?pr=4018) are failing after my last last commits from today? When I run them locally, only 2 of a total 3175 unit tests are failing (which were failing also before my todays commits):

```
poetry run pytest -n auto --cov=./prowler --cov-report=xml tests
...
=========================================================================================== short test summary info ===========================================================================================
FAILED tests/providers/aws/services/route53/route53_dangling_ip_subdomain_takeover/route53_dangling_ip_subdomain_takeover_test.py::Test_route53_dangling_ip_subdomain_takeover::test_hosted_zone_dangling_public_record - urllib.error.URLError: <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:997)>
FAILED tests/providers/aws/services/route53/route53_dangling_ip_subdomain_takeover/route53_dangling_ip_subdomain_takeover_test.py::Test_route53_dangling_ip_subdomain_takeover::test_hosted_zone_external_record - urllib.error.URLError: <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:997)>
=========================================================================== 2 failed, 3173 passed, 36 warnings in 399.73s (0:06:39) ===========================================================================
~
```
And these 2 failing unit tests are not from me. I didn't touch them.
Why the [github workflow](https://github.com/prowler-cloud/prowler/actions/runs/9174232628/job/25224637564?pr=4018) failed on so many unit tests?
Thanks for your help!

jfagoagas:
> Hi @jfagoagas could you please advise, why the `pr-lint-test` [tests](https://github.com/prowler-cloud/prowler/actions/runs/9174232628/job/25224637564?pr=4018) are failing after my last last commits from today? When I run them locally, only 2 of a total 3175 unit tests are failing (which were failing also before my todays commits):
> 
> ```
> poetry run pytest -n auto --cov=./prowler --cov-report=xml tests
> ...
> =========================================================================================== short test summary info ===========================================================================================
> FAILED tests/providers/aws/services/route53/route53_dangling_ip_subdomain_takeover/route53_dangling_ip_subdomain_takeover_test.py::Test_route53_dangling_ip_subdomain_takeover::test_hosted_zone_dangling_public_record - urllib.error.URLError: <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:997)>
> FAILED tests/providers/aws/services/route53/route53_dangling_ip_subdomain_takeover/route53_dangling_ip_subdomain_takeover_test.py::Test_route53_dangling_ip_subdomain_takeover::test_hosted_zone_external_record - urllib.error.URLError: <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:997)>
> =========================================================================== 2 failed, 3173 passed, 36 warnings in 399.73s (0:06:39) ===========================================================================
> ~
> ```
> 
> And these 2 failing unit tests are not from me. I didn't touch them. Why the [github workflow](https://github.com/prowler-cloud/prowler/actions/runs/9174232628/job/25224637564?pr=4018) failed on so many unit tests? Thanks for your help!

I think your forked repo wasn't up to date with Prowler's `master`. I've sync'ed your `master` branch and then merge that branch against this. Probably failing test happened due to this and also Github actions are degraded today so that could be also.

kagahd:
Hi @jfagoagas, the github workflow failed again with the same errors. Do you have another idea what the reason could be? In [line 155](https://github.com/prowler-cloud/prowler/actions/runs/9176744167/job/25232742488?pr=4018#step:15:196) it says `AttributeError: module 'prowler.providers.common' has no attribute 'common'` but other unit tests do the same what's marked in [line 105](https://github.com/prowler-cloud/prowler/actions/runs/9176744167/job/25232742488?pr=4018#step:15:106):
```
        with mock.patch(
            "prowler.providers.common.common.get_global_provider",
            return_value=aws_provider,
        ),
```
What's wrong with it?

jfagoagas:
> Hi @jfagoagas, the github workflow failed again with the same errors. Do you have another idea what the reason could be? In [line 155](https://github.com/prowler-cloud/prowler/actions/runs/9176744167/job/25232742488?pr=4018#step:15:196) it says `AttributeError: module 'prowler.providers.common' has no attribute 'common'` but other unit tests do the same what's marked in [line 105](https://github.com/prowler-cloud/prowler/actions/runs/9176744167/job/25232742488?pr=4018#step:15:106):
> 
> ```
>         with mock.patch(
>             "prowler.providers.common.common.get_global_provider",
>             return_value=aws_provider,
>         ),
> ```
> 
> What's wrong with it?

Yesterday we merged a PR which changes the location of the `get_global_provider` function, that was the issue. I've pushed a fix.

tests/providers/aws/services/ec2/ec2_securitygroup_allow_ingress_from_internet_to_all_ports/ec2_securitygroup_allow_ingress_from_internet_to_all_ports_test.py
```diff
@@ -332,3 +332,75 @@ def test_ec2_default_sgs_ignoring_vpc_in_use(self):
...
+
+            from prowler.providers.aws.services.ec2.ec2_securitygroup_allow_ingress_from_internet_to_all_ports.ec2_securitygroup_allow_ingress_from_internet_to_all_ports import (
+                ec2_securitygroup_allow_ingress_from_internet_to_all_ports,
+            )
+
+            check = ec2_securitygroup_allow_ingress_from_internet_to_all_ports()
+            result = check.execute()
+
+            # Verify set_failed_check was called with the correct arguments
+            mock_set_failed_check.assert_called_with(
```
jfagoagas, line 386: Pretty smart 💯

kagahd:
> Yesterday we merged a PR which changes the location of the `get_global_provider` function, that was the issue. I've pushed a fix.

That makes sense, thanks a bunch!

jfagoagas (approved):
Thanks for this super improvement again @kagahd 👏 

It's been a pleasure to work with you and review a PR with such a great level of quality 🚀

jfagoagas:
I have just one question left because I'm not sure, what was the previous behaviour when running an specific port check with the `public_ports` set to `True`? I thought we were not raising a finding but I'm not sure I have to test it, I mean to not modify that behaviour now.

kagahd:
> what was the previous behaviour when running an specific port check with the `public_ports` set to `True`? I thought we were not raising a finding

Exactly, you were not raising a finding, the status was "PASS". 
This behaviour is still the same, the status is still "PASS". However, I adapted the `report.status_extended` in order to mention that "has all ports open to the Internet and therefore was not checked against the specific XYZ port". Without my adaption of  `report.status_extended` you would still have a message like "Security group ABC does not have XYZ port open to the Internet." which is quite wrong because **all ports** are open.

jfagoagas:
> > what was the previous behaviour when running an specific port check with the `public_ports` set to `True`? I thought we were not raising a finding
> 
> Exactly, you were not raising a finding, the status was "PASS". This behaviour is still the same, the status is still "PASS". However, I adapted the `report.status_extended` in order to mention that "has all ports open to the Internet and therefore was not checked against the specific XYZ port". Without my adaption of `report.status_extended` you would still have a message like "Security group ABC does not have XYZ port open to the Internet." which is quite wrong because **all ports** are open.

Thanks for confirming that, maybe we can remove that `PASS` finding in the future, but at the end the important ones are the `FAIL` findings and not to raise 2 with the same information.

PR #4024, opened 2024-05-17
fix(mutelist): Handle exceptions before match

jfagoagas:
### Description

Handle a case when the exceptions where matching before the real mutelist match.

### License


drewkerrigan (approved):
Tested and confirmed locally, thanks @jfagoagas! 🚢

pedrooot (approved):
LGTM!

PR #4029, opened 2024-05-20
feat(ec2): add checks for EC2 instances with exposed ports to the internet

MrCloudSec:
### Description

Add checks `ec2_instance_port_X_exposed_to_internet` to check if there are EC2 Instances exposed to the internet through the specific port.

EC2 Instances with exposed port open to the Internet will be flagged as FAIL with a severity of medium if the instance has no public IP, high if the instance has a public IP but is in a private subnet, and critical if the instance has a public IP and is in a public subnet.

- [x] ec2_instance_port_mongodb_exposed_to_internet
- [x] ec2_instance_port_ftp_exposed_to_internet
- [x] ec2_instance_port_ssh_exposed_to_internet
- [x] ec2_instance_port_rdp_exposed_to_internet
- [x] ec2_instance_port_cassandra_exposed_to_internet
- [x] ec2_instance_port_elasticsearch_kibana_exposed_to_internet
- [x] ec2_instance_port_kafka_exposed_to_internet
- [x] ec2_instance_port_memcached_exposed_to_internet
- [x] ec2_instance_port_mysql_exposed_to_internet
- [x] ec2_instance_port_oracle_exposed_to_internet
- [x] ec2_instance_port_postgres_exposed_to_internet
- [x] ec2_instance_port_redis_exposed_to_internet
- [x] ec2_instance_port_sql_server_exposed_to_internet
- [x] ec2_instance_port_telnet_exposed_to_internet
- [x] ec2_instance_port_kerberos_exposed_to_internet
- [x] ec2_instance_port_ldap_exposed_to_internet
- [x] ec2_instance_port_cifs_exposed_to_internet

### License


prowler/providers/aws/services/ec2/ec2_instance_ssh_port_exposed_to_internet/ec2_instance_ssh_port_exposed_to_internet.py
```diff
@@ -0,0 +1,42 @@
...
+                        if sg.id == instance_sg["GroupId"]:
+                            for ingress_rule in sg.ingress_rules:
+                                if check_security_group(
+                                    ingress_rule, "tcp", check_ports, any_address=True
+                                ):
+                                    # The port is open, now check if the instance is in a public subnet with a public IP
+                                    report.status = "FAIL"
+                                    if instance.public_ip:
+                                        report.status_extended = f"Instance {instance.id} has SSH exposed to 0.0.0.0/0 on public ip address {instance.public_ip} but in private subnet {instance.subnet_id}."
+                                        report.check_metadata.Severity = "high"
```
jchrisfarris: Per my slack message, this seems like you're hardcoding the severity here, not letting the user set it in metadata.yaml.

MrCloudSec: It is because the severity changes depending on each scenario.

jfagoagas: Yes but in this case that feature wouldn't work.

MrCloudSec: That's right, maybe we can increase the severity in each scenario without hardcoding it.

prowler/providers/aws/services/ec2/ec2_instance_port_sqlserver_exposed_to_internet/ec2_instance_port_sqlserver_exposed_to_internet.py
```diff
@@ -0,0 +1,43 @@
...
+                                        report.check_metadata.Severity = "high"
+                                        if vpc_client.vpc_subnets[
+                                            instance.subnet_id
+                                        ].public:
+                                            report.status_extended = f"Instance {instance.id} has SQL Server exposed to 0.0.0.0/0 on public ip address {instance.public_ip} in public subnet {instance.subnet_id}."
+                                            report.check_metadata.Severity = "critical"
+                                    else:
+                                        report.status_extended = f"Instance {instance.id} has SQL Server exposed to 0.0.0.0/0 but with no public ip address."
+                                        report.check_metadata.Severity = "medium"
+                                    break
```
jfagoagas: Can you think on a way to have this just in one place? All the checks have the same logic.

jchrisfarris: I'm not sure it is that important to say "SQL Server", which means you could have the status extended just use {port} instead.

jfagoagas: I think the same, at the end any port can be bound to whatever service.

MrCloudSec: Okay, sounds good.

MrCloudSec: @jchrisfarris we will need to change the `check_security_group` function to return the ports that are open. I have included that task in our roadmap, until then, we will leave it as it is now saying the specific service that is open.

prowler/providers/aws/services/ec2/ec2_instance_port_cassandra_exposed_to_internet/ec2_instance_port_cassandra_exposed_to_internet.py
```diff
@@ -0,0 +1,39 @@
...
+            report = Check_Report_AWS(self.metadata())
+            report.region = instance.region
+            report.status = "PASS"
+            report.status_extended = f"Instance {instance.id} does not have Cassandra ports open to the Internet."
+            report.resource_id = instance.id
+            report.resource_arn = instance.arn
+            report.resource_tags = instance.tags
+            if instance.security_groups:
+                for instance_sg in instance.security_groups:
+                    for sg in ec2_client.security_groups:
```
jfagoagas: Could we change this to be a map? I think it is too much to loop all over the security groups in each check for each instance, we need to improve the performance here.

prowler/providers/aws/services/ec2/ec2_instance_port_cassandra_exposed_to_internet/ec2_instance_port_cassandra_exposed_to_internet.py
```diff
@@ -0,0 +1,39 @@
...
+                                ):
+                                    # The port is open, now check if the instance is in a public subnet with a public IP
+                                    report.status = "FAIL"
+                                    (
+                                        report.status_extended,
+                                        report.check_metadata.Severity,
+                                    ) = get_instance_public_status(
+                                        vpc_client.vpc_subnets, instance, "Cassandra"
+                                    )
+                                    break
```
jfagoagas: This `break` only aborts the `for ingress_rule in sg.ingress_rules:` loop, is that on purpose or do we need to add more checks to exit the outer loops?

prowler/providers/aws/services/ec2/ec2_instance_port_cassandra_exposed_to_internet/ec2_instance_port_cassandra_exposed_to_internet.py
jfagoagas: This comment applies to all the checks within this PR.

prowler/providers/aws/services/ec2/ec2_instance_port_cassandra_exposed_to_internet/ec2_instance_port_cassandra_exposed_to_internet.py
jfagoagas: This comment applies to all the checks within this PR.

prowler/providers/aws/services/ec2/ec2_instance_port_kerberos_exposed_to_internet/ec2_instance_port_kerberos_exposed_to_internet.metadata.json
```diff
@@ -0,0 +1,34 @@
...
+  "RelatedUrl": "",
+  "Remediation": {
+    "Code": {
+      "CLI": "",
+      "NativeIaC": "",
+      "Other": "",
+      "Terraform": ""
+    },
+    "Recommendation": {
+      "Text": "Modify the security group to remove the rule that allows ingress from the internet to TCP port 88 or 464 (Kerberos).",
```
jfagoagas:
```suggestion
      "Text": "Modify the security group to remove the rule that allows ingress from the internet to TCP port 88, 464, 749 or 750 (Kerberos).",
```

prowler/providers/aws/services/ec2/ec2_service.py
```diff
@@ -487,6 +468,8 @@ class Instance(BaseModel):
     http_tokens: Optional[str]
     http_endpoint: Optional[str]
     monitoring_state: str
+    security_groups: list[dict]
```
jfagoagas: Could you use `list[SecurityGroup]` here? With that you can map the attributes.

jfagoagas: Or maybe just store the security group ID.

prowler/providers/aws/services/ec2/ec2_service.py
```diff
@@ -99,13 +72,21 @@ def __describe_instances__(self, regional_client):
                                     image_id=instance["ImageId"],
                                     launch_time=instance["LaunchTime"],
                                     private_dns=instance["PrivateDnsName"],
```
jfagoagas, line 74: I'd say to add a `get()` to the `Instance` attributes in this function.

prowler/providers/aws/services/ec2/ec2_service.py
```diff
@@ -99,13 +72,21 @@ def __describe_instances__(self, regional_client):
...
+                                    ),
+                                    http_endpoint=instance.get(
+                                        "MetadataOptions", {}
+                                    ).get("HttpEndpoint"),
+                                    instance_profile=instance.get("IamInstanceProfile"),
+                                    monitoring_state=instance.get(
+                                        "Monitoring", {"State": "disabled"}
+                                    ).get("State", "disabled"),
+                                    security_groups=instance["SecurityGroups"],
+                                    subnet_id=instance["SubnetId"],
```
jfagoagas: Use the `get` also here.

prowler/providers/aws/services/ec2/lib/instance.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+def get_instance_public_status(
+    vpc_subnets: dict[VpcSubnet], instance: Instance, service: str
+) -> tuple:
+    """
+    Get the status and severity of an instance based on the service exposed to internet.
+    Args:
+        instance (Instance): The instance to check.
+        service (str): The service to verify if it is exposed to internet.
```
jfagoagas: Please update this accordingly.

prowler/providers/aws/services/ec2/lib/instance.py
```diff
@@ -0,0 +1,26 @@
...
+    Returns:
+        tuple: The status and severity of the instance status.
+    """
+    status = f"Instance {instance.id} has {service} exposed to 0.0.0.0/0 but with no public ip address."
```
jfagoagas:
```suggestion
    status = f"Instance {instance.id} has {service} exposed to 0.0.0.0/0 but with no public IP address."
```

prowler/providers/aws/services/ec2/lib/instance.py
```diff
@@ -0,0 +1,26 @@
...
+    severity = "medium"
+
+    if instance.public_ip:
+        status = f"Instance {instance.id} has {service} exposed to 0.0.0.0/0 on public ip address {instance.public_ip} but in private subnet {instance.subnet_id}."
```
jfagoagas:
```suggestion
        status = f"Instance {instance.id} has {service} exposed to 0.0.0.0/0 on public IP address {instance.public_ip} but it is placed in a private subnet {instance.subnet_id}."
```

prowler/providers/aws/services/ec2/lib/instance.py
```diff
@@ -0,0 +1,26 @@
...
+        severity = "high"
+        if vpc_subnets[instance.subnet_id].public:
+            status = f"Instance {instance.id} has {service} exposed to 0.0.0.0/0 on public ip address {instance.public_ip} in public subnet {instance.subnet_id}."
```
jfagoagas:
```suggestion
            status = f"Instance {instance.id} has {service} exposed to 0.0.0.0/0 on public IP address {instance.public_ip} but it is placed in a public subnet {instance.subnet_id}."
```

jfagoagas (changes requested):

prowler/providers/aws/services/ec2/ec2_service.py
MrCloudSec: Done, I am storing the ID.

prowler/providers/aws/services/ec2/ec2_service.py
MrCloudSec: We don't know if putting a None in the attribute will break something.

prowler/providers/aws/services/ec2/ec2_instance_port_cassandra_exposed_to_internet/ec2_instance_port_cassandra_exposed_to_internet.py
MrCloudSec: I agree with you, but changing the security groups to a map would need a lot of checks to be changed. As this is a performance improvement, we can add it in the future.

prowler/providers/aws/services/ec2/ec2_instance_port_cassandra_exposed_to_internet/ec2_instance_port_cassandra_exposed_to_internet.py
```diff
@@ -0,0 +1,38 @@
...
+                            ):
+                                # The port is open, now check if the instance is in a public subnet with a public IP
+                                report.status = "FAIL"
+                                (
+                                    report.status_extended,
+                                    report.check_metadata.Severity,
+                                ) = get_instance_public_status(
+                                    vpc_client.vpc_subnets, instance, "Cassandra"
+                                )
+                                break
```
jfagoagas, line 38: Review the comment about the loop break.

jfagoagas (changes requested):

jfagoagas (approved):

PR #4055, opened 2024-05-21
fix(gcp): handle projects API Call error

MrCloudSec:
### Description

Handle projects API Call error since the `Cloud Resource Manager API has not been used` problem was not being catch correctly.

### License


jfagoagas (approved):
🌟

PR #4068, opened 2024-05-23
chore(scan): New scan() function and fix an issue while scanning with only logs mode

jfagoagas:
### Context

We need to adapt our code for the future 😄 

### Description

- Create a new `scan()` function.
- Fix a bug while calling `execute_checks` in the `only_logs` mode.

### License


prowler/lib/check/check.py
```diff
@@ -591,8 +592,7 @@ def execute_checks(
                 check_findings = execute(
                     service,
                     check_name,
-                    global_provider.type,
-                    global_provider.identity,
+                    global_provider,
```
jfagoagas, line 595: This is to fix a bug, since the `execute` function only accepts the `global_provider` object.

pedrooot (approved):

PR #4080, opened 2024-05-23
chore(slack): change Slack channel name env variable

MrCloudSec:
### Description

Change Slack channel name environment variable from `SLACK_CHANNEL_ID` to `SLACK_CHANNEL_NAME` to avoid confusions.

### License

### DIFF
--- a/prowler/__main__.py
+++ b/prowler/__main__.py
@@ -248,16 +248,22 @@ def prowler():
     stats = extract_findings_statistics(findings)
 
     if args.slack:
-        if "SLACK_API_TOKEN" in environ and "SLACK_CHANNEL_ID" in environ:
+        if "SLACK_API_TOKEN" in environ and (
+            "SLACK_CHANNEL_NAME" in environ or "SLACK_CHANNEL_ID" in environ
+        ):
             _ = send_slack_message(
                 environ["SLACK_API_TOKEN"],
-                environ["SLACK_CHANNEL_ID"],
+                (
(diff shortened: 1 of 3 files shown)


prowler/__main__.py
```diff
@@ -248,16 +248,22 @@ def prowler():
     stats = extract_findings_statistics(findings)
 
     if args.slack:
-        if "SLACK_API_TOKEN" in environ and "SLACK_CHANNEL_ID" in environ:
+        if "SLACK_API_TOKEN" in environ and (
+            "SLACK_CHANNEL_NAME" in environ or "SLACK_CHANNEL_ID" in environ
+        ):
             _ = send_slack_message(
```
jfagoagas, line 254: Are both values valid in the Slack API?

MrCloudSec: What API? Those values are parsed by Prowler and then used them as args to the `send_slack_message` function.

jfagoagas: I mean that you are now passing either `SLACK_CHANNEL_NAME` or `SLACK_CHANNEL_ID` to the `send_slack_message` function, and then is used in the following way:
```python
response = client.chat_postMessage(
            username="Prowler",
            icon_url=square_logo_img,
            channel=f"#{channel}",
            blocks=create_message_blocks(identity, logo, stats),
        )
```
Are both valid when used in the Slack API as `channel`?

jfagoagas (approved):
🔝

PR #4081, opened 2024-05-23
chore(output): create generic output library

MrCloudSec:
### Description
We need to make the Prowler outputs a component in v5, for that reason we need a base Output class as the parent of each output class like SecurityHub, Slack or S3.

- Create a base Output class
- Define child classes
- Each output needs to handle authentication by their own

### License

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,7 +30,7 @@ tags
 *.DS_Store
 
 # Prowler output
-output/
+/output/
 
 # Prowler found secrets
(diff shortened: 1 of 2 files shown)


jfagoagas:
I'm closing this PR since we are going to tackle this in different parts, see PRs:
- https://github.com/prowler-cloud/prowler/pull/4315
- https://github.com/prowler-cloud/prowler/pull/4133

PR #4085, opened 2024-05-24
fix(rds): solve ParameterValue KeyError

MrCloudSec:
### Description

Solve `ParameterValue` KeyError and use the values defined in https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/mysql-ssl-connections.html#mysql-ssl-connections.require-ssl.

### License

### DIFF
--- a/prowler/providers/aws/services/rds/rds_instance_transport_encrypted/rds_instance_transport_encrypted.py
+++ b/prowler/providers/aws/services/rds/rds_instance_transport_encrypted/rds_instance_transport_encrypted.py
@@ -42,7 +42,7 @@ class rds_instance_transport_encrypted(Check):
                     for parameter in db_instance.parameters:
                         if (
                             parameter["ParameterName"] == "rds.force_ssl"
-                            and parameter["ParameterValue"] == "1"
+                            and parameter.get("ParameterValue", "0") == "1"
                         ):
                             report.status = "PASS"
                             report.status_extended = f"RDS Instance {db_instance.id} connections use SSL encryption."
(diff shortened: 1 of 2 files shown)


prowler/providers/aws/services/rds/rds_instance_transport_encrypted/rds_instance_transport_encrypted.py
```diff
@@ -42,15 +42,15 @@ def execute(self):
                     for parameter in db_instance.parameters:
                         if (
                             parameter["ParameterName"] == "rds.force_ssl"
-                            and parameter["ParameterValue"] == "1"
+                            and parameter.get("ParameterValue", "0") == "1"
```
jfagoagas, line 45: Why this is not done at the service level?

tests/providers/aws/services/rds/rds_instance_transport_encrypted/rds_instance_transport_encrypted_test.py
```diff
@@ -250,7 +250,7 @@ def test_mysql_rds_instance_no_ssl(self):
             Parameters=[
                 {
                     "ParameterName": "require_secure_transport",
-                    "ParameterValue": "0",
+                    "ParameterValue": "OFF",
```
jfagoagas, line 253: There are some tests failing, I don't know if they are related with this but please take a look at it. Thanks!

prowler/providers/aws/services/rds/rds_instance_transport_encrypted/rds_instance_transport_encrypted.py
MrCloudSec: Because in the service we are storing all the parameters in an attribute. It is not like the DB Cluster.

tests/providers/aws/services/rds/rds_instance_transport_encrypted/rds_instance_transport_encrypted_test.py
MrCloudSec: I don't know why they failed, but know they work correctly.

prowler/providers/aws/services/rds/rds_instance_transport_encrypted/rds_instance_transport_encrypted.py
jfagoagas: okay, it'd be better to do this on the service side but it's good for now.

jfagoagas (approved):
🔝

PR #4086, opened 2024-05-24
fix(opensearch): handle non existing SAMLOptions in domain

MrCloudSec:
### Description

Handle non existing `SAMLOptions` in OpenSearch Domain.

### License


pedrooot (approved):
LGTM!

jfagoagas (changes requested):
Could you please add `get()` for all the calls in the function? Thanks!

MrCloudSec:
> Could you please add `get()` for all the calls in the function? Thanks!

At which calls do you refer? Remember that we have to do this one by one because since we have to make sure to handle each specific case correctly.

jfagoagas:
> > Could you please add `get()` for all the calls in the function? Thanks!
> 
> At which calls do you refer? Remember that we have to do this one by one because since we have to make sure to handle each specific case correctly.

I meant for all the calls within that function but it's good for now.

jfagoagas (approved):
🔝

PR #4094, opened 2024-05-27
chore(AWS): allow ingress to any port for user defined network interface types

kagahd:
### Context

This PR is to implement the suggestion brought up in the discussion #4065 

### Description

Copied from the discussion #4065 : 

> prowler has checks to ensure that there are no AWS Security Groups without ingress filtering (0.0.0.0/0) being used.
In prowler v2 it's `extra74` and in prowler v3/v4 it's `ec2_securitygroup_allow_ingress_from_internet_to_any_port`.
However, there are cases where an ec2 instance requires access from or to external IP addresses. In these cases, to reduce the attack surface, it's best practice to put a load balancer or an API gateway in front of the ec2 instance whereas the ec2 instance must only allow traffic from/to its load balancer. Furthermore, the load balancer or API gateway should allow traffic only on the IP range, the protocols and the ports that are required for the application running on the ec2 instance.
Additionally, separation of concerns should be applied: each resource or resource group should have its own security group. Especially if the security group allows public access, it must only be attached to the load balancer or an API gateway.
Additionally, security groups may get automatically created by AWS for VPC endpoints as for example `GuardDuty`. These, by AWS automatically created, security groups allow all in- and outbound traffic (0.0.0.0/0) but since their endpoint is on AWS side, we consider it's not our responsibility to protect AWS resources.
Therefore, we didn't want to get alarmed by the prowler checks in these special cases.
To achieve this goal, we patched prowler v2 check extra74 as follows, the changed code line is:
`SG_NO_INGRESS_FILTER=$($AWSCLI ec2 describe-network-interfaces $PROFILE_OPT --region $regx --filters "Name=group-id,Values=$SG_ID" --query "length(NetworkInterfaces[?Attachment.InstanceOwnerId != 'amazon-elb' && InterfaceType != 'api_gateway_managed' && InterfaceType != 'vpc_endpoint'])" --output text)`
instead of:
`SG_NO_INGRESS_FILTER=$($AWSCLI ec2 describe-network-interfaces $PROFILE_OPT --region $regx --filters "Name=group-id,Values=$SG_ID" --query "length(NetworkInterfaces)" --output text)`

The implementation of the check `ec2_securitygroup_allow_ingress_from_internet_to_any_port` in this PR behaves exactly as our patched prowler v2 check `extra74` as I described above.
As you can see in the code, apart from the config, everything was already there. I just needed to adapt the check to read the config and behave accordingly.
However, two unit tests in `Test_ec2_securitygroup_allow_ingress_from_internet_to_any_port` are still [failing](https://github.com/prowler-cloud/prowler/actions/runs/9258609871/job/25468999822?pr=4094#step:15:359) (their status is `PASS` instead of `FAIL`) because their open-to-the-internet security group does not have a network interface attached, so the check let them correctly `PASS`. Once a network interface is attached that is not allowed, these both unit tests should correctly `FAIL`.
Could you please show me how the unit tests have to be adapted so that their security group is attached to a network interface?

### License


prowler/config/config.yaml
```diff
@@ -253,6 +253,16 @@ aws:
   # aws.rds_instance_backup_enabled
   # Whether to check RDS instance replicas or not
   check_rds_instance_replicas: False
+  #
+  check_ec2_securitygroup_allow_ingress_from_internet_to_any_port_allowed_interface_types:
```
jfagoagas:
```suggestion
  ec2_allowed_interface_types:
```

prowler/providers/aws/services/ec2/ec2_securitygroup_allow_ingress_from_internet_to_any_port/ec2_securitygroup_allow_ingress_from_internet_to_any_port.py
jfagoagas, line 1: Could you please add a test for this new behaviour?

jfagoagas (commented):
@sergargar could you please also review this PR? It's an interesting improvement to reduce noise and false positives.

prowler/providers/aws/services/ec2/ec2_securitygroup_allow_ingress_from_internet_to_any_port/ec2_securitygroup_allow_ingress_from_internet_to_any_port.py
kagahd: Hi @jfagoagas, I would like to do that but I'm not sure how to attach a mocked network interface to a mocked security group. Could you help me with that?

jfagoagas: For sure @kagahd but the tests for that check uses https://github.com/getmoto/moto to create VPCs, Subnets, Security Groups, etc, so you will need to use the same logic as with `boto3` to create whatever you need. Also you can modify the returned objects within the `EC2` client after the `mocks` to inject whatever values you need.

MrCloudSec (changes requested):
I like this idea @kagahd !
Please can you add a test for the case of: `f"Security group {security_group.name} ({security_group.id}) has at least one port open to the Internet but is exclusively attached to none or to an allowed network interface type."`. Thanks!

prowler/providers/aws/services/ec2/ec2_securitygroup_allow_ingress_from_internet_to_any_port/ec2_securitygroup_allow_ingress_from_internet_to_any_port.py
kagahd: It should be fine now. Both previously failed unit tests pass now.
I also added a new one `test_ec2_open_sg_attached_to_allowed_eni` to test whether an **open** security PASSes when attached to an **allowed** network interface. This unit test passes as well.

kagahd:
> Please can you add a test for the case of: `f"Security group {security_group.name} ({security_group.id}) has at least one port open to the Internet but is exclusively attached to none or to an allowed network interface type."`.

I did so just a few minutes before. We were obviously cross posting. The unit test is called `test_ec2_open_sg_attached_to_allowed_eni`. Is this what you asked for, @sergargar?

MrCloudSec:
> > Please can you add a test for the case of: `f"Security group {security_group.name} ({security_group.id}) has at least one port open to the Internet but is exclusively attached to none or to an allowed network interface type."`.
> 
> I did so just a few minutes before. We were obviously cross posting. The unit test is called `test_ec2_open_sg_attached_to_allowed_eni`. Is this what you asked for, @sergargar?

Yes, thank you!

MrCloudSec:
@kagahd can you add also the new config values to the documentation in `prowler/config/config.yaml` and `docs/tutorials/configuration_file.md`? Thanks.

kagahd:
> @kagahd can you add also the new config values to the documentation in `prowler/config/config.yaml` and `docs/tutorials/configuration_file.md`? Thanks.

Sure, I just committed and pushed them.

MrCloudSec:
Thanks @kagahd, I also forgot to say to put that config variables in the `# AWS EC2 Configuration` section of the yaml and use the same pattern, please.

kagahd:
The workflow `Pull Request Documentation Link` [failed](https://github.com/prowler-cloud/prowler/actions/runs/9293428914/job/25576534567?pr=4094#step:2:18) because of `Resource not accessible by integration`. This corresponds to the expected functionality as documented in the repository of the workflow used:

> In public repositories this action does not work in pull_request workflows when triggered by forks. Any attempt will be met with the error, Resource not accessible by integration. 

Source: https://github.com/peter-evans/create-or-update-comment?tab=readme-ov-file#action-inputs

So for me it should be fine.

kagahd:
> Thanks @kagahd, I also forgot to say to put that config variables in the `# AWS EC2 Configuration` section of the yaml and use the same pattern, please.

I think I've already done that, or did you mean something else, @sergargar?

prowler/providers/aws/services/ec2/ec2_securitygroup_allow_ingress_from_internet_to_any_port/ec2_securitygroup_allow_ingress_from_internet_to_any_port.py
```diff
@@ -35,8 +35,26 @@ def execute(self):
...
+                                    and isinstance(eni.attachment, dict)
+                                    and "InstanceOwnerId" in eni.attachment
+                                    and eni.attachment["InstanceOwnerId"]
+                                    not in ec2_client.audit_config.get(
+                                        "ec2_allowed_instance_owners",
+                                        [],
+                                    )
+                                ):
+                                    report.status = "FAIL"
+                                    report.status_extended = f"Security group {security_group.name} ({security_group.id}) has at least one port open to the Internet and is not exclusively attached to an allowed network interface type."
```
jfagoagas: What about the `InstanceOwnerId`? Maybe we should put something into the `report.status_extended?

What do you think? @kagahd @sergargar

kagahd: I think that the logic is still wrong even though the unit tests pass. Let me explain later but don't merge yet.

jfagoagas: For sure, I'll add the `no-merge` label.

MrCloudSec:
> > Thanks @kagahd, I also forgot to say to put that config variables in the `# AWS EC2 Configuration` section of the yaml and use the same pattern, please.
> 
> I think I've already done that, or did you mean something else, @sergargar?

I meant to put it in the AWS EC2 Configuration section where the shodan config variable is, and put something like:
 `# aws.ec2_securitygroup_allow_ingress_from_internet_to_any_port`

prowler/providers/aws/services/ec2/ec2_securitygroup_allow_ingress_from_internet_to_any_port/ec2_securitygroup_allow_ingress_from_internet_to_any_port.py
kagahd: I cleaned up the code to make the logic easier to understand even though it stays the same boolean logic: if `eni_type` is allowed **OR** `eni_owner` is allowed (or both), then the check PASSes even though the Security Group is open to the Internet for a given port.
I added also still more unit tests to cover all cases.
I think it's now ready to be merged.

kagahd:
> I meant to put it in the AWS EC2 Configuration section where the shodan config variable is, and put something like:
> `# aws.ec2_securitygroup_allow_ingress_from_internet_to_any_port`

Sorry, I don't understand what you mean. Excuse my ignorance, I don't know what shodan is, unless karate is meant. 😄

MrCloudSec:
> > I meant to put it in the AWS EC2 Configuration section where the shodan config variable is, and put something like:
> > `# aws.ec2_securitygroup_allow_ingress_from_internet_to_any_port`
> 
> Sorry, I don't understand what you mean. Excuse my ignorance, I don't know what shodan is, unless karate is meant. 😄

haha I meant to put that config variables up in the yaml, just where the AWS EC2 Configuration is.

kagahd:
> I meant to put it in the AWS EC2 Configuration section where the shodan config variable is, and put something like:
> `# aws.ec2_securitygroup_allow_ingress_from_internet_to_any_port`

Now I got it (hopefully) 😄 and added that line 🎉

tests/providers/aws/services/ec2/ec2_securitygroup_allow_ingress_from_internet_to_any_port/ec2_securitygroup_allow_ingress_from_internet_to_any_port_test.py
```diff
@@ -240,6 +464,56 @@ def test_ec2_compliant_default_sg_only_open_to_one_port(self):
                     assert sg.resource_details == default_sg_name
                     assert sg.resource_tags == []
 
+    @mock_aws
+    def test_ec2_compliant_default_sg_only_open_to_one_port(self):
```
jfagoagas: Is this test name right?

jfagoagas (dismissed):
Thanks again @kagahd for taking care of this. Is a great improvement for the Security Group check in order to keep reducing false positives/noise/things that cannot be fixed in Prowler.

💯

tests/providers/aws/services/ec2/ec2_securitygroup_allow_ingress_from_internet_to_any_port/ec2_securitygroup_allow_ingress_from_internet_to_any_port_test.py
kagahd: > Is this test name right?

Of course not, eagle-eyed! 😄 It comes from the **former** check `ec2_securitygroup_allow_ingress_from_internet_to_any_port` which was renamed to `ec2_securitygroup_allow_ingress_from_internet_to_all_ports` which is compliant if not all ports but only 1 port is open (see [code here](https://github.com/prowler-cloud/prowler/pull/3962/files#diff-a7e83fa8c24ed98f345add4b8110752aae3fcec9359046d72d86b764100868bdR186)).
I corrected it and even caught another one of its kind (`test_ec2_default_sgs_with_all_ports_check`versus `test_ec2_default_sgs_with_any_ports_check`). 🎉

prowler/providers/aws/services/ec2/ec2_securitygroup_allow_ingress_from_internet_to_any_port/ec2_securitygroup_allow_ingress_from_internet_to_any_port.py
kagahd: > For sure, I'll add the `no-merge` label.

@jfagoagas, the `no-merge` label can be removed from my point of view. Thanks!

MrCloudSec:
Hi @kagahd, I have moved the config variables to the EC2 section as I mentioned.

MrCloudSec (approved):
Thanks for your work @kagahd !!

PR #4116, opened 2024-05-28
 fix(rds): ParameterValue MySQL and MariaDB RDS Instances

sansns:
### Context

MySQL and MariaDB Instances use a ```require_secure_transport = 1``` while the clusters use ```require_secure_transport = ON``` to set the ParameterValue for requiring encryption.

### Description

The current checks have been updated for DB Instances to check for  ```require_secure_transport = ON``` which will not provide the correct information. 

RDS MySQL Cluster Encryption Required

RDS MySQL DB Instance Encryption Required

### License


MrCloudSec (approved):
Great catch, thank you @sansns !!

PR #4136, opened 2024-05-30
fix(mutelist): Handle items starting by *

jfagoagas:
### Context

We are aborting the execution if the allowlist regex fails unexpectedly and also not handling right when the pattern starts by `*` which is not a valid regex.

### Description

- Change `critical` log to `error`
- Return `False` if something not handled happens.
- If the pattern starts by `*` add a `.` in front of it.

### License


MrCloudSec (approved):
LGTM

PR #4140, opened 2024-05-30
fix(allowlist): return False if something fails

jfagoagas:
### Description

Remove the `logger.critical` and `sys.exit()` from the Allowlist logic.

### License


MrCloudSec (approved):
LGTM

PR #4147, opened 2024-05-31
fix(defender): Add new parameter required by new API version

puchy22:
### Context

New API version break the call to get defender pricings, comes from https://github.com/prowler-cloud/prowler/pull/4034

### Description

Add new required parameter to call.

### License


jfagoagas (approved):
🥷

PR #4148, opened 2024-05-31
fix(rds): Handle DBParameterGroupNotFound

jfagoagas:
### Description

Handle `DBParameterGroupNotFound` when calling `describe_db_parameters` RDS API.

### License

### DIFF
--- a/prowler/providers/aws/services/rds/rds_service.py
+++ b/prowler/providers/aws/services/rds/rds_service.py
@@ -198,57 +198,97 @@ class RDS(AWSService):
                 "describe_db_clusters"
             )
             for page in describe_db_clusters_paginator.paginate():
-                for cluster in page["DBClusters"]:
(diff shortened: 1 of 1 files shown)


prowler/providers/aws/services/rds/rds_service.py
```diff
@@ -198,57 +198,97 @@ def __describe_db_clusters__(self, regional_client):
...
+                                    )
+                                    # We must use a unique value as the dict key to have unique keys
+                                    self.db_clusters[db_cluster_arn] = db_cluster
+
+                                    # Get DB Cluster Parameters
+                                    describe_db_parameters_paginator = (
+                                        regional_client.get_paginator(
+                                            "describe_db_parameters"
+                                        )
+                                    )
```
MrCloudSec, line 246: I would put another try before this for loop (which is the one that is failing), so even if the cluster does not have parameter group name, its class can be created.

MrCloudSec (approved):
LGTM

MrCloudSec:
Removing backport-v3 tag in favor of the new PR of this issue https://github.com/prowler-cloud/prowler/pull/4150

PR #4149, opened 2024-05-31
chore(iam): Downgrade AWS IAM check severity

### LINKED ISSUE #4142: [Bug]: Inconsistent severity assigned to iam_no_custom_policy_permissive_role_assumption
### Steps to Reproduce

This issue concerns the severity classification for check ID `iam_no_custom_policy_permissive_role_assumption`.

The check has `critical` severity. It flags policies that allow unconstrained `sts:AssumeRole`.

**Why is this check classified as critical, especially when compared to other related checks, which are classified only as high? How can an attacker abuse unconstrained `sts:AssumeRole` to escalate privileges in a generic way?**
I understand that `sts:AssumeRole` allows assuming any role _in theory_. However, the role's trust policy is also consulted. Roles can only be assumed if their trust policy explicitly allowlists the caller. In addition, the role's name might not be public and is not always known.

Other AWS IAM checks that are similar are
- `iam_policy_allows_privilege_escalation`: classified as `high`, generically covering multiple options to escalate privileges
- `iam_inline_policy_no_administrative_privileges`: classified as `high`, checks for administrative permissions assigned inline.

### Expected behavior

I expect severity scores to be comparable across similarly-severe issue. However, I don't understand why the severities for the checks mentioned above differ. Part of this might be that **the check's metadata does not sufficiently explain the risk, which is stated as: `If not restricted unintended access could happen.`.**

See above.

### Environment Resource

Not relevant / Workstation

### OS used

Not relevant / macOS

### Prowler version

4.1.0

### Pip version

24.0

### Context

**Note:** I've been writing up relatively minor points as issues on GitHub. I do this because I think it's useful for others to be able to revisit and find issues, and I believe GitHub's issue search to be more accessible than Slack. **Let me know if you'd prefer I raise these points via chat.**

rieck-srlabs:
### Context

Corrects inconsistent check severity, fixes #4142.

### Description

Changes severity of `iam_no_custom_policy_permissive_role_assumption` from `Critical` to `High` to align with other existing permission-related checks.

### License

### DIFF
--- a/prowler/providers/aws/services/iam/iam_no_custom_policy_permissive_role_assumption/iam_no_custom_policy_permissive_role_assumption.metadata.json
+++ b/prowler/providers/aws/services/iam/iam_no_custom_policy_permissive_role_assumption/iam_no_custom_policy_permissive_role_assumption.metadata.json
@@ -12,7 +12,7 @@
   "ResourceType": "AwsIamPolicy",
   "Description": "Ensure that no custom IAM policies exist which allow permissive role assumption (e.g. sts:AssumeRole on *)",
   "Risk": "If not restricted unintended access could happen.",
-  "RelatedUrl": "",
+  "RelatedUrl": "https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_use_permissions-to-switch.html#roles-usingrole-createpolicy",
   "Remediation": {
     "Code": {
       "CLI": "",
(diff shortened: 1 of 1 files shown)


jfagoagas (changes requested):
Thanks for the fix @rieck-srlabs !!

Could you add this link https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_use_permissions-to-switch.html#roles-usingrole-createpolicy  to the `RelatedUrl`?

rieck-srlabs:
Done, thanks for the quick and helpful handling as always.

jfagoagas (approved):
💯

PR #4150, opened 2024-05-31
fix(rds): use correct API call for cluster parameters

MrCloudSec:
### Context

We were using an invalid API Call to describe the parameters of a DB Cluster.

### Description

Use `describe_db_cluster_parameters` call and separate it into a new function.

### License

### DIFF
--- a/prowler/providers/aws/services/rds/rds_service.py
+++ b/prowler/providers/aws/services/rds/rds_service.py
@@ -26,6 +26,7 @@ class RDS(AWSService):
         self.__threading_call__(self.__describe_db_snapshots__)
         self.__threading_call__(self.__describe_db_snapshot_attributes__)
         self.__threading_call__(self.__describe_db_clusters__)
+        self.__threading_call__(self.__describe_db_cluster_parameters__)
         self.__threading_call__(self.__describe_db_cluster_snapshots__)
         self.__threading_call__(self.__describe_db_cluster_snapshot_attributes__)
         self.__threading_call__(self.__describe_db_engine_versions__)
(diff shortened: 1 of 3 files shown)


toniblyx (dismissed):

prowler/providers/aws/services/rds/rds_service.py
```diff
@@ -436,8 +434,8 @@ class DBCluster(BaseModel):
     auto_minor_version_upgrade: bool
     multi_az: bool
     parameter_group: str
-    force_ssl: Optional[bool]
-    require_secure_transport: Optional[str]
+    force_ssl: Optional[str] = "0"
+    require_secure_transport: Optional[str] = "OFF"
```
jfagoagas: Maybe the optional is not needed here since you have default values, is not going to be `None`.

prowler/providers/aws/services/rds/rds_service.py
```diff
@@ -298,6 +251,51 @@ def __describe_db_clusters__(self, regional_client):
...
+                    try:
+                        describe_db_cluster_parameters_paginator = (
+                            regional_client.get_paginator(
+                                "describe_db_cluster_parameters"
+                            )
+                        )
+                        for page in describe_db_cluster_parameters_paginator.paginate(
+                            DBClusterParameterGroupName=cluster.parameter_group
+                        ):
+                            for parameter in page["Parameters"]:
```
jfagoagas, line 268: Do we need to include two `try/except` here? One for each loop?

jfagoagas (changes requested):
Great catch!!

prowler/providers/aws/services/rds/rds_service.py
MrCloudSec: I don't think it is necessary, what we usually do is one try for the whole function and another one for each resource

prowler/providers/aws/services/rds/rds_service.py
MrCloudSec: True!

prowler/providers/aws/services/rds/rds_service.py
jfagoagas: yes but if something breaks within the loop we are not going to loop over and we should.

jfagoagas (approved):

PR #4153, opened 2024-06-03
fix(cloudtrail): trail.region must be home region

jfagoagas:
### Context

We are storing the CloudTrail Trail region as the `regional_client.region` instead of `HomeRegion` for multi region trails. This is due to the fact that a multi region trail can be checked in whatever region where it is working but we must store the `HomeRegion` as the Trail region since it’s where it was created.

## Description

`trail.region` must have the region where the trail was created.

### License


pedrooot (dismissed):
Good approach! Thanks @jfagoagas

jfagoagas:
Another option should be to leave the service as is and update the checks to use `finding.region = trail.home_region`.

MrCloudSec:
> Another option should be to leave the service as is and update the checks to use `finding.region = trail.home_region`.

I would do this approach, setting the `home_region` as region in all the checks except from the two `cloudtrail_multi_region_enabled*` checks.

MrCloudSec (approved):
Good catch!

PR #4186, opened 2024-06-04
docs(reporting): fix mapping of json-ocsf field cloud.account.type

kagahd:
### Context

The documentation of the mapping between the native JSON Prowler v3 and JSON-OCSF in Prowler v4 was wrong.

### Description

The Prowler v3 field `Provider` was mapped in documentation to Prowler v4 field `cloud.account.type`.
I think that's wrong because if the v3 field `Provider` contained for example the value `aws`, the v4 field `cloud.account.type` would have be `AWS_Account`. So I changed the v4 mapping from `cloud.account.type` to `cloud.provider` which contains also `aws` like the v3 field `Provider` in this case.

### License


MrCloudSec (approved):
Good catch @kagahd, thank you!

PR #4193, opened 2024-06-05
fix(glue): check if get dev endpoints call is supported

MrCloudSec:
### Description

Set a WARNING log if the GetDevEndpoints call is not supported in the region.

### License


pedrooot (approved):
💯

PR #4202, opened 2024-06-07
fix(custom): execute custom checks

sejimhp:
### Context

Custom rules were not available and I would like to to fix it.

It is caused by copying the custom checks folder after loading the checks.
so, added when files are copied

### Description

copying the custom check in the `parse_checks_from_folder` function [here](https://github.com/prowler-cloud/prowler/blob/master/prowler/__main__.py#L181-L183)
But check loading is called in [load_checks_to_execute](https://github.com/prowler-cloud/prowler/blob/master/prowler/__main__.py#L150-L161) before that

There was no process to assign a custom check for that transition, so the custom check was not executed!

Added support for updating pre-loaded sets at the time of copying custom check folders.

### Check

Here is an example of gcp.
Prepare the following directory structure rules.

```
custom_rule
└── cloudsql_test_custom_rule
    ├── __init__.py
    ├── cloudsql_test_custom_rule.metadata.json
    └── cloudsql_test_custom_rule.py
```

Execute the following command

without custome rule scan
```
prowler gcp --project-ids {project_id}
```

with custome rule scan
```
prowler gcp --project-ids {project_id} -x custom_rule
```

Confirmed that the number of scans performed and the number of compliant evaluations changes

### License

### DIFF
--- a/prowler/__main__.py
+++ b/prowler/__main__.py
@@ -180,7 +180,8 @@ def prowler():
 
     # Import custom checks from folder
     if checks_folder:
-        parse_checks_from_folder(global_provider, checks_folder)
+        custom_checks = parse_checks_from_folder(global_provider, checks_folder)
+        checks_to_execute.update(custom_checks)
 
     # Exclude checks if -e/--excluded-checks
     if excluded_checks:
--- a/prowler/lib/check/check.py
+++ b/prowler/lib/check/check.py
@@ -126,9 +126,9 @@ def parse_checks_from_file(input_file: str, provider: str) -> set:
 
 
 # Load checks from custom folder
-def parse_checks_from_folder(provider, input_folder: str) -> int:
+def parse_checks_from_folder(provider, input_folder: str) -> set:
     try:
-        imported_checks = 0
+        custom_checks = set()
         # Check if input folder is a S3 URI
         if provider.type == "aws" and re.search(
             "^s3://([^/]+)/(.*?([^/]+))/$", input_folder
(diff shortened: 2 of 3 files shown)


pedrooot:
Hey! @sejimhp This looks great! Could you fix the tests in order to ensure that all it's correct? Thank you 💯

sejimhp:
@pedrooot Thank you for review!
I fixed the test.

sejimhp:
I had to add an additional push because it failed in the format.

pedrooot:
If you have a look in the tests, now blake it's failing. Could you try to solve it? Using pre-commit should be enough

sejimhp:
The coverage drop seems to be caused by __main__.py, but I couldn't find a test for it...

MrCloudSec (approved):
Good catch @sejimhp, thank you so much for the fix!

prowler/__main__.py
```diff
@@ -180,7 +180,8 @@ def prowler():
 
     # Import custom checks from folder
     if checks_folder:
-        parse_checks_from_folder(global_provider, checks_folder)
+        custom_checks = parse_checks_from_folder(global_provider, checks_folder)
+        checks_to_execute.update(custom_checks)
```
kagahd, line 184: No, this is wrong and a regression! Including checks from the `checks_folder` does not necessarily mean that we want to execute **all** of them! You may want to execute only checks that are defined by parameter `-c` or `--checks`. See [prowlers documentation](https://docs.prowler.com/projects/prowler-open-source/en/latest/tutorials/misc/#custom-checks). Your code prevents this behavior. It worked still in the previous prowler version 4.2.3 but it's broken now in the actual v4.2.4.

jfagoagas: This workaround fixed that https://github.com/prowler-cloud/prowler/pull/4256

kagahd: > This workaround fixed that #4256

That's great! 
PS: I clicked on "submit review" only the next day of my comment. That's why my comment only appeared now, after having fixed the issue already.

jfagoagas: No problem, thanks!

PR #4206, opened 2024-06-07
fix(html): fix status from HTML outputs

pedrooot:
### Description

Finding status were showing with a bad format on HTML outputs
Screenshot 2024-06-07 at 12 09 11

### License


MrCloudSec (approved):
Good catch!

PR #4242, opened 2024-06-11
chore(network): Reduce network watchers azure check findings

puchy22:
### Context

The check to ensure if network watchers was enabled in every region introduce some noise to the final report.

### Description

Check logic and tests have been changed to make only one report per subscription.

### License

### DIFF
--- a/prowler/providers/azure/services/network/network_watcher_enabled/network_watcher_enabled.py
+++ b/prowler/providers/azure/services/network/network_watcher_enabled/network_watcher_enabled.py
@@ -18,10 +18,10 @@ class network_watcher_enabled(Check):
 
             if missing_locations:
                 report.status = "FAIL"
-                report.status_extended = f"Network Watcher is not enabled for the following locations in subscription '{subscription}': {', '.join(missing_locations)}"
+                report.status_extended = f"Network Watcher is not enabled for the following locations in subscription '{subscription}': {', '.join(missing_locations)}."
             else:
                 report.status = "PASS"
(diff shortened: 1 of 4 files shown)


prowler/providers/azure/services/network/network_watcher_enabled/network_watcher_enabled.py
```diff
@@ -5,24 +5,25 @@
...
-                    report.status = "PASS"
-                    report.status_extended = f"Network Watcher is enabled for the location {location} in subscription {subscription}."
-                    findings.append(report)
+            report = Check_Report_Azure(self.metadata())
+            report.subscription = subscription
+            report.resource_name = "Network Watcher"
+            report.location = "Global"
+            report.resource_id = f"/subscriptions/{network_client.subscriptions[subscription]}/resourceGroups/NetworkWatcherRG/providers/Microsoft.Network/networkWatchers/NetworkWatcher_*"
+            report.status = "FAIL"
+            report.status_extended = f"Network Watcher is not enabled for all locations in subscription '{subscription}'."
```
MrCloudSec: Would it be possible to indicate which are the locations that do not have network watchers?

puchy22: Good point, thanks to that I start thinking in how to do and I believe that I improve so much the business logic of the check in terms of efficiency and readability. Review it again and tell me if you see better than before.

MrCloudSec (approved):
Love it! Thanks @puchy22

PR #4285, opened 2024-06-21
fix(rds): handle not existing endpoint

MrCloudSec:
### Description

Handle not existing endpoint by setting it as an empty value and avoid the following error:
`rds_instance_no_public_access -- AttributeError[26]: 'NoneType' object has no attribute 'get'`

### License


jfagoagas (approved):
Great catch!

PR #4291, opened 2024-06-24
feat(gcp): add service account impersonation

### LINKED ISSUE #4006: Add Service Account impersonation in GCP as a new authentication method
### New feature motivation

Add [service account impersonation in GCP](https://cloud.google.com/iam/docs/service-account-impersonation) as a new authentication method in Prowler.

### Solution Proposed

Adding the flag `--impersonate-service-account <target service account email>`

Instead of saving credentials file or run prowler inside GCP workload (for the metadata server option), add the mentioned authentication method.
This is actually security best practice so it should be high priority because now the `credentials.json` approach have private keys on the machines which is not a good idea.

### Additional context

Thanks to Ben Lahav via Slack for the idea!

MrCloudSec:
### Description

Add the ability of impersonate a GCP Service Account using the argument `--impersonate-service-account <service-account-email>`.

### License

### DIFF
--- a/prowler/providers/gcp/gcp_provider.py
+++ b/prowler/providers/gcp/gcp_provider.py
@@ -3,7 +3,8 @@ import re
 import sys
 
 from colorama import Fore, Style
-from google import auth
+from google.auth import default, impersonated_credentials
+from google.auth.transport.requests import Request
 from google.oauth2.credentials import Credentials
 from googleapiclient import discovery
 from googleapiclient.errors import HttpError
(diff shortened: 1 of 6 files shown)


prowler/providers/gcp/gcp_provider.py
```diff
@@ -197,6 +230,10 @@ def print_credentials(self):
             f"GCP Account: {Fore.YELLOW}{self.identity.profile}{Style.RESET_ALL}",
             f"GCP Project IDs: {Fore.YELLOW}{', '.join(self.project_ids)}{Style.RESET_ALL}",
         ]
+        if self.impersonated_service_account:
```
jfagoagas, line 233: We need to add tests for this part.

jfagoagas (changes requested):
That's a great addition! Please think about a way to create tests for this new code, thanks!

jfagoagas (approved):
💯

PR #4359, opened 2024-07-02
chore(iam): Remove unnecesary attached policy in a inline policy

puchy22:
### Context

Ensure if a inline policy is attached when they are always attached.

### Description

Removed the unnecessary verification from `iam_inline_policy_no_administrative_privileges` check.

### License


jfagoagas (approved):
🥷

PR #4368, opened 2024-07-03
refactor(ASFF): create class

jfagoagas:
### Context

We are creating modular outputs for Prowler, now is the time for the AWS Security Finding Format.

### Description

- Create the `ASFF(Output)` class.
- Add documentation for all the functions.
- Refactor the `filter_security_hub_findings_per_region` since now we don't need to recreate the ASFF findings, we've been doing it twice 🤦 
- Adapt the tests.

### License


prowler/lib/outputs/csv/models.py
```diff
@@ -39,6 +39,7 @@ def batch_write_data_to_file(self) -> None:
                 csv_writer.writeheader()
                 for finding in self._data:
                     csv_writer.writerow(finding)
+                self._file_descriptor.close()
```
jfagoagas, line 42: Added to the CSV class since it was missing.

prowler/lib/outputs/csv/models.py
```diff
@@ -30,7 +30,7 @@ def transform(self, findings: list[Finding]) -> None:
     def batch_write_data_to_file(self) -> None:
         """Writes the findings to a file using the CSV format using the `Output._file_descriptor`."""
         try:
-            if self._file_descriptor:
+            if self._file_descriptor and not self._file_descriptor.closed:
```
jfagoagas, line 33: Good to check that the file descriptor is not closed.

prowler/lib/outputs/json_asff/json_asff.py
jfagoagas, line 1: Move to the the `asff.py`

prowler/lib/outputs/json_asff/models.py
jfagoagas, line 1: Move to the the `asff.py`

prowler/lib/outputs/file_descriptors.py
```diff
@@ -77,6 +76,8 @@ def fill_file_descriptors(output_modes, output_directory, output_filename, provi
                 # FIXME: Remove this once we always use the new CSV(Output)
                 if output_mode == "csv":
                     continue
+                elif output_mode == "json-asff":
+                    continue
```
MrCloudSec, line 73: Is this needed? As compliance does not use it, I think we can remove it

MrCloudSec:
Tested locally and it perfectly works!

prowler/lib/outputs/file_descriptors.py
jfagoagas: I think is needed for now until we solve Compliance, am I right @pedrooot ?

pedrooot: when all the output PRs are merged we can delete all the ifs and maintain only the specifics for compliance.

pedrooot:
Now, `batch_send_to_security_hub` use `_send_findings_to_security_hub` passing the findings as a dict and the finding format in `AWSSecurityFindingFormat`. The real problem it's that `resolve_security_hub_previous_findings` pass the findings in dict function. I've made a "parse" in `_send_findings_to_security_hub` and depending the data structure it makes a different behaviour. This is not correct so I'll work on standardize all

pedrooot:
The final approach it's that all the findings passed to `_send_findings_to_security_hub` are dict

prowler/lib/outputs/file_descriptors.py
MrCloudSec: I still think that this can be removed, this `output_mode` is not used in compliance...

pedrooot (approved):
giphy

PR #4424, opened 2024-07-10
fix(cis): add missing fields and reorder

MrCloudSec:
### Description

Add missing fields in CIS and reorder them to respect the current order of the fields in the output file.

### License


prowler/lib/check/compliance_models.py
```diff
@@ -91,7 +91,6 @@ class CIS_Requirement_Attribute(BaseModel):
     AdditionalInformation: str
     DefaultValue: Optional[str]
     References: str
-    DefaultValue: Optional[str]
```
jfagoagas, line 94: Why are you removing this? Out of curiosity.

jfagoagas (approved):
Great catch @sergargar 👏

prowler/lib/check/compliance_models.py
MrCloudSec: It is duplicated, you can see the same value in line 92

PR #4446, opened 2024-07-15
fix(iam_avoid_root_usage): change timestamp format

pedrooot:
### Description

Change timestamp format to use the same format as the user data.

### License


jfagoagas (changes requested):
I'm curious about the origin of this change: Why do we need to change the timestamp format?

Could you add some tests to prevent this from happening again?

jfagoagas:
AWS IAM Credential Report now returns the timestamps in this format:
- After -> `2024-04-22T19:47:52Z`
- Before -> `2024-04-22T19:47:52+00:00`

`Z` and `+00:00` are equivalent but our parsing is not handling it.

prowler/providers/aws/services/iam/iam_avoid_root_usage/iam_avoid_root_usage.py
```diff
@@ -27,23 +27,23 @@ def execute(self) -> Check_Report_AWS:
                             datetime.datetime.now()
                             - datetime.datetime.strptime(
                                 user["password_last_used"],
-                                "%Y-%m-%dT%H:%M:%S+00:00",
+                                "%Y-%m-%dT%H:%M:%SZ",
```
jfagoagas: Python's https://dateutil.readthedocs.io/en/stable/ treats both formats equally, could be good to change it to prevent this from happening again.
`parser.parse(timestamp)`

pedrooot: Done!

jfagoagas (approved):
Neat!

PR #4457, opened 2024-07-16
chore(s3): create class and refactor

jfagoagas:
### Context

We are working to decouple all outputs and integrations and now is the turn for the S3 integration to send Prowler's output to a bucket.

### Description

- Create a new S3 class
- Send all the generated outputs to S3

### License


prowler/providers/aws/lib/quick_inventory/quick_inventory.py
```diff
@@ -285,23 +284,32 @@ def create_output(resources: list, provider: AwsProvider, args):
...
+        #     {args.output_directory} / {output_file + csv_file_suffix},
+        #     output_bucket,
+        #     object_name,
+        # )
+        # JSON
+        # s3_client.upload_file(
+        #     {args.output_directory} / {output_file + csv_file_suffix},
+        #     output_bucket,
+        #     object_name,
+        # )
```
jfagoagas: @sergargar could you please fix this. We should not use the S3 class anymore here since the QuickInventory has not an output created yet.

MrCloudSec: Done!

MrCloudSec (approved):
👏🏼 👏🏼 👏🏼

PR #4465, opened 2024-07-17
chore(output): review report function

pedrooot:
### Description

Report function reviewed and new tests added

### License


tests/lib/outputs/outputs_test.py
```diff
@@ -318,3 +323,57 @@ def test_extract_findings_statistics_all_fail_are_not_muted(self):
         assert stats["resources_count"] == 1
         assert stats["findings_count"] == 2
         assert not stats["all_fails_are_muted"]
+
+    def test_report_with_aws_provider(self):
```
jfagoagas: I'm not sure about the point of this test.

tests/lib/outputs/outputs_test.py
MrCloudSec, line 1: Could you add tests for all the use cases of the function?

pedrooot: True! I have to check multiple providers and findings info

tests/lib/outputs/outputs_test.py
pedrooot: I'll improve it to check the stdout output

prowler/lib/outputs/outputs.py
```diff
@@ -8,7 +8,7 @@ def stdout_report(finding, color, verbose, status, fix):
     if finding.check_metadata.Provider == "aws":
         details = finding.region
     if finding.check_metadata.Provider == "azure":
-        details = finding.check_metadata.ServiceName
+        details = finding.location
```
MrCloudSec, line 11: Why this?

tests/lib/outputs/outputs_test.py
MrCloudSec: @pedrooot please, add tests for GCP, Kubernetes and for the MANUAL finding too. Thanks!

prowler/lib/outputs/outputs.py
pedrooot: because we were using ServiceName since we didn't manage locations for Azure

tests/lib/outputs/outputs_test.py
pedrooot: Done!

MrCloudSec (approved):
👏🏼 👏🏼 👏🏼

PR #4477, opened 2024-07-18
fix(main): change module name

pedrooot:
### Description

`from prowler.lib.outputs.csv.models import CSV → from prowler.lib.outputs.csv.csv import CSV`

### License


jfagoagas (approved):
🥷

PR #4495, opened 2024-07-19
fix(inspector2): add more efficient way to check if any active findings

sergargar:
### Context
For the check `inspector2_active_findings_exist` we are getting all Inspector2 findings in the service just to check if there is any active finding, which can lead to API timeouts in big AWS accounts with thousands of findings.

### Description

Only check if there is one active inspector finding in the account and in the check Prowler can still verify whether the AWS account has active inspector findings or not, only by iterating one single finding.

### License


prowler/providers/aws/services/inspector2/inspector2_service.py
```diff
@@ -34,54 +33,30 @@ def __batch_get_account_status__(self, regional_client):
...
-                                        arn=finding["findingArn"],
-                                        region=regional_client.region,
-                                        severity=finding.get("severity"),
-                                        status=finding.get("status"),
-                                        title=finding.get("title"),
-                                    )
-                                )
+            regional_client = self.regional_clients[inspector.region]
+            active_findings = regional_client.list_findings(
+                filterCriteria={
```
jfagoagas, line 41: Is this `filterCriteria` applied at server side or client side?

jfagoagas (commented):
That looks great, I left some comments to clarify things.

prowler/providers/aws/services/inspector2/inspector2_service.py
sergargar: Server side, that is why is way faster now.

jfagoagas (approved):
Great catch @sergargar thanks for this!

jfagoagas (approved):

PR #4498, opened 2024-07-22
fix(sns): add condition to sns topics

pedrooot:
### Description

It's needed to add new conditions related with organizations. One example is: `"aws:PrincipalOrgID": "<our-org-id>"`
To apply this change we will need to add info for the check and modify the output status using is_condition_block_restrictive_organization

The final approach it's that if the condition has "*" for the "aws:PrincipalOrgID" it won't pass.

### License


prowler/providers/aws/lib/policy_condition_parser/policy_condition_parser.py
```diff
@@ -1,5 +1,8 @@
 def is_condition_block_restrictive(
-    condition_statement: dict, source_account: str, is_cross_account_allowed=False
+    condition_statement: dict,
+    source_account: str,
+    is_cross_account_allowed=False,
+    org_id: str = None,
```
jfagoagas: Do we need this? Can we not just check if the condition is set?

jfagoagas (changes requested):

prowler/providers/aws/lib/policy_condition_parser/policy_condition_parser.py
pedrooot: added org_id since the format for this sort of conditions where like: `"aws:resourceaccount": "<aws_account>"` but for the new condition the org_id it´s needed. This way we ensure that the condition it´s well applied.

tests/providers/aws/services/sns/sns_topics_not_publicly_accessible/sns_topics_not_publicly_accessible_test.py
```diff
@@ -226,3 +239,37 @@ def test_topic_public(self):
...
+                sns_topics_not_publicly_accessible,
+            )
+
+            check = sns_topics_not_publicly_accessible()
+            result = check.execute()
+            assert len(result) == 1
+            assert result[0].status == "PASS"
+            assert (
+                result[0].status_extended
+                == f"SNS topic {topic_name} is not public because its policy only allows access from the same account."
```
MrCloudSec: We will have to add another status, if they have an organization in the conditions it has cross-account access for all the accounts inside the org.

jfagoagas: Good catch!

MrCloudSec:
Please @pedrooot, update the PR with the explanation of this fix.

pedrooot:
Done! @sergargar

tests/providers/aws/services/sns/sns_topics_not_publicly_accessible/sns_topics_not_publicly_accessible_test.py
mlanders-spectra: I'd replace "from the same account" with "from a specific org or specific accounts"

jfagoagas: Agree with @mlanders-spectra.

poetry.lock
jfagoagas, line 1: Could you remove changes in this file?

prowler/providers/aws/services/sns/sns_topics_not_publicly_accessible/sns_topics_not_publicly_accessible.py
```diff
@@ -33,13 +33,18 @@ def execute(self):
...
-                            if (
-                                "Condition" in statement
-                                and is_condition_block_restrictive(
-                                    statement["Condition"], sns_client.audited_account
+                            if "Condition" in statement:
+                                account_condition, org_condition = (
+                                    is_condition_block_restrictive(
+                                        statement["Condition"],
+                                        sns_client.audited_account,
+                                        sns_client.audited_org_id,
```
jfagoagas: Where does the `audited_org_id` come from?

I see the point of calling that function with the AWS Account ID since is something we always have but we can not always know the Organization ID audited. In that case something must be passed from config if not set.

prowler/providers/aws/lib/policy_condition_parser/policy_condition_parser.py
jfagoagas: I'm more in favor of just checking if this condition is met. This function just checks if the `condition` block of an IAM policy is restrictive, which means that at least one of the conditions down there is set and with an specific value, not an `*`.

Also, I don't like the need of changing the function result since applies to many other checks.

jfagoagas (changes requested):

.github/workflows/pull-request.yml
```diff
@@ -73,7 +73,7 @@ jobs:
       - name: Safety
         if: steps.are-non-ignored-files-changed.outputs.any_changed == 'true'
         run: |
-          poetry run safety check --ignore 70612
+          poetry run safety check --ignore 70612,40459
```
jfagoagas: Why is this needed in this PR?

docs/tutorials/configuration_file.md
```diff
@@ -355,6 +356,10 @@ aws:
   # aws.acm_certificates_expiration_check
   days_to_expire_threshold: 7
 
+  # AWS SNS Configuration
+  # aws.sns_topics_not_publicly_accessible
+  organization_id: null
```
jfagoagas: What about several organizations sharing a resource?

prowler/providers/aws/lib/policy_condition_parser/policy_condition_parser.py
```diff
@@ -31,6 +36,7 @@ def is_condition_block_restrictive(
             "aws:sourcearn",
             "aws:sourcevpc",
             "aws:sourcevpce",
+            "aws:principalorgid",
```
jfagoagas: I think this is no longer needed right?

prowler/providers/aws/lib/policy_condition_parser/policy_condition_parser.py
```diff
@@ -41,6 +47,7 @@ def is_condition_block_restrictive(
             "aws:principalaccount",
             "aws:sourcevpc",
             "aws:sourcevpce",
+            "aws:principalorgid",
```
jfagoagas: I think this is no longer needed right?

prowler/providers/aws/lib/policy_condition_parser/policy_condition_parser.py
```diff
@@ -90,3 +97,76 @@ def is_condition_block_restrictive(
                                 is_condition_valid = True
 
     return is_condition_valid
+
+
+def is_condition_block_restrictive_organization(
```
jfagoagas, line 100: Why not to check just if the condition is present at least for now? I mean not to introduce more configuration and logic just to check if a condition is met.

jfagoagas: What do you think @sergargar?

prowler/providers/aws/lib/policy_condition_parser/policy_condition_parser.py
```diff
@@ -1,8 +1,10 @@
-def is_condition_block_restrictive(
-    condition_statement: dict, source_account: str, is_cross_account_allowed=False
+def is_condition_block_restrictive_account(
```
jfagoagas: I'd also leave this function with their original name. We can rename that later not to mix things in the PR.

jfagoagas (changes requested):

.github/workflows/pull-request.yml
pedrooot: I'll add it in other PR

prowler/providers/aws/lib/policy_condition_parser/policy_condition_parser.py
pedrooot: True

prowler/providers/aws/lib/policy_condition_parser/policy_condition_parser.py
pedrooot: True

prowler/providers/aws/lib/policy_condition_parser/policy_condition_parser.py
pedrooot: Done

prowler/providers/aws/lib/policy_condition_parser/policy_condition_parser.py
```diff
@@ -90,3 +95,64 @@ def is_condition_block_restrictive(
...
+                }
+
+                if value in condition_statement[condition_operator]:
+                    # values are a list
+                    if isinstance(
+                        condition_statement[condition_operator][value],
+                        list,
+                    ):
+                        is_condition_valid = True
+                        # i
```
jfagoagas:
```suggestion
```

tests/config/config_test.py
```diff
@@ -262,6 +262,7 @@ def mock_prowler_get_latest_release(_, **kwargs):
     ],
     "check_rds_instance_replicas": False,
     "days_to_expire_threshold": 7,
+    "organization_id": None,
```
jfagoagas:
```suggestion
```

jfagoagas (dismissed):
That's fantastic 👏

prowler/config/config.yaml
```diff
@@ -283,7 +283,6 @@ aws:
       "controllerManager",
       "scheduler",
     ]
```
MrCloudSec:
```suggestion
    ]

```

tests/config/fixtures/config.yaml
```diff
@@ -284,7 +284,6 @@ aws:
         "scheduler",
     ]
```
MrCloudSec:
```suggestion


```

MrCloudSec (approved):
👑 👑

PR #4500, opened 2024-07-22
fix(gcp): false positive for `iam_sa_no_administrative_privilege` check

### LINKED ISSUE #4393: False Positive in iam_sa_no_administrative_privileges Check
### Steps to Reproduce

Hey folks, when I run a Prowler scan on the IAM service in GCP, one of the checks, iam_sa_no_administrative_privileges, produces a false positive.

Steps to Reproduce
- Create a service account in the project, or use an existing service account.
-  Assign the` Access Approval Config Editor role` to the service account in the project.
Run a scan using the command below:
```bash
prowler gcp --project-ids <project-id> -b --services iam
```
- Open the result. You will find that the service account is shown as having admin privileges, even though it doesn't.

### Expected behavior

According to the CIS benchmark, for the admin role, we should check if the word is present. For the owner and editor roles, we should look for an exact match.
For example only `roles/Editor` should be considered as the admin role not the `Access Approval Config Editor role`

Service account with role that has name `editor` or `owner` has marked as the admin privileges

### Environment Resource

Running Locally

### OS used

Ubuntu 22.04.4 LTS

### Prowler version

Prowler 4.2.1

### Pip version

pip 20.3.4

### Context

_No response_

JOSHUAJEBARAJ:
### Context

Fix #4393

### Description

The `iam_sa_no_administrative_privileges` check previously flagged a vulnerability if a service account had roles containing "editor" or "owner". Now, it has been updated to use exact matching instead of partial string matching for these role names according to the CIS Benchmark: "each serviceaccount does not contain *Admin or *admin or does not match roles/editor or does not match roles/owner."

### License


jfagoagas:
Hello @JOSHUAJEBARAJ could you please add some tests or adapt the existing ones to handle this new behaviour? Could you also point out the documentation about this change?

Thanks for contributing with Prowler 🚀

MrCloudSec:
I have updated the PR description with the CIS documentation:

> each serviceaccount does not contain *Admin or *admin or does not match roles/editor or does not match roles/owner.

MrCloudSec (approved):
Thanks for the fix @JOSHUAJEBARAJ ! I have added the necessary tests for this PR.

PR #4507, opened 2024-07-22
fix(aws): only check artifacts that can be scanned for vulnerabilities by `ecr_repositories_scan_vulnerabilities_in_latest_image`

kagahd:
### Context

Until now, the check `ecr_repositories_scan_vulnerabilities_in_latest_image` just takes the newest pushed image from the ECR repository to check its scan report for vulnerabilities. However, the newest pushed image in a repository is not always an artifact that can be scanned, hence there is no report for it and the prowler check fails. 

### Description

AWS ECR is generally **able to scan** artifacts of the following **artifact** media types (not to confound with **manifest** media type):

- `application/vnd.docker.container.image.v1+json` Docker image configuration
- `application/vnd.docker.image.rootfs.diff.tar` Docker image layer as a tar archive
- `application/vnd.docker.image.rootfs.diff.tar.gzip` Docker image layer that is compressed using gzip
- `application/vnd.oci.image.config.v1+json` OCI image configuration
- `application/vnd.oci.image.layer.v1.tar` Uncompressed OCI image layer
- `application/vnd.oci.image.layer.v1.tar+gzip` Compressed OCI image layer

However, tools like [Google container tool Jib](https://github.com/GoogleContainerTools/jib), use `application/vnd.oci.image.config.v1+json` also for **signatures**, which are **not** scannable. Luckily, these are tagged with `sha-<HASH-CODE>.sig`,  so that they can still be easily recognized.

In contrast, **non-scannable** artifacts are for example:
- Image Indexes (Manifest Lists) `application/vnd.docker.distribution.manifest.list.v2+json`
  - An image index or manifest list groups multiple image versions for different platforms. The index itself does not contain layers and thus cannot be directly scanned. Instead, the individual images referenced by the index are scanned.
- Notary Signatures `application/vnd.cncf.notary.v2.signature`
  - Notary signatures are used to verify the integrity and authenticity of container images but do not contain image layers that can be scanned for vulnerabilities.
- OCI Artifact Types (such as Helm Charts) `application/vnd.oci.artifact.v1+json`
  - General OCI artifacts like Helm charts are not container images and do not contain layers in the traditional sense. They cannot be directly scanned for vulnerabilities, although other tools or processes might perform security checks on specific types of artifacts.

For example, if a **signed** container image was built with the [Google container tool Jib](https://github.com/GoogleContainerTools/jib), multiple artifacts are pushed to the registry.

AWS_ECR_signed_image

As you can see in the screenshot, the artifacts are ordered by `Pushed at`. The four first artifacts are **not scannable** because they are not Docker container images. The first three are signatures and the fourth is an "Image Index". All these four artifacts do not contain any layers that could be scanned.

Furthermore, as you can see in the screenshot, the fifth artifact, which is a scannable Docker container image, **does not have any tags** which is represented by the `-` sign. This can have several reasons as for example:

- **security considerations**
  - In certain cases, images could be uploaded without tags to ensure that only the digest is used, which provides a higher level of security. The digest guarantees that the image is unchanged, while tags can be changed manually. 
- **backwards compatibility and image promotion** 
  - Sometimes images are intentionally uploaded without tags to allow for later promotion. This may be part of a strategy to ensure backwards compatibility or to manage image versioning. For example, an image is initially uploaded without a tag and a tag is added after successful testing.

However, the current prowler report refers to an image tag that not only may not be present, but is also used by different artifacts, so the user would have difficulty understanding which image is meant in the prowler report (the prowler report also didn't mention that the newest image of that tag is meant).

#### Implemented solution

1. The check `ecr_repositories_scan_vulnerabilities_in_latest_image` verifies now the scan report of newest pushed **scannable** artifact.
2. The prowler report refers now to the **image digest**. However, the image tag ist still mentioned in the report. The image digest is a cryptographic hash value that uniquely identifies the content of a Docker image. The digest changes when the content of the image changes, even if the image tag remains the same. The user can therefore easily identify the image to which the prowler report refers.

### License


jfagoagas:
WOW @kagahd this is impressive 🤯 Thank you for this contribution, we will review it and get back to you!

Thanks for contributing with Prowler always taking care of it ❤️

prowler/providers/aws/services/ecr/ecr_service.py
```diff
@@ -141,6 +141,9 @@ def __get_image_details__(self, regional_client):
                                 if image is not None:
                                     severity_counts = None
                                     last_scan_status = None
+                                    artifact_media_type = image.get(
```
jfagoagas, line 142: This got my attention without having started the PR review. 

What if we just store the artifacts with media type `"application/vnd.docker.container.image.v1+json"`? We are not doing anything else with the rest of media types so  for the sake of memory storage and speed I think it'd be better just to store what we are going to analyze. Then in the future we can extend it.

kagahd: > What if we just store the artifacts with media type `"application/vnd.docker.container.image.v1+json"`?

I was not sure if prowler is using other types of artifacts but if prowler only cares about Docker images, then yes, the `ecr_service.py` could sort them out early already to save RAM and CPU cycles.

kagahd: It seems that detecting a scannable image is not as easy as I initially thought because some artifact media types are used for scannable and not scannable artifacts 😞 I have to dive deeper to solve this problem.
Furthermore, AWS introduced a new version of basic scanning which does not support imageScanFindingsSummary and imageScanStatus in the DescribeImages API. To view these, we have to use the DescribeImageScanFindings API. 
I'll adapt prowler's `ecr_servive.py` to take this change into account.
So maybe it's better to **review this PR a bit later** when I'm done with it.

jfagoagas: That's totally fine! Thank you very much for your work on this it an absolutely pleasure to work with you 🚀 

I'm going to convert the PR to `draft` and once it will be ready for review you can mark it as `ready for review` again.

Thanks a lot!

kagahd: ready for review 😄

jfagoagas: Great, I'll see if @sergargar can review that today, if not we will review it on Friday. Thanks for your work Kay!

jfagoagas:
Hello @kagahd I'm in the middle of the review but I'm not sure I'll be able to finish it today. If not I'll do that on Monday.

Thank you!

kagahd:
> Hello @kagahd I'm in the middle of the review but I'm not sure I'll be able to finish it today. If not I'll do that on Monday.

Hello @jfagoagas, thanks for the info. I won't be at work for the next few weeks anyway, so there's no rush 😄

tests/providers/aws/services/ecr/ecr_repositories_scan_vulnerabilities_in_latest_image/ecr_repositories_scan_vulnerabilities_in_latest_image_test.py
MrCloudSec, line 1: Please @kagahd, add a new test for a non scannable artifact.

prowler/providers/aws/services/ecr/ecr_service.py
```diff
@@ -139,48 +160,92 @@ def __get_image_details__(self, regional_client):
...
+
+                                        # If imageScanStatus is not present or imageScanFindingsSummary is missing,
+                                        # we need to call DescribeImageScanFindings because AWS' new version of
+                                        # basic scanning does not support imageScanFindingsSummary and imageScanStatus
+                                        # in the DescribeImages API.
+                                        if "imageScanStatus" not in image:
+                                            try:
+                                                # use "image" for scan findings to get data the same way as for an image
+                                                image = (
+                                                    client.describe_image_scan_findings(
```
MrCloudSec, line 172: Please, create a new function in the ECR class for this API Call.

prowler/providers/aws/services/ecr/ecr_service.py
```diff
@@ -121,6 +121,27 @@ def __get_repository_lifecycle_policy__(self, regional_client):
 
     def __get_image_details__(self, regional_client):
         logger.info("ECR - Getting images details...")
+
+        def is_artifact_scannable(artifact_media_type, tags):
```
MrCloudSec: Please, put this function outside of the parent function and use try/except.

MrCloudSec (changes requested):
Love this @kagahd ❤️ Please, review my comments.

jfagoagas:
No new permissions needed, `describe:*` covers it.

MrCloudSec (approved):

PR #4522, opened 2024-07-23
fix(checks): ensure CheckID is correct in check's metadata

MrCloudSec:
### Context

When doing `prowler <provider> -l ` some checks were not appearing because the CheckID was incorrect in check's metadata.

### Description

Make sure those checks appear in the listing and do a test to prevent this in the future.

### License

### DIFF
--- a/tests/lib/check/check_test.py
+++ b/tests/lib/check/check_test.py
@@ -9,7 +9,6 @@ from pkgutil import ModuleInfo
 
 from boto3 import client
 from colorama import Fore, Style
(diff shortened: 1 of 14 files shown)


tests/lib/check/check_test.py
jfagoagas, line 1: If the code is the same for all the providers, you can create a function within the test class and then call it from each test. What do you think?

jfagoagas (approved):
That's fantastic 👏

PR #4530, opened 2024-07-24
fix(s3): enhance threading in s3 service

MrCloudSec:
### Description

Enhance threading in S3 service so it is handled with our global function.

### License

### DIFF
--- a/prowler/providers/aws/services/s3/s3_service.py
+++ b/prowler/providers/aws/services/s3/s3_service.py
@@ -1,5 +1,4 @@
 import json
-import threading
 from typing import Optional
 
 from botocore.client import ClientError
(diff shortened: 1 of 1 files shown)


pedrooot (commented):
Loving this! 😍 Do you think that we should add `.get` for the attributes?
For example:
`bucket_versioning["Status"]` -> `bucket_versioning.get("Status", <bool or string>)`

jfagoagas (approved):
Thanks for taking care of this!

MrCloudSec:
> Loving this! 😍 Do you think that we should add `.get` for the attributes? For example: `bucket_versioning["Status"]` -> `bucket_versioning.get("Status", <bool or string>)`

Good idea! But it is out of the scope of this PR.

PR #4550, opened 2024-07-26
test(GCP): Add remaining GCP tests for KMS checks

danibarranqueroo:
### Context

There was some checks without tests for KMS in GCP provider.

### Description

Added new tests for the following checks:

- [x] `kms_key_not_publicly_accessible check`
- [x] `kms_key_rotation_enabled check`

### License

### DIFF
--- a/prowler/providers/gcp/services/kms/kms_key_not_public_access/kms_key_not_public_access.json
+++ b/prowler/providers/gcp/services/kms/kms_key_not_public_access/kms_key_not_public_access.json
@@ -1,6 +1,6 @@
 {
   "Provider": "gcp",
-  "CheckID": "kms_key_not_publicly_accessible",
+  "CheckID": "kms_key_not_public_access",
   "CheckTitle": "Check for Publicly Accessible Cloud KMS Keys",
   "CheckType": [],
   "ServiceName": "kms",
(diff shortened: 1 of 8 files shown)


danibarranqueroo:
The GitHub checks failed because there was an existing AWS KMS test with the same name as this new GCP KMS test. I’ve changed the name of the GCP KMS `kms_key_not_publicly_accessible` check and test to `kms_key_not_public_access`, and it should work correctly now.

jfagoagas:
> The GitHub checks failed because there was an existing AWS KMS test with the same name as this new GCP KMS test. I’ve changed the name of the GCP KMS `kms_key_not_publicly_accessible` check and test to `kms_key_not_public_access`, and it should work correctly now.

I think you can also rename the test class file with something else to avoid this and be able to have the same check name for both providers. I think is only an issue with how `pytest` looks for test files. There should be some similar already addressed cases for IAM.

danibarranqueroo:
Hi @jfagoagas!
I’ve followed your suggestions and updated the names accordingly. I kept the old name for the check and only changed the test file and class names by adding "gcp" to avoid pytest failures.

MrCloudSec (approved):
Great job @danibarranqueroo !

PR #4553, opened 2024-07-26
feat(aws): make check `eks_control_plane_logging_all_types_enabled` configurable

kagahd:
### Context

The prowler check `eks_control_plane_logging_all_types_enabled` verifies that **all** possible AWS EKS control plan logging types are enabled and fails if not.

### Description

AWS EKS Control plane has five different log types:
- API server
  - Logs pertaining to API requests to the cluster.
- Audit
  - Logs pertaining to cluster access via the Kubernetes API.
- Authenticator
  - Logs pertaining to authentication requests into the cluster.
- Controller manager
  - Logs pertaining to state of cluster controllers.
- Scheduler 
  - Logs pertaining to scheduling decisions.

While we understand that all five log types can be beneficial to improve IT security, they are of varying importance for IT security. Since logging can be expensive depending on the use case, companies should be able to decide which log types they require for optimum IT security at balanced cost. 
Therefore the check `eks_control_plane_logging_all_types_enabled` should be configurable, which this PR implements.
The default configuration is still that all log types must be enabled to PASS the check, so nothing changes, unless the user decides otherwise by customizing their prowler configuration specifically to their needs.

### License


MrCloudSec (approved):
Love this @kagahd ! The main reason of that logic was because the control 2.1.1 of the CIS Amazon Elastic Kubernetes Service (EKS) Benchmark requires all log types to be enabled.

PR #4563, opened 2024-07-29
feat(aws): Add a test_connection method

jfagoagas:
### Context

We need to be able to test the connection to each provider using one of the available authentication methods, starting from the environment variables.

### Description

- Add a `test_connection` method for the AWS provider.
- Create a `test_connection` abstract method in the `Provider` class but not enforced for now not to break anything.

### Notes for reviewers
- Review the removal of `sys.exit()` in favor of catching the exceptions during the `__init__`.
- You can test it with the following:
```python
from prowler.providers.aws.aws_provider import AwsProvider


# - Using environment variables
t = AwsProvider.test_connection()
print(t)

# - Using profile
t = AwsProvider.test_connection(profile="dev")
print(t)


# - With MFA
t = AwsProvider.test_connection(profile="dev", mfa_enabled=True)
print(t)


# - With Role without external ID
t = AwsProvider.test_connection(
    role_arn="arn:aws:iam::111122223333:role/ProwlerRole"
)
print(t)

# - With Role with external ID from ENV
t = AwsProvider.test_connection(
    role_arn="arn:aws:iam::111122223333:role/ProwlerRole",
    external_id="8eabb29e-e591-4828-919f-a9b2839a137d",
)
print(t)


t = AwsProvider.test_connection(
    role_arn="arn:aws:iam::111122223333:role/ProwlerRole",
    external_id="8eabb29e-e591-4828-919f-a9b2839a137d",
    session_duration=900,
    profile="test",
)
print(t)
```

### License

### DIFF
--- a/prowler/providers/aws/aws_provider.py
+++ b/prowler/providers/aws/aws_provider.py
@@ -10,6 +10,7 @@ from boto3 import client
 from boto3.session import Session
 from botocore.config import Config
 from botocore.credentials import RefreshableCredentials
+from botocore.exceptions import ClientError, ProfileNotFound
 from botocore.session import Session as BotocoreSession
 from colorama import Fore, Style
 from pytz import utc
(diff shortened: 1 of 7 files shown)


prowler/providers/aws/aws_provider.py
```diff
@@ -106,9 +107,13 @@ def __init__(self, arguments: Namespace):
             self.session.current_session.region_name, input_regions
         )
 
-        caller_identity = validate_aws_credentials(
+        connected, caller_identity = self.test_connection(
             self.session.current_session, sts_region
         )
+        # TODO: review this for CLI and SDK usage
+        if not connected:
```
MrCloudSec: I would raise a critical error here.

MrCloudSec: Or maybe not using the boolean, only for the API.

jfagoagas: The `test_connection` raises an ERROR log, I think with that is more than enough. We need to think how to handle that for the CLI.

MrCloudSec: why then using the `connected` attribute here?

jfagoagas: because we need to abort the provider's execution if it is not connected.

prowler/providers/aws/aws_provider.py
```diff
@@ -881,7 +886,134 @@ def get_checks_to_execute_by_audit_resources(self) -> set[str]:
...
+                account=caller_identity.get("Account"),
+                arn=ARN(caller_identity.get("Arn")),
+                region=aws_region,
+            )
+
+        except (ClientError, ProfileNotFound, ArgumentTypeError) as error:
+            logger.error(
+                f"{error.__class__.__name__}[{error.__traceback__.tb_lineno}]: {error}"
+            )
+            raise error
```
vicferpoy: Can you rename `error` to something more descriptive? This is a _controlled_ error and somehow _expected_ error. Just for clarity. Something like `credentials_error` or something like that. Also, ideally in the future, this raised exception should be one of the custom SDK exceptions so the API can identify them easier.

prowler/providers/aws/aws_provider.py
```diff
@@ -881,7 +886,134 @@ def get_checks_to_execute_by_audit_resources(self) -> set[str]:
...
+        - session.Session.client: The STS session client.
+
+        Example:
+            session = boto3.session.Session()
+            sts_client = create_sts_session(session, 'us-west-2')
+        """
+        try:
+            sts_endpoint_url = (
+                f"https://sts.{aws_region}.amazonaws.com"
+                if "cn-" not in aws_region
```
vicferpoy: This could potentially lead to unwanted behaviors, unless you are 100% certain about it. Consider using compiled regexes, `.startswith` (if that is a prefix) or something similar. Just a suggestion, not a requested change.

vicferpoy (changes requested):
This looks good! I left some comments and I also wondered how is the API going to use this? 

As far as I know, we only store the `provider_id` (relevant to what I see in the `test_connection ` expected arguments). Would that be the `external_id` field? Would that be enough to test the connection?

prowler/providers/aws/aws_provider.py
vicferpoy: I think I highlighted too many lines. I was referring to:

```python
if "cn-" not in aws_region
```

jfagoagas:
> This looks good! I left some comments and I also wondered how is the API going to use this?
> 
> As far as I know, we only store the `provider_id` (relevant to what I see in the `test_connection ` expected arguments). 

It mostly depends on the AWS Credentials method used, you can see more about it here https://boto3.amazonaws.com/v1/documentation/api/latest/guide/credentials.html. Currently Prowler supports out of the box methods 3, 4, 5 and 6. For example using the environment variables or the profile/config will make the SDK to automatically look for them and then try to authenticate with AWS.


> Would that be the `external_id` field? Would that be enough to test the connection?

The `external_id` will be used while assuming AWS IAM roles, you can get more information about it in the following links:
- https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_create_for-user_externalid.html
- https://docs.aws.amazon.com/IAM/latest/UserGuide/confused-deputy.html
It won't be enough just with the `external_id` itself, it needs to be used in conjunction with an AWS IAM Role ARN.

prowler/providers/aws/aws_provider.py
jfagoagas: sure thing!

prowler/providers/aws/aws_provider.py
jfagoagas: Great catch 👏

vicferpoy (dismissed):
LGTM! Thanks for the explanation too!

jfagoagas:
We are still discussing about the format of the `test_connection`, probably this is going to be the chosen one:
```python
@staticmethod
    def test_connection(
        session: Session = None,
        profile: str = None,
        aws_region: str = AWS_STS_GLOBAL_ENDPOINT_REGION,
        role_arn: str = None,
        role_session_name: str = ROLE_SESSION_NAME,
        session_duration: int = 3600,
        external_id: str = None,
        mfa_enabled: bool = False,
        raise_on_exception: bool = True,
    ) -> TestConnection:
        """
        Validates AWS credentials using the provided session and AWS region.

        If no session is provided, the method will create a new session using the Boto3 default session.

        Args:
            session (Session): The AWS session object.
            profile (str): The AWS profile to use for the session.
            aws_region (str): The AWS region to validate the credentials in.
            role_arn (str): The ARN of the IAM role to assume.
            role_session_name (str): The name of the role session.
            session_duration (int): The duration of the assumed role session in seconds.
            external_id (str): The external ID to use when assuming the role.
            mfa_enabled (bool): Whether MFA (Multi-Factor Authentication) is enabled.
            raise_on_exception (bool): Whether to raise an exception if an error occurs.

        Returns:

            TestConnection: A named tuple containing the result of the validation.
                - connected (bool): Indicates whether the validation was successful.
                - result (AWSCallerIdentity): An object representing the caller's identity if the validation was successful.
                - error (Exception): An exception object if an error occurs during the validation.

        Raises:
            Exception: If an error occurs during the validation process.

        Examples:
            >>> AwsProvider.test_connection(
                role_arn="arn:aws:iam::111122223333:role/ProwlerRole",
                external_id="67f7a641-ecb0-4f6d-921d-3587febd379c"
            )
            AWSCallerIdentity(user_id='AROAAAAAAAAAAAAAAAAAA:ProwlerAssessmentSession', account='111122223333', arn=ARN(arn='arn:aws:sts::111122223333:assumed-role/ProwlerRole/ProwlerAssessmentSession', partition='aws', service='sts', region=None, account_id='111122223333', resource='ProwlerRole/ProwlerAssessmentSession', resource_type='assumed-role'), region='us-east-1')

            >>> AwsProvider.test_connection(profile="test")
            AWSCallerIdentity(user_id='AROAAAAAAAAAAAAAAAAAA:test-user', account='111122223333', arn=ARN(arn='arn:aws:sts::111122223333:user/test-user', partition='aws', service='sts', region=None, account_id='111122223333', resource='test-user', resource_type='user'), region='us-east-1')
        """
        try:
            # Create the default session if no session is given
            session = (
                session if session else AwsProvider.setup_session(mfa_enabled, profile)
            )

            # Test Connection using the IAM Role
            if role_arn:
                session_duration = validate_session_duration(session_duration)
                role_session_name = validate_role_session_name(role_session_name)
                role_arn = parse_iam_credentials_arn(role_arn)
                assumed_role_information = AWSAssumeRoleInfo(
                    role_arn=role_arn,
                    session_duration=session_duration,
                    external_id=external_id,
                    mfa_enabled=mfa_enabled,
                    role_session_name=role_session_name,
                )
                assumed_role_credentials = AwsProvider.assume_role(
                    session,
                    assumed_role_information,
                )
                session = Session(
                    aws_access_key_id=assumed_role_credentials.aws_access_key_id,
                    aws_secret_access_key=assumed_role_credentials.aws_secret_access_key,
                    aws_session_token=assumed_role_credentials.aws_session_token,
                    region_name=aws_region,
                    profile_name=profile,
                )

            sts_client = AwsProvider.create_sts_session(session, aws_region)
            caller_identity = sts_client.get_caller_identity()
            # Include the region where the caller_identity has validated the credentials
            return TestConnection(
                connected=True,
                result=AWSCallerIdentity(
                    user_id=caller_identity.get("UserId"),
                    account=caller_identity.get("Account"),
                    arn=ARN(caller_identity.get("Arn")),
                    region=aws_region,
                ),
            )

        except (ClientError, ProfileNotFound) as credentials_error:
            logger.error(
                f"{credentials_error.__class__.__name__}[{credentials_error.__traceback__.tb_lineno}]: {credentials_error}"
            )
            if raise_on_exception:
                raise credentials_error
            else:
                return TestConnection(error=credentials_error)

        except ArgumentTypeError as validation_error:
            logger.error(
                f"{validation_error.__class__.__name__}[{validation_error.__traceback__.tb_lineno}]: {validation_error}"
            )
            if raise_on_exception:
                raise validation_error
            else:
                return TestConnection(error=validation_error)

        except Exception as error:
            logger.critical(
            )
            raise error
```

and the `TestConnection` object:
```python
@dataclass
class TestConnection:
    _connected: bool = False
    _error: Exception = None
    _result: Any = None

    @property
    def connected(self) -> bool:
        return self._connected

    @property
    def error(self) -> Exception:
        return self._error

    @property
    def result(self) -> Any:
        return self._result
```

This PR needs to be updated accordingly.

Also the init should call it with:
```
caller_identity = self.test_connection(
            session=self.session.current_session,
            aws_region=sts_region,
            raise_exception=True,
        )
```

jfagoagas:
Do not merge until the above change is addressed.

prowler/providers/common/test_connection_dataclass.py
```diff
@@ -0,0 +1,9 @@
+from dataclasses import dataclass
+from typing import Any
+
+
+@dataclass
+class TestConnection:
+    connected: bool = False
+    error: Exception = None
+    result: Any = None
```
thejaywhy: I'm curious how the `result` should be used? 

Is it needed? For all providers?

vicferpoy: I don't like using `Any` outside the testing scope. I understand it could be a placeholder until we have every provider test_connection implemented but if possible, let's change it in the future.

prowler/providers/common/test_connection_dataclass.py
vicferpoy, line 1: Modules with the `test_` prefix are used for testing. Could we add this dataclass to another module? Or at least rename it.

prowler/providers/aws/aws_provider.py
```diff
@@ -881,7 +890,153 @@ def get_checks_to_execute_by_audit_resources(self) -> set[str]:
...
+                ),
+            )
+
+        except (ClientError, ProfileNotFound) as credentials_error:
+            logger.error(
+                f"{credentials_error.__class__.__name__}[{credentials_error.__traceback__.tb_lineno}]: {credentials_error}"
+            )
+            if raise_on_exception:
+                raise credentials_error
+            else:
```
vicferpoy: Not needed. We can reduce indentation.

prowler/providers/aws/aws_provider.py
```diff
@@ -881,7 +890,153 @@ def get_checks_to_execute_by_audit_resources(self) -> set[str]:
...
+                return TestConnection(error=credentials_error)
+
+        except ArgumentTypeError as validation_error:
+            logger.error(
+                f"{validation_error.__class__.__name__}[{validation_error.__traceback__.tb_lineno}]: {validation_error}"
+            )
+            if raise_on_exception:
+                raise validation_error
+            else:
```
vicferpoy: Same here.

vicferpoy (changes requested):
This implementation looks better, great job. I left some small requested changes.

prowler/providers/aws/aws_provider.py
jfagoagas: Done.

prowler/providers/aws/aws_provider.py
jfagoagas: Done.

prowler/providers/common/test_connection_dataclass.py
jfagoagas: Done.

prowler/providers/common/test_connection_dataclass.py
jfagoagas: I decided to remove the `result` object since the API does not need it, the SDK will not use the test-connection at least to init the provider.

jfagoagas: Not needed at the end, since this method is just a `test-connection` and makes no sense to return credentials, it is not a `setup_credentials` method.

prowler/providers/aws/aws_provider.py
```diff
@@ -881,7 +888,174 @@ def get_checks_to_execute_by_audit_resources(self) -> set[str]:
...
+                role_arn="arn:aws:iam::111122223333:role/ProwlerRole",
+                external_id="67f7a641-ecb0-4f6d-921d-3587febd379c"
+            )
+            AWSCallerIdentity(user_id='AROAAAAAAAAAAAAAAAAAA:ProwlerAssessmentSession', account='111122223333', arn=ARN(arn='arn:aws:sts::111122223333:assumed-role/ProwlerRole/ProwlerAssessmentSession', partition='aws', service='sts', region=None, account_id='111122223333', resource='ProwlerRole/ProwlerAssessmentSession', resource_type='assumed-role'), region='us-east-1')
+
+            >>> AwsProvider.test_connection(profile="test")
+            AWSCallerIdentity(user_id='AROAAAAAAAAAAAAAAAAAA:test-user', account='111122223333', arn=ARN(arn='arn:aws:sts::111122223333:user/test-user', partition='aws', service='sts', region=None, account_id='111122223333', resource='test-user', resource_type='user'), region='us-east-1')
+        """
+        try:
+            # Create the default session if no session is given
```
pedrooot:
```suggestion
```

prowler/providers/aws/aws_provider.py
```diff
@@ -881,7 +888,174 @@ def get_checks_to_execute_by_audit_resources(self) -> set[str]:
...
+            )
+
+        except (ClientError, ProfileNotFound, NoCredentialsError) as credentials_error:
+            logger.error(
+                f"{credentials_error.__class__.__name__}[{credentials_error.__traceback__.tb_lineno}]: {credentials_error}"
+            )
+            if raise_on_exception:
+                raise credentials_error
+
+            return Connection(error=credentials_error)
```
pedrooot, line 1013:
```suggestion
            return Connection(is_connected=False, error=credentials_error)
```

prowler/providers/aws/aws_provider.py
```diff
@@ -881,7 +888,174 @@ def get_checks_to_execute_by_audit_resources(self) -> set[str]:
...
+
+        except ArgumentTypeError as validation_error:
+            logger.error(
+                f"{validation_error.__class__.__name__}[{validation_error.__traceback__.tb_lineno}]: {validation_error}"
+            )
+            if raise_on_exception:
+                raise validation_error
+
+            return Connection(error=validation_error)
```
pedrooot, line 1022:
```suggestion
            return Connection(is_connected=False, error=validation_error)
```
Same here

pedrooot (commented):
I left some comments but 😍

prowler/providers/aws/aws_provider.py
jfagoagas: `Connection.is_connected` defaults to `False`.

prowler/providers/aws/aws_provider.py
jfagoagas: `Connection.is_connected` defaults to `False`.

pedrooot: Fact

prowler/providers/aws/aws_provider.py
pedrooot: Fact

pedrooot (approved):
🥇 🐐

vicferpoy (approved):
Great job!

PR #4570, opened 2024-07-30
feat(opensearch): Add domain inside VPC case for public domain check

### LINKED ISSUE #4494: False Positive: AWS Opensearch domain is reported as publicly accessible when Opensearch is within VPC
### Steps to Reproduce

An Amazon Opensearch service domain within a VPC.
Running prowler reports that `Opensearch domain {domain.name} policy allows access (Principal: '*').`

`python -m prowler aws -c opensearch_service_domains_not_publicly_accessible`

### Expected behavior

Prowler should either detect that the service domain is running within a VPC (and therefore is not inherently publicly accessible), or provide a false-positive notice.

```json
{
...
 "severity_id": 5,
 "severity": "Critical",
 "status": "New",
 "status_code": "FAIL",
 "status_detail": "Opensearch domain test-es policy allows access (Principal: '*').",
 "status_id": 1,
...
}
```

### Environment Resource

Workstation

### OS used

Kali Linux

### Prowler version

4.2.4

### Pip version

24.0

### Context

You can't apply IP-based access policies to domains that reside within a VPC because security groups already enforce IP-based access policies.

[Reference](https://docs.aws.amazon.com/opensearch-service/latest/developerguide/vpc.html)

puchy22:
### Context

Check `opensearch_service_domains_not_publicly_accessible` did not take into account that a domain was inside a VPC, which would make it non-publicly accessible regardless of the policies.

Fixes https://github.com/prowler-cloud/prowler/issues/4494
### Description

- Change `endpoint_vpc` to `vpc_endpoints` and support for multiple endpoints
- Add new verification if domain is inside of VPC
- Add new check with this case

### License

### DIFF
--- a/prowler/providers/aws/services/opensearch/opensearch_service.py
+++ b/prowler/providers/aws/services/opensearch/opensearch_service.py
@@ -91,12 +91,16 @@ class OpenSearchService(AWSService):
                     DomainName=domain.name
                 )
                 domain.arn = describe_domain["DomainStatus"]["ARN"]
-                domain.endpoint_vpc = None
+                domain.vpc_endpoints = None
                 if "Endpoints" in describe_domain["DomainStatus"]:
                     if "vpc" in describe_domain["DomainStatus"]["Endpoints"]:
-                        domain.endpoint_vpc = describe_domain["DomainStatus"][
-                            "Endpoints"
-                        ]["vpc"]
+                        domain.vpc_endpoints = [
+                            vpc
+                            for vpc in describe_domain["DomainStatus"][
+                                "Endpoints"
+                            ].values()
+                        ]
+
                 domain.vpc_id = None
                 if "VPCOptions" in describe_domain["DomainStatus"]:
                     domain.vpc_id = describe_domain["DomainStatus"]["VPCOptions"][
@@ -156,7 +160,7 @@ class OpenSearchDomain(BaseModel):
     region: str
     arn: str = None
     logging: list[PublishingLoggingOption] = []
-    endpoint_vpc: str = None
+    vpc_endpoints: list[str] = None
     vpc_id: str = None
     access_policy: dict = None
     cognito_options: bool = None
--- a/prowler/providers/aws/services/opensearch/opensearch_service.py
+++ b/prowler/providers/aws/services/opensearch/opensearch_service.py
@@ -53,12 +53,9 @@ class OpenSearchService(AWSService):
                     "INDEX_SLOW_LOGS",
                     "AUDIT_LOGS",
                 ]:
-                    if (
-                        logging_key
-                        in describe_domain["DomainConfig"]["LogPublishingOptions"][
-                            "Options"
-                        ]
-                    ):
+                    if logging_key in describe_domain["DomainConfig"].get(
+                        "LogPublishingOptions", {}
+                    ).get("Options", {}):
                         domain.logging.append(
                             PublishingLoggingOption(
                                 name=logging_key,
--- a/prowler/providers/aws/services/opensearch/opensearch_service_domains_not_publicly_accessible/opensearch_service_domains_not_publicly_accessible.py
+++ b/prowler/providers/aws/services/opensearch/opensearch_service_domains_not_publicly_accessible/opensearch_service_domains_not_publicly_accessible.py
@@ -19,8 +19,7 @@ class opensearch_service_domains_not_publicly_accessible(Check):
             )
 
             if domain.vpc_id:
-                report.status = "PASS"
-                report.status_extended = f"Opensearch domain {domain.name} is in a VPC."
+                report.status_extended = f"Opensearch domain {domain.name} is in a VPC, then it is not publicly accessible."
             elif domain.access_policy:
                 for statement in domain.access_policy["Statement"]:
                     # look for open policies
--- a/prowler/providers/aws/services/opensearch/opensearch_service_domains_not_publicly_accessible/opensearch_service_domains_not_publicly_accessible.py
+++ b/prowler/providers/aws/services/opensearch/opensearch_service_domains_not_publicly_accessible/opensearch_service_domains_not_publicly_accessible.py
@@ -17,7 +17,11 @@ class opensearch_service_domains_not_publicly_accessible(Check):
             report.status_extended = (
                 f"Opensearch domain {domain.name} does not allow anonymous access."
             )
-            if domain.access_policy:
+
+            if domain.vpc_id:
+                report.status = "PASS"
+                report.status_extended = f"Opensearch domain {domain.name} is in a VPC."
+            elif domain.access_policy:
                 for statement in domain.access_policy["Statement"]:
                     # look for open policies
                     if (
--- a/tests/providers/aws/services/opensearch/opensearch_service_domains_not_publicly_accessible/opensearch_service_domains_not_publicly_accessible_test.py
+++ b/tests/providers/aws/services/opensearch/opensearch_service_domains_not_publicly_accessible/opensearch_service_domains_not_publicly_accessible_test.py
@@ -308,14 +308,14 @@ class Test_opensearch_service_domains_not_publicly_accessible:
             assert result[0].resource_tags == []
 
     def test_domain_inside_vpc(self):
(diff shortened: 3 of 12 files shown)


prowler/providers/aws/services/opensearch/opensearch_service.py
```diff
@@ -91,12 +91,16 @@ def __describe_domain__(self, regional_clients):
...
-                        domain.endpoint_vpc = describe_domain["DomainStatus"][
-                            "Endpoints"
-                        ]["vpc"]
+                        domain.vpc_endpoints = [
+                            vpc
+                            for vpc in describe_domain["DomainStatus"][
+                                "Endpoints"
+                            ].values()
+                        ]
+
```
jfagoagas, line 100: What is the reason behind this?

tests/providers/aws/services/opensearch/opensearch_service_domains_not_publicly_accessible/opensearch_service_domains_not_publicly_accessible_test.py
```diff
@@ -246,3 +246,35 @@ def test_policy_data_not_restricted_whole_internet(self):
...
+            check = opensearch_service_domains_not_publicly_accessible()
+            result = check.execute()
+            assert len(result) == 1
+            assert result[0].status == "PASS"
+            assert (
+                result[0].status_extended
+                == f"Opensearch domain {domain_name} is in a VPC, then it is not publicly accessible."
+            )
+            assert result[0].resource_id == domain_name
+            assert result[0].resource_arn == domain_arn
```
jfagoagas: Please assert also the region and the tags. Review if they are asserted in the rest of tests.

prowler/providers/aws/services/opensearch/opensearch_service.py
```diff
@@ -156,7 +160,7 @@ class OpenSearchDomain(BaseModel):
     region: str
     arn: str = None
     logging: list[PublishingLoggingOption] = []
-    endpoint_vpc: str = None
+    vpc_endpoints: list[str] = None
```
jfagoagas, line 160: This name could be confusing, it is referring to VPC Endpoints?

jfagoagas (dismissed):

prowler/providers/aws/services/opensearch/opensearch_service.py
puchy22: This is because I do not understand why only one vpc_endpoint was supported when several could have been, moreover IPv4 VPCs only had in count with these changes we have all in count. However, I do not understand why this attribute was set as it is not used in any checks for now. Should I remove it?

jfagoagas: If we are sure that this is not being used in the checks please remove it. Double check this with @sergargar please.

prowler/providers/aws/services/opensearch/opensearch_service.py
puchy22: Yes, it is referring VPC Endpoints used by OpenSearch domain.

prowler/providers/aws/services/opensearch/opensearch_service.py
MrCloudSec: I confirm that `endpoint_vpc` is not used.

prowler/providers/aws/services/opensearch/opensearch_service.py
```diff
@@ -91,12 +91,16 @@ def __describe_domain__(self, regional_clients):
...
                 if "Endpoints" in describe_domain["DomainStatus"]:
                     if "vpc" in describe_domain["DomainStatus"]["Endpoints"]:
-                        domain.endpoint_vpc = describe_domain["DomainStatus"][
-                            "Endpoints"
-                        ]["vpc"]
+                        domain.vpc_endpoints = [
+                            vpc
+                            for vpc in describe_domain["DomainStatus"][
+                                "Endpoints"
+                            ].values()
```
MrCloudSec, line 98: Have you check if the `Endpoints` value can have endpoints that are not from VPC?

prowler/providers/aws/services/opensearch/opensearch_service_domains_not_publicly_accessible/opensearch_service_domains_not_publicly_accessible.py
```diff
@@ -17,7 +17,10 @@ def execute(self):
             report.status_extended = (
                 f"Opensearch domain {domain.name} does not allow anonymous access."
             )
-            if domain.access_policy:
+
+            if domain.vpc_id:
+                report.status_extended = f"Opensearch domain {domain.name} is in a VPC, then it is not publicly accessible."
```
MrCloudSec, line 22: Please, check if the domain is in a VPC public subnet or in a private one.

prowler/providers/aws/services/opensearch/opensearch_service.py
puchy22: No, the VPC endpoints for an Amazon OpenSearch Service domain must be within the same VPC. However, you can have multiple VPC endpoints within the same VPC to allow access from different subnets within that VPC.


Yet as these are not used by any check should I keep them or remove them?

prowler/providers/aws/services/opensearch/opensearch_service_domains_not_publicly_accessible/opensearch_service_domains_not_publicly_accessible.py
puchy22: I don't think it's sufficient to check whether the subnet is public or private. Even if a subnet is public, the domain won't necessarily be public without additional configurations, such as AWS PrivateLink, a VPN, or a proxy. For more details, you can refer to this discussion: [Access to AWS OpenSearch in a VPC](https://repost.aws/questions/QUMsQz3kFkSUKIIXVRmxSh2A/access-to-aws-opensearch-in-a-vpc).

Simply checking the subnet type doesn't definitively determine whether the check should be a PASS or FAIL. If we want to avoid false positives checking if really is public through this mechanism we need to perform a more comprehensive check, and I think that more depth is not worth it, because if you have configured to access is because it is not a risk.

tests/providers/aws/services/opensearch/opensearch_service_domains_not_publicly_accessible/opensearch_service_domains_not_publicly_accessible_test.py
MrCloudSec, line 1: Could you do the tests with moto? As you can see here the API calls for opensearch are implemented: https://github.com/getmoto/moto/blob/master/IMPLEMENTATION_COVERAGE.md

prowler/providers/aws/services/opensearch/opensearch_service_domains_not_publicly_accessible/opensearch_service_domains_not_publicly_accessible.py
MrCloudSec: So if I have an OpenSearch in a VPC inside a public subnet exposed to the public, it is not publicly accessible?

puchy22: Yes, I checked with real infrastructure and I was not able to connect even the subnet was public. In this [docs](https://docs.aws.amazon.com/opensearch-service/latest/developerguide/vpc.html#vpc-architecture) there are some ways to access but you have to configure some things, by default there is no access. I think with the configurations like VPN or proxy the check still being a PASS because this methods are not bad practices however they are "exposing" the domain to Internet.

tests/providers/aws/services/opensearch/opensearch_service_domains_not_publicly_accessible/opensearch_service_domains_not_publicly_accessible_test.py
puchy22: Done!

prowler/providers/aws/services/opensearch/opensearch_service_domains_not_publicly_accessible/opensearch_service_domains_not_publicly_accessible.py
MrCloudSec: Awesome, thanks for the testing!

MrCloudSec (approved):
👏🏼

PR #4593, opened 2024-07-31
fix(aws): Pass backup retention check if retention period is equal to minimum

### LINKED ISSUE #4588: documentdb_cluster_backup_enabled fails when retention time is equal to minimum
### Steps to Reproduce

Run prowler (`docker run toniblyx/prowler`) against an AWS account with a DocumentDB cluster configured to retain backups for 7 days.

### Expected behavior

`documentdb_cluster_backup_enabled` should pass when the retention period is configured to satisfy the minimum.

`documentdb_cluster_backup_enabled` check still fails when backup retention is set to the minimum value given in the extended status (7 days).

> DocumentDB Cluster [redacted] has backup enabled with retention period 7 days. Recommended to increase the backup retention period to a minimum of 7 days.

`pip install prowler`

### Environment Resource

Workstation/GitHub Actions

### OS used

`ubuntu-latest` (GitHub Actions)/Fedora 40 Workstation

### Prowler version

4.3.0

### Pip version

24.2

### Context

The status message is not aligned with the condition in the check, which requires the configured retention period to be *greater* than 7 days.
https://github.com/prowler-cloud/prowler/blob/f639dc8bf4e3b2180ec60ff79557bb702625fad1/prowler/providers/aws/services/documentdb/documentdb_cluster_backup_enabled/documentdb_cluster_backup_enabled.py#L20-L29

Note that the same config value is used by `neptune_cluster_backup_enabled`, which has the same issue.
https://github.com/prowler-cloud/prowler/blob/f639dc8bf4e3b2180ec60ff79557bb702625fad1/prowler/providers/aws/services/neptune/neptune_cluster_backup_enabled/neptune_cluster_backup_enabled.py#L18-L27

I'm happy to send a PR for this, I'm just not sure whether the status message or the condition is to be considered correct.

cetteup:
### Context

Fix #4588.

### Description

Compare `actual` and `expected` using `>=` rather than `>`.

### Checklist

- [ ] Review if the code is being covered by checks.

### License

### DIFF
--- a/prowler/providers/aws/services/documentdb/documentdb_cluster_backup_enabled/documentdb_cluster_backup_enabled.py
+++ b/prowler/providers/aws/services/documentdb/documentdb_cluster_backup_enabled/documentdb_cluster_backup_enabled.py
@@ -17,7 +17,7 @@ class documentdb_cluster_backup_enabled(Check):
             report.status_extended = (
                 f"DocumentDB Cluster {cluster.id} does not have backup enabled."
             )
-            if cluster.backup_retention_period > documentdb_client.audit_config.get(
+            if cluster.backup_retention_period >= documentdb_client.audit_config.get(
                 "minimum_backup_retention_period", 7
             ):
                 report.status = "PASS"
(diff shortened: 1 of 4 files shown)


pedrooot (approved):
Loving this! Thanks @cetteup for the fix 🔝

PR #4606, opened 2024-08-01
fix(mutelist): Fix tags match

### LINKED ISSUE #4597: Cannot mute failed checks based on AWS tags using 4.3.0
### Steps to Reproduce

1. Create tagged AWS resources (tested with EC2 instances, IAM roles)
2. Add tag-based mute list entry matching those resources
3. Run prowler with the mutelist

### Expected behavior

Findings related to the muted checks are reported as muted for the tagged resources.

Related findings are not muted, causing prowler to exit with a non-zero exit code. Additionally, no findings are muted for checks in the mutelist containing both specific resources as well as tags. Meaning not only do tag-based mutelist entries not work, they seem to break the entire mutelist entry for the relevant check.

Note: When the mutelist contains an empty array of tags for a check, the resource-based exclusions still work. Adding an empty string value also has no negative impact. But even adding one of the examples (`environment=dev`) breaks the resource-based muting.

### Environment Resource

Workstation/GitHub Actions

### OS used

`ubuntu-latest` (GitHub Actions)/Fedora 40 Workstation

### Prowler version

4.3.0

### Pip version

24.2

### Context

No warnings/errors are logged regarding our mutelist. When reverting back to 4.2.4, all findings are muted as expected using the exact same mutelist.

I could not find any obvious (to me) reason for the tags breaking while quickly looking through the mutelist code. My best guess is that this issue was somehow introduced by #4413 (just looking for changes to the mutelist added with 4.3.0).

jfagoagas:
### Context

Fix #4597

### Description

Fix how the `finding.tags` are used when matching the Mutelist.

### Checklist

### License

### DIFF
--- a/prowler/lib/mutelist/mutelist.py
+++ b/prowler/lib/mutelist/mutelist.py
@@ -8,6 +8,26 @@ from prowler.lib.mutelist.models import mutelist_schema
 
 
 class Mutelist(ABC):
+    """
+    Abstract base class for managing a mutelist.
+
+    Attributes:
+        _mutelist (dict): Dictionary containing information about muted checks for different accounts.
+        _mutelist_file_path (str): Path to the mutelist file.
+        MUTELIST_KEY (str): Key used to access the mutelist in the mutelist file.
+
+    Methods:
+        __init__: Initializes a Mutelist object.
+        mutelist: Property that returns the mutelist dictionary.
+        mutelist_file_path: Property that returns the mutelist file path.
+        is_finding_muted: Abstract method to check if a finding is muted.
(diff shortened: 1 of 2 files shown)


MrCloudSec:
Thank you, good idea 🚀 
Are you covering the case of `environment=dev|environment=prod` in the mutelist tags?

MrCloudSec (approved):
Love it ❤️

jfagoagas:
> Thank you, good idea 🚀 Are you covering the case of `environment=dev|environment=prod` in the mutelist tags?

Great catch, I've included a test for it but the case is well handled by the code.

PR #4609, opened 2024-08-01
test(awslambda): Cover possible checks with moto instead MagicMock

puchy22:
### Context

Now moto covers AWS lambda functions calls, so it is necessary to port actual MagickMock tests to moto to adjust more the tests with the reality, for check `awslambda_function_not_publicly_accessible`.

### Description

Modify public and not public tests, CanonicalUser test was not supported by moto.

### Checklist

### License


tests/providers/aws/services/awslambda/awslambda_function_not_publicly_accessible/awslambda_function_not_publicly_accessible_test.py
```diff
@@ -134,21 +176,21 @@ def test_function_not_public(self):
...
             assert result[0].resource_arn == function_arn
             assert result[0].status == "PASS"
             assert (
                 result[0].status_extended
                 == f"Lambda function {function_name} has a policy resource-based policy not public."
             )
-            assert result[0].resource_tags == []
+            assert result[0].resource_tags == [{}]
 
     def test_function_public_with_canonical(self):
```
MrCloudSec, line 191: Why you don't use moto in this test?

tests/providers/aws/services/awslambda/awslambda_function_not_publicly_accessible/awslambda_function_not_publicly_accessible_test.py
MrCloudSec, line 1: Could you create the lambda functions with tags, please?

tests/providers/aws/services/awslambda/awslambda_function_not_publicly_accessible/awslambda_function_not_publicly_accessible_test.py
puchy22: I put in PR Description, I did not find a way to mock the CanonicalUser with moto.

prowler/providers/aws/services/awslambda/awslambda_service.py
```diff
@@ -161,7 +161,7 @@ def __list_tags_for_resource__(self):
                 try:
                     regional_client = self.regional_clients[function.region]
                     response = regional_client.list_tags(Resource=function.arn)["Tags"]
-                    function.tags = [response]
```
MrCloudSec: Why are you changing this? This is breaking the standard format of the AWS tagging in Prowler (it has to be a list)

puchy22: I changed it because it did not make much sense to me. We're using a list that will not be used, the length of the list will always be 1 no matter how many tags the resource had. If it is in this way because of formatting reason I could change it back.

MrCloudSec: Yes, it is how AWS works, in most of the services it return a list of dictionaries where each dictionary has an unique tag. That is why we are forcing it to be a list in this service.

MrCloudSec: That's why we use the unroll_tags function in Prowler for the outputs:
```
>>> tags = [{"key": "name", "value": "John"}, {"key": "age", "value": "30"}]
>>> unroll_tags(tags)
{'name': 'John', 'age': '30'}
```

MrCloudSec (approved):
👏🏼 👏🏼

Issue #3905, opened 2024-04-30
Add configuration for secret related checks


pr3l14t0r:
### New feature motivation

All checks related to secret detection lead to a massive amount of false positives. This could get prevented if there would be a possibility to configure those checks, since most of them use the `detect-secrets`  package. 

Use case: 
The checks [awslambda_function_no_secrets_in_variables](https://github.com/prowler-cloud/prowler/blob/484cf6f49d558da344b2f8dd149d831b693c9c2d/prowler/providers/aws/services/awslambda/awslambda_function_no_secrets_in_variables/awslambda_function_no_secrets_in_variables.py#L12) and [ecs_task_definitions_no_environment_secrets](https://github.com/prowler-cloud/prowler/blob/484cf6f49d558da344b2f8dd149d831b693c9c2d/prowler/providers/aws/services/ecs/ecs_task_definitions_no_environment_secrets/ecs_task_definitions_no_environment_secrets.py#L12) do use the [detect-secrets](https://github.com/Yelp/detect-secrets) package.

Unfortunately there is no possibility (at least no documented one) that allows to configure the checks done with that package. The repository documents how you can [use a customized settings file](https://github.com/Yelp/detect-secrets?tab=readme-ov-file#usage-in-other-python-scripts), which can for example adjust the threshold of `Base64HighEntropyString` etc. 

Could it be possible to add the JSON like config to the prowler's [config.yaml](https://github.com/prowler-cloud/prowler/blob/master/prowler/config/config.yaml)? 

Furthermore it would be nice to have an option to filter out environment variable names. 

Example: 
In an AWS `TaskDefinition` i use a variable named `GCP_SECRET_ARN`. This contains an ARN to `AWS_SECRETS_MANAGER` service which will be used during runtime of a script to retrieve a JSON, since it is not really an option to mount a JSON with newline characters into an environment variable as secret. Thus i would like to filter out for name patterns like "`*SECRET_ARN`". 
Would this make sense to you, too? 

As documented for `detect-secrets`, one could use the following parameter:
```
  --exclude-secrets EXCLUDE_SECRETS
                        If secrets match this regex, it will be ignored.
```

This could already help me out for my use-case. 

### Solution Proposed

Add an option to the config.yaml that would apply to all scans which use `detect-secrets`. 

My pseudo example: 

```yaml
aws:
  # configure entropy threshold
  awslambda_function_no_secrets_in_variables_Base64HighEntropyString_limit: 7.0
[...]

PR #4632, opened 2024-08-05
feat(aws): Add new CodeBuild check to validate environment variables

danibarranqueroo:
### Context

To enhance security and follow best practices for handling sensitive information, we are adding a new check within the CodeBuild framework. This check ensures that sensitive environment variables (AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY) in CodeBuild projects are not stored as clear text credentials.

To facilitate this check, modifications were necessary in the CodeBuild service because the `batch_get_projects` method previously did not utilize this parameter. Additionally, I have renamed all the methods according to the Python naming convention.

In addition, I did not use Moto for testing because the way it creates a new CodeBuild project seems not to accept environment variables, or something similar, because when I debug, they are always empty. I have tested the check using the AWS CLI and it seems to be working great, so for the test file, I had use MagicMock instead of Moto.

### Description

I have added this new check called `codebuild_project_no_plaintext_credentials` with his respective test to prove that it works.

### License


prowler/providers/aws/services/codebuild/codebuild_project_no_plaintext_credentials/codebuild_project_no_plaintext_credentials.py
```diff
@@ -0,0 +1,23 @@
...
+    def execute(self):
+        findings = []
+        for project in codebuild_client.projects.values():
+            report = Check_Report_AWS(self.metadata())
+            report.region = project.region
+            report.resource_id = project.name
+            report.resource_arn = project.arn
+            report.status = "PASS"
+            report.status_extended = f"CodeBuild project {project.name} does not have sensitive environment plaintext credentials."
+            sensitive_vars = ["AWS_ACCESS_KEY_ID", "AWS_SECRET_ACCESS_KEY"]
```
jfagoagas: What do you think about using the `audit_config` to allow users to add/remove sensitive environment variables? It'd be a great addition!

https://docs.prowler.com/projects/prowler-open-source/en/latest/developer-guide/checks/#using-the-audit-config

danibarranqueroo: Good point! I've just added it.

MrCloudSec:
Please use the detect secrets library for this check like in `awslambda_function_no_secrets_in_variables`.

jfagoagas:
> Please use the detect secrets library for this check like in `awslambda_function_no_secrets_in_variables`.

But what is the point of the check? I think is not to have these sensitive environment variables, as per https://docs.aws.amazon.com/securityhub/latest/userguide/codebuild-controls.html#codebuild-2

MrCloudSec:
> > Please use the detect secrets library for this check like in `awslambda_function_no_secrets_in_variables`.
> 
> But what is the point of the check? I think is not to have these sensitive environment variables, as per https://docs.aws.amazon.com/securityhub/latest/userguide/codebuild-controls.html#codebuild-2

We don't have to limit Prowler checks to CIS or SecurityHub controls, if we detect all type of secrets, the check would be more advance and we are also covering the SecurityHub check.

It makes no sense to make the user setting in the `config.yaml` the env variables that he wants to check...

prowler/providers/aws/services/codebuild/codebuild_project_no_plaintext_credentials/codebuild_project_no_plaintext_credentials.metadata.json
```diff
@@ -0,0 +1,32 @@
+{
+  "Provider": "aws",
+  "CheckID": "codebuild_project_no_plaintext_credentials",
```
MrCloudSec:
```suggestion
  "CheckID": "codebuild_project_no_secrets_in_variables",
```

jfagoagas:
> the user setting in the `config.yaml` the env variables that he wants to check

@danibarranqueroo I apologise because I explained my point just the opposite as I wanted. What I wanted to say is that it could be helpful for our users to be able to exclude some environment variables via `audit_config` since the Mutelist does not work for this kind of checks since the resource is not the environment variable analysed, you can read more about this here https://github.com/prowler-cloud/prowler/issues/3905

danibarranqueroo:
Finally, after following all your advises, I've completed this check using Detect Secrets to find credentials/secrets in the value of the environmental variables, but only when they are stored as plaintext. Additionally, I've used Audit Config to allow users to exclude specific environment variables from the check by adding their names to the config.yaml.

Take a look at it when you can! 😄 :rocket:

prowler/providers/aws/services/codebuild/codebuild_project_no_secrets_in_variables/codebuild_project_no_secrets_in_variables.metadata.json
```diff
@@ -0,0 +1,32 @@
...
+  "Severity": "critical",
+  "ResourceType": "AwsCodeBuildProject",
+  "Description": "This check ensures that AWS CodeBuild projects do not use plaintext AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY as environment variables.",
+  "Risk": "Using plaintext AWS credentials in CodeBuild project environment variables can expose these sensitive keys, leading to unauthorized access and potential security breaches.",
+  "RelatedUrl": "https://docs.aws.amazon.com/codebuild/latest/userguide/how-to-create-project-console.html",
+  "Remediation": {
+    "Code": {
+      "CLI": "",
+      "NativeIaC": "",
+      "Other": "https://docs.aws.amazon.com/securityhub/latest/userguide/codebuild-controls.html",
```
MrCloudSec:
```suggestion
      "Other": "https://docs.aws.amazon.com/securityhub/latest/userguide/codebuild-controls.html#codebuild-2",
```

prowler/providers/aws/services/codebuild/codebuild_project_no_secrets_in_variables/codebuild_project_no_secrets_in_variables.metadata.json
```diff
@@ -0,0 +1,32 @@
...
+  "CheckType": [
+    "Security Best Practices"
+  ],
+  "ServiceName": "codebuild",
+  "SubServiceName": "",
+  "ResourceIdTemplate": "arn:partition:service:region:account-id:resource-id",
+  "Severity": "critical",
+  "ResourceType": "AwsCodeBuildProject",
+  "Description": "This check ensures that AWS CodeBuild projects do not use plaintext AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY as environment variables.",
+  "Risk": "Using plaintext AWS credentials in CodeBuild project environment variables can expose these sensitive keys, leading to unauthorized access and potential security breaches.",
```
MrCloudSec: Do this description generic, this check is not only flagging AWS credentials.

prowler/providers/aws/services/codebuild/codebuild_project_no_secrets_in_variables/codebuild_project_no_secrets_in_variables.metadata.json
```diff
@@ -0,0 +1,32 @@
+{
+  "Provider": "aws",
+  "CheckID": "codebuild_project_no_secrets_in_variables",
+  "CheckTitle": "Ensure CodeBuild projects do not contain plaintext AWS credentials",
```
MrCloudSec: Do this title generic, this check is not only flagging AWS credentials.

MrCloudSec (changes requested):
Great work @danibarranqueroo ! Please, update the metadata of the check according to the new changes.

MrCloudSec (approved):
👏🏼

PR #4633, opened 2024-08-05
fix(gcp): check next rotation time in KMS keys

### LINKED ISSUE #4617: Bug while running the kms_key_rotation_enabled
### Steps to Reproduce

Hey folks, when I run a Prowler scan on the KMS service in the GCP, one of the checks doesn't check kms_key_rotation_enabled doesn't check for all condition 

1. Go to the console and create the KMS key 
2. Under Key rotation select the Key rotation period as 90 days and for the `Starting on` field select the date which is more than 90 days 

Now run the scan using the below command 

```bash
python3 prowler.py gcp --project-ids xxxxx --check kms_key_rotation_enabled
```

### Expected behavior

According to the CIS benchmark  for this check 
Ensure outcome values for `rotationPeriod` and `nextRotationTime` satisfy the below
criteria:
- rotationPeriod is <= 129600m
- rotationPeriod is <= 7776000s
- rotationPeriod is <= 2160h
- rotationPeriod is <= 90d
and 
- nextRotationTime is <= 90days from current DATE
Taken from the CIS benchmark 
But currently prowler is only looking for only checking for [rotation-period](https://github.com/prowler-cloud/prowler/blob/master/prowler/providers/gcp/services/kms/kms_service.py#L93) and [here](https://github.com/prowler-cloud/prowler/blob/master/prowler/providers/gcp/services/kms/kms_key_rotation_enabled/kms_key_rotation_enabled.py#L18)
But actually we have to check for both the fields 

Currently its looking for only the rotationPeriod field which leads to wrong result as the key has the next-rotation-time more than 90 days marked as pass 

### Environment Resource

Local

### OS used

Mac is 

### Prowler version

Prowler 4.2.1

### Pip version

pip 20.3.4

### Context

n/A

pedrooot:
### Context

It is needed to check the parameter `nextRotationTime`

Fix #4617

### Description

`nextRotationTime` must to be checked since it should be smaller than 90 days from the current date. 

Thanks @JOSHUAJEBARAJ for the report 🥇 

### Checklist

### License


prowler/providers/gcp/services/kms/kms_key_rotation_enabled/kms_key_rotation_enabled.py
```diff
@@ -11,18 +13,31 @@ def execute(self) -> Check_Report_GCP:
...
-                        f"Key {key.name} is rotated every 90 days or less."
-                    )
+                )
+            if condition_rotation_period and condition_next_rotation_time:
+                report.status = "PASS"
+                report.status_extended = f"Key {key.name} is rotated every 90 days or less and the next rotation time is in less than 90 days."
+            else:
+                report.status = "FAIL"
+                if condition_rotation_period:
+                    report.status_extended = f"Key {key.name} is rotated every 90 days or less but the next rotation time is in more than 90 days."
```
MrCloudSec, line 36: How a key can have a rotation period of 90 days and its next rotation time in more than 90 days?

JOSHUAJEBARAJ: In the `Key Rotation` options, you need to select both `Key Rotation Period` and `Starting On`. If you choose a Starting On date that is more than 90 days in the future, the key will not be rotated for the first 90 days of the initial version.
image

MrCloudSec: Just the info that I needed, thank you @JOSHUAJEBARAJ !

MrCloudSec (approved):
Great fix 👏🏼

PR #4684, opened 2024-08-07
feat(kubernetes): add a test_connection method

MrCloudSec:
### Context

We need to be able to test the connection to each provider using one of the available authentication methods, starting from the environment variables.

### Description

Add a test_connection method for the Kubernetes provider.
Create a test_connection abstract method in the Provider class but not enforced for now not to break anything.

### Checklist

### License

### DIFF
--- a/prowler/providers/kubernetes/kubernetes_provider.py
+++ b/prowler/providers/kubernetes/kubernetes_provider.py
@@ -40,7 +40,9 @@ class KubernetesProvider(Provider):
             arguments (dict): A dictionary containing configuration arguments.
         """
         logger.info("Instantiating Kubernetes Provider ...")
-        self._session = self.setup_session(arguments.kubeconfig_file, arguments.context)
+        self._session = KubernetesProvider.test_connection(
+            arguments.kubeconfig_file, arguments.context
+        )
         if not arguments.namespace:
             logger.info("Retrieving all namespaces ...")
             self._namespaces = self.get_all_namespaces()
@@ -135,7 +137,8 @@ class KubernetesProvider(Provider):
             # "partition": "identity.partition",
         }
 
-    def setup_session(self, kubeconfig_file, input_context) -> KubernetesSession:
+    @staticmethod
+    def setup_session(kubeconfig_file, input_context) -> KubernetesSession:
         """
         Sets up the Kubernetes session.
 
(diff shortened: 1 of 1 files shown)


prowler/providers/kubernetes/kubernetes_provider.py
```diff
@@ -146,42 +149,58 @@ def setup_session(self, kubeconfig_file, input_context) -> KubernetesSession:
...
-                        if context_item["name"] == input_context:
-                            context = context_item
-                else:
-                    context = config.list_kube_config_contexts()[1]
-            return KubernetesSession(api_client=client.ApiClient(), context=context)
+                context = config.list_kube_config_contexts()[1]
+        return KubernetesSession(api_client=client.ApiClient(), context=context)
+
+    @staticmethod
+    def test_connection(kubeconfig_file, input_context) -> KubernetesSession:
```
```suggestion
    def test_connection(kubeconfig_file: str, input_context: str, raise_on_exception: bool = True) -> Connection:
```

Merge from `master` to get the `Connection` object:
```python

@dataclass
class Connection:
    is_connected: bool = False
    error: Exception = None
```

prowler/providers/kubernetes/kubernetes_provider.py
```diff
@@ -146,42 +149,58 @@ def setup_session(self, kubeconfig_file, input_context) -> KubernetesSession:
...
+        try:
+            session = KubernetesProvider.setup_session(kubeconfig_file, input_context)
+            client.CoreV1Api().list_namespace(timeout_seconds=2, _request_timeout=2)
+            return session
         except Exception as error:
             logger.critical(
                 f"{error.__class__.__name__}[{error.__traceback__.tb_lineno}]: {error}"
             )
-            sys.exit(1)
+            raise error
```
```suggestion
            if raise_on_exception:
               raise error
            return Connection(error=error)
```

prowler/providers/kubernetes/kubernetes_provider.py
```diff
@@ -146,42 +149,58 @@ def setup_session(self, kubeconfig_file, input_context) -> KubernetesSession:
...
+            kubeconfig_file (str): Path to the kubeconfig file.
+            input_context (str): Context name.
+
+        Returns:
+            KubernetesSession: A Kubernetes session.
+        """
+        try:
+            session = KubernetesProvider.setup_session(kubeconfig_file, input_context)
+            client.CoreV1Api().list_namespace(timeout_seconds=2, _request_timeout=2)
+            return session
```
```suggestion
            return Connection(is_connected=True)
```

prowler/providers/kubernetes/kubernetes_provider.py
```diff
@@ -40,7 +40,9 @@ def __init__(self, arguments: Namespace):
             arguments (dict): A dictionary containing configuration arguments.
         """
         logger.info("Instantiating Kubernetes Provider ...")
-        self._session = self.setup_session(arguments.kubeconfig_file, arguments.context)
+        self._session = KubernetesProvider.test_connection(
+            arguments.kubeconfig_file, arguments.context
+        )
```
jfagoagas: This should be a call to setup_session.

jfagoagas (changes requested):

prowler/providers/kubernetes/kubernetes_provider.py
```diff
@@ -146,42 +147,62 @@ def setup_session(self, kubeconfig_file, input_context) -> KubernetesSession:
...
-                            context = context_item
-                else:
-                    context = config.list_kube_config_contexts()[1]
-            return KubernetesSession(api_client=client.ApiClient(), context=context)
+                context = config.list_kube_config_contexts()[1]
+        return KubernetesSession(api_client=client.ApiClient(), context=context)
+
+    @staticmethod
+    def test_connection(
+        kubeconfig_file: str, input_context: str, raise_on_exception: bool = True
```
jfagoagas: We need to add default values here because as far as I know this function in K8s can be called without them.

jfagoagas (dismissed):

prowler/providers/kubernetes/kubernetes_provider.py
```diff
@@ -146,42 +147,64 @@ def setup_session(self, kubeconfig_file, input_context) -> KubernetesSession:
...
+                context = config.list_kube_config_contexts()[1]
+        return KubernetesSession(api_client=client.ApiClient(), context=context)
+
+    @staticmethod
+    def test_connection(
+        kubeconfig_file: str = "~/.kube/config",
+        input_context: str = "",
+        raise_on_exception: bool = True,
+    ) -> Connection:
+        """
```
pedrooot, line 187: Could you add docstrings following the guide from https://github.com/google/styleguide/blob/gh-pages/pyguide.md#38-comments-and-docstrings please? Thanks!

pedrooot (changes requested):

pedrooot (approved):
🚀  Thanks for this amazing feature!!

PR #4712, opened 2024-08-09
fix(aws): run Prowler as IAM Root or Federated User

MrCloudSec:
### Description

Fix ARN lib so Prowler can be run as IAM Root or Federated User.

### Checklist

### License

### DIFF
--- a/tests/providers/aws/lib/arn/arn_test.py
+++ b/tests/providers/aws/lib/arn/arn_test.py
@@ -245,6 +245,73 @@ class Test_ARN_Parsing:
                     "resource": IAM_ROLE,
                 },
             },
+            # Root user
+            {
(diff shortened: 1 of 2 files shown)


tests/providers/aws/lib/arn/arn_test.py
```diff
@@ -319,6 +386,7 @@ def test_is_valid_arn(self):
             "arn:aws:lambda:eu-west-1:123456789012:function:lambda-function"
         )
         assert is_valid_arn("arn:aws:sns:eu-west-1:123456789012:test.fifo")
+
```
MrCloudSec:
```suggestion
```

puchy22 (approved):
Love this!

PR #4725, opened 2024-08-12
feat(aws): Add new RDS check to verify that cluster minor version upgrade is enabled

danibarranqueroo:
### Context

This new check assesses whether automatic minor version upgrades are enabled for Amazon RDS Multi-AZ DB clusters. Enabling automatic upgrades ensures that the database clusters are promptly updated with the latest minor versions, which may include new features, bug fixes, security patches, and performance improvements. 

This check is already done for instances but it should be done for both because enabling this on clusters only verify that new instances created will have by default this option enabled, if an instance belongs to a cluster with minor upgrades enabled but it has minor upgrades disabled it won’t be updated so both checks are needed.

Also, Moto doesn’t support the parameter MultiAZ when creating a new cluster so I’ll be using Magic Mock for the test unit of the check.

### Description

I added `rds_cluster_minor_version_upgrade_enabled` with his respective unit test.

### License


tests/providers/aws/services/rds/rds_cluster_minor_version_upgrade_enabled/rds_cluster_minor_version_upgrade_enabled_test.py
```diff
@@ -0,0 +1,201 @@
...
+                    "DBEngineVersionDescription": "description",
+                },
+            ]
+        }
+    return make_api_call(self, operation_name, kwarg)
+
+
+@mock.patch("botocore.client.BaseClient._make_api_call", new=mock_make_api_call)
+class Test_rds_cluster_minor_version_upgrade_enabled:
+    @mock_aws
```
MrCloudSec: Why are you using moto in the tests?

MrCloudSec (approved):
👏🏼👏🏼👏🏼

PR #4730, opened 2024-08-13
feat(aws): Split the checks that mix RDS Instances and Clusters

danibarranqueroo:
### Context

There were two checks labeled as instance checks that were scanning both instances and clusters. That's why I have separated these into different checks to enhance the scanning process.

These two checks have something in common: they are "heritable." If a cluster has a certain attribute, then the instances that belong to it will have the same attribute. That's why in the instance checks, we only scan instances that are not part of a cluster.

Additionally, in the IAM authentication check, some engines do not support this type of authentication, so I have decided not to include those in the scan.

### Description

I added `rds_cluster_iam_authentication`, `rds_cluster_default_admin` both with its respective unit tests and modified `rds_instance_iam_authentication` and `rds_instance_default_admin`

### License


MrCloudSec (approved):
Love this!

PR #4732, opened 2024-08-13
fix(acm): Change check logic to scan only in use certificates

MarioRgzLpz:
### Context

We noticed that the ACM check `acm_certificates_transparency_logs_enabled` is scanning unused certificates too something that shouldn’t happen. In order to fix this we have to modify the check so it doesn't scan unused certificates unless the user wants to do it. Adding a test to the check to make sure it is working properly.

### Description

Change `acm_certificates_transparency_logs_enabled` and added respective unit test .

### License


MrCloudSec (approved):
Thank you!

PR #4738, opened 2024-08-14
feat(aws): Add new RDS check for deletion protection enabled on clusters

danibarranqueroo:
### Context

This new check will verify that deletion protection is enabled for Amazon RDS DB clusters. Enabling deletion protection provides a critical safeguard against accidental or unauthorized deletions, ensuring that the clusters cannot be removed without explicitly disabling this setting first.

When you try to create a new RDS database, there is an option in the additional configuration that allows you to enable deletion protection with the following description: "Protects the database from being deleted accidentally. While this option is enabled, you can’t delete the database." This check verifies that every DB cluster has this option enabled.

### Description

I added `rds_cluster_deletion_protection` with its respective unit test.

### License

### DIFF
--- a/prowler/providers/aws/services/rds/rds_cluster_deletion_protection/rds_cluster_deletion_protection.metadata.json
+++ b/prowler/providers/aws/services/rds/rds_cluster_deletion_protection/rds_cluster_deletion_protection.metadata.json
@@ -13,7 +13,7 @@
   "RelatedUrl": "https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_DeleteInstance.html",
   "Remediation": {
     "Code": {
-      "CLI": "aws rds modify-db-instance --db-instance-identifier <db_instance_id> --deletion-protection --apply-immediately",
+      "CLI": "aws rds modify-db-cluster --db-cluster-identifier <db_cluster_id> --deletion-protection --apply-immediately",
       "NativeIaC": "",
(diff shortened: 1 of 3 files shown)


prowler/providers/aws/services/rds/rds_cluster_deletion_protection/rds_cluster_deletion_protection.metadata.json
```diff
@@ -0,0 +1,30 @@
...
+  "SubServiceName": "",
+  "ResourceIdTemplate": "arn:aws:rds:region:account-id:db-cluster",
+  "Severity": "low",
+  "ResourceType": "AwsRdsDbCluster",
+  "Description": "Check if RDS clusters have deletion protection enabled.",
+  "Risk": "You can only delete clusters that do not have deletion protection enabled.",
+  "RelatedUrl": "https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_DeleteInstance.html",
+  "Remediation": {
+    "Code": {
+      "CLI": "aws rds modify-db-instance --db-instance-identifier <db_instance_id> --deletion-protection --apply-immediately",
```
MrCloudSec: Please, change the CLI command for the cluster one.

MrCloudSec (changes requested):
I think we should remove the cluster logic from the check `rds_instance_deletion_protection`, what do you think?

danibarranqueroo:
Yes, good point. We can do it the same way we did in #4730

MrCloudSec (approved):
👏🏼

PR #4746, opened 2024-08-14
fix(ecr): handle non-existing findingSeverityCounts key

MrCloudSec:
### Description

Handle non-existing `findingSeverityCounts` key when an image has no findings and getting the following error:

`KeyError[207]: 'findingSeverityCounts'`
### Checklist

### License


drewkerrigan (approved):
Thank you!

puchy22 (approved):
Great catch 💯

Issue #4714, opened 2024-08-09
iam_role_cross_service_confused_deputy_prevention remediation broke specific process


migs017:
### Steps to Reproduce

1. It isn't a command but a setup where cloudtrail will send logs to cloudwatch
2. AWS environment
3. Single account
4. Error
![iam role last activity date](https://github.com/user-attachments/assets/eb0d54ab-ec4f-4989-ae4e-014217839232)

### Expected behavior

Prowler recommends to remediate/prevent confused deputy its either use aws:SourceArn or aws:SourceAccount or both. If the specfici resource has been added in the condition nothing should break our process.

### Actual Result with Screenshots or Logs

![iam role last activity date](https://github.com/user-attachments/assets/bc5b5ee8-78c6-48ec-a7e6-e775120417b5)
Prowlers solution works on some roles but for cloudtrail to cloudwatch process that a role will handle it breaks. We also encounter in IAM role that's assumed by aws transcoder when we add either aws:SourceArn or aws:SourceAccount or both. Our process will stop working.

The exact trust relationship policy for the transcoder role:
{
    "Version": "2008-10-17",
    "Statement": [
        {
            "Sid": "1",
            "Effect": "Allow",
            "Principal": {
                "Service": [
                    "elastictranscoder.amazonaws.com",
                    "transcribe.amazonaws.com"
                ]
            },
            "Action": "sts:AssumeRole",
            "Condition": {
                "StringEquals": {
                    "aws:SourceAccount": "<account_id>"
                },
                "ArnLike": {
                    "aws:SourceArn": [
                        "arn:aws:elastictranscoder:<region>:<account_id>:pipeline/example_pipeline",
                        "arn:aws:elastictranscoder:<region>:<account_id>:job/*",
                        "arn:aws:elastictranscoder:<region>:<account_id>:preset/*"
                    ]
                }
            }
        }
    ]
}

### How did you install Prowler?

From pip package (pip install prowler)

### Environment Resource

Workstation

### OS used

Windows

### Prowler version

4.2.4

### Pip version

pip 23.2.1

### Context

We consulted the cloudtrail to cloudwatch process role to AWS Support and they mention "no documentation around which services do or do not support the aws:SourceArn or aws:SourceAccount condition keys because they're global condition keys and technically available to all services. The keys are supported in any situation where a service tries to access another service's resource with a call from their service principal. If it's not from a service principal, we don't expect those condition keys to be set."

MrCloudSec:
[...]

PR #4759, opened 2024-08-16
chore(iam): add exception for public policy in EKS service

### LINKED ISSUE #4745: Confused Deputy Attack finding for AWS EKS service
### Steps to Reproduce

Run scan, inspect finding.

### Expected behavior

I want for prowler to detect defects in configuration, and it does an excellent job on this front.  However, I'm getting a "confused deputy" finding on our AWS EKS cluster role.  The remediation guidance suggests following AWS guidance, to include "aws:SourceAccount" and/or "aws:SourceArn" condition keys.  This was not successful.  Opening a ticket with AWS produced language from Omar M.:

> You mentioned that you were following the AWS recommendations by adding the “aws:SourceAccount” and “aws:SourceArn” condition keys and that you were still getting the error. I was able to reproduce this similar issue from my end as well and I have check and verified with our internal team that as of now EKS services does not support these conditions to prevent cross-account confused deputy attacks. Also, currently there are no other condition that you can use. It seems that there is currently no way to get around this finding for your use case (with EKS).

> I reached out to our service team to provide this feedback but as a support engineer we do not have any ETA when will EKS provide support to these conditions. I was able to find an active feature request to add support for these conditions. I have gone ahead and have added a +1 and have added your case to the request in support of it. Unfortunately, I do not have any ETA for when/if this feature will be released. However, I do recommend keeping an eye on the AWS What’s New page [2] and AWS News Blog [3] for information on new feature releases.

I think it would be helpful to add an explanation to the finding guidance that it is not possible (according to AWS) to clear the finding using the AWS guidance.

Screenshot from 2024-08-14 15-17-46

### Environment Resource

EC2 instance

### OS used

RHEL 9

### Prowler version

Prowler 4.3.1 (latest is 4.3.3, upgrade for the latest features)

### Pip version

pip 21.2.3 from /usr/lib/python3.9/site-packages/pip (python 3.9)

### Context

_No response_

puchy22:
### Context

Fix #4745

### Description

Update function `is_policy_public` adding exception for EKS service.

### Checklist

### License


MrCloudSec (changes requested):
I think we should improve the logic of the check to avoid false negatives, like in #4714 too.

MrCloudSec (approved):

PR #4785, opened 2024-08-19
fix(ec2): Manage `UnicodeDecodeError` when reading user data

puchy22:
### Context

Manage error when decoding invalid UTF-8 user data:

`ec2_instance_secrets_user_data -- UnicodeDecodeError[32]: 'utf-8' codec can't decode byte 0xdf in position 0: invalid continuation byte`

### Description

- [x] Manage the error and log it with warning level instead an error
- [x] Add tests to cover this case

### Checklist

### License


MrCloudSec (approved):
great fix 👏🏼

PR #4786, opened 2024-08-19
fix(mutelist): change logic for tags in aws mutelist

### LINKED ISSUE #4782: Conflicting documentation for mutelist tags on AWS
### Steps to Reproduce

The mutelist tags specification on AWS appears to be applied as an OR of the tags. This is consistent with the documentation at <https://docs.prowler.com/projects/prowler-open-source/en/latest/tutorials/mutelist/#how-the-mutelist-works>:

> The Mutelist uses an "ANDed" and "ORed" logic to determine which resources, checks, regions, and tags should be muted. For each check, the Mutelist checks if the account, region, and resource match the specified criteria, using an "ANDed" logic. If tags are specified, the mutelist uses and "ORed" logic to see if at least one tag is present in the resource.

But then below on the yaml example:

```yaml
### Tags is an optional list that matches on tuples of 'key=value' and are "ANDed" together.
### Use an alternation Regex to match one of multiple tags with "ORed" logic.
### For each check you can except Accounts, Regions, Resources and/or Tags.
```

And then proceeds to use an example where it assumes an AND:

```yaml
Mutelist:
  Accounts:
    "123456789012":
        ...
        "*":
          Regions:
            - "*"
          Resources:
            - "test"
          Tags:
            - "test=test"         # Will mute every resource containing the string "test" and the tags 'test=test' and
            - "project=test|project=stage" # either of ('project=test' OR project=stage) in account 123456789012 and every region
```

In my tests with the latest version 4.3.3, passing a tag multiple times with different values is applying an OR, as I expect, e.g.:

```yaml
# Mutes anything with a `Type=type1`, or with `Name=host1`
Tags:
  - "Type=type1"
  - "Name=host1"
```

A related question in this case is: what is the syntax for applying an AND of tags? Exactly like the yaml example in the documentation.

### Expected behavior

Consistent documentation

See <https://docs.prowler.com/projects/prowler-open-source/en/latest/tutorials/mutelist/#how-the-mutelist-works>

### Environment Resource

Docker container and pip install locally

### OS used

macOS and Amazon Linux 2023

### Prowler version

4.3.3

### Pip version

23.2.1

### Context

_No response_

pedrooot:
### Context

Mutelist allow the user to mute certain findings.

Fix #4782

### Description

To mute a finding, all the tags should be in the mutelist. In this PR this logic is applied and docs are updated.

Thanks @dlouzan for the report!

### Checklist

### License

### DIFF
--- a/docs/tutorials/mutelist.md
+++ b/docs/tutorials/mutelist.md
@@ -9,10 +9,6 @@ Mutelist option works along with other options and will modify the output in the
 
 ## How the Mutelist Works
 
-Here’s the improved version in English:
-
----
-
 The **Mutelist** uses both "AND" and "OR" logic to determine which resources, checks, regions, and tags should be muted. For each check, the Mutelist evaluates whether the account, region, and resource match the specified criteria using "AND" logic. If tags are specified, the Mutelist can apply either "AND" or "OR" logic:
 
 - **"AND" logic:** The Mutelist will mute all findings whose resource tags contain all the specified tags. Here’s an example:
@@ -46,9 +42,9 @@ The **Mutelist** uses both "AND" and "OR" logic to determine which resources, ch
             Resources:
               - "test"
             Tags:
-              - "test=test | project=test"
+              - "test=test|project=(test|dev)"
   ```
-  This will mute all findings that contain EITHER the `test=test` tag OR the `project=test` tag.
+  This will mute all findings that contain EITHER the `test=test` OR `project=test` OR `project=dev`.
 
 To use both logics simultaneously, here’s an example:
 
@@ -67,6 +63,9 @@ To use both logics simultaneously, here’s an example:
               - "project=test|project=stage"
   ```
 This will mute every resource containing the string "test" and the tags `test=test` and either `project=test` OR `project=stage` in every account and region.
+
+???+ note
+    Remember that mutelist can be used with regular expressions
 
 If any of the criteria do not match, the check is not muted.
 
--- a/docs/tutorials/mutelist.md
+++ b/docs/tutorials/mutelist.md
@@ -9,7 +9,7 @@ Mutelist option works along with other options and will modify the output in the
 
 ## How the Mutelist Works
 
-The Mutelist uses an "ANDed" and "ORed" logic to determine which resources, checks, regions, and tags should be muted. For each check, the Mutelist checks if the account, region, and resource match the specified criteria, using an "ANDed" logic. If tags are specified, the mutelist uses and "ORed" logic to see if at least one tag is present in the resource.
+The Mutelist uses an "ANDed" and "ORed" logic to determine which resources, checks, regions, and tags should be muted. For each check, the Mutelist checks if the account, region, and resource match the specified criteria, using an "ANDed" logic. If tags are specified, the mutelist uses an "ANDed" logic to see if all of the tags are present in the resource.
 
 If any of the criteria do not match, the check is not muted.
 
--- a/docs/tutorials/mutelist.md
+++ b/docs/tutorials/mutelist.md
@@ -65,7 +65,7 @@ To use both logics simultaneously, here’s an example:
 This will mute every resource containing the string "test" and the tags `test=test` and either `project=test` OR `project=stage` in every account and region.
 
 ???+ note
-    Remember that mutelist can be used with regular expressions
+    Remember that mutelist can be used with regular expressions.
 
 If any of the criteria do not match, the check is not muted.
 
--- a/docs/tutorials/mutelist.md
+++ b/docs/tutorials/mutelist.md
@@ -9,65 +9,12 @@ Mutelist option works along with other options and will modify the output in the
 
 ## How the Mutelist Works
 
-The **Mutelist** uses both "AND" and "OR" logic to determine which resources, checks, regions, and tags should be muted. For each check, the Mutelist evaluates whether the account, region, and resource match the specified criteria using "AND" logic. If tags are specified, the Mutelist can apply either "AND" or "OR" logic:
+The **Mutelist** uses both "AND" and "OR" logic to determine which resources, checks, regions, and tags should be muted. For each check, the Mutelist evaluates whether the account, region, and resource match the specified criteria using "AND" logic. If tags are specified, the Mutelist can apply either "AND" or "OR" logic.
 
-- **"AND" logic:** The Mutelist will mute all findings whose resource tags contain all the specified tags. Here’s an example:
-
-  ```yaml
-  Mutelist:
-    Accounts:
-      "*":
-        Checks:
-          "*":
-            Regions:
-              - "*"
-            Resources:
-              - "test"
-            Tags:
-              - "test=test"
-              - "project=test"
-  ```
-  This will mute all findings that contain BOTH tags at the same time.
-
-- **"OR" logic:** The Mutelist will mute findings that contain any of the specified resource tags. The `|` symbol is used to apply this logic:
-
-  ```yaml
-  Mutelist:
-    Accounts:
-      "*":
-        Checks:
-          "*":
-            Regions:
-              - "*"
-            Resources:
-              - "test"
-            Tags:
-              - "test=test|project=(test|dev)"
-  ```
-  This will mute all findings that contain EITHER the `test=test` OR `project=test` OR `project=dev`.
-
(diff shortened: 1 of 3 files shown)


docs/tutorials/mutelist.md
```diff
@@ -9,7 +9,7 @@ Mutelist option works along with other options and will modify the output in the
 
 ## How the Mutelist Works
 
-The Mutelist uses an "ANDed" and "ORed" logic to determine which resources, checks, regions, and tags should be muted. For each check, the Mutelist checks if the account, region, and resource match the specified criteria, using an "ANDed" logic. If tags are specified, the mutelist uses and "ORed" logic to see if at least one tag is present in the resource.
+The Mutelist uses an "ANDed" and "ORed" logic to determine which resources, checks, regions, and tags should be muted. For each check, the Mutelist checks if the account, region, and resource match the specified criteria, using an "ANDed" logic. If tags are specified, the mutelist uses an "ANDed" logic to see if all of the tags are present in the resource.
```
dlouzan: I'd say that this is still a bit confusing, as it mentions first to be using both AND or OR, but then later only mentions using AND. You guys could probably add some of the explanation of https://github.com/prowler-cloud/prowler/issues/4782#issuecomment-2296766981 in here to make clear how to use both use cases 🙇

pedrooot: @dlouzan I've just updated the docs, tell me your thoughts!

dlouzan: @pedrooot Now it's much clearer, but this has led to some more questions 😃 

Without having delved too much into the codebase, I see that this part is doing a `re.match` on the tags:

- What is the syntax for multiple tags that is obtained in the input `finding_items` when the function is called? Is that a comma-separated list of tags in a string?
- Is the code doing any kind of cleanup/filtering of the passed string from the mutelist rules for tags? or it goes as-is from the yaml user input?
- If it is not, and it is passed directly to `re.match`, can't users do here a lot of magic (intendedly or unintendedly 😅). E.g. would your OR sample also match with `project=(test|stage)`?
- If that's the case, it might be worth noting in the docs the fact that this is actually an interpreted regex (with characters that might need escaping, interpreted quantifiers, etc).
- Some other cases where I'm wondering what happens is substrings, do we need anchors here? E.g. I have a rule for `Name=myhost` but the actual resource has `Name=myhost.mydomain`, I think this will match the rule in your example? But that would also depend on the syntax that is passed on `finding_items`.

prowler/lib/mutelist/mutelist.py
```diff
@@ -211,7 +211,13 @@ def is_muted_in_check(
...
                         muted_resources, finding_resource
                     )
-                    muted_in_tags = self.is_item_matched(muted_tags, finding_tags)
+                    muted_in_tags = self.is_item_matched(
+                        muted_tags, finding_tags, tag=True
+                    )
+
+                    print(muted_in_tags)
+                    print(muted_tags)
+                    print(finding_tags)
```
MrCloudSec:
```suggestion
```

prowler/lib/mutelist/mutelist.py
```diff
@@ -317,12 +325,23 @@ def is_item_matched(matched_items, finding_items):
...
                     if item.startswith("*"):
                         item = ".*" + item[1:]
-                    if re.search(item, finding_items):
-                        is_item_matched = True
-                        break
+                    if tag:
+                        print(item)
+                        print(finding_items)
+                        value = re.search(item, finding_items)
+                        print(value)
```
MrCloudSec:
```suggestion
                        value = re.search(item, finding_items)
```

tests/providers/aws/lib/mutelist/aws_mutelist_test.py
```diff
@@ -1132,7 +1132,7 @@ def test_is_muted_tags(self):
         }
         mutelist = AWSMutelist(mutelist_content=mutelist_content)
 
-        assert mutelist.is_muted(
```
MrCloudSec, line 1135: Why some test changes?

docs/tutorials/mutelist.md
pedrooot: It's using re.search for this part, findings_items are a string whose format can be seen on `unroll_dict` docstring from `prowler/prowler/lib/outputs/utils.py`. 
For your second and third point, we check that the mutelist follow a defined schema (`prowler/lib/mutelist/models.py`) and yes, `project=(test|stage)` will mute `project=test` and `project=stage`. I'll add this to docs and tests to cover this cases. 
Regarding to the last point: yes, it will match the rule because the mutelist contains the substring from the resource. If you don't want this to happen remember that mutelist is using regular expressions, just modify your rule to `Name=myhost(?!\.)` to ensure that `Name=myhost.mydomain` won't be muted.

prowler/lib/mutelist/mutelist.py
pedrooot: 🤦🏼

prowler/lib/mutelist/mutelist.py
pedrooot: 🤦🏼

tests/providers/aws/lib/mutelist/aws_mutelist_test.py
pedrooot: This test was checking for a bad behaviour, in this case we have two tags on the mutelist so passing only one on the resource_tags fiel should not be muted.

MrCloudSec (approved):
Great fix 👏🏼

docs/tutorials/mutelist.md
dlouzan: @pedrooot Thanks for the clarifications, appreciated. AFAIU the substring case might need some more thinking depending on the case, e.g. a mute rule with `Name=my_host` would also match a resource with `Name=my_host_is_not_really_this`. We probably would need something like `Name=my_host\\b`, but honestly I have not really thought this through and I'm starting my holidays today 😁 

The fact that the regex depends on matching the pipechar-separated string from `unroll_dict` might make this a bit weird. Perhaps it'd be best if the match was done iterating over a list for each tag of the resource individually, that would make possible to anchor searches or provide options to specify whether to do regex or substring searches. But here I'm just speculating as I do not know the codebase and there might be good reasons for the current behaviour.

prowler/lib/mutelist/mutelist.py
```diff
@@ -317,12 +321,19 @@ def is_item_matched(matched_items, finding_items):
         try:
             is_item_matched = False
             if matched_items and (finding_items or finding_items == ""):
+                if tag:
+                    is_item_matched = True
```
jfagoagas, line 325: What is the point of this?

prowler/lib/mutelist/mutelist.py
```diff
@@ -303,7 +307,7 @@ def is_excepted(
             return False
 
     @staticmethod
-    def is_item_matched(matched_items, finding_items):
+    def is_item_matched(matched_items, finding_items, tag=False) -> bool:
```
jfagoagas, line 310: Can you update the method documentation?

prowler/lib/mutelist/mutelist.py
```diff
@@ -317,12 +321,19 @@ def is_item_matched(matched_items, finding_items):
...
                 for item in matched_items:
                     if item.startswith("*"):
                         item = ".*" + item[1:]
-                    if re.search(item, finding_items):
-                        is_item_matched = True
-                        break
+                    if tag:
+                        if not re.search(item, finding_items):
+                            is_item_matched = False
+                            break
```
jfagoagas, line 332: Why do you need this? Out of curiosity

prowler/lib/mutelist/mutelist.py
pedrooot: To apply the AND logic, I'll explain it in the comment below.

prowler/lib/mutelist/mutelist.py
pedrooot: Sure

prowler/lib/mutelist/mutelist.py
pedrooot: If I find a tag that it's not inside finding_items, it won't match the item because of AND logic.

jfagoagas: Understood, thanks for clarifying it!

PR #4821, opened 2024-08-21
fix(aws): enhance resource arn filtering

MrCloudSec:
### Description

The flag `--resource-arn` was not working because we were handling the audited regions as a list and not as a set and getting the error:

`2024-08-21 13:55:44,520 [File: aws_provider.py:761]     [Module: aws_provider]   CRITICAL: TypeError[758]: 'set' object is not subscriptable`

Also, the regex pattern of the ARN was not valid for the ARN of CloudWatch log groups, e.g. `arn:aws:logs:eu-west-1:123456789012:log-group:/ecs/test:*`

### Checklist

### License

### DIFF
--- a/prowler/providers/aws/aws_provider.py
+++ b/prowler/providers/aws/aws_provider.py
@@ -124,7 +124,7 @@ class AwsProvider(Provider):
         self._identity = self.set_identity(
             caller_identity=caller_identity,
             input_profile=input_profile,
-            input_regions=input_regions,
+            input_regions=set(input_regions),
             profile_region=profile_region,
         )
         ########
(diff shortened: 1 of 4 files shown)


prowler/providers/aws/aws_provider.py
```diff
@@ -124,7 +124,7 @@ def __init__(self, arguments: Namespace):
         self._identity = self.set_identity(
             caller_identity=caller_identity,
             input_profile=input_profile,
-            input_regions=input_regions,
+            input_regions=set(input_regions),
```
jfagoagas: Why not to make it a `set` for the whole `init` logic?

jfagoagas (commented):
There is also a failing test.

jfagoagas (approved):

PR #4840, opened 2024-08-22
fix(action): remove cache in setup python action

MrCloudSec:
### Description

Remove cache in setup python action to avoid false errors in releasing Prowler to PyPi:
`Error: Cache folder path is retrieved for poetry but doesn't exist on disk: /home/runner/.cache/pypoetry/virtualenvs`

### Checklist

### License


jfagoagas (changes requested):
Why this? We need to use the Python cache to speed up the execution.

jfagoagas:
Closing this since we cannot remove the cache.

PR #4845, opened 2024-08-22
fix(aws): handle AWS key-only tags

MrCloudSec:
### Description
Handle AWS Tags that only have keys and non-existing values so no KeyError can raise.
 ```
>>> tags = [{"Key": "name"}]
>>> unroll_tags(tags)
{'name': ''}
```

### Checklist

### License


jfagoagas (commented):
Several questions:
- What is the origin of this issue?
- Does it need to be backported to v4.3 and v3?

jfagoagas (approved):
Great catch!

PR #4851, opened 2024-08-23
chore(aws): Change RDS instance type from list to dict

danibarranqueroo:
### Context

For a new check that will be done soon, it's necessary to iterate over two different services. Using a dictionary type for this is more efficient than using a list.
As a result of this change, the rest of the RDS instances checks that currently iterate over a list will need to be modified to ensure they continue functioning properly.

### Description

Modified all the checks related to iterating over instances and updated the necessary tests accordingly.

### Checklist

### License


MrCloudSec (approved):
Love this!

PR #4852, opened 2024-08-23
feat(ec2): Ensure EC2 launch templates do not assign public IPs 

MarioRgzLpz:
### Context

This check verifies whether Amazon EC2 launch templates are configured to assign public IP addresses to network interfaces upon launch. The check fails if an EC2 launch template is set to assign a public IP address to network interfaces or if any network interface has a public IP address.

### Description

Added `ec2_launch_template_no_public_ip` check with metadata and respective unit tests.

### Checklist

### License


prowler/providers/aws/services/ec2/ec2_launch_template_no_public_ip/ec2_launch_template_no_public_ip.py
```diff
@@ -0,0 +1,33 @@
...
+            for version in template.versions:
+                # Check if the launch template version assigns a public IP address
+                if version.template_data.associate_public_ip_address:
+                    versions_with_public_ip.append(str(version.version_number))
+
+            if len(versions_with_public_ip) > 0:
+                report.status = "FAIL"
+                report.status_extended = (
+                    f"EC2 Launch Template {template.name} in template versions: "
+                    f"{', '.join(versions_with_public_ip)} is configured to assign a public IP address."
```
MrCloudSec:
```suggestion
                    f"EC2 Launch Template {template.name} is configured to assign a public IP address to network interfaces upon launch in template versions: f"{', '.join(versions_with_public_ip)}."
```

prowler/providers/aws/services/ec2/ec2_launch_template_no_public_ip/ec2_launch_template_no_public_ip.py
```diff
@@ -0,0 +1,33 @@
...
+                )
+            else:
+                report.status = "PASS"
+                report.status_extended = f"No versions of EC2 Launch Template {template.name} are configured to assign a public IP address."
```
MrCloudSec:
```suggestion
                report.status_extended = f"EC2 Launch Template {template.name} is not configured to assign a public IP address to network interfaces upon launch."
```

MrCloudSec:
Can we also check the second part of the check?
"The control fails if an EC2 launch template is configured to assign a public IP address to network interfaces **or if there is at least one network interface that has a public IP address.**"

prowler/providers/aws/services/ec2/ec2_launch_template_no_public_ip/ec2_launch_template_no_public_ip.py
```diff
@@ -0,0 +1,54 @@
...
+                    versions_with_autoassign_public_ip.append(
+                        str(version.version_number)
+                    )
+                if version.template_data.network_interfaces:
+                    for network_interface in version.template_data.network_interfaces:
+                        if network_interface.public_ip_addresses != []:
+                            versions_with_network_interfaces_public_ip.append(
+                                str(version.version_number)
+                            )
+                            break
```
MrCloudSec, line 29: Why a break here? Don't we want to add all the versions with this issue?

prowler/providers/aws/services/ec2/ec2_launch_template_no_public_ip/ec2_launch_template_no_public_ip.py
```diff
@@ -0,0 +1,54 @@
...
+
+            for version in template.versions:
+                # Check if the launch template version assigns a public IP address
+                if version.template_data.associate_public_ip_address:
+                    versions_with_autoassign_public_ip.append(
+                        str(version.version_number)
+                    )
+                if version.template_data.network_interfaces:
+                    for network_interface in version.template_data.network_interfaces:
+                        if network_interface.public_ip_addresses != []:
```
MrCloudSec:
```suggestion
                        if network_interface.public_ip_addresses:
```

prowler/providers/aws/services/ec2/ec2_launch_template_no_public_ip/ec2_launch_template_no_public_ip.py
```diff
@@ -0,0 +1,54 @@
...
+                    extended_messages.append(
+                        f"EC2 Launch Template {template.name} is configured to assign a public IP address to network interfaces upon launch in template versions: {', '.join(versions_with_autoassign_public_ip)}."
+                    )
+
+                if versions_with_network_interfaces_public_ip:
+                    extended_messages.append(
+                        f"EC2 Launch Template {template.name} is using a network interface with public IP addresses in template versions: {', '.join(versions_with_network_interfaces_public_ip)}."
+                    )
+
+                report.status_extended = " ".join(extended_messages)
```
MrCloudSec: It is better to create 3 different status extended that using this logic to mix two of them.

prowler/providers/aws/services/ec2/ec2_launch_template_no_public_ip/ec2_launch_template_no_public_ip.py
MarioRgzLpz: That break is only to stop looking in the network interfaces for public IPs, it will still add every version that is using a network interface with a public IP. That is only to optimize because a version can have a ton of network interfaces and it could take so long if not adding that break. My first idea was to add the network interface's id but that would  require to iterate over all the network interfaces and that could cause the check to take a lot of time.

MrCloudSec (approved):

PR #4866, opened 2024-08-26
fix(rds): get the db_instances values

### LINKED ISSUE #4865: rds_instance_certificate_expiration reports failed on compliant resource
### Steps to Reproduce

When running prowler, the check `rds_instance_certificate_expiration` fails even though the affected RDS instances have up-to-date certificates.

### Expected behavior

The check should correctly detect non-expired certificates

image
image

### Environment Resource

1. On Premise K8s Cluster

### OS used

1. Ubuntu Server 22.04

### Prowler version

Prowler 4.4.0 (You are running the latest version, yay!)

### Pip version

pip 24.2 from /usr/local/lib/python3.12/site-packages/pip (python 3.12)

### Context

_No response_

pedrooot:
### Context
Fix #4865
### Description

`self.db_intances` is a `dict` so if we want to access to this `dict` values, `self.db_intances.values()` is needed.

### Checklist

### License

### DIFF
--- a/tests/providers/aws/services/rds/rds_service_test.py
+++ b/tests/providers/aws/services/rds/rds_service_test.py
@@ -213,9 +213,7 @@ class Test_RDS_Service:
 
             rds = RDS(rds_client)
             assert len(rds.db_instances) == 1
-            print(rds.db_instances)
             db_instance_arn, db_instance = next(iter(rds.db_instances.items()))
-            print(db_instance.ca_cert)
             assert db_instance.id == "db-master-1"
             assert db_instance.region == AWS_REGION_US_EAST_1
             assert len(db_instance.cert) == 1
--- a/tests/providers/aws/services/rds/rds_service_test.py
+++ b/tests/providers/aws/services/rds/rds_service_test.py
@@ -208,6 +208,8 @@ class Test_RDS_Service:
             "prowler.providers.aws.services.rds.rds_service.RDS",
             new=rds_client,
         ):
+            from prowler.providers.aws.services.rds.rds_service import RDS
+
             rds = RDS(rds_client)
             assert len(rds.db_instances) == 1
             db_instance_arn, db_instance = next(iter(rds.db_instances.items()))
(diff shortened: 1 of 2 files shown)


tests/providers/aws/services/rds/rds_service_test.py
```diff
@@ -161,31 +162,68 @@ def test__describe_db_parameters__(self):
...
-        assert db_instance.id == "db-master-1"
-        assert db_instance.region == AWS_REGION_US_EAST_1
-        for cert in db_instance.cert:
-            assert cert["ValidTill"] < datetime.now()
+        with mock.patch(
+            "prowler.providers.aws.services.rds.rds_service.RDS",
+            new=rds_client,
+        ):
+            # Test Check
+            from prowler.providers.aws.services.rds.rds_service import RDS
```
jfagoagas:
```suggestion
```

This is not needed here since it's been already imported at the beginning.

tests/providers/aws/services/rds/rds_service_test.py
```diff
@@ -161,31 +162,68 @@ def test__describe_db_parameters__(self):
...
-        # RDS client for this test class
-        aws_provider = set_mocked_aws_provider([AWS_REGION_US_EAST_1])
-        rds = RDS(aws_provider)
-        assert len(rds.db_instances) == 1
-        db_instance_arn, db_instance = next(iter(rds.db_instances.items()))
-        assert db_instance.id == "db-master-1"
-        assert db_instance.region == AWS_REGION_US_EAST_1
-        for cert in db_instance.cert:
-            assert cert["ValidTill"] < datetime.now()
+        with mock.patch(
```
jfagoagas, line 207: Is there any real value on this tests? Because you are creating an object which is later patched and then all the asserts are performed agains it.

tests/providers/aws/services/rds/rds_service_test.py
pedrooot: True! Thanks

jfagoagas (dismissed):
Great catch! Thanks for the fix

jfagoagas (approved):

PR #4871, opened 2024-08-27
feat(azure): add custom exception class

pedrooot:
### Description

Add custom exceptions for azure provider:
- `AzureEnvironmentVariableError`
- `AzureNoSubscriptionsError`
- `AzureSetUpIdentityError`
- `AzureNoAuthenticationMethodError`
- `AzureBrowserAuthNoTenantIDError`
- `AzureTenantIDNoBrowserAuthError`
- `AzureArgumentTypeValidationError`
- `AzureSetUpRegionConfigError`
- `AzureDefaultAzureCredentialError`
- `AzureInteractiveBrowserCredentialError`
- `AzureHTTPResponseError`

### Checklist

### License


prowler/providers/azure/azure_provider.py
```diff
@@ -244,23 +258,25 @@ def validate_arguments(
             SystemExit: If browser authentication is enabled but the tenant ID is not provided.
             SystemExit: If tenant ID is provided but browser authentication is not enabled.
         """
-        if (
+        if not browser_auth and tenant_id:
+            raise AzureTenantIDNoBrowserAuthError(
+                file=os.path.basename(__file__),
+                original_exception="Azure Tenant ID (--tenant-id) is required only for browser authentication mode",
```
jfagoagas: Why this? We should keep that exception, it is a validation one.

jfagoagas: I see the same code at line 276, please check it.

prowler/providers/azure/azure_provider.py
```diff
@@ -244,23 +258,25 @@ def validate_arguments(
...
             not az_cli_auth
             and not sp_env_auth
             and not browser_auth
             and not managed_identity_auth
         ):
-            raise SystemExit(
-                "Azure provider requires at least one authentication method set: [--az-cli-auth | --sp-env-auth | --browser-auth | --managed-identity-auth]"
+            raise AzureNoAuthenticationMethodError(
+                file=os.path.basename(__file__),
+                original_exception="Azure provider requires at least one authentication method set: [--az-cli-auth | --sp-env-auth | --browser-auth | --managed-identity-auth]",
```
jfagoagas: Since there is no original exception, that line is just the exception we generate.

This comment applies to the rest of the file.

prowler/providers/azure/azure_provider.py
```diff
@@ -423,10 +453,22 @@ def test_connection(
...
             Exception: If failed to test the connection to Azure subscription.
+            AzureNoAuthenticationMethodError: If no authentication method is found.
+            AzureBrowserAuthNoTenantIDError: If browser authentication is enabled but the tenant ID is not found.
+            AzureTenantIDNoBrowserAuthError: If tenant ID is provided but browser authentication is not enabled.
+            AzureArgumentTypeValidationError: If there is an error in the argument type validation.
+            AzureSetUpRegionConfigError: If there is an error in setting up the region configuration.
+            AzureDefaultAzureCredentialError: If there is an error in retrieving the Azure credentials.
+            AzureInteractiveBrowserCredentialError: If there is an error in retrieving the Azure credentials using browser authentication.
+            AzureHTTPResponseError: If there is an HTTP response error.
+            SystemExit: If there is an error in the system.
```
jfagoagas: Are we still raising this?

prowler/providers/azure/azure_provider.py
```diff
@@ -451,35 +493,74 @@ def test_connection(
...
             if raise_on_exception:
-                raise validation_error
-            return Connection(error=validation_error)
-
+                raise AzureHTTPResponseError(
+                    file=os.path.basename(__file__),
+                    original_exception=http_response_error,
+                )
+            return Connection(error=http_response_error)
         except SystemExit as exit_error:
```
jfagoagas: We should replace this with our exceptions, we should not raise `SystemExit` anymore.

prowler/providers/azure/exceptions/exceptions.py
```diff
@@ -0,0 +1,162 @@
...
+
+class AzureBaseException(ProwlerException):
+    """Base class for Azure Errors."""
+
+    AZURE_ERROR_CODES = {
+        (1914, "AzureEnvironmentVariableError"): {
+            "message": "Azure environment variable error",
+            "remediation": "Check the Azure environment variables and ensure they are properly set.",
+            "file": "{file}",
+            "provider": "Azure",
```
jfagoagas: Why do we need to set `"provider": "Azure"` if these are just for the Azure provider?

jfagoagas (changes requested):

prowler/providers/azure/azure_provider.py
pedrooot: Of course, it could come from `validate_arguments`, `setup_region_config` and `setup_session`

prowler/providers/azure/azure_provider.py
pedrooot: I add this in order to get more info about why we are raising the exception, I can just to delete it but we won't get info.

prowler/providers/azure/azure_provider.py
pedrooot: I just moved it above, otherwise this error will never raise. If we check all the attributes first it will raise the `AzureNoAuthenticationMethodError` because by default, all auth methods will have the False values.

pedrooot: It won't change the behaviour, I think that is better now

prowler/providers/azure/azure_provider.py
pedrooot: True, it shouldn't raises SystemExit never.

prowler/providers/azure/exceptions/exceptions.py
pedrooot: True, I will change this.

prowler/providers/azure/azure_provider.py
pedrooot: P.D: SystemExit deleted.

prowler/providers/azure/azure_provider.py
```diff
@@ -240,27 +254,30 @@ def validate_arguments(
...
+            AzureNoAuthenticationMethodError: If no authentication method is found.
+            AzureBrowserAuthNoTenantIDError: If browser authentication is enabled but the tenant ID is not found.
+            AzureTenantIDNoBrowserAuthError: If tenant ID is provided but browser authentication is not enabled.
+            AzureArgumentTypeValidationError: If there is an error in the argument type validation.
         """
-        if (
+        if not browser_auth and tenant_id:
+            raise AzureTenantIDNoBrowserAuthError(
+                file=os.path.basename(__file__),
+                message="Azure Tenant ID (--tenant-id) is required only for browser authentication mode",
```
jfagoagas:
```suggestion
                message="Azure Tenant ID (--tenant-id) is required for browser authentication mode",
```

prowler/providers/azure/azure_provider.py
```diff
@@ -289,12 +306,18 @@ def setup_region_config(region):
...
+            )
         except Exception as validation_error:
             logger.error(
                 f"{validation_error.__class__.__name__}[{validation_error.__traceback__.tb_lineno}]: {validation_error}"
             )
-            raise validation_error
+            raise AzureSetUpRegionConfigError(
+                file=os.path.basename(__file__),
+                original_exception=validation_error,
+            )
```
jfagoagas: I think we can rename `validation_error` to `error` since it is a generic exception, right?

prowler/providers/azure/azure_provider.py
```diff
@@ -503,8 +576,10 @@ def check_service_principal_creds_env_vars():
                 logger.critical(
                     f"Azure provider: Missing environment variable {env_var} needed to authenticate against Azure"
                 )
-                # TODO: add custom exception once we have the AzureException
-                raise SystemExit
+                raise AzureEnvironmentVariableError(
+                    file=os.path.basename(__file__),
+                    message=f"Missing environment variable {env_var}",
```
jfagoagas: I think we should have the same message as the logger, something like:
`Missing environment variable {env_var} required to authenticate`.

prowler/providers/azure/azure_provider.py
```diff
@@ -619,8 +694,10 @@ async def get_azure_identity():
                 logger.critical(
                     "It was not possible to retrieve any subscriptions, please check your permission assignments"
                 )
-                # TODO: add custom exception once we have the AzureException
-                raise SystemExit
+                raise AzureNoSubscriptionsError(
+                    file=os.path.basename(__file__),
+                    message="No subscriptions were found",
```
jfagoagas: The same as the above comment,
`No subscriptions were found, please check your permission assignments`

jfagoagas (changes requested):

prowler/providers/azure/exceptions/exceptions.py
```diff
@@ -9,67 +9,67 @@ class AzureBaseException(ProwlerException):
...
             "file": "{file}",
-            "provider": "Azure",
+            "provider": "{provider}",
         },
         (1922, "AzureDefaultAzureCredentialError"): {
             "message": "Error in DefaultAzureCredential",
             "remediation": "Check that all the attributes are properly set up for the DefaultAzureCredential.",
             "file": "{file}",
-            "provider": "Azure",
+            "provider": "{provider}",
```
jfagoagas: Can we remove this part since it is always the same?

prowler/providers/azure/azure_provider.py

prowler/providers/azure/azure_provider.py

prowler/providers/azure/azure_provider.py

prowler/providers/azure/azure_provider.py

prowler/providers/azure/exceptions/exceptions.py
pedrooot: Done

prowler/providers/azure/azure_provider.py
jfagoagas: But I meant to change `original_exception` to `message` since there is no original exception.

prowler/providers/azure/azure_provider.py
```diff
@@ -92,7 +106,7 @@ def __init__(
         logger.info("Checking if any credentials mode is set ...")
 
         # Validate the authentication arguments
-        self.validate_arguments(
+        self.validate_arguments_cli(
```
jfagoagas, line 109: Why this? It should follow a different path than the CLI.

jfagoagas (dismissed):

jfagoagas (approved):

PR #4888, opened 2024-08-29
feat(cloudfront): Ensure distributions use SNI to serve HTTPS requests

HugoPBrito:
### Context

This check ensures that Amazon CloudFront distributions are configured to use Server Name Indication (SNI) for serving HTTPS requests with custom SSL/TLS certificates, which is essential for secure network configurations.

### Description

The new check `cloudfront_distributions_using_sni_https_requests` verifies if CloudFront distributions use SNI when serving HTTPS requests with a custom SSL/TLS certificate. The check fails if the distribution is set to use a dedicated IP address instead of SNI.

### Checklist

### License


prowler/providers/aws/services/cloudfront/cloudfront_distributions_using_sni_https_requests/cloudfront_distributions_using_sni_https_requests.py
MrCloudSec, line 1: If the distribution does not have a certificate, that does not mean we have to alert a FAIL.

MrCloudSec (approved):
Good job! I have changed the name of the check and only alerting those distributions with certificates.

PR #4892, opened 2024-08-29
feat(core): Save memory at execution

cr0hn:
### Context

This PR has a new feature that significantly reduces the memory footprint when Prowler is running.

### Description

This PR aims to reduce Prowler's memory footprint by switching from using the regular Python list and dictionaries to using SQLite3 as a database.

In the first place, it could sound a bit weird, but it's actually a good idea to reduce Prowler's memory footprint, and the performance impact is really low.

In some tests I've run, Prowler's memory footprint went from 5GB to 400-500MB, which is a huge improvement.

In this PR not all checks of all providers are implemented, but only these ones:

- AWS

And for the following AWS services:

- CloudWatch
- S3
- EC2
- ECS
- CloudWatch Logs
- SageMaker
- Lambda
- SSM

In addition, this PR also includes a quick fix that aims to use Prowler with local AWS emulators (e.g., localstack or moto).

### License


jfagoagas:
Hello @cr0hn, this is great! We will review as soon as we can, in the meantime could you merge your branch with the `master` branch since it seems that some things changed since the point your branch was created.

Thanks!

prowler/__main__.py
```diff
@@ -302,56 +304,74 @@ def prowler():
...
-    finding_outputs = [
-        Finding.generate_output(global_provider, finding) for finding in findings
-    ]
+    finding_outputs = mklist()
 
-    generated_outputs = {"regular": [], "compliance": []}
+    for finding in findings:
+        finding_outputs.append(Finding.generate_output(global_provider, finding))
+
+    generated_outputs = GeneratedOutputs(
```
jfagoagas, line 315: I'm not sure about this because it is just a workaround for the current behaviour with the S3 integration. Maybe it'll be easier to have a flag to keep the files if needed later in the process.

prowler/lib/outputs/outputs.py
```diff
@@ -2,6 +2,34 @@
 
 from prowler.config.config import orange_color
 from prowler.lib.logger import logger
+from prowler.lib.persistence import mklist
+
+
+class GeneratedOutputs:
```
jfagoagas, line 8: Probably this could be added to the S3 class, and move the responsibility there.

prowler/lib/outputs/outputs.py
```diff
@@ -2,6 +2,34 @@
...
+
+    def __init__(self, output_bucket=None, output_bucket_no_assume: None = None):
+        self._store = True if output_bucket or output_bucket_no_assume else False
+        self.regular = mklist() if self._store else None
+        self.compliance = mklist() if self._store else None
+
+    def add_compliance(self, cmp):
```
jfagoagas:
```suggestion
    def add_compliance(self, compliance):
```

prowler/lib/outputs/outputs.py
```diff
@@ -2,6 +2,34 @@
...
+
+        if not self._store:
+            return
+
+        if cmp:
+            self.compliance.append(cmp)
+
+    def add_regular(self, rg):
```
jfagoagas:
```suggestion
    def add_regular(self, regular):
```

prowler/lib/persistence/README.md
```diff
@@ -0,0 +1,104 @@
...
+del b["key1"]   
+
+print("key1" in b)
+```
+
+> Note: The above code is NOT using the SQLite database. By default, it uses native Python data structures.
+
+### Using the non-memory based data structures
+
+When you want to use a non-memory, this is SQLite, you need to set the `PROWLER_DB_CONNECTION` environment variable to `sqlite://`.
```
jfagoagas: Note: We need to think if we want to make this the default behaviour.

prowler/lib/persistence/manager.py
```diff
@@ -0,0 +1,52 @@
...
+from .sqlite import SQLiteList, SQLiteDict
+
+default_list = SQLiteList
+
+
+def mklist() -> list:
+    """
+    Create a new list with the given name and data.
+    """
+    prowler_db_connection = os.environ.get('PROWLER_DB_CONNECTION')
```
jfagoagas: If the default value is `in-memory` we'd need to set that default.
```suggestion
    prowler_db_connection = os.environ.get('PROWLER_DB_CONNECTION', "memory://")
```

prowler/lib/persistence/manager.py
```diff
@@ -0,0 +1,52 @@
...
+
+    else:
+        raise ValueError(f"Unsupported database connection: {prowler_db_connection}")
+
+
+def mkdict() -> dict:
+    """
+    Create a new dictionary with the given name and data.
+    """
+    prowler_db_connection = os.environ.get('PROWLER_DB_CONNECTION')
```
jfagoagas: The same as above.

prowler/lib/persistence/manager.py
jfagoagas, line 1: I'm not sure about the idea of checking environment variables and creating a new connection to SQLite each time we create a list/dict.

prowler/lib/persistence/sqlite.py
```diff
@@ -0,0 +1,267 @@
+import os
+import sqlite3
+import tempfile
+
+import dill
+
+DEFAULT_CACHE_SIZE = int(os.environ.get('PROWLER_CACHE_SIZE', 2000))
+
+
+class SQLiteDict:
```
jfagoagas, line 10: I think we need to add documentation to this class and their methods.

prowler/lib/persistence/README.md
```diff
@@ -0,0 +1,104 @@
...
+
+When you want to set the SQLite cache size, you need to set the `PROWLER_DB_CACHE_SIZE` environment variable.
+
+For example:
+
+```bash
+> export PROWLER_DB_CACHE_SIZE=1000
+```
+
+The cache size is the number of items that will be stored in memory. The default value is 2000.
```
jfagoagas: Why are you referring here to items? Because it is being set as a negative value so that means the cache size in KiB in SQLite.

Is `PROWLER_DB_CACHE_SIZE` the same as `PROWLER_CACHE_SIZE`?

prowler/providers/aws/aws_provider.py
```diff
@@ -1165,6 +1165,10 @@ def create_sts_session(
                 if not aws_region.startswith("cn-")
                 else f"https://sts.{aws_region}.amazonaws.com.cn"
             )
+
+            if os.environ.get("PROWLER_LOCAL_DEBUG") == "1":
+                sts_endpoint_url = os.environ["AWS_ENDPOINT_URL"]
```
jfagoagas: I'm not sure about this, because just setting the `AWS_ENDPOINT_URL` should be enough right? That environment value is not usually set unless you want to route API calls to a different endpoint.

prowler/lib/persistence/sqlite.py
```diff
@@ -0,0 +1,267 @@
...
+        self._tmp_path = tempfile.NamedTemporaryFile(prefix='prowler-dict-')
+        self.db_name = self._tmp_path.name
+        self.conn = sqlite3.connect(self.db_name)
+        self.cache_size = cache_size or DEFAULT_CACHE_SIZE
+        self._configure_cache()
+        self._create_table()
+
+    def _configure_cache(self):
+        with self.conn:
+            self.conn.execute(f'PRAGMA cache_size = {-self.cache_size}')
```
jfagoagas: Why not to add a parameter to this query?

prowler/lib/persistence/__init__.py
```diff
@@ -0,0 +1 @@
+from .manager import *
```
jfagoagas: Is this a good practice? All the linters are flagging it and I think it is just to skip one part of the import path, right?

jfagoagas (changes requested):
Hello @cr0hn, I just did a first pass to all the new business logic to improve the memory management and I expect to do some functional tests later today to verify that everything is running as expected.

Some other colleagues are reviewing this PR too due to their importance. I left some comments/questions because some parts are not clear to me, please review it when you get a chance.

Regarding tests I didn't review it yet but I'm thinking that if we are introducing a new `end_to_end` folder, we will probably need to wrap the rest into a `unit` folder.

By the way, I fixed all the conflicts and I'm applying all the linters/formatters we have configured. Please, next time remember to follow our contribution guidelines here https://docs.prowler.com/projects/prowler-open-source/en/latest/developer-guide/introduction/#contributing-with-your-code-or-fixes-to-prowler

Thanks!

jfagoagas:
@cr0hn could you please review the errors raised from our formatters here https://github.com/prowler-cloud/prowler/actions/runs/10773270870/job/29872693660?pr=4892?

prowler/lib/persistence/__init__.py
cr0hn: It's a good practice if you use in the files __all__ meta-var

cr0hn:
I just fixed the linter errors @jfagoagas

prowler/lib/persistence/README.md
```diff
@@ -0,0 +1,104 @@
...
+> Note: The above code is NOT using the SQLite database. By default, it uses native Python data structures.
+
+### Using the non-memory based data structures
+
+When you want to use a non-memory, this is SQLite, you need to set the `PROWLER_DB_CONNECTION` environment variable to `sqlite://`.
+
+For example:
+
+```bash
+> export PROWLER_DB_CONNECTION=sqlite://
```
jfagoagas, line 79: Is there any real intention of having the SQLite connection chain in the environment variable? If not I think we can just set `sqlite` because it may lead to confusion, seems that the string needs to be completed.

prowler/lib/utils/gc.py
```diff
@@ -0,0 +1,35 @@
+import gc
+
+from contextlib import contextmanager
+
+
+@contextmanager
+def force_gc(disable_gc=False):
```
jfagoagas, line 7: What is the point of this context manager? Is it intended for testing and benchmarking?

prowler/lib/utils/gc.py
```diff
@@ -0,0 +1,35 @@
...
+    if disable_gc:
+        gc.disable()
+
+    try:
+        yield
+    finally:
+        gc.collect()
+
+    if disable_gc:
+        gc.enable()
```
jfagoagas, line 32: Would not be easier to just call explicitly the GC from `main`?

prowler/lib/utils/gc.py
cr0hn: This context manager is used for the Python Garbage collector to run instead of waiting for the next planner run

prowler/lib/utils/gc.py
cr0hn: You could do it, but I don't recommend it. 

IMO, a CM improves the readability and allows a future new feature implementation, like, temporal disabling the GC (with a high memory data load, for example)

prowler/lib/persistence/README.md
cr0hn: Imagine this scenario:

If the local run has only a few resources in AWS, the user may not need the SQLite feature (which works nice, but it is a bit slower than memory). 

Imagine this second scenario:

If you want to add a distributed DB, like MongoDB or Redis, by using the connection string, nothing will change on configuring Prowler.

The idea behind it is to prepare Prowler for future new DB (or other data stores) integrations.

prowler/lib/persistence/sqlite.py
cr0hn: This query fixes the maximum memory cache used by SQLite. The idea is to limit it

prowler/lib/persistence/manager.py
```diff
@@ -0,0 +1,52 @@
...
+    """
+    prowler_db_connection = os.environ.get('PROWLER_DB_CONNECTION', "memory://")
+    prowler_db_cache_size = os.environ.get('PROWLER_DB_CACHE_SIZE')
+
+    if not prowler_db_connection:
+        return list()
+
+    # In-memory. Default.
+    if prowler_db_connection.startswith('memory://'):
+        return list()
```
vicferpoy: This can be unified in a single if.

prowler/providers/aws/aws_provider.py
cr0hn: If you want to create a simulated AWS environment in your local (with Local Stack, for example), setting the env var `AWS_ENDPOINT_URL` is not enough. You have to modify this Python line.

If you don't use STS, it won't be necessary.

prowler/lib/persistence/manager.py
```diff
@@ -0,0 +1,52 @@
...
+    Create a new list with the given name and data.
+    """
+    prowler_db_connection = os.environ.get('PROWLER_DB_CONNECTION', "memory://")
+    prowler_db_cache_size = os.environ.get('PROWLER_DB_CACHE_SIZE')
+
+    if not prowler_db_connection:
+        return list()
+
+    # In-memory. Default.
+    if prowler_db_connection.startswith('memory://'):
```
vicferpoy: I don't like using `startswith` here, if the env variable remains as string.

prowler/lib/persistence/manager.py
```diff
@@ -0,0 +1,52 @@
...
+from .sqlite import SQLiteList, SQLiteDict
+
+default_list = SQLiteList
+
+
+def mklist() -> list:
+    """
+    Create a new list with the given name and data.
+    """
+    prowler_db_connection = os.environ.get('PROWLER_DB_CONNECTION', "memory://")
```
vicferpoy: I'm confused about the usage of this environment variable. Why does it have to be a string? Wouldn't it make more sense to be a flag to either use sqlite or not?

vicferpoy: I can only see this variable usage as a flag, unless I'm missing something. That would make its name not descriptive enough.

prowler/lib/persistence/sqlite.py
```diff
@@ -0,0 +1,267 @@
...
+
+DEFAULT_CACHE_SIZE = int(os.environ.get('PROWLER_CACHE_SIZE', 2000))
+
+
+class SQLiteDict:
+
+    def __init__(self, cache_size=2000):
+        self._tmp_path = tempfile.NamedTemporaryFile(prefix='prowler-dict-')
+        self.db_name = self._tmp_path.name
+        self.conn = sqlite3.connect(self.db_name)
```
vicferpoy, line 14: Same concerns as Pepe about creating a database and a connection every time we create a list or dict.

prowler/lib/outputs/outputs.py
cr0hn: Maybe te name of the class should be something like `S3Output`

prowler/lib/persistence/sqlite.py
```diff
@@ -0,0 +1,267 @@
...
+            for item in other:
+                new_list.append(item)
+            return new_list
+        elif isinstance(other, list):
+            new_list = SQLiteList()
+            for item in self:
+                new_list.append(item)
+            for item in other:
+                new_list.append(item)
+            return new_list
```
vicferpoy:
```suggestion
        if isinstance(other, (SQLiteList, list)):
            new_list = SQLiteList()
            new_list.extend(self)
            new_list.extend(other)
            return new_list
```

prowler/lib/persistence/sqlite.py
```diff
@@ -0,0 +1,267 @@
...
+        else:
+            raise TypeError("Unsupported operand type(s) for +: 'SQLiteList' and '{}'".format(type(other).__name__))
+
+    def __iadd__(self, other):
+        if isinstance(other, (SQLiteList, list)):
+            for item in other:
+                self.append(item)
+            return self
+        else:
+            raise TypeError("Unsupported operand type(s) for +=: 'SQLiteList' and '{}'".format(type(other).__name__))
```
vicferpoy:
```suggestion
        if isinstance(other, (SQLiteList, list)):
            self.extend(other)
            return self
        raise TypeError(f"Unsupported operand type(s) for +=: 'SQLiteList' and '{type(other).__name__}'")
```

vicferpoy (changes requested):

cr0hn (commented):
Answer comments

prowler/__main__.py
jfagoagas: We will review it soon, for now it is ok.

prowler/lib/persistence/README.md
jfagoagas: It is going to be the default behaviour.

prowler/lib/persistence/manager.py
jfagoagas: For now, we will leave it as is, looking to have a DB connection string within the environment variable.

prowler/lib/persistence/manager.py
jfagoagas: For now, we will leave it as is.

prowler/providers/aws/aws_provider.py
jfagoagas: I'll leave this commented for now since it is just intended for testing.

jfagoagas: STS is a key point of everything around the AWS provider, we will see what to do here.

prowler/lib/persistence/README.md
jfagoagas: Thanks!

prowler/lib/utils/gc.py
jfagoagas: We are not sure about this, but we will review that later on.

prowler/lib/persistence/manager.py
jfagoagas: Added a TODO

prowler/lib/persistence/README.md
cr0hn: I'm referring to the number of memory pages used by SQLite. In the example, the 1000 value will store pages in memory up to a maximum of 1000kiB

jfagoagas:

PR #4908, opened 2024-08-30
feat(gcp): add custom exceptions clas

pedrooot:
### Description

Add custom exception class for GCP provider.

### Checklist

### License

### DIFF
--- a/prowler/providers/gcp/exceptions/exceptions.py
+++ b/prowler/providers/gcp/exceptions/exceptions.py
@@ -1,0 +1,9 @@
+from prowler.exceptions.exceptions import ProwlerException
+
+
+class GCPBaseException(ProwlerException):
+    def __init__(self, message, remediation, file):
+        self.message = message
+        self.remediation = remediation
+        self.file = file
+        self.provider = "GCP"
--- a/prowler/providers/gcp/exceptions/exceptions.py
+++ b/prowler/providers/gcp/exceptions/exceptions.py
@@ -8,38 +8,26 @@ class GCPBaseException(ProwlerException):
         (1925, "GCPCloudResourceManagerAPINotUsedError"): {
             "message": "Cloud Resource Manager API not used",
             "remediation": "Enable the Cloud Resource Manager API for the project.",
-            "file": "{file}",
-            "provider": "{provider}",
         },
         (1926, "GCPHTTPError"): {
             "message": "HTTP error",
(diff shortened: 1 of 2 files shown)


jfagoagas:
Is this still WIP?

prowler/providers/gcp/exceptions/exceptions.py
```diff
@@ -0,0 +1,90 @@
...
+
+class GCPBaseException(ProwlerException):
+    """Base class for GCP Errors."""
+
+    GCP_ERROR_CODES = {
+        (1925, "GCPCloudResourceManagerAPINotUsedError"): {
+            "message": "Cloud Resource Manager API not used",
+            "remediation": "Enable the Cloud Resource Manager API for the project.",
+            "file": "{file}",
+            "provider": "GCP",
```
jfagoagas: The same as with the other providers, probably we don't need this here.

prowler/providers/gcp/gcp_provider.py
```diff
@@ -73,8 +81,10 @@ def __init__(
         accessible_projects = self.get_projects()
         if not accessible_projects:
             logger.critical("No Project IDs can be accessed via Google Credentials.")
-            # TODO: add custom exception once we have the GCP exceptions
-            raise SystemExit
+            raise GCPNoAccesibleProjectsError(
+                file=__file__,
+                original_exception="No Project IDs can be accessed via Google Credentials.",
```
jfagoagas:
```suggestion
                message="No Project IDs can be accessed via Google Credentials.",
```

Why not in this way?

jfagoagas (changes requested):

prowler/providers/gcp/gcp_provider.py
jfagoagas: This comment applies to the rest of the file with the same logic while creating exceptions.

pedrooot: Because the point of this type of exceptions is to have this exceptions is to have "static" values for all the exceptions fields except the file and the original exception. I could get your point but if at the end the original_exception will appear too on the log, why to change the message?

prowler/providers/gcp/exceptions/exceptions.py
pedrooot: Removed ✅

jfagoagas (approved):
This looks great, another step to have the Prowler SDK 🚀

PR #4910, opened 2024-08-30
chore(prowler): change all methods from services from format double underscore to single underscore

puchy22:
### Context

Now we were using incorrect format to naming methods for services. To indicate internal use for methods is better to use `_*`.

[Ref](https://peps.python.org/pep-0008/#descriptive-naming-styles)

### Description

Changes all methods from services form all providers defined in the same class. Not having in count father methods.

### Checklist

### License


pedrooot (approved):
LGTM!

PR #4912, opened 2024-09-02
feat(kubernetes): add custom exception class

pedrooot:
### Description

Add custom exception class for Kubernetes provider.

### Checklist

- [x] Review if the code is being covered by tests.

### License


prowler/providers/kubernetes/exceptions/exceptions.py
```diff
@@ -0,0 +1,73 @@
...
+    KUBERNETES_ERROR_CODES = {
+        (1925, "KubernetesCloudResourceManagerAPINotUsedError"): {
+            "message": "Cloud Resource Manager API not used",
+            "remediation": "Enable the Cloud Resource Manager API for the project.",
+            "file": "{file}",
+            "provider": "Kubernetes",
+        },
+        (1926, "KubernetesSetUpSessionError"): {
+            "message": "Error setting up session",
+            "remediation": "Check the session setup and ensure it is properly set up.",
```
jfagoagas: Can we improve this remediation?

prowler/providers/kubernetes/exceptions/exceptions.py
```diff
@@ -0,0 +1,73 @@
...
+
+class KubernetesBaseException(ProwlerException):
+    """Base class for Kubernetes errors."""
+
+    KUBERNETES_ERROR_CODES = {
+        (1925, "KubernetesCloudResourceManagerAPINotUsedError"): {
+            "message": "Cloud Resource Manager API not used",
+            "remediation": "Enable the Cloud Resource Manager API for the project.",
+            "file": "{file}",
+            "provider": "Kubernetes",
```
jfagoagas: Do we need to have always the `"provider": "Kubernetes",` key? I think we can get it from the exception class.

prowler/providers/kubernetes/exceptions/exceptions.py
jfagoagas, line 1: Please, check the remediation and errors, we need to make user's life easier and each message needs to be helpful.

prowler/providers/kubernetes/kubernetes_provider.py
```diff
@@ -245,8 +252,7 @@ def search_and_save_roles(
             logger.critical(
                 f"{error.__class__.__name__}[{error.__traceback__.tb_lineno}]: {error}"
             )
-            # TODO: add custom exception once we have the Kubernetes exceptions
-            raise error
+            raise KubernetesSearchAndSaveRolesError(original_exception=error)
```
jfagoagas: I'm not sure about this kind of exception, since in this function we are just getting and setting values in a list.

prowler/providers/kubernetes/kubernetes_provider.py
```diff
@@ -299,8 +307,7 @@ def get_all_namespaces(self) -> list[str]:
             logger.critical(
                 f"{error.__class__.__name__}[{error.__traceback__.tb_lineno}]: {error}"
             )
-            # TODO: add custom exception once we have the Kubernetes exceptions
-            raise error
+            raise KubernetesGetAllNamespacesError(original_exception=error)
```
jfagoagas: We should catch the possible Kubernetes APIs exceptions here and see how to handle them.

prowler/providers/kubernetes/kubernetes_provider.py
```diff
@@ -276,12 +282,14 @@ def get_context_user_roles(self):
             )
             logger.info("Context user roles retrieved successfully.")
             return roles
+        except KubernetesSearchAndSaveRolesError as error:
+            logger.critical(str(error))
```
jfagoagas: Why is this critical here?

jfagoagas (changes requested):

prowler/providers/kubernetes/kubernetes_provider.py
pedrooot: It's a way to get the specific exception. Do you prefer to raise just the error?

prowler/providers/kubernetes/kubernetes_provider.py
pedrooot: Because the exception from below is critical too

prowler/providers/kubernetes/exceptions/exceptions.py
pedrooot: Done

prowler/providers/kubernetes/kubernetes_provider.py
pedrooot: Done

prowler/providers/kubernetes/kubernetes_provider.py
```diff
@@ -1,8 +1,9 @@
 import os
 
 from colorama import Fore, Style
-from kubernetes.client.exceptions import ApiException, TimeoutError
+from kubernetes.client.exceptions import ApiException
```
jfagoagas, line 4: Why did you remove the `TimeoutError` from the `kubernetes` API?

prowler/providers/kubernetes/kubernetes_provider.py
jfagoagas: I think so, in this case there is nothing specifically related to K8s happening in the function.

prowler/providers/kubernetes/kubernetes_provider.py
jfagoagas: I think we are logging the errors in the above function so there is no need to duplicate the `critical` log.

prowler/providers/kubernetes/kubernetes_provider.py
```diff
@@ -295,12 +309,39 @@ def get_all_namespaces(self) -> list[str]:
             namespaces = [item.metadata.name for item in namespace_list.items]
             logger.info("All namespaces retrieved successfully.")
             return namespaces
+        except ApiException as api_error:
+            logger.critical(
+                f"ApiException[{api_error.__traceback__.tb_lineno}]: {api_error}"
+            )
+            raise KubernetesApiError(
```
```suggestion
            raise KubernetesAPIError(
```

prowler/providers/kubernetes/exceptions/exceptions.py
```diff
@@ -0,0 +1,108 @@
...
+        },
+        (1928, "KubernetesGetContextUserRolesError"): {
+            "message": "Failed to retrieve context user roles, possibly due to misconfiguration.",
+            "remediation": "Check the user roles in the current context and ensure they are correctly set up. Refer to the Kubernetes documentation for guidance: https://kubernetes.io/docs/reference/access-authn-authz/rbac/#user-facing-roles",
+        },
+        (1929, "KubernetesGetAllNamespacesError"): {
+            "message": "Error occurred while fetching all namespaces in the Kubernetes cluster.",
+            "remediation": "Verify that namespaces are correctly configured and accessible. Follow this guide to troubleshoot namespace issues: https://kubernetes.io/docs/tasks/administer-cluster/namespaces-walkthrough/",
+        },
+        (1930, "KubernetesApiError"): {
```
jfagoagas:
```suggestion
        (1930, "KubernetesAPIError"): {
```

prowler/providers/kubernetes/kubernetes_provider.py
```diff
@@ -295,12 +309,39 @@ def get_all_namespaces(self) -> list[str]:
...
+            raise KubernetesTypeError(
+                original_exception=type_error, file=os.path.abspath(__file__)
+            )
         except Exception as error:
             logger.critical(
                 f"{error.__class__.__name__}[{error.__traceback__.tb_lineno}]: {error}"
             )
-            # TODO: add custom exception once we have the Kubernetes exceptions
-            raise error
+            raise KubernetesGetAllNamespacesError(original_exception=error)
```
jfagoagas: I'm thinking that maybe it is better to create a generic exception instead one per function, something like `KubernetesBaseException` with some code to wrap all the generic ones.

jfagoagas (changes requested):

prowler/providers/kubernetes/exceptions/exceptions.py
pedrooot: Done

prowler/providers/kubernetes/kubernetes_provider.py
pedrooot: Done

prowler/providers/kubernetes/kubernetes_provider.py
pedrooot: Done

prowler/providers/kubernetes/kubernetes_provider.py
pedrooot: It was a typo, there is no TimeoutError from kubernetes.client.exceptions

prowler/providers/kubernetes/kubernetes_provider.py
pedrooot: Done

prowler/providers/kubernetes/kubernetes_provider.py
```diff
@@ -276,12 +289,13 @@ def get_context_user_roles(self):
...
             return roles
+        except KubernetesError as error:
+            raise error
         except Exception as error:
             logger.critical(
                 f"{error.__class__.__name__}[{error.__traceback__.tb_lineno}]: {error}"
             )
-            # TODO: add custom exception once we have the Kubernetes exceptions
-            raise error
+            raise KubernetesGetContextUserRolesError(original_exception=error)
```
jfagoagas: Would not be `KubernetesError` here?

jfagoagas (changes requested):
I think all K8s API calls could return `ApiException`, wouldn't we catch them in each call and encapsulate it under `KubernetesAPIException`?

prowler/providers/kubernetes/exceptions/exceptions.py
```diff
@@ -0,0 +1,82 @@
...
+    def __init__(self, code, file=None, original_exception=None, message=None):
+        super().__init__(code, file, original_exception, message)
+
+
+class KubernetesTimeoutError(KubernetesBaseException):
+    def __init__(self, code, file=None, original_exception=None, message=None):
+        super().__init__(code, file, original_exception, message)
+
+
+class KubernetesValueError(KubernetesBaseException):
```
jfagoagas: Please remove this since it is no longer needed.

prowler/providers/kubernetes/exceptions/exceptions.py
```diff
@@ -0,0 +1,82 @@
...
+    def __init__(self, code, file=None, original_exception=None, message=None):
+        super().__init__(code, file, original_exception, message)
+
+
+class KubernetesTypeError(KubernetesBaseException):
```
jfagoagas: Please remove this since it is no longer needed.

jfagoagas (changes requested):
Everything is fine apart from the two exceptions that we should remove, thanks for getting this to the final line 🏁 closer to have a real working SDK 🙌

jfagoagas (dismissed):

jfagoagas (approved):

PR #4914, opened 2024-09-02
Add ClientProxy for lazy object loading

cr0hn:
### Context

This PR adds an object proxy for the lazy loading of Python objects and some unit tests.

### Description

The idea is to have a proxy object that can be used to load Python objects lazily.

This is useful for keeping the application's memory footprint low and for tracking the objects that are being loaded.

It also includes a `alive_clients` global variable that can be used to track all the Python classes being loaded by using the proxy. This global variable is backed by the `_AliveClients` class, which also includes an example method to clean up the clients that are no longer needed.

### License

### DIFF
--- a/prowler/lib/utils/proxy.py
+++ b/prowler/lib/utils/proxy.py
@@ -1,0 +1,143 @@
+import gc
+from typing import List, Set, Dict
+
+
+class ClientProxy:
+    """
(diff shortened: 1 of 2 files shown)


jfagoagas:

cr0hn:
That's ok. It was a pleasure!

PR #4915, opened 2024-09-02
feat(secrets): improve detect secrets checks and add config

### LINKED ISSUE #3905: Add configuration for secret related checks

### LINKED ISSUE #4893: Improve `detect-secrets` engine: configuration + plugins
### Steps to Reproduce

Prowler relies on https://github.com/Yelp/detect-secrets to detect secrets for some AWS resources in some checks:

- autoscaling_find_secrets_ec2_launch_configuration
- awslambda_function_no_secrets_in_code
- awslambda_function_no_secrets_in_variables
- cloudformation_stack_outputs_find_secrets
- cloudwatch_log_group_no_secrets_in_logs
- ec2_instance_secrets_user_data
- ec2_launch_template_no_secrets
- ecs_task_definitions_no_environment_secrets
- ssm_document_secrets

Currently the detection engine is raising some false positives we need to work on to improve detection.

Some examples:
- `MYSQL_ALLOW_EMPTY_PASSWORD` is flagged as a secret when the value is just `yes` or `no`.
- `detect-secrets` TelegramBotToken plugin is flagging some strings to be a secret when it is not.

### Expected behavior

Prowler should not raise false positives regarding secrets detection.

- We need to investigate more about adding/removing plugins.
- Adjust the `base64-limit` and the `hex-limit`

N/A

### Environment Resource

N/A

### OS used

N/A

### Prowler version

4.3.5

### Pip version

24.0

### Context

_No response_

pedrooot:
### Context

Fix #4893
Fix #3905

### Description

Add `secrets` configuration to make the output more accurate. Checks modified:
- `autoscaling_find_secrets_ec2_launch_configuration `
- `awslambda_function_no_secrets_in_code `
-  `awslambda_function_no_secrets_in_variables`
-  `cloudformation_stack_outputs_find_secrets`
-  `cloudwatch_log_group_no_secrets_in_logs`
-  `ec2_instance_secrets_user_data`
-  `ec2_launch_template_no_secrets`
-  `ecs_task_definitions_no_environment_secrets`
-  `ssm_document_secrets`
- `codebuild_project_no_secrets_in_variables`

Also, `config.yaml` will have the field `secrets_ignore_patterns`, a list that can be changed to set a list of regex from your known secrets. This way, `detect-secrets` will discard the secrets that match the regex.

### Checklist

### License


prowler/lib/utils/utils.py
```diff
@@ -80,31 +81,57 @@ def hash_sha512(string: str) -> str:
...
-    temp_data_file = tempfile.NamedTemporaryFile(delete=False)
-    temp_data_file.write(bytes(data, encoding="raw_unicode_escape"))
-    temp_data_file.close()
+def is_known_non_secret(line):
+    known_non_secrets = [
+        "MYSQL_ALLOW_EMPTY_PASSWORD",
+        "GCP_SECRET_ARN",
+        "S3_BUCKET",
+        "AWS_STACK",
+    ]
```
jfagoagas: I think this list could end up with thousands of values, I'm not sure about placing it here. Do we have any other way not to flag them as secrets?

prowler/lib/utils/utils.py
```diff
@@ -80,31 +81,57 @@ def hash_sha512(string: str) -> str:
...
 
     secrets = SecretsCollection()
     with transient_settings(
         {
             "plugins_used": [
-                {"name": "Base64HighEntropyString", "limit": 4.5},
-                {"name": "HexHighEntropyString", "limit": 3.0},
+                {"name": "KeywordDetector"},
+                {"name": "Base64HighEntropyString", "limit": 4.0},
+                {"name": "HexHighEntropyString", "limit": 1.5},
```
jfagoagas: Why did you chose these values? Can you document it?

prowler/lib/utils/utils.py
```diff
@@ -80,31 +81,57 @@ def hash_sha512(string: str) -> str:
...
                 {"name": "AWSKeyDetector"},
             ],
             "filters_used": [
                 {"path": "detect_secrets.filters.common.is_invalid_file"},
-                {"path": "detect_secrets.filters.heuristic.is_likely_id_string"},
+                {"path": "detect_secrets.filters.common.is_known_non_secret"},
+                {"path": "detect_secrets.filters.common.is_known_false_positive"},
```
jfagoagas: What is the purpose of these filters?

tests/providers/aws/services/awslambda/awslambda_function_no_secrets_in_code/awslambda_function_no_secrets_in_code_test.py
```diff
@@ -107,7 +107,7 @@ def test_function_code_with_secrets(self):
             assert result[0].status == "FAIL"
             assert (
                 result[0].status_extended
-                == f"Potential secret found in Lambda function {LAMBDA_FUNCTION_NAME} code -> lambda_function.py: Secret Keyword on line 3."
+                == f"Potential secrets found in Lambda function {LAMBDA_FUNCTION_NAME} code -> lambda_function.py: Base64 High Entropy String on line 3, Secret Keyword on line 3; lambda_function.py: Base64 High Entropy String on line 3, Secret Keyword on line 3."
```
jfagoagas: Why this change? We should not flag for the same secret several times.

tests/providers/aws/services/awslambda/awslambda_function_no_secrets_in_variables/awslambda_function_no_secrets_in_variables_test.py
```diff
@@ -115,6 +115,49 @@ def test_function_secrets_in_variables(self):
...
+        function_arn = f"arn:aws:lambda:{AWS_REGION_US_EAST_1}:{AWS_ACCOUNT_NUMBER}:function/{function_name}"
+
+        lambda_client.functions = {
+            "function_name": Function(
+                name=function_name,
+                security_groups=[],
+                arn=function_arn,
+                region=AWS_REGION_US_EAST_1,
+                runtime=function_runtime,
+                environment={"TELEGRAM_BOT_TOKEN": "telegram-token"},
```
jfagoagas, line 131: You should add a valid Telegram Bot Token because detect-secrets flags it as a secret for the environment value not by their name.

tests/providers/aws/services/ec2/ec2_instance_secrets_user_data/ec2_instance_secrets_user_data_test.py
```diff
@@ -256,7 +300,7 @@ def test_one_ec2_file_with_secrets_gzip(self):
             assert result[0].status == "FAIL"
             assert (
                 result[0].status_extended
-                == f"Potential secret found in EC2 instance {instance.id} User Data -> Secret Keyword on line 1, Hex High Entropy String on line 3, Secret Keyword on line 3, Secret Keyword on line 4."
+                == f"Potential secret found in EC2 instance {instance.id} User Data -> Secret Keyword on line 1, Hex High Entropy String on line 3, Secret Keyword on line 3, Hex High Entropy String on line 4, Secret Keyword on line 4."
```
jfagoagas: The same as before, why are we adding more for the same secret?

jfagoagas (changes requested):

prowler/lib/utils/utils.py
jfagoagas: I think I'll remove that for now until we explore all the options to try not to flag them as secrets.

jfagoagas: Let's try to improve the configuration for the test that will fail once we remove this block.

jfagoagas (changes requested):

prowler/lib/utils/utils.py
```diff
@@ -81,29 +80,26 @@ def hash_sha512(string: str) -> str:
...
-    common.is_known_non_secret = is_known_non_secret
-    if not file:
-        temp_data_file = tempfile.NamedTemporaryFile(delete=False)
-        temp_data_file.write(bytes(data, encoding="raw_unicode_escape"))
-        temp_data_file.close()
-
-    secrets = SecretsCollection()
-    with transient_settings(
-        {
+def detect_secrets_scan(data=None, file=None, excluded_secrets: list[str] = None):
```
jfagoagas: Add the types for the arguments and the return value please.

tests/providers/aws/services/ec2/ec2_instance_secrets_user_data/ec2_instance_secrets_user_data_test.py
jfagoagas: Review this please.

jfagoagas (changes requested):

prowler/lib/utils/utils.py
pedrooot: Done

jfagoagas (dismissed):
What a great improvement in our way to detect-secrets!

jfagoagas (approved):

cetteup:
@pedrooot This is a great improvement! When will this be released?

PR #4935, opened 2024-09-05
feat(ec2): Amazon EC2 Instances Should Not Use Multiple ENIs

MarioRgzLpz:
### Context

Using multiple ENIs can lead to increased network security complexity and unintended network paths, particularly in instances with multiple subnets. I added a new check that helps ensure that instances are not overly complex or exposed to additional network risks.

Checks whether an Amazon EC2 instance uses multiple Elastic Network Interfaces (ENIs) or Elastic Fabric Adapters (EFAs). The control passes if the instance is using a single network adapter.

### Description

Change `ec2_service `to add a new attribute network interfaces to Instance model saving ENIs ids to check type. Added the  check logic in `ec2_instance_multiple_eni`. Added tests for service and check.

### Checklist

### License


prowler/providers/aws/services/ec2/ec2_service.py
```diff
@@ -72,6 +72,11 @@ def _describe_instances(self, regional_client):
                         if not self.audit_resources or (
                             is_resource_filtered(arn, self.audit_resources)
                         ):
+                            enis = []
+                            for eni in instance.get("NetworkInterfaces"):
```
MrCloudSec:
```suggestion
                            for eni in instance.get("NetworkInterfaces, []"):
```

prowler/providers/aws/services/ec2/ec2_instance_multiple_eni/ec2_instance_multiple_eni.py
```diff
@@ -0,0 +1,49 @@
...
+                )
+            else:
+                for eni_id in instance.network_interfaces:
+                    if ec2_client.network_interfaces[eni_id].type in eni_types:
+                        eni_types[ec2_client.network_interfaces[eni_id].type].append(
+                            eni_id
+                        )
+
+                message_status_extended = ""
+                if len(eni_types["efa"]) + len(eni_types["interface"]) > 1:
```
MrCloudSec: Why not trunk here?

prowler/providers/aws/services/ec2/ec2_instance_multiple_eni/ec2_instance_multiple_eni.metadata.json
```diff
@@ -0,0 +1,34 @@
+{
+  "Provider": "aws",
+  "CheckID": "ec2_instance_multiple_eni",
```
MrCloudSec:
```suggestion
  "CheckID": "ec2_instance_uses_single_eni",
```

tests/providers/aws/services/ec2/ec2_instance_multiple_eni/ec2_instance_multiple_eni_test.py
```diff
@@ -0,0 +1,317 @@
...
+            check = ec2_instance_multiple_eni()
+            result = check.execute()
+
+            assert len(result) == 1
+            assert result[0].status == "FAIL"
+            assert result[0].region == AWS_REGION_US_EAST_1
+            assert result[0].resource_id == "i-0123456789abcdef0"
+            assert (
+                result[0].status_extended
+                == "EC2 Instance i-0123456789abcdef0 uses multiple ENIs: "
```
MrCloudSec: Can you prettify this message please?

MarioRgzLpz: What do you mean by prettify? Should I add global attributes for the instance id and enis ids?

prowler/providers/aws/services/ec2/ec2_instance_multiple_eni/ec2_instance_multiple_eni.py
MarioRgzLpz: SecurityHub recommendation is to check only for multiple EFAs or interfaces. Thats because a trunk ENI is used to increase the number of ENIs you can attach to a instance. For more information look at the first paragraph here https://docs.aws.amazon.com/AmazonECS/latest/developerguide/container-instance-eni.html and at the AWSVPC trunking section here https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-account-settings.html. I don't know if there should be a limit to trunk ENIs maybe we have to discuss that.

prowler/providers/aws/services/ec2/ec2_instance_uses_single_eni/ec2_instance_uses_single_eni.metadata.json
```diff
@@ -0,0 +1,34 @@
+{
+  "Provider": "aws",
+  "CheckID": "ec2_instance_uses_single_eni",
+  "CheckTitle": "Amazon EC2 instances should not use multiple ENIs",
+  "CheckType": [
+    "Software and Configuration Checks/AWS Security Best Practices"
```
sergargar, line 6:
```suggestion
    "Software and Configuration Checks", "AWS Security Best Practices"
```

tests/providers/aws/services/ec2/ec2_instance_multiple_eni/ec2_instance_multiple_eni_test.py
MrCloudSec: Something like: 
```
"EC2 Instance i-0123456789abcdef0 uses only one ENI: (EFAs: ['eni-2'], Trunks: ['eni-1', 'eni-3'].)"
```

prowler/providers/aws/services/ec2/ec2_instance_multiple_eni/ec2_instance_multiple_eni.py
MrCloudSec: Can we then ignore trunk ENIs in the check? So the user do not get confused when seeing the status extended since he is not going to be able to delete those trunk ENIs.

MarioRgzLpz: Why aren't you able to delete trunk ENIs? Looked for it in the docs but didn't find it. Still I think is a good idea to ignore trunk ENIs since they only add noise and no real value to the provided information.

MrCloudSec: We saw that trunks are normal interfaces with a specific option so I would count it as an ENI too.

MrCloudSec (approved):

PR #4961, opened 2024-09-09
chore(refactor): make Provider generation global

MrCloudSec:
### Context

The SDK as of today has two different ways to instantiate a Provider class which is causing some issues since if it is used from the SDK is not being set as global, because each service needs to get the provider.

### Description

Refactor the code to have just one way to init a Provider and each provider now calls `Provider.set_global_provider(self)` at the end of their `__init__()` method.

### Checklist

### License

### DIFF
--- a/prowler/__main__.py
+++ b/prowler/__main__.py
@@ -71,6 +71,8 @@ from prowler.providers.aws.lib.security_hub.security_hub import SecurityHub
 from prowler.providers.common.provider import Provider
 from prowler.providers.common.quick_inventory import run_provider_quick_inventory
 
+providers_path = "prowler.providers"
+
 
 def prowler():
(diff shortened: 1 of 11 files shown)


prowler/__main__.py
```diff
@@ -71,6 +71,8 @@
 from prowler.providers.common.provider import Provider
 from prowler.providers.common.quick_inventory import run_provider_quick_inventory
 
+providers_path = "prowler.providers"
+
```
MrCloudSec:
```suggestion
```

pedrooot (dismissed):
Needed refactor 💪🏼 Thanks for this!

pedrooot (approved):
LGTM!

PR #4973, opened 2024-09-10
feat(aws): Add new check to ensure RDS instances are not using default database engine ports

danibarranqueroo:
### Context

This new check verifies if an instance uses a port other than the default port of the database engine. The control fails if the RDS instance uses the default port or a non-default port that is not configured in the security group.

Since RDS instances belong to VPC Security Groups, which are managed by EC2, I've used this service to iterate over all the security groups and verify if an instance belongs to any of them. If so, I've checked that the security group has the port that the instance is listening to open. With this implementation, I can verify that if the instance uses a non-default port, it's configured correctly in the security group.

### Description

Added `rds_instance_not_default_port` check with its unit tests.

### Checklist

### License


prowler/providers/aws/services/rds/rds_instance_not_default_port/rds_instance_not_default_port.metadata.json
```diff
@@ -0,0 +1,32 @@
+{
+  "Provider": "aws",
+  "CheckID": "rds_instance_not_default_port",
```
MrCloudSec:
```suggestion
  "CheckID": "rds_instance_non_default_port",
```

prowler/providers/aws/services/rds/rds_instance_not_default_port/rds_instance_not_default_port.metadata.json
```diff
@@ -0,0 +1,32 @@
...
+  "CheckTitle": "Check if RDS instances are using non-default ports.",
+  "CheckType": [
+    "Software and Configuration Checks, AWS Security Best Practices"
```
MrCloudSec:
```suggestion
    "Software and Configuration Checks", "AWS Security Best Practices"
```

prowler/providers/aws/services/rds/rds_instance_not_default_port/rds_instance_not_default_port.py
```diff
@@ -0,0 +1,72 @@
...
+                    security_group,
+                ) in ec2_client.security_groups.items():
+                    if security_group.id in db_instance.security_groups:
+                        for ingress_rule in security_group.ingress_rules:
+                            if check_security_group(
+                                ingress_rule,
+                                "tcp",
+                                [db_instance.port],
+                                any_address=True,
+                            ):
```
MrCloudSec: This will only check if the port is public, not if it can be accessed by an IP address, is that what we want?

MrCloudSec: I would only check the ports for each instance in `db_instance.port`.

prowler/providers/aws/services/rds/rds_instance_not_default_port/rds_instance_not_default_port.py
```diff
@@ -0,0 +1,72 @@
+from prowler.lib.check.models import Check, Check_Report_AWS
+from prowler.providers.aws.services.ec2.ec2_client import ec2_client
+from prowler.providers.aws.services.ec2.lib.security_groups import check_security_group
+from prowler.providers.aws.services.rds.rds_client import rds_client
+
+
+class rds_instance_not_default_port(Check):
+    def execute(self):
+        findings = []
+        default_ports = {
```
MrCloudSec: Where did you get this list from?

MrCloudSec (approved):

PR #5003, opened 2024-09-11
fix(gcp): add default project for org level checks

MrCloudSec:
### Description

To avoid a KeyError in `finding.py` in the following line:
```
provider.projects[check_output.project_id]
```
We need to always add a Project ID in the `project_id` attribute of the checks, therefore, for Organization level checks, I created the `default_project_id` variable containing the GCP default project ID associated to the GCP credentials that Prowler is using.
### Checklist

### License

### DIFF
--- a/prowler/providers/gcp/gcp_provider.py
+++ b/prowler/providers/gcp/gcp_provider.py
@@ -68,7 +68,7 @@ class GcpProvider(Provider):
         logger.info("Instantiating GCP Provider ...")
         self._impersonated_service_account = impersonate_service_account
 
-        self._session = self.setup_session(
+        self._session, self._default_project_id = self.setup_session(
             credentials_file, self._impersonated_service_account
         )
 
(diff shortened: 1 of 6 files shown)


jfagoagas (commented):
Why is this needed now? What is the origin? If it is related to a new GCP SDK's update please look for it and link both PRs, thanks!

MrCloudSec:
> Why is this needed now? What is the origin? If it is related to a new GCP SDK's update please look for it and link both PRs, thanks!

This happened because in the checks that are at the organization level like `iam_organization_essential_contacts_configured`, we are setting the GCP organization ID as the Project ID which is causing a Key Error in the new `finding.py` file [here](https://github.com/prowler-cloud/prowler/blob/13c95ba131d7025f086b85ebbb849ee85b9215a8/prowler/lib/outputs/finding.py#L145) since it is not an actual Project ID.

jfagoagas (approved):

PR #5005, opened 2024-09-11
chore(ssm): add trusted accounts variable to ssm check

MrCloudSec:
### Description

Add the possibility of setting trusted AWS accounts for the check `ssm_documents_set_as_public` so Prowler gives only a FAIL when it is public or shared with a non-trusted account.

### Checklist

### License

### DIFF
--- a/tests/config/config_test.py
+++ b/tests/config/config_test.py
@@ -93,7 +93,7 @@ config_aws = {
         8080,
         8088,
     ],
-    "trusted_account_ids": ["580944000508"],
+    "trusted_account_ids": [],
     "log_group_retention_days": 365,
     "max_idle_disconnect_timeout_in_seconds": 600,
     "max_disconnect_timeout_in_seconds": 300,
--- a/tests/config/config_test.py
+++ b/tests/config/config_test.py
@@ -93,7 +93,7 @@ config_aws = {
         8080,
         8088,
     ],
-    "trusted_account_ids": [],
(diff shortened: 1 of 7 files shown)


tests/config/config_test.py
```diff
@@ -93,7 +93,7 @@ def mock_prowler_get_latest_release(_, **kwargs):
         8080,
         8088,
     ],
-    "trusted_account_ids": [],
+    "trusted_account_ids": ["580944000508"],
```
MrCloudSec:
```suggestion
    "trusted_account_ids": [],
```

tests/config/fixtures/config.yaml
```diff
@@ -58,10 +58,11 @@ aws:
     ]
 
   # AWS VPC Configuration (vpc_endpoint_connections_trust_boundaries, vpc_endpoint_services_allowed_principals_trust_boundaries)
+  # AWS SSM Configuration (aws.ssm_documents_set_as_public)
   # Single account environment: No action required. The AWS account number will be automatically added by the checks.
   # Multi account environment: Any additional trusted account number should be added as a space separated list, e.g.
   # trusted_account_ids : ["123456789012", "098765432109", "678901234567"]
-  trusted_account_ids: []
+  trusted_account_ids: ["580944000508"]
```
MrCloudSec:
```suggestion
  trusted_account_ids: []
```

pedrooot (approved):
Thanks for this improvement! 🦾

PR #5029, opened 2024-09-13
feat(rds): add new check `rds_instance_inside_vpc`

danibarranqueroo:
### Context

Implemented a check to ensure that Amazon RDS instances are deployed within a Virtual Private Cloud (VPC). This check helps enhance security and network isolation by verifying that RDS instances are not publicly accessible and are properly integrated within a VPC for controlled access.

The way to know if an instance is deployed in a VPC comes across the DBSubnetGroup parameter of the Amazon RDS API, it allows to know if an instance is within a VPC, as this group defines the subnets within a VPC where an RDS instance can be deployed. If the field is present, the instance is in a VPC, and if it's absent, the instance is likely in EC2-Classic.

Another problem that came up during testing of the check is that AWS no longer allows creating an instance outside of a VPC. So, this check would only be useful for verifying old RDS instances, but all new ones will be in a VPC.

This issue also affects the unit test to verify instances not in a VPC because neither boto nor Moto allows me to create an instance outside of a VPC or in the old EC2-Classic environment. Therefore, this unit test needs to be mocked using MagicMock.

### Description

Added new check `rds_instance_inside_vpc` with its unit tests.

### Checklist

### License

### DIFF
--- a/prowler/providers/aws/services/rds/rds_service.py
+++ b/prowler/providers/aws/services/rds/rds_service.py
@@ -100,6 +100,14 @@ class RDS(AWSService):
                                 copy_tags_to_snapshot=instance.get(
                                     "CopyTagsToSnapshot"
                                 ),
+                                db_subnet_group={
+                                    "name": instance.get("DBSubnetGroup", {}).get(
+                                        "DBSubnetGroupName", ""
+                                    ),
+                                    "vpc_id": instance.get("DBSubnetGroup", {}).get(
+                                        "VpcId", ""
+                                    ),
+                                },
                             )
         except Exception as error:
             logger.error(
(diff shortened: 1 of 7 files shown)


prowler/providers/aws/services/rds/rds_service.py
```diff
@@ -53,6 +53,7 @@ def _describe_db_instances(self, regional_client):
                         is_resource_filtered(arn, self.audit_resources)
                     ):
                         if instance["Engine"] != "docdb":
+                            print(instance.get("DBSubnetGroup", {}))
```
MrCloudSec:
```suggestion
```

MrCloudSec (approved):
Good job! Please, see my changes to the code, you have to make sure that the check is working as expected. The `db_subnet_group` variable was never empty since you were always adding the keys `name` and `vpc_id` making the check to always reporting a PASS.

PR #5067, opened 2024-09-17
chore(aws): improve IAM Resource Policy public logic

### LINKED ISSUE #4870: Many false positives for check awslambda_function_not_publicly_accessible
### Steps to Reproduce

Ahoy! I am running prowler periodically on AWS. Since the latest version (`Prowler 4.3.5`) i am getting a lot of false positives on check [awslambda_function_not_publicly_accessible](https://github.com/prowler-cloud/prowler/blob/f487bda1fe7c57650d4446189c1b4e6db7a5614f/prowler/providers/aws/services/awslambda/awslambda_function_not_publicly_accessible/awslambda_function_not_publicly_accessible.py). 

From my understanding, the code is missing a check whether a `condition` is set in the resource based policy. 

Here's a redacted example policy that i have in place for a lambda-script:

```json
{
  "Version": "2012-10-17",
  "Id": "default",
  "Statement": [
    {
      "Sid": "awslambda-myLambdaScript-LambdaInvokePermission",
      "Effect": "Allow",
      "Principal": {
        "Service": "ses.amazonaws.com"
      },
      "Action": "lambda:InvokeFunction",
      "Resource": "arn:aws:lambda:<REGION>:<ACCOUNT-ID>:function:myLambdaScript",
      "Condition": {
        "StringEquals": {
          "AWS:SourceAccount": "<ACCOUNT-ID>"
        }
      }
    }
  ]
}
```

Notice that there's a condition set to mitigate access from public. This condition allows SES to only invoke this function from within the account. 

But still this policy would get flagged as failed check, stating:
```
Lambda function myLambdaFunction has a policy resource-based policy with public access.
```

Neither am i a python expert, nor a prowler-developer, but shouldn't a `condition` statement get taken into consideration to determine whether a function is publicly accessible or not? 

Anyway - I have multiple of such Resource-based policy documents and all of these get flagged as publicly accessible, when they are in fact not. 

Could you double check and help me out verifying whether it's a false positive // the check's code is wrong? 

Cheers. :) 

### Expected behavior

The check `awslambda_function_not_publicly_accessible`  should not fail when the Resource-based policy document contains a `Condition` which restricts access, like:

```json
      "Condition": {
        "StringEquals": {
          "AWS:SourceAccount": "<ACCOUNT-ID>"
        }
      }
```

When using a `Condition`, the check would still fail and state:

```
Lambda function myLambdaFunction has a policy resource-based policy with public access.
```

### Environment Resource

2. Fargate Task

### OS used

Container source image: `docker.io/library/python:3.12-slim`

`cat /etc/os-release`

```
PRETTY_NAME="Debian GNU/Linux 12 (bookworm)"
NAME="Debian GNU/Linux"
VERSION_ID="12"
VERSION="12 (bookworm)"
```

### Prowler version

4.3.5

### Pip version

24.2

### Context

_No response_

MrCloudSec:
### Context

We were using different logics for checks where the public access on resource policies have to be verified.
Therefore, in some check Prowler was not checking the conditions.
Unify all those logics into a single function: `is_policy_public` function.

### Description
`is_policy_public` function allows to check if a resource policy is public and customize the checking with additional not allowed actions (it always checks `*`), an AWS Account to check if it is restrict in the policy condition and if the cross account access is allowed or not.

- Consolidate function `is_public_access_allowed` to `is_policy_public` function.
- Use `is_policy_public` function in checks where the IAM Resource Policy must be verified.
- Use `is_condition_block_restrictive`, `is_condition_block_restrictive_organization` and `is_condition_restricting_from_private_ip` in `is_policy_public` function.
- Replace `has_secure_conditions` by `is_condition_block_restrictive`, `is_condition_block_restrictive_organization` and `is_condition_restricting_from_private_ip` functions.

### Checklist

### License


MrCloudSec:
- [x] Call the `is_condition_block_restrictive` function with `is_cross_account_allowed = True` to the `awslambda_function_not_publicly_accessible` check.
- [ ] Improve the `is_condition_block_restrictive` to handle S3 ARNs which does not contains the AWS Account Number.

The second point cannot be addressed since S3 ARNs do not contain AWS account IDs.

pedrooot (dismissed):
🔝 Needed change, thanks for this!

pedrooot (dismissed):
LGTM!

pedrooot (approved):
🔝

PR #5084, opened 2024-09-18
fix(elasticache): get correct automatic failover attribute

HugoPBrito:
### Context

I noticed a mistake in a published check I made recently, so I’ll correct it. The service is picking an incorrect parameter from the dict response of the API. The check might work well because it was tested but it can have been a bad testing due to the mistake itself. 

### Description

We should pick:

```
{
    'Marker': 'string',
    'ReplicationGroups': [
        {
        'AutomaticFailover': 'enabled'|'disabled'|'enabling'|'disabling',
        }
    ]
}
```

But we were picking instead:

```
{
    'Marker': 'string',
    'ReplicationGroups': [
        {
            'PendingModifiedValues': {
                'AutomaticFailoverStatus': 'enabled'|'disabled',
            }
        }
    ]
} 
```   

### Checklist

### License

### DIFF
--- a/prowler/providers/aws/services/elasticache/elasticache_service.py
+++ b/prowler/providers/aws/services/elasticache/elasticache_service.py
@@ -107,7 +107,7 @@ class ElastiCache(AWSService):
                                 "AutoMinorVersionUpgrade", False
                             ),
                             automatic_failover=repl_group.get(
-                                "AutomaticFailoverStatus", "disabled"
+                                "AutomaticFailover", "disabled"
                             ),
                         )
                 except Exception as error:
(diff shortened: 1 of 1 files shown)


MrCloudSec (dismissed):
Good catch!

prowler/providers/aws/services/elasticache/elasticache_service.py
```diff
@@ -180,4 +180,4 @@ class ReplicationGroup(BaseModel):
     multi_az: str
     tags: Optional[list]
     auto_minor_version_upgrade: bool
-    automatic_failover: str
+    automatic_failover: Optional[str]
```
MrCloudSec:
```suggestion
    automatic_failover: str
```

MrCloudSec (approved):

PR #5099, opened 2024-09-19
feat(prowler-check-kreator): `ProwlerChecKreator` first version

puchy22:
### Context

I created a script that helps to create checks with AI (Gemini for now).

### Description

Added some scripts in utils.

### Checklist

- Are there new checks included in this PR? No?
    - If so, do we need to update permissions for the provider? 

### License


MrCloudSec (changes requested):
Please, do the following changes:

- Add the required Python library to the documentation.

- Fix -> Error: cannot access local variable ‘test_template’ when it has not been assigned a value.

- Rearrange questions and simplify them: "Do you want to use AI?" -> "Which model?" -> "Do you want to overwrite existing files?"

- Automatically include 10 reference examples by default, rather than relying on the user to provide them.

- Add support for additional providers.

- If the GEMINI_API_KEY environment variable is not set, prompt the user to enter it.

MrCloudSec (approved):
I love it @puchy22 !!

PR #5113, opened 2024-09-20
feat(rds): add new check `rds_cluster_non_default_port`

danibarranqueroo:
### Context

There’s a [control](https://docs.aws.amazon.com/securityhub/latest/userguide/rds-controls.html#rds-23) on SH that its tittle says it’s about instances but it verifies whether instances or clusters, so I preferred to separate this control into two new checks.
### Description

Added new check `rds_cluster_non_default_port` with its unit tests.

### Checklist

### License

### DIFF
--- a/prowler/providers/aws/services/rds/rds_cluster_non_default_port/rds_cluster_non_default_port.metadata.json
+++ b/prowler/providers/aws/services/rds/rds_cluster_non_default_port/rds_cluster_non_default_port.metadata.json
@@ -9,7 +9,7 @@
   "SubServiceName": "",
   "ResourceIdTemplate": "arn:aws:rds:region:account-id:cluster:db-cluster",
   "Severity": "low",
-  "ResourceType": "AwsRdsDbInstance",
+  "ResourceType": "AwsRdsDbCluster",
   "Description": "Checks if an cluster uses a port other than the default port of the database engine. The control fails if the RDS cluster uses the default port.",
   "Risk": "Using a default database port exposes the cluster to potential security vulnerabilities, as attackers are more likely to target known, commonly-used ports. This may result in unauthorized access to the database or increased susceptibility to automated attacks.",
   "RelatedUrl": "https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Overview.DBInstance.Modifying.html",
--- a/prowler/providers/aws/services/rds/rds_cluster_non_default_port/rds_cluster_non_default_port.metadata.json
+++ b/prowler/providers/aws/services/rds/rds_cluster_non_default_port/rds_cluster_non_default_port.metadata.json
@@ -1,0 +1,32 @@
+{
+  "Provider": "aws",
+  "CheckID": "rds_cluster_non_default_port",
+  "CheckTitle": "Check if RDS clusters are using non-default ports.",
+  "CheckType": [
+    "Software and Configuration Checks/AWS Security Best Practices"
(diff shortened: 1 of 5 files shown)


prowler/providers/aws/services/rds/rds_cluster_non_default_port/rds_cluster_non_default_port.metadata.json
```diff
@@ -0,0 +1,32 @@
...
+  "CheckID": "rds_cluster_non_default_port",
+  "CheckTitle": "Check if RDS clusters are using non-default ports.",
+  "CheckType": [
+    "Software and Configuration Checks/AWS Security Best Practices"
+  ],
+  "ServiceName": "rds",
+  "SubServiceName": "",
+  "ResourceIdTemplate": "arn:aws:rds:region:account-id:cluster:db-cluster",
+  "Severity": "low",
+  "ResourceType": "AwsRdsDbInstance",
```
MrCloudSec:
```suggestion
  "ResourceType": "AwsRdsDbCluster",
```

prowler/providers/aws/services/rds/rds_cluster_non_default_port/rds_cluster_non_default_port.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+class rds_cluster_non_default_port(Check):
+    def execute(self):
+        findings = []
+        default_ports = {
+            3306: ["mysql"],
+            5432: ["postgres"],
+            8182: ["neptune"],
+        }
```
MrCloudSec: Why only this ones?

danibarranqueroo: Those are the only engines that support clustered instances, so the difference with the instance check is that here we have fewer engine types, which is why I list these ports differently.

You can check it [here](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/rds/client/create_db_cluster.html):

**Engine (string) – [REQUIRED]**

The database engine to use for this DB cluster.

**Valid for Cluster Type**: Aurora DB clusters and Multi-AZ DB clusters

**Valid Values**:
- aurora-mysql
- aurora-postgresql
- mysql
- postgres
- neptune

If I'm wrong with this assumption, please ping me, and I'll make the necessary changes.

prowler/providers/aws/services/rds/rds_cluster_non_default_port/rds_cluster_non_default_port.metadata.json
danibarranqueroo: Done!

prowler/providers/aws/services/rds/rds_cluster_non_default_port/rds_cluster_non_default_port.py
MrCloudSec: These are the different options for creating a DB Cluster in the AWS Console, remember that not all engines are supported for creating clusters in the AWS CLI.
image
Can you add all of them? They are the same as the instances check, right?

danibarranqueroo: Okay, yes they are the same as the instances so I'll add them, thanks for the help! 😄

MrCloudSec (approved):

PR #5120, opened 2024-09-20
fix(iam): fill resource id with inline policy entity

pedrooot:
### Description

Some checks from inline policies were not adding info related with the policy entity inside the resource id.

### Checklist

### License


MrCloudSec (changes requested):
Please, make sure that the status extended for the 4 `iam_inline_policy_*` checks are like:

`Inline policy <POLICY_NAME> attached to <ENTITY_TYPE> <ENTITY_NAME> ...`

Like in check `iam_inline_policy_no_administrative_privileges`.

MrCloudSec (changes requested):
Please, can you put the resource name instead of the resource Arn in the status extended?

Also, we can get the resource type using `arn.split(":")[-1].split("/")[1]` and not using that amount of lines per code for each check.

MrCloudSec (approved):

PR #5128, opened 2024-09-20
feat(ecs): Ensure public IP addresses are not assigned automatically

MarioRgzLpz:
### Context

Added a new check that verifies whether Amazon ECS services are configured to automatically assign public IP addresses. The check fails if AssignPublicIP is ENABLED and passes if AssignPublicIP is DISABLED.

A public IP address is an IP address that is reachable from the internet. If you launch your Amazon ECS instances with a public IP address, then your Amazon ECS instances are reachable from the internet. Amazon ECS services should not be publicly accessible, as this may allow unintended access to your container application servers.

### Description

Changed `ecs_service.py` to add the new `Service` model and two new methods to list and describe services. Added new `ecs_service_no_assign_public_ip.py` with respective unit tests.

### Checklist

### License

### DIFF
--- a/prowler/providers/aws/services/ecs/ecs_service.py
+++ b/prowler/providers/aws/services/ecs/ecs_service.py
@@ -15,12 +15,14 @@ class ECS(AWSService):
         super().__init__(__class__.__name__, provider)
         self.task_definitions = {}
         self.services = {}
+        self.clusters = {}
         self.__threading_call__(self._list_task_definitions)
         self.__threading_call__(
             self._describe_task_definition, self.task_definitions.values()
         )
-        self.__threading_call__(self._list_services)
-        self.__threading_call__(self._describe_services, self.services.values())
+        self.__threading_call__(self._list_clusters)
+        self.__threading_call__(self._describe_clusters, self.clusters.values())
+        self.__threading_call__(self._describe_services, self.clusters.values())
 
     def _list_task_definitions(self, regional_client):
         logger.info("ECS - Listing Task Definitions...")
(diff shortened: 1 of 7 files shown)


MrCloudSec (changes requested):
In order to `list_services`, you need to list the clusters first. Remember that the services are inside a ECS Cluster. If you don't input a cluster, the `list_services` gets the default one which may not exist -> https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ecs/client/list_services.html
Please, test it in a real scenario since you are not using moto.

MarioRgzLpz:
@sergargar Change the ecs service to get the services using the clusters arn (added Cluster model too) and add unit tests for this changes -> [17bbcd5](https://github.com/prowler-cloud/prowler/pull/5128/commits/17bbcd5a255e1c5cab5bdc885a00b1797d07be35)

prowler/providers/aws/services/ecs/ecs_service.py
```diff
@@ -79,42 +81,74 @@ def _describe_task_definition(self, task_definition):
...
-            list_ecs_paginator = regional_client.get_paginator("list_services")
+            client = self.regional_clients[cluster.region]
+
+            list_ecs_paginator = client.get_paginator("list_services")
+            service_arns = []
+            for page in list_ecs_paginator.paginate(cluster=cluster.arn):
+                service_arns.extend(page["serviceArns"])
+
+            if service_arns:
+                describe_response = client.describe_services(
```
MrCloudSec: Can you check this limit?
`You may specify up to 10 services to describe in a single operation.`
Maybe you can call the API for each service arn.

MrCloudSec (changes requested):
Thank you for the changes, good job! 

I left you one last comment to address.

prowler/providers/aws/services/ecs/ecs_service.py
MarioRgzLpz: Changed to call the API once for each service arn -> [82ebb90](https://github.com/prowler-cloud/prowler/pull/5128/commits/82ebb9036909132aea78ea4e6b5be609e517aa74)

prowler/providers/aws/services/ecs/ecs_service.py
```diff
@@ -39,41 +46,110 @@ def _list_task_definitions(self, regional_client):
...
+            list_ecs_paginator = client.get_paginator("list_services")
+            service_arns = []
+            for page in list_ecs_paginator.paginate(cluster=cluster.arn):
+                service_arns.extend(page["serviceArns"])
+
+            if service_arns:
+                for service_arn in service_arns:
+                    describe_response = client.describe_services(
+                        cluster=cluster.arn,
+                        services=service_arn,
```
MrCloudSec:
```suggestion
                        services=[service_arn],
```

MrCloudSec (approved):

PR #5129, opened 2024-09-20
feat(dynamodb): add new check `dynamodb_table_autoscaling_enabled`

danibarranqueroo:
### Context

This new check verifies that Amazon DynamoDB tables are configured to automatically scale their capacity based on demand. The control fails if the table does not use on-demand capacity mode or provisioned mode is used without auto scaling enabled.
Auto scaling helps ensure that DynamoDB tables adjust their read and write capacity dynamically in response to traffic patterns, which prevents throttling exceptions and maintains application availability. Tables in on-demand mode automatically adjust capacity to handle varying workloads, while provisioned mode with auto scaling adjusts based on predefined parameters.

As we can see here, the billing mode of the table set if the capacity is adjust to on-demand or provisioned:

```
BillingMode (string) –
Controls how you are charged for read and write throughput and how you manage capacity. This setting can be changed later.
PROVISIONED - Sets the read/write capacity mode to PROVISIONED. We recommend using PROVISIONED for predictable workloads.
PAY_PER_REQUEST - Sets the read/write capacity mode to PAY_PER_REQUEST. We recommend using PAY_PER_REQUEST for unpredictable workloads.
```
And then, if the table is set as PROVISIONED, you can choose to have or not enabled read/write capacity to auto-scaling. This can be verify using the ApplicationAutoScaling service of AWS.

### Description

Added new check `dynamodb_table_autoscaling_enabled`, modified the service and added unit tests.

### Checklist

### License


prowler/providers/aws/services/dynamodb/dynamodb_service.py
```diff
@@ -120,6 +123,37 @@ def _get_resource_policy(self):
...
+                            ScalableDimension="dynamodb:table:WriteCapacityUnits",
+                        )
+                    )
+                    if write_response["ScalableTargets"]:
+                        table.write_autoscaling = True
+        except Exception as error:
+            logger.error(
+                f"{error.__class__.__name__}:{error.__traceback__.tb_lineno} -- {error}"
+            )
+
```
MrCloudSec: Please, as you are using another client, put this function in the `autoscaling` service in a new class.

MrCloudSec (approved):

PR #5144, opened 2024-09-23
feat(ecr): Ensure ECR repositories have tag immutability configured

MarioRgzLpz:
### Context

Added new logic to check whether a ECR repository has tag immutability enabled. The check fails if a private ECR repository has tag immutability disabled and passes if tag immutability is set to IMMUTABLE.
Amazon ECR Tag Immutability enables customers to rely on the descriptive tags of an image as a reliable mechanism to track and uniquely identify images. An immutable tag is static, which means each tag refers to a unique image. This improves reliability and scalability as the use of a static tag will always result in the same image being deployed. When configured, tag immutability prevents the tags from being overridden, which reduces the attack surface.

### Description

Change `Repository` model in `ecr_service` to add a new atributte `immutability` for the check. Added new check `ecr_repositories_immutability_tag_configured` with respective unit tests.

### Checklist

### License

### DIFF
--- a/prowler/providers/aws/services/ecr/ecr_repositories_immutability_tag_configured/ecr_repositories_immutability_tag_configured.metadata.json
+++ b/prowler/providers/aws/services/ecr/ecr_repositories_immutability_tag_configured/ecr_repositories_immutability_tag_configured.metadata.json
@@ -1,0 +1,32 @@
+{
+  "Provider": "aws",
+  "CheckID": "ecr_repositories_immutability_tag_configured",
+  "CheckTitle": "ECR private repositories should have tag immutability configured",
+  "CheckType": [
+    "Software and Configuration Checks/AWS Security Best Practices"
+  ],
+  "ServiceName": "ecr",
+  "SubServiceName": "",
+  "ResourceIdTemplate": "arn:aws:ecr:{region}:{account-id}:repository/{repository-name}",
+  "Severity": "medium",
+  "ResourceType": "AwsEcrRepository",
+  "Description": "This control checks whether a private ECR repository has tag immutability enabled. The control fails if a private ECR repository has tag immutability disabled and passes if tag immutability is set to IMMUTABLE.",
+  "Risk": "Disabling tag immutability in an ECR repository increases the risk of image tampering by allowing tags to be overridden, which can lead to deploying unintended images and security vulnerabilities.",
+  "RelatedUrl": "https://docs.aws.amazon.com/config/latest/developerguide/ecr-private-tag-immutability-enabled.html",
+  "Remediation": {
+    "Code": {
+      "CLI": "aws ecr put-image-tag-mutability --repository-name <repository-name> --image-tag-mutability IMMUTABLE",
+      "NativeIaC": "",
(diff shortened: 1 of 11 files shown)


prowler/providers/aws/services/ecr/ecr_repositories_immutability_tag_configured/ecr_repositories_immutability_tag_configured.metadata.json
```diff
@@ -0,0 +1,32 @@
+{
+  "Provider": "aws",
+  "CheckID": "ecr_repositories_immutability_tag_configured",
```
MrCloudSec:
```suggestion
  "CheckID": "ecr_private_repository_tag_immutability",
```

prowler/providers/aws/services/ecr/ecr_repositories_immutability_tag_configured/ecr_repositories_immutability_tag_configured.py
MrCloudSec, line 1: You should only check private repositories.

prowler/providers/aws/services/ecr/ecr_service.py
```diff
@@ -47,6 +47,8 @@ def _describe_registries_and_repositories(self, regional_client):
                                 scan_on_push=repository["imageScanningConfiguration"][
                                     "scanOnPush"
                                 ],
+                                immutability=repository["imageTagMutability"]
+                                == "IMMUTABLE",
```
MrCloudSec:
```suggestion
                                immutability=repository.get("imageTagMutability", "MUTABLE")
```

Please, compare the strings in the check.

prowler/providers/aws/services/ecr/ecr_repositories_immutability_tag_configured/ecr_repositories_immutability_tag_configured.py
MarioRgzLpz: As far as I understand using [ECR](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ecr.html) we are only returning the private repositories. To return the public repositories we should use [ECRPublic](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ecr-public.html). Maybe Im missing something and we are indeed returning all repositories, in that case we should change some checks that right now are checking every repository including public ones.

MarioRgzLpz:
@sergargar Requested changes added in commit -> [6e3083d](https://github.com/prowler-cloud/prowler/pull/5144/commits/6e3083d4c16d05088ee54c850235784f47ed5edc)

prowler/providers/aws/services/ecr/ecr_repositories_immutability_tag_configured/ecr_repositories_immutability_tag_configured.py
MrCloudSec: You're right! Thanks for the info

MrCloudSec (approved):
Good job!

I changed the name again to follow the same pattern as the rest of ECR checks.

PR #5166, opened 2024-09-24
feat(threat-detection): Use IAM Identity for Cloudtrail Threat Detection instead of IP

### LINKED ISSUE #5057: Use IAM identity to calculate threat detection threshold
### New feature motivation

The current implementation of threat detection (both privilege escalation and permission enumeration) uses number of potential priv esc/enumeration attempts per source IP address. While it detects simple bruteforce from single IP, the bypass is really easy - combine pacu/iam enumerate with IP rotation tools. The threshold will stay under default threshold limit and doesn't detect the attack.

### Solution Proposed

Trying to find threshold based on IAM identity could help detect attacks. Let's say attacker uses pacu/iam enumerate with IP rotation tools, the IAM identity that the attacker uses is still the same. Using threshold based on IAM identity, prowler can find and flag the identity.

N/A

### Additional context

_No response_

abant07:
### Context

Currently Cloudtrail threat detection has two checks, enumeration and privilege escalation, and the way that these were detected was by seeing how frequently a ip address was accessing a resource in a specific amount of time, and then applying a threshold. If the rate exceeds the threshold, it flags the cloudtrail threat detection. However, a attacker could get around this and just combine pacu/iam enumerate with IP rotation tools and then be free to make api calls to these services unlimited times without triggering any check.

Fix #5057

### Description

The change is that instead of detecting the rate of the ip address that are making the programmatical calls to resources, we instead check the frequency of the AWS account that is making the calls, because this is something that cannot be changed with any tools unlike a ip address. In terms of changing the code, I just changed the threat detection files to look for ["userIdentity"]["priviledgeId"] instead of ['sourceIpAddress"]

### License


MrCloudSec:
Thanks for the PR @abant07 ! 

However, the principal ID of an IAM identity is sometimes hard to find in AWS. Could you use the ARN, and then, in the status extended something like:

`f"Potential enumeration attack detected from AWS {aws_identity_type} {aws_identity_arn.split("/")[-1]} with an threshold of {identity_threshold}."`

What do you think?

abant07:
Sure I can do that

MrCloudSec:
@abant07 can you solve the actions?

If you install our pre-commit, it can do it automatically for you.

abant07:
Oh how can I install it?

abant07:
Hey @sergargar 

It seems there are two cloudfront tests failing however I did not edit those? Everything else seems to be fine

MrCloudSec:
> Oh how can I install it?

`pre-commit install`

MrCloudSec:
> Hey @sergargar
> 
> It seems there are two cloudfront tests failing however I did not edit those? Everything else seems to be fine

Hi @abant07, no worries, I will address those errors.

MrCloudSec (approved):
Thanks for your contribution @abant07 !

PR #5175, opened 2024-09-24
feat(dynamodb): add new check `dynamodb_table_protected_by_backup_plan`

danibarranqueroo:
### Context

This new check verifies whether Amazon DynamoDB tables are included in a backup plan. The control fails if the table is not covered by a backup plan.
AWS Backup provides a centralized and automated way to manage backups across AWS services. By including DynamoDB tables in your backup plans, you protect against data loss or unintended deletion. SH uses a backupVaultLockCheck parameter that ensures backups are stored in a secure vault with additional protection, but here in prowler we decided a couple of similar checks ago that we won’t use that parameter.

As I need to iterate over all the backup plans and verify if any table is inside one of them, I decided to change tables from list to dict, this makes me change some checks already merged.

### Description

Added new check `dynamodb_table_protected_by_backup_plan` with its unit tests and modify code to change tables from list to dict.

### Checklist

### License

### DIFF
--- a/prowler/providers/aws/services/dynamodb/dynamodb_service.py
+++ b/prowler/providers/aws/services/dynamodb/dynamodb_service.py
@@ -32,7 +32,6 @@ class DynamoDB(AWSService):
                         is_resource_filtered(arn, self.audit_resources)
                     ):
                         table = Table(
-                            arn=arn,
                             name=table,
                             encryption_type=None,
                             kms_arn=None,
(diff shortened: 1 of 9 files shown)


prowler/providers/aws/services/dynamodb/dynamodb_service.py
```diff
@@ -31,15 +31,13 @@ def _list_tables(self, regional_client):
...
-                                kms_arn=None,
-                                region=regional_client.region,
-                            )
+                        table = Table(
+                            name=table,
+                            encryption_type=None,
+                            kms_arn=None,
+                            region=regional_client.region,
                         )
+                        self.tables[arn] = table
```
MrCloudSec:
```suggestion
                        self.tables[arn] = Table(
                            name=table,
                            encryption_type=None,
                            kms_arn=None,
                            region=regional_client.region,
                        )
```

MrCloudSec (changes requested):
This check and all `protected_by_backup_plan` checks give a false negative when resources are protected by a backup plan but they are not still in a backup vault.

Please use the API Calls `list-backup-selections` and `get-backup-selection` to get a more accurate information.

PD: change the logic of the Backup service and related checks in another PR.

MrCloudSec (approved):
Approve but pending to change the backup logic.

PR #5204, opened 2024-09-26
fix(aws): change `protected_by_backup_plan` checks

danibarranqueroo:
### Context

Change AWS checks protected_by_backup_plan since the logic is not correct.

@sergargar has discovered that when a new backup plan is created with a resource, we return a FAIL because it doesn't show up in the list_protected_resources call. It only appears after the first backup has been made. This means we need to update the checks related to protected_by_backup_plan and use the list-backup-selections and get-backup-selection calls to view resources associated with a backup plan.

Selections can include: "Resources": [ "arn:aws:ec2:*:*:instance/*" ] which applies to all instances, or a wildcard (*) covering all resources, so this has been added too.

### Description

Modify the code and the unit tests for this checks:

- `dynamodb_table_protected_by_backup_plan`
- `ec2_ebs_volume_protected_by_backup_plan`
- `rds_instance_protected_by_backup_plan`

### Checklist

### License


MrCloudSec (approved):
Good job!

I modified the code since you were only getting the protected resources of the last iterated backup selection. Please @danibarranqueroo, be careful with that type of logic.

PR #5216, opened 2024-09-26
fix(aws): handle none type attributes

MrCloudSec:
### Description

Handle `None` type attributes in checks:
- `organizations_scp_check_deny_regions`
- `ecr_repositories_not_publicly_accessible`

Also, fix typo in work `accessible`.

### Checklist

### License

### DIFF
--- a/prowler/providers/aws/services/organizations/organizations_scp_check_deny_regions/organizations_scp_check_deny_regions.py
+++ b/prowler/providers/aws/services/organizations/organizations_scp_check_deny_regions/organizations_scp_check_deny_regions.py
@@ -12,86 +12,87 @@ class organizations_scp_check_deny_regions(Check):
         )
 
         for org in organizations_client.organizations:
-            report = Check_Report_AWS(self.metadata())
-            report.resource_id = org.id
-            report.resource_arn = org.arn
-            report.region = organizations_client.region
-            report.status = "FAIL"
-            report.status_extended = (
-                "AWS Organizations is not in-use for this AWS Account."
-            )
-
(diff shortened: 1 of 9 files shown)


prowler/providers/aws/services/organizations/organizations_scp_check_deny_regions/organizations_scp_check_deny_regions.py
jfagoagas, line 1: Could you please add tests for the following lines:
- 74
- 82 to 85
- 88 to 90

If we are changing the logic of the check we should cover those scenarios.

jfagoagas (changes requested):

prowler/providers/aws/services/organizations/organizations_scp_check_deny_regions/organizations_scp_check_deny_regions.py
MrCloudSec: @jfagoagas the logic of the check was not changed, only the following line was added:
`if org.policies is not None: # Access denied to list policies`

jfagoagas: Yes but even with that I think those lines are not being tested, you should add a test to verify when `org.policies is None`, don't you?

MrCloudSec: Yes, test was added!

jfagoagas (approved):
Great one, thanks @sergargar 🚀

PR #5259, opened 2024-10-01
fix(ecs): Adjust code to the new ARN formats in the ECS service

MarioRgzLpz:
### Context

While doing a check I noticed that AWS was updating some ARNs that would break our current ecs service. In order to stop that from happening we need to adjust the code to the new ARN format and making it that works with the actual ARN. 

Link to the ARN updates [here](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-account-settings.html#ecs-resource-ids). Chapter: Amazon Resource Name (ARN) format

### Description

Changed how we get the name of services and task definitions to avoid it to fail once the new ARN format is implemented.

### Checklist

### License


MrCloudSec (approved):
🔝

PR #5271, opened 2024-10-02
feat(redshift): add new check `redshift_cluster_in_transit_encryption_enabled`

danibarranqueroo:
### Context

This control ensures that connections to Amazon Redshift clusters are encrypted in transit by requiring TLS/SSL encryption. The check specifically verifies that the Redshift cluster parameter require_SSL is set to True, mandating that all connections use TLS. This encryption protects against man-in-the-middle attacks and other security vulnerabilities by securing data as it moves between clients and the Redshift cluster. 

During the creation of the unit tests, I realized that Moto does not cover the API call `modify_cluster_parameter_group`, so I wasn't able to set the `require_ssl` parameter to the desired value. That's why I needed to use Botocore for these tests.

### Description

Added new check `redshift_cluster_in_transit_encryption_enabled` with its unit tests and create a new API call in `redshift_service` in order to scan cluster parameters. 

### Checklist

### License


MrCloudSec (approved):
👏🏼

PR #5275, opened 2024-10-02
feat(efs): add new check `efs_mount_target_not_publicly_accesible`

MarioRgzLpz:
### Context

Added a new check that make sure Amazon Elastic File System (EFS) mount targets are associated with a private subnet. The control fails if any EFS mount target is associated with a public subnet.

By default, EFS file systems are accessible only from the VPC in which they are created. To enhance security, it's recommended that EFS mount targets be created in private subnets. This configuration ensures that the EFS file system is protected from unauthorized access and potential attacks, as it will not be directly accessible from the internet.

### Description

Added new `MountTarget` model to `efs_service`. Added new check `efs_mount_target_not_publicly_accesible` with respective unit tests. Change filesystem from list to dick in the `efs service`. Changed old efs checks and tests to adjust to the service changes.

### Checklist

    - If so, do we need to update permissions for the provider?No

### License


MrCloudSec (commented):
👏🏼 👏🏼

MrCloudSec (approved):

PR #5285, opened 2024-10-03
feat(efs): add new check `efs_access_point_enforce_user_identity`

MarioRgzLpz:
### Context

Add a new check to ensure Amazon Elastic File System (EFS) access points are configured to enforce a POSIX user identity. The control fails if a POSIX user identity is not defined during the creation of the EFS access point.

Enforcing a user identity on EFS access points ensures that file system requests through the access point adhere to specified POSIX user and group settings. This configuration helps manage and restrict access to shared datasets, ensuring that only authorized users can access or modify the data.

### Description

Added a new attribute `posix_user` to the `AccessPoint` model in the `efs_service`. Added new check `efs_access_point_enforce_user_identity` with respective unit tests.

### Checklist

    - If so, do we need to update permissions for the provider?No

### License


prowler/providers/aws/services/efs/efs_access_point_enforce_user_identity/efs_access_point_enforce_user_identity.py
```diff
@@ -0,0 +1,28 @@
...
+        for fs in efs_client.filesystems.values():
+            if fs.access_points:
+                report = Check_Report_AWS(self.metadata())
+                report.region = fs.region
+                report.resource_id = fs.id
+                report.resource_arn = fs.arn
+                report.resource_tags = fs.tags
+                report.status = "PASS"
+                report.status_extended = (
+                    f"EFS {fs.id} has all access point with defined POSIX user."
```
MrCloudSec:
```suggestion
                    f"EFS {fs.id} has all access points with defined POSIX user."
```

MrCloudSec (approved):

PR #5286, opened 2024-10-03
chore(cloudwatch): Improve checks related with function `check_cloudwatch_log_metric_filter`

puchy22:
### Context

The function `check_cloudwatch_log_metric_filter` is used in several CloudWatch checks and it is defined in a strange way. It is returning a report that is the same object that is getting passed as parameter, this happens because Python with mutable objects that are passed to a function it works by reference so the return in the function now is not making anything. For that reason it should be changed to create a new object and return it, instead modifying a object passed as argument.

### Description

- Create report to return instead change reference in `check_cloudwatch_log_metric_filter`
- Add logic for not permissions for metrics filter in `metric_filters`
- Change check logic to use new metric filter function

### Checklist

### License


MrCloudSec (approved):
Thanks!!

PR #5287, opened 2024-10-03
chore(ocsf): adapt mapping for version 1.3.0

MrCloudSec:
### Context

OCSF version 1.3.0 introduces some breaking changes that have been changed in https://github.com/prowler-cloud/py-ocsf-models/pull/123.

### Description

Adapt Prowler mapping to the new version of OCSF 1.3.0, including also the recommendations from https://github.com/ocsf/examples/pull/78.

### Checklist

### License


prowler/lib/outputs/ocsf/ocsf.py
```diff
@@ -129,26 +151,19 @@ def transform(self, findings: List[Finding]) -> None:
...
                             type_id=cloud_account_type.value,
-                            type=cloud_account_type.name,
+                            type=cloud_account_type.name.replace("_", " "),
                             uid=finding.account_uid,
                             labels=unroll_dict_to_list(finding.account_tags),
                         ),
                         org=Organization(
                             uid=finding.account_organization_uid,
                             name=finding.account_organization_name,
+                            # TODO: add the org unit id and name
```
jfagoagas, line 166: Are we going to implement it in this PR?

prowler/lib/outputs/ocsf/ocsf.py
```diff
@@ -59,21 +58,24 @@ def transform(self, findings: List[Finding]) -> None:
                 finding_severity = getattr(
                     SeverityID, finding.severity.capitalize(), SeverityID.Unknown
                 )
-                finding_status = self.get_finding_status_id(
-                    finding.status, finding.muted
-                )
+                finding_status = self.get_finding_status_id(finding.muted)
 
                 detection_finding = DetectionFinding(
+                    message=finding.status_extended,
```
jfagoagas, line 64: Why this new field? I saw that it is recommended but it is the same as `status_detail`

tests/lib/outputs/ocsf/ocsf_test.py
```diff
@@ -87,6 +99,36 @@ def test_transform(self):
...
+        ]
+
+        output = OCSF(findings)
+        output._file_descriptor = mock_file
+
+        with patch.object(mock_file, "close", return_value=None):
+            output.batch_write_data_to_file()
+
+        mock_file.seek(0)
+        content = mock_file.read()
```
jfagoagas, line 125: do you need to write the contents to that mock object? You can just load the object into the request as a JSON body.

jfagoagas (changes requested):
That's great @sergargar 👏 left some tiny questions/changes.

prowler/lib/outputs/ocsf/ocsf.py
MrCloudSec: Nope.

prowler/lib/outputs/ocsf/ocsf.py
MrCloudSec: Yes, but I think it is good to see it at first.

jfagoagas (approved):
🚀

PR #5305, opened 2024-10-07
feat(scan): add scan duration

pedrooot:
### Description

Scan duration is needed for the API usage. This PR adds this property to the `Scan` class.

### Checklist

### License

### DIFF
--- a/prowler/lib/scan/scan.py
+++ b/prowler/lib/scan/scan.py
@@ -19,7 +19,7 @@ class Scan:
     _service_checks_completed: dict[str, set[str]]
     _progress: float = 0.0
     _findings: list = []
-    _duration: float = 0.0
+    _duration: int = 0
 
     def __init__(self, provider: Provider, checks_to_execute: list[str]):
         """
@@ -64,7 +64,7 @@ class Scan:
         )
 
     @property
-    def duration(self) -> float:
+    def duration(self) -> int:
         return self._duration
 
     @property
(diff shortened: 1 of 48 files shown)


prowler/lib/scan/scan.py
```diff
@@ -18,6 +19,7 @@ class Scan:
     _service_checks_completed: dict[str, set[str]]
     _progress: float = 0.0
     _findings: list = []
+    _duration: float = 0.0
```
jfagoagas: I think an `int` is more than enough here.

prowler/lib/scan/scan.py
```diff
@@ -159,6 +166,8 @@ def scan(
                     logger.error(
                         f"{check_name} - {error.__class__.__name__}[{error.__traceback__.tb_lineno}]: {error}"
                     )
+            # Update the scan duration when all checks are completed
+            self._duration = (datetime.datetime.now() - start_time).total_seconds()
```
jfagoagas: Can we remove the microsencond precision, more than enough with seconds.

jfagoagas (changes requested):

prowler/lib/scan/scan.py
pedrooot: Done

prowler/lib/scan/scan.py
pedrooot: Done

jfagoagas (approved):

PR #5306, opened 2024-10-07
FIX Update elasticache_redis_cluster_backup_enabled.metadata.json

nunofranciscomoreira:
Set the Severity with the same value as the Severity set on the check.

### Context

The check `prowler/providers/aws/services/elasticache/elasticache_redis_cluster_backup_enabled/elasticache_redis_cluster_backup_enabled.py` sets the Severity as `low` on line 26. 
The metadata file `prowler/providers/aws/services/elasticache/elasticache_redis_cluster_backup_enabled/elasticache_redis_cluster_backup_enabled.metadata.json` sets the Severity as `high`. 

### Description

This makes prowler publish results with a Low severity when running with --severity high.

The check test is asserting against "low" so the Metadata file has to be changed to low in order to prevent it from being launched when running with --severity high.

### Checklist

### License

### DIFF
--- a/prowler/providers/aws/services/elasticache/elasticache_redis_cluster_backup_enabled/elasticache_redis_cluster_backup_enabled.metadata.json
+++ b/prowler/providers/aws/services/elasticache/elasticache_redis_cluster_backup_enabled/elasticache_redis_cluster_backup_enabled.metadata.json
@@ -6,7 +6,7 @@
   "ServiceName": "elasticache",
   "SubServiceName": "",
   "ResourceIdTemplate": "arn:partition:service:region:account-id:resource-id",
-  "Severity": "high",
+  "Severity": "low",
   "ResourceType": "Other",
   "Description": "Ensure Elasticache Redis cache cluster has automatic backups enabled.",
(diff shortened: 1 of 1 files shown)


MrCloudSec:
Hi @nunofranciscomoreira, this check only has a low severity if the cluster has automated snapshot backups enabled but with a retention period of less than 7 days.

nunofranciscomoreira:
Hi @sergargar, I understand your point, perhaps this change is not required.
But please consider the case where you only run for high severity findings.
That low should not appear on the output. or am I wrong?

nunofranciscomoreira:
And not only regarding this check. I believe there are more cases where the severity is decreased during the check. 
Should these findings appear on the output if they don't match the required severity?

MrCloudSec:
> And not only regarding this check. I believe there are more cases where the severity is decreased during the check. Should these findings appear on the output if they don't match the required severity?

@nunofranciscomoreira  that's a really good point, can you open an issue for this request? So we can include it in our roadmap 😄 thanks! By now, I close this PR.

PR #5307, opened 2024-10-07
feat(scan): execute all checks if no checks are provided

pedrooot:
### Description

If no checks are provided when the scan is initialized, all checks for the provider will be executed. Threat detection checks will also be discarded.

### Checklist

### License


prowler/lib/scan/scan.py
```diff
@@ -19,7 +20,7 @@ class Scan:
     _progress: float = 0.0
     _findings: list = []
 
-    def __init__(self, provider: Provider, checks_to_execute: list[str]):
+    def __init__(self, provider: Provider, checks_to_execute: list[str] = []):
```
jfagoagas:
```suggestion
    def __init__(self, provider: Provider, checks_to_execute: list[str] = None):
```

rationale -> https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

scan.py
jfagoagas, line 1: Watch out this file, do you want to push it?

prowler/lib/scan/scan.py
```diff
@@ -29,11 +30,31 @@ def __init__(self, provider: Provider, checks_to_execute: list[str]):
         """
         self._provider = provider
         # Remove duplicated checks and sort them
-        self._checks_to_execute = sorted(list(set(checks_to_execute)))
+        self._checks_to_execute = (
```
jfagoagas, line 33: Why are you doing this here? I mean the `check[0]`.

jfagoagas (changes requested):

prowler/lib/scan/scan.py
pedrooot: In this case, `recover_checks_from_provider` returns a tuple with the check name and the check path, as I'm just storing the check name I take the first value from the tuple.

prowler/lib/scan/scan.py
pedrooot: Ok! Thanks

scan.py
pedrooot: My bad 🤦🏼

prowler/lib/scan/scan.py
jfagoagas: Good catch, I forgot about it!

jfagoagas (approved):
🚀

PR #5325, opened 2024-10-08
feat(autoscaling): add new check `autoscaling_group_multiple_instance_types`

danibarranqueroo:
### Context

This new check ensures that Amazon EC2 Auto Scaling groups use multiple instance types and deploy instances across multiple Availability Zones. Using multiple instance types enhances availability by allowing the Auto Scaling group to launch alternative instance types if there is insufficient capacity in the chosen Availability Zones, improving resilience and high availability.

I have done the check following [this](https://www.trendmicro.com/cloudoneconformity/knowledge-base/aws/AutoScaling/asg-multiple-instance-type-az.html) guide, but I have some concerns. While testing the check on real infrastructure, I realized that there are two ways to add instance types: either by adding a new instance attached to the Auto Scaling group or by adding a new override to the Launch Template to include new supported instance types.

Also, for the unit tests I needed to use botocore because moto does not cover the `attach_instance` api call.

### Description

Added new check `autoscaling_group_multiple_instance_types` with its unit tests.

### Checklist

### License


prowler/providers/aws/services/autoscaling/autoscaling_group_multiple_instances_types_in_multiaz/autoscaling_group_multiple_instances_types_in_multiaz.metadata.json
```diff
@@ -0,0 +1,32 @@
+{
+  "Provider": "aws",
+  "CheckID": "autoscaling_group_multiple_instances_types_in_multiaz",
```
MrCloudSec:
```suggestion
  "CheckID": "autoscaling_group_multiple_instance_types",
```

prowler/providers/aws/services/autoscaling/autoscaling_group_multiple_instances_types_in_multiaz/autoscaling_group_multiple_instances_types_in_multiaz.py
MrCloudSec, line 1: Can we check if an AZ has more than one instance types?
For example, if there are 2 AZs and in one there are multiple instance types but in the other one only one instance type, it should fail.

MrCloudSec (changes requested):

danibarranqueroo:
Hi @sergargar!

These past few days, I’ve been working on other checks from this same service, and since they’re all merged now, I’ve finally made the changes you suggested. To do this, I needed to merge those checks into this branch and then start working on how to verify if each AZ has multiple instances. I decided to store each instance's AZ in a dictionary, using the AZ name as the key and the instance type as the value. During the check, I verify if each AZ has 2 or more values.

Since Moto still doesn’t support the `attach_instances` API call, I’ve kept the old `describe_autoscaling_groups` test, which covers the other Auto Scaling checks, and created a new test specifically for this Auto Scaling group with attached instances.

Hope this message helps you understand what I’ve done more quickly :rocket:😄

MrCloudSec (approved):

PR #5327, opened 2024-10-08
chore(aws): fix resource types to match with ASFF resources

puchy22:
### Context

Some ResourceTypes did not match with https://docs.aws.amazon.com/securityhub/latest/userguide/asff-resources.html

### Description

Changed ResourceTypes that does not match with ASFF format.

### Checklist

### License


puchy22:
After discussion it has been decided that this will not be added and that CloudFormation types will be accepted with a specific format that will be documented.

PR #5332, opened 2024-10-08
feat(s3): add test_connection method

pedrooot:
### Description

This PR adds `test_connection` method to the S3 class.

### Checklist

### License


prowler/providers/aws/lib/s3/s3.py
```diff
@@ -147,3 +148,33 @@ def send_to_bucket(
                 f"{error.__class__.__name__}[{error.__traceback__.tb_lineno}] -- {error}"
             )
         return uploaded_objects
+
+    @staticmethod
+    def test_connection(session, bucket_name: str) -> bool:
```
MrCloudSec: Please, add `raise_on_exception` and return the `Connection` object.

prowler/providers/aws/lib/s3/exceptions/exceptions.py
```diff
@@ -0,0 +1,32 @@
+from prowler.exceptions.exceptions import ProwlerException
+
+
+class S3BaseException(ProwlerException):
+    """Base class for S3 exceptions."""
+
+    S3_ERROR_CODES = {
+        (2500, "S3TestConnectionError"): {
```
jfagoagas: We need to add a new number I think.

prowler/providers/aws/lib/s3/s3.py
```diff
@@ -147,3 +150,38 @@ def send_to_bucket(
                 f"{error.__class__.__name__}[{error.__traceback__.tb_lineno}] -- {error}"
             )
         return uploaded_objects
+
+    @staticmethod
+    def test_connection(
+        session, bucket_name: str, raise_on_exception: bool = True
```
jfagoagas, line 163: I think for now is good to request a session but probably we will need to setup the same credentials as the provider.

prowler/providers/aws/lib/s3/s3.py
```diff
@@ -147,3 +150,38 @@ def send_to_bucket(
...
+    ) -> Connection:
+        """
+        Test the connection to the S3 bucket.
+
+        Parameters:
+        - session: An instance of the `Session` class representing the AWS session.
+        - bucket_name: A string representing the name of the S3 bucket.
```
jfagoagas, line 170: Are we supporting also the bucket URI here?

prowler/providers/aws/lib/s3/s3.py
```diff
@@ -147,3 +150,38 @@ def send_to_bucket(
...
+
+        Raises:
+        - Exception: An exception indicating that the connection test failed.
+        """
+        try:
+            # Set a Temp file to upload
+            with tempfile.TemporaryFile() as temp_file:
+                temp_file.write(b"Test Prowler Connection")
+                temp_file.seek(0)
+                session.upload_fileobj(
```
jfagoagas: What is the region being used in the client? https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/get_bucket_location.html

To test it use a session from a different region where the bucket belongs because a while ago if bucket does not belong to the same region than the API loads from env (`eu-west-1`) the calls return a 301, causing the process to fail.

prowler/providers/aws/lib/s3/s3.py
```diff
@@ -147,3 +150,38 @@ def send_to_bucket(
...
+                    temp_file, bucket_name, "test-prowler-connection.txt"
+                )
+
+            # Try to delete the file
+            session.delete_object(Bucket=bucket_name, Key="test-prowler-connection.txt")
+            return True
+        except Exception as error:
```
jfagoagas, line 236: We need to catch exceptions here coming from S3 and create our own.

jfagoagas (changes requested):

prowler/providers/aws/lib/s3/s3.py
pedrooot: I've checked the region from the S3 client, if it doesn't match the bucket location I create a new session for the given location.

prowler/providers/aws/lib/s3/s3.py
pedrooot: Good point! Done

prowler/providers/aws/lib/s3/s3.py
pedrooot: Nice

prowler/providers/aws/lib/s3/exceptions/exceptions.py
pedrooot: Done

prowler/providers/aws/lib/s3/s3.py
pedrooot: Done

jfagoagas (approved):
🚀

PR #5345, opened 2024-10-09
chore(sns): manage `ResourceNotFoundException` and add paralelism

puchy22:
### Context

SNS Topics list tags was not taking advantage of paralelism and was not managing exceptions correctly

### Description

Use threading call for function `list_tags_for_resource` and manage `ResourceNotFoundException` error.

### Checklist

### License

### DIFF
--- a/prowler/providers/aws/services/sns/sns_service.py
+++ b/prowler/providers/aws/services/sns/sns_service.py
@@ -61,13 +61,12 @@ class SNS(AWSService):
                 f"{regional_client.region} -- {error.__class__.__name__}[{error.__traceback__.tb_lineno}]: {error}"
             )
 
-    def _list_tags_for_resource(self, resource: any):
+    def _list_tags_for_resource(self, resource):
         logger.info("SNS - Listing Tags...")
         try:
-            if getattr(resource, "region", "") and getattr(resource, "arn", ""):
-                resource.tags = self.regional_clients[
-                    resource.region
-                ].list_tags_for_resource(ResourceArn=resource.arn)["Tags"]
+            resource.tags = self.regional_clients[
+                resource.region
+            ].list_tags_for_resource(ResourceArn=resource.arn)["Tags"]
         except ClientError as error:
             if error.response["Error"]["Code"] == "ResourceNotFoundException":
                 logger.warning(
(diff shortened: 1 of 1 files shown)


prowler/providers/aws/services/sns/sns_service.py
```diff
@@ -61,18 +61,25 @@ def _get_topic_attributes(self, regional_clients):
                 f"{regional_client.region} -- {error.__class__.__name__}[{error.__traceback__.tb_lineno}]: {error}"
             )
 
-    def _list_tags_for_resource(self):
-        logger.info("SNS - List Tags...")
+    def _list_tags_for_resource(self, resource: any):
```
jfagoagas:
```suggestion
    def _list_tags_for_resource(self, resource: Topic):
```
?

prowler/providers/aws/services/sns/sns_service.py
```diff
@@ -61,18 +61,25 @@ def _get_topic_attributes(self, regional_clients):
...
+        logger.info("SNS - Listing Tags...")
         try:
-            for topic in self.topics:
-                regional_client = self.regional_clients[topic.region]
-                response = regional_client.list_tags_for_resource(
-                    ResourceArn=topic.arn
-                )["Tags"]
-                topic.tags = response
+            if getattr(resource, "region", "") and getattr(resource, "arn", ""):
```
jfagoagas: Why this if you are expecting a `Topic`?

jfagoagas (changes requested):

prowler/providers/aws/services/sns/sns_service.py
puchy22: In the function's parameters, I defined the resource as `any` in case we want to list tags for another resource in the future. However, I believe this is more theoretical than practical. Although I defined it as `any` we control when the function is called, and it's unlikely we'd call something that doesn't have an ARN. I can remove it if you think it's unnecessary.

jfagoagas: I think for now is more descriptive if we just use the `Topic` with their attributes, although I appreciate having that generic point of doing things :D

jfagoagas (approved):
Great one, thanks for the fix!

PR #5358, opened 2024-10-10
feat(azure): add authentication method from static credentials

pedrooot:
### Description

This Pull Request (PR) adds a new method for instantiating the Azure provider exclusively using `client_id`, `tenant_id` and `client_secret`.

Also, adds new exceptions.

### Checklist

### License

### DIFF
--- a/prowler/providers/azure/azure_provider.py
+++ b/prowler/providers/azure/azure_provider.py
@@ -5,7 +5,11 @@ from os import getenv
 
 import requests
 from azure.core.exceptions import HttpResponseError
-from azure.identity import DefaultAzureCredential, InteractiveBrowserCredential
+from azure.identity import (
+    ClientSecretCredential,
+    DefaultAzureCredential,
+    InteractiveBrowserCredential,
+)
 from azure.mgmt.subscription import SubscriptionClient
 from colorama import Fore, Style
 from msgraph import GraphServiceClient
@@ -16,8 +20,10 @@ from prowler.lib.utils.utils import print_boxes
 from prowler.providers.azure.exceptions.exceptions import (
     AzureArgumentTypeValidationError,
     AzureBrowserAuthNoTenantIDError,
+    AzureConfigCredentialsError,
     AzureDefaultAzureCredentialError,
     AzureEnvironmentVariableError,
+    AzureGetTokenIdentityError,
     AzureHTTPResponseError,
     AzureInteractiveBrowserCredentialError,
     AzureNoAuthenticationMethodError,
@@ -91,6 +97,8 @@ class AzureProvider(Provider):
         subscription_ids: list = [],
         audit_config: dict = {},
         fixer_config: dict = {},
+        client_id: str = None,
+        client_secret: str = None,
     ):
         """
         Initializes the Azure provider.
@@ -105,6 +113,8 @@ class AzureProvider(Provider):
             subscription_ids (list): List of subscription IDs.
             audit_config (dict): The audit configuration for the Azure provider.
             fixer_config (dict): The fixer configuration.
+            client_id (str): The Azure client ID.
+            client_secret (str): The Azure client secret.
 
         Returns:
             None
@@ -114,6 +124,8 @@ class AzureProvider(Provider):
             AzureSetUpRegionConfigError: If there is an error in setting up the region configuration.
             AzureDefaultAzureCredentialError: If there is an error in retrieving the Azure credentials.
             AzureInteractiveBrowserCredentialError: If there is an error in retrieving the Azure credentials using browser authentication.
+            AzureConfigCredentialsError: If there is an error in configuring the Azure credentials from a dictionary.
+            AzureGetTokenIdentityError: If there is an error in getting the token from the Azure identity.
         """
         logger.info("Setting Azure provider ...")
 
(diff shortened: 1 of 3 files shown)


prowler/providers/azure/azure_provider.py
```diff
@@ -114,26 +124,43 @@ def __init__(
...
+            tenant_id,
+            client_id,
+            client_secret,
         )
 
         logger.info("Checking if region is different than default one")
         self._region_config = self.setup_region_config(region)
 
+        # Get the dict from the static credentials
+        azure_config = None
```
jfagoagas: Could we name this `credentials` or `azure_credentials`?

prowler/providers/azure/azure_provider.py
```diff
@@ -114,26 +124,43 @@ def __init__(
...
+        if tenant_id and client_id and client_secret:
+            azure_config = self.get_statics_credentials(
```
jfagoagas: Since you are not doing anything but creating a `dict` I think the function is not needed nor the `if`.

prowler/providers/azure/azure_provider.py
```diff
@@ -246,30 +276,40 @@ def validate_arguments(
...
+                    message="Azure provider requires at least one authentication method set: [--az-cli-auth | --sp-env-auth | --browser-auth | --managed-identity-auth]",
+                )
+            elif browser_auth and not tenant_id:
+                raise AzureBrowserAuthNoTenantIDError(
+                    file=os.path.basename(__file__),
+                    message="Azure Tenant ID (--tenant-id) is required for browser authentication mode",
+                )
+        else:
+            if not tenant_id:
+                raise AzureTenantIDNoBrowserAuthError(
```
jfagoagas: I'd create a different error here since we are not using browser auth.

prowler/providers/azure/azure_provider.py
```diff
@@ -378,21 +423,40 @@ def setup_session(
...
-                    exclude_visual_studio_code_credential=True,
-                    # Azure Auth using Shared Token Cache is not supported
-                    exclude_shared_token_cache_credential=True,
-                    # Azure Auth using PowerShell is not supported
-                    exclude_powershell_credential=True,
-                    # set Authority of a Microsoft Entra endpoint
-                    authority=region_config.authority,
-                )
+                if azure_config:
+                    try:
```
jfagoagas, line 440: Why another `try`?

prowler/providers/azure/azure_provider.py
```diff
@@ -346,6 +386,7 @@ def setup_session(
         browser_auth: bool,
         managed_identity_auth: bool,
         tenant_id: str,
+        azure_config: dict,
```
jfagoagas: I think it'd be simpler if you just add three new arguments here and the pass them using the `**dict` so you can unpack it into arguments. But this is up to you, nothing super important.

prowler/providers/azure/azure_provider.py
```diff
@@ -378,21 +423,40 @@ def setup_session(
...
-                )
+                if azure_config:
+                    try:
+                        credentials = ClientSecretCredential(
+                            tenant_id=azure_config["tenant_id"],
+                            client_id=azure_config["client_id"],
+                            client_secret=azure_config["client_secret"],
+                        )
+                        return credentials
+                    except Exception as error:
```
jfagoagas: We need to catch the possible exceptions from the `ClientSecretCredential`, if any.

prowler/providers/azure/azure_provider.py
```diff
@@ -378,21 +423,40 @@ def setup_session(
...
+                        )
+                        logger.critical(
+                            f"{error.__class__.__name__}[{error.__traceback__.tb_lineno}] -- {error}"
+                        )
+                        raise AzureConfigCredentialsError(
+                            file=os.path.basename(__file__), original_exception=error
+                        )
+                else:
+                    # Since the input vars come as True when it is wanted to be used, we need to inverse it since
+                    # DefaultAzureCredential sets the auth method excluding the others
```
jfagoagas:
```suggestion
                    # Since the authentication method to be used will come as True, we have to negate it since
                    # DefaultAzureCredential sets just one authentication method, excluding the others
```

prowler/providers/azure/exceptions/exceptions.py
```diff
@@ -147,3 +155,17 @@ def __init__(self, file=None, original_exception=None, message=None):
         super().__init__(
             1924, file=file, original_exception=original_exception, message=message
         )
+
+
+class AzureConfigCredentialsError(AzureCredentialsError):
```
jfagoagas: Can we create more? Like testing for:
- Invalid `client_id`
- Invalid `client_secret`
- Both valid but not belonging to `tenant_id`

jfagoagas (changes requested):

prowler/providers/azure/azure_provider.py

prowler/providers/azure/azure_provider.py
pedrooot: Sure! Done

prowler/providers/azure/azure_provider.py
pedrooot: To handle the specific exceptions for the method from inside

prowler/providers/azure/azure_provider.py

prowler/providers/azure/azure_provider.py

prowler/providers/azure/exceptions/exceptions.py
pedrooot: Done

prowler/providers/azure/azure_provider.py
pedrooot: Done!

jfagoagas (approved):
🤯 Azure always requires a ton, great one 👏

PR #5367, opened 2024-10-10
feat(opensearch): add new check `opensearch_service_domains_at_least_three_master_nodes`

MarioRgzLpz:
### Context

Add a new check to ensure `Elasticsearch/Opensearch` domains are configured with at least three dedicated primary nodes. This control fails if the domain does not use dedicated primary nodes.  

A domain with fewer than three master nodes is vulnerable to failures during resource strain, such as during blue/green deployments. While using five dedicated master nodes may improve availability, three nodes are generally sufficient to mitigate availability risks without incurring additional unnecessary costs. Having at least three dedicated master nodes ensures sufficient capacity and reliable cluster operations in case of node failure.

### Description

Add check `opensearch_service_domains_at_least_three_master_nodes` with respective unit test and metadata. Add new attribute `dedicated_master_count` to `Domain` model in the `opensearch_service`. 

### Checklist

### License

### DIFF
--- a/prowler/providers/aws/services/opensearch/opensearch_service.py
+++ b/prowler/providers/aws/services/opensearch/opensearch_service.py
@@ -13,141 +13,141 @@ class OpenSearchService(AWSService):
     def __init__(self, provider):
         # Call AWSService's __init__
         super().__init__("opensearch", provider)
-        self.opensearch_domains = []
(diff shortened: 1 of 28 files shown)


MrCloudSec:
Implemented by https://github.com/prowler-cloud/prowler/pull/5393

PR #5368, opened 2024-10-10
feat(color): add `--no-color` flag

### LINKED ISSUE #5294: Add flag to disable color output
### New feature motivation

When running prowler in some automated environments (e.g. Jenkins) it's too hard to read the tool output because of ANSI escape symbols. If there were a flag, e.g. `--no-color` to disable this behavior it would be great!
image

### Solution Proposed

Add a flag to disable the color output, e.g. `--no-color` 

There is a Jenkins AnsiColor plugin https://plugins.jenkins.io/ansicolor/

### Additional context

_No response_

MrSecure:
### Context

Fix #5294 

### Description

- Import colorama's `init`
- Add `--no-color` command line option (default `False`)
- Use `no_color` option to control colorama init.

### Checklist

  - Tests added to the parser testing library

### License

### DIFF
--- a/prowler/__main__.py
+++ b/prowler/__main__.py
@@ -115,8 +115,6 @@ def prowler():
 
     if args.no_color:
         colorama_init(strip=True)
-    else:
-        colorama_init(strip=False)
 
     if not args.no_banner:
         legend = args.verbose or getattr(args, "fixer", None)
--- a/prowler/__main__.py
+++ b/prowler/__main__.py
@@ -5,6 +5,7 @@ import sys
 from os import environ
 
 from colorama import Fore, Style
+from colorama import init as colorama_init
 
 from prowler.config.config import (
     csv_file_suffix,
@@ -111,6 +112,11 @@ def prowler():
         and not checks_file
         and not checks_folder
     )
+
+    if args.no_color:
+        colorama_init(strip=True)
+    else:
+        colorama_init(strip=False)
 
     if not args.no_banner:
         legend = args.verbose or getattr(args, "fixer", None)
--- a/docs/tutorials/misc.md
+++ b/docs/tutorials/misc.md
@@ -27,7 +27,7 @@ prowler <provider> -b/--no-banner
 ## Disable Colors
 Prowler can run without showing colors:
 ```console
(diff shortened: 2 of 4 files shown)


MrCloudSec:
Hi @MrSecure, thanks a lot for your contribution! We will review this asap.

MrSecure:
Should the `--only-logs` flag set `--no-color` in addition to `--no-banner` ?
Or does that complicate things since `--only-logs` is slated for removal in v5?

jfagoagas:
> Should the `--only-logs` flag set `--no-color` in addition to `--no-banner` ? Or does that complicate things since `--only-logs` is slated for removal in v5?

The `--only-logs` flag removes all the pure CLI output so there is no need to set also `--no-color` and it is going to be deprecated in v5 since it is no longer needed because Prowler is usable as an SDK.

Regarding the `--no-banner`, I think they should work independently.

prowler/__main__.py
```diff
@@ -112,6 +113,11 @@ def prowler():
         and not checks_folder
     )
 
+    if args.no_color:
+        colorama_init(strip=True)
+    else:
+        colorama_init(strip=False)
```
pedrooot:
```suggestion
```
Is this needed?

pedrooot (changes requested):
Thanks for your contribution! Please review my comments 🙏🏼

prowler/__main__.py
MrSecure: There's a good chance it's not strictly required. However, as the default is 'None' with an OS-dependent behavior I thought it was wise to keep.  Reviewing the OS-dependent behavior, I agree, this should be remove.

MrSecure: fixed.

pedrooot: Thanks! I'll add to your PR info about this flag inside the docs, thanks for your contribution!

jfagoagas:
Hello @MrSecure @pedrooot I've just removed the `-Q` option since for now we only want to add long flags to the CLI. 

Thank you!

pedrooot (approved):
🔝

toniblyx:
Thank you @MrSecure :D

PR #5370, opened 2024-10-10
feat(k8s): Add kubeconfig content static authentication

MrCloudSec:
### Description

Add `kubeconfig_content` static authentication. Also, add the necessary testing and fix the exceptions.

### Checklist

### License

### DIFF
--- a/prowler/providers/kubernetes/kubernetes_provider.py
+++ b/prowler/providers/kubernetes/kubernetes_provider.py
@@ -207,6 +207,7 @@ class KubernetesProvider(Provider):
     def test_connection(
         kubeconfig_file: str = "~/.kube/config",
         kubeconfig_content: dict = None,
+        namespace: str = None,
         input_context: str = "",
         raise_on_exception: bool = True,
     ) -> Connection:
@@ -216,6 +217,7 @@ class KubernetesProvider(Provider):
         Args:
             kubeconfig_file (str): Path to the kubeconfig file.
             kubeconfig_content (dict): Content of the kubeconfig file.
+            namespace (str): Namespace name.
             input_context (str): Context name.
             raise_on_exception (bool): Whether to raise an exception on error.
         Returns:
@@ -225,7 +227,12 @@ class KubernetesProvider(Provider):
             KubernetesProvider.setup_session(
                 kubeconfig_file, kubeconfig_content, input_context
             )
-            client.CoreV1Api().list_namespace(timeout_seconds=2, _request_timeout=2)
+            if namespace:
+                client.CoreV1Api().list_namespaced_pod(
+                    namespace, timeout_seconds=2, _request_timeout=2
+                )
+            else:
+                client.CoreV1Api().list_namespace(timeout_seconds=2, _request_timeout=2)
             return Connection(is_connected=True)
         except KubernetesSetUpSessionError as setup_session_error:
             logger.critical(
--- a/prowler/providers/kubernetes/kubernetes_provider.py
+++ b/prowler/providers/kubernetes/kubernetes_provider.py
@@ -4,6 +4,7 @@ from colorama import Fore, Style
 from kubernetes.client.exceptions import ApiException
 from kubernetes.config.config_exception import ConfigException
 from requests.exceptions import Timeout
+from yaml import safe_load
 
 from kubernetes import client, config
 from prowler.config.config import get_default_mute_file_path
@@ -15,6 +16,7 @@ from prowler.providers.kubernetes.exceptions.exceptions import (
     KubernetesAPIError,
     KubernetesCloudResourceManagerAPINotUsedError,
     KubernetesError,
+    KubernetesInvalidKubeConfigFileError,
     KubernetesSetUpSessionError,
     KubernetesTimeoutError,
 )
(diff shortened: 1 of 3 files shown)


prowler/providers/kubernetes/kubernetes_provider.py
```diff
@@ -42,19 +42,21 @@ def __init__(
         namespace: list = None,
         audit_config: dict = {},
         fixer_config: dict = {},
+        kubeconfig_content: dict = {},
```
```suggestion
        kubeconfig_content: dict = None,
```

prowler/providers/kubernetes/kubernetes_provider.py
```diff
@@ -139,51 +141,72 @@ def get_output_mapping(self):
         }
 
     @staticmethod
-    def setup_session(kubeconfig_file, input_context) -> KubernetesSession:
+    def setup_session(
+        kubeconfig_file: str = None,
+        kubeconfig_content: dict = {},
```
```suggestion
        kubeconfig_content: dict = None,
```

prowler/providers/kubernetes/kubernetes_provider.py
```diff
@@ -139,51 +141,72 @@ def get_output_mapping(self):
...
-                    "user": "service-account-name",  # Also a placeholder
-                },
-            }
-        else:
+            if kubeconfig_content:
+                config.load_kube_config_from_dict(
+                    kubeconfig_content, context=input_context
+                )
+            else:
+                try:
```
jfagoagas, line 169: Why do you need the double `try/except`?

prowler/providers/kubernetes/kubernetes_provider.py
```diff
@@ -139,51 +141,72 @@ def get_output_mapping(self):
...
                         context = context_item
             else:
                 context = config.list_kube_config_contexts()[1]
-        return KubernetesSession(api_client=client.ApiClient(), context=context)
+            return KubernetesSession(api_client=client.ApiClient(), context=context)
+        except Exception as error:
+            logger.critical(
+                f"{error.__class__.__name__}[{error.__traceback__.tb_lineno}]: {error}"
+            )
+            raise KubernetesError(
```
```suggestion
            raise KubernetesSetUpSessionError(
```

Could you please use the exception created for the setup session?

prowler/providers/kubernetes/kubernetes_provider.py
```diff
@@ -139,51 +141,72 @@ def get_output_mapping(self):
...
+                original_exception=error, file=os.path.abspath(__file__)
+            )
 
     @staticmethod
     def test_connection(
         kubeconfig_file: str = "~/.kube/config",
+        kubeconfig_content: dict = {},
```
```suggestion
        kubeconfig_content: dict = None,
```

prowler/providers/kubernetes/kubernetes_provider.py
```diff
@@ -192,13 +215,16 @@ def test_connection(
...
             Connection: A Connection object.
         """
         try:
-            KubernetesProvider.setup_session(kubeconfig_file, input_context)
+            KubernetesProvider.setup_session(
+                kubeconfig_file, kubeconfig_content, input_context
+            )
             client.CoreV1Api().list_namespace(timeout_seconds=2, _request_timeout=2)
             return Connection(is_connected=True)
         except ApiException as api_error:
```
jfagoagas, line 255: You need to catch the `KubernetesSetUpSessionError` here and add the following:
```python
except KubernetesSetUpSessionError as setup_session_error:
            logger.critical(
                f"KubernetesSetUpSessionError[{setup_session_error.__traceback__.tb_lineno}]: {setup_session_error}"
            )
            if raise_on_exception:
                raise setup_session_error
            return Connection(error=setup_session_error)
```

prowler/providers/kubernetes/kubernetes_provider.py
```diff
@@ -192,13 +215,16 @@ def test_connection(
...
+            raise_on_exception (bool): Whether to raise an exception on error.
         Returns:
             Connection: A Connection object.
         """
         try:
-            KubernetesProvider.setup_session(kubeconfig_file, input_context)
+            KubernetesProvider.setup_session(
+                kubeconfig_file, kubeconfig_content, input_context
+            )
             client.CoreV1Api().list_namespace(timeout_seconds=2, _request_timeout=2)
```
jfagoagas: Would we need to add an optional `namespace` here for validation purposes?

jfagoagas (changes requested):

prowler/providers/kubernetes/kubernetes_provider.py
MrCloudSec: Listing namespaces is a simple and non-invasive Kubernetes API call. It confirms that the Kubernetes client (and thus, the user or service account) can establish a connection to the cluster’s API server, which is the primary indicator of a valid session.

jfagoagas: That's great but my point was focused on adding an specific `namespace` as an optional validation for the client, since we allow that argument to be used.

prowler/providers/kubernetes/kubernetes_provider.py
MrCloudSec: We were not handling errors in this method.

jfagoagas (approved):

PR #5379, opened 2024-10-11
chore(lambda): update obsolete lambda runtime

jbro-tp:
### Context

Adding additional lambda runtimes to check for in AWS.

https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html#runtimes-deprecated

### Description

No fix, just providing more criteria to check.

### Checklist

### License


prowler/config/config.yaml
```diff
@@ -99,7 +99,9 @@ aws:
       "nodejs10.x",
       "nodejs12.x",
       "nodejs14.x",
+      "nodejs16.x",
       "dotnet5.0",
+      "dotnet7.0",
```
MrCloudSec:
```suggestion
      "dotnet7",
```
The identifier for AWS is like that as per https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html#runtimes-deprecated.

MrCloudSec (changes requested):
Thanks for the update @jonathanbro ! Please, review my comment.

jbro-tp:
I've committed your suggestion @sergargar :)

MrCloudSec (approved):

PR #5381, opened 2024-10-11
feat(gcp): add provider id validation inside test_connection

pedrooot:
### Description

This PR adds a validation inside test_connection in order to check that the `provider_id` provided is reachable for the given credentials.

### Checklist

### License

### DIFF
--- a/prowler/providers/gcp/exceptions/exceptions.py
+++ b/prowler/providers/gcp/exceptions/exceptions.py
@@ -37,6 +37,10 @@ class GCPBaseException(ProwlerException):
             "message": "Error loading static credentials",
             "remediation": "Check the credentials and ensure they are properly set up. client_id, client_secret and refresh_token are required.",
         },
+        (1933, "GCPNotValidProviderIdError"): {
+            "message": "Provider does not match with the expected project_id",
+            "remediation": "Check the provider and ensure it matches the expected project_id.",
+        },
     }
 
     def __init__(self, code, file=None, original_exception=None, message=None):
@@ -113,4 +117,11 @@ class GCPStaticCredentialsError(GCPCredentialsError):
     def __init__(self, file=None, original_exception=None, message=None):
         super().__init__(
             1932, file=file, original_exception=original_exception, message=message
+        )
+
+
+class GCPNotValidProviderIdError(GCPBaseException):
+    def __init__(self, file=None, original_exception=None, message=None):
+        super().__init__(
+            1933, file=file, original_exception=original_exception, message=message
         )
--- a/prowler/providers/gcp/exceptions/exceptions.py
+++ b/prowler/providers/gcp/exceptions/exceptions.py
@@ -37,7 +37,7 @@ class GCPBaseException(ProwlerException):
             "message": "Error loading static credentials",
             "remediation": "Check the credentials and ensure they are properly set up. client_id, client_secret and refresh_token are required.",
         },
-        (1933, "GCPNotValidProviderIdError"): {
+        (1933, "GCPInvalidAccountCredentials"): {
             "message": "Provider does not match with the expected project_id",
             "remediation": "Check the provider and ensure it matches the expected project_id.",
         },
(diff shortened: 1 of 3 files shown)


prowler/providers/gcp/exceptions/exceptions.py
```diff
@@ -37,6 +37,10 @@ class GCPBaseException(ProwlerException):
             "message": "Error loading static credentials",
             "remediation": "Check the credentials and ensure they are properly set up. client_id, client_secret and refresh_token are required.",
         },
+        (1933, "GCPNotValidProviderIdError"): {
```
jfagoagas:
```suggestion
        (1933, "GCPInvalidAccountCredentials"): {
```

The point here is to validate that the provided credentials does not match with the `project_id` that wants to be used.

prowler/providers/gcp/gcp_provider.py
```diff
@@ -528,3 +550,28 @@ def validate_static_arguments(
...
             "type": "authorized_user",
         }
+
+    @staticmethod
+    def validate_project_id(provider_id: str, credentials: str = None) -> bool:
+        """
+        Validate the provider ID
+
+        Args:
+            provider_id: str
```
jfagoagas, line 559: Please improve the docstring with:
- Better explanation of what the function does
- Args and their description

prowler/providers/gcp/gcp_provider.py
```diff
@@ -316,9 +324,16 @@ def test_connection(
...
                 )
 
-            session, _ = GcpProvider.setup_session(
+            session, project_id = GcpProvider.setup_session(
                 credentials_file, service_account, gcp_credentials
             )
+
+            if provider_id and project_id != provider_id:
+                # Logic to check if the provider ID matches the project ID
+                GcpProvider.validate_project_id(
```
jfagoagas, line 332: I think it'll be easier if you include the code within the function, so it does not match raise Exception and if matches return `connnected = True`. What do you think?

prowler/providers/gcp/gcp_provider.py
jfagoagas: Nothing to do if you remove it.

prowler/providers/gcp/gcp_provider.py
```diff
@@ -316,9 +324,16 @@ def test_connection(
                     client_id, client_secret, refresh_token
                 )
 
-            session, _ = GcpProvider.setup_session(
+            session, project_id = GcpProvider.setup_session(
                 credentials_file, service_account, gcp_credentials
             )
+
+            if provider_id and project_id != provider_id:
```
jfagoagas, line 330:
```suggestion
            if provider_id and project_id != provider_id and provider_id not in GcpProvider.get_projects(credentials=credentials):
```

but my question is why do we need to get all the projects if the setup session returns the project id.

jfagoagas (changes requested):
Thanks @pedrooot, I left some questions/changes. How can we add tests for this code?

prowler/providers/gcp/exceptions/exceptions.py
pedrooot: Done!

prowler/providers/gcp/gcp_provider.py
pedrooot: DOne!

prowler/providers/gcp/gcp_provider.py
pedrooot: The setup_session returns the default project_id, could happen that the given one doesn't match with the default

prowler/providers/gcp/gcp_provider.py
pedrooot: I would prefer to modularize the code into the `validate_project_id` method, this way I can document what the method does in order to clarify the usage for the user.

jfagoagas (approved):
Great one 👏

PR #5389, opened 2024-10-14
refactor(WAF): Rename `WAF` to `WAFRegional` and Add Global `WAF` Service

HugoPBrito:
### Context

I identified a misconception in how the AWS WAF services and clients were referenced and utilized across our services. Specifically, `AWS WAFRegional` was being referred to as simply `WAF`, which led to confusion. Additionally, `AWS WAF` (global) was not represented separately, which could cause potential issues when differentiating between global and regional WAF services.

### Description

This PR addresses the following:

- **Renaming**: The existing `WAF` service and client have been renamed to `WAFRegional` to align with AWS’s naming convention for regional WAF services.
- **Introducing Global WAF**: A new `WAF` service and client have been introduced to represent `AWS WAF` (global), ensuring that both regional and global WAF services are distinctly covered.
This change mirrors the structure used by AWS and is similar approach to how we handle`S3` and `S3Control` services under the same S3 service file with different client files.

### Checklist

- Are there new checks included in this PR?  No.

### License


MrCloudSec (changes requested):
Good catch!
Please, change the waf client that is used in check `elbv2_waf_acl_attached` to the new regional one.

HugoPBrito:
> Good catch! Please, change the waf client that is used in check `elbv2_waf_acl_attached` to the new regional one.

Done! I also updated the new client name to avoid any possible confusions.

MrCloudSec (approved):

PR #5391, opened 2024-10-14
feat(azure): add provider id validation inside test_connection

pedrooot:
### Description
Add provider validation inside the test_connection method in order to ensure that the given provider_id it's a valid subscription_id

### Checklist

### License


prowler/providers/azure/exceptions/exceptions.py
```diff
@@ -97,6 +97,10 @@ class AzureBaseException(ProwlerException):
             "message": "The provided tenant ID and client ID do not belong to the provided client secret",
             "remediation": "Check the tenant ID and client ID and ensure they belong to the provided client secret.",
         },
+        (1937, "AzureInvalidAccountCredentialsError"): {
+            "message": "The provided provider_id does not math with the valid subscriptions",
```
jfagoagas:
```suggestion
            "message": "The provided provider_id does not match with the available subscriptions",
```

jfagoagas (approved):

PR #5453, opened 2024-10-17
fix(bedrock): add filtering and handle different ARNs

MrCloudSec:
### Description

As per https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/iam-access-control-overview-cwl.html, Log Groups can have two different ARNs, so we need to check both types.
Also, add the resource filtering to Bedrock Guardrails.

### Checklist

### License


puchy22 (approved):
👍

PR #5465, opened 2024-10-18
feat(waf): add new check `waf_global_rule_with_conditions`

HugoPBrito:
### Context

`AWS WAF Classic global rules` are used to manage and control web traffic by specifying conditions that determine how the firewall handles requests. These conditions help identify, inspect, and mitigate potential security threats, such as malicious or suspicious traffic, thereby improving the overall security posture. Without any conditions in place, web traffic flows freely, which can lead to vulnerabilities being exploited, bypassing security policies.

### Description

This check ensures that `AWS WAF Classic global rules` contain at least one condition. If no conditions are present, the rule becomes ineffective, allowing all traffic to pass unchecked.

### Checklist

- Are there new checks included in this PR? Yes.

### License


MrCloudSec (changes requested):
I have made necessary changes in the PR @HugoPBrito, but the call `_list_resources_for_web_acl` does not exist for the WAF API. Please, take a look!

MrCloudSec (approved):

PR #5472, opened 2024-10-18
chore: vendorize `detect-secrets` dependency

MrCloudSec:
### Context
This PR resolves the PyPI upload issue caused by the use of a direct GitHub dependency for `detect-secrets`, which is not allowed by PyPI.

### Description
- Vendorized `detect-secrets`: The library has been added to `prowler/lib/detect-secrets` to include the latest changes from the master branch, as the PyPI version is outdated.
- Updated `pyproject.toml`: The dependency is now referenced locally from `prowler/lib/detect-secrets` instead of GitHub.

### Checklist

### License


MrCloudSec:
This approach was discarded.

PR #5493, opened 2024-10-21
feat(aws): add new check `dynamodb_accelerator_cluster_multi_az`

sansns:
### Context

Add DynamoDB DAX nodes existing in multiple AZ check

### Description

Checks(Modified and Added):
Check if DAX nodes exist in multiple AZs
Fix DynamoDB Table KMS Encryption verbiage. Previously it stated the correct FAIL or PASS, but it stated wrong encryption type for the table.
```
DynamoDB table {table.name} does have DEFAULT encryption enabled.
```
```
DynamoDB table {table.name} is using DEFAULT encryption.
```

Tests(Modified and Added):
Add DAX Multi AZ checks
Remove Search function from two tests

### Checklist

    - If so, do we need to update permissions for the provider? Please review this carefully. No

### License


MrCloudSec (approved):
Thanks for this new check @sansns !

PR #5524, opened 2024-10-24
feat(gcp): add `--organization-id` flag

### LINKED ISSUE #5478: Add "--organization-id" argument for GCP assessments
### New feature motivation

For GCP, the easiest way to run prowler is in the Cloud Shell where you are already authenticated to Google Cloud and prowler will use those creds. But if that account has access to multiple organizations then prowler, by default, will scan all projects from all those organizations given no `--project-id` filter. But what if those organizations contain hundreds/thousands of projects and you want  to run prowler just for one of them? Then you will end up with a super long command line containing hundreds/thousands of project ids.

### Solution Proposed

It would be useful to have an argument like `--organization-id` that prowler can use to then list & scan all projects in the given organization id (and accessible by the credentials used).

Getting a list of project ids for an organization with the following command and pass that list to prowler.
```
gcloud asset search-all-resources --asset-types="cloudresourcemanager.googleapis.com/Project" --scope=organizations/123456789 --format=json | jq -r '.[].additionalAttributes.projectId'
```

or with python:
```
from google.cloud.asset_v1 import AssetServiceClient, SearchAllResourcesRequest

def get_projects_under_org(org_id):
    client = AssetServiceClient()
    request = SearchAllResourcesRequest(
        scope = org_id, 
        asset_types = ["cloudresourcemanager.googleapis.com/Project"],
        read_mask = "additionalAttributes"
    )
    response = client.search_all_resources(request=request)
    projects = []
    for item in response:
        projects.append(item.additional_attributes.get("projectId"))
    return projects
```
### Additional context

_No response_

MrCloudSec:
### Context

Fix #5478 

### Description

Add new flag `--organization-id` flag to scan only GCP projects from a specific organization.

### Checklist

### License

### DIFF
--- a/docs/tutorials/gcp/organization.md
+++ b/docs/tutorials/gcp/organization.md
@@ -1,9 +1,12 @@
 # GCP Organization
 
-By default, Prowler will scan all the Google Cloud projects that the authenticated user has access to.
+By default, Prowler scans all Google Cloud projects accessible to the authenticated user.
 
-If you want to scan only projects from a specific organization, you can use the `--organization-id` argument.
+To limit the scan to projects within a specific Google Cloud organization, use the `--organization-id` option with the GCP organization ID:
 
 ```console
 prowler gcp --organization-id organization-id
 ```
+
+???+ note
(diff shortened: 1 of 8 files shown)


jfagoagas (approved):
Great addition! Thanks @sergargar

PR #5543, opened 2024-10-28
fix: added s3 origin comprobation in `cloudfront_distributions_s3_origin_non_existent_bucket`

### LINKED ISSUE #5532: False positives for cloudfront_distributions_s3_origin_non_existent_bucket because not all options are considered
### Steps to Reproduce

Ahoy! I am running prowler in version `4.4.0` and have encountered a lot of false positives for the check `cloudfront_distributions_s3_origin_non_existent_bucket`. 

According to the checks' [code](https://github.com/prowler-cloud/prowler/blob/84ce7a8b52672260b5f7efeb6d4e76a5b61b84a4/prowler/providers/aws/services/cloudfront/cloudfront_distributions_s3_origin_non_existent_bucket/cloudfront_distributions_s3_origin_non_existent_bucket.py#L8), the `origin` is received by a split on `domain_name` and it is just assumed that it has to be an S3 bucket. 

```python
            for origin in distribution.origins:
                bucket_name = origin.domain_name.split(".")[0]
                if not s3_client._head_bucket(bucket_name):
                    non_existent_buckets.append(bucket_name)
```

This is not correct. An origin for a CloudFront distribution resource could also be an ELB, ApiGatetway and even more. The various options of origins are listed in this documentation:
[Use various origins with CloudFront distributions](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/DownloadDistS3AndCustomOrigins.html)

So what happens? If you use ELB for all of your CloudFront Distributions, the check will fail for each resource resulting in a lot of false positives.. 

The command i am running is: 

```bash
prowler aws --log-level WARNING --output-filename "prowler-output-cloudfront" --checks cloudfront_distributions_s3_origin_non_existent_bucket --output-modes csv json-asff json-ocsf --ignore-exit-code-3 --no-banner
```

### Expected behavior

The check `cloudfront_distributions_s3_origin_non_existent_bucket` should validate the origin-type of a CloudFront distribution resource. 

Example: Let's assume the following `domain_name` of an origin: `important-api-12345.production.company.org`.
According to the code

```python
bucket_name = origin.domain_name.split(".")[0]
```

The extracted name would be `important-api-12345` . Since the origin is an ELB, the check would fail because it is just looking for S3 buckets. The check should therefore consider all other documented origin-types. If no S3, ApiGateway, ELB etc. does match the domain's name, the check should fail with something like "No valid resource for configured origin of CloudFront distribution ABCDEFG found".. Or similar.. 

Not applicable

### Environment Resource

Docker container locally

### OS used

Debian GNU/Linux 12 (bookworm)

### Prowler version

4.4.0

### Pip version

24.2

### Context

_No response_

HugoPBrito:
### Context

In the `cloudfront_distributions_s3_origin_non_existent_bucket` check, we initially didn't verify if the origin was a `S3 bucket`, which resulted in numerous false positives.

It fixes issue: fix #5532

### Description

A comprobation has been added to confirm that the origin is a `S3 bucket`.

### Checklist

### License

### DIFF
--- a/prowler/providers/aws/services/cloudfront/cloudfront_distributions_s3_origin_non_existent_bucket/cloudfront_distributions_s3_origin_non_existent_bucket.metadata.json
+++ b/prowler/providers/aws/services/cloudfront/cloudfront_distributions_s3_origin_non_existent_bucket/cloudfront_distributions_s3_origin_non_existent_bucket.metadata.json
@@ -1,7 +1,7 @@
 {
   "Provider": "aws",
   "CheckID": "cloudfront_distributions_s3_origin_non_existent_bucket",
-  "CheckTitle": "CloudFront distributions should not point to non-existent S3 origins.",
+  "CheckTitle": "CloudFront distributions should not point to non-existent S3 origins without static website hosting.",
   "CheckType": [
     "Software and Configuration Checks/Industry and Regulatory Standards/NIST 800-53 Controls"
   ],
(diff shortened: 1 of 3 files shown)


prowler/providers/aws/services/cloudfront/cloudfront_distributions_s3_origin_non_existent_bucket/cloudfront_distributions_s3_origin_non_existent_bucket.metadata.json
```diff
@@ -1,7 +1,7 @@
 {
   "Provider": "aws",
   "CheckID": "cloudfront_distributions_s3_origin_non_existent_bucket",
-  "CheckTitle": "CloudFront distributions should not point to non-existent S3 origins.",
+  "CheckTitle": "CloudFront distributions should not point to non-existent S3 origins without static website hosting.",
```
sergargar, line 4: How are we checking that the S3 bucket does not have static website hosting?

MrCloudSec (approved):
Good catch !!

PR #5035, opened 2024-09-13
chore(aws): handle NotAction cases in IAM policies

MrCloudSec:
### Context

Thanks to @Chan9390, we discovered that policies allowing invalid actions in NotAction statements grant Administrator access:

```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Sid": "Statement1",
            "Effect": "Allow",
            "NotAction": "prowler:action",
            "Resource": "*"
        }
    ]
}
```

And AWS does not have a validation yet to prevent this policies to be added.

### Description

Handle NotAction cases to avoid false negatives as well as when the action is invalid in checks:
 - `iam_policy_allows_privilege_escalation`
 - `iam_inline_policy_allows_privilege_escalation`
 - `iam_aws_attached_policy_no_administrative_privileges`
 - `iam_customer_attached_policy_no_administrative_privileges`
 - `iam_customer_unattached_policy_no_administrative_privileges`
 - `iam_inline_policy_no_administrative_privileges`

### Checklist

- Are there new checks included in this PR? Yes / No
    - If so, do we need to update permissions for the provider? Please review this carefully.
- [x] Review if the code is being covered by tests.
- [x] Review if code is being documented following this specification https://github.com/google/styleguide/blob/gh-pages/pyguide.md#38-comments-and-docstrings
- [x] Review if backport is needed.

### License

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #5560, opened 2024-10-29
fix(aws): findings in IAM policies were not reported

kagahd:
### Context

The refactoring done in PR #5035 introduced a bug so that findings in IAM policies were not reported anymore. Furthermore, the corresponding unit tests did not cover theses cases.
These bugs were first released in prowler v4.4.0 and are still present.

### Description

The bug consists of a wrong `break` statement which ends the loop as soon as the first finding was found. All the following IAM policies were wrongly disregarded.
Furthermore, the report is not added to the findings because this is done within the `for` loop which the `break` statement left already.
I fixed this by removing the `break` statement. 
I also added code to the unit tests to cover these cases.

### Checklist

### License


jfagoagas:
Hey @kagahd, thanks for this fix. We will review it as soon as possible.

MrCloudSec (approved):
Great catch @kagahd, thanks!

PR #5561, opened 2024-10-29
chore(azure): deprecate `AzureGermanCloud`

puchy22:
### Context

The Germany Dedicated zone is no longer supported as an authority host, as it was phased out in favor of the public cloud, as indicated in [this Microsoft post](https://news.microsoft.com/europe/2018/08/31/microsoft-to-deliver-cloud-services-from-new-datacentres-in-germany-in-2019-to-meet-evolving-customer-needs/).

The decision to deprecate it now follows a recent discussion in the [official Microsoft SDK repository](https://github.com/Azure/azure-sdk-for-python/pull/37654#issuecomment-2386968839), where they suggested issuing a warning instead of immediately deprecating the feature. Since it was not functioning in our code due to the non-existence of the referenced endpoint, we have decided to remove all code related to this region.

### Description

- Update version of `azure-identity`
- Remove all code related with `AzureGermanCloud`

### Checklist

### License

### DIFF
--- a/docs/tutorials/azure/use-non-default-cloud.md
+++ b/docs/tutorials/azure/use-non-default-cloud.md
@@ -7,7 +7,6 @@ At the time of writing this documentation the available Azure Clouds from differ
 - AzureCloud
 - AzureChinaCloud
 - AzureUSGovernment
-- AzureGermanCloud
 
 If you want to change the default one you must include the flag `--azure-region`, i.e.:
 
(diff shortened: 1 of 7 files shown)


jfagoagas (approved):
Approving this since it is not a breaking change in Prowler because the Germany Azure region was deprecated even before having Azure as a provider in Prowler.

Thank you for taking care of this @puchy22 👏

PR #5571, opened 2024-10-29
fix(gcp): enforce correct severity levels in CloudSQL PostgreSQL `log_min_messages`

MrCloudSec:
### Context
This update aligns the `cloudsql_instance_postgres_log_min_messages_flag` check with CIS GCP framework recommendations, which specify that log_min_messages should be set to ERROR or higher (e.g., ERROR, LOG, FATAL, PANIC). Configurations below ERROR do not meet the recommended standards.

Description:
The check now fails only if `log_min_messages` is set to any level below ERROR, ensuring compliance and accuracy in flagging only non-compliant configurations.

### Checklist

### License

### DIFF
--- a/prowler/providers/gcp/services/cloudsql/cloudsql_instance_postgres_log_min_messages_flag/cloudsql_instance_postgres_log_min_messages_flag.py
+++ b/prowler/providers/gcp/services/cloudsql/cloudsql_instance_postgres_log_min_messages_flag/cloudsql_instance_postgres_log_min_messages_flag.py
@@ -4,7 +4,17 @@ from prowler.providers.gcp.services.cloudsql.cloudsql_client import cloudsql_cli
 
 class cloudsql_instance_postgres_log_min_messages_flag(Check):
     def execute(self) -> Check_Report_GCP:
-        desired_log_min_messages = "error"
+        failing_log_levels = [
+            "DEBUG5",
+            "DEBUG4",
+            "DEBUG3",
+            "DEBUG2",
+            "DEBUG1",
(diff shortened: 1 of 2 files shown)


prowler/providers/gcp/services/cloudsql/cloudsql_instance_postgres_log_min_messages_flag/cloudsql_instance_postgres_log_min_messages_flag.py
```diff
@@ -4,7 +4,17 @@
...
-        desired_log_min_messages = "error"
+        failing_log_levels = [
+            "DEBUG5",
+            "DEBUG4",
+            "DEBUG3",
+            "DEBUG2",
+            "DEBUG1",
+            "INFO",
+            "NOTICE",
+            "WARNING",
```
benlahav: We should remove this "WARNING" because the CIS benchmark suggest AT LEAST WARNING but recommend ERROR, image attached:
image

MrCloudSec:
```suggestion
```

MrCloudSec: Good catch @benlahav, thank you!

puchy22 (approved):
LGTM! ✅

PR #5593, opened 2024-10-30
fix(aws): do not flag cross-service confused deputy as public

### LINKED ISSUE #5454: "public" definition of resources
### Steps to Reproduce

looking at your code based on some customer questions we've received.

### Expected behavior

Hello. 

re this code here:

https://github.com/prowler-cloud/prowler/blob/778edd5fec7d0cf8ebae6d1471d54fbb18db81ea/prowler/providers/aws/services/iam/lib/policy.py#L112

While this is useful check, it looks to me like you're not checking for aws:SourceArn, aws:SourceOrgId, or aws:SourceOrgPaths which can be used as an alternative to aws:SourceAccount to help protect against the cross-service confused deputy issue.

Also this finding is really for "lack of cross-service confused deputy condition key present" - this does not mean the resource is public. I think that's confusing because that's not how AWS talks about public resources, and implies that there is for sure a way an unauthorized party to access your resource which is usually not the case. 

I can appreciate that nuance may be hard to message, but I think something that's probably easy to do is to identify this as a cross service confused deputy finding, and provide the relevant documentation to the prowler user:

https://docs.aws.amazon.com/IAM/latest/UserGuide/confused-deputy.html#cross-service-confused-deputy-prevention

N/a

### Environment Resource

N/A

### OS used

N/A

### Prowler version

N/A

### Pip version

N/A

### Context

AWS receiving some customer outreach around what does this finding mean

MrCloudSec:
### Context

Fix #5454.

### Description

Do not flag IAM policies with cross service confused deputy as public unless the argument `check_cross_service_confused_deputy` is passed to the `is_policy_public` function.
### Checklist

### License

### DIFF
--- a/prowler/providers/aws/services/iam/iam_role_cross_service_confused_deputy_prevention/iam_role_cross_service_confused_deputy_prevention.py
+++ b/prowler/providers/aws/services/iam/iam_role_cross_service_confused_deputy_prevention/iam_role_cross_service_confused_deputy_prevention.py
@@ -20,6 +20,7 @@ class iam_role_cross_service_confused_deputy_prevention(Check):
                     if not is_policy_public(
                         role.assume_role_policy,
                         iam_client.audited_account,
+                        check_cross_service_confused_deputy=True,
                         not_allowed_actions=["sts:AssumeRole", "sts:*"],
                     ):
                         report.status = "PASS"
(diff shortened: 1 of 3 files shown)


jfagoagas (approved):
Great fix!!

PR #5602, opened 2024-10-31
feat(appsync): add new check `appsync_field_level_logging_enabled`

MarioRgzLpz:
### Context

This control ensures that field-level logging is enabled for AWS AppSync APIs. The control fails if field-level logging is not enabled or is neither ALL nor ERROR.

Enabling field-level logging helps you identify, troubleshoot, and optimize GraphQL queries by providing detailed information about API requests and responses. It is essential for addressing issues and meeting regulatory requirements.

### Description

Add new check `appsync_field_level_logging_enabled` with respective unit tests and metadata.

### Checklist

### License


MrCloudSec (changes requested):
You need to get all APIs since this check not only applies to GraphQL APIs.
Please, use the function `list_apis` to do so.

MrCloudSec (approved):

PR #5645, opened 2024-11-05
fix(docker): add g++ to Dockerfile for presidio-analyzer compatibility

MrCloudSec:
### Context
The absence of `g++` was causing docker build to fail due to compilation errors in dependencies like `cymem` and `murmurhash`. Adding `g++` ensures successful installation of presidio-analyzer and related packages.
### Description

Adds`g++` to the Dockerfile to resolve build issues with `presidio-analyzer,` which requires a C++ compiler for certain dependencies.

### Checklist

### License

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,9 +2,9 @@ FROM python:3.12-alpine
 
 LABEL maintainer="https://github.com/prowler-cloud/prowler"
 
-# Install essential tool to compile presidio-analyze library
-# hadolint ignore=DL3018
-RUN apk --no-cache add g++
+# Update system dependencies and install essential tools
+#hadolint ignore=DL3018
+RUN apk --no-cache upgrade && apk --no-cache add curl git g++
 
 # Create non-root user
 RUN mkdir -p /home/prowler && \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -2,11 +2,11 @@ FROM python:3.12-alpine
 
 LABEL maintainer="https://github.com/prowler-cloud/prowler"
 
-# Update system dependencies and install essential tools
-#hadolint ignore=DL3018
-RUN apk --no-cache upgrade && apk --no-cache add curl git
+# Install essential tool to compile presidio-analyze library
+# hadolint ignore=DL3018
+RUN apk --no-cache add g++
```
jfagoagas: Why are you removing the other packages installation?

MrCloudSec: I think they are not needed, but I can leave them if necessary.

MrCloudSec:
```suggestion
# Update system dependencies and install essential tools
#hadolint ignore=DL3018
RUN apk --no-cache upgrade && apk --no-cache add curl git g++
```

jfagoagas (approved):

PR #5657, opened 2024-11-07
feat(core): Save memory at execution 

cr0hn:
### Context

This PR has a new feature that significantly reduces the memory footprint when Prowler is running.

### Description

This PR aims to reduce Prowler's memory footprint by switching from the regular Python list and dictionaries to SQLite3 as a database.

In the first place, it could sound a bit weird, but reducing Prowler's memory footprint is actually a good idea, and the performance impact is really low.

### License


jfagoagas:
Duplicates https://github.com/prowler-cloud/prowler/pull/4892

jfagoagas:

PR #5697, opened 2024-11-08
fix(ec2): unique finding per Security Group in high risk ports check

MarioRgzLpz:
### Context

An issue was detected within `ec2_securitygroup_allow_ingress_from_internet_to_high_risk_tcp_ports` check because it loops over a list of ports and all the findings are generated with the same finding.uid which can’t be a possibility.

### Description

Changed `ec2_securitygroup_allow_ingress_from_internet_to_high_risk_tcp_ports` logic to create one finding per security group instead of one finding per high risk port.

### Checklist

### License


MrCloudSec (approved):
Great job!!

PR #5736, opened 2024-11-12
fix(aws): remove `cloudwatch_log_group_no_critical_pii_in_logs` check

MrCloudSec:
### Description

Remove check `cloudwatch_log_group_no_critical_pii_in_logs` since `presidio-analyzer` is downloading a package during Prowler execution.
### Checklist

### License


jfagoagas (approved):
We are removing the check `cloudwatch_log_group_no_critical_pii_in_logs` for the following reasons:
- It uses the [`presidio-analyzer`](https://microsoft.github.io/presidio/analyzer/) library which by default loads during runtime the NLP module and other PII recognizers from external sources not defined in the Prowler dependencies. That's something we could not do as a security product taking also into account that Prowler can run in environments with little to zero access to the Internet.
- Those dependencies cannot be installed via `poetry` since they are not available in `PyPI` and would require adding more steps to our installation procedure to get them.
- The NLP module installed, [`en-core-web-lg`](https://huggingface.co/spacy/en_core_web_lg) is a large vector table with ~500k entries with a size of 400MiB. This is a huge package to be donwloaded during runtime.
- The installation procedure of the above dependencies adds several lines to our terminal UI with the download process and some logs.
-  [`presidio-analyzer`](https://microsoft.github.io/presidio/analyzer/) requires a ton of new libraries that we need to review first.

We will work to reintroduce the `cloudwatch_log_group_no_critical_pii_in_logs` in a way that all dependencies are explicitly defined, required and verified.

poetry.lock
jfagoagas, line 1: Please review this file taking into account the new dependency updates. We should not add changes here.

jfagoagas (changes requested):

jfagoagas (approved):

PR #5751, opened 2024-11-13
refactor(azure): get locations with self session

pedrooot:
### Description

This pull request includes changes to the `azure_provider.py` file within the `prowler/providers/azure` directory. The main focus of these changes is to simplify the `get_locations` method by removing the need to pass credentials as an argument and instead using the session directly.

Key changes include:

* Simplification of the `__init__` method by removing the credentials argument from the `get_locations` method call.
* Modification of the `get_locations` method to no longer require credentials as an argument and directly use the session for authentication.

### Checklist

### License

### DIFF
--- a/prowler/providers/azure/azure_provider.py
+++ b/prowler/providers/azure/azure_provider.py
@@ -198,7 +198,7 @@ class AzureProvider(Provider):
         )
 
         # TODO: should we keep this here or within the identity?
-        self._locations = self.get_locations(self.session)
+        self._locations = self.get_locations()
 
         # Audit Config
         if config_content:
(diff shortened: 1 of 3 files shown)


prowler/providers/azure/azure_provider.py
```diff
@@ -942,33 +942,29 @@ async def get_azure_identity():
...
-                    "Content-Type": "application/json",
-                }
-                response = requests.get(url, headers=headers)
-                if response.status_code == 200:
-                    data = response.json()
-                    for location in data["value"]:
-                        locations[display_name].append(location["name"])
+        credentials = self.session
+        locations = {}
+        token = credentials.get_token("https://management.azure.com/.default").token
```
jfagoagas: Can you do that with the Azure SDK?

prowler/providers/azure/azure_provider.py
```diff
@@ -942,33 +942,29 @@ async def get_azure_identity():
...
-    def get_locations(self, credentials) -> dict[str, list[str]]:
+    def get_locations(self) -> dict[str, list[str]]:
         """
         Retrieves the locations available for each subscription using the provided credentials.
 
-        Args:
-            credentials: The credentials object used to authenticate the request.
-
         Returns:
             A dictionary containing the locations available for each subscription. The dictionary
```
jfagoagas, line 950: Can you add an example of the output?

prowler/providers/azure/azure_provider.py
```diff
@@ -942,33 +942,29 @@ async def get_azure_identity():
...
+            url = f"https://management.azure.com/subscriptions/{subscription_id}/locations?api-version=2022-12-01"
+            headers = {
+                "Authorization": f"Bearer {token}",
+                "Content-Type": "application/json",
+            }
+            response = requests.get(url, headers=headers)
+            if response.status_code == 200:
+                data = response.json()
+                for location in data["value"]:
+                    locations[display_name].append(location["name"])
```
jfagoagas: What is this function used for? Can you review the output format?

jfagoagas (changes requested):

prowler/providers/azure/azure_provider.py
pedrooot: Done

prowler/providers/azure/azure_provider.py

prowler/providers/azure/azure_provider.py
pedrooot: Output format added on the docstrings. This method it's used to set `_locations` for AzureProvider but this it's never used. We should discuss what to do with this attributes.

jfagoagas (dismissed):

jfagoagas (approved):

PR #5773, opened 2024-11-14
feat(rds): add new fixer `rds_snapshots_public_access_fixer`

danibarranqueroo:
### Context

Develop a fixer that modifies the access permissions of RDS snapshots to ensure they are private and not accessible to the public. This will help prevent unauthorized access to sensitive data stored in database backups, enhancing the security of RDS resources.

As this fixer is executed when the `rds_snapshots_public_access` check fails and that check scans instances and clusters snapshots I've had to adapt a little bit the code of the fixer in order to know only with the ResourceID if what I'm trying to modify is an instance or a cluster.

### Description

Added new fixer `rds_snapshots_public_access_fixer` with its unit tests.

### Checklist

### License


MrCloudSec (approved):
Good job !

PR #5813, opened 2024-11-18
fix(aws): exclude member accounts in IAM Root Credentials check

MrCloudSec:
### Context

The recently added check `iam_root_credentials_management_enabled` uses the `ListOrganizationsFeatures` API to verify if centralized root credentials management is enabled. However, this API is restricted to the management account or delegated administrators. AWS Support confirmed that member accounts cannot access this information, resulting in the `AccountNotManagementOrDelegatedAdministratorException` error.

### Description

The check now skips findings for member accounts, as they are unable to retrieve the required data due to API restrictions. This ensures accurate results without unnecessary findings for member accounts.
### Checklist

### License

### DIFF
--- a/prowler/providers/aws/services/iam/iam_service.py
+++ b/prowler/providers/aws/services/iam/iam_service.py
@@ -982,6 +982,21 @@ class IAM(AWSService):
                 logger.warning(
                     f"{self.region} -- {error.__class__.__name__}[{error.__traceback__.tb_lineno}]: {error}"
                 )
+            elif (
+                error.response["Error"]["Code"]
+                == "OrganizationNotInAllFeaturesModeException"
+            ):
+                logger.warning(
+                    f"{self.region} -- {error.__class__.__name__}[{error.__traceback__.tb_lineno}]: {error}"
+                )
+            elif (
+                error.response["Error"]["Code"]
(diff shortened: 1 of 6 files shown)


prowler/providers/aws/services/iam/iam_service.py
```diff
@@ -982,6 +982,21 @@ def _list_organizations_features(self):
...
+                    f"{self.region} -- {error.__class__.__name__}[{error.__traceback__.tb_lineno}]: {error}"
+                )
+            elif (
+                error.response["Error"]["Code"]
+                == "AccountNotManagementOrDelegatedAdministratorException"
+            ):
+                logger.warning(
+                    f"{self.region} -- {error.__class__.__name__}[{error.__traceback__.tb_lineno}]: {error}"
+                )
+                self.organization_features = None
```
jfagoagas, line 999: Is this the only scenario where we need to set it to `None`?

MrCloudSec: Yes, it is treated like an access denied.

jfagoagas: Right, thanks!

jfagoagas (approved):

PR #5855, opened 2024-11-21
feat(s3): add new fixer `s3_bucket_public_write_acl_fixer`

danibarranqueroo:
### Context

Develop a fixer that removes public write access from the ACLs of S3 buckets. The fixer will ensure that only authorized users or IAM roles can write to the bucket, preventing unauthorized modifications and protecting sensitive data stored in S3.

This fixer is already done and tested in moto but I don’t want to test it on real infra until I discuss with someone else because for creating a bucket with public write access it’s needed to change the “BlockPublicAcls” parameter at bucket and account level and this could affect to the other s3 buckets that are currently on our AWS ProwlerDev account. So, for this reason, I'll left this PR in draft.

### Description

Added a new fixer `s3_bucket_public_write_acl_fixer` with its unit tests.

### Checklist

### License


danibarranqueroo:
I have now my personal account working and I've been able to test on real infrastructure the fixer, as it works as expected this PR is now ready for review.

MrCloudSec (approved):

PR #5904, opened 2024-11-26
fix(ec2): Change `ec2_sg_high_risk_ports` configurable parameter name

MarioRgzLpz:
### Context

Changed configurable parameter `ec2_sg_high_risk_ports` to `ec2_high_risk_ports` so its the same in the check and in the config file and correctly works.

### Description

Changed configurable parameter `ec2_sg_high_risk_ports` to `ec2_high_risk_ports` in every config file needed.

### Checklist

### License


MrCloudSec (approved):
Good catch!

PR #5911, opened 2024-11-26
feat(sqs): add new fixer `sqs_queues_not_publicly_accessible_fixer`

danibarranqueroo:
### Context

Developed a new fixer that restricts public access to SQS queues, ensuring that only authorized users and applications can interact with the queue. This includes modifying the queue's access policy to block any public access permissions and limit access to trusted IAM roles or specific AWS accounts.

The way to solve this is modifying the attributes of the failed queue, here we add a new policy which is the old one but changing the public statements to a trusted account.

### Description

Added new fixer `sqs_queues_not_publicly_accessible_fixer` with its unit tests.

### Checklist

### License

### DIFF
--- a/prowler/providers/aws/services/sqs/sqs_queues_not_publicly_accessible/sqs_queues_not_publicly_accessible_fixer.py
+++ b/prowler/providers/aws/services/sqs/sqs_queues_not_publicly_accessible/sqs_queues_not_publicly_accessible_fixer.py
@@ -33,6 +33,8 @@ def fixer(resource_id: str, region: str) -> bool:
 
         regional_client = sqs_client.regional_clients[region]
 
+        queue_name = resource_id.split("/")[-1]
+
         trusted_policy = {
             "Version": "2012-10-17",
             "Statement": [
@@ -43,6 +45,7 @@ def fixer(resource_id: str, region: str) -> bool:
                         "AWS": f"arn:{audited_partition}:iam::{account_id}:root"
                     },
                     "Action": "sqs:*",
+                    "Resource": f"arn:{audited_partition}:sqs:{region}:{queue_name}",
                 }
             ],
         }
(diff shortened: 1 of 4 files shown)


prowler/providers/aws/services/sqs/sqs_queues_not_publicly_accessible/sqs_queues_not_publicly_accessible_fixer.py
```diff
@@ -0,0 +1,61 @@
...
+    Returns:
+        bool: True if the operation is successful (policy updated), False otherwise.
+    """
+    try:
+        account_id = sqs_client.audited_account
+        audited_partition = sqs_client.audited_partition
+
+        regional_client = sqs_client.regional_clients[region]
+
+        trusted_policy = {
```
MrCloudSec, line 38: You have to also set the resource, look at this example https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-basic-examples-of-sqs-policies.html#grant-one-permission-to-one-account

danibarranqueroo: Done! Also to follow that example I needed to use the name instead of the resource_id so take a look of how I've done it please 🚀

prowler/providers/aws/services/sqs/sqs_queues_not_publicly_accessible/sqs_queues_not_publicly_accessible_fixer.py
```diff
@@ -0,0 +1,64 @@
...
+            "Version": "2012-10-17",
+            "Statement": [
+                {
+                    "Sid": "ProwlerFixerStatement",
+                    "Effect": "Allow",
+                    "Principal": {
+                        "AWS": f"arn:{audited_partition}:iam::{account_id}:root"
+                    },
+                    "Action": "sqs:*",
+                    "Resource": f"arn:{audited_partition}:sqs:{region}:{account_id}:{queue_name}",
```
MrCloudSec, line 48:
```suggestion
                    "Resource": account_id,
```

MrCloudSec (approved):

PR #5926, opened 2024-11-27
feat(opensearch): add new fixer `opensearch_service_domains_not_publicly_accessible_fixer`

danibarranqueroo:
### Context

Developed a new fixer that restricts public access to OpenSearch Service domains, ensuring they are only accessible from within the VPC or specific, authorized IP ranges. This will enhance security by preventing unauthorized users from accessing sensitive data stored in OpenSearch domains.

### Description

Added new fixer `opensearch_service_domains_not_publicly_accessible_fixer` with its unit tests.

### Checklist

### License

### DIFF
--- a/prowler/providers/aws/services/opensearch/opensearch_service_domains_not_publicly_accessible/opensearch_service_domains_not_publicly_accessible_fixer.py
+++ b/prowler/providers/aws/services/opensearch/opensearch_service_domains_not_publicly_accessible/opensearch_service_domains_not_publicly_accessible_fixer.py
@@ -1,0 +1,68 @@
+import json
+
+from prowler.lib.logger import logger
+from prowler.providers.aws.services.opensearch.opensearch_client import (
+    opensearch_client,
+)
+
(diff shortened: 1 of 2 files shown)


prowler/providers/aws/services/opensearch/opensearch_service_domains_not_publicly_accessible/opensearch_service_domains_not_publicly_accessible_fixer.py
MrCloudSec, line 1: We need to delete the policy here since only the fine-grained access control should be used.

danibarranqueroo: For this one, there is no API call to delete the policy so I've used the call that update the Domain Configuration and put an empty policy.

MrCloudSec (approved):

PR #5939, opened 2024-11-27
fix(k8s): handle Kubernetes kubeconfig content correctly

MrCloudSec:
### Description
- Ensured the SDK uses consistent Kubernetes credentials regardless of whether the `kubeconfig` is provided as a multi-line string or a file.

### Checklist

### License


prowler/providers/kubernetes/kubernetes_provider.py
```diff
@@ -74,7 +74,7 @@ def __init__(
         fixer_config: dict = {},
         mutelist_path: str = None,
         mutelist_content: dict = {},
-        kubeconfig_content: dict = None,
+        kubeconfig_content: str = None,
```
jfagoagas: I think this is still in draft but we should keep the ability of passing a `dict` too, the SDK does not need to be adapted because the CLI or the API are the interfaces that needs to pass the data in the required format. The good point is that `safe_load` is idempotent so you can pass either a `str` or a `dict` and if the format is valid a `dict` will be always returned.

api/src/backend/api/models.py
```diff
@@ -187,12 +187,14 @@ def validate_gcp_uid(value):
...
-                code="kubernetes-uid",
-                pointer="/data/attributes/uid",
+        if not re.match(
+            r"(^[a-z0-9]([-a-z0-9]{1,61}[a-z0-9])?$)|(^arn:aws(-cn|-us-gov|-iso|-iso-b)?:[a-zA-Z0-9\-]+:([a-z]{2}-[a-z]+-\d{1})?:(\d{12})?:[a-zA-Z0-9\-_\/:\.\*]+(:\d+)?$)",
+            value,
+        ):
+            raise ValueError(
+                "The value must either be a valid Kubernetes UID (up to 63 characters, "
+                "starting and ending with a lowercase letter or number, containing only "
+                "lowercase alphanumeric characters and hyphens) or a valid EKS ARN."
```
vicferpoy: Please, update regexes and detail messages, but don't modify the current logic. Use `ModelValidationError` with all the required fields. This breaks this exception.

vicferpoy (dismissed):

prowler/providers/kubernetes/kubernetes_provider.py
```diff
@@ -273,17 +280,17 @@ def setup_session(
...
+                    contexts = config.list_kube_config_contexts(
+                        config_file=kubeconfig_file
+                    )[0]
+                    for context_item in contexts:
+                        if context_item["name"] == context:
+                            context = context_item
+                else:
+                    context = config.list_kube_config_contexts(
+                        config_file=kubeconfig_file
+                    )[1]
```
jfagoagas, line 295: Why are you getting the `[1]`?

MrCloudSec: Because that is the active context, that was not changed.

api/src/backend/api/models.py
MrCloudSec: Good catch, I will change it now!

api/src/backend/api/models.py
```diff
@@ -187,12 +187,14 @@ def validate_gcp_uid(value):
...
+            value,
+        ):
             raise ModelValidationError(
-                detail="K8s provider ID must be up to 63 characters, start and end with a lowercase letter or number, "
-                "and contain only lowercase alphanumeric characters and hyphens.",
-                code="kubernetes-uid",
-                pointer="/data/attributes/uid",
+                "The value must either be a valid Kubernetes UID (up to 63 characters, "
+                "starting and ending with a lowercase letter or number, containing only "
+                "lowercase alphanumeric characters and hyphens) or a valid EKS ARN."
```
jfagoagas: We need to create a ticket for this to see AKS and GKE, also I think the API should not be responsible of this validation.

prowler/providers/kubernetes/kubernetes_provider.py
jfagoagas: If the `active` is the second element, what is in the first one?

MrCloudSec: The first one is a list of all the contexts, the second one is an object of the active context.

ui/components/ui/custom/custom-input.tsx
```diff
@@ -91,20 +91,33 @@ export const CustomInput = <T extends FieldValues>({
...
-              placeholder={inputPlaceholder}
-              type={inputType}
-              variant={variant}
-              size={size}
-              isInvalid={isInvalid}
-              defaultValue={defaultValue}
-              endContent={endContent}
-              {...field}
-            />
+            {type === "textarea" ? (
```
paabloLC: @MrCloudSec Thanks for the contribution. 

We're currently using this component 36 times across 15 different files. It's likely the most important component in Prowler v5 (it's present in every form we have). During our discussion today, I briefly explained how it works.

The `<input>` element is an HTML component, while `<textarea>` is a completely distinct HTML component. By making changes to CustomInput, you’re merging two separate components into one, which is not a viable approach.

To summarize, we would need to create a CustomTextArea component and customize it in the same way as CustomInput, including all the necessary props so it can be reused consistently without breaking anything.

I'm happy to assist with any changes needed to move forward, but this will have to wait until the release is finished as there are higher priority items

Please ensure this file is not modified in this PR.

Thank you.

paabloLC (dismissed):
Please check my comment on the CustomInput component.

jfagoagas (approved):
Great fix @MrCloudSec 👏

PR #5950, opened 2024-11-28
feat(glacier): add new fixer `glacier_vaults_policy_public_access_fixer`

danibarranqueroo:
### Context

Developed a new fixer that updates the access policies for Glacier vaults to restrict public access, ensuring they are only accessible to authorized AWS accounts and trusted IAM roles. This will help prevent unauthorized access to archived data and improve the security of Glacier vault resources.

Also, while testing, I noticed that the unit tests from the check `glacier_vaults_policy_public_access` did not have the right patch in order to work without problems with MagicMock so I've fixed them. 

Aditionally, as Moto does not cover the `set_vault_access_policy` API call I needed to use Botocore.

### Description

Added new fixer `glacier_vaults_policy_public_access_fixer` with its unit tests.

### Checklist

### License

### DIFF
--- a/prowler/providers/aws/services/glacier/glacier_vaults_policy_public_access/glacier_vaults_policy_public_access_fixer.py
+++ b/prowler/providers/aws/services/glacier/glacier_vaults_policy_public_access/glacier_vaults_policy_public_access_fixer.py
@@ -1,0 +1,60 @@
+import json
+
+from prowler.lib.logger import logger
+from prowler.providers.aws.services.glacier.glacier_client import glacier_client
+
+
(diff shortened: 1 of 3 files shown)


prowler/providers/aws/services/glacier/glacier_vaults_policy_public_access/glacier_vaults_policy_public_access_fixer.py
MrCloudSec, line 1: Same here, we need to delete the vault policy.

MrCloudSec (approved):

PR #5954, opened 2024-11-28
feat(only_logs): deprecate only_logs flag

pedrooot:
### Description

Remove `only_logs` flag from Prowler

### Checklist

### License


jfagoagas (commented):
I've reviewed the PR and I'm not sure if we can remove this since there is an execution method for Windows that will fail.

jfagoagas:
We are not going to deprecate this, we will refactor it for the API to use it.

PR #5968, opened 2024-11-29
fix(scan): fix deadlock on resource transactions during concurrent scans

vicferpoy:
### Context

During concurrent scans in different Celery workers, there are situations where the same resource is accessed in different transactions, causing a deadlock.

### Description

A configurable retry mechanism has been added to handle these situations, terminating the locked transaction and trying again after a short time.

image

### License


jfagoagas (approved):
Great fix, thanks @vicferpoy 🚀

PR #5973, opened 2024-11-30
feat(dynamic_filters): add dynamic filters system

AdriiiPRodri:
### Context

In this PR I have added the endpoint to get the filters dynamically depending on the findings of a date.

### Description

Using the specified date, all the findings of that date and the services and regions associated with them are extracted.
```
{
    "data": {
        "type": "findings-services-regions",
        "id": null,
        "attributes": {
            "services": [
                "accessanalyzer",
                "account"
            ],
            "regions": [
                "ap-northeast-1",
                "ap-northeast-2",
                "ap-northeast-3",
                "ap-south-1",
                "ap-southeast-1",
                "ap-southeast-2",
                "ca-central-1",
                "eu-central-1",
                "eu-central-2",
                "eu-north-1",
                "eu-south-2",
                "eu-west-1",
                "eu-west-2",
                "eu-west-3",
                "sa-east-1",
                "us-east-1",
                "us-east-2",
                "us-west-1",
                "us-west-2"
            ]
        }
    }
}
```

image

### Checklist

### License

### DIFF
--- a/api/src/backend/api/filters.py
+++ b/api/src/backend/api/filters.py
@@ -426,6 +426,16 @@ class FindingFilter(FilterSet):
         return dt
 
 
+class FindingDynamicFilter(FilterSet):
+    updated_at = DateFilter(field_name="updated_at", lookup_expr="date")
+
+    class Meta:
+        model = Finding
+        fields = {
+            "updated_at": ["exact"],
+        }
+
+
 class ProviderSecretFilter(FilterSet):
     inserted_at = DateFilter(field_name="inserted_at", lookup_expr="date")
     updated_at = DateFilter(field_name="updated_at", lookup_expr="date")
--- a/api/src/backend/api/filters.py
+++ b/api/src/backend/api/filters.py
@@ -426,31 +426,6 @@ class FindingFilter(FilterSet):
         return dt
 
 
-class FindingDynamicFilter(FilterSet):
-    inserted_at = DateFilter(
-        method="filter_inserted_at", lookup_expr="date", required=True
-    )
-
-    class Meta:
-        model = Finding
-        fields = {
-            "inserted_at": ["exact"],
-        }
-
-    def filter_inserted_at(self, queryset, name, value):
-        value = self.maybe_date_to_datetime(value)
-        start = uuid7_start(datetime_to_uuid7(value))
-
-        return queryset.filter(id__gte=start).filter(inserted_at__date=value)
-
-    @staticmethod
-    def maybe_date_to_datetime(value):
-        dt = value
-        if isinstance(value, date):
-            dt = datetime.combine(value, datetime.min.time(), tzinfo=timezone.utc)
-        return dt
-
-
 class ProviderSecretFilter(FilterSet):
     inserted_at = DateFilter(field_name="inserted_at", lookup_expr="date")
     updated_at = DateFilter(field_name="updated_at", lookup_expr="date")
--- a/api/src/backend/api/filters.py
+++ b/api/src/backend/api/filters.py
@@ -427,13 +427,28 @@ class FindingFilter(FilterSet):
 
 
 class FindingDynamicFilter(FilterSet):
-    updated_at = DateFilter(field_name="updated_at", lookup_expr="date")
+    inserted_at = DateFilter(
+        method="filter_inserted_at", lookup_expr="date", required=True
+    )
 
     class Meta:
         model = Finding
         fields = {
-            "updated_at": ["exact"],
+            "inserted_at": ["exact"],
         }
+
+    def filter_inserted_at(self, queryset, name, value):
+        value = self.maybe_date_to_datetime(value)
+        start = uuid7_start(datetime_to_uuid7(value))
+
+        return queryset.filter(id__gte=start).filter(inserted_at__date=value)
+
+    @staticmethod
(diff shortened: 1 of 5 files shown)


api/src/backend/api/filters.py
```diff
@@ -426,6 +426,16 @@ def maybe_date_to_datetime(value):
         return dt
 
 
+class FindingDynamicFilter(FilterSet):
+    updated_at = DateFilter(field_name="updated_at", lookup_expr="date")
```
vicferpoy: We should filter by `inserted_at`, not `updated_at`. Findings are never updated and it does not make much sense.

Also, you can define `required=True` here to prevent validation on view level and add the field and required tag in the documentation without further changes.

api/src/backend/api/v1/serializers.py
```diff
@@ -707,6 +707,14 @@ class Meta:
     }
 
 
+class FindingDynamicFilterSerializer(serializers.Serializer):
+    services = serializers.ListField(child=serializers.CharField())
+    regions = serializers.ListField(child=serializers.CharField())
+
+    class Meta:
+        resource_name = "findings-services-regions"
```
vicferpoy: I would call this `finding-dynamic-filters`.

api/src/backend/api/v1/views.py
```diff
@@ -978,6 +982,22 @@ def get_queryset(self):
         summary="Retrieve data from a specific finding",
         description="Fetch detailed information about a specific finding by its ID.",
     ),
+    findings_services_regions=extend_schema(
+        tags=["Finding"],
+        summary="Retrieve the services and regions that are impacted by a finding in a specific date",
+        description="Fetch services and regions affected in a finding by date.",
```
```suggestion
        summary="Retrieve the services and regions that are impacted by findings in a specific date",
        description="Fetch services and regions affected in findings by date.",
```

api/src/backend/api/v1/views.py
```diff
@@ -978,6 +982,22 @@ def get_queryset(self):
...
+        parameters=[
+            OpenApiParameter(
+                name="filter[updated_at]",
+                required=True,
+                type=OpenApiTypes.DATE,
+                location=OpenApiParameter.QUERY,
+                description="Date in the format YYYY-MM-DD",
```
vicferpoy: Remember to update this after the changes in the filterset. Although It might not be necessary to add anything after the changes. Up to you.

api/src/backend/api/v1/views.py
```diff
@@ -1008,6 +1027,24 @@ def inserted_at_to_uuidv7(self, inserted_at):
...
 
+    def get_serializer_class(self):
+        if self.action == "findings_services_regions":
+            return FindingDynamicFilterSerializer
+
+        return super().get_serializer_class()
+
+    def get_filterset_class(self):
+        if self.action in ["findings_services_regions"]:
+            self.ordering_fields = []
```
vicferpoy: This is covered in `get_ordering_fields` right? Or is it necessary here as well?

api/src/backend/api/v1/views.py
```diff
@@ -1039,6 +1076,32 @@ def get_queryset(self):
...
+    @action(detail=False, methods=["get"], url_name="findings_services_regions")
+    def findings_services_regions(self, request):
+        try:
+            updated_at = datetime.strptime(
+                request.query_params["filter[updated_at]"], "%Y-%m-%d"
+            ).date()
+            if updated_at > date.today():
+                raise ValidationError(detail="The date must be a date in the past.")
+        except ValueError:
+            raise ValidationError(detail="Invalid date format.")
```
vicferpoy: Was this asked by UI or something? We have a lot of date filters without this kind of logic. It simply should return no values at all if we provide a future date. If it is required, this validation should belong in the filter module in my opinion.

vicferpoy (changes requested):
This looks really good, great job! I left a few suggestions and one required change. Feel free to discuss anything since I lack some context about the requirements here.

vicferpoy:
@AdriiiPRodri Also, I forgot to mention this filter could be vastly improved using the logic from the `inserted_at` filter from the base Findings filterset:

```python
    def filter_inserted_at(self, queryset, name, value):
        value = self.maybe_date_to_datetime(value)
        start = uuid7_start(datetime_to_uuid7(value))

        return queryset.filter(id__gte=start).filter(inserted_at__date=value)
```

Please take a look at how it is done in the base findings filters, because it takes partitioned tables into consideration and improves efficiency by a lot.

api/src/backend/api/v1/views.py
AdriiiPRodri: I remove get_ordering_fields, this functions doesn't work for the doc generation

AdriiiPRodri:
> @AdriiiPRodri Also, I forgot to mention this filter could be vastly improved using the logic from the `inserted_at` filter from the base Findings filterset:
> 
> ```python
>     def filter_inserted_at(self, queryset, name, value):
>         value = self.maybe_date_to_datetime(value)
>         start = uuid7_start(datetime_to_uuid7(value))
> 
>         return queryset.filter(id__gte=start).filter(inserted_at__date=value)
> ```
> 
> Please take a look at how it is done in the base findings filters, because it takes partitioned tables into consideration and improves efficiency by a lot.

Fixed in the last commit, thanks!

PR #6064, opened 2024-12-09
chore(RLS): Improve policy comparison

jfagoagas:
### Motivation

When an RLS policy is applied to a query, the conditions specified in the policy are evaluated for each row that the query touches. In queries affecting thousands of rows, this behavior can drastically reduce query performance, as the overhead of executing these functions adds up quickly. -- https://supabase.com/docs/guides/database/database-advisors?queryGroups=lint&lint=0003_auth_rls_initplan#the-performance-issue
### Description

Improve Postgres Row Level Security policies:
- Remove the `CASE` since it is not needed, a comparison with `null` returns `false`, so calling `current_setting('setting', True)` is more than enough.
- Wrap `current_setting` in a subquery to be executed in an `InitPlan` which will be executed just once and then the result will be cached for the rest of the rows.

### TODO
- [ ] Pending to create migrations to replace these policies.

### Checklist

- [ ] Review if backport is needed. -> **Not needed.**

### License

### DIFF
--- a/api/src/backend/api/rls.py
+++ b/api/src/backend/api/rls.py
@@ -79,7 +79,7 @@ class RowLevelSecurityConstraint(models.BaseConstraint):
         grant_queries = ""
         for statement in self.statements:
             # SELECT and DELETE uses USING since applies to rows selected
-            # INSERT uses WITH CHECK becauseonly applies in cases where records are being added
+            # INSERT uses WITH CHECK because only applies in cases where records are being added
             # UPDATE requires USING and WITH CHECK but if only a USING clause is specified, then that clause will be used for both USING and WITH CHECK cases.
             # https://www.postgresql.org/docs/current/sql-createpolicy.html
             clause = f"{'WITH CHECK' if statement == 'INSERT' else 'USING'}"
--- a/api/src/backend/api/rls.py
+++ b/api/src/backend/api/rls.py
@@ -2,8 +2,7 @@ from typing import Any
 from uuid import uuid4
 
 from django.core.exceptions import ValidationError
-from django.db import DEFAULT_DB_ALIAS
-from django.db import models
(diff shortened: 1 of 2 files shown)


api/src/backend/api/rls.py
```diff
@@ -45,10 +44,7 @@ class RowLevelSecurityConstraint(models.BaseConstraint):
         FOR {statement}
         TO %(db_user)s
         {clause} (
-            CASE
-                WHEN current_setting('%(tenant_setting)s', True) IS NULL THEN FALSE
-                ELSE %(field_column)s = current_setting('%(tenant_setting)s')::uuid
-            END
+            (SELECT current_setting('%(tenant_setting)s', True)::uuid) = %(field_column)s
```
jfagoagas, line 47: Pending to create migrations to replace these policies.

api/src/backend/api/rls.py
```diff
@@ -82,6 +78,10 @@ def create_sql(self, model: Any, schema_editor: Any) -> Any:
         policy_queries = ""
         grant_queries = ""
         for statement in self.statements:
+            # SELECT and DELETE uses USING since applies to rows selected
+            # INSERT uses WITH CHECK becauseonly applies in cases where records are being added
```
vicferpoy:
```suggestion
            # INSERT uses WITH CHECK because only applies in cases where records are being added
```

vicferpoy (changes requested):
Thanks for the improvement! However, we are missing changes on migrations. This change will only exist for users who don't have the initial migration applied.

jfagoagas:
bugbot run

jfagoagas:
bugbot run

jfagoagas:
Closing this PR as it's stale. I'll re-open it.

PR #6066, opened 2024-12-09
fix(RLS): enforce config security

jfagoagas:
### Context

Protect the queries setting the `tenant_id` for RLS using parameters to prevent SQLi. Although the scenario to exploit this is difficult since the `tenant_id` comes from the token which is signed by us and cannot be verified otherwise.

### Description

- Use placeholders with `cursor.execute` as recommended https://docs.djangoproject.com/en/5.1/topics/db/sql/#executing-custom-sql-directly
- Unify RLS parameter setting in just one place

### Checklist

### License

### DIFF
--- a/api/src/backend/api/db_utils.py
+++ b/api/src/backend/api/db_utils.py
@@ -1,4 +1,5 @@
 import secrets
+import uuid
 from contextlib import contextmanager
 from datetime import datetime, timedelta, timezone
 
@@ -8,6 +9,7 @@ from django.core.paginator import Paginator
 from django.db import connection, models, transaction
 from psycopg2 import connect as psycopg2_connect
 from psycopg2.extensions import AsIs, new_type, register_adapter, register_type
+from rest_framework_json_api.serializers import ValidationError
 
 DB_USER = settings.DATABASES["default"]["USER"] if not settings.TESTING else "test"
 DB_PASSWORD = (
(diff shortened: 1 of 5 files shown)


api/src/backend/api/db_utils.py
```diff
@@ -47,7 +47,12 @@ def psycopg_connection(database_alias: str):
 def tenant_transaction(tenant_id: str):
     with transaction.atomic():
         with connection.cursor() as cursor:
-            cursor.execute(f"SELECT set_config('api.tenant_id', '{tenant_id}', TRUE);")
+            # TODO
+            # try:
+            #     uuid.UUID(tenant_id)
+            # except ValueError:
+            #     raise ValidationError("Tenant ID must be a valid UUID")
```
AdriiiPRodri: Can we remove this code?

api/src/backend/api/decorators.py
```diff
@@ -43,9 +43,13 @@ def wrapper(*args, **kwargs):
             tenant_id = kwargs.pop("tenant_id")
         except KeyError:
             raise KeyError("This task requires the tenant_id")
-
+        # TODO
+        # try:
+        #     uuid.UUID(tenant_id)
+        # except ValueError:
+        #     raise ValidationError("Tenant ID must be a valid UUID")
```
AdriiiPRodri: Same as before

AdriiiPRodri (approved):
LGTM

PR #6088, opened 2024-12-09
fix(db-manager): set health check and conn_max_age to 1 hour

jfagoagas:
### Context

Django is revoking connections with the following exception: `django.db.utils.InterfaceError: connection already closed`

### Description

- Configure `CONN_MAX_AGE` for an hour instead of 0 which automatically closes the connection.

### Checklist

### License


jfagoagas:
Closing this since the issue was not introduced by Django.

PR #6109, opened 2024-12-10
fix(deploy): temporal fix for the alpine-python segmentation fault

AdriiiPRodri:
### Context

There is a problem with the current version of Python with Alpine 3.21, according to our tests there is a segmentation fault when building images that depend on this base: https://github.com/docker-library/python/issues/993

### Description

We have decided to fix the Alpine versions to 3.20.

### Checklist

### License


jfagoagas (approved):
🔝

PR #6111, opened 2024-12-10
fix(autoscaling): `autoscaling_group_launch_configuration_requires_imdsv2` fails if Launch Template is used

### LINKED ISSUE #6047: The control autoscaling_group_launch_configuration_requires_imdsv2 fails if AWS Launch Template is used
### Steps to Reproduce

* Create a Launch Template with the required option imdsv2
* Create an ASG using this template
* Run the Prowler scan
```
prowler aws -c autoscaling_group_launch_configuration_requires_imdsv2
```

### Expected behavior

If I use a Launch Template instead of a Launch Configuration for creating an ASG the check should not fail

image

### Environment Resource

Workstation

### OS used

Windows

### Prowler version

Prowler 5.0.0 (You are running the latest version, yay!)

### Pip version

pip 24.2 (python 3.12)

### Context

_No response_

danibarranqueroo:
### Context

Fix #6047

If you use a Launch Template instead of a Launch Configuration, the check `autoscaling_group_launch_configuration_requires_imdsv2` will return a fail by default. I have modified both this check and the check `autoscaling_group_launch_configuration_no_public_ip`, because in both cases, when there is no Launch Configuration in the Auto Scaling group, the default report status is incorrect.

For the first check, it returned a FAIL (which is what was reported in the issue), and for the second check, it incorrectly reported a PASS, which should not occur when there is no Launch Configuration. The solution here was to avoid returning findings if there is no Launch Configuration. With this fix, launching an Auto Scaling group with a Launch Template will result in no findings.

To handle the control for IMDSv2 required in Auto Scaling groups that use a Launch Template, I have created a new check called `ec2_launch_template_imdsv2_required`.

### Description

Modified `autoscaling_group_launch_configuration_requires_imdsv2` and `autoscaling_group_launch_configuration_no_public_ip` checks, modify `ec2_service` and create new check `ec2_launch_template_imdsv2_required` with its unit tests.

### Checklist

### License


MrCloudSec (changes requested):
Please, create another PR for the new check, and only include here the fix of the existing check.

tests/providers/aws/services/autoscaling/autoscaling_group_launch_configuration_no_public_ip/autoscaling_group_launch_configuration_no_public_ip_test.py
```diff
@@ -171,7 +171,7 @@ def test_groups_without_launch_configuration(self):
             AvailabilityZones=["us-east-1a", "us-east-1b"],
         )
 
-        autoscaling_group_arn = autoscaling_client.describe_auto_scaling_groups(
+        autoscaling_client.describe_auto_scaling_groups(
             AutoScalingGroupNames=[autoscaling_group_name]
         )["AutoScalingGroups"][0]["AutoScalingGroupARN"]
```
MrCloudSec:
```suggestion
```

tests/providers/aws/services/autoscaling/autoscaling_group_launch_configuration_requires_imdsv2/autoscaling_group_launch_configuration_requires_imdsv2_test.py
```diff
@@ -119,7 +119,7 @@ def test_groups_without_launch_configuration(self):
             AvailabilityZones=["us-east-1a", "us-east-1b"],
         )
 
-        autoscaling_group_arn = autoscaling_client.describe_auto_scaling_groups(
+        autoscaling_client.describe_auto_scaling_groups(
             AutoScalingGroupNames=[autoscaling_group_name]
         )["AutoScalingGroups"][0]["AutoScalingGroupARN"]
```
MrCloudSec:
```suggestion
```

MrCloudSec (approved):

PR #6119, opened 2024-12-10
fix(aws): get firewall manager managed rule groups

### LINKED ISSUE #6030: False positive on `wafv2_webacl_with_rules` when when ACL provisioned with AWS Firewall Manager
### Steps to Reproduce

1. Depoloy ACL with AWS Firewall Manager including any rule group
2. Run prowler against AWS account where ACL is provisioned with Firewall Manager
3. Prowler will add that ACL as failed because the list of managed rules is empty, but it has AWS Firewall Manager provisioned rules

### Expected behavior

it's expected that ACL is managed centrally and may not have rules added directly

image

image

### Environment Resource

4. EKS

### OS used

1. Amazon Linux 2023

### Prowler version

5.0.0

### Pip version

24.0

### Context

_No response_

HugoPBrito:
### Context

`Firewall Manager` managed `rule groups` were previously not being accounted for, resulting in false positives.

Fix #6030

### Description

This issue has now been corrected, and false positives should no longer occur.

### Checklist

### License

### DIFF
--- a/prowler/providers/aws/services/wafv2/wafv2_service.py
+++ b/prowler/providers/aws/services/wafv2/wafv2_service.py
@@ -150,6 +150,22 @@ class WAFv2(AWSService):
                         else:
                             acl.rules.append(new_rule)
 
+                    firewall_manager_managed_rg = get_web_acl.get("WebACL", {}).get(
+                        "PreProcessFirewallManagerRuleGroups", []
+                    ) + get_web_acl.get("WebACL", {}).get(
+                        "PostProcessFirewallManagerRuleGroups", []
+                    )
+
+                    for rule in firewall_manager_managed_rg:
+                        acl.rules.append(
+                            Rule(
+                                name=rule.get("Name", ""),
+                                cloudwatch_metrics_enabled=rule.get(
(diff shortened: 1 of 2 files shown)


prowler/providers/aws/services/wafv2/wafv2_service.py
```diff
@@ -150,6 +150,22 @@ def _get_web_acl(self, acl: str):
...
                             acl.rules.append(new_rule)
 
+                    firewall_manager_managed_rg = get_web_acl.get("WebACL", {}).get(
+                        "PreProcessFirewallManagerRuleGroups", []
+                    ) + get_web_acl.get("WebACL", {}).get(
+                        "PostProcessFirewallManagerRuleGroups", []
+                    )
+
+                    for rule in firewall_manager_managed_rg:
+                        acl.rules.append(
```
MrCloudSec:
```suggestion
                        acl.rule_groups.append(
```

MrCloudSec (changes requested):
Awesome! Please, check my comment and add a test in the service for this use case. Thanks!

MrCloudSec (approved):

PR #6122, opened 2024-12-10
fix(compliance_tables): add correct values for findings

pedrooot:
### Description

Add correct values for compliance specific tables. Percentage should be calculated depending on the findings associated with the compliance.

### Checklist

### License


MrCloudSec (approved):
Awesome !!!

PR #6149, opened 2024-12-11
fix(azure): custom Prowler Role for Azure assignableScopes

puchy22:
### Context

The current assignableScopes in the Prowler Custom Role for Azure is / which is not accepted, it should be indicated by the user, something like /"subscriptions" OR "providers/Microsoft.Management/managementGroups"/{Your Subscription or Management Group ID} should work.

### Description

Change the custom role in the JSON role file definition and let a note in documentation.

### Checklist

### License


jfagoagas (approved):
Thanks for this fix @puchy22 👏

PR #6159, opened 2024-12-12
feat(ec2): add new fixer `ec2_instance_port_cifs_exposed_to_internet_fixer`

danibarranqueroo:
### Context

Developed a new fixer that updates the security group rules for EC2 instances to restrict access to the CIFS/SMB port, ensuring it is only accessible from trusted IP addresses or private networks. This will prevent unauthorized access and reduce the risk of data leaks and potential malware infections.

In order to do this, I have followed the same structure of the check but at the moment that the report is done on the check (there is a FAIL) in the fixer I use the `revoke_security_group_ingress` boto3 call.

### Description

Added new fixer `ec2_instance_port_cifs_exposed_to_internet_fixer` with its unit tests.

### Checklist

### License


MrCloudSec (approved):
Great work!

PR #6164, opened 2024-12-12
feat(s3): add new fixer `s3_bucket_public_access_fixer`

danibarranqueroo:
### Context

Developed a new fixer that modifies the permissions on S3 buckets to restrict public access. This includes setting up the Public Access Block configuration for the bucket, which prevents both public reads and writes to the bucket. The fixer will ensure that only authorized users or roles have access to the contents of the S3 bucket.

This fixer could be done in two ways, modifying the public access block (what I've done is this) and when you block all public access the ACL grantee will automatically be private and the bucket will no longer be public. Other way to fix this is setting the public ACL to private, the same as it's done on the fixer `s3_bucket_public_write_acl`. Both options fix the public access and I've used one in the other fixer and the other on this fixer due to the name of check, but if the reviewer thinks an option is better than the other I'll change what is needed.

### Description

Added new fixer `s3_bucket_public_access_fixer` with its unit tests.

### Checklist

### License

### DIFF
--- a/prowler/providers/aws/services/s3/s3_bucket_public_access/s3_bucket_public_access_fixer.py
+++ b/prowler/providers/aws/services/s3/s3_bucket_public_access/s3_bucket_public_access_fixer.py
@@ -6,7 +6,7 @@ def fixer(resource_id: str, region: str) -> bool:
     """
     Modify the S3 bucket's public access settings to block all public access.
     Specifically, this fixer configures the bucket's public access block settings to
-    prevent any public access (ACLs and policies). Requires the s3:PutBucketAcl
+    prevent any public access (ACLs and policies). Requires the s3:PutBucketPublicAccessBlock
     permission to modify the public access settings.
     Permissions:
     {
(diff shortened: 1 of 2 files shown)


prowler/providers/aws/services/s3/s3_bucket_public_access/s3_bucket_public_access_fixer.py
```diff
@@ -0,0 +1,46 @@
+from prowler.lib.logger import logger
+from prowler.providers.aws.services.s3.s3_client import s3_client
+
+
+def fixer(resource_id: str, region: str) -> bool:
```
MrCloudSec, line 5: Check this docstring and the permission that is needed.

MrCloudSec (approved):

prowler/providers/aws/services/s3/s3_bucket_public_access/s3_bucket_public_access_fixer.py
MrCloudSec: Fixed.

PR #6211, opened 2024-12-16
chore(rls): Add tenant_id filters in views and improve querysets

jfagoagas:
### Description

Improve queries using RLS policies adding the `tenant_id` filter.

### ToDo
- [x] Review the rest of locations where queries are being done with/without filters.

### Checklist

### License


api/src/backend/api/v1/views.py
```diff
@@ -945,7 +953,7 @@ def connection(self, request, pk=None):
         get_object_or_404(Provider, pk=pk)
         with transaction.atomic():
             task = check_provider_connection_task.delay(
-                provider_id=pk, tenant_id=request.tenant_id
+                provider_id=pk, tenant_id=self.request.tenant_id
```
jfagoagas, line 956: Why `self.request` instead of the `request` passed?

vicferpoy: Consistency across the module. It's the same.

AdriiiPRodri (approved):

Issue #5710, opened 2024-11-09
merge custom config with prowler's default config


kagahd:
### New feature motivation

When using prowler with the option `--config-file`, the given config file **replaces** prowler's own default config file. Therefore, users have to merge their config with prowler's config on every new release of prowler which is not only quite tedious but can also be forgotten. 
Usually, users put in their custom config only those values that they want to match their specific requirements. These values should just **overwrite** prowlers default config values. 
It is counterintuitive to require the user to provide a complete configuration that is already provided by prowler when they only want to specify a few custom configuration values.

### Solution Proposed

I suggest that prowler **merges** both configs by giving precedence the values of the custom config file passed by `--config-file`. This behaviour must be documented.

### Describe alternatives you've considered

I don't see any alternative. At the very least, the behavior of how a user-defined configuration file is used by prowler, i.e. overwriting or replacing, and also the consequences when upgrading prowler should be clearly documented.

jfagoagas:
Hello @kagahd, we were thinking about this when we realised that Prowler's config has default values in the places where the configuration values are used, so I think that if you just pass your config the rest of values should be defined as the default ones. Could you please verify it?

If this is not what you are suggesting, please let me know.

Thanks!

kagahd:
Hello @jfagoagas, thanks for your feedback!

> so I think that if you just pass your config the rest of values should be defined as the default ones. Could you please verify it?

Yes, I just verified it for the check `awslambda_function_using_supported_runtimes`:
If I use my custom `config.yaml` by using prowlers `--config-file` parameter, the check finds obsolete runtimes **only if** my custom config contains the `obsolete_lambda_runtimes` array with the obsolete runtimes. 
As soon as I remove `obsolete_lambda_runtimes` from my custom config, the check `awslambda_function_using_supported_runtimes` **finds nothing** anymore.

That's why [I asked in the prowler forum](https://prowler-workspace.slack.com/archives/C0451NDLC4X/p1731087123908399) if prowlers parameter `--config-file` **replaces** prowler's config instead to merge both. 
Since your colleague @drewkerrigan confirmed my assumption, I opened this feature request.

jfagoagas:
Hello @kagahd, you're right and that's an issue in Prowler because each parameter should have the default value configured regardless if the config is passed/merged or not.

We will review and get back to you once we have an update.

Thanks!

puchy22:
[...]

PR #6225, opened 2024-12-17
chore(config): set default values for empty config fields

pedrooot:
### Context

Related with #5710

### Description

This PR set the correct values for the config when it's used on the checks and the field is missing on the `config.yaml` file.

### Checklist

### License


MrCloudSec (changes requested):
Good job @pedrooot! Can you please change `original_*` to `default_*`, please?

MrCloudSec (approved):

PR #6226, opened 2024-12-17
ref(rbac): improve rbac implementation for views

AdriiiPRodri:
### Context

Some endpoints were not properly protected with RBAC, in this PR we seek to add such protection.

### Description

In this PR we have added the missing permissions for some endpoints. In addition, we perform a refactor to the RBAC-views related code.

### Checklist

### License

### DIFF
--- a/api/src/backend/api/base_views.py
+++ b/api/src/backend/api/base_views.py
@@ -35,10 +35,10 @@ class BaseViewSet(ModelViewSet):
         """
         Sets required_permissions before permissions are checked.
         """
-        self.get_required_permissions()
+        self.set_required_permissions()
         super().initial(request, *args, **kwargs)
 
-    def get_required_permissions(self):
+    def set_required_permissions(self):
         """This is an abstract method that must be implemented by subclasses."""
         NotImplemented
 
--- a/api/src/backend/api/base_views.py
+++ b/api/src/backend/api/base_views.py
@@ -11,11 +11,13 @@ from api.db_router import MainRouter
 from api.db_utils import POSTGRES_USER_VAR, rls_transaction
 from api.filters import CustomDjangoFilterBackend
 from api.models import Role, Tenant
+from api.rbac.permissions import HasPermissions
 
 
 class BaseViewSet(ModelViewSet):
     authentication_classes = [JWTAuthentication]
-    permission_classes = [permissions.IsAuthenticated]
+    required_permissions = []
+    permission_classes = [permissions.IsAuthenticated, HasPermissions]
     filter_backends = [
         filters.QueryParameterValidationFilter,
         filters.OrderingFilter,
@@ -28,6 +30,17 @@ class BaseViewSet(ModelViewSet):
 
     ordering_fields = "__all__"
     ordering = ["id"]
+
+    def initial(self, request, *args, **kwargs):
+        """
+        Sets required_permissions before permissions are checked.
+        """
+        self.get_required_permissions()
+        super().initial(request, *args, **kwargs)
+
+    def get_required_permissions(self):
+        """This is an abstract method that must be implemented by subclasses."""
+        NotImplemented
 
     def get_queryset(self):
         raise NotImplementedError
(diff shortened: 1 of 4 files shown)


api/src/backend/api/v1/views.py
```diff
@@ -290,26 +290,19 @@ class UserViewSet(BaseUserViewset):
...
         Returns the required permissions based on the request method.
         """
         if self.action == "me":
             # No permissions required for me request
-            return []
+            self.required_permissions = []
         else:
             # Require permission for the rest of the requests
-            return [Permissions.MANAGE_USERS]
+            self.required_permissions = [Permissions.MANAGE_USERS]
```
vicferpoy, line 305: I believe this is not correct. This function should return something, not update a internal attribute. Why did you modify the former logic? Since you are setting the base required permissions as `required_permissions = [Permissions.MANAGE_USERS]`, I believe this function should be something like:

```python
    def get_required_permissions(self):
        """
        Returns the required permissions based on the request method.
        """
        if self.action == "me":
            # No permissions required for me request
            return []
        else:
            # Require permission for the rest of the requests
            return self.required_permissions
```

api/src/backend/api/v1/views.py
```diff
@@ -1349,15 +1322,8 @@ class FindingViewSet(BaseRLSViewSet):
...
-    permission_classes = BaseRLSViewSet.permission_classes + [HasPermissions]
-
-    def initial(self, request, *args, **kwargs):
-        """
-        Sets required_permissions before permissions are checked.
-        """
-        self.required_permissions = ResourceViewSet.required_permissions
-        super().initial(request, *args, **kwargs)
+    # RBAC required permissions
+    required_permissions = [Permissions.MANAGE_PROVIDERS, Permissions.MANAGE_SCANS]
```
vicferpoy: Why `Permissions.MANAGE_SCANS`?

This is a readonly viewset. Same with overviews. Maybe I'm missing something.

api/src/backend/api/base_views.py
```diff
@@ -29,6 +31,17 @@ class BaseViewSet(ModelViewSet):
...
+    def initial(self, request, *args, **kwargs):
+        """
+        Sets required_permissions before permissions are checked.
+        """
+        self.get_required_permissions()
+        super().initial(request, *args, **kwargs)
+
+    def get_required_permissions(self):
+        """This is an abstract method that must be implemented by subclasses."""
+        NotImplemented
```
vicferpoy: There are a few viewsets that do not implement this method. How is it working?

vicferpoy (changes requested):
I left some questions regarding implementation.

api/src/backend/api/base_views.py
AdriiiPRodri: They use their required_permissions parameter, so the function get_required_permissions does not return anything, maybe the interesting thing would be to change the name of this function to set_required_permissions.

api/src/backend/api/v1/views.py
AdriiiPRodri: Maybe we don't have the same idea behind the permission, I think we should have a common place where this is defined, we can discuss it but in principle in that endpoint you can see findings and for me the findings are related to scans.

api/src/backend/api/v1/views.py
AdriiiPRodri: Making this change implies modifying code and having to define the function get_required_permissions in all views, for me the interesting thing would be to change the name to set_required_permissions.

api/src/backend/api/base_views.py
vicferpoy: Yes, I agree. Thanks for clarifying.

api/src/backend/api/v1/views.py
vicferpoy: I thought this was one of the Django viewset functions. If it is auxiliary and our own, it's fine. Maybe consider renaming it as you suggested in other comment.

vicferpoy (approved):
LGTM! Good job

PR #6230, opened 2024-12-17
fix(.env): remove comment

jfagoagas:
### Description

Comments are being interpreted as the environment value.

### Checklist

### License

### DIFF
--- a/.env
+++ b/.env
@@ -41,8 +41,10 @@ DJANGO_LOGGING_FORMATTER=human_readable
 # Applies to both Django and Celery Workers
 DJANGO_LOGGING_LEVEL=INFO
 DJANGO_WORKERS=4 # Defaults to the maximum available based on CPU cores if not set.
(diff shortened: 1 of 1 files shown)


MrCloudSec (dismissed):

.env
```diff
@@ -41,8 +41,10 @@ DJANGO_LOGGING_FORMATTER=human_readable
 # Applies to both Django and Celery Workers
 DJANGO_LOGGING_LEVEL=INFO
 DJANGO_WORKERS=4 # Defaults to the maximum available based on CPU cores if not set.
```
vicferpoy: What about this one?

We should review the whole `.env` file.

vicferpoy (changes requested):

.env
jfagoagas: Done!

vicferpoy (approved):

PR #6238, opened 2024-12-18
ref(rbac): enable relationship creation when objects is created

AdriiiPRodri:
### Context

The POST and PATCH doesn't allow the relationships creation. In this PR we have enabled it.

### Checklist

### License

### DIFF
--- a/api/src/backend/api/v1/serializers.py
+++ b/api/src/backend/api/v1/serializers.py
@@ -235,13 +235,10 @@ class UserCreateSerializer(BaseWriteSerializer):
 
 class UserUpdateSerializer(BaseWriteSerializer):
     password = serializers.CharField(write_only=True, required=False)
-    roles = serializers.ResourceRelatedField(
-        queryset=Role.objects.all(), many=True, required=False
-    )
 
     class Meta:
         model = User
-        fields = ["id", "name", "password", "email", "company_name", "roles"]
+        fields = ["id", "name", "password", "email", "company_name"]
         extra_kwargs = {
             "id": {"read_only": True},
         }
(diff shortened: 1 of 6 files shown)


api/src/backend/api/v1/serializers.py
```diff
@@ -235,10 +235,13 @@ def create(self, validated_data):
 
 class UserUpdateSerializer(BaseWriteSerializer):
     password = serializers.CharField(write_only=True, required=False)
+    roles = serializers.ResourceRelatedField(
+        queryset=Role.objects.all(), many=True, required=False
+    )
```
vicferpoy: Remove this from this serializer.

api/src/backend/api/v1/serializers.py
```diff
@@ -235,10 +235,13 @@ def create(self, validated_data):
...
 
     class Meta:
         model = User
-        fields = ["id", "name", "password", "email", "company_name"]
+        fields = ["id", "name", "password", "email", "company_name", "roles"]
```
vicferpoy: Same here.

api/src/backend/api/v1/serializers.py
```diff
@@ -475,21 +483,94 @@ class Meta:
...
         model = ProviderGroup
-        fields = ["id", "name"]
+        fields = [
+            "id",
+            "name",
+            "inserted_at",
+            "updated_at",
+            "providers",
+            "roles",
+            "url",
```
vicferpoy: This can only be present in read serializers.

vicferpoy (changes requested):

vicferpoy (approved):

PR #6264, opened 2024-12-19
refactor(mutelist): use jsonschema on mutelist

pedrooot:
### Description

Use `jsonschema` in mutelist.

### Checklist

### License


prowler/lib/mutelist/models.py
```diff
@@ -1,24 +0,0 @@
...
-                        "Regions": list,
-                        "Resources": list,
-                        Optional("Tags"): list,
-                        Optional("Exceptions"): {
-                            Optional("Accounts"): list,
-                            Optional("Regions"): list,
-                            Optional("Resources"): list,
-                            Optional("Tags"): list,
-                        },
-                        Optional("Description"): str,
```
jfagoagas, line 18: Can you verify how these fields can be set as optional in jsonschema?

jfagoagas (changes requested):

prowler/lib/mutelist/models.py
pedrooot: Done in the next commit!

jfagoagas (approved):
Awesome 🚀

PR #6283, opened 2024-12-20
fix(db-utils): fix batch_delete function

vicferpoy:
### Description

When using `Paginator`, it would not iterate over more objects than the value of the `batch_size`.

We have fixed this and included a proper test for it.

### License


AdriiiPRodri (dismissed):
LGTM

AdriiiPRodri (approved):

PR #6284, opened 2024-12-20
fix(users): fix /users/me behavior when having more than 1 users in the same tenant

vicferpoy:
### Context

We were retrieving the first user ordered by memberships in the `GET /users/me` endpoint, leading to incorrect behavior when having more than user in the same tenant.

### Description

It has been fixed to properly return the requesting user info.

#### User1 info

image

#### User2 info

image

### License


AdriiiPRodri (approved):
LGTM

PR #6351, opened 2025-01-03
fix(Azure TDE): add filter for master DB

johannes-engler-mw:
### Context

The check for TDE enabled DBs on Azure SQL servers fails everytime because the MASTER DB cant be TDE enabled.

### Description

This PR fixes the Azure SQL TDE Encryption check, which fails on the MASTER DB. The MASTER DB is existing on any SQL DB in Azure and cant be TDE enabled, so the check will fail every time.

Also it adds an exclusion for the Vulture pre-commit hook, which skips the api/src/backend/ folder.

### Checklist

### License


.pre-commit-config.yaml
```diff
@@ -103,7 +103,7 @@ repos:
       - id: vulture
         name: vulture
         description: "Vulture finds unused code in Python programs."
-        entry: bash -c 'vulture --exclude "contrib" --min-confidence 100 .'
```
puchy22: This change should not be in this PR, we know that with the new app the pre-commits hooks should be improved but it should be done in a different PR.

prowler/providers/azure/services/sqlserver/sqlserver_tde_encryption_enabled/sqlserver_tde_encryption_enabled.py
```diff
@@ -12,6 +12,8 @@ def execute(self) -> Check_Report_Azure:
                 )
                 if len(databases) > 0:
                     for database in databases:
+                        if database.name.lower() == "master":
+                            continue
```
puchy22, line 16: Please, could you create a test for this case?

puchy22 (changes requested):

.pre-commit-config.yaml
johannes-engler-mw: The problem is that without that change I cant commit because Vulture is failing in the API directory.

prowler/providers/azure/services/sqlserver/sqlserver_tde_encryption_enabled/sqlserver_tde_encryption_enabled.py
johannes-engler-mw: Will do.

johannes-engler-mw: is implemented, please review

.pre-commit-config.yaml
puchy22: Sorry for the inconvenience, while we work on it you can use the `--no-verify` argument to skip the pre-commit when you need it. For the rest of the PR it seems correct, so when you remove this I'll approve it.

puchy22: Hi @johannes-engler-mw, the problem with pre-commit have been solved in the PR #6393.

johannes-engler-mw: HI @puchy22, merged the new version but still had to skip verification because of a security issue:
```
 VULNERABILITIES REPORTED
+==============================================================================+

-> Vulnerability found in pyjwt version 2.9.0
   Vulnerability ID: 74429
   Affected spec: <2.10.1
   ADVISORY: Affected versions of pyjwt are vulnerable to Partial
   Comparison (CWE-187). This flaw allows attackers to bypass issuer (iss)...
   CVE-2024-53861
   For more information about this vulnerability, visit
   https://data.safetycli.com/v/74429/97c
   To ignore this vulnerability, use PyUp vulnerability id 74429 in safety’s
   ignore command-line argument or add the ignore to your safety policy file.
```

Anyway I have removed my changes from the pre-commit config.

puchy22 (approved):
Nice catch! 🥇

MrCloudSec (approved):

PR #6353, opened 2025-01-03
docs: add new format CloudFormation for ResourceType in check metadata

puchy22:
### Context

Add to Metadata doc that in ResourceType is accepted the [CloudFormation types](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-template-resource-type-ref.html) but removing the :: and only using capital letters at the beginning of each part

### Description

Changed doc to indicate this new CloudFormation format.

### Checklist

### License

### DIFF
--- a/docs/developer-guide/checks.md
+++ b/docs/developer-guide/checks.md
@@ -279,7 +279,7 @@ Each Prowler check has metadata associated which is stored at the same level of 
   "Severity": "critical",
   # ResourceType only for AWS, holds the type from here
   # https://docs.aws.amazon.com/securityhub/latest/userguide/asff-resources.html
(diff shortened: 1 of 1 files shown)


docs/developer-guide/checks.md
```diff
@@ -279,6 +279,9 @@ Each Prowler check has metadata associated which is stored at the same level of
   "Severity": "critical",
   # ResourceType only for AWS, holds the type from here
   # https://docs.aws.amazon.com/securityhub/latest/userguide/asff-resources.html
+  # In case of not exists, use CloudFormation type but removing the "::" and using capital letters only at the beginning of each word. Example: "AWS::EC2::Instance" -> "AwsEc2Instance"
```
MrCloudSec:
```suggestion
  # In case of not existing, use CloudFormation type but removing the "::" and using capital letters only at the beginning of each word. Example: "AWS::EC2::Instance" -> "AwsEc2Instance"
```

MrCloudSec (approved):

PR #6398, opened 2025-01-09
fix(cloudformation): fix flaky tests

danibarranqueroo:
### Context

Pepe noticed that some tests from CloudFormation service are failing, I've made an easy solution continuing using MagicMock to see if they stop failing, if this does not work it’ll be necessary to change them to Moto (if they are covered) or Botocore.

### Description

Modified tests from all the CloudFormation checks.

### Checklist

### License


jfagoagas (approved):
It works!

PR #6460, opened 2025-01-13
feat(finding): add first_seen attribute

AdriiiPRodri:
### Description

The findings that we currently store in the database have a `delta` parameter that indicates among other things whether the finding is `new` or already existed previously in the system.

In case it is new the `first_seen` parameter will be set to the date when the finding was created. In case the `delta` parameter is not `new` the first_seen attribute will have the value of the first_seen attribute of the finding with the same `uid` and `delta = “new”`.

### Checklist

### License


api/src/backend/api/migrations/0006_findings_first_seen.py
```diff
@@ -0,0 +1,58 @@
+from django.db import migrations, models
+
+
+def populate_first_seen(apps, schema_editor):
```
jfagoagas: Do we want to backfill all previous findings? This blocks the app startup while updating all the findings.

api/src/backend/api/migrations/0006_findings_first_seen.py
```diff
@@ -0,0 +1,58 @@
...
+    # Then update all other findings
+    schema_editor.execute(
+        """
+        UPDATE findings f1
+        SET first_seen = (
+            SELECT MIN(f2.inserted_at)
+            FROM findings f2
+            WHERE f2.uid = f1.uid
+            AND f2.delta = 'new'
+        )
```
jfagoagas: I don't know if the following query will be simpler and more performant but we only need to get the first one ordering by `finding.id` and limiting the result to 1, because as is using UUIDv7 the timestamp is present, so with that you wouldn't need the `MIN()` to get it.

api/src/backend/api/migrations/0006_findings_first_seen.py
```diff
@@ -0,0 +1,58 @@
...
+
+    operations = [
+        # Add first_seen field as nullable initially
+        migrations.AddField(
+            model_name="finding",
+            name="first_seen",
+            field=models.DateTimeField(null=True),
+        ),
+        # Populate the data
+        migrations.RunPython(populate_first_seen),
```
jfagoagas: Not sure about adding this, what do you think @drewkerrigan @vicferpoy ?

api/src/backend/tasks/jobs/scan.py
```diff
@@ -237,6 +237,14 @@ def perform_prowler_scan(
 
                     status = FindingStatus[finding.status]
                     delta = _create_finding_delta(last_status, status)
+                    first_seen = datetime.now(tz=timezone.utc)
+                    # When the delta attribute is different from new it means that previously there was a finding with the same uid and delta new, so we will obtain the first_seen attribute of that first finding
+                    if delta != Finding.DeltaChoices.NEW:
```
jfagoagas: If the backfill is not included this should be changed to always get the previous finding `first_seen` if the `delta != new`.

jfagoagas (dismissed):
That's a great job @AdriiiPRodri please review my comments when you get a chance, thanks!

api/src/backend/tasks/jobs/scan.py
```diff
@@ -237,6 +237,14 @@ def perform_prowler_scan(
...
+                        first_seen = (
+                            Finding.objects.filter(uid=finding_uid, delta=Finding.DeltaChoices.NEW.value)
+                            .values("first_seen")
+                            .first()["first_seen"]
+                        )
```
vicferpoy: Can't we move this logic inside the `_create_finding_delta` (consider renaming it if necessary) to avoid making extra queries when not needed? In that function, we retrieve the latest finding with the same `uid` to get the `delta` value. If that finding's `first_seen` is not `None`, we can just copy it from there without having to query again.

vicferpoy (changes requested):
In addition to the requested change:

- We are calling all `date` fields with `*_at`, right? Why isn't this field called `first_seen_at`?
- Adapt unit tests when you implement this logic into the `delta` function to also test this. We need to cover all situations.

vicferpoy (changes requested):
In addition to the question, I'm still missing unit tests.

api/src/backend/api/models.py
```diff
@@ -615,6 +615,7 @@ class DeltaChoices(models.TextChoices):
     id = models.UUIDField(primary_key=True, default=uuid7, editable=False)
     inserted_at = models.DateTimeField(auto_now_add=True, editable=False)
     updated_at = models.DateTimeField(auto_now=True, editable=False)
+    first_seen_at = models.DateTimeField(editable=False)
```
vicferpoy: What happens when we perform this migration on an already existing database with findings? What value do the older findings take? I believe we are missing, at least, `null=True` here and in the migration.

vicferpoy: > Sorry, this comment was meant to be added with the last review but I missed it somehow.

vicferpoy (changes requested):

AdriiiPRodri:
> In addition to the question, I'm still missing unit tests.

Due to the priority of other tasks it has been decided to keep pending the tests to evaluate this new field, we will do these tests as soon as we finish the more priority tasks.

vicferpoy (approved):

PR #6470, opened 2025-01-13
fix(rbac): block admin role deletion

AdriiiPRodri:
### Description

In this PR we have fixed the bug that caused that the administrator role could be removed. In addition to this we have added tests to confirm that this cannot happen.

### Checklist

### License

### DIFF
--- a/api/src/backend/api/v1/views.py
+++ b/api/src/backend/api/v1/views.py
@@ -1636,6 +1636,17 @@ class RoleViewSet(BaseRLSViewSet):
             request.data["manage_account"] = str(user_role.manage_account).lower()
         return super().partial_update(request, *args, **kwargs)
 
+    def destroy(self, request, *args, **kwargs):
+        instance = self.get_object()
+        serializer = self.get_serializer(
+            instance,
+            data=request.data,
+            partial=True,
+        )
+        serializer.is_valid(raise_exception=True)
+
+        return super().destroy(request, *args, **kwargs)
+
 
 @extend_schema_view(
     create=extend_schema(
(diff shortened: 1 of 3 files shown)


api/src/backend/api/v1/views.py
```diff
@@ -1636,6 +1636,17 @@ def partial_update(self, request, *args, **kwargs):
             request.data["manage_account"] = str(user_role.manage_account).lower()
         return super().partial_update(request, *args, **kwargs)
 
+    def destroy(self, request, *args, **kwargs):
+        instance = self.get_object()
+        serializer = self.get_serializer(
+            instance,
+            data=request.data,
+            partial=True,
```
vicferpoy: Why is the partial part needed?

vicferpoy (changes requested):

api/src/backend/api/v1/views.py
AdriiiPRodri: If I remove the partial then we will get this error:

```
{
    "errors": [
        {
            "detail": "This field is required.",
            "status": "400",
            "source": {
                "pointer": "/data/attributes/name"
            },
            "code": "required"
        }
    ]
}
```

So adding `partial=True` makes that we don't require fields

vicferpoy: I think the issue is that serializer validation is not suitable here. This is a patch to reutilize code in my opinion, but not a good use of serializers since they expect a payload and DESTROY methods do not include that. I believe the correct approach would be to perform the check at view level, like we've done in the past. For instance, invitations:

```python
    def destroy(self, request, *args, **kwargs):
        instance = self.get_object()
        if instance.state != Invitation.State.PENDING:
            raise ValidationError(detail="This invitation cannot be revoked.")
        instance.state = Invitation.State.REVOKED
        instance.save()
        return Response(status=status.HTTP_204_NO_CONTENT)
```

AdriiiPRodri: This was done to continue with the deployment of RBAC and to minimize the time required. I am going to apply this change as it will not take much time.

vicferpoy (approved):

PR #6471, opened 2025-01-13
ref(rbac): disable some checks

AdriiiPRodri:
### Description

In this PR we have disabled some permissions since we are not using it at this moment.

### Checklist

### License

### DIFF
--- a/api/src/backend/api/v1/serializers.py
+++ b/api/src/backend/api/v1/serializers.py
@@ -1416,8 +1416,8 @@ class RoleSerializer(RLSSerializer, BaseWriteSerializer):
             "id",
             "name",
             "manage_users",
+            "manage_account",
             # Disable for the first release
-            # "manage_account",
             # "manage_billing",
             # "manage_integrations",
             # /Disable for the first release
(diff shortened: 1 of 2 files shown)


api/src/backend/api/v1/serializers.py
```diff
@@ -1416,10 +1416,12 @@ class Meta:
             "id",
             "name",
             "manage_users",
-            "manage_account",
-            "manage_billing",
+            # Disable for the first release
+            # "manage_account",
+            # "manage_billing",
+            # "manage_integrations",
+            # /Disable for the first release
```
vicferpoy: Is this enough? These permissions still exist in the database and are applied to some of our current views. What happens then?

For instance:

```python
class InvitationViewSet(BaseRLSViewSet):
    queryset = Invitation.objects.all()
    serializer_class = InvitationSerializer
    filterset_class = InvitationFilter
    http_method_names = ["get", "post", "patch", "delete"]
    search_fields = ["email"]
    ordering = ["-inserted_at"]
    ordering_fields = [
        "inserted_at",
        "updated_at",
        "expires_at",
        "state",
        "inviter",
    ]
    # RBAC required permissions
    required_permissions = [Permissions.MANAGE_ACCOUNT]

    def get_queryset(self):
        return Invitation.objects.filter(tenant_id=self.request.tenant_id)
```
 
Would this mean that unless users have all permissions (admin), they wouldn't be able to create invitations?

vicferpoy (changes requested):

vicferpoy (approved):

PR #6479, opened 2025-01-13
feat(kubernetes): add resource metadata to report

MrCloudSec:
### Description

Refactor the `Check_Report_Kubernetes` initialization for the listed services to pass the `resource_metadata` object directly.
Remove assignments for `resource_name`, `resource_id`, and `namespace` within each check.
Ensure the `Check_Report_Kubernetes` constructor dynamically extracts this information from `resource_metadata`.

### Checklist

### License


prowler/lib/check/models.py
```diff
@@ -413,7 +413,11 @@ class Check_Report:
     def __init__(self, metadata, resource=None):
         self.status = ""
         self.check_metadata = CheckMetadata.parse_raw(metadata)
-        self.resource_metadata = resource.dict() if resource else {}
+        self.resource_metadata = (
+            resource.dict()
+            if hasattr(resource, "dict")
+            else resource.to_dict() if hasattr(resource, "to_dict") else {}
+        )
```
pedrooot: Should we change this in order to handle dataclass?

prowler/lib/check/models.py
```diff
@@ -487,11 +491,17 @@ class Check_Report_Kubernetes(Check_Report):
...
+        super().__init__(metadata, resource_metadata)
+        self.resource_id = (
+            getattr(resource_metadata, "uid", None)
+            or getattr(resource_metadata, "name", None)
+            or ""
+        )
+        self.resource_name = getattr(resource_metadata, "name", "")
+        self.namespace = getattr(resource_metadata, "namespace", "cluster-wide")
+        if not self.namespace:
+            self.namespace = "cluster-wide"
```
pedrooot, line 541: Do we need this? `self.namespace = getattr(resource_metadata, "namespace", "cluster-wide")` will assign the "cluster-wide" value

prowler/providers/kubernetes/services/core/core_service.py
```diff
@@ -144,18 +142,16 @@ def _in_worker_node(self):
...
 
-@dataclass
-class Container:
+class Container(BaseModel):
     name: str
     image: str
     command: Optional[List[str]]
     ports: Optional[List[dict]]
     env: Optional[List[dict]]
-    security_context: Optional[V1SecurityContext]
```
pedrooot, line 154: Why this? Just asking

pedrooot (changes requested):
Do you think that we should change from:
`report = Check_Report_Kubernetes(self.metadata(), pod)`
to
`report = Check_Report_Kubernetes(metadata=self.metadata(), resource_metadata=pod)` ??

prowler/lib/check/models.py
MrCloudSec: Done

prowler/lib/check/models.py
MrCloudSec: Yes, since the attribute namespace could exist but it can be empty.

prowler/providers/kubernetes/services/core/core_service.py
MrCloudSec: It is necessary to change it to BaseModel.

pedrooot (approved):
🔝 Thanks for this!!

PR #6499, opened 2025-01-14
feat(prowler-role): Add templates to deploy it in AWS

jfagoagas:
### Description

Add Terraform and CloudFormation templates to deploy `ProwlerScan` AWS IAM Role to be used from Prowler to scan AWS Cloud Providers.

Key changes:
- `permissions/prowler-security-hub.json` deleted since it is already included in the Prowler additions policy.
- Changed `aws` by `*` in the `AllowAPIGatewayReadOnly` statement to cover all partitions.
- Terraform template uses `permissions/prowler-additions-policy.json` from the parent's folder.

### Checklist

### License


.gitignore
```diff
@@ -45,6 +45,7 @@ junit-reports/
 # Terraform
 .terraform*
 *.tfstate
+*.tfstate.*
```
MrCloudSec, line 48: Why not a single `*.tfstate*` line?

permissions/prowler-additions-policy.json
MrCloudSec, line 1: I think we have to review this and keep the `securityhub:BatchImportFindings` permission separated since it is not a read only permission.

.gitignore
jfagoagas: Both are needed for files ending in `tfstate`

permissions/prowler-additions-policy.json
jfagoagas: I think for now we should keep it since users are using this template.

MrCloudSec (approved):

PR #6440, opened 2025-01-09
feat(ec2): include resource metadata in Check_Report

MrCloudSec:
### Context

This PR refactors the `Check_Report` and `Check_Report_AWS` classes to dynamically handle `resource_metadata`. For now, the changes are applied only to the EC2 service, allowing EC2-specific details to be included in the `data` field of the OCSF output. This improves the context of findings for EC2 resources while maintaining flexibility for future expansion to other services.

### Description

- Added `resource_metadata` to `Check_Report`: Dynamically extracts attributes from the EC2 metadata object and serializes them into a dictionary.
- Updated `Check_Report_AWS`: Automatically populates `resource_id`, `resource_arn`, and region for EC2 findings using `resource_metadata`.
- Improved OCSF Output: Includes `resource_metadata` in the `data` field under `metadata`, providing enriched context for EC2 findings.
- Initialization Updates: EC2 `Check_Report_AWS` instances now require `resource_metadata` as part of their initialization to ensure consistent data handling.

### Checklist

- Are there new checks included in this PR? Yes / No
    - If so, do we need to update permissions for the provider? Please review this carefully.
- [ ] Review if the code is being covered by tests.
- [ ] Review if code is being documented following this specification https://github.com/google/styleguide/blob/gh-pages/pyguide.md#38-comments-and-docstrings
- [ ] Review if backport is needed.

### License

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #6505, opened 2025-01-14
feat(azure): include resource metadata in Check_Report_Azure

puchy22:
### Context

Due to the new way in which the report metadata is being generated in the PR #6440 for AWS, this PR does the same thing but for Azure.

### Description

Pass resource metadata to Check_Report and automate the extraction of basic attributes from resource metadata.

### Checklist

### License


MrCloudSec (approved):
👏🏼

PR #6527, opened 2025-01-15
feat(findings): Add resource_tag filters for findings endpoint

vicferpoy:
### Description

New filters have been added to the following endpoints:

- `GET /findings`
- `GET /findings/metadata`

image

In addition, the `GET /findings/metadata` endpoint will not return a list of values for each tag.

### Examples

Every filter is self explanatory except for the `filter[resource_tags]`, which expects a list of key:value pairs separated by commas:

image

> Same behavior for `GET /findings`

### License

### DIFF
--- a/api/src/backend/api/filters.py
+++ b/api/src/backend/api/filters.py
@@ -319,6 +319,27 @@ class FindingFilter(FilterSet):
         field_name="resources__type", lookup_expr="icontains"
     )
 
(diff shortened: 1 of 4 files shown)


jfagoagas:
We need to backport this PR to v5.1 since this feature was backported there and requires the code present here https://github.com/prowler-cloud/prowler/pull/6524

PR #6537, opened 2025-01-15
fix(detect_secrets): refactor logic for detect-secrets

### LINKED ISSUE #6467: report of DetectSecrets line number is wrong
### Steps to Reproduce

Run a check which uses DetectSecrets e.g. `ecs_task_definitions_no_environment_secrets` on a resource with potential secrets. The check will report the line number where the potential secret was found.
Prowler's `ecs_task_definitions_no_environment_secrets` metadata states:

> If a secret is detected, the line number shown in the finding matches with the environment variable `Name` attribute starting to count at the `environment` key from the ECS Task Definition in JSON format.

Given the following example the key `environment` of a AWS ECS task definition in json format which is shown in AWS console as follows:

```
"environment": [
  {
      "name": "FIRST_PARAM",
      "value": "first value"
  },
  {
      "name": "SECOND_PARAM",
      "value": "second value"
  },
  {
      "name": "POTENTIAL_SECRET",
      "value": "potential secret"
  }
]
```

If the third entry of the `environment` key of the task definition contains a potential secret, the check would indicate **line number four**. Why is it so? Because the json that prowler is passing to DetectSecrets is transformed as follows:

```
{
  "FIRST_PARAM": "first value",
  "SECOND_PARAM": "second value",
  "POTENTIAL_SECRET": "potential secret",
}
```
Obviously, DetectSecrets counts the opening curly bracket from the json expression as line number one.

### Expected behavior

I'd expect to read the correct line number in prowler's report.

doesn't matter

### Environment Resource

Docker container locally

### OS used

MacOS

### Prowler version

5.0.5

### Pip version

doesn't matter

### Context

_No response_

pedrooot:
### Context

Fix #6467
Thanks @kagahd for the heads up! 🚀 
### Description

This PR change the logic from the checks: `ecs_task_definitions_no_environment_secrets`, `awslambda_function_no_secrets_in_variables`
The main problem here is that we are re-writing the ecs env vars into a format that Detect-Secrets could find them and we were using the line from the output to create the `status_extended`. That's why the line was incorrect (it was taking the lines from the re-written variables).

The new approach changes the status extended from:
`...Secret keyword on line 2.`
to
`...Secret keyword on env var DB_PASSWORD.`

This way, the user will have the env name on the output.

### Checklist

### License

### DIFF
--- a/prowler/providers/aws/services/ecs/ecs_task_definitions_no_environment_secrets/ecs_task_definitions_no_environment_secrets.py
+++ b/prowler/providers/aws/services/ecs/ecs_task_definitions_no_environment_secrets/ecs_task_definitions_no_environment_secrets.py
@@ -1,3 +1,4 @@
+import hashlib
 from json import dumps
 
 from prowler.lib.check.models import Check, Check_Report_AWS
@@ -25,8 +26,16 @@ class ecs_task_definitions_no_environment_secrets(Check):
 
                 if container.environment:
                     dump_env_vars = {}
+                    original_env_vars = {}
                     for env_var in container.environment:
                         dump_env_vars.update({env_var.name: env_var.value})
+                        original_env_vars.update(
+                            {
+                                hashlib.sha1(  # nosec B324 SHA1 is used here for non-security-critical unique identifiers
+                                    env_var.value.encode("utf-8")
+                                ).hexdigest(): env_var.name
+                            }
+                        )
 
                     env_data = dumps(dump_env_vars, indent=2)
                     detect_secrets_output = detect_secrets_scan(
@@ -35,7 +44,7 @@ class ecs_task_definitions_no_environment_secrets(Check):
                     if detect_secrets_output:
                         secrets_string = ", ".join(
                             [
-                                f"{secret['type']} on line {secret['line_number']}"
+                                f"{secret['type']} on env var {original_env_vars[secret['hashed_secret']]}"
                                 for secret in detect_secrets_output
                             ]
                         )
--- a/prowler/providers/aws/services/ecs/ecs_task_definitions_no_environment_secrets/ecs_task_definitions_no_environment_secrets.py
+++ b/prowler/providers/aws/services/ecs/ecs_task_definitions_no_environment_secrets/ecs_task_definitions_no_environment_secrets.py
@@ -44,7 +44,7 @@ class ecs_task_definitions_no_environment_secrets(Check):
                     if detect_secrets_output:
                         secrets_string = ", ".join(
                             [
-                                f"{secret['type']} on env var {original_env_vars[secret['hashed_secret']]}"
+                                f"{secret['type']} on the environment variable {original_env_vars[secret['hashed_secret']]}"
                                 for secret in detect_secrets_output
                             ]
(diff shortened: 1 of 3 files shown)


kagahd:
I like the fix because it will be easier spot the potential secret. However, I think other checks that use DetectSecrets may have the same problem if they refer to the line number, aren't they?

prowler/providers/aws/services/ecs/ecs_task_definitions_no_environment_secrets/ecs_task_definitions_no_environment_secrets.py
```diff
@@ -25,8 +26,16 @@ def execute(self):
...
+                    original_env_vars = {}
                     for env_var in container.environment:
                         dump_env_vars.update({env_var.name: env_var.value})
+                        original_env_vars.update(
+                            {
+                                hashlib.sha1(  # nosec B324 SHA1 is used here for non-security-critical unique identifiers
+                                    env_var.value.encode("utf-8")
+                                ).hexdigest(): env_var.name
+                            }
+                        )
```
jfagoagas, line 38: This needs to be done for all checks using `detect-secrets`.

prowler/providers/aws/services/ecs/ecs_task_definitions_no_environment_secrets/ecs_task_definitions_no_environment_secrets.py
```diff
@@ -35,7 +44,7 @@ def execute(self):
                     if detect_secrets_output:
                         secrets_string = ", ".join(
                             [
-                                f"{secret['type']} on line {secret['line_number']}"
+                                f"{secret['type']} on env var {original_env_vars[secret['hashed_secret']]}"
```
jfagoagas:
```suggestion
                                f"{secret['type']} on the environment variable {original_env_vars[secret['hashed_secret']]}"
```

jfagoagas (changes requested):

prowler/providers/aws/services/ecs/ecs_task_definitions_no_environment_secrets/ecs_task_definitions_no_environment_secrets.py
pedrooot: Done for the awslambda_function_no_secrets_in_variables check too since it has the same logic.

prowler/providers/aws/services/ecs/ecs_task_definitions_no_environment_secrets/ecs_task_definitions_no_environment_secrets.py
pedrooot: Done

pedrooot:
@kagahd I've been reviewing the logic from the remaining checks using detect-secrets on the same way and the numbers from the lines are correct. 
Also, I've modified the check `awslambda_function_no_secrets_in_variables` in order to apply the same logic when getting the env var name in the status_extended.

jfagoagas (approved):
That's fantastic, what a great improvement 👏

PR #6544, opened 2025-01-15
feat(detect-secrets): get secrets plugins from config.yaml

### LINKED ISSUE #6522: make DetectSecrets detectors configurable
### New feature motivation

Instead to have all the detectors hard coded in [utils.py](https://github.com/prowler-cloud/prowler/blob/master/prowler/lib/utils/utils.py#L112-L138), they could/should be in  [config.yaml](https://github.com/prowler-cloud/prowler/blob/master/prowler/config/config.yaml). This would  make it easier to deactivate detectors aka plugins. 

### Solution Proposed

For example, the `KeywordDetector` finds a lot of false-positives because even if the keyword is suspicious (e.g. `DB_PASSWORD`) the value may be harmless (e.g. the ARN to the secret in AWS secrets manager). Even worse, the user could only mitigate the "potential risk" by **renaming the keyword**, not its value, to make the check pass.
However, I wouldn't wipe the `KeywordDetector` one for all because it can be used for manual audits. Thus, making the used detectors configurable would be a good solution.

None so far

### Additional context

_No response_

pedrooot:
### Context

Fix #6522

### Description

Make plugins from detect-secrets configurable. 
Thanks @kagahd for the feature-request 🚀 

### Checklist

### License

### DIFF
--- a/prowler/providers/aws/services/cloudformation/cloudformation_stack_outputs_find_secrets/cloudformation_stack_outputs_find_secrets.py
+++ b/prowler/providers/aws/services/cloudformation/cloudformation_stack_outputs_find_secrets/cloudformation_stack_outputs_find_secrets.py
@@ -1,5 +1,5 @@
 from prowler.lib.check.models import Check, Check_Report_AWS
-from prowler.lib.utils.utils import detect_secrets_scan
+from prowler.lib.utils.utils import default_detect_secrets_plugins, detect_secrets_scan
 from prowler.providers.aws.services.cloudformation.cloudformation_client import (
     cloudformation_client,
 )
@@ -30,7 +30,7 @@ class cloudformation_stack_outputs_find_secrets(Check):
                     data=data,
                     excluded_secrets=secrets_ignore_patterns,
                     detect_secrets_plugins=cloudformation_client.audit_config.get(
-                        "detect_secrets_plugins", None
+                        "detect_secrets_plugins", default_detect_secrets_plugins
                     ),
                 )
                 # If secrets are found, update the report status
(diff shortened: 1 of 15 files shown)


kagahd:
I opened a https://github.com/Yelp/detect-secrets/pull/910 to add a new detector for Aiven tokens. However, I doubt that it will be merged soon considering all the [open, unmerged other PR's for new detectors](https://github.com/Yelp/detect-secrets/pulls?q=is%3Apr+is%3Aopen+plugin+).
Do you mind integrating the AivenTokenDetector into Prowler? It's just copying detect_secrets/plugins/aiven_token.py to the site-packages folder. That would save me patching the Prowler image.

prowler/providers/aws/services/cloudformation/cloudformation_stack_outputs_find_secrets/cloudformation_stack_outputs_find_secrets.py
```diff
@@ -27,7 +27,11 @@ def execute(self):
                     data += f"{output}\n"
 
                 detect_secrets_output = detect_secrets_scan(
-                    data=data, excluded_secrets=secrets_ignore_patterns
+                    data=data,
+                    excluded_secrets=secrets_ignore_patterns,
+                    detect_secrets_plugins=cloudformation_client.audit_config.get(
```
jfagoagas, line 32: I think you should get the default list of plugins instead of `None` or `{}`. Please review it in all checks.

jfagoagas (changes requested):

prowler/providers/aws/services/cloudformation/cloudformation_stack_outputs_find_secrets/cloudformation_stack_outputs_find_secrets.py
pedrooot: I agree with you, changes on the next commit.

jfagoagas (approved):

PR #6605, opened 2025-01-20
fix(aws): list tags for DocumentDB clusters

kagahd:
### Context

Until now, only the tags of the DocumentDB **instances** and **snapshots** were read out but not the tags of the **cluster**.

### Description

This fix lists also the tags of the cluster.

### Checklist

### License


MrCloudSec (approved):
Thanks @kagahd , good catch!!

PR #6682, opened 2025-01-24
fix(cloudsql): add trusted client certificates case for `cloudsql_instance_ssl_connections`

### LINKED ISSUE #6668: GCP cloudsql_instance_ssl_connections check false positive
### Steps to Reproduce

Run a GCP scan turns up a false positive for a CloudSQL instance with this configuration:

```
"ipConfiguration": {
      "enablePrivatePathForGoogleCloudServices": false,
      "ipv4Enabled": false,
      "privateNetwork": "...",
      "requireSsl": true,
      "sslMode": "TRUSTED_CLIENT_CERTIFICATE_REQUIRED"
    },
```

There are 3 options: https://cloud.google.com/sql/docs/mysql/configure-ssl-instance#enforcing-ssl

`ALLOW_UNENCRYPTED_AND_ENCRYPTED` - Allow unencrypted network traffic (not recommended)
`ENCRYPTED_ONLY` - This option only allows connections using SSL/TLS encryption. Certificates aren't validated.
`TRUSTED_CLIENT_CERTIFICATE_REQUIRED` - This option only allows connections from clients that use a valid client certificate and are SSL encrypted.

The cloudsql_instance_ssl_connections check has this logic (https://github.com/prowler-cloud/prowler/blob/862137ae7d0c87dd3cd270ff72f0f52632592eb3/prowler/providers/gcp/services/cloudsql/cloudsql_instance_ssl_connections/cloudsql_instance_ssl_connections.py#L14)

```
if not instance.require_ssl or instance.ssl_mode != "ENCRYPTED_ONLY":
  report.status = "FAIL"
  report.status_extended = f"Database Instance {instance.name} does not require SSL connections."
```

The logic either needs to exclude both ENCRYPTED_ONLY and TRUSTED_CLIENT_CERTIFICATE_REQUIRED, or be flipped to `or instance.ssl_mode == "ALLOW_UNENCRYPTED_AND_ENCRYPTED"`. Thoughts on which direction is better?

### Expected behavior

Using the TRUSTED_CLIENT_CERTIFICATE_REQUIRED configuration should not throw the check.

N/A

### Environment Resource

Workstation

### OS used

MacOS

### Prowler version

Prowler 5.1.5

### Pip version

pip 24.3.1

### Context

_No response_

puchy22:
### Context

Check `cloudsql_instance_ssl_connections` was not supporting for trusted clients certificates that is a way of SSL connection.

Fix #6668

### Description

- Change check logic to only FAIL with unencrypted connections
- Cover the case with tests.

### Checklist

### License

### DIFF
--- a/prowler/providers/gcp/services/cloudsql/cloudsql_instance_ssl_connections/cloudsql_instance_ssl_connections.py
+++ b/prowler/providers/gcp/services/cloudsql/cloudsql_instance_ssl_connections/cloudsql_instance_ssl_connections.py
@@ -11,7 +11,10 @@ class cloudsql_instance_ssl_connections(Check):
             report.status_extended = (
                 f"Database Instance {instance.name} requires SSL connections."
             )
-            if not instance.require_ssl or instance.ssl_mode != "ENCRYPTED_ONLY":
+            if (
+                not instance.require_ssl
+                or instance.ssl_mode == "ALLOW_UNENCRYPTED_AND_ENCRYPTED"
+            ):
                 report.status = "FAIL"
                 report.status_extended = f"Database Instance {instance.name} does not require SSL connections."
             findings.append(report)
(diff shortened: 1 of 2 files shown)


ejohn20:
@puchy22 Looks great! This PR gives me a good idea of what the corresponding test case would need to look like. Thank you for adjusting that.

MrCloudSec (approved):
Good catch!! 👏🏼👏🏼👏🏼👏🏼

PR #6691, opened 2025-01-27
fix(gcp): fix wrong provider value in check

secretcod3r:
minor PR for fixing a wrong entry for the "provider" key in this check
value should be "gcp"
fixed


jfagoagas (approved):

jfagoagas:
Thanks for the fix @secretcod3r 👏

PR #6700, opened 2025-01-27
feat(schedules): Rework daily schedule to always show the next scan

vicferpoy:
### Description

With this development, daily schedules will always add a new `scheduled` scan, with the correct `scheduled_at` datetime in its details. We are soft deprecating the `next_scan_at` field for it.

In addition, we added a new Django migration to migrate the former scheduled scans to this "format".

image

When adding new providers for the first time, after requesting `POST /schedules/daily`, a new `scheduled` scan with `available` state will be created right away and will be executed once a worker is available, improving the UX.

### License


AdriiiPRodri (dismissed):
LGTM

jfagoagas (dismissed):
I've just tested out the functionality and works as expected, thanks @vicferpoy 👏

PR #6710, opened 2025-01-28
fix(aws): key error for detect-secrets

kagahd:
### Context

PR #6537 introduced a bug which this PR will fix.

### Description

Both prowler checks `ecs_task_definitions_no_environment_secrets` and `awslambda_function_no_secrets_in_variables` will throw a key error if the potential secret is not the **whole** but only a **part** of the value. 
Since PR #6537, the **whole** value was sha1 hashed and associated to its variable name in order to get the variable name back by passing the found potential secret which was also sha1 hashed by detect-secrets. However, detect-secrets hashes only the **part** of the value which was detected to be a potential secret but not the **whole** value as PR #6537 did.
For example, if the value is `srv://admin:pass@db`, detect-secrets will hash only `pass` and not the whole value. Therefore, the hash will differ from the hash computed by PR #6537 and as such it will not be retrievable, leading to a key error and therefore **missing the finding**. 

As you see in this PR, the fix consists in associating the line number with the variable name. Since detect-secrets returns the line number, we are still able to get the variable name for easy spotting the potential secret.
The line number is off by 2 because the array `original_env_vars` starts at 0 while the json, passed to detect-secrets, starts with a curly bracket in line 1.
I added some more unit tests to cover cases with secrets found on different lines in different order with different detectors.

### Checklist

### License

### DIFF
--- a/prowler/providers/aws/services/awslambda/awslambda_function_no_secrets_in_variables/awslambda_function_no_secrets_in_variables.py
+++ b/prowler/providers/aws/services/awslambda/awslambda_function_no_secrets_in_variables/awslambda_function_no_secrets_in_variables.py
@@ -1,4 +1,3 @@
-import hashlib
 import json
 
 from prowler.lib.check.models import Check, Check_Report_AWS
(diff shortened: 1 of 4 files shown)


pedrooot:
Hey! @kagahd thanks for your contribution! I'll review it when I get a chance

pedrooot (approved):
Thanks for this @kagahd 👏🏼 🔝

PR #6711, opened 2025-01-28
fix(scans, findings): Improve API performance ordering by inserted_at instead of id

vicferpoy:
### Context

Postgres v16.x does not include native support for ordering UUIDv7, leading to a really bad performance when trying to order by UUID7.

### Description

Scans and findings queries will no longer order by ID. Instead, `inserted_at` will be used.

### License


jfagoagas (approved):
thanks for this!

PR #6746, opened 2025-01-30
feat(report): add export system and its endpoint

AdriiiPRodri:
### Description

In this PR we have refactored the code in charge of performing the scans from the API side, now all the report files will always be generated, they will be compressed and can be uploaded to S3 or just keep them locally.

We have added an endpoint to get the files from local and from the S3 bucket.

### Checklist

### License


api/src/backend/api/v1/views.py
```diff
@@ -1127,6 +1138,101 @@ def create(self, request, *args, **kwargs):
             },
         )
 
+    @extend_schema(
+        tags=["Scan"],
+        summary="Download ZIP report",
+        description="Returns a ZIP file containing the requested report",
+        request=ScanReportSerializer,
+        responses={
+            200: OpenApiResponse(description="Report obtanined successfully"),
```
jfagoagas, line 1147:
```suggestion
            200: OpenApiResponse(description="Report obtained successfully"),
```

api/src/backend/api/v1/views.py
```diff
@@ -1127,6 +1138,101 @@ def create(self, request, *args, **kwargs):
             },
         )
 
+    @extend_schema(
+        tags=["Scan"],
+        summary="Download ZIP report",
+        description="Returns a ZIP file containing the requested report",
+        request=ScanReportSerializer,
```
vicferpoy, line 1145: Is this required? there shouldn't be any payload in this endpoint.

api/src/backend/api/v1/views.py
```diff
@@ -1127,6 +1138,101 @@ def create(self, request, *args, **kwargs):
...
+        responses={
+            200: OpenApiResponse(description="Report obtanined successfully"),
+            404: OpenApiResponse(description="Report not found"),
+        },
```
vicferpoy, line 1149: This should be autogenerated since this is detailed action view. Please avoid this kind of documentation details unless you add them for every viewset in the module.

.env
jfagoagas, line 1: All the `ARTIFACTS_*` environment variables should be prefixed with `DJANGO` because it is the convention we've been using for the ones used in Django.

api/src/backend/api/v1/views.py
```diff
@@ -1127,6 +1138,101 @@ def create(self, request, *args, **kwargs):
...
+                )
+
+            try:
+                file_path = zip_files[0]
+                with open(file_path, "rb") as f:
+                    file_content = f.read()
+                filename = os.path.basename(file_path)
+            except IOError:
+                return Response(
+                    {"detail": "Error reading local file"},
```
vicferpoy, line 1226: I don't think this belongs in here. It also makes unit testing way more difficult. Please create a service layer to encapsulate all the logic related to the s3 integration.

api/src/backend/tasks/jobs/scan.py
vicferpoy, line 1: Some of these changes seem unnecessary, and a few appear to introduce incorrect behavior.

Could you go through the diff again and clean up any unintended modifications? It’d be good to double-check that all changes align with our existing patterns and requirements. 

Also, same as with the view logic, export specific functions that are used here should be moved to the export service module and imported here.

api/src/backend/tasks/jobs/scan.py
```diff
@@ -144,116 +310,66 @@ def perform_prowler_scan(
...
-        for progress, findings in prowler_scan.scan():
+        all_findings = []
+        for progress, findings, stats in prowler_scan.scan():
             for finding in findings:
                 if finding is None:
                     logger.error(f"None finding detected on scan {scan_id}.")
                     continue
                 for attempt in range(CELERY_DEADLOCK_ATTEMPTS):
                     try:
-                        with rls_transaction(tenant_id):
```
jfagoagas, line 160: Why are you modifying the logic to store findings and resources? If this is out of the scope of this functionality I'd remove it from this PR.

vicferpoy (changes requested):
Thanks for the job Adri! I added some questions and suggestions. Also, I'm missing a lot of tests.

api/src/backend/tasks/jobs/scan.py
```diff
@@ -266,92 +382,96 @@ def perform_prowler_scan(
...
+                    if region_data.get(finding.check_id) != "FAIL":
+                        region_data[finding.check_id] = finding.status.value
 
-            # Update scan progress
+            # Progress updates and output generation
             with rls_transaction(tenant_id):
                 scan_instance.progress = progress
                 scan_instance.save()
 
+            all_findings.extend(findings)
```
jfagoagas, line 400: I think we should not do that, for large scans this is going to increase memory usage a lot and could end up in OOM errors.

api/src/backend/tasks/jobs/scan.py
```diff
@@ -266,92 +382,96 @@ def perform_prowler_scan(
...
+            kwargs = dict(config["kwargs"])
+            if mode == "html":
+                kwargs["provider"] = prowler_provider
+                kwargs["stats"] = stats
+            config["class"](
+                findings=all_findings,
+                create_file_descriptor=True,
+                file_path=output_directory,
+                file_extension=config["suffix"],
+            ).batch_write_data_to_file(**kwargs)
```
jfagoagas, line 413: I thought we talked about writing findings to file in _streaming_ instead of all at once to reduce memory overhead.

api/src/backend/tasks/jobs/scan.py
```diff
@@ -144,116 +310,66 @@ def perform_prowler_scan(
...
                 with rls_transaction(tenant_id):
                     finding_uid = finding.uid
-                    last_first_seen_at = None
                     if finding_uid not in last_status_cache:
-                        most_recent_finding = (
-                            Finding.all_objects.filter(
-                                tenant_id=tenant_id, uid=finding_uid
-                            )
+                        most_recent = (
+                            Finding.objects.filter(uid=finding_uid)
```
jfagoagas, line 355: That's concerning... why are you removing the `tenant_id` filter? It is a key part of our multi tenant system with RLS, we could not remove it unless there is a strong reason behind.

api/src/backend/tasks/jobs/scan.py
```diff
@@ -478,29 +598,28 @@ def aggregate_findings(tenant_id: str, scan_id: str):
...
-                fail_changed=agg["fail_changed"],
-                pass_new=agg["pass_new"],
-                pass_changed=agg["pass_changed"],
-                muted_new=agg["muted_new"],
-                muted_changed=agg["muted_changed"],
-            )
-            for agg in aggregation
-        }
-        ScanSummary.objects.bulk_create(scan_aggregations, batch_size=3000)
+        ScanSummary.objects.bulk_create(
```
jfagoagas, line 601: Why the `rls_transaction` is removed here? Is this actually working? An insert into this table won't work without the `tenant_id` configured.

And I don't know if modifying this part is within the PR scope.

prowler/config/config.py
```diff
@@ -59,6 +59,7 @@ def get_available_compliance_frameworks(provider=None):
 # gcp_zones_json_file = "gcp_zones.json"
 
 default_output_directory = getcwd() + "/output"
+tmp_output_directory = "/tmp/prowler_api_output"
```
jfagoagas, line 62: Is this part of the SDK config?

prowler/lib/check/check.py
jfagoagas, line 1: This is a little change but for next time please keep the bound code changes to PR scope. Is this covered with tests?

prowler/lib/outputs/output.py
```diff
@@ -38,16 +38,18 @@ def __init__(
         file_extension: str = "",
     ) -> None:
         self._data = []
+        self.file_path = file_path
```
jfagoagas, line 41: What is the purpose of this `file_path`?

prowler/lib/scan/scan.py
```diff
@@ -74,151 +72,190 @@ def __init__(
             ScanInvalidStatusError: If the status does not exist in the provider.
         """
         self._provider = provider
+        self._statuses = self._validate_statuses(status)
```
jfagoagas, line 75: I really like all this changes to encapsulate this logic but please move it to a different PR, we need also to cover the changes with tests now that we have several new methods for input validation.

This is one of the critical paths, we need to make sure that all changes are tested.

jfagoagas (changes requested):
I'm worried about several parts of this implementation:

1. What happens if the output generation fails during a Scan? How are we going to recover that fail and generate the outputs for the Scan?
2. We are introducing a lot of changes not covered by tests, we could not skip testing part.
3. I'd like to talk about API responses when something fails, e.g.: `ClientError` from boto3 should not be a 500 in our API since we are not responsible of that error.

I did not finish the review because I think we need to talk about the design of this first.

jfagoagas:
Closing this PR since the work was done in https://github.com/prowler-cloud/prowler/pull/6878

PR #6748, opened 2025-01-30
feat(findings): Improve /findings/metadata performance

vicferpoy:
### Context

To improve UX, we found that `GET /findings/metadata` was delaying the findings page rendering. 

### Description

The query strategy to retrieve the dynamic filters has been changed. In addition, tags dynamic filters have been temporarily disabled until we implement them in the UI and find a proper solution for them.

This PR also includes bug-proof logic for null findings or empty regions during scans, due to a recent bug in our SDK.

### License

### DIFF
--- a/api/poetry.lock
+++ b/api/poetry.lock
@@ -253,6 +253,20 @@ files = [
 cryptography = "*"
 
 [[package]]
+name = "autopep8"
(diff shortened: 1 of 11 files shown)


jfagoagas (dismissed):
Awesome 👏

PR #6753, opened 2025-01-30
feat(scans): Optimize read queries during scans

vicferpoy:
### Context

During scans, when retrieving information from the latest common finding, we did:

```sql
SELECT "findings"."status", "findings"."first_seen_at" FROM "findings" INNER JOIN "scans" ON ("findings"."scan_id" = "scans"."id") INNER JOIN "providers" ON ("scans"."provider_id" = "providers"."id") WHERE (NOT "providers"."is_deleted" AND NOT "providers"."is_deleted" AND "findings"."uid" = ?) ORDER BY "findings"."id" DESC LIMIT ?
```

### Description

We got rid of the JOINs for that specific query and added a filter for `tenant_id`, increasing performance for that query.

### Checklist

### License

### DIFF
--- a/api/src/backend/tasks/jobs/scan.py
+++ b/api/src/backend/tasks/jobs/scan.py
@@ -228,8 +228,10 @@ def perform_prowler_scan(
                     last_first_seen_at = None
                     if finding_uid not in last_status_cache:
                         most_recent_finding = (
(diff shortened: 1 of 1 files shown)


jfagoagas (approved):
🚀

PR #6781, opened 2025-02-03
fix(directoryservice): handle ClientException

danibarranqueroo:
### Context

The sentry scanner notifies us about two issues that comes from an exception that should be handle as a warning instead of an error. This happened during a calling to DescribeEventTopics and GetSnapshotLimits.

### Description

Modified those exceptions to catch the error and show it as a warning.

### Checklist

### License

### DIFF
--- a/prowler/providers/aws/services/directoryservice/directoryservice_service.py
+++ b/prowler/providers/aws/services/directoryservice/directoryservice_service.py
@@ -131,9 +131,15 @@ class DirectoryService(AWSService):
                             self.directories[directory.id].event_topics = event_topics
                         except ClientError as error:
                             if error.response["Error"]["Code"] == "ClientException":
-                                logger.warning(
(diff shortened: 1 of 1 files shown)


prowler/providers/aws/services/directoryservice/directoryservice_service.py
```diff
@@ -107,21 +107,44 @@ def _describe_event_topics(self, regional_client):
...
+                                        status=event_topic["Status"],
+                                        created_date_time=event_topic[
+                                            "CreatedDateTime"
+                                        ],
+                                    )
+                                )
+                            self.directories[directory.id].event_topics = event_topics
+                        except ClientError as error:
+                            if error.response["Error"]["Code"] == "ClientException":
+                                error_message = error.response["Error"]["Message"]
```
MrCloudSec: Please, only verify the attribute `["Error"]["Message"]`

MrCloudSec (approved):

PR #6863, opened 2025-02-07
feat(findings): Use ArrayAgg and subqueries on metadata endpoint

vicferpoy:
### Context

After the latest changes on `GET /findings/metadata`, performance on production environment got worse.

### Description

This PR reworks that endpoint logic again to perform array aggregations and reduce the query weight by using subqueries and removing duplicates/sorting at Python level.

### Checklist

### License


api/src/backend/api/v1/views.py
```diff
@@ -1418,55 +1418,31 @@ def metadata(self, request):
...
-            # "tags": tags_dict
+            "services": services,
+            "regions": regions,
+            "resource_types": resource_types,
         }
 
         serializer = self.get_serializer(data=result)
         serializer.is_valid(raise_exception=True)
-        return Response(serializer.data, status=status.HTTP_200_OK)
+        return Response(serializer.data, status=200)
```
```suggestion
        return Response(serializer.data, status=status.HTTP_200_OK)
```

jfagoagas (approved):

PR #6874, opened 2025-02-10
 feat(api): setup sentry for OSS API

cesararroba:
### Context

Setup Sentry for OSS API to be able to manage errors and tracing

### Description

- Added Sentry default configuration
- Added `DJANGO_SENTRY_DSN` variable to add your own Sentry key

### Checklist

### License


api/src/backend/config/django/base.py
```diff
@@ -207,4 +209,18 @@
...
+    # Add data like request headers and IP for users,
+    # see https://docs.sentry.io/platforms/python/data-management/data-collected/ for more info
+    send_default_pii=True,
+    _experiments={
+        # Set continuous_profiling_auto_start to True
+        # to automatically start the profiler on when
+        # possible.
+        "continuous_profiling_auto_start": True,
+    },
+)
```
vicferpoy: Is Sentry switched on by default? Does it affect to logs or performance? Maybe we should add an `if` clause depending on the environment variable.

.env
```diff
@@ -92,3 +92,4 @@ jQIDAQAB
 # openssl rand -base64 32
 DJANGO_SECRETS_ENCRYPTION_KEY="oE/ltOhp/n1TdbHjVmzcjDPLcLA41CVI/4Rk+UB5ESc="
 DJANGO_BROKER_VISIBILITY_TIMEOUT=86400
+DJANGO_SENTRY_DSN=
```
vicferpoy, line 95: Does Sentry work if this variable is not set?

vicferpoy (changes requested):
Thank you! I left some questions.

.env
cesararroba: It will be disabled as the default value is `""`, but I added it into the `.env` file in case someone wants to use it.

api/src/backend/config/django/base.py
cesararroba: I've seen if the dsn attribute is empty, Sentry is disabled. That's why I've used `""` by default. This won't affect to the logs or performance. I did a local test and there is nothing about Sentry when the dsn is empty.

vicferpoy (dismissed):
Thanks for clarifying! LGTM

.env
jfagoagas, line 95: We should add these two to have more info in all the errors:
- `SENTRY_ENVIRONMENT`
- `SENTRY_RELEASE`

api/src/backend/config/django/base.py
```diff
@@ -207,4 +209,18 @@
 
 
 TESTING = False
+
 FINDINGS_MAX_DAYS_IN_RANGE = env.int("DJANGO_FINDINGS_MAX_DAYS_IN_RANGE", 7)
+
+sentry_sdk.init(
```
jfagoagas: I'm thinking about start collecting also `log.error` to catch SDK's errors too.

.env
cesararroba: Added missing variables with `local` as default value

api/src/backend/config/django/base.py
cesararroba: If I'm not reading wrong, this is configured by default in the logging integration (active by default too): https://docs.sentry.io/platforms/python/integrations/logging/#options

jfagoagas: We should also add `before_send=before_send` not to send several credential errors coming from the SDK.


```python
ignored_exceptions = [
    # Authentication Errors from AWS
    "InvalidToken",
    "AccessDeniedException",
    "AuthorizationErrorException",
    "UnrecognizedClientException",
    "UnauthorizedOperation",
    "AuthFailure",
    "InvalidClientTokenId",
    "AccessDenied",
    # Shodan Check
    "No Shodan API Key",
    # For now we don't want to log the RequestLimitExceeded errors
    "RequestLimitExceeded",
    "ThrottlingException",
    "Rate exceeded",
    # The following comes from urllib3
    # eu-west-1 -- HTTPClientError[126]: An HTTP Client raised an unhandled exception: AWSHTTPSConnectionPool(host='hostname.s3.eu-west-1.amazonaws.com', port=443): Pool is closed.
    "Pool is closed",
]


def before_send(event, hint):
    """
    before_send handles the Sentry events in order to sent them or not
    """
    # Ignore logs with the ignored_exceptions
    # https://docs.python.org/3/library/logging.html#logrecord-objects
    if "log_record" in hint:
        log_msg = hint["log_record"].msg
        log_lvl = hint["log_record"].levelno

        # Handle Error events and discard the rest
        if log_lvl == 40 and any(ignored in log_msg for ignored in ignored_exceptions):
            return
    return event
```

This is not tested 😄

cesararroba: Added and tested 👍

cesararroba:
@vicferpoy could you please check the `poetry.lock` file conflicts please? 🙏

vicferpoy (approved):
Great, thank you!

jfagoagas (approved):

PR #6878, opened 2025-02-10
feat(export): add API export system

AdriiiPRodri:
### Description

In this PR we have refactored the code in charge of performing the scans from the API side, now all the report files will always be generated, they will be compressed and can be uploaded to S3 or just keep them locally. This happens in a separate task.

We have added an endpoint to get the files from local and from the S3 bucket.

### Checklist

#### API

### License


prowler/lib/outputs/finding.py
```diff
@@ -267,3 +274,183 @@ def generate_output(
...
+        Finding object is structured for use in API responses or further processing.
+
+        Args:
+            finding (API Finding): An API Finding instance containing data from the database.
+
+        Returns:
+            Finding: A new Finding instance populated with data from the provided model.
+        """
+        output_data = {}
+        output_data["auth_method"] = ""
```
jfagoagas: @jfagoagas review

prowler/lib/outputs/finding.py
```diff
@@ -267,3 +274,183 @@ def generate_output(
...
+        output_data["timestamp"] = finding.inserted_at
+        output_data["account_uid"] = finding.scan.provider.uid
+        output_data["account_name"] = ""
```
jfagoagas: @jfagoagas review

prowler/lib/outputs/finding.py
```diff
@@ -267,3 +274,183 @@ def generate_output(
...
+            ResourceIdTemplate=finding.check_metadata["resourceidtemplate"],
+            Categories=finding.check_metadata["categories"],
+            DependsOn=finding.check_metadata["dependson"],
+            RelatedTo=finding.check_metadata["relatedto"],
+            Notes=finding.check_metadata["notes"],
+        )
+        output_data["uid"] = finding.uid
+        output_data["status"] = Status(finding.status)
+        output_data["status_extended"] = finding.status_extended
+        output_data["resource_uid"] = finding.resources.first().uid
```
jfagoagas: Could you call just once to `finding.resources.first()` to get the resource and the get the required attributes.

prowler/lib/outputs/finding.py
```diff
@@ -267,3 +274,183 @@ def generate_output(
                 f"{error.__class__.__name__}[{error.__traceback__.tb_lineno}]: {error}"
             )
             raise error
+
+    @classmethod
+    def transform_api_finding(cls, finding) -> "Finding":
```
jfagoagas: Please add tests to cover this function.

prowler/lib/outputs/finding.py
```diff
@@ -267,3 +274,183 @@ def generate_output(
...
+        resource_tags = finding.resources.first().tags.all()
+        output_data["resource_tags"] = unroll_tags(
+            [{"key": tag.key, "value": tag.value} for tag in resource_tags]
+        )
+        output_data["region"] = finding.resources.first().region
+        output_data["compliance"] = {}
+
+        return cls(**output_data)
+
+    def _transform_findings_stats(scan_summaries: list[dict]) -> dict:
```
jfagoagas, line 363: We need to add tests to this function too.

docker-compose-dev.yml
jfagoagas, line 1: Add this to the `docker-compose.yml`.

api/src/backend/tasks/tasks.py
```diff
@@ -152,3 +168,116 @@ def perform_scan_summary_task(tenant_id: str, scan_id: str):
...
+    output_writers = {}
+    scan_summary = FindingOutput._transform_findings_stats(
+        ScanSummary.objects.filter(scan_id=scan_id)
+    )
+
+    # Retrieve findings queryset
+    findings_qs = Finding.objects.filter(scan_id=scan_id).order_by("uid")
+
+    # Process findings in batches
+    for batch, is_last_batch in batched(findings_qs.iterator(), 50):
```
jfagoagas: Can we add an environment variable for the batch size?

api/src/backend/tasks/tasks.py
```diff
@@ -152,3 +168,116 @@ def perform_scan_summary_task(tenant_id: str, scan_id: str):
...
+    os.makedirs("/".join(output_directory.split("/")[:-1]), exist_ok=True)
+
+    # Define auxiliary variables
+    output_writers = {}
+    scan_summary = FindingOutput._transform_findings_stats(
+        ScanSummary.objects.filter(scan_id=scan_id)
+    )
+
+    # Retrieve findings queryset
+    findings_qs = Finding.objects.filter(scan_id=scan_id).order_by("uid")
```
jfagoagas:
```suggestion
    findings_qs = Finding.objects.filter(scan_id=scan_id).order_by("service")
```

I think this will be more performant since `uid` is a huge string.

api/src/backend/tasks/tasks.py
```diff
@@ -152,3 +168,116 @@ def perform_scan_summary_task(tenant_id: str, scan_id: str):
...
+                    findings=finding_outputs,
+                    file_path=output_directory,
+                    file_extension=config["suffix"],
+                )
+                writer.close_file = is_last_batch
+                output_writers[writer_class] = writer
+
+            # Write the current batch using the writer
+            writer.batch_write_data_to_file(**kwargs)
+            writer._data = []
```
jfagoagas, line 254: Could you add a TODO to refactor this line and modify it in the SDK?

api/src/backend/tasks/tasks.py
jfagoagas, line 1: We need to review output file name to be:
`prowler-output-<provider.uid>-<timestamp>.<extension>`

api/src/backend/tasks/tasks.py
```diff
@@ -152,3 +168,116 @@ def perform_scan_summary_task(tenant_id: str, scan_id: str):
...
+
+    # Compress output files
+    output_directory = _compress_output_files(output_directory)
+
+    # Save to configured storage
+    uploaded = _upload_to_s3(tenant_id, output_directory, scan_id)
+
+    if uploaded:
+        output_directory = uploaded
+        uploaded = True
```
jfagoagas, line 264: We need to add a cleanup if using S3 to remove files from `/tmp`

api/src/backend/tasks/tasks.py
```diff
@@ -152,3 +168,116 @@ def perform_scan_summary_task(tenant_id: str, scan_id: str):
...
+    else:
+        uploaded = False
+
+    # Update the scan instance with the output path and upload status
+    Scan.objects.filter(id=scan_id).update(
+        output_path=output_directory, upload_to_s3=uploaded
+    )
+
+    return {
+        "output_path": output_directory,
```
jfagoagas: We need to hide this if using it in Cloud with S3.

api/src/backend/api/v1/views.py
```diff
@@ -1090,6 +1099,88 @@ def partial_update(self, request, *args, **kwargs):
...
         return Response(data=read_serializer.data, status=status.HTTP_200_OK)
 
+    @extend_schema(
+        tags=["Scan"],
+        summary="Download ZIP report",
+        description="Returns a ZIP file containing the requested report",
+        request=ScanReportSerializer,
+        responses={
+            200: OpenApiResponse(description="Report obtained successfully"),
+            423: OpenApiResponse(
```
```suggestion
            403: OpenApiResponse(
```

api/src/backend/api/migrations/0010_scan_report_output.py
```diff
@@ -0,0 +1,22 @@
...
+        migrations.AddField(
+            model_name="scan",
+            name="output_path",
+            field=models.CharField(blank=True, max_length=200, null=True),
+        ),
+        migrations.AddField(
+            model_name="scan",
+            name="upload_to_s3",
+            field=models.BooleanField(blank=True, null=True),
+        ),
```
jfagoagas: We decided to remove this field.

.env
jfagoagas, line 1: I think we need to add this to the `base.py`

jfagoagas (changes requested):

api/src/backend/tasks/tasks.py
AdriiiPRodri: The UID field is indexed and the service field is in the resource table, so this query will be more slow

prowler/lib/outputs/finding.py
jfagoagas: @AdriiiPRodri these are the contents for the `authe_method` field regarding each provider allowed in the API?
- GCP ->`Principal: {gcp_provider.identity.profile}`, we need to get `gcp_provider.identity.profile`
- Azure -> `{azure_provider.identity.identity_type}: {azure_provider.identity.identity_id}`, we need to get `azure_provider.identity.identity_type` and  `azure_provider.identity.identity_id`
- K8s -> `kubeconfig` as a string, since we don't support `in-cluster` which is the other `auth_method` supported.
- AWS -> `profile: None` as a string since `profile` is not supported, in the future we'll need to adjust this but that's the format in the SDK

I think you should use `get_nested_attribute` function to get all of the above as it is done in the `Finding` class, e.g.: `get_nested_attribute(gcp_provider, "identity.profile")` since it defaults to `""` if not present.

jfagoagas: `partition` is also missing and the content by provider is:

- GCP -> Nothing, so empty string
- Azure -> `{azure_provider.region_config.name}`
- K8s -> Nothing, so empty string
- AWS -> `{aws_provider.identity.partition}`

I think you should use `get_nested_attribute` function to get all of the above as it is done in the `Finding` class, e.g.: `get_nested_attribute(gcp_provider, "identity.profile")` since it defaults to `""` if not present.

prowler/lib/outputs/finding.py
jfagoagas: - GCP -> `api_provider.uid`
- Azure -> `api_provider.uid`
- K8s -> `context: {api_provider.uid}`
- AWS -> `{aws_provider.organizations_metadata.account_name}` 

I think you should use `get_nested_attribute` function to get all of the above as it is done in the `Finding` class, e.g.: `get_nested_attribute(gcp_provider, "identity.profile")` since it defaults to `""` if not present.

jfagoagas: I'm thinking that if we can pass the `provider` object to this function we could get also:
- `account_name`
- `account_email`
- `account_organization_uid`
- `account_organization_name`
- `account_tags`

for the different providers as it is done in the `generate_output` class method.

api/src/backend/config/django/base.py
```diff
@@ -208,3 +208,14 @@
 
 TESTING = False
 FINDINGS_MAX_DAYS_IN_RANGE = env.int("DJANGO_FINDINGS_MAX_DAYS_IN_RANGE", 7)
+
+
+# API export settings
+TMP_OUTPUT_DIRECTORY = "/tmp/prowler_api_output"
+FINDINGS_BATCH_SIZE = 1000
```
jfagoagas: What about `env.int("DJANGO_FINDINGS_BATCH_SIZE", 1000)`

api/src/backend/api/v1/views.py
```diff
@@ -1181,6 +1190,72 @@ def partial_update(self, request, *args, **kwargs):
...
+        request=ScanReportSerializer,
+        responses={
+            200: OpenApiResponse(description="Report obtained successfully"),
+            403: OpenApiResponse(
+                description="There is a problem with the AWS credentials"
+            ),
+        },
+    )
+    @action(detail=True, methods=["get"], url_name="report")
+    def report(self, request, pk=None):
```
vicferpoy, line 1208: Missing report task status check to return a customized response if it's in progress or it failed.
Same for the scan status.

api/src/backend/tasks/tasks.py
```diff
@@ -152,3 +170,125 @@ def perform_scan_summary_task(tenant_id: str, scan_id: str):
...
+    )
+
+    # Define auxiliary variables
+    output_writers = {}
+    scan_summary = FindingOutput._transform_findings_stats(
+        ScanSummary.objects.filter(scan_id=scan_id)
+    )
+
+    # Retrieve findings queryset
+    findings_qs = Finding.objects.filter(scan_id=scan_id).order_by("uid")
```
vicferpoy: You can use `all_objects` here to avoid unnecessary JOINs.

api/src/backend/tasks/tasks.py
```diff
@@ -152,3 +170,125 @@ def perform_scan_summary_task(tenant_id: str, scan_id: str):
...
+    uploaded = _upload_to_s3(tenant_id, output_directory, scan_id)
+
+    if uploaded:
+        output_directory = uploaded
+        uploaded = True
+    else:
+        uploaded = False
+
+    # Update the scan instance with the output path
+    Scan.objects.filter(id=scan_id).update(output_path=output_directory)
```
vicferpoy: Same here.

api/src/backend/tasks/tasks.py
```diff
@@ -68,13 +80,20 @@ def perform_scan_task(
...
         scan_id=scan_id,
         provider_id=provider_id,
         checks_to_execute=checks_to_execute,
     )
 
+    chain(
+        perform_scan_summary_task.s(tenant_id, scan_id),
+        generate_outputs.si(scan_id, provider_id, tenant_id=tenant_id),
+    ).apply_async()
+
```
vicferpoy: Why `perform_scan_summary_task` uses a mutable signature now (`.s`)?

api/src/backend/tasks/tasks.py
```diff
@@ -152,3 +170,125 @@ def perform_scan_summary_task(tenant_id: str, scan_id: str):
...
+            - is_last_batch (bool): True if this is the final batch, False otherwise.
+    """
+    batch = []
+    for item in iterable:
+        batch.append(item)
+        if len(batch) == batch_size:
+            yield batch, False
+            batch = []
+
+    yield batch, True
```
vicferpoy: Why is a tool function defined in the task module?

api/src/backend/tasks/tasks.py
```diff
@@ -152,3 +170,125 @@ def perform_scan_summary_task(tenant_id: str, scan_id: str):
...
+
+
+@shared_task(base=RLSTask, name="scan-output", queue="scans")
```
vicferpoy: We talk about `reports` everywhere. Since it is a `RLSTask`, it means this task can be checked by the user using the `/tasks` endpoints. Shouldn't we call it `scan-report`?

Also, the queue concerns me. Regarding Prowler Cloud, we do not have active workers for the `scans` queue. This means that we will have to adapt more logic to deploy another ECS for this kind of tasks.

vicferpoy (changes requested):
I have some extra questions:

- If the report task is heavy, should we add a new queue for it? `scans` queue has some issues, as stated in a comment.
- Should we add a retry policy for this task?

api/docker-entrypoint.sh
```diff
@@ -28,7 +28,7 @@ start_prod_server() {
 
 start_worker() {
   echo "Starting the worker..."
-  poetry run python -m celery -A config.celery worker -l "${DJANGO_LOGGING_LEVEL:-info}" -Q celery,scans -E --max-tasks-per-child 1
+  poetry run python -m celery -A config.celery worker -l "${DJANGO_LOGGING_LEVEL:-info}" -Q celery,scans,scans-report -E --max-tasks-per-child 1
```
jfagoagas: I'd rename it to `scan-reports` since it's more consistent for a queue handling multiple "scan reports," which makes sense if multiple reports are processed.

api/src/backend/api/migrations/0010_scan_report_output.py
```diff
@@ -0,0 +1,17 @@
...
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("api", "0009_increase_provider_uid_maximum_length"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="scan",
+            name="output_path",
```
jfagoagas:
```suggestion
            name="output_location",
```

What about `output_location`? I think it'll be more generic for future approaches, like S3 now but other file locations in the future. I'd also like to define if we are going to call this `reports` or `outputs`.

api/src/backend/api/specs/v1.yaml
```diff
@@ -4105,6 +4105,39 @@ paths:
...
+        description: A UUID string identifying this scan.
+        required: true
+      tags:
+      - Scan
+      security:
+      - jwtAuth: []
+      responses:
+        '200':
+          description: Report obtained successfully
+        '423':
```
jfagoagas: `423` or `403`? I think we should handle 404's too.

jfagoagas: Maybe it only needs to be regenerated.

api/src/backend/config/django/base.py
```diff
@@ -219,3 +219,20 @@
...
 FINDINGS_MAX_DAYS_IN_RANGE = env.int("DJANGO_FINDINGS_MAX_DAYS_IN_RANGE", 7)
+
+
+# API export settings
+DJANGO_TMP_OUTPUT_DIRECTORY = env.str(
+    "DJANGO_TMP_OUTPUT_DIRECTORY", "/tmp/prowler_api_output"
+)
+DJANGO_FINDINGS_BATCH_SIZE = env.str("DJANGO_FINDINGS_BATCH_SIZE", 1000)
+
+DJANGO_ARTIFACTS_AWS_S3_OUTPUT_BUCKET = env.str(
```
jfagoagas: I have the same issue with `ARTIFACTS` maybe reports is more consistent across the codebase. This is not a blocker whatsoever.

api/src/backend/tasks/jobs/export.py
jfagoagas, line 1: I'd add tests here for the AWS integration, I can help you with that.

api/src/backend/tasks/tasks.py
```diff
@@ -135,12 +154,11 @@ def perform_scheduled_scan_task(self, tenant_id: str, provider_id: str):
...
             )
 
-    perform_scan_summary_task.apply_async(
-        kwargs={
-            "tenant_id": tenant_id,
-            "scan_id": str(scan_instance.id),
-        }
-    )
+    chain(
+        perform_scan_summary_task.s(tenant_id, scan_instance.id),
```
jfagoagas:
```suggestion
        perform_scan_summary_task.si(tenant_id, scan_instance.id),
```
maybe?

api/src/backend/tasks/tasks.py
jfagoagas: Was this included?

prowler/lib/outputs/finding.py
jfagoagas: Was this included?

prowler/lib/outputs/output.py
```diff
@@ -33,21 +32,27 @@ class Output(ABC):
...
         file_path: str = None,
         file_extension: str = "",
+        from_cli: bool = True,
     ) -> None:
         self._data = []
+        self.close_file = False
+        self.file_path = file_path
+        self._file_descriptor = None
+        # This parameter is to avoid refactoring more code, the CLI does not write in batches, the API does
+        self._from_cli = from_cli
```
jfagoagas, line 44: Is there any other way? I'm worried about having something like this, maybe something like:
- `batch_mode`
- `write_in_batches`
- `use_batches`

api/CHANGELOG.md
```diff
@@ -11,6 +11,13 @@ All notable changes to the **Prowler API** are documented in this file.
 
 ---
 
+## [v1.5.0] (Prowler v5.4.0) - 2025-XX-XX
```
jfagoagas: There is a section for v1.5.0, please review it.

jfagoagas (changes requested):

api/CHANGELOG.md
AdriiiPRodri: When this was created it was not

prowler/lib/outputs/output.py
AdriiiPRodri: It is called this way to emphasize that this is because it comes from the CLI and that really what should be done is to use this also in the CLI but that is another issue, you can change the name but still what should be done is to refactor the CLI

api/src/backend/config/django/base.py
AdriiiPRodri: It is called artifacts because it is a file containing the reports but it is not a report itself, it can be discussed. It is the same as Gitlab artifacts for example

api/src/backend/tasks/tasks.py
AdriiiPRodri: Didn't we agree that the AWS machines are ephemeral so we don't mind deleting anything and locally it is inside the temporary files folder?

jfagoagas: Sure, but then I realised that the exports are going to run in a different queue, so it won't be ephemeral.

api/CHANGELOG.md
jfagoagas: Ah okay

prowler/lib/outputs/output.py
jfagoagas: I don't know what will be in the SDK but I think for now I'd keep a name for batches instead of something like `cli`.

api/src/backend/api/v1/views.py
```diff
@@ -1181,6 +1190,70 @@ def partial_update(self, request, *args, **kwargs):
...
+
+        if scan_instance.state != StateChoices.COMPLETED:
+            return Response(
+                {
+                    "detail": f"The scan is not finished yet. State is: {scan_instance.state}"
+                },
+                status=status.HTTP_400_BAD_REQUEST,
+            )
+
+        if scan_instance.output_path.startswith("s3://"):
```
jfagoagas: Here we need to check if `output_path` is not `None` because this is going to be the default value for all the scans before this feature. We should return a `404 Not Found`.

jfagoagas (changes requested):
I've made several tests end to end and I saw the following:
- The folder within the `.zip` should be called `output` instead of artifacts
- Following the conversation with @vicferpoy we need to handle more statuses while getting the outputs:
```
# if scan in progress -> 202 Accepted with scan resource task
# if scan complete -> check scan_report task
# # if scan_report in progress -> 202 Accepted with scan_report resource task
# # if scan_report completed
# # # file found -> 200 with report (current response)
# # # file not found -> 404 NOT FOUND
```

- To fill the outputs we need to pass the following values regarding the provider:
```
# AWS
sdk_finding.resource_name = api_finding.resource.name
sdk_finding.resource_arn = api_finding.resource.uid
# GCP
sdk_finding.resource_name = api_finding.resource.name
sdk_finding.resource_id = api_finding.resource.uid
# Azure
sdk_finding.resource_name = api_finding.resource.name
sdk_finding.resource_id = api_finding.resource.uid
# Kubernetes
sdk_finding.resource_name = api_finding.resource.name
sdk_finding.resource_id = api_finding.resource.uid
```

- We need to delete the files from `/tmp` if the file was uploaded to S3.

api/CHANGELOG.md
```diff
@@ -8,6 +8,14 @@ All notable changes to the **Prowler API** are documented in this file.
...
 ### Added
 - Social login integration with Google and GitHub [(#6906)](https://github.com/prowler-cloud/prowler/pull/6906)
+- Add API scan report system, now all scans launched from the API will generate a compressed file with the report in OCSF, CSV and HTML formats [(#6878)](https://github.com/prowler-cloud/prowler/pull/6878).
+
+---
+
+## [v1.5.0] (Prowler v5.4.0) - 2025-XX-XX
+
+### Added
+- Add API scan report system, now all scans launched from the API will generate a compressed file with the report in OCSF, CSV and HTML formats [(#6878)](https://github.com/prowler-cloud/prowler/pull/6878).
```
```suggestion
```

api/src/backend/api/v1/views.py
```diff
@@ -1181,6 +1190,92 @@ def partial_update(self, request, *args, **kwargs):
...
+    @extend_schema(
+        tags=["Scan"],
+        summary="Download ZIP report",
+        description="Returns a ZIP file containing the requested report",
+        request=ScanReportSerializer,
+        responses={
+            200: OpenApiResponse(description="Report obtained successfully"),
+            202: OpenApiResponse(description="The task is in progress"),
+            403: OpenApiResponse(
+                description="There is a problem with the AWS credentials"
```
```suggestion
                description="There is a problem with credentials"
```

.env
```diff
@@ -30,6 +30,30 @@ VALKEY_HOST=valkey
...
+
+# An optional AWS session token
+DJANGO_OUTPUT_AWS_SESSION_TOKEN=""
+
+# The AWS region where your S3 bucket is located (e.g., "us-east-1")
+DJANGO_OUTPUT_AWS_DEFAULT_REGION=""
+
+# The name of the S3 bucket where scan output should be stored
+DJANGO_OUTPUT_AWS_S3_OUTPUT_BUCKET=""
+
```
jfagoagas: As you add `S3` to `DJANGO_OUTPUT_AWS_S3_OUTPUT_BUCKET` I think it'd be a great idea to also put it in the rest:
```suggestion
# The path to the directory where scan output should be stored
DJANGO_TMP_OUTPUT_DIRECTORY = "/tmp/prowler_api_output"

# The maximum number of findings to process in a single batch
DJANGO_FINDINGS_BATCH_SIZE = 1000

# The AWS access key to be used when uploading scan output to an S3 bucket
# If left empty, default AWS credentials resolution behavior will be used
DJANGO_OUTPUT_S3_AWS_ACCESS_KEY_ID=""

# The AWS secret key to be used when uploading scan output to an S3 bucket
DJANGO_OUTPUT_S3_AWS_SECRET_ACCESS_KEY=""

# An optional AWS session token
DJANGO_OUTPUT_S3_AWS_SESSION_TOKEN=""

# The AWS region where your S3 bucket is located (e.g., "us-east-1")
DJANGO_OUTPUT_S3_AWS_DEFAULT_REGION=""

# The name of the S3 bucket where scan output should be stored
DJANGO_OUTPUT_S3_AWS_OUTPUT_BUCKET=""

```

prowler/lib/outputs/finding.py
```diff
@@ -267,3 +276,179 @@ def generate_output(
...
+            Finding: A new Finding instance populated with data from the provided model.
+        """
+        # Missing Finding's API values
+        finding.muted = False
+        finding.resource_details = ""
+        resource = finding.resources.first()
+        finding.resource_arn = resource.uid
+        finding.resource_name = resource.name
+
+        finding.resource = model_to_dict(resource)
```
jfagoagas:
```suggestion
        finding.resource = {}
```

I'd exclude this as it is only done for the OCSF in the SDK because the API has not all the values that this field is intended to have, I prefer to add a TODO and work on it on the future once we have the full resource data into the Prowler API DB.

api/src/backend/config/django/base.py
jfagoagas: That's a good point.

api/src/backend/tasks/jobs/export.py
```diff
@@ -0,0 +1,156 @@
...
+        ClientError, NoCredentialsError, or ParamValidationError if both attempts to create a client fail.
+    """
+    s3_client = None
+    try:
+        s3_client = boto3.client(
+            "s3",
+            aws_access_key_id=env.str("DJANGO_OUTPUT_AWS_ACCESS_KEY_ID"),
+            aws_secret_access_key=env.str("DJANGO_OUTPUT_AWS_SECRET_ACCESS_KEY"),
+            aws_session_token=env.str("DJANGO_OUTPUT_AWS_SESSION_TOKEN"),
+            region_name=env.str("DJANGO_OUTPUT_AWS_DEFAULT_REGION"),
```
jfagoagas: Since these variables are declared in the `base.py` I think you can access them using
```python
from django.conf import settings

settings.<ENV>
```

jfagoagas: Not having `DJANGO_OUTPUT_AWS_DEFAULT_REGION` makes the endpoint to raise a `HTTP 500 Internal Server Error`. I think we should handle that too.

api/src/backend/api/v1/views.py
```diff
@@ -1181,6 +1190,92 @@ def partial_update(self, request, *args, **kwargs):
...
+                return Response(
+                    {"detail": "There is a problem with the AWS credentials."},
+                    status=status.HTTP_403_FORBIDDEN,
+                )
+
+            bucket_name = env.str("DJANGO_OUTPUT_AWS_S3_OUTPUT_BUCKET")
+            key = output_path[len(f"s3://{bucket_name}/") :]
+            s3_object = s3_client.get_object(Bucket=bucket_name, Key=key)
+            file_content = s3_object["Body"].read()
+            filename = os.path.basename(output_path.split("/")[-1])
```
jfagoagas: We need to handle `NoSuchKey` if the file got deleted from S3.

jfagoagas (changes requested):
We are getting closer to the 🏁 line! Apart from the comments above I detected the following while testing:

- In the generated outputs `resource_name` has the same value as `resource_uid` and that also causes to have the `resource_uid` within the `finding_uid` instead of `resource_name`, e.g.: in AWS for an `accessanalyzer` check, `resource_name = analyzer/unknown` and `resource_uid = arn:aws:accessanalyzer:eu-north-1:111122223333:analyzer/unknown`.
- I realised that we need to verify that `DJANGO_OUTPUT_AWS_DEFAULT_REGION` must not be empty because the API will raise 500's while trying to connect with AWS.
- We need to handle if S3 raises `NoSuchKey`, e.g.: the object was deleted, to return a 404 instead of raising a 500.
- For some reason pushing to S3 is not storing `scan.output_path`  but the file is correctly pushed to the configured S3 bucket.

I resolved the addressed comments and left open the ones that weren't.

UPDATE:
- We need to add a test to verify that the API Finding is transformed in an SDK Finding with all the required fields.

prowler/lib/outputs/finding.py
AdriiiPRodri: Yes it is included

jfagoagas: Awesome, thank you!

api/src/backend/api/v1/views.py
jfagoagas: Tested and fixed ✅

api/src/backend/tasks/jobs/export.py
jfagoagas: Done!

jfagoagas: Fixed and tested

api/src/backend/api/v1/views.py
```diff
@@ -1181,6 +1190,102 @@ def partial_update(self, request, *args, **kwargs):
...
+        summary="Download ZIP report",
+        description="Returns a ZIP file containing the requested report",
+        request=ScanReportSerializer,
+        responses={
+            200: OpenApiResponse(description="Report obtained successfully"),
+            202: OpenApiResponse(description="The task is in progress"),
+            403: OpenApiResponse(description="There is a problem with credentials"),
+            404: OpenApiResponse(description="The scan has no reports"),
+        },
+    )
```
vicferpoy: Could we move this with the rest of Scan schemas? To maintain consistency.

vicferpoy (changes requested):

api/src/backend/tasks/jobs/export.py
```diff
@@ -0,0 +1,157 @@
...
+        s3_client = boto3.client(
+            "s3",
+            aws_access_key_id=settings.DJANGO_OUTPUT_S3_AWS_ACCESS_KEY_ID,
+            aws_secret_access_key=settings.DJANGO_OUTPUT_S3_AWS_SECRET_ACCESS_KEY,
+            aws_session_token=settings.DJANGO_OUTPUT_S3_AWS_SESSION_TOKEN,
+            region_name=settings.DJANGO_OUTPUT_S3_AWS_DEFAULT_REGION,
+        )
+        s3_client.list_buckets()
+    except (ClientError, NoCredentialsError, ParamValidationError, ValueError):
+        s3_client = boto3.client("s3")
```
jfagoagas, line 84: If credentials are not configured this is raising an exception, could you please check it when you get a chance?

When this happens the call to _upload_to_s3 within generate_outputs raises an exception and the output location is not stored nor locally.

AdriiiPRodri: `get_s3_client is` only invoked if you have configured `DJANGO_OUTPUT_S3_AWS_OUTPUT_BUCKET`, which indicates that you want the outputs to be uploaded to S3, then it gives an exception because if a user only configures this it is understood that he is running Prowler in AWS (it is not necessary to set the credentials) or he has made a mistake because he has not set the other variables if he wants access through credentials

jfagoagas: Good point, thanks for pointing that out. The only issue I see with that is that if something fails there is no fallback to the local storage.

jfagoagas (dismissed):
Finally this get to an end, super happy about this new functionality, it required a great effort but it is done 👏 awesome work 🚀

vicferpoy (dismissed):

vicferpoy (approved):

jfagoagas (approved):

PR #6896, opened 2025-02-11
fix(aws): SNS threw IndexError if SubscriptionArn is PendingConfirmation

kagahd:
### Context

The `sns_service.py` threw the error `IndexError[100]: list index out of range` upon extracting the region from the field `SubscriptionArn` of a subscription if the field `SubscriptionArn` was not an ARN but the value `PendingConfirmation`.

### Description

This PR fixes it by verifying that the `SubscriptionArn` has at least 4 parts after the split by the character `:` in order to extract the 4th part, which is the `region`. If it doesn't, the `region` is set to `unkown`. Instead of `unknown` it could have been set also to `regional_client.region` or `topic.region` but I'm not sure if they can differ from the region of a subscription. Feel free to adapt it.

### Checklist

#### API

### License

### DIFF
--- a/prowler/providers/aws/services/sns/sns_service.py
+++ b/prowler/providers/aws/services/sns/sns_service.py
@@ -97,7 +97,11 @@ class SNS(AWSService):
                             owner=sub["Owner"],
                             protocol=sub["Protocol"],
                             endpoint=sub["Endpoint"],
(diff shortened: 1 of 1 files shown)


MrCloudSec (approved):
Good catch @kagahd !!

PR #6906, opened 2025-02-12
feat(social-login): Add social login integration for Google and Github OAuth providers

vicferpoy:
### Description

We have implemented social login though Google and GitHub OAuth providers. To use this new feature, we must configure a redirection URL to one of our new endpoints. These endpoints are:

- `POST /tokens/google`
- `POST /tokens/github`

Both endpoints must include a JSON body:

```json
{
  "code": "<code>"
}
```

This code is retrieved from requesting the OAuth provider the authentication code.

The response will be the same as `POST /tokens`, but with different `type` depending on the OAuth provider.

If the authenticated user's email does not exist in the database already, under the `users` table, it will be created in addition to a default tenant and a default set of RBAC admin permissions. If it does, a relationship between the existent user and the social account will be created in order to obtain a token through this authentication mechanism.

In order to allow switching tenants using this method, a new endpoint has been added to ask for a new token, providing a required `tenant_id`. In order to request to this endpoint, the user must be logged in through a Bearer token as every other endpoint:

`POST /tokens/switch`

```json
{
  "data": {
    "type": "tokens-switch-tenant",
    "attributes": {
      "tenant_id": "0412980b-06e3-436a-ab98-3c9b1d0333d3"
    }
  }
}
```

image

### Checklist

#### API

### License


api/src/backend/api/tests/integration/test_authentication.py
```diff
@@ -177,3 +179,122 @@ def test_user_me_when_inviting_users(create_test_user, tenants_fixture, roles_fi
     user2_me = client.get(reverse("user-me"), headers=user2_headers)
     assert user2_me.status_code == 200
     assert user2_me.json()["data"]["attributes"]["email"] == user2_email
+
+
+@pytest.mark.django_db
+class TestTokenSwitchTenant:
+    def test_switch_tenant_with_valid_token(self, tenants_fixture, providers_fixture):
```
drewkerrigan, line 186: Is there a test to verify a user cannot switch to a tenant they don't belong to?

api/src/backend/config/settings/social_login.py
```diff
@@ -0,0 +1,54 @@
...
+        },
+    },
+    "github": {
+        "APP": {
+            "client_id": GITHUB_OAUTH_CLIENT_ID,
+            "secret": GITHUB_OAUTH_CLIENT_SECRET,
+        },
+        "SCOPE": [
+            "user",
+            "repo",
```
drewkerrigan: Do we need the `repo` scope? According to the docs:

"Grants full access to public and private repositories including read and write access to code, commit statuses, repository invitations, collaborators, deployment statuses, and repository webhooks." (https://docs.github.com/en/apps/oauth-apps/building-oauth-apps/scopes-for-oauth-apps)

api/src/backend/api/tests/integration/test_authentication.py
vicferpoy: There is! We already talked about this in a meeting.

api/src/backend/config/settings/social_login.py
vicferpoy: We don't. Thank you for noticing.

drewkerrigan (approved):
Looks great! Thank you for reviewing with me IRL.

PR #6908, opened 2025-02-12
fix(aws): handle `AccessDenied` when retrieving resource policy

MrCloudSec:
### Description

When Prowler was getting an Access Denied in retrieving resource policies, it was giving a false positive in the policy checks. Therefore, this PR makes sure that Prowler only gives findings when a resource policy is correctly retrieved.

Thanks @wunzeco for the heads up!
### Checklist

#### API

### License

### DIFF
--- a/prowler/providers/aws/services/efs/efs_service.py
+++ b/prowler/providers/aws/services/efs/efs_service.py
@@ -164,7 +164,7 @@ class FileSystem(BaseModel):
     id: str
     arn: str
     region: str
-    policy: Optional[dict] = {}
+    policy: Optional[dict]
     backup_policy: Optional[str] = "DISABLED"
     encrypted: bool
     availability_zone_id: Optional[str]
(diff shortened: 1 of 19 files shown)


prowler/providers/aws/services/efs/efs_service.py
```diff
@@ -164,7 +164,7 @@ class FileSystem(BaseModel):
     id: str
     arn: str
     region: str
-    policy: Optional[dict] = {}
+    policy: Optional[dict]
```
pedrooot, line 167: Do you think that we should make this? 
```suggestion
    policy: Optional[dict] = None
```
It's done this way in `prowler/providers/aws/services/s3/s3_service.py`

prowler/providers/aws/services/s3/s3_service.py
```diff
@@ -681,7 +681,7 @@ class Bucket(BaseModel):
     logging: bool = False
     public_access_block: Optional[PublicAccessBlock]
     acl_grantees: List[ACL_Grantee] = Field(default_factory=list)
-    policy: Dict = Field(default_factory=dict)
+    policy: Optional[dict] = None
```
pedrooot:
```suggestion
    policy: Optional[dict]
```

pedrooot (approved):
Thanks for this!! 🔝

PR #6919, opened 2025-02-13
fix(aws): codebuild service threw KeyError for projects type CODEPIPELINE

kagahd:
### Context

The `codebuild_service.py` threw the error `KeyError[87]: 'location'` for projects of type `CODEPIPELINE`.

### Description
This PR fixes it by verifying that the field `location` exists before accessing it. In case of projects of type `CODEPIPELINE`, the field `location` does not exist and is therefore set to an empty string.

See [AWS documentation](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/codebuild/client/batch_get_projects.html) about the `location` property within the response of `batch_get_projects()` method: 

> **location** (string) –
> 
> Information about the location of the source code to be built. Valid values include:
> 
> For source code settings that are specified in the source action of a pipeline in **CodePipeline, location should not be specified.** If it is specified, CodePipeline ignores it. This is because CodePipeline uses the settings in a pipeline’s source action instead of this value.

### Checklist

#### API

### License

### DIFF
--- a/prowler/providers/aws/services/codebuild/codebuild_service.py
+++ b/prowler/providers/aws/services/codebuild/codebuild_service.py
@@ -84,21 +84,13 @@ class Codebuild(AWSService):
             if project_info["source"]["type"] != "NO_SOURCE":
                 project.source = Source(
                     type=project_info["source"]["type"],
-                    location=(
-                        project_info["source"]["location"]
-                        if project_info["source"].get("location") is not None
-                        else ""
-                    ),
+                    location=project_info["source"].get("location", ""),
                 )
             project.secondary_sources = []
             for secondary_source in project_info.get("secondarySources", []):
                 source_obj = Source(
                     type=secondary_source["type"],
-                    location=(
-                        secondary_source["location"]
-                        if secondary_source.get("location") is not None
-                        else ""
-                    ),
+                    location=secondary_source.get("location", ""),
                 )
                 project.secondary_sources.append(source_obj)
             environment = project_info.get("environment", {})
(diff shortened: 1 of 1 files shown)


prowler/providers/aws/services/codebuild/codebuild_service.py
```diff
@@ -84,12 +84,21 @@ def _batch_get_projects(self, project):
             if project_info["source"]["type"] != "NO_SOURCE":
                 project.source = Source(
                     type=project_info["source"]["type"],
-                    location=project_info["source"]["location"],
+                    location=(
+                        project_info["source"]["location"]
+                        if project_info["source"].get("location") is not None
+                        else ""
+                    ),
```
jfagoagas: You can also do `project_info["source"].get("location","") to have it in one line without the `if/else`

prowler/providers/aws/services/codebuild/codebuild_service.py
```diff
@@ -84,12 +84,21 @@ def _batch_get_projects(self, project):
...
                 )
             project.secondary_sources = []
             for secondary_source in project_info.get("secondarySources", []):
                 source_obj = Source(
-                    type=secondary_source["type"], location=secondary_source["location"]
+                    type=secondary_source["type"],
+                    location=(
+                        secondary_source["location"]
+                        if secondary_source.get("location") is not None
+                        else ""
```
jfagoagas: You can also do `secondary_source["location","") to have it in one line without the `if/else`

jfagoagas (dismissed):

MrCloudSec (approved):
Thanks @kagahd, good catch!!

PR #6952, opened 2025-02-17
refactor(microsoft365): `CheckReportMicrosoft365` and `resource metadata`

HugoPBrito:
### Context

Prowler’s resource metadata structure could be improved, and this PR aims to establish a foundation for best practices in provider development, benefiting future providers as well.

### Description

Several changes have been applied to refactor the `Microsoft365` provider’s reporting model, including:

- Renaming `Check_Report_Microsoft365` → `CheckReportMicrosoft365`.
- Adding `resource_name` and `resource_id` to `__init__`.
- Updating tests accordingly.
- `Removing `report.resource_id` and `report.resource_name` from existing Microsoft365` checks.
- Adding `asserts` in check tests to verify `check_report.resource`.
- Adding `docstrings` to checks.

### Checklist

#### API

### License

### DIFF
--- a/prowler/lib/check/models.py
+++ b/prowler/lib/check/models.py
@@ -547,8 +547,8 @@ class CheckReportMicrosoft365(Check_Report):
         self,
         metadata: Dict,
         resource: Any,
-        resource_name: str = "",
-        resource_id: str = "",
+        resource_name: str,
+        resource_id: str,
         resource_location: str = "global",
     ) -> None:
         """Initialize the Microsoft365 Check's finding information.
--- a/prowler/lib/check/models.py
+++ b/prowler/lib/check/models.py
@@ -536,26 +536,34 @@ class Check_Report_Kubernetes(Check_Report):
 
 
 @dataclass
-class Check_Report_Microsoft365(Check_Report):
+class CheckReportMicrosoft365(Check_Report):
     """Contains the Microsoft365 Check's finding information."""
 
     resource_name: str
     resource_id: str
     location: str
 
-    def __init__(self, metadata: Dict, resource: Any) -> None:
+    def __init__(
+        self,
+        metadata: Dict,
+        resource: Any,
+        resource_name: str = "",
+        resource_id: str = "",
+        resource_location: str = "global",
+    ) -> None:
         """Initialize the Microsoft365 Check's finding information.
 
         Args:
             metadata: The metadata of the check.
             resource: Basic information about the resource. Defaults to None.
+            resource_name: The name of the resource related with the finding.
+            resource_id: The id of the resource related with the finding.
+            resource_location: The location of the resource related with the finding.
         """
         super().__init__(metadata, resource)
-        self.resource_name = getattr(
-            resource, "name", getattr(resource, "resource_name", "")
-        )
-        self.resource_id = getattr(resource, "id", getattr(resource, "resource_id", ""))
-        self.location = getattr(resource, "location", "global")
+        self.resource_name = resource_name
+        self.resource_id = resource_id
+        self.location = resource_location
 
 
 # Testing Pending
--- a/prowler/providers/microsoft365/services/entra/entra_thirdparty_integrated_apps_not_allowed/entra_thirdparty_integrated_apps_not_allowed.py
+++ b/prowler/providers/microsoft365/services/entra/entra_thirdparty_integrated_apps_not_allowed/entra_thirdparty_integrated_apps_not_allowed.py
@@ -25,23 +25,32 @@ class entra_thirdparty_integrated_apps_not_allowed(Check):
         """
         findings = []
         auth_policy = entra_client.authorization_policy
+
         report = CheckReportMicrosoft365(
             metadata=self.metadata(),
-            resource=auth_policy,
-            resource_name=getattr(auth_policy, "name", "Authorization Policy"),
-            resource_id=getattr(auth_policy, "id", "authorizationPolicy"),
+            resource=auth_policy if auth_policy else {},
+            resource_name=auth_policy.name if auth_policy else "Authorization Policy",
+            resource_id=auth_policy.id if auth_policy else "authorizationPolicy",
         )
-        report.status = "FAIL"
-        report.status_extended = "App creation is not disabled for non-admin users."
 
-        # Check if the policy disables app creation for non-admin users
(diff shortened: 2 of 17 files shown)


prowler/lib/check/models.py
```diff
@@ -536,26 +536,34 @@ def __init__(self, metadata: Dict, resource: Any) -> None:
...
     resource_id: str
     location: str
 
-    def __init__(self, metadata: Dict, resource: Any) -> None:
+    def __init__(
+        self,
+        metadata: Dict,
+        resource: Any,
+        resource_name: str = "",
+        resource_id: str = "",
```
jfagoagas: I think is highly important not to set default values as empty strings for this two fields to enforce that are always passed with the expected values.

prowler/providers/microsoft365/services/entra/entra_thirdparty_integrated_apps_not_allowed/entra_thirdparty_integrated_apps_not_allowed.py
```diff
@@ -1,18 +1,40 @@
...
+        findings = []
         auth_policy = entra_client.authorization_policy
-        report = Check_Report_Microsoft365(self.metadata(), auth_policy)
-        report.resource_name = getattr(auth_policy, "name", "Authorization Policy")
-        report.resource_id = getattr(auth_policy, "id", "authorizationPolicy")
+        report = CheckReportMicrosoft365(
+            metadata=self.metadata(),
+            resource=auth_policy,
+            resource_name=getattr(auth_policy, "name", "Authorization Policy"),
+            resource_id=getattr(auth_policy, "id", "authorizationPolicy"),
```
jfagoagas:
```suggestion
            resource_name=auth_policy.name,
            resource_id=auth_policy.id,
```

Why are you using the `getattr` if the object has `name` and `id`?

tests/providers/microsoft365/services/admincenter/admincenter_groups_not_public_visibility/admincenter_groups_not_public_visibility_test.py
```diff
@@ -60,6 +60,11 @@ def test_admincenter_user_no_admin(self):
             assert len(result) == 1
             assert result[0].status == "PASS"
             assert result[0].status_extended == "Group Group1 has Private visibility."
```
jfagoagas, line 62: Do not forget to assert `location`, review it in all the checks please.

jfagoagas (changes requested):

prowler/providers/microsoft365/services/entra/entra_thirdparty_integrated_apps_not_allowed/entra_thirdparty_integrated_apps_not_allowed.py
HugoPBrito: Mario designed it this way to prevent the check from crashing when the authentication policy couldn't be retrieved. If the access is directly made the test `Test_entra_thirdparty_integrated_apps_not_allowed` crashes instead of returning a `FAIL` finding.

jfagoagas: but as far as I can see these fields are present in the object created in the service, am I missing something?

HugoPBrito: If the call to `msgraph` fails by any chance:


```python
auth_policy = await self.client.policies.authorization_policy.get()
```

 The object will be stored as `None` at the direct access will break the execution with `'NoneType' object has no attribute 'name'`.

It seems Mario implemented a safe retrieval flow to ensure that if we encounter issues fetching data from the API, the process doesn’t break. Instead, the tool returns a `FAIL` finding, due to the inability to properly assess the tenant's configuration.

Instead of that, we can craft a default object with non compliant configuration on purpose intead of `None` in the cases we can not retrieve the information, to standarize `CheckReportMicrosoft365` creation in all checks with direct access.

jfagoagas: That's interesting for sure, I did not realise of that part 🤔

AdriiiPRodri: For me it is fine as it is or if we want it to be understood better I think this way is more readable, although `getattr` doesn't seem wrong in this case:

```
resource_name = auth_policy.name if auth_policy and auth_policy.name else “Authorization policy”.
resource_id = auth_policy.id if auth_policy and auth_policy.id else “authorizationPolicy”
```

Another option is to make a refactor so that the object that is returned always has all the fields (good practice), in case of not being able to dedicate time to this I would stay with the option above

HugoPBrito: I have implemented a new output for `report.status_extended` to let the user know exactly what is happening in addition to your comments @AdriiiPRodri. Thanks for your help!

jfagoagas (dismissed):
That's great, thanks for this work @HugoPBrito 👏 

I've only one question left:
- Is `"Authorization Policy was not found."` expected? I mean, is it a security issue if the Authorization Policy is not found? What happens if Prowler has no permissions? I think raising a `WARNING` error in the service is more appropiate but again I don't have much context in M365, so please shed a light 💡

HugoPBrito:
> That's great, thanks for this work @HugoPBrito 👏 
> 
> I've only one question left:
> - Is `"Authorization Policy was not found."` expected? I mean, is it a security issue if the Authorization Policy is not found? What happens if Prowler has no permissions? I think raising a `WARNING` error in the service is more appropiate but again I don't have much context in M365, so please shed a light 💡 

I implemented it thinking about:
- That configuration does not exist in the tenant
- There are not enough permissions
- An unexpected error occurs

Rather than telling the user that their configuration is incorrect, it's better to inform them that it couldn't be read or that they simply don't have it.

jfagoagas (dismissed):

MrCloudSec (approved):

PR #6985, opened 2025-02-19
feat(aws): add check `secretsmanager_has_restrictive_resource_policy`

kagahd:
### Context

This feature request offers a new AWS check `secretsmanager_has_restrictive_resource_policy`. 
The check evaluates resource-based policies for AWS Secrets Manager secrets.

### Description

To pass the check, resource-based policies attached to secrets must meet **all** of the following criteria:

## 1. Explicit "Deny" Statement for Unauthorized Principals

The resource policy must include an explicit `Deny` statement that applies to **all principals**, **all actions**, and **all resources**, unless exceptions are defined in the `StringNotEquals` Condition block.

**Example**

```
{
  "Sid": "DenyUnauthorizedPrincipals",
  "Effect": "Deny",
  "Principal": "*",
  "Action": "*",
  "Resource": "*",
  "Condition": {
    "StringNotEquals": {
      "aws:PrincipalArn": [
        "<ASSUMED-ROLE-ARN-FOR-SECURY-AUDIT>",
        "arn:aws:iam::<YOUR_AWS_ACCOUNT_ID>:role/SomeOtherRole"
      ]
    }
  }
}
...<further required policy statements>...
```
This structure ensures that unauthorized access is denied across the board.

**Principal: `"*"`**  
Applies the denial to all entities, internal or external, preventing unauthorized principals from accessing the secret.

**Action: `"*"` or `"secretsmanager:*"`**  
Covers all possible actions, ensuring no unauthorized operation—directly or indirectly—can be performed on the secret. Using `"secretsmanager:*"` is acceptable to limit the denial to actions related specifically to Secrets Manager.

**Resource: `"*"`**  
Because resource-based policies are inherently tied to the secret they protect, this field can either use a wildcard (`"*"` ) for broader applicability or the exact ARN of the secret for stricter targeting.

The `StringNotEquals` Condition block refines the denial by allowing **exceptions**, such as specific roles defined in `aws:PrincipalArn`.

One **mandatory** exception must be given to `<ASSUMED-ROLE-ARN-FOR-SECURY-AUDIT>` because this role, which is deployed in every organization's AWS account, must be allowed to audit the secret by the IT security team.

All Principal ARNs that will have their own policy entry must also be listed here. Otherwise, this rule would already block any access attempts.

### Using wildcards in Principal ARNs

The check allows using an `ArnNotLike` condition with key `aws:PrincipalArn` within the explicit `Deny` statement for unauthorized principals. The role name must contain **at least 12 characters before the wildcard**.

**Example**

```
{
  "Sid": "DenyUnauthorizedPrincipals",
  "Effect": "Deny",
  "Principal": "*",
  "Action": "*",
  "Resource": "*",
  "Condition": {
    "StringNotEquals": {
      "aws:PrincipalArn": [
        "<ASSUMED-ROLE-ARN-FOR-SECURY-AUDIT>",
        "arn:aws:iam::<YOUR_AWS_ACCOUNT_ID>:role/SomeOtherRole"
      ]
    },
    "ArnNotLike": {
      "aws:PrincipalArn": [
        "arn:aws:iam::<YOUR_AWS_ACCOUNT_ID>:role/<PREFIX>*",
        "arn:aws:iam::<YOUR_AWS_ACCOUNT_ID>:role/<PREFIX>*"
      ]
    }
  }
}
...<further required policy statements>...
```

### Excluding AWS services from the Deny

If AWS services must also be excluded from the Deny statement, the `StringNotEqualsIfExists` condition operator must be used instead of `StringNotEquals`. This ensures that both principals and services can be excluded even though only one of them may exist in the request.

Additionally, a `Null` condition block must be included which contains both `aws:PrincipalArn` and `aws:PrincipalServiceName` with value `"true"`. This ensures that requests missing both keys are denied.

**Example**

```
{
  "Sid": "DenyUnauthorizedPrincipals",
  "Effect": "Deny",
  "Principal": "*",
  "Action": "*",
  "Resource": "*",
  "Condition": {
    "StringNotEqualsIfExists": {
      "aws:PrincipalArn": [
        "<ASSUMED-ROLE-ARN-FOR-SECURY-AUDIT>",
        "arn:aws:iam::<YOUR_AWS_ACCOUNT_ID>:role/SomeOtherRole"
      ],
      "aws:PrincipalServiceName": "appflow.amazonaws.com"
    },
    "Null": {
      "aws:PrincipalArn": "true",
      "aws:PrincipalServiceName": "true"
    }
  }
}
...<further required policy statements>...
```

## 2. Restrict Access Outside the Organization

The second **mandatory** `Deny` statement ensures that principals **outside the AWS Organization** are denied access to the secret.
```
{
  "Sid": "DenyOutsideOrganization",
  "Effect": "Deny",
  "Principal": "*",
  "Action": "*",
  "Resource": "*",
  "Condition": {
    "StringNotEquals": {
      "aws:PrincipalOrgID": "<YOUR_ORG_ID>"
    }
  }
}
...<further required policy statements>...
```

The `StringNotEquals` condition ensures that the denial applies to any request coming from outside of the organization, as identified by the `aws:PrincipalOrgID` condition key.

The value(s) of `aws:PrincipalOrgID` can be configured via the key `organizations_trusted_ids` in prowler's [config.yaml](https://github.com/prowler-cloud/prowler/blob/master/prowler/config/config.yaml).

If the AWS account is **not part of an organization**, this value can be left empty in the configuration. In that case, the check `secretsmanager_has_restrictive_resource_policy` will **not require** the `DenyOutsideOrganization` statement.

### Dealing with AWS Services

AWS services (e.g. `"appflow.amazonaws.com"`) do not send a `PrincipalOrgID` in their requests. Therefore the service name must also be added to the condition block.

**Example**

```
{
  "Sid": "DenyOutsideOrganization",
  "Effect": "Deny",
  "Principal": "*",
  "Action": "*",
  "Resource": "*",
  "Condition": {
    "StringNotEquals": {
      "aws:PrincipalOrgID": "<YOUR_ORG_ID>",
      "aws:PrincipalServiceName": "appflow.amazonaws.com"
    }
  }
}
...<further required policy statements>...
```

This statement alone does **not** guarantee that the service runs within the same AWS account or organization.  
Therefore the service must additionally be restricted in its `Allow` statement using `aws:SourceAccount` (see section **4. Allow Statement for AWS services**).

## 3. Deny Statement to Permit Only Specific Actions for Designated Principals

These `Deny` statements ensure that specific principals are restricted to performing **only specific actions**.

These statements are **optional**, but if omitted the principals listed in `DenyUnauthorizedPrincipals` will effectively be denied all actions.

Any principal listed in the `StringNotEquals` condition of `DenyUnauthorizedPrincipals` must have a corresponding `Deny` statement with a tailored `NotAction` clause.

Wildcards (`*`) are **not allowed** in `NotAction`. Refer to the supported [AWS Secrets Manager actions](https://docs.aws.amazon.com/secretsmanager/latest/apireference/API_Operations.html).

The IT security team must be able to audit the resource-based policy. Therefore the principal `<ASSUMED-ROLE-ARN-FOR-SECURY-AUDIT>` must be permitted the actions:

- `secretsmanager:DescribeSecret`
- `secretsmanager:GetResourcePolicy`

**Example**

```
{
  "Sid": "AllowAuditPolicyRead",
  "Effect": "Deny",
  "Principal": {
    "AWS": "<ASSUMED-ROLE-ARN-FOR-SECURY-AUDIT>"
  },
  "NotAction": [
    "secretsmanager:DescribeSecret",
    "secretsmanager:GetResourcePolicy"
  ],
  "Resource": "*"
},
{
  "Sid": "AllowSecretAccessForSomeOtherRole",
  "Effect": "Deny",
  "Principal": {
    "AWS": "arn:aws:iam::<YOUR_AWS_ACCOUNT_ID>:role/SomeOtherRole"
  },
  "NotAction": [
    "secretsmanager:GetSecretValue"
  ],
  "Resource": "*"
}
...<further required policy statements>...
```

### Using wildcards for principals

The check allows using `ArnLike` with `aws:PrincipalArn` to define the allowed actions for principals that were specified via `ArnNotLike` in the `DenyUnauthorizedPrincipals` statement.

The role name must again contain **at least 12 characters before the wildcard**.

**Example**

```
{
  "Sid": "AllowReadOnlyRolePrefixAccess",
  "Effect": "Deny",
  "Principal": "*",
  "NotAction": [
    "secretsmanager:GetSecretValue",
    "secretsmanager:DescribeSecret"
  ],
  "Resource": "*",
  "Condition": {
    "ArnLike": {
      "aws:PrincipalArn": [
        "arn:aws:iam::<YOUR_AWS_ACCOUNT_ID>:role/<PREFIX>*",
        "arn:aws:iam::<YOUR_AWS_ACCOUNT_ID>:role/<PREFIX>*"
      ]
    }
  }
}
...<further required policy statements>...
```

## 4. Allow Statement for AWS Services

AWS services require an explicit `Allow` statement to perform actions on the secret.

Two requirements apply:

1. **No wildcard (`*`) in any Action**
2. The AWS **SourceAccount must match the current account** to prevent access from services running in external AWS accounts. Additional restrictive conditions (e.g. `ArnLike` on `aws:SourceArn`) are accepted alongside `aws:SourceAccount` and will not cause the check to fail.

Both simple service principals (e.g. `appflow.amazonaws.com`) and regionalized service principals (e.g. `logs.eu-central-1.amazonaws.com`) are supported.

**Example**

```
{
  "Sid": "AllowAppFlowAccess",
  "Effect": "Allow",
  "Principal": {
    "Service": "appflow.amazonaws.com"
  },
  "Action": [
    "secretsmanager:GetSecretValue",
    "secretsmanager:PutSecretValue",
    "secretsmanager:DeleteSecret",
    "secretsmanager:DescribeSecret",
    "secretsmanager:UpdateSecret"
  ],
  "Resource": "*",
  "Condition": {
    "StringEquals": {
      "aws:SourceAccount": "<YOUR_AWS_ACCOUNT_ID>"
    }
  }
}
...<further required policy statements>...
```

### Cross-account access

Cross-account access to secrets is **not allowed**.  
Resource-based policies that allow cross-account access will cause the check to **fail on purpose**. This behavior is intentional to ensure that any cross-account access is made explicitly visible during reviews and compliance checks, since from an IT Security perspective it expands the trust boundary and therefore requires deliberate assessment and clear justification.

## Complete Resource-Based Policy Example

Below is a full example resource policy for scenarios where **only IAM principals require access**.

```
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Sid": "DenyUnauthorizedPrincipals",
      "Effect": "Deny",
      "Principal": "*",
      "Action": "*",
      "Resource": "*",
      "Condition": {
        "StringNotEquals": {
          "aws:PrincipalArn": [
            "<ASSUMED-ROLE-ARN-FOR-SECURY-AUDIT>",
            "arn:aws:iam::<YOUR_AWS_ACCOUNT_ID>:role/SomeOtherRole"
          ]
        }
      }
    },
    {
      "Sid": "DenyOutsideOrganization",
      "Effect": "Deny",
      "Principal": "*",
      "Action": "*",
      "Resource": "*",
      "Condition": {
        "StringNotEquals": {
          "aws:PrincipalOrgID": "<YOUR_ORG_ID>"
        }
      }
    },
    {
      "Sid": "AllowAuditPolicyRead",
      "Effect": "Deny",
      "Principal": {
        "AWS": "<ASSUMED-ROLE-ARN-FOR-SECURY-AUDIT>"
      },
      "NotAction": [
        "secretsmanager:DescribeSecret",
        "secretsmanager:GetResourcePolicy"
      ],
      "Resource": "*"
    },
    {
      "Sid": "AllowSecretAccessForSomeOtherRole",
      "Effect": "Deny",
      "Principal": {
        "AWS": "arn:aws:iam::<YOUR_AWS_ACCOUNT_ID>:role/SomeOtherRole"
      },
      "NotAction": [
        "secretsmanager:GetSecretValue"
      ],
      "Resource": "*"
    }
  ]
}
```

---

### Unit Tests

As shown in `secretsmanager_has_restrictive_resource_policy_test.py`, several unit tests were created.  
Three of them alone cover already **105 different scenarios** by modifying parts of an otherwise valid resource policy to ensure the check correctly **fails**.
- `test_secretsmanager_policies_for_principals` = 73 test cases
- `test_secretsmanager_policies_for_principals_and_services` = 29 test cases
- `test_policy_with_arn_not_like` = 3 test cases

Each negative test case is followed by a **counter-test** that removes the invalid change to ensure the policy **passes**, verifying that the test logic itself is correct.

`@pytest.mark.parametrize` is used to reduce redundancy while maintaining readability.

Additionally, **3 regression tests** cover order-independent cross-account detection, service Allow with extra restrictive conditions, and regionalized service principals.

---

### Checklist

    - If so, do we need to update permissions for the provider? **No**
- [x] Review if code is documented following the Google Python style guide.
- [x] Review if changes to `README.md` are required.

#### API

- [x] Check if version updates are required.
- [x] Ensure new entries are added to `CHANGELOG.md` if applicable.

---

### License


tests/providers/aws/services/secretsmanager/secretsmanager_has_restrictive_resource_policy/secretsmanager_has_restrictive_resource_policy_test.py
jfagoagas, line 1: I think https://docs.pytest.org/en/stable/example/parametrize.html would be ideal for this tests, let me know what you think 🙌

kagahd: Oh, I've never heard of it but seems that's really nice! Let me adapt the code asap.

jfagoagas:
Hello @kagahd I apologise because this PR got lost in our inbox. The team is going to review it as soon as possible.

Thanks!

prowler/providers/aws/services/secretsmanager/secretsmanager_has_restrictive_resource_policy/secretsmanager_has_restrictive_resource_policy.py
MrCloudSec, line 1: Please @kagahd, can you add the logic of handling the policies in `prowler/providers/aws/services/iam/lib/policy.py`.

kagahd: The check is about to change because of new use cases we were made aware of. Once integrated, I'll push it here.

MrCloudSec: Sounds good @kagahd, let us know when you push it!

HugoPBrito:
Hi @kagahd,

Are you planning on continuing this development? Just to know if we should proceed to close it/try to finish it ourselves.

Thanks!

kagahd:
> Are you planning on continuing this development? Just to know if we should proceed to close it/try to finish it ourselves.


Thanks @HugoPBrito for reminding me. 
Yes, we are running already a "finalized" version of this check in our company and I think it's valuable to share it with the community. As soon as I have time for it, I will update this PR or do you think it's better to create a new one?

HugoPBrito:
> > Are you planning on continuing this development? Just to know if we should proceed to close it/try to finish it ourselves.
> 
> 
> Thanks @HugoPBrito for reminding me. 
> Yes, we are running already a "finalized" version of this check in our company and I think it's valuable to share it with the community. As soon as I have time for it, I will update this PR or do you think it's better to create a new one?

Hi @kagahd!

As you prefer. As long as they're linked in case you create a new one, that's up to what's best for you.

kagahd:
Hi @HugoPBrito, I have now updated the PR to the latest version, including the revised description.
We are aware that this check is quite complex and extensive, and that some engineers responsible for keeping AWS environments compliant with IT Security requirements may find it challenging to implement and remediate all related findings. Nevertheless, we consider this check highly valuable from a security perspective, because it helps make risky access patterns visible and enforces a more deliberate review of Secrets Manager resource policies.
For that reason, we continue to strongly support this check and will work closely with the affected teams to help them understand the requirements and remediate the identified risks.

HugoPBrito:
Hi @kagahd,

I get your point, but we will need to evaluate it. I believe this will end up merged, since it brings a lot of value, and if someone don't want to deal with it, they can always mute it.

Since the pr is huge and complex, the review will be most likely slow. In the meantime, please adapt the metadata to the new standard, defined in our docs on the [metadata guidelines](https://docs.prowler.com/developer-guide/check-metadata-guidelines) page.

Please feel free to add information you find valuable to the `Notes` field.

kagahd:
Hi @HugoPBrito 

> In the meantime, please adapt the metadata to the new standard, defined in our docs on the [metadata guidelines](https://docs.prowler.com/developer-guide/check-metadata-guidelines) page.

It's done. Let me know if anything needs further adjustment. Here's a summary of the changes:

- **CheckTitle**: Rewritten as a declarative, singular statement (removed "Ensure" verb)
- **CheckType**: Added `Effects/Data Exposure`
- **Description**: Rewritten to focus on the finding using markdown formatting, describing the four evaluated controls
- **Risk**: Now explains specific CIA triad impact (confidentiality, integrity, lateral movement, privilege escalation)
- **Recommendation**: Principle-based guidance (deny-by-default, least privilege) instead of imperative "Ensure…"
- **Remediation Code**: Added CLI, CloudFormation, Terraform, and manual console steps (were all empty before)
- **Categories**: Changed from `access-control` to `secrets` and `trust-boundaries`
- **Notes**: Added a description of the four layered controls and the intentional cross-account failure behavior
- Added `ResourceGroup`, `AdditionalURLs` fields to match the current upstream format

prowler/providers/aws/services/secretsmanager/secretsmanager_has_restrictive_resource_policy/secretsmanager_has_restrictive_resource_policy.py
```diff
@@ -0,0 +1,531 @@
...
+                arn_not_like_principals = []  # Store ARN patterns from ArnNotLike
+
+                # Check for an explicit Deny that applies to all Principals except those defined in the Condition
+                has_explicit_deny_for_all = False
+
+                # Track cross-account access detection
+                cross_account_principals = []
+
+                # Process all statements
+                for statement in statements:
```
HugoPBrito, line 109: This loop mixes cross-account Allow detection and deny validation. Because you break on the first valid Deny, any later Allow with a cross-account principal is never inspected. I reproduced a PASS with an ExternalRole Allow appended after the Deny blocks. This needs a second pass over statements or a separate scan for Allow statements so the result is order-independent.

prowler/providers/aws/services/secretsmanager/secretsmanager_has_restrictive_resource_policy/secretsmanager_has_restrictive_resource_policy.py
```diff
@@ -0,0 +1,531 @@
...
+                                )
+                                # STRICT VALIDATION: Ensure only "StringEquals" with "aws:SourceAccount" exists
+                                # No additional condition operators or keys allowed
+
+                                # Collect specific issues for this service
+                                issues = []
+                                has_wildcard_action = any(
+                                    "*" in action for action in actions
+                                )
+                                has_correct_condition = condition == {
```
HugoPBrito: This exact-equality check rejects service Allow statements that add extra restrictive conditions such as ArnLike on aws:SourceArn. If the requirement is must include aws:SourceAccount, I would expect additional restrictive conditions to be accepted rather than rejected.

prowler/providers/aws/services/secretsmanager/secretsmanager_has_restrictive_resource_policy/secretsmanager_has_restrictive_resource_policy.py
```diff
@@ -0,0 +1,531 @@
...
+                        )
+                        for statement in statements
+                    )
+                )
+
+                # Check for "NotActions" without wildcard * for not_denied_principals and not_denied_services
+                failed_principals = []
+                failed_services = []
+
+                # Validate that NotAction does not contain wildcards for specified principals
```
HugoPBrito, line 344: Can you confirm whether NotAction is intentionally optional? The tests explicitly PASS when the per-principal Deny/NotAction statements are removed, and the implementation only validates them if they already exist. If the contract is must have least-privilege per-principal Deny statements, this check currently doesn't enforce it.

prowler/providers/aws/services/secretsmanager/secretsmanager_has_restrictive_resource_policy/secretsmanager_has_restrictive_resource_policy.py
```diff
@@ -0,0 +1,531 @@
...
+class secretsmanager_has_restrictive_resource_policy(Check):
+    def execute(self):
+        findings = []
+        organizations_trusted_ids = secretsmanager_client.audit_config.get(
+            "organizations_trusted_ids", []
+        )
+        # Regular expression to match IAM roles or users without wildcard * in their name
+        arn_pattern = r"arn:aws:iam::\d{12}:(role|user)/([^*]+)$"
+        # Regular expression to match AWS service names
+        service_pattern = r"^[a-z0-9-]+\.amazonaws\.com$"
```
HugoPBrito: Can you confirm the intended service-principal format here? This regex only accepts service.amazonaws.com, but AWS has valid regionalized and multi-label service principals too. If this check is meant to support only simple principals like appflow.amazonaws.com, that constraint should be documented explicitly; otherwise the pattern is too narrow.

prowler/providers/aws/services/secretsmanager/secretsmanager_has_restrictive_resource_policy/secretsmanager_has_restrictive_resource_policy.py
kagahd: You're absolutely right — this is a bug. The current implementation processes Allow and Deny statements in a single pass, and the `break` on line 217/229 exits the loop as soon as a valid Deny is found. Any Allow statement with a cross-account principal that appears *after* that Deny is never evaluated, leading to a false PASS.

I'll refactor this to decouple the two concerns: first, collect all cross-account Allow principals across *all* statements, then separately validate the Deny statements. This will make the result completely order-independent, as it should be.

prowler/providers/aws/services/secretsmanager/secretsmanager_has_restrictive_resource_policy/secretsmanager_has_restrictive_resource_policy.py
kagahd: Good catch. The current check (`condition == {"StringEquals": {"aws:SourceAccount": ...}}`) rejects policies that add *additional* restrictive conditions like `ArnLike` on `aws:SourceArn`. That's overly strict — a policy that is *more* restrictive than the baseline should still pass.

I'll switch this to a subset/superset check: verify that the required `StringEquals` / `aws:SourceAccount` condition is *present*, without rejecting additional conditions that only tighten the policy further.

prowler/providers/aws/services/secretsmanager/secretsmanager_has_restrictive_resource_policy/secretsmanager_has_restrictive_resource_policy.py
kagahd: This is an intentional design decision: the per-principal Deny/NotAction statements are considered an *optional hardening layer* on top of the required baseline (which is the global Deny with `StringNotEquals` / `aws:PrincipalArn`). The rationale is that the global Deny already restricts access to only the listed principals — the per-principal NotAction blocks are an additional defense-in-depth measure that limits *what* each allowed principal can do, but they're not strictly required for the policy to be considered "restrictive."

I fully agree this needs to be documented explicitly though. I'll add a clear comment in the code and a note in the metadata explaining this design choice.

prowler/providers/aws/services/secretsmanager/secretsmanager_has_restrictive_resource_policy/secretsmanager_has_restrictive_resource_policy.py
kagahd: Fair point. The current regex only matches simple, non-regionalized principals like `appflow.amazonaws.com`. For now, this is intentional — the services we've encountered in practice with Secrets Manager resource policies all use simple principals.

However, I'll document this constraint explicitly and adjust the regex to also handle regionalized and multi-label principals (e.g., `logs.eu-central-1.amazonaws.com`) to be future-proof.

kagahd:
Hi @HugoPBrito,

Thanks a lot for the detailed and well-researched review. It's clear you spent real time understanding the logic, and I really appreciate that, especially given the complexity of this check. I'll push fixes for points 1 and 2 as priority, update the documentation for point 3, and widen the regex for point 4.

kagahd:
Hi @HugoPBrito,

All four points from your review have been addressed in commit f26cba1:

1. **Loop ordering bug (line 79)** - Decoupled Allow/Deny scanning into two separate passes. Cross-account Allow principals are now collected independently before Deny validation, making the result fully order-independent. Added a regression test that places a cross-account Allow after the Deny blocks.

2. **Condition exact-equality (line 354)** - Switched from exact-equality to a subset check. The condition now verifies that `StringEquals` with `aws:SourceAccount` is present, without rejecting additional restrictive conditions like `ArnLike`. Added a test with extra conditions that confirms PASS.

3. **NotAction optionality (line 319)** - Documented the design decision in a code comment and updated the metadata (Description, Recommendation, Notes) to clearly state that per-principal Deny/NotAction statements are an optional defense-in-depth layer.

4. **Service principal regex (line 18)** - Widened the regex to accept regionalized and multi-label principals (e.g. `logs.eu-central-1.amazonaws.com`). Added a test to verify. Updated the metadata to document support for both formats.

kagahd:
Hi @HugoPBrito,

Just a quick ping - all four points from your review have been addressed in the latest commits. Would you have a chance to take another look when you get a moment? Happy to make any further adjustments if needed.

Thanks!

HugoPBrito:
> Hi @HugoPBrito,
> 
> Just a quick ping - all four points from your review have been addressed in the latest commits. Would you have a chance to take another look when you get a moment? Happy to make any further adjustments if needed.
> 
> Thanks!

Thanks @kagahd ! I'll be taking a look today.

prowler/providers/aws/services/secretsmanager/secretsmanager_has_restrictive_resource_policy/secretsmanager_has_restrictive_resource_policy.py
```diff
@@ -0,0 +1,532 @@
...
+                else:
+                    final_role_arn = "None"
+
+            report.status_extended = (
+                f"SecretsManager secret '{secret.name}' does not have a resource-based policy "
+                f"or access to the policy is denied for the role '{final_role_arn}'"
+            )
+
+            if secret.policy:
+                statements = secret.policy.get("Statement", [])
```
HugoPBrito, line 69: `Statement` in IAM policies can be either a single object or a list. Here we iterate over `secret.policy.get("Statement", [])` assuming it is always a list, so a valid policy with a single statement would make us iterate over dict keys and then fail on `statement.get(...)`. Please normalize `Statement` to a list first, like we already do in other policy helpers.

prowler/providers/aws/services/secretsmanager/secretsmanager_has_restrictive_resource_policy/secretsmanager_has_restrictive_resource_policy.py
```diff
@@ -0,0 +1,532 @@
...
+
+                # Allow-Statement for not-denied services must not have any wildcards in "Action"
+                # and "SourceAccount" must be the audited account
+                for statement in statements:
+                    if statement.get("Effect") == "Allow":
+                        principals = self.extract_field(statement.get("Principal", {}))
+                        for service in principals:
+                            if service in not_denied_services:
+                                condition = statement.get("Condition", {})
+                                actions = self.extract_field(
```
HugoPBrito: This validation only checks that `Action` does not contain wildcards, but it does not require `Action` to exist at all. That means a service-principal `Allow` statement using `NotAction` plus `StringEquals/aws:SourceAccount` could still pass here, even though it would be broader than the contract described in the PR. I think we should explicitly require `Action` to be present and reject `NotAction` for these service `Allow` statements.

prowler/providers/aws/services/secretsmanager/secretsmanager_has_restrictive_resource_policy/secretsmanager_has_restrictive_resource_policy.py
```diff
@@ -0,0 +1,532 @@
...
+                                # Collect specific issues for this service
+                                issues = []
+                                has_wildcard_action = any(
+                                    "*" in action for action in actions
+                                )
+                                # Accept the condition if it contains at least
+                                # StringEquals with aws:SourceAccount matching the
+                                # audited account. Additional restrictive conditions
+                                # (e.g. ArnLike on aws:SourceArn) are acceptable.
+                                has_correct_condition = (
```
HugoPBrito: IAM condition keys are case-insensitive, but this check is doing an exact lookup for `aws:SourceAccount`. A policy using `AWS:SourceAccount` or `aws:sourceaccount` would be semantically valid and still fail here. We should normalize the condition keys before reading them, or reuse the shared IAM policy helper logic that already handles this correctly.

prowler/providers/aws/services/secretsmanager/secretsmanager_has_restrictive_resource_policy/secretsmanager_has_restrictive_resource_policy.py
```diff
@@ -0,0 +1,532 @@
...
+                                    "*" in action
+                                    for action in self.extract_field(
+                                        statement.get("NotAction", [])
+                                    )
+                                ):
+                                    failed_principals.append(principal)
+
+                # Allow-Statement for not-denied services must not have any wildcards in "Action"
+                # and "SourceAccount" must be the audited account
+                for statement in statements:
```
HugoPBrito, line 349: Non-blocking, but I think this service-principal `Allow` validation block should be extracted into a shared IAM-policy helper, or at least a dedicated private method. Right now this check is implementing its own condition-parsing semantics instead of leaning on the common policy utilities, and that is already causing drift in behavior. Pulling it out would make the check easier to reason about and reduce the chance of policy-validation bugs creeping back in.

HugoPBrito (commented):
Sorry for the amount of requested changes throughout this PR review. I know this has been a long and detailed iteration, and I really appreciate the effort you have put into it, @kagahd. Thanks for staying with it and addressing the previous rounds so carefully. I have a few final review comments below.

prowler/providers/aws/services/secretsmanager/secretsmanager_has_restrictive_resource_policy/secretsmanager_has_restrictive_resource_policy.py
kagahd: Thanks for catching this! You are right, I have added the defensive normalization now, consistent with the pattern used in other checks (e.g. `organizations_scp_check_deny_regions`). Please verify in 1e86400c3.

prowler/providers/aws/services/secretsmanager/secretsmanager_has_restrictive_resource_policy/secretsmanager_has_restrictive_resource_policy.py
kagahd: Good catch. I have extracted the service-principal Allow validation into a dedicated method `_validate_service_allow_statement` and it now explicitly rejects both `NotAction` and `NotResource` (since `NotResource` has the same scope-broadening effect). If neither is present, it also requires `Action` to be explicitly set. Please verify in 1e86400c3.

prowler/providers/aws/services/secretsmanager/secretsmanager_has_restrictive_resource_policy/secretsmanager_has_restrictive_resource_policy.py
kagahd: Valid point. I have added a `_normalize_condition_keys` method that lowercases all condition keys upfront, following the same approach as `is_condition_block_restrictive()` in `prowler/providers/aws/services/iam/lib/policy.py`. All internal comparisons now use lowercase keys. Please verify in 1e86400c3.

prowler/providers/aws/services/secretsmanager/secretsmanager_has_restrictive_resource_policy/secretsmanager_has_restrictive_resource_policy.py
kagahd: Agreed. I have extracted the service-principal Allow validation into `_validate_service_allow_statement` as a private method. Moving it into a shared IAM policy helper would be a reasonable next step, but for now this at least isolates the logic and makes it easier to reason about. Please verify in 1e86400c3.

kagahd:
@HugoPBrito, there is no need to apologize! Having a thorough reviewer on a check this complex is genuinely valuable. Your comments have consistently improved the robustness of the implementation (the case-insensitive condition keys and the NotAction gap were real bugs that would have slipped through). Appreciate the detailed feedback, it makes the end result better.

prowler/providers/aws/services/secretsmanager/secretsmanager_has_restrictive_resource_policy/secretsmanager_has_restrictive_resource_policy.py
```diff
@@ -0,0 +1,580 @@
...
+            actions = self.extract_field(statement.get("Action", []))
+            if any(isinstance(action, str) and "*" in action for action in actions):
+                issues.append("contains wildcard in Action field")
+
+        # Validate condition: require at least StringEquals with aws:SourceAccount
+        # Additional restrictive conditions (e.g. ArnLike on aws:SourceArn) are acceptable.
+        condition = statement.get("Condition", {})
+        has_correct_condition = (
+            "StringEquals" in condition
+            and condition.get("StringEquals", {}).get("aws:sourceaccount")
```
HugoPBrito: Non-blocking edge case: `_validate_service_allow_statement()` currently only accepts `aws:SourceAccount` as a scalar string. AWS also allows a single-value list in `StringEquals`, so a valid policy like `{"aws:SourceAccount": ["123456789012"]}` is currently marked as FAIL. I would not block the PR on this, but it would be good to support if you want to make the validation more AWS-compatible.

prowler/providers/aws/services/secretsmanager/secretsmanager_has_restrictive_resource_policy/secretsmanager_has_restrictive_resource_policy.py
```diff
@@ -0,0 +1,580 @@
...
+    # If the field is empty or invalid, return the default_value in the array.
+    def extract_field(self, field, default_value=None):
+        if isinstance(field, str):
+            return [field]
+        elif isinstance(field, list):
+            return field
+        elif isinstance(field, dict):
+            for key in ("AWS", "Service"):
+                if key in field:
+                    return [field[key]] if isinstance(field[key], str) else field[key]
```
HugoPBrito: One last blocking issue here: `extract_field()` only returns one branch from a dict `Principal`, so an `Allow` like `{"AWS": "...", "Service": "appflow.amazonaws.com"}` ignores the service principal during validation. That means the check can incorrectly PASS without enforcing `aws:SourceAccount` for the service. Please either flatten both `AWS` and `Service` values or explicitly reject mixed-principal `Allow` statements, and add a regression test for that case.

prowler/providers/aws/services/secretsmanager/secretsmanager_has_restrictive_resource_policy/secretsmanager_has_restrictive_resource_policy.py
kagahd: Thanks for catching this edge case! You're right, AWS allows condition values as either a scalar string or a single-value list, so `{"aws:SourceAccount": ["123456789012"]}` is perfectly valid. I've updated `_validate_service_allow_statement()` to normalize the value via `extract_field()` and use `in` instead of `==`, which handles both forms correctly. Added a regression test (`test_source_account_as_list_passes`) to cover this. Fixed in c98e488.

prowler/providers/aws/services/secretsmanager/secretsmanager_has_restrictive_resource_policy/secretsmanager_has_restrictive_resource_policy.py
kagahd: Great catch, thanks for the thorough review! This was indeed a real bug - `extract_field()` had an early `return` inside the loop over `("AWS", "Service")`, so it only ever returned the first matching branch. A mixed-principal Allow like `{"AWS": "...", "Service": "appflow.amazonaws.com"}` would silently skip the service principal, bypassing SourceAccount validation entirely.

I've fixed `extract_field()` to flatten both branches into a single list. This has no side effects on other callers since the `{"AWS": ..., "Service": ...}` dict structure only appears in Principal fields. Added a regression test (`test_mixed_principal_allow_must_validate_service`) that would have failed before the fix. Fixed in c98e488.

kagahd:
@HugoPBrito Both points addressed in c98e488. The mixed-principal bug in `extract_field()` was a real find - it would have silently skipped service-principal validation in production. Thanks again for the careful review, the check keeps getting more solid with each round of feedback!

HugoPBrito:
Thanks for sticking with this one. I went through the latest diff, the commit history, and the earlier review threads, not just the current snapshot.

The PR is in much better shape now. The issues raised in the previous rounds look addressed, and I also checked the thread status: all 12 review threads are resolved, so I do not see any old comments still pending.

I still have one blocker before merge:

- `prowler/providers/aws/services/secretsmanager/secretsmanager_has_restrictive_resource_policy/secretsmanager_has_restrictive_resource_policy.py`

  The check is still hardcoded to the commercial AWS partition. Right now it only accepts IAM ARNs like `arn:aws:iam::...`, service principals ending in `.amazonaws.com`, and it also rebuilds assumed-role ARNs back into `arn:aws:iam::...`.

  That is fine for the standard partition, but Prowler also supports `aws-cn` and `aws-us-gov`, and Secrets Manager is available there too. In the current state, valid principals like `arn:aws-cn:iam::...` or service principals ending in `.amazonaws.com.cn` would be treated as invalid, so this can produce false FAILs in supported partitions.

Because of that, I would block merge until the partition handling is made partition-aware.

kagahd:
@HugoPBrito Good catch, fully agree. All patterns were indeed hardcoded to the commercial partition.

Fixed in 3257ab6 - the check now reads `secretsmanager_client.audited_partition` and builds all regex patterns dynamically:

- **ARN patterns** (`arn_pattern`, `iam_arn_with_account_pattern`, `iam_root_arn_pattern`, `arn_wildcard_pattern`) use the actual partition instead of hardcoded `aws`
- **Service principal pattern** uses `.amazonaws.com.cn` for `aws-cn` and `.amazonaws.com` for all other partitions (including `aws-us-gov`, which shares the commercial DNS suffix)
- **Assumed-role ARN transformation** (STS -> IAM) also respects the partition

Added two regression tests:
- `test_china_partition_principals_and_services` - verifies that `arn:aws-cn:iam::...` ARNs and `.amazonaws.com.cn` service principals are accepted when the partition is `aws-cn`
- `test_china_partition_rejects_commercial_arns` - verifies that commercial `arn:aws:iam::...` ARNs are correctly rejected in a `aws-cn` context

**Note on ISO partitions:** The isolated partitions (`aws-iso`, `aws-iso-b`, `aws-iso-e`, `aws-iso-f`) use entirely different DNS suffixes (e.g. `.c2s.ic.gov`, `.sc2s.sgov.gov`). I chose not to implement these since Secrets Manager availability in those environments is unclear and Prowler is unlikely to be deployed there. If that changes, the fix would be a simple lookup table - the pattern infrastructure is already in place.

HugoPBrito:
Thank you very much for every one of the changes you've made @kagahd ! This is the way 🚀 

Could you please provide some evidences for the current state of the check?

Also, I'm thinking about adding it to `threat-detection` category since its execution time is huge. Taking into account that it's a very specific and deep check, I find it the right decision. What do you think?

jfagoagas (dismissed):

kagahd:
Thanks for the green light @HugoPBrito! Happy to provide evidences. To make sure I share the right thing: would sanitized output from a test run against a sandbox account (one secret with a compliant policy, one without) work for you? I'd rather not post production findings for obvious reasons. Alternatively, the 80 unit tests cover all validation paths if that's sufficient.

Regarding the `threat-detection` category: I'm not sure it fits here since this is a static configuration compliance check, not an active threat detection mechanism. Adding it to that category could set the wrong expectation about what the check does. If execution time is a concern, I'm happy to look into performance optimizations or discuss whether a dedicated scan profile would be a better fit. What are your thoughts?

HugoPBrito:
> Thanks for the green light @HugoPBrito! Happy to provide evidences. To make sure I share the right thing: would sanitized output from a test run against a sandbox account (one secret with a compliant policy, one without) work for you? I'd rather not post production findings for obvious reasons. Alternatively, the 80 unit tests cover all validation paths if that's sufficient.
> 
> Regarding the `threat-detection` category: I'm not sure it fits here since this is a static configuration compliance check, not an active threat detection mechanism. Adding it to that category could set the wrong expectation about what the check does. If execution time is a concern, I'm happy to look into performance optimizations or discuss whether a dedicated scan profile would be a better fit. What are your thoughts?

Sure! That'll be fine.

Regarding performance, feel free to give it a try. If not, I can also take a look into it. It concerns me because it could take a really long times in big accounts and I'm worried about Prowler Cloud customers.

kagahd:
@HugoPBrito Thanks for the quick reply! Here are the sandbox evidences and performance benchmark results.

## Performance Benchmark

We ran this check across **301 active AWS accounts** (one GitHub Actions job per account) and compared it against the native `secretsmanager_secret_unused` check on the same set of accounts:

| Metric | `secretsmanager_has_restrictive_resource_policy` | `secretsmanager_secret_unused` (baseline) |
|---|---|---|
| Mean duration | 225s | 226s |
| Median duration | 222s | 225s |
| p95 duration | 308s | 297s |
| Difference | **-0.5%** | baseline |

The custom check adds no measurable overhead. Runtime is dominated by Prowler's provider initialization and API calls, not by the check logic itself. The check only iterates over already-fetched secrets and their policies, so it should scale the same way regardless of account size.

## Sandbox Evidence (PASS / FAIL)

Tested on a sandbox account with two purpose-built secrets (account IDs and ARNs anonymized). Prowler was configured with `organizations_trusted_ids` to validate the organization boundary check. The two mandatory requirements (deny unauthorized principals + deny outside organization) are both present. The remaining validations (per-principal NotAction, service-principal Allow, cross-account detection) only apply when those policy elements are present. These paths are thoroughly covered by the 80 unit tests.

**PASS** - Secret with a restrictive resource policy (Deny unauthorized principals + Deny outside organization):
```json
{
  "status_code": "PASS",
  "status_detail": "SecretsManager secret 'prowler-test-PASS' has a sufficiently restrictive resource-based policy.",
  "resource_policy": {
    "Version": "2012-10-17",
    "Statement": [
      {
        "Sid": "DenyUnauthorizedPrincipals",
        "Effect": "Deny",
        "Principal": "*",
        "Action": "*",
        "Resource": "*",
        "Condition": {
          "StringNotEquals": {
            "aws:PrincipalArn": [
              "arn:aws:iam::111111111111:role/AuthorizedAdminRole",
              "arn:aws:iam::111111111111:role/AuditRole"
            ]
          }
        }
      },
      {
        "Sid": "DenyOutsideOrganization",
        "Effect": "Deny",
        "Principal": "*",
        "Action": "secretsmanager:*",
        "Resource": "*",
        "Condition": {
          "StringNotEquals": {
            "aws:PrincipalOrgID": "o-exampleorgid"
          }
        }
      }
    ]
  }
}
```

**FAIL** - Secret with no resource policy:
```json
{
  "status_code": "FAIL",
  "status_detail": "SecretsManager secret 'prowler-test-FAIL' does not have a resource-based policy or access to the policy is denied for the role 'arn:aws:iam::111111111111:role/AuthorizedAdminRole'",
  "resource_policy": null
}
```

Let me know if you need anything else or if the evidences look good to you. Looking forward to getting this merged! 🚀

HugoPBrito:
Thank you very much @kagahd! It's an absolute pleasure to be your reviewer.

prowler/CHANGELOG.md
HugoPBrito, line 1: Please update this.

HugoPBrito (changes requested):
Please change that and it'll be at v5.25! If not done by the needed time I'll do it myself so it's included in the release.

kagahd:
@HugoPBrito Done! Moved the changelog entry to the `[5.25.0]` section and merged upstream/master to resolve the conflicts. Let me know if there's anything else needed on my side!

HugoPBrito (approved):

HugoPBrito:
Thanks for your awesome work @kagahd!

PR #7053, opened 2025-02-26
refactor(stats): Use Finding instead of Check_Report

jfagoagas:
### Description

Refactor `extract_findings_statistics` to use `Finding` instead of `Check_Report_XXX`.

### Checklist

#### API

### License

### DIFF
--- a/prowler/__main__.py
+++ b/prowler/__main__.py
@@ -305,6 +305,18 @@ def prowler():
             print(f"{Style.BRIGHT}{Fore.GREEN}\nNo findings to fix!{Style.RESET_ALL}\n")
         sys.exit()
 
+    # Outputs
+    # TODO: this part is needed since the checks generates a Check_Report_XXX and the output uses Finding
+    # This will be refactored for the outputs generate directly the Finding
+    finding_outputs = []
+    for finding in findings:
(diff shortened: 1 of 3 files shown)


prowler/__main__.py
```diff
@@ -305,6 +305,18 @@ def prowler():
...
+    # TODO: this part is needed since the checks generates a Check_Report_XXX and the output uses Finding
+    # This will be refactored for the outputs generate directly the Finding
+    finding_outputs = []
+    for finding in findings:
+        try:
+            finding_outputs.append(
+                Finding.generate_output(global_provider, finding, output_options)
+            )
+        except Exception:
+            continue
```
MrCloudSec, line 318: Why do we continue and ignore the exception?

jfagoagas: We will need to review why it's there, I can't remember now.

pedrooot (approved):
Thanks for fixing this bug! 🔝

pedrooot (dismissed):
Thanks for fixing this bug! 🔝

MrCloudSec (approved):
👏🏼 👏🏼 👏🏼 👏🏼

PR #7057, opened 2025-02-27
feat(m365): add sharepoint service with 4 checks

danibarranqueroo:
### Context

This PR introduces a new service for Microsoft365: SharePoint, which includes four new checks along with comprehensive tests for each and also for the service. All changes have been fully adapted to the latest refactor, ensuring consistency across the codebase.

The four new checks cover settings for SharePoint. They are:

- `sharepoint_external_sharing_managed`: Validates that external sharing is managed in accordance with organizational policies (allowing or blocking access to certain domains).
- `sharepoint_external_sharing_restricted`: Ensures that external sharing is appropriately restricted to safeguard data.
- `sharepoint_guest_sharing_restricted`: Verifies that guest sharing is limited to prevent unauthorized access.
- `sharepoint_modern_authentication_required`: Confirms that modern authentication is enabled for enhanced security, this means that access for applications that don't use modern authentication is blocked.

### Description

Added new service with new checks and all the unit tests needed.

### Checklist

#### API

### License


MrCloudSec (changes requested):
Good job! Please, make sure you only pass a single object to CheckReportMicrosoft365

MrCloudSec (dismissed):
Good job 👏🏼  👏🏼  👏🏼  👏🏼

prowler/providers/microsoft365/services/sharepoint/sharepoint_service.py
```diff
@@ -0,0 +1,62 @@
...
+            settings = SharePointSettings(
+                sharingCapability=(
+                    str(global_settings.sharing_capability).split(".")[-1]
+                    if global_settings.sharing_capability
+                    else None
+                ),
+                sharingAllowedDomainList=global_settings.sharing_allowed_domain_list,
+                sharingBlockedDomainList=global_settings.sharing_blocked_domain_list,
+                sharingDomainRestrictionMode=global_settings.sharing_domain_restriction_mode,
+                modernAuthentication=global_settings.is_legacy_auth_protocols_enabled,
```
andoniaf: If legacy auth protocols are enabled, modernAuth shouldn't be false instead?

prowler/providers/microsoft365/services/sharepoint/sharepoint_modern_authentication_required/sharepoint_modern_authentication_required.py
```diff
@@ -0,0 +1,49 @@
...
+            report = CheckReportMicrosoft365(
+                self.metadata(),
+                resource=settings if settings else {},
+                resource_name="SharePoint Settings",
+                resource_id=sharepoint_client.tenant_domain,
+            )
+            report.status = "PASS"
+            report.status_extended = "Microsoft 365 SharePoint does not allow access to apps that don't use modern authentication."
+
+            if settings.modernAuthentication:
```
andoniaf: If [this](https://github.com/prowler-cloud/prowler/pull/7057/files#r1973386537) comment is correct, this should be `if not settings.modernAuthentication:`

prowler/providers/microsoft365/services/sharepoint/sharepoint_service.py
MrCloudSec: Good catch! I think the naming is confusing.

MrCloudSec: It should be `basicAuthentication`

andoniaf: I'm not 100% sure legacy auth includes only basicAuth, maybe we could just call it `LegacyAuth` or `LegacyAuthentication`. To match [their docs](https://learn.microsoft.com/en-us/graph/api/sharepointsettings-update?view=graph-rest-1.0&tabs=http) description.

danibarranqueroo: The naming is a bit confusing, that's true.

When the parameter from the API `is_legacy_auth_protocols_enabled` returns True means that SharePoint allows access to applications that do NOT use Modern Authentication (this means that a True should be a FAIL because applications using Basic Auth are allowed).

We can just call it LegacyAuth as @andoniaf suggested and the logic of the check will be the same as now (also I suggest to add what I said above in a comment because it's very easy to confuse it)

prowler/providers/microsoft365/services/sharepoint/sharepoint_modern_authentication_required/sharepoint_modern_authentication_required.py
danibarranqueroo: Answered in the other conversation 😄

MrCloudSec (approved):

PR #7059, opened 2025-02-27
fix(ecs): ensure unique finding id in ECS checks

### LINKED ISSUE #7050: Finding UID for check ecs_service_no_assign_public_ip is not unique
### Steps to Reproduce

1. In an AWS region create two different clusters
2. In each one create a service that assigns a public IP. The service name should be the same in both clusters
3. Scan the account using prowler and export results to CSV

You'll get two findings with the same FINDING_UID

### Expected behavior

It seems that the pattern used is prowler-aws-ecs_service_no_assign_public_ip-ACCOUNT_NUMBER-REGION-SERVICE_NAME

However, if you have multiple clusters in the same region you can have services with duplicate names.

The expected behaviour is that FINDING_UID is unique per account and resource.

Proposal: use the pattern prowler-aws-ecs_service_no_assign_public_ip-ACCOUNT_NUMBER-REGION-CLUSTER_NAME#SERVICE_NAME

The separator should be anything except for valid cluster/service name allowed characters: uppercase and lowercase, numbers, underscores, and hyphens.

Not required

### Environment Resource

1

### OS used

1

### Prowler version

Prowler 5.3.0

### Pip version

pip 25.0.1

### Context

_No response_

MrCloudSec:
### Context

Fix #7050

### Description

Ensure Prowler creates unique Finding UIDs in ECS by using also the cluster name inside the service ID.
### Checklist

#### API

### License

### DIFF
--- a/prowler/providers/aws/services/ecs/ecs_service.py
+++ b/prowler/providers/aws/services/ecs/ecs_service.py
@@ -115,6 +115,7 @@ class ECS(AWSService):
                     service_arn = service_desc["serviceArn"]
                     service_obj = Service(
                         name=sub(":.*", "", service_arn.split("/")[-1]),
+                        id=f"{sub(':.*', '', service_arn.split('/')[-2])}/{sub(':.*', '', service_arn.split('/')[-1])}",
                         arn=service_arn,
                         region=cluster.region,
                         assign_public_ip=(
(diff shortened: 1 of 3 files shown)


andoniaf (approved):
🙌

PR #7061, opened 2025-02-27
feat(entra): add new check `entra_password_hash_sync_enabled`

danibarranqueroo:
### Context

This PR introduces a new check for M365 service: Entra. This new check verifies if Password hash synchronization is enabled. This is one of the sign-in methods used to accomplish hybrid identity synchronization. 

The only way to have this enabled is using Microsoft Entra Connect, an application from Microsoft that synchronizes a hash of a user's password from an on-premises Active Directory instance to a cloud-based Entra ID instance in order to have hybrid environment safer.

### Description

Added new check `entra_password_hash_sync_enabled` with its unit tests and modifed the service to add `_get_organization` api call

### Checklist

#### API

### License


prowler/providers/microsoft365/services/entra/entra_password_hash_sync_enabled/entra_password_hash_sync_enabled.py
```diff
@@ -0,0 +1,45 @@
...
+class entra_password_hash_sync_enabled(Check):
+    """
+    Check if password hash synchronization is enabled for hybrid Microsoft Entra deployments.
+
+    This check verifies that password hash synchronization is enabled in hybrid Microsoft Entra deployments.
+    Enabling password hash sync ensures that on-premises passwords are synchronized to Microsoft Entra,
+    facilitating seamless authentication and enhancing leaked credential protection. Without password hash
+    synchronization, users might have to manage multiple passwords and detection of leaked credentials would be compromised.
+
+    Note: This control applies only to hybrid deployments using Microsoft Entra Connect sync and does not apply to federated domains.
```
andoniaf, line 16: Should we check then if the domain is federated to avoid false positives?

andoniaf: But you can have multiple domains... 🤔 Not sure how this setting would behave in that case.

MrCloudSec:
@danibarranqueroo what happens if the organization is not hybrid? The check has to be ignored, right?

danibarranqueroo:
> @danibarranqueroo what happens if the organization is not hybrid? The check has to be ignored, right?

Yes @MrCloudSec, I found that problem too, but I wasn't able to find any endpoint to know if an organization is hybrid. If it's not clear we can modify the status extended to notify the user that this check only applies to hybrid deployments.

MrCloudSec (approved):
LGTM

PR #7078, opened 2025-03-02
chore(deployment): New Helm chart

### LINKED ISSUE #7016: helm refactoring to scale and ease of use
### New feature motivation

The current helm chart is relativly hard to use and misses a bunch of features.

- The current containers don't scale independently, if I want a new worker I would have to add it to the same deployment as the API.
- There is no difference between configuration and secrets, which makes it hard to read.
- You can't differentiate between celery worker and API. My understanding of prowler is that it supports workload identity, this should make it possible to only have the celery SA to actually have access to Kubernetes/Cloud APIs. I haven't looked if the API support workload identity to connect to DB. But it's something that we should solve in the long run.
- It's hard to manage multiple helm charts, sure you can do sub charts, but why would you in this case?
- There is no release for the helm chart that follows the container tag, which makes it hard to get new
features.
- The helm chart don't follow security best practices, which is kind of boring for a security related product.
- No documentation at all

### Solution Proposed

- Remove the helm chart and start from scratch, this will of course not be backwards compatible.
- Have the API, UI, celery worker, celery broker in the same helm chart in different deployment with different SA and RBAC.
- Make it possible to autoScale the celery workers, for example based on CPU
- Don't have valkey or postgres be a part of the helm chart (not that the current one does ether)
- Document the helm chart
  - Come with suggestions on how to deploy postgrs and valkey
- Create an automated release flow, that releases when the container is built and store the resulting artifact in  github OCI.
- Potentially create a DJANGO_SECRETS/DJANGO_TOKEN through a Kubernetes job at initial creation. Here I don't know enough about DJANGO on what is needed, but I can't find any documentation about it when it comes to make prowler production ready. Would love some help from the maintainers, all I see that it's hard coded in a .env files, but I would assume you should generate your own.

N/A

### Additional context

_No response_

nissessenap:
### Context

Fix: https://github.com/prowler-cloud/prowler/issues/7016

### Description

Creates a new helm chart, that is possible to build on and scale separate deployments individually.
Breaking out secrets from values files and providing documentation around helm.

### Checklist

#### API

### License

### DIFF
--- a/contrib/k8s/prowler/.helmignore
+++ b/contrib/k8s/prowler/.helmignore
@@ -1,0 +1,23 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
+.DS_Store
+# Common VCS dirs
+.git/
+.gitignore
+.bzr/
+.bzrignore
+.hg/
+.hgignore
+.svn/
+# Common backup files
+*.swp
+*.bak
+*.tmp
+*.orig
+*~
+# Various IDEs
+.project
+.idea/
+*.tmproj
+.vscode/
--- a/contrib/k8s/prowler/Chart.yaml
+++ b/contrib/k8s/prowler/Chart.yaml
@@ -1,0 +1,24 @@
+apiVersion: v2
+name: prowler
+description: A Helm chart for Kubernetes
+
+# A chart can be either an 'application' or a 'library' chart.
+#
+# Application charts are a collection of templates that can be packaged into versioned archives
+# to be deployed.
+#
+# Library charts provide useful utilities or functions for the chart developer. They're included as
+# a dependency of application charts to inject those utilities and functions into the rendering
+# pipeline. Library charts do not define any templates and therefore cannot be deployed.
+type: application
+
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.1.0
+
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
+# It is recommended to use it with quotes.
+appVersion: "5.3.0"
--- a/contrib/k8s/prowler/README.md
+++ b/contrib/k8s/prowler/README.md
@@ -1,0 +1,428 @@
+---
+title: "Helm installation"
+linkTitle: "Helm installation"
+---
+
+# prowler
+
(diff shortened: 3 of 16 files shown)


nissessenap:
This is an initial draft, but I don't think I will have time to work much more on this in the near feature, but I wanted to upload something. If someone else got time, I would love to see improvements to it.

In the PR there is a TODO that cover most things that I personally think should be solved before merge.
But I don't know if some of the features are supported by the application itself.

## Questions

- Does the API/cellery instance got any readiness probes? I can't find any in docker-compose at least.
- What config is needed in the UI frontend compared to the backend? In docker-compose everything is just mounted. Looking at the old helm chart I think I only need to add DJANGO_AUTH_SECRET. But I haven't had time to add the UI to test it out.
- Does the API actually support autoscaling? In the docker entrypoint a DB migration always happens. How good is the migration framework that you are using when it comes to looking the database? What would happen if two new API instances starts relatively at the same time? Will the first on update some SHA to tell the other replica that it already has started the DB migration?

## Issue

- The container is massive, 1.92Gb. Could we go over to some multi-stage container build? Does the application need any CLIs of any kind? Or is it just priority/time?
- What I can see it's not possible to any env variable to configure the config file location: https://github.com/prowler-cloud/prowler/blob/8a144a4046d27d0dfb406638d7220f681cfde73f/prowler/lib/cli/parser.py#L344-L353
Would be nice to add this feature, that way we can remove horrible hard-coded paths that depends on the dockerfile.

jfagoagas:
Hello @nissessenap, what are your plans on this? Do you think you'll find time to continue working on this?

Thanks!

nissessenap:
Hi @jfagoagas , probably not (I decided against prowler at work). But I might use more of my own time if I get help with the questions that I state in my comment above.

But if I remember correctly I don't do anything magical in the chart (I normally don't) so it shouldn't be hard for someone else to pick this up.

aaronpejakovic-elmo:
I might potentially take a look at this if i get some time. We have recently just implemented prowler within our company and had to make some helm changes to deploy and was also faced with the same questions around scaling etc.

andoniaf:
> I might potentially take a look at this if i get some time. We have recently just implemented prowler within our company and had to make some helm changes to deploy and was also faced with the same questions around scaling etc.

That would be great @ap-elmo ! Thanks for considering it. Let me check with the team to see if we can help answer those questions.

Utwo:
I can help with this as well. At my company, we were using this chart as an inspiration https://promptlylabs.github.io/prowler-helm-chart/, and we paired it with Keda autoscaler for the prowler worker.

nissessenap:
@Utwo feel free to create a new PR and take a stab at it.

danibarranqueroo:
Closing this PR as the helm chart was done here https://github.com/prowler-cloud/prowler/pull/9835

PR #7167, opened 2025-03-10
feat(integrations): Added new endpoints to allow configuring integrations

vicferpoy:
### Description

This PR includes the new model for `integrations`, in addition to serializers and endpoints for them. Thanks to this PR, new API integrations will be possible through easier development. This means that this PR will serve as a baseline, but does not necessarily include any particular integration.

> A bug with Django enum types when reverting migrations was also fixed in this PR.

image

image

**NOTE:** Relationships between integrations and providers are defined through a many-to-many table, but relationships endpoints will not be developed for this. The following endpoints allow defining relationships between them:

- `POST /integrations`
- `PATCH /integrations/<id>`

All relationships used in these endpoints will overwrite the former ones for the given integration object.

### Checklist

#### API

### License


jfagoagas (dismissed):
Reviewed IRL, looks fantastic 👏

api/CHANGELOG.md
```diff
@@ -26,6 +30,6 @@ All notable changes to the **Prowler API** are documented in this file.
 - Daily scheduled scan instances are now created beforehand with `SCHEDULED` state [(#6700)](https://github.com/prowler-cloud/prowler/pull/6700).
 - Findings endpoints now require at least one date filter [(#6800)](https://github.com/prowler-cloud/prowler/pull/6800).
 - Findings metadata endpoint received a performance improvement [(#6863)](https://github.com/prowler-cloud/prowler/pull/6863).
-- Increase the allowed length of the provider UID for Kubernetes providers [(#6869)](https://github.com/prowler-cloud/prowler/pull/6869).
+- Increased the allowed length of the provider UID for Kubernetes providers [(#6869)](https://github.com/prowler-cloud/prowler/pull/6869).
```
AdriiiPRodri, line 33: Normally we use the present tense and it is the most established when making commits and writing Changelogs, why this change?

api/src/backend/api/rls.py
```diff
@@ -59,11 +58,11 @@ class RowLevelSecurityConstraint(models.BaseConstraint):
     drop_sql_query = """
         ALTER TABLE %(table_name)s NO FORCE ROW LEVEL SECURITY;
         ALTER TABLE %(table_name)s DISABLE ROW LEVEL SECURITY;
-        REVOKE ALL ON TABLE %(table_name) TO %(db_user)s;
+        REVOKE ALL ON TABLE %(table_name)s FROM %(db_user)s;
```
AdriiiPRodri, line 61: This change looks so important, can you give me more information? Was this working before this change?

AdriiiPRodri (changes requested):

api/src/backend/api/rls.py
vicferpoy: It wasn't working, it's part of the fix for the revert migrations. We haven't used this until now, that I split the migrations and tested if we could revert them.

api/CHANGELOG.md
vicferpoy: Because I created the CHANGELOG and used past tense for every other line, but I didn't add that one.

Present tense is widely used for commits, but CHANGELOG entries affect the past. That was my reasoning and what I've read, but there's certainly some debate about it.

It's a good conversation to have outside this PR. I would leave this one as it is, but feel free to open a thread with the team to establish the way to do it in the future and we can open a new PR to modify it if we decide so. Thank you for the input.

AdriiiPRodri: There is some discussion on this, yes, at some point it will have to be decided and can be discussed with the team, some entries are in the present tense.

api/src/backend/api/rls.py
AdriiiPRodri: Understood, thank you

vicferpoy:
After discussing it with the team, we have decided this is the place for the new integrations RBAC permissions. Switching back to draft until it's done.

jfagoagas (approved):

PR #7112, opened 2025-03-04
chore(poetry): Upgrade to v2

jfagoagas:
### Description

Upgrade `poetry` to be compatible with v2, notable changes:
- `black` formatted tons of `tests` in Prowler and several files in the API.
- `Dockerfile` was changed to use `poetry` as the package manager.
- Update `pyproject.toml` to Poetry v2 spec.

### Checklist

- Are there new checks included in this PR? Yes / No
    - If so, do we need to update permissions for the provider? Please review this carefully.
- [ ] Review if the code is being covered by tests.
- [ ] Review if code is being documented following this specification https://github.com/google/styleguide/blob/gh-pages/pyguide.md#38-comments-and-docstrings
- [ ] Review if backport is needed.
- [ ] Review if is needed to change the [Readme.md](https://github.com/prowler-cloud/prowler/blob/master/README.md)

#### API
- [ ] Verify if API specs need to be regenerated.
- [ ] Check if version updates are required (e.g., specs, Poetry, etc.).
- [ ] Ensure new entries are added to [CHANGELOG.md](https://github.com/prowler-cloud/prowler/blob/master/api/CHANGELOG.md), if applicable.

### License

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #7172, opened 2025-03-10
fix(exports): change the way to remove the local export files after s3 upload

AdriiiPRodri:
### Context

Currently we are getting `FileNotFoundError` type errors, this is due to the fact that several processes are used in the exports worker. Because of this when performing the deletion there was a race condition between scans.

### Description

The workaround has been to delete even the scan folder, so that only the files created within the scan itself and the scan itself will be deleted.

### Checklist

#### API

### License


jfagoagas (dismissed):
Thanks for this fix 👏

api/CHANGELOG.md
```diff
@@ -8,6 +8,13 @@ All notable changes to the **Prowler API** are documented in this file.
 
 ---
 
+## [v1.5.1] (Prowler v5.4.1)
```
vicferpoy, line 15: I don't see any API version updates in the rest of the code.

api/src/backend/tasks/tasks.py
```diff
@@ -263,17 +264,24 @@ def generate_outputs(scan_id: str, provider_id: str, tenant_id: str):
...
-        rmtree(DJANGO_TMP_OUTPUT_DIRECTORY, ignore_errors=True)
     else:
         uploaded = False
 
     # Update the scan instance with the output path
     Scan.all_objects.filter(id=scan_id).update(output_location=output_directory)
 
-    logger.info(f"Scan output files generated, output location: {output_directory}")
+    if not exception_raised:
+        logger.info(f"Scan output files generated, output location: {output_directory}")
```
vicferpoy: Is this log correct? `exception_raised` is flagged as `True` when the local files could not be removed, but the reports are already generated, isn't it?

vicferpoy (changes requested):

api/src/backend/tasks/tasks.py
AdriiiPRodri: Right, the files remain, the idea was to emphasize that there was a problem but it is true that the files are created

AdriiiPRodri:
The last commit uses poetry 2.1.1 since that is what we are going to use from now on, as I don't know if this PR will be merge before or after https://github.com/prowler-cloud/prowler/pull/7112, I add the API version changes and regenerate the `.lock`

vicferpoy (changes requested):
Thank you for the changes. We are still missing the version in `views.py`, but after that change everything should be ready for merge.

api/CHANGELOG.md
jfagoagas: Is this right? Prowler v5.4.1 is not released _yet_

jfagoagas (dismissed):
Should we merge Poetry first?

AdriiiPRodri:
> Should we merge Poetry first?

Yes

jfagoagas (approved):

vicferpoy (approved):

PR #7183, opened 2025-03-11
fix: add a handled response in case local files are missing

AdriiiPRodri:
### Description

In this PR we have added a new response in case an attempt is made to retrieve the local export files and they do not exist. In this case the API will return the 404 code.

### Checklist

#### API

### License


api/src/backend/api/v1/views.py
```diff
@@ -1280,7 +1280,13 @@ def report(self, request, pk=None):
             filename = os.path.basename(output_location.split("/")[-1])
         else:
             zip_files = glob.glob(output_location)
-            file_path = zip_files[0]
+            try:
+                file_path = zip_files[0]
+            except IndexError:
+                return Response(
```
jfagoagas, line 1298: You can also log this in Sentry with
```python
import sentry_sdk

sentry_sdk.capture_exception(exception)
```

AdriiiPRodri: Thanks! Added

api/CHANGELOG.md
```diff
@@ -8,6 +8,13 @@ All notable changes to the **Prowler API** are documented in this file.
 
 ---
 
+## [v1.5.1] (Prowler v5.4.1)
```
vicferpoy, line 15: We need to update the API version. Do it here or in the other PR you have opened.

vicferpoy (changes requested):

api/CHANGELOG.md
AdriiiPRodri: In the other PR

jfagoagas (dismissed):

vicferpoy (dismissed):

vicferpoy (approved):

PR #7201, opened 2025-03-12
fix(route53): solve false positive in `route53_public_hosted_zones_cloudwatch_logging_enabled`

### LINKED ISSUE #7173: route53_public_hosted_zones_cloudwatch_logging_enabled passes for every zone if there is one with enabled logging
### Steps to Reproduce

1. Create two hosted zones in Route53 (without query logging set up)
2. Run `prowler aws --checks route53_public_hosted_zones_cloudwatch_logging_enabled`
3. `route53_public_hosted_zones_cloudwatch_logging_enabled` fails for both zones
4. Set up query logging for only one of the zones
5. Run `prowler aws --checks route53_public_hosted_zones_cloudwatch_logging_enabled`

### Expected behavior

`route53_public_hosted_zones_cloudwatch_logging_enabled` should pass for the zone with the logging configuration and fail for the zone without the logging configuration

`route53_public_hosted_zones_cloudwatch_logging_enabled` passes for both zones

From brew (brew install prowler)

### Environment Resource

Workstation

### OS used

MacOS

### Prowler version

5.3.0

### Pip version

25.0

### Context

It seems the issue hides in `_list_query_logging_configs` in `route53_service.py`: https://github.com/prowler-cloud/prowler/blob/b7bce6008f81071310094aedcf23e7386002ccae/prowler/providers/aws/services/route53/route53_service.py#L83-L95

The code reads like this:
```
for every hosted zone
  for every query logging config
    use the configuration as the query logging config for the hosted zone
```
I think it needs to be something like this:

```
for every hosted zone
  for every query logging config
    **if the HostedZoneId of the config matches the hosted zone**
      use the configuration as the query logging config for the hosted zone 
```

danibarranqueroo:
### Context

This PR fixes a problem in Route53. When you have more than one zone and one of them have logging enabled, we return all of the zones a PASS instead of having FAIL or PASS depending on the state for each zone. 

The problem was that we were assigning the last Logging Config to every Hosted Zone, so if there was just one Logging Config every Hosted Zone will return a PASS, so the solution was to add a condition and only assign the Logging Config to the Hosted Zone if they are matching IDs.

Fix #7173 

### Description

Modified Route53 service and test it on infra.

### Checklist

#### API

### License

### DIFF
--- a/prowler/providers/aws/services/route53/route53_service.py
+++ b/prowler/providers/aws/services/route53/route53_service.py
@@ -86,13 +86,14 @@ class Route53(AWSService):
                 )
                 for page in list_query_logging_configs_paginator.paginate():
                     for logging_config in page["QueryLoggingConfigs"]:
-                        self.hosted_zones[hosted_zone.id].logging_config = (
-                            LoggingConfig(
-                                cloudwatch_log_group_arn=logging_config[
-                                    "CloudWatchLogsLogGroupArn"
-                                ]
+                        if logging_config["HostedZoneId"] == hosted_zone.id:
+                            self.hosted_zones[hosted_zone.id].logging_config = (
+                                LoggingConfig(
+                                    cloudwatch_log_group_arn=logging_config[
+                                        "CloudWatchLogsLogGroupArn"
+                                    ]
+                                )
                             )
-                        )
 
         except Exception as error:
             logger.error(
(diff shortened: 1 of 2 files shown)


prowler/providers/aws/services/route53/route53_service.py
MrCloudSec, line 1: Please, add a test to cover this case.

andoniaf (approved):

PR #7213, opened 2025-03-12
revert: Poetry upgrade to v2

jfagoagas:
### Description

Revert `poetry` upgrade to `v2` since the project's structure is not valid.


jfagoagas:
Fixed in https://github.com/prowler-cloud/prowler/pull/7217

PR #7221, opened 2025-03-13
fix(vm): handle NoneType accessing security_profile

andoniaf:
### Description
Handle NoneType accessing security_profile attributes as `vm.security_profile` can be `None`.

https://learn.microsoft.com/en-us/python/api/azure-mgmt-computefleet/azure.mgmt.computefleet.models.basevirtualmachineprofile?view=azure-python#azure-mgmt-computefleet-models-basevirtualmachineprofile-security-profile

### Checklist

#### API

### License

### DIFF
--- a/prowler/providers/azure/services/vm/vm_service.py
+++ b/prowler/providers/azure/services/vm/vm_service.py
@@ -50,7 +50,7 @@ class VirtualMachines(AzureService):
                                     else None
                                 ),
                                 location=vm.location,
-                                security_profile=vm.get("security_profile", None),
+                                security_profile=getattr(vm, "security_profile", None),
                                 extensions=[
                                     VirtualMachineExtension(id=extension.id)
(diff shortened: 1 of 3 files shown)


prowler/providers/azure/services/vm/vm_service.py
```diff
@@ -50,7 +50,7 @@ def _get_virtual_machines(self):
                                     else None
                                 ),
                                 location=vm.location,
-                                security_profile=vm.security_profile,
+                                security_profile=vm.get("security_profile", None),
```
MrCloudSec: Is not `.get` only for dictionaries? I think you should use `getattr`.

andoniaf: You're right, fixed.

MrCloudSec (approved):

PR #7230, opened 2025-03-13
fix(entra): check name and logic of `entra_admin_users_have_mfa_enabled`

HugoPBrito:
### Context

A logic error was found in `entra_admin_mfa_enabled_for_administrative_roles` check of `Entra service`. If a policy is non-compliant, we should continue iterating instead of stopping execution.

As it stands, the execution stops if it encounters a policy that is `Enabled for reporting`, but there may be another policy that is fully `Enabled` that we never reach because the previous one was selected for the report.

### Description

The name and the logic have been changed and a test has been added to validate the new behavior.

### Checklist

#### API

### License


pedrooot (dismissed):
Great work! Thanks

prowler/providers/microsoft365/services/entra/entra_capolicy_administrative_roles_have_mfa_enabled/entra_capolicy_administrative_roles_have_mfa_enabled.py
```diff
@@ -68,13 +68,13 @@ def execute(self) -> List[CheckReportMicrosoft365]:
...
-                report.status_extended = f"Conditional Access Policy '{policy.display_name}' enforces MFA for administrative roles."
-
                 if policy.state == ConditionalAccessPolicyState.ENABLED_FOR_REPORTING:
                     report.status = "FAIL"
                     report.status_extended = f"Conditional Access Policy '{policy.display_name}' only reports MFA for administrative roles but does not enforce it."
-                break
+                else:
+                    report.status = "PASS"
+                    report.status_extended = f"Conditional Access Policy '{policy.display_name}' enforces MFA for administrative roles."
+                    break
```
AdriiiPRodri: Is this change necessary? Before the default value was set to PASS and changed if `if policy.state == ConditionalAccessPolicyState.ENABLED_FOR_REPORTING`, I think the behavior now is the same, it can be done both ways but I want to know if I'm missing something.

AdriiiPRodri (changes requested):

prowler/providers/microsoft365/services/entra/entra_capolicy_administrative_roles_have_mfa_enabled/entra_capolicy_administrative_roles_have_mfa_enabled.py
HugoPBrito: Before, the `break` statement would came into action either it was `ENABLED` (compliant `PASS` finding) or `ENABLED_FOR_REPORTING` (not compliant and therefore `FAIL` finding).

This would trigger a **false positive** in cases where you have 2 well configured policies excepting the status.

If you had 1 non compliant and 1 compliant (1 enabled for reporting and 1 enabled), if the check read the non compliant first, it would have broken the execution and returned a `FAIL` besides having a policy that is compliant, which must have been a `PASS` finding.

HugoPBrito: Now we only break execution when you found a **truly compliant enabled policy**. If you find one that's well set up but not enabled, you keep searching for an enabled one.

AdriiiPRodri: Understood!

AdriiiPRodri (dismissed):

HugoPBrito:
I changed the name because the last one could lead to confusion as `admin_mfa` does not exist, there are no different types of MFA.

MrCloudSec (changes requested):
Please, rename the check to `entra_admin_users_have_mfa_enabled`

MrCloudSec:
Also, `entra_admin_mfa_enabled_for_administrative_roles` to `entra_admin_users_mfa_enabled` and `entra_admin_portals_role_limited_access` to `entra_admin_portals_access_restriction`

MrCloudSec (approved):

PR #7233, opened 2025-03-13
feat(jira): add basic auth method

pedrooot:
### Description

This PR adds basic auth for the jira class. It will need the following items:
* `user_mail`: the mail from the user
* `api_token`: the token generated in order to use the jira API
* `domain`: the site name

### Checklist

#### API

### License

### DIFF
--- a/prowler/lib/outputs/jira/jira.py
+++ b/prowler/lib/outputs/jira/jira.py
@@ -235,6 +235,9 @@ class Jira:
 
         Returns:
             - None
+
+        Raises:
+            - JiraBasicAuthError: Failed to authenticate using basic auth
         """
         try:
             user_string = f"{self._client_mail}:{self._api_token}"
--- a/prowler/lib/outputs/jira/jira.py
+++ b/prowler/lib/outputs/jira/jira.py
@@ -10,6 +10,7 @@ from prowler.lib.logger import logger
 from prowler.lib.outputs.finding import Finding
 from prowler.lib.outputs.jira.exceptions.exceptions import (
     JiraAuthenticationError,
+    JiraBasicAuthError,
     JiraCreateIssueError,
     JiraGetAccessTokenError,
     JiraGetAuthResponseError,
@@ -84,6 +85,7 @@ class Jira:
         - JiraCreateIssueError: Failed to create an issue in Jira
         - JiraSendFindingsResponseError: Failed to send the findings to Jira
         - JiraTestConnectionError: Failed to test the connection
+        - JiraBasicAuthError: Failed to authenticate using basic auth
 
     Usage:
         jira = Jira(
@@ -98,6 +100,10 @@ class Jira:
     _client_id: str = None
     _client_secret: str = None
     _access_token: str = None
+    _client_mail: str = None
+    _api_token: str = None
+    _site_name: str = None
+    _using_basic_auth: bool = False
     _refresh_token: str = None
     _expiration_date: int = None
     _cloud_id: str = None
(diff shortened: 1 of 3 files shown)


prowler/lib/outputs/jira/jira.py
```diff
@@ -157,6 +174,10 @@ def cloud_id(self, value):
     def scopes(self):
         return self._scopes
 
+    @property
+    def using_basic_auth(self):
+        return self._using_basic_auth
```
jfagoagas, line 187: Is this property needed?

prowler/lib/outputs/jira/jira.py
```diff
@@ -98,6 +100,10 @@ class Jira:
     _client_id: str = None
     _client_secret: str = None
     _access_token: str = None
+    _client_mail: str = None
+    _api_token: str = None
```
jfagoagas: For me `user_mail`/`mail` and `token` are more appropriate here.

prowler/lib/outputs/jira/jira.py
```diff
@@ -98,6 +100,10 @@ class Jira:
...
+    _site_name: str = None
```
jfagoagas: what about `domain`? Please review how Atlassian names this.

prowler/lib/outputs/jira/jira.py
```diff
@@ -120,14 +126,25 @@ def __init__(
...
         self._client_secret = client_secret
+        self._client_mail = client_mail
+        self._api_token = api_token
+        self._site_name = site_name
         self._scopes = ["read:jira-user", "read:jira-work", "write:jira-work"]
-        auth_url = self.auth_code_url()
-        authorization_code = self.input_authorization_code(auth_url)
-        self.get_auth(authorization_code)
+        # If the client mail, api token and site name are present, use basic auth
+        if client_mail and api_token and site_name:
```
jfagoagas: I think here we should do more validation about the arguments passed, to force each path if the required attributes are set.

prowler/lib/outputs/jira/jira.py
```diff
@@ -209,6 +230,31 @@ def get_timestamp_from_seconds(seconds: int) -> datetime:
         """
         return (datetime.now() + timedelta(seconds=seconds)).isoformat()
 
+    def get_basic_auth(self) -> None:
+        """Get the access token using the mail and api token
```
jfagoagas:
```suggestion
        """Get the access token using the mail and API token.
```

prowler/lib/outputs/jira/jira.py
```diff
@@ -120,14 +126,25 @@ def __init__(
...
         self._client_id = client_id
         self._client_secret = client_secret
+        self._client_mail = client_mail
+        self._api_token = api_token
+        self._site_name = site_name
         self._scopes = ["read:jira-user", "read:jira-work", "write:jira-work"]
-        auth_url = self.auth_code_url()
-        authorization_code = self.input_authorization_code(auth_url)
-        self.get_auth(authorization_code)
+        # If the client mail, api token and site name are present, use basic auth
```
jfagoagas:
```suggestion
        # If the client mail, API token and site name are present, use basic auth
```

prowler/lib/outputs/jira/jira.py
```diff
@@ -209,6 +230,31 @@ def get_timestamp_from_seconds(seconds: int) -> datetime:
...
+            - None
+
+        Raises:
+            - JiraBasicAuthError: Failed to authenticate using basic auth
+        """
+        try:
+            user_string = f"{self._client_mail}:{self._api_token}"
+            self._access_token = base64.b64encode(user_string.encode("utf-8")).decode(
+                "utf-8"
+            )
```
jfagoagas, line 254: Why do you need to decode?

jfagoagas (changes requested):

prowler/lib/outputs/jira/jira.py
pedrooot: I'm using it on the `test_connection`method

prowler/lib/outputs/jira/jira.py
pedrooot: I changed it from `client_mail` -> `user_mail` but for the token, I like `api_token` since we are already using token var in other parts of the code and `api_token` sounds more distinctive for me

prowler/lib/outputs/jira/jira.py
pedrooot: done, domain is the correct way to name it

prowler/lib/outputs/jira/jira.py
pedrooot: Done

prowler/lib/outputs/jira/jira.py
pedrooot: Done

prowler/lib/outputs/jira/jira.py
pedrooot: Done

prowler/lib/outputs/jira/jira.py
pedrooot: It's done this way because I need an string at the end. If I remove the decode part i'll get only the bytes

jfagoagas: Good to know, thanks!

prowler/lib/outputs/jira/jira.py
jfagoagas: Thanks for the clarification!

jfagoagas (dismissed):

jfagoagas (approved):

PR #7322, opened 2025-03-19
feat(aws): add static credentials for S3 and SH

pedrooot:
### Description

This PR allows creating a session inside each specific class (S3 and SH). This way there is no need to create a provider each time that we want to send findings using S3 or SH.

### Checklist

#### API

### License


prowler/providers/aws/lib/security_hub/security_hub.py
```diff
@@ -58,14 +60,63 @@ class SecurityHub:
...
+        - aws_access_key_id: The AWS access key ID.
+        - aws_secret_access_key: The AWS secret access key.
+        - aws_session_token: The AWS session token, optional.
+        - retries_max_attempts: The maximum number of retries for the AWS client.
+        - regions: A set of regions to audit.
+        """
+        if aws_session:
+            self._session = aws_session
+        else:
+            aws_setup_session = AwsSetUpSession(
```
jfagoagas, line 106: @AdriiiPRodri @vicferpoy what do you think it is better? Setup the session using the current approach with `AwsSetUpSession` or make the S3 and SecurityHub classes to inherit from a class called `AWSService` (or something similar) and then call the parent's `__init__`? 

Thanks!

AdriiiPRodri: I think that all AWS services have something in common such as the session, so I would create a parent class from which these classes would inherit

pedrooot: The thing is that hierarchy can't be applied here, we have to allow the class to use the `session` from the provider or create it inside.

jfagoagas (dismissed):
Finally 👏 There is room for improvement as we'll refactor the `AWSSession` to hold all the `AwsSetUpSession` logic, but it is fine for now.

jfagoagas (dismissed):

jfagoagas (approved):

PR #7348, opened 2025-03-24
feat(gcp): add check for dormant (unused) SA keys

bgdanix:
### Description

Enhance the Google Cloud security checks by checking if there are dormant user-managed keys for the Service Accounts. A dormant key / account is [considered dormant by Google](https://cloud.google.com/iam/docs/service-account-overview) if it hasn't been used for 180 days.

### Checklist

#### API

### License

### DIFF
--- a/prowler/compliance/gcp/mitre_attack_gcp.json
+++ b/prowler/compliance/gcp/mitre_attack_gcp.json
@@ -141,6 +141,7 @@
         "iam_organization_essential_contacts_configured",
         "iam_role_kms_enforce_separation_of_duties",
         "iam_role_sa_enforce_separation_of_duties",
+        "iam_sa_dormant_user_managed_keys",
         "iam_sa_no_administrative_privileges",
         "iam_sa_no_user_managed_keys",
         "iam_sa_user_managed_key_rotate_90_days",
--- a/prowler/compliance/gcp/mitre_attack_gcp.json
+++ b/prowler/compliance/gcp/mitre_attack_gcp.json
@@ -141,7 +141,7 @@
         "iam_organization_essential_contacts_configured",
         "iam_role_kms_enforce_separation_of_duties",
         "iam_role_sa_enforce_separation_of_duties",
-        "iam_sa_dormant_user_managed_keys",
+        "iam_sa_user_managed_key_unused",
         "iam_sa_no_administrative_privileges",
         "iam_sa_no_user_managed_keys",
         "iam_sa_user_managed_key_rotate_90_days",
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ It contains hundreds of controls covering CIS, NIST 800, NIST CSF, CISA, RBI, Fe
 | Provider | Checks | Services | [Compliance Frameworks](https://docs.prowler.com/projects/prowler-open-source/en/latest/tutorials/compliance/) | [Categories](https://docs.prowler.com/projects/prowler-open-source/en/latest/tutorials/misc/#categories) |
 |---|---|---|---|---|
 | AWS | 564 | 82 | 33 | 10 |
-| GCP | 77 | 13 | 6 | 3 |
+| GCP | 78 | 13 | 6 | 3 |
 | Azure | 140 | 18 | 7 | 3 |
 | Kubernetes | 83 | 7 | 4 | 7 |
 | Microsoft365 | 5 | 2 | 1 | 0 |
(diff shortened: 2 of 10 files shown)


bgdanix:
@MrCloudSec  Could you re-run the failed checks? Looks like an issue with the GitHub runner

bgdanix:
@MrCloudSec is it ok to be merged?

MrCloudSec (commented):
Thanks for your contribution @bgdanix ! 
I have changed the check to align it to Prowler.
One question, would not be more interesting just checking if there are User-Managed Service Account Keys, having them is already a misconfiguration.

bgdanix:
>One question, would not be more interesting just checking if there are User-Managed Service Account Keys, having them is already a misconfiguration.

@MrCloudSec it can be a misconfiguration (sometimes you depend on software that only works with user-managed keys: [see this](https://cloud.google.com/iam/docs/best-practices-service-accounts#choose-when-to-use)). This check helps narrowing by finding out where you have keys that you haven't used for a while and just forgot about them.

Once this is merged, I plan to open another PR to also look out for dormant SAs as well. These kind of findings are in line with the [SCC threat findings](https://cloud.google.com/security-command-center/docs/concepts-event-threat-detection-overview) (look out for "dormant")

bgdanix:
@MrCloudSec I've had a look at the changes you did and there's another one you need to change: add the "iam_sa_user_managed_key_unused" in the "prowler/compliance/gcp/mitre_attack_gcp.json" as part of MITRE T1078 (that's what Security Command Center puts it under).

MrCloudSec:
> > One question, would not be more interesting just checking if there are User-Managed Service Account Keys, having them is already a misconfiguration.
> 
> @MrCloudSec it can be a misconfiguration (sometimes you depend on software that only works with user-managed keys: [see this](https://cloud.google.com/iam/docs/best-practices-service-accounts#choose-when-to-use)). This check helps narrowing by finding out where you have keys that you haven't used for a while and just forgot about them.
> 
> Once this is merged, I plan to open another PR to also look out for dormant SAs as well. These kind of findings are in line with the [SCC threat findings](https://cloud.google.com/security-command-center/docs/concepts-event-threat-detection-overview) (look out for "dormant")

But even if a software can only use user-managed keys, it is still unsafe, is like having exposed resources to the internet, you may need to use a mutelist for those findings that even if they are failed, it is control for the workload needs.

prowler/compliance/gcp/mitre_attack_gcp.json
```diff
@@ -141,6 +141,7 @@
         "iam_organization_essential_contacts_configured",
         "iam_role_kms_enforce_separation_of_duties",
         "iam_role_sa_enforce_separation_of_duties",
+        "iam_sa_dormant_user_managed_keys",
```
MrCloudSec:
```suggestion
        "iam_sa_user_managed_key_unused",
```

bgdanix:
> But even if a software can only use user-managed keys, it is still unsafe, is like having exposed resources to the internet, you may need to use a mutelist for those findings that even if they are failed, it is control for the workload needs.

I fully agree. I see it more as a complementary detection, an easy answer to the question of: "do I have any user-managed keys unused for a long time?" so that I can get a quick-win, especially when considering a large GCP footprint with hundreds/thousand findings about generic user-managed keys.

MrCloudSec:
> > But even if a software can only use user-managed keys, it is still unsafe, is like having exposed resources to the internet, you may need to use a mutelist for those findings that even if they are failed, it is control for the workload needs.
> 
> I fully agree. I see it more as a complementary detection, an easy answer to the question of: "do I have any user-managed keys unused for a long time?" so that I can get a quick-win, especially when considering a large GCP footprint with hundreds/thousand findings about generic user-managed keys.

Maybe we can have this check to detect user-managed keys, and also, in the same check saying in the status extended when was the last time the key was used.

bgdanix:
> Maybe we can have this check to detect user-managed keys, and also, in the same check saying in the status extended when was the last time the key was used.

Yeah, that's an option. I would rather keep it separate though, the same way SCC has findings for both just to be on par with the official findings.

MrCloudSec:
> > Maybe we can have this check to detect user-managed keys, and also, in the same check saying in the status extended when was the last time the key was used.
> 
> Yeah, that's an option. I would rather keep it separate though, the same way SCC has findings for both just to be on par with the official findings.

Okay, we can have another one with lower severity that looks for user-managed keys independently of their last use. Would you be able to do it in another PR? If not, we can add the check to our internal check roadmap.

bgdanix:
> Okay, we can have another one with lower severity that looks for user-managed keys independently of their last use. Would you be able to do it in another PR?

From what I've seen there is already one that checks for user-managed keys so I could just make use of that one and add additional info in case that key hasn't been used for a long time.

MrCloudSec (approved):
Thank you for your effort @bgdanix !

PR #7361, opened 2025-03-24
fix(iam): handle `UnboundLocalError` cannot access local variable 'report'

danibarranqueroo:
### Context

This PR introduces changes in the check iam_custom_role_has_permissions_to_administer_resource_locks from IAM service in Azure. There we’re adding to the findings the variable report between the two loops, so if there is no roles there is no problem but as it's a dict if there's no roles but the key of the dict exists the report is not initialized and we’re trying to add to the findings nothing so it causes the error.

### Description

Modified the check and added a test case.

### Checklist

#### API

### License


andoniaf (approved):

MrCloudSec (approved):
Good catch!

PR #7381, opened 2025-03-26
fix(redshift): validation error for `Cluster.multi_az`

danibarranqueroo:
### Context

This PR introduces a fix for Redshift service in AWS. We’re having a validation error:
```
us-east-1 -- ValidationError[31]: 1 validation error for Cluster
multi_az - value could not be parsed to a boolean (type=type_error.bool)
```
This comes from the api call describe cluster of Redshift service, in boto3 documentation we can see:

`MultiAZ (string) – A boolean value that, if true, indicates that the cluster is deployed in two Availability Zones.`

So, the sdk is returning a string and we assume on our code that it’s a boolean value,  I've changed it so now it'll work in both cases, for String and for Boolean. Can't added test for the string case because moto create the parameter MultiAZ as a boolean.

### Description

Modified multi_az parameter.

### Checklist

#### API

### License

### DIFF
--- a/prowler/providers/aws/services/redshift/redshift_service.py
+++ b/prowler/providers/aws/services/redshift/redshift_service.py
@@ -45,8 +45,7 @@ class Redshift(AWSService):
                                 "AllowVersionUpgrade", False
                             ),
                             encrypted=cluster.get("Encrypted", False),
-                            multi_az=str(cluster.get("MultiAZ", False)).lower()
-                            == "true",
+                            multi_az=cluster.get("MultiAZ", ""),
                             region=regional_client.region,
                             tags=cluster.get("Tags"),
                             master_username=cluster.get("MasterUsername", ""),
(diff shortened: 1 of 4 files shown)


prowler/providers/aws/services/redshift/redshift_service.py
```diff
@@ -45,7 +45,8 @@ def _describe_clusters(self, regional_client):
                                 "AllowVersionUpgrade", False
                             ),
                             encrypted=cluster.get("Encrypted", False),
-                            multi_az=cluster.get("MultiAZ", False),
+                            multi_az=str(cluster.get("MultiAZ", False)).lower()
```
MrCloudSec:
```suggestion
                            multi_az=cluster.get("MultiAZ", "False)).lower()
```

If it is always a string, we don't need the cast, right? Can we test this, please?

danibarranqueroo: I've tested it and it strange, in their docs they say that `it's a string and if true it's enabled` but the sdk parameter is `cluster['MultiAZ'] = 'Enabled'`. So, I'll change that and save that parameter the same way it's returned by Boto3.

MrCloudSec (approved):

PR #7383, opened 2025-03-26
fix(rds): handle Certificate `rds-ca-2019` not found

danibarranqueroo:
### Context

We’re having an issue when calling the DescribeCertificates operation: `Certificate rds-ca-2019 not found`

The problem here is that `rds-ca-2019` is deprecated so it’s obvious that it cannot be found, the problem is that if as user you are still using a deprecated certificated on your RDS instance without changing the error will continue here:

[Amazon Docs where appears that CA as deprecated](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.SSL-certificate-rotation.html) 

So, my solution has been adding a try-except to that DescribeCertificates call and put a warning in case the certificate CA is rds-ca-2019 (deprecated in 2024) or rds-ca-2015 (deprecated in 2020)

### Description

Modified RDS service and added tests for the exception

### Checklist

#### API

### License

### DIFF
--- a/prowler/providers/aws/services/rds/rds_service.py
+++ b/prowler/providers/aws/services/rds/rds_service.py
@@ -149,24 +149,41 @@ class RDS(AWSService):
                         "describe_certificates"
                     )
                     if instance.ca_cert:
-                        for page in describe_db_certificates_paginator.paginate(
-                            CertificateIdentifier=instance.ca_cert
-                        ):
-                            for certificate in page["Certificates"]:
-                                instance.cert.append(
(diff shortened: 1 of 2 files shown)


prowler/providers/aws/services/rds/rds_service.py
MrCloudSec, line 1: I would do the warning generic by only handling this error,  no need to use the deprecated_certs variable since needs maintenance.

MrCloudSec (approved):

PR #7411, opened 2025-04-01
fix(report): log as error when Resource ID or Name do not exist

MrCloudSec:
### Description

Log as error when Resource ID or Name do not exist in the finding.

### Checklist

#### API

### License

### DIFF
--- a/prowler/lib/outputs/finding.py
+++ b/prowler/lib/outputs/finding.py
@@ -111,18 +111,6 @@ class Finding(BaseModel):
             finding_output (Finding): the finding output object
 
         """
-        if not check_output.resource_id:
-            logger.error(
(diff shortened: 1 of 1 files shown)


prowler/lib/outputs/finding.py
```diff
@@ -111,6 +111,18 @@ def generate_output(
             finding_output (Finding): the finding output object
 
         """
+        if not check_output.resource_id:
```
jfagoagas: Why not to always at the end of this function when the output is unified?

MrCloudSec: Done, good idea!

jfagoagas (approved):

PR #7423, opened 2025-04-03
chore(deletion): Add environment variable for batch size

jfagoagas:
### Description

 Add environment variable for the deletion tasks batch size.

### Checklist

#### API

### License


.env
```diff
@@ -133,3 +133,6 @@ SOCIAL_GOOGLE_OAUTH_CLIENT_SECRET=""
 SOCIAL_GITHUB_OAUTH_CALLBACK_URL="${AUTH_URL}/api/auth/callback/github"
 SOCIAL_GITHUB_OAUTH_CLIENT_ID=""
 SOCIAL_GITHUB_OAUTH_CLIENT_SECRET=""
+
+# Deletion Task Batch Size
+DJANGO_DELETION_BATCH_SIZE=5000
```
vicferpoy, line 138: Since this is the default value, we don't need to add this to the `.env`.

vicferpoy (changes requested):

.env
vicferpoy:
```suggestion
```
It's the default value, we don't need to add it.

vicferpoy (changes requested):

vicferpoy (approved):

PR #7466, opened 2025-04-08
fix(provider): disable periodic task on views before deleting

AdriiiPRodri:
### Context

Currently a user is unable to delete a provider because the API returns a 500. This is due to the deletion of periodic tasks at view level before deleting the provider itself.

### Description

In this PR we are going to disable the periodic task at the view level. After deleting the provider, we will remove the related periodic task.

### Checklist

#### API

### License

### DIFF
--- a/api/CHANGELOG.md
+++ b/api/CHANGELOG.md
@@ -16,6 +16,13 @@ All notable changes to the **Prowler API** are documented in this file.
 
 ---
 
+## [v1.5.4] (Prowler v5.4.4)
+
+### Fixed
+- Disabled periodic task deletion at view level due to performance issues ([#7466])(https://github.com/prowler-cloud/prowler/pull/7466).
+
+---
+
 ## [v1.5.3] (Prowler v5.4.3)
 
 ### Fixed
--- a/api/CHANGELOG.md
+++ b/api/CHANGELOG.md
@@ -19,7 +19,7 @@ All notable changes to the **Prowler API** are documented in this file.
 ## [v1.5.4] (Prowler v5.4.4)
 
 ### Fixed
-- Disabled periodic task deletion at view level due to performance issues ([#7466])(https://github.com/prowler-cloud/prowler/pull/7466).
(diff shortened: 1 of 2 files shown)


api/CHANGELOG.md
```diff
@@ -16,6 +16,13 @@ All notable changes to the **Prowler API** are documented in this file.
 
 ---
 
+## [v1.5.4] (Prowler v5.4.4)
+
+### Fixed
+- Disabled periodic task deletion at view level due to performance issues ([#7466])(https://github.com/prowler-cloud/prowler/pull/7466).
```
```suggestion
- Fixed a bug with periodic tasks when trying to delete a provider ([#7466])(https://github.com/prowler-cloud/prowler/pull/7466).
```

api/src/backend/api/v1/views.py
```diff
@@ -1089,7 +1089,12 @@ def destroy(self, request, *args, pk=None, **kwargs):
         provider = get_object_or_404(Provider, pk=pk)
         provider.is_deleted = True
         provider.save()
-        delete_related_daily_task(str(provider.id))
+        # Temporarily disabled due to performance issues
+        # delete_related_daily_task(str(provider.id))
+
+        # TODO: Disable instead of delete, review performance
```
```suggestion

```

vicferpoy (changes requested):

api/src/backend/api/v1/views.py
```diff
@@ -1089,7 +1089,8 @@ def destroy(self, request, *args, pk=None, **kwargs):
         provider = get_object_or_404(Provider, pk=pk)
         provider.is_deleted = True
         provider.save()
-        delete_related_daily_task(str(provider.id))
+        task_name = f"scan-perform-scheduled-{str(provider.id)}"
```
```suggestion
        task_name = f"scan-perform-scheduled-{pk}"
```

vicferpoy (changes requested):

vicferpoy (approved):
Thank you! Let's wait a bit before merging to decide which version we release this on 💯

PR #7498, opened 2025-04-14
fix(findings): increase uid max length to 600

jfagoagas:
### Context

The `finding.uid` field has a maximum length of 300 with for some resources is not enough.

### Description

Increate `finding.uid` max length from `varchar(300)` to `varchar(600)`. The migration generated does not block the `findings.table` or requires a rewrite because:
- PostgreSQL does not physically rewrite the table just because the length limit is increased.
- It only updates the system catalog metadata to reflect the new limit. 

### Checklist

#### API
- [x] Verify if API specs need to be regenerated. -> to 1.6.1

### License


AdriiiPRodri (dismissed):

api/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to the **Prowler API** are documented in this file.
 
+## [v1.6.1] (Prowler UNRELEASED)
+
+###
+- Changed `findings.uid` field length from `varchar(300)` to `varchar(300)` [(#7498)](https://github.com/prowler-cloud/prowler/pull/7498).
```
```suggestion
### Changed
- Changed `findings.uid` field length from `varchar(300)` to `varchar(600)` [(#7498)](https://github.com/prowler-cloud/prowler/pull/7498).

```

vicferpoy (changes requested):
The PR title says 500. Also, could you add a unit tests for this?

api/CHANGELOG.md
jfagoagas: 600 is correct, I've updated the PR title.

vicferpoy (approved):
Thanks for your contribution! 🫡

PR #7540, opened 2025-04-16
chore(sentry): use exceptions instead of logger.error to handle check exceptions

andoniaf:
### Context

Sending exceptions as a log error means we lose valuable info when reviewing those exceptions, like the trace pointing to where exactly the exception happened in the code.

### Description

- Replace all `logger.error` with `logger.exception` inside `prowler/providers`.
  - Important to remember to start using `logger.exception` from now on.
- Add `ignore_errors` to sentry configuration so it honors the same ignore list we already had.
- Minor fixes in the docs.

### Checklist

#### API

### License

### DIFF
--- a/api/README.md
+++ b/api/README.md
@@ -235,6 +235,7 @@ To view the logs for any component (e.g., Django, Celery worker), you can use th
 
 ```console
 docker logs -f $(docker ps --format "{{.Names}}" | grep 'api-')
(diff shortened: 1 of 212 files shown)


andoniaf:
Replaced by https://github.com/prowler-cloud/prowler/pull/7598

PR #7545, opened 2025-04-17
fix(aws): update bucket naming validation to accept dots

### LINKED ISSUE #7543: [Prowler-cli] Allow . in bucket name
### New feature motivation

When using prowler cli, options like `--output-bucket` / `-B` allow users to push scan reports to s3 buckets.

A check enforces AWS naming rules:
https://github.com/prowler-cloud/prowler/blob/abae5f1626f3c1ccfd5e73fb8d2a83186a9d0940/prowler/providers/aws/lib/arguments/arguments.py#L227

Said naming rules do not recommend, **yet allow** uses of dots. the regex forbids dots.

### Solution Proposed

Allow dots in regex

`"(?!(^xn--|.+-s3alias$))^[a-z0-9][a-z0-9-\.]{1,61}[a-z0-9]$"`

Remove check entirely. It's up to the user to provide the bucket he wants. If aws let me create a bucket, I should be able to use it with prowler.

### Additional context

_No response_

andoniaf:
### Description

> Bucket names can consist only of lowercase letters, numbers, periods (.), and hyphens (-).
from [AWS Docs](https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html)

Solves #7543 

### Checklist

#### API

### License

### DIFF
--- a/prowler/CHANGELOG.md
+++ b/prowler/CHANGELOG.md
@@ -19,6 +19,7 @@ All notable changes to the **Prowler SDK** are documented in this file.
 - Fix package name location in pyproject.toml while replicating for prowler-cloud [(#7531)](https://github.com/prowler-cloud/prowler/pull/7531).
 - Remove cache in PyPI release action [(#7532)](https://github.com/prowler-cloud/prowler/pull/7532).
 - Add the correct values for logger.info inside iam service [(#7526)](https://github.com/prowler-cloud/prowler/pull/7526).
+- Update S3 bucket naming validation to accept dots [(#7545)](https://github.com/prowler-cloud/prowler/pull/7545).
 
 ---
 
--- a/prowler/providers/aws/lib/arguments/arguments.py
+++ b/prowler/providers/aws/lib/arguments/arguments.py
@@ -224,7 +224,10 @@ def validate_arguments(arguments: Namespace) -> tuple[bool, str]:
 
 def validate_bucket(bucket_name: str) -> str:
     """validate_bucket validates that the input bucket_name is valid"""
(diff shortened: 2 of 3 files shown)


MrCloudSec (changes requested):
Great work on the regex, it covers almost all the key rules for valid S3 bucket names. Just a quick note: AWS recently added a few more reserved suffixes like `.mrap`, `--x-s3`, and `--table-s3` that we might want to block as well. Also, small tweak suggestion to ensure it enforces the 3-character minimum.

Overall though, very solid work 👏

metahertz:
Love the addition of known bad bucket prefixes! Great update, agree with @MrCloudSec otherwise looks awesome.

andoniaf:
> Great work on the regex, it covers almost all the key rules for valid S3 bucket names. Just a quick note: AWS recently added a few more reserved suffixes like `.mrap`, `--x-s3`, and `--table-s3` that we might want to block as well. Also, small tweak suggestion to ensure it enforces the 3-character minimum.
> 
> Overall though, very solid work 👏

I thought on adding those too, but as this is for validating buckets that use can use (not create) I wasn't sure as people could be using those kind of buckets.

Even more, I'm thinking on removing the `-s3alias` and `--ol-s3` restrictions. We do not have a reason to restrict which kind of bucket you're using for this and people may have "different architectures" that rely on those kind of buckets 😅 

Also, the regex covers the 3 char limitation.  I know it doesn't looks explicit enough so I'll add a 3-char test case to make it clear.

MrCloudSec (approved):

PR #7547, opened 2025-04-18
fix(aws): remove `SHA-1` from ACM insecure key algorithms

F30:
### Context

Listing a hash function gives the impression that one could check the hashing algorithm of a cert's signature.

But this is not possible, the code only ever compares to `certificate.key_algorithm`, which is either RSA or a curve.

### Description

Remove SHA-1 from default insecure key algorithms.

### Checklist

- Are there new checks included in this PR? → No.
- [ ] Ensure new entries are added to [CHANGELOG.md](https://github.com/prowler-cloud/prowler/blob/master/prowler/CHANGELOG.md), if applicable. → **Not quite sure if this qualifies as a "change to the Prowler SDK".**

### License

### DIFF
--- a/contrib/k8s/helm/prowler-api/values.yaml
+++ b/contrib/k8s/helm/prowler-api/values.yaml
@@ -399,7 +399,6 @@ mainConfig:
       [
         "RSA-1024",
         "P-192",
-        "SHA-1",
       ]
 
     # AWS EKS Configuration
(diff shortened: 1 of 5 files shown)


MrCloudSec (approved):
Good catch @F30. You're right that `key_algorithm` never returns "SHA-1", so including it here could be misleading. Thanks for the cleanup.

PR #7568, opened 2025-04-22
feat(teams): add new check `teams_meeting_external_lobby_bypass_disabled`

HugoPBrito:
### Context

Allowing unrestricted meeting access can compromise the confidentiality of meetings, especially when sensitive information is being discussed. Without proper controls, external or unauthenticated participants might gain access before being vetted by the organizer.

### Description

This check ensures that only individuals within the organization can bypass the lobby when joining a Teams meeting. Anyone outside the organization, including dial-in callers, will be placed in the lobby and must be admitted by an organizer, co-organizer, or presenter before joining. This measure adds a layer of control and security for meetings involving sensitive information.

### Checklist

#### API

### License


andoniaf (dismissed):
🙌

prowler/providers/m365/services/teams/teams_meeting_external_lobby_bypass_disabled/teams_meeting_external_lobby_bypass_disabled.py
```diff
@@ -0,0 +1,47 @@
...
+                resource_id="teamsMeetingsGlobalPolicy",
+            )
+            report.status = "FAIL"
+            report.status_extended = (
+                "People outside the organization can bypass the lobby."
+            )
+
+            if (
+                global_meeting_policy.allow_external_users_to_bypass_lobby
+                == "EveryoneInCompanyExcludingGuests"
```
MrCloudSec: Which are the possible values here? Maybe other values like "NoOne" can also be a PASS.

MrCloudSec (changes requested):
@HugoPBrito review my comment.

prowler/providers/m365/services/teams/teams_meeting_external_lobby_bypass_disabled/teams_meeting_external_lobby_bypass_disabled.py
HugoPBrito: These are the current options:

image

If we want to support more options than the recommended one, we could add an `Enum` with all possible values, although that would take some extra time.

andoniaf: [Here](https://learn.microsoft.com/en-us/powershell/module/teams/set-csteamsmeetingpolicy?view=teams-ps#-autoadmittedusers) is documented what does every option.

Maybe we should include as PASS `OrganizerOnly` too.

PS: I love how the options are not named the same in different docs. 🙃

MrCloudSec: I would say to include these ones that all of them exclude guests:
image

MrCloudSec (approved):

PR #7569, opened 2025-04-22
feat(exchange): add new check `exchange_transport_rules_whitelist_disabled`

danibarranqueroo:
### Context

This PR introduces a new check for Exchange service in M365. Mail flow rules (transport rules) in Exchange Online are used to identify and take action on messages that flow through the organization.
Whitelisting domains in transport rules bypasses regular malware and phishing scanning, which can enable an attacker to launch attacks against your users from a safe haven domain, to ensure this is not happening we need to verify there is no rule with spam confidence level (SCL) to '-1' and affecting to any domain.

### Description

Added new check `exchange_transport_rules_whitelist_disabled`, modified the service and added unit tests.

### Checklist

#### API

### License


prowler/providers/m365/services/exchange/exchange_service.py
```diff
@@ -53,6 +56,28 @@ def _get_mailbox_audit_config(self):
...
+        logger.info("Microsoft365 - Getting transport rules configuration...")
+        transport_rules = []
+        try:
+            rules_data = self.powershell.get_transport_rules()
+            if isinstance(rules_data, dict):
+                rules_data = [rules_data]
+            for rule in rules_data:
+                if rule:
+                    transport_rules.append(
+                        TransportRule(
```
andoniaf: This implementation checks `if rule:` inside the loop, but won't be better to validate the input data earlier, adding an early return if `rules_data` is empty?
```suggestion
            rules_data = self.powershell.get_transport_rules()
            if not rules_data:
                return transport_rules
            if isinstance(rules_data, dict):
                rules_data = [rules_data]
            for rule in rules_data:
                transport_rules.append(
                    TransportRule(
```

tests/providers/m365/services/exchange/exchange_transport_rules_whitelist_disabled/exchange_transport_rules_whitelist_disabled_test.py
```diff
@@ -0,0 +1,126 @@
...
+                == "Transport rule 'Rule1' does not whitelist any domains."
+            )
+            assert result[0].resource_name == "Rule1"
+            assert result[0].resource_id == "ExchangeTransportRule"
+            assert result[1].status == "PASS"
+            assert (
+                result[1].status_extended
+                == "Transport rule 'Rule2' does not whitelist any domains."
+            )
+            assert result[1].resource_name == "Rule2"
```
andoniaf, line 53: The result here is FAIL because the rule has domain and the scl is `-1`, do we want to change the assert or the mocked rule?

tests/providers/m365/services/exchange/exchange_transport_rules_whitelist_disabled/exchange_transport_rules_whitelist_disabled_test.py
```diff
@@ -0,0 +1,126 @@
...
+
+            assert len(result) == 1
+            assert result[0].status == "FAIL"
+            assert (
+                result[0].status_extended
+                == "Transport rule 'WhitelistRule' whitelists domains: whitelist.com."
+            )
+            assert result[0].resource_name == "WhitelistRule"
+            assert result[0].resource_id == "ExchangeTransportRule"
+            assert result[0].location == "global"
```
andoniaf, line 98: Similar to [this](https://github.com/prowler-cloud/prowler/pull/7569/files#r2056113429). Not sure if we prefer to change the assert or the mocked rule.

prowler/providers/m365/services/exchange/exchange_service.py
MrCloudSec: Good idea, but this would not check if the rule inside the `rules_data` exists.

danibarranqueroo: To cover both cases I've added @andoniaf suggestion maintaining the `if rule` inside the `rules_data` loop.

tests/providers/m365/services/exchange/exchange_transport_rules_whitelist_disabled/exchange_transport_rules_whitelist_disabled_test.py
danibarranqueroo: Pass and fail cases were inverted in the unit tests, I'd fix it now. Nice catch!

tests/providers/m365/services/exchange/exchange_transport_rules_whitelist_disabled/exchange_transport_rules_whitelist_disabled_test.py
danibarranqueroo: Same thing as [here](https://github.com/prowler-cloud/prowler/pull/7569#discussion_r2056545516).

MrCloudSec (approved):

PR #7580, opened 2025-04-22
feat(exchange): add new check `exchange_external_email_tagging_enabled`

danibarranqueroo:
### Context

This PR introduces a new check for Exchange service in M365. Tagging emails from external senders helps to inform end users about the origin of the email. This can allow them to proceed with more caution and make informed decisions when it comes to identifying spam or phishing emails. This feature can be enabled/disabled only via PowerShell.

### Description

Added new check `exchange_external_mail_tagged`, modified the service and added unit tests.

### Checklist

#### API

### License


prowler/providers/m365/services/exchange/exchange_service.py
```diff
@@ -53,6 +54,29 @@ def _get_mailbox_audit_config(self):
             )
         return mailboxes_config
 
+    def _get_external_mail_config(self):
+        logger.info("Microsoft365 - Getting external mail configuration...")
+        external_mail_config = []
+        try:
+            external_mail_configuration = self.powershell.get_external_mail_config()
```
andoniaf, line 64: Similar to this [comment](https://github.com/prowler-cloud/prowler/pull/7569/files#r2056061012).

I think I would be better to exit early instead of checking inside the loop.

danibarranqueroo: Same as [this](https://github.com/prowler-cloud/prowler/pull/7569#discussion_r2056544451), we still need to maintain the if inside the loop but is good also to verify if the parameter exists before it.

MrCloudSec (dismissed):

MrCloudSec (approved):

PR #7614, opened 2025-04-25
feat(teams): add new checks `teams_security_reporting_enabled` and `defender_chat_report_policy_configured`

andoniaf:
### Context
Without proper security reporting enabled, users cannot effectively report suspicious or malicious messages, potentially allowing security threats to go unnoticed.

### Description
This checks ensure:
- Users are allowed to report security concerns in Teams settings.
- Report Submission Policy is properly configured in Defender settings.

### Checklist

#### API

### License


MrCloudSec (dismissed):
@andoniaf, please solve the conflicts in this one.

prowler/providers/m365/services/teams/teams_security_reporting_enabled/teams_security_reporting_enabled.py
MrCloudSec, line 1: @andoniaf thanks for solving conflicts. Please, add different status extended so the user can know which service is misconfigured.

HugoPBrito: I suggest dividing the control into two checks, as done in this [PR](https://github.com/prowler-cloud/prowler/pull/7658), since several services are being audited. Then, we can include both checks in the compliance mapping. This approach maintains the finding-to-resource philosophy and we keep services isolated.

MrCloudSec: I really like that idea @HugoPBrito! Can you do the changes according to that, @andoniaf please?

andoniaf: Good idea, changes done ✅

prowler/providers/m365/services/defender/defender_report_policy_configured/defender_report_policy_configured.metadata.json
MrCloudSec, line 1: @andoniaf I think "Ensure Monitor reported messages in Microsoft Teams" should be another check since it is separated from "Send reported messages to:" setting in Defender.

MrCloudSec: Also, the `Get-ReportSubmissionPolicy` seems to be from the Exchange service, maybe we should redefine the check. What do you think?

andoniaf: > @andoniaf I think "Ensure Monitor reported messages in Microsoft Teams" should be another check since it is separated from "Send reported messages to:" setting in Defender.

We could create another check only for verifying that Report Email addresses are configured, but I don't think we can remove it from this one as we need to check all the configuration. 
Maybe we should rename it so it specifies this check is only for Teams (even it's checked in Defender service). Something like "Ensure chat report submission policy is properly configured in Defender".



> Also, the `Get-ReportSubmissionPolicy` seems to be from the Exchange service, maybe we should redefine the check. What do you think?

Yes, the cmdlet it's from Exchange, but in the UI is configured through Defender. We already have some checks like this.

andoniaf: I updated the check name, let me know what do you think.

HugoPBrito: @andoniaf, renaming the check was a good decision, as the previous name could have led to misunderstandings. I believe this check meets the requirements of the CIS control we need to asses, and implemented as necessary.

@MrCloudSec, Defender service always uses the Exchange module. In this case, we'll prioritize our resource-service structure for the users over the one used by the PowerShell cmdlets Microsoft uses, so I think it's perfect since the resource comes from Defender.

prowler/providers/m365/services/defender/defender_chat_report_policy_configured/defender_chat_report_policy_configured.py
```diff
@@ -0,0 +1,53 @@
...
+        Returns:
+            List[CheckReportM365]: A list of reports containing the result of the check.
+        """
+        findings = []
+        report_submission_policy = defender_client.report_submission_policy
+
+        if report_submission_policy:
+            report = CheckReportM365(
+                metadata=self.metadata(),
+                resource=report_submission_policy if report_submission_policy else {},
```
HugoPBrito:
```suggestion
                resource=report_submission_policy,
```
It's already inside a `if report_submission_policy:` statement.

prowler/providers/m365/services/teams/teams_security_reporting_enabled/teams_security_reporting_enabled.py
```diff
@@ -0,0 +1,50 @@
...
+        Returns:
+            List[CheckReportM365]: A list of reports containing the result of the check.
+        """
+        findings = []
+        global_messaging_policy = teams_client.global_messaging_policy
+
+        if global_messaging_policy:
+            report = CheckReportM365(
+                metadata=self.metadata(),
+                resource=global_messaging_policy if global_messaging_policy else {},
```
HugoPBrito:
```suggestion
                resource=global_messaging_policy,
```
It's already inside a `if global_messaging_policy:` statement.

prowler/providers/m365/services/teams/teams_service.py
```diff
@@ -75,6 +77,9 @@ def _get_global_meeting_policy(self):
                     designated_presenter_role_mode=global_meeting_policy.get(
                         "DesignatedPresenterRoleMode", "EveryoneUserOverride"
                     ),
+                    allow_security_end_user_reporting=global_meeting_policy.get(
+                        "AllowSecurityEndUserReporting", True
```
HugoPBrito:
```suggestion
                        "AllowSecurityEndUserReporting", False
```
I would set the non compliant value if we couldn't retrieve the actual one.

prowler/providers/m365/services/teams/teams_service.py
```diff
@@ -85,6 +90,23 @@ def _get_global_meeting_policy(self):
...
 
+    def _get_global_messaging_policy(self):
+        logger.info("M365 - Getting Teams global messaging policy...")
+        global_messaging_policy = None
+        try:
+            global_messaging_policy = self.powershell.get_global_messaging_policy()
+            if global_messaging_policy:
+                global_messaging_policy = GlobalMessagingPolicy(
+                    allow_security_end_user_reporting=global_messaging_policy.get(
+                        "AllowSecurityEndUserReporting", True
```
HugoPBrito:
```suggestion
                        "AllowSecurityEndUserReporting", False
```
Same here.

prowler/providers/m365/services/teams/teams_service.py
```diff
@@ -130,6 +152,10 @@ class GlobalMeetingPolicy(BaseModel):
     meeting_chat_enabled_type: str = "EnabledForEveryone"
 
 
+class GlobalMessagingPolicy(BaseModel):
+    allow_security_end_user_reporting: bool = True
```
HugoPBrito:
```suggestion
    allow_security_end_user_reporting: bool = False
```
Same here.

HugoPBrito (changes requested):
Please, take a look and let me know what you think!

prowler/providers/m365/services/teams/teams_service.py
HugoPBrito: I say this because, from my point of view, it's better to create a false positive finding than a false negative.

prowler/providers/m365/services/defender/defender_chat_report_policy_configured/defender_chat_report_policy_configured.py
andoniaf: True! 👍

prowler/providers/m365/services/teams/teams_service.py
andoniaf: yes, agree 👍

HugoPBrito (approved):
Perfect, thanks!

PR #7618, opened 2025-04-26
feat(m365): automate `PowerShell` modules installation

HugoPBrito:
### Context

There are several `PowerShell modules` needed to successfully execute M365 Provider.

### Description

This PR includes the implementation of the installation automation of these modules during the provider initialization.

### Checklist

#### API

### License


util/m365_powershell_modules_installation.py
jfagoagas, line 1: I think it is a good time to create a `m365` folder within `util`.

What happens if a user wants to run this script from Powershell to install the modules?

andoniaf: From PS works the same:
```
PS /home/prowler> poetry run python ./m365_powershell_modules_installation.py
```

andoniaf (dismissed):

andoniaf (dismissed):

prowler/lib/powershell/powershell.py
```diff
@@ -191,8 +191,15 @@ def error_reader_thread():
...
-        error_result = error_queue.get(timeout=1)
+        try:
+            result = result_queue.get(timeout=timeout) or default
+        except queue.Empty:
+            result = default
+
+        try:
+            error_result = error_queue.get(timeout=1)
+        except queue.Empty:
+            error_result = ""
```
AdriiiPRodri: Maybe we can change the code like this?:

```suggestion
        try:
            result = result_queue.get(timeout=timeout) or default
            error_result = error_queue.get(timeout=1)
        except queue.Empty:
            result = default
```

prowler/providers/m365/lib/arguments/arguments.py
```diff
@@ -35,16 +35,9 @@ def init_parser(self):
...
-        help="Microsoft 365 user email",
-    )
-    m365_parser.add_argument(
-        "--encypted-password",
-        nargs="?",
-        default=None,
-        help="Microsoft 365 encrypted password",
+        "--init-modules",
+        action="store_true",
+        help="Initialize Microsoft 365 PowerShell modules",
```
AdriiiPRodri, line 40: Now how the user can specify the user and password? This affects the API behaviour?

prowler/providers/m365/lib/powershell/m365_powershell.py
```diff
@@ -467,20 +487,61 @@ def get_inbound_spam_filter_policy(self) -> dict:
...
+                logger.error(f"Failed to initialize module {module}: {str(error)}")
+                return False
+
+        return True
+    finally:
+        pwsh.close()
+
+
+def main():
+    initialize_m365_powershell_modules()
```
AdriiiPRodri: This function haven't unittests, it is worth right now but we need to add it when we have a chance

AdriiiPRodri (changes requested):

prowler/providers/m365/lib/arguments/arguments.py
HugoPBrito: This change should not affect API as they:
1. Were not being used at all by the sdk
2. In the hypothetical case they were, they should have been being initialized from CLI only

prowler/lib/powershell/powershell.py
HugoPBrito: Of course. Thanks for the suggestion.

AdriiiPRodri (approved):

PR #7653, opened 2025-05-02
feat: add compliance to API report files and its endpoint

AdriiiPRodri:
### Description

In this PR we have added the compliance output to the ZIP of the API outputs.

We have also added an endpoint to be able to retrieve a specific compliance within a scan, whether the results are uploaded to S3 or locally.
image

### Checklist

#### API

### License


api/src/backend/api/specs/v1.yaml
jfagoagas, line 1: Please review all texts to match with the syntax of the rest of endpoints.

api/src/backend/api/v1/views.py
```diff
@@ -150,6 +150,7 @@
     UserSerializer,
     UserUpdateSerializer,
 )
+from prowler.config.config import get_available_compliance_frameworks
```
jfagoagas: I think the API has something during runtime mapping all checks and compliance frameworks. Please review what is done during the `scan` to generate the compliance overview.

api/CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ All notable changes to the **Prowler API** are documented in this file.
 ### Added
 
 - Added M365 as a new provider [(#7563)](https://github.com/prowler-cloud/prowler/pull/7563).
+- Added compliance folder to export outputs as ZIP functionality [(#7653)](https://github.com/prowler-cloud/prowler/pull/7653).
+- Added a new endpoint to download the specified compliance file [(#7653)](https://github.com/prowler-cloud/prowler/pull/7653).
```
jfagoagas: Why are you adding two entries to the changelog?

api/pyproject.toml
```diff
@@ -7,7 +7,7 @@ authors = [{name = "Prowler Engineering", email = "engineering@prowler.com"}]
 dependencies = [
   "celery[pytest] (>=5.4.0,<6.0.0)",
   "dj-rest-auth[with_social,jwt] (==7.0.1)",
-  "django==5.1.7",
+  "django==5.1.8",
```
jfagoagas, line 10: Are we finally including this in the PR?

api/src/backend/api/v1/views.py
```diff
@@ -1219,70 +1246,86 @@ def partial_update(self, request, *args, **kwargs):
...
-            bucket_name = env.str("DJANGO_OUTPUT_S3_AWS_OUTPUT_BUCKET")
-            key = output_location[len(f"s3://{bucket_name}/") :]
+            if list_objects:
+                # list keys under prefix then match suffix
+                prefix = os.path.dirname(path_pattern)
+                suffix = os.path.basename(path_pattern)
+                try:
+                    resp = client.list_objects_v2(Bucket=bucket, Prefix=prefix)
+                except ClientError:
+                    return Response(
```
jfagoagas, line 1329: I think this needs to be refined as we previously did
```python
try:
                s3_object = s3_client.get_object(Bucket=bucket_name, Key=key)
            except ClientError as e:
                error_code = e.response.get("Error", {}).get("Code")
                if error_code == "NoSuchKey":
                    return Response(
                        {"detail": "The scan has no reports."},
                        status=status.HTTP_404_NOT_FOUND,
                    )
                return Response(
                    {"detail": "There is a problem with credentials."},
                    status=status.HTTP_403_FORBIDDEN,
                )
```

We should not raise `HTTP 500 Internal Server Error`.

api/src/backend/api/v1/views.py
```diff
@@ -1219,70 +1246,86 @@ def partial_update(self, request, *args, **kwargs):
...
+            data=serializer.data,
+            status=status.HTTP_202_ACCEPTED,
+            headers={
+                "Content-Location": reverse(
+                    "task-detail", kwargs={"pk": serializer.data["id"]}
+                )
+            },
+        )
+
+    def _load_file(self, path_pattern, s3=False, bucket=None, list_objects=False):
```
jfagoagas, line 1294: Do we have tests for this specific function? I mean specific unit tests not as part of the view's.

api/src/backend/api/v1/views.py
```diff
@@ -1219,70 +1246,86 @@ def partial_update(self, request, *args, **kwargs):
...
-                headers={
-                    "Content-Location": reverse(
-                        "task-detail", kwargs={"pk": output_serializer.data["id"]}
-                    )
-                },
-            )
+    def _get_running_task_response(self, scan_instance):
+        """
+        If the scan or its report-generation task is still executing,
+        return a 202 Response with the task payload and Content-Location.
```
```suggestion
        return an `HTTP 202 Accepted` response with the task payload and Content-Location.
```

api/src/backend/api/v1/views.py
```diff
@@ -1291,28 +1334,97 @@ def report(self, request, pk=None):
...
+
         return response
 
+    @action(detail=True, methods=["get"], url_name="report")
+    def report(self, request, pk=None):
+        scan = self.get_object()
+        # Check for executing tasks
+        running_resp = self._get_running_task_response(scan)
+        if running_resp:
+            return running_resp
```
jfagoagas: For me something like `get_task_status` or `check_task_in_progress` is more readable, but that's a personal preference.

api/src/backend/api/v1/views.py
jfagoagas: Could you specify the return type? But it is a little bit weird to me to play with either `Response` or the tuple `content` and `filename`. Again, not expert on this regard, but not so readable at first.

api/src/backend/api/v1/views.py
```diff
@@ -1291,28 +1334,97 @@ def report(self, request, pk=None):
...
+                os.path.dirname(key_prefix), "compliance", f"{name}.csv"
+            )
+            loader = self._load_file(prefix, s3=True, bucket=bucket, list_objects=True)
+        else:
+            base = os.path.dirname(scan.output_location)
+            pattern = os.path.join(base, "compliance", f"*_{name}.csv")
+            loader = self._load_file(pattern, s3=False)
+
+        if isinstance(loader, Response):
+            return loader
```
jfagoagas, line 1448: I think this is related to the previous comment, not so readable at first.

api/src/backend/tasks/jobs/export.py
```diff
@@ -154,3 +244,41 @@ def _generate_output_directory(
     os.makedirs("/".join(path.split("/")[:-1]), exist_ok=True)
 
     return path
+
+
+def _generate_compliance_output_directory(
```
jfagoagas: I think you can remove this function and implement a simple condition within `_generate_output_directory`.

api/src/backend/tasks/tasks.py
```diff
@@ -251,84 +256,110 @@ def generate_outputs(scan_id: str, provider_id: str, tenant_id: str):
...
+        In both cases set `.close_file = is_last`.
+        """
+        initialization = False
+        if name not in writer_map:
+            writer_map[name] = factory()
+            initialization = True
+        w = writer_map[name]
+        w.close_file = is_last
+
+        return w, initialization
```
jfagoagas, line 281: Is this function within the `generate_outputs` function? If so, what is the reason behind?

api/src/backend/tasks/tasks.py
```diff
@@ -251,84 +256,110 @@ def generate_outputs(scan_id: str, provider_id: str, tenant_id: str):
...
+            compliance_obj = frameworks_bulk[name]
+            klass = next(
+                (
+                    c
+                    for cond, c in COMPLIANCE_CLASS_MAP.get(provider_type, [])
+                    if cond(name)
+                ),
+                GenericCompliance,
+            )
+            filename = f"{comp_dir}_{name}.csv"
```
jfagoagas, line 328: Does the compliance output have the same name as in the SDK? `prowler-output-{provider.uid}-{unixtimestamp}_{framework_name}` like: `prowler-output-106908755756-20250505125041_aws_account_security_onboarding_aws`.

prowler/lib/outputs/compliance/compliance_output.py
```diff
@@ -31,14 +31,21 @@ def __init__(
...
+        self.file_path = file_path
         self.file_descriptor = None
+        # This parameter is to avoid refactoring more code, the CLI does not write in batches, the API does
+        self._from_cli = from_cli
 
         if not file_extension and file_path:
             self._file_extension = "".join(Path(file_path).suffixes)
         if file_extension:
             self._file_extension = file_extension
+            self.file_path = f"{file_path}{self.file_extension}"
```
jfagoagas, line 48: It is a simple addition but it'd be great to include it in the SDK's tests.

jfagoagas (changes requested):
What a great job @AdriiiPRodri 👏 

Some things that we need to review:
- Review API version, if we are creating a new endpoint we should bump for a new minor.
- We need to add tests to cover changes in both existing and new code.

Regarding some parts of the API I need @vicferpoy to review the code.

api/src/backend/api/v1/views.py
```diff
@@ -1291,28 +1334,97 @@ def report(self, request, pk=None):
...
-            file_content = s3_object["Body"].read()
-            filename = os.path.basename(output_location.split("/")[-1])
+            content = s3_obj["Body"].read()
+            filename = os.path.basename(key)
         else:
-            zip_files = glob.glob(output_location)
-            try:
-                file_path = zip_files[0]
-            except IndexError as e:
-                sentry_sdk.capture_exception(e)
```
jfagoagas, line 1301: Why did you remove this?

AdriiiPRodri:
> What a great job @AdriiiPRodri 👏
> 
> Some things that we need to review:
> 
> * Review API version, if we are creating a new endpoint we should bump for a new minor.
> * We need to add tests to cover changes in both existing and new code.
> 
> Regarding some parts of the API I need @vicferpoy to review the code.

There is a lack of tests prior to this PR, so tests have been added to the modules that previously had them. Adding new test files may take more time and as I say it is something that was missing previously. Then we will have to decide whether to create those tests in this or another PR

api/src/backend/api/v1/views.py
AdriiiPRodri: Because I made a refactor to avoid code repetition, now we check by using suffixes and patterns if the file exists or not:

```
keys = [obj["Key"] for obj in contents if obj["Key"].endswith(suffix)]
if not keys:
    return Response(
        {
            "detail": f"No compliance file found for name '{os.path.splitext(suffix)[0]}'."
        },
        status=status.HTTP_404_NOT_FOUND,
    )
```

api/src/backend/tasks/tasks.py
jfagoagas: Again, to me this is really hard to follow.

AdriiiPRodri: Yes, has the same name:

API:
`prowler-output-106908755756-20250505085215_ffiec_aws.csv`

CLI:
`prowler-output-106908755756-20250425131030_ffiec_aws.csv`

api/src/backend/tasks/tasks.py
AdriiiPRodri: This function is to reduce as much as possible the code duplication. This function is in charge of return the correct writer and handled the initialisation of them

api/src/backend/tasks/jobs/export.py
AdriiiPRodri: It was separated to see more clearly what each one does but merging them into one can be done and I will explain how to use it in the docstring

api/src/backend/api/v1/views.py
AdriiiPRodri: This type of functions have been written to avoid duplicating code between endpoints and to delegate to each function only one specific action, so we can encapsulate the different actions

The functions returns content and filename but if something related with the file handling happens it will return it as this function is in charge of file handling

For the tests, more cases can be mocked, the function is being tested with the endpoints as part of them since they are a set. It is possible to make a specific test and mock the rest but I did not see it necessary since they are functions created specifically for the endpoints and to encapsulate actions

api/src/backend/api/v1/views.py
AdriiiPRodri: I set a 500 because at that point, if a ClientError really appears, the error is not on the part of the user, it is something internal to us. In my opinion the errors of type 400 are a problem of configuration or in the request, in short, a user problem. The errors 500 are our errors and should never happen but if they happen in this case it is controlled. We can discuss this and change it as we decide

api/pyproject.toml
AdriiiPRodri: Since it is very simple and this is easy to add and does not involve change of minor I thought that the quickest and easiest way is to enter here, it is to save time, if you see it convenient another PR will be created

api/CHANGELOG.md
AdriiiPRodri: Because they are two different things, on the one hand the compliance folder has been added to the ZIP and on the other hand an endpoint has been added, but both functionalities can exist without the need of the other one. This was a last minute requirement and compliance development would not be complete without it

jfagoagas: Thanks for the clarification.

```suggestion
- Added a `compliance/` folder and ZIP‐export functionality for all compliance reports.[(#7653)](https://github.com/prowler-cloud/prowler/pull/7653).
- Added a new API endpoint to fetch and download any specific compliance file by name [(#7653)](https://github.com/prowler-cloud/prowler/pull/7653).
```

Just a suggestion.

api/src/backend/api/v1/views.py
jfagoagas: I do not agree with you. A `ClientError` in `botocore`/`boto3` could not end up in an `HTTP 500 Internal Server Error` because is it a wrapper on top of different types of errors. If we are doing a `list_objects_v2` we should handle what's happening by:
- Logging the error
- Sending the exception to Sentry

We can maybe raise 500's for some cases but not for all the possible exceptions under `ClientError`.

api/pyproject.toml
jfagoagas: If we are sure that is thoroughly tested go ahead, but for the long-term and if something happens it is best to use an specific PR for this change. What's your opinion @vicferpoy?

prowler/lib/outputs/compliance/compliance_output.py
AdriiiPRodri: There is no tests for this file, this is tested in the specified compliance tests like `aws_well_architected_test.py`

api/src/backend/api/v1/views.py
AdriiiPRodri: I will change it but in my opinion even if boto fails it is still not the user's fault and in any case it is the server that has failed. At that point there is no way the user could have caused the crash

api/src/backend/tasks/tasks.py
AdriiiPRodri: In order to simplify this implies more code and more testing, the MAP with the compliance is a dictionary whose key is the provider_type and inside a list of tuples, where the first element is the condition and the second one is the class

Then what we do here is to iterate that list and check the condition, if the condition is True we use the class, if none is True we use the Generic. I think that the formatting in this case of ruff makes it difficult to read

api/src/backend/api/v1/views.py
AdriiiPRodri: What I mean is that at this point all errors are unexpected

AdriiiPRodri: Change applied

api/src/backend/tasks/tasks.py
AdriiiPRodri: If you want we can change to this version, it's the same:

```
klass = GenericCompliance
for condition, cls in COMPLIANCE_CLASS_MAP.get(provider_type, []):
        if condition(framework_name):
            klass = cls
            break
```

api/src/backend/api/v1/views.py
jfagoagas: In that case, do we need to keep `sentry_sdk.capture_exception`?

api/src/backend/tasks/tasks.py
jfagoagas: To me @vicferpoy has the final word on this PR.

api/pyproject.toml
vicferpoy: Agree @jfagoagas . But as you said, if it's already tested let's leave it here to not delay this development more. Let's keep it in mind in the future.

api/src/backend/api/v1/views.py
vicferpoy: Did you check `compliance.py` in the API @AdriiiPRodri ? To see if we can reuse any of that or adapt it somehow. It could also help with the compliance ID filters later on.

api/src/backend/api/v1/views.py
```diff
@@ -1150,6 +1152,28 @@ def destroy(self, request, *args, pk=None, **kwargs):
             404: OpenApiResponse(description="The scan has no reports"),
         },
     ),
+    compliance=extend_schema(
+        tags=["Scan"],
+        summary="Retrieve compliance report as CSV",
+        description="Download a specific compliance report (e.g., 'cis_1.4_aws') as a CSV file.",
+        operation_id="scan_compliance_download",
```
```python
operation_id="scan_compliance_download",
```

Are we also adding this field from now on? To keep it in mind. I believe it just serves a purpose for documentation.

api/src/backend/api/v1/views.py
```diff
@@ -1219,70 +1247,109 @@ def partial_update(self, request, *args, **kwargs):
...
+            status=status.HTTP_202_ACCEPTED,
+            headers={
+                "Content-Location": reverse(
+                    "task-detail", kwargs={"pk": serializer.data["id"]}
+                )
+            },
+        )
 
-        if scan_instance.output_location.startswith("s3://"):
+    def _load_file(self, path_pattern, s3=False, bucket=None, list_objects=False):
```
vicferpoy, line 1294: It is not a good practice to have this kind of functions defined at view level. Ideally, you work with specific exceptions and handle them in the view to return a specific HTTP response or keep going.

My two cents. Do as you wish.

api/src/backend/tasks/tasks.py
```diff
@@ -251,84 +255,106 @@ def generate_outputs(scan_id: str, provider_id: str, tenant_id: str):
         logger.info(f"No findings found for scan {scan_id}")
         return {"upload": False}
 
-    # Initialize the prowler provider
-    prowler_provider = initialize_prowler_provider(Provider.objects.get(id=provider_id))
+    provider_obj = Provider.objects.get(id=provider_id)
```
vicferpoy, line 258: Could we take this opportunity to wrap this code out of the task logic itself and move it to `export.py`? To improve readability throughout the defined tasks and also allow seamless migrations to another Task Runner framework if required in the future.

I know there are old tasks that should be migrated to that pattern as well, just asking.

vicferpoy (changes requested):
Great job with this. I left a few suggestions. Up to you to follow or discard them.

api/src/backend/tasks/tasks.py
vicferpoy: I believe Adri already changed this. I agree it's better this way. We're getting older and I would have had issues trying to understand that piece of code in a few months 🤣

api/src/backend/api/v1/views.py
AdriiiPRodri: For the moment I remove it and if we decide to add it we will do it at the same time in a PR

AdriiiPRodri:
I can't respond directly in the comment: https://github.com/prowler-cloud/prowler/pull/7653#discussion_r2075142943

I checked it, there is nothing faster than calling `get_available_compliance_frameworks`, moved this logic to this file and implemented a simple cache to only do this calculation once per provider.

api/src/backend/tasks/jobs/export.py
```diff
@@ -13,13 +13,86 @@
...
+    "kubernetes": [
+        (lambda name: name.startswith("cis_"), KubernetesCIS),
+        (lambda name: name.startswith("iso27001_"), KubernetesISO27001),
+    ],
+    "m365": [
+        (lambda name: name.startswith("cis_"), M365CIS),
+    ],
+    "nhn": [
+        (lambda name: name.startswith("iso27001_"), NHNISO27001),
+    ],
```
jfagoagas: Please remove `nhn` because it is not an official supported provider, so it is not present in the API.

api/src/backend/tasks/tasks.py
```diff
@@ -251,84 +255,106 @@ def generate_outputs(scan_id: str, provider_id: str, tenant_id: str):
...
+    compressed = _compress_output_files(out_dir)
+    upload_uri = _upload_to_s3(tenant_id, compressed, scan_id)
 
-    if uploaded:
-        # Remove the local files after upload
+    if upload_uri:
         try:
-            rmtree(Path(output_directory).parent, ignore_errors=True)
-        except FileNotFoundError as e:
+            rmtree(Path(compressed).parent, ignore_errors=True)
```
jfagoagas, line 351: Are we testing the files/path deletion? I'm not sure about but probably we are, just to double check.

api/src/backend/tasks/tasks.py
AdriiiPRodri: This implies a lot of changes in the tests and since we know that this module has to be refactored I think it is more appropriate to do it all in a PR. The tests added in this PR have been done because we have no date for this refactor

api/src/backend/api/v1/views.py
jfagoagas: Thanks for the clarification.

api/src/backend/api/v1/views.py
jfagoagas: @vicferpoy is this related with my previous comment? If so I'd say to leave it as we want to follow the practices we want to do in the API.

AdriiiPRodri: Right now it takes time to change this because exception responses are also handled, so they are also common to both endpoints. I didn't know the point that in Django having helper functions inside the view was a bad practice, the simplest and less time consuming solution is to repeat some code, other solutions will involve more testing and test changes

api/src/backend/tasks/tasks.py
AdriiiPRodri: Yes, we are testing it, I change the test to be more specified

api/src/backend/tasks/jobs/export.py
AdriiiPRodri: Removed

api/src/backend/api/v1/views.py
jfagoagas: Thanks, I know that we need to improve the way we handle this things, but I'm not sure about it.

api/src/backend/tasks/tasks.py
jfagoagas: Totally agree, I'm getting older.

api/src/backend/tasks/tasks.py
jfagoagas: Could you create a ticket for it?

api/src/backend/api/v1/views.py
jfagoagas: Could you create a ticket for it?

api/src/backend/api/v1/views.py
jfagoagas: It was restored.

api/src/backend/api/v1/views.py
jfagoagas: Is this already reviewed? It is.

jfagoagas (approved):
Finally 🥇 Thank you so much for the effort 👏

vicferpoy (approved):

PR #7655, opened 2025-05-02
fix: error about page number persistence when filters change

alejandrobailo:
### Context

When you filter in any view, you will get new results. What is happening is that if you are on a page other than the first one, that page could not exist for those new filters.

### Description

We have many ways to handle this scenario, I chose to create a custom hook to handle filter interactions. In this way, we avoid duplicating code and we have a centralized logic to handle all views with filters.
The new custom hook can update and clear filters, always setting the page to the first one when any of these actions are triggered.

### Checklist

#### API

### License


paabloLC (approved):
LGTM

PR #7661, opened 2025-05-05
feat(defender): add new check `defender_malware_policy_comprehensive_attachments_filter_applied`

danibarranqueroo:
### Context

This PR introduces a new check for Defender in M365. This check verifies that the Common Attachment Types Filter is enabled in Microsoft Defender for Office 365 and that it applies a comprehensive list of blocked file types. The check is configurable via the audit_config and uses a default list of Microsoft-recommended extensions unless explicitly overridden (blocking file types is something that depends a lot on the organization so that is why we have the default list and it can be changed with whatever extension is needed).

It evaluates whether:

- The EnableFileFilter flag is set to true.
- The anti-malware policy is enabled (either via an active rule or if it's the default policy).
- All file types defined in recommended_blocked_file_types (from audit_config) are present in the policy.

If any required extensions are missing or the policy is not enabled, the check fails.
This allows organizations to enforce strict or customized filtering policies based on their specific risk models.

### Description

### Checklist

#### API

### License


prowler/providers/m365/services/defender/defender_malware_policy_comprehensive_attachments_filter_applied/defender_malware_policy_comprehensive_attachments_filter_applied.py
```diff
@@ -0,0 +1,142 @@
...
+            findings.append(report)
+        else:
+            for policy in defender_client.malware_policies:
+                report = CheckReportM365(
+                    metadata=self.metadata(),
+                    resource=policy,
+                    resource_name="Defender Malware Policy",
+                    resource_id="defenderMalwarePolicy",
+                )
+                report.status = "PASS"
```
MrCloudSec: I think we should set the default status as FAIL.

andoniaf (approved):
🙌

PR #7665, opened 2025-05-06
feat(compliance): add a button to download the report in compliance card

paabloLC:
### Description

- This PR adds a download button to the compliance card component, allowing users to download the compliance report as a CSV.  
- The button is disabled when a region filter is applied in the URL to prevent partial or scoped downloads.

image

### Checklist

#### API

### License


ui/app/(prowler)/compliance/page.tsx
```diff
@@ -170,6 +171,8 @@ const SSRComplianceGrid = async ({
             totalRequirements={total}
             prevPassingRequirements={passed}
             prevTotalRequirements={total}
+            scanId={scanId}
+            complianceId={compliance_id}
```
alejandrobailo, line 175: know python uses snake_case and typescript uses camelCase but in the future it would be nice to see everything in camelCase, we could have a mapper that runs on the actions.

ui/components/ui/download-icon-button/download-icon-button.tsx
```diff
@@ -0,0 +1,40 @@
...
+  ariaLabel?: string;
+  isDisabled?: boolean;
+  textTooltip?: string;
+}
+
+export const DownloadIconButton = ({
+  paramId,
+  onDownload,
+  ariaLabel = "Download report",
+  isDisabled = false,
```
alejandrobailo: In CustomButton isDisabled is already false by default so we don't need to set it here as default.

alejandrobailo (changes requested):
Good job, I was wondering why we need to serialize the data, but yeah thats the way to manage binaries in server actions 👍

ui/app/(prowler)/compliance/page.tsx
paabloLC: exactly, it's for consistency with the API

alejandrobailo (approved):

PR #7690, opened 2025-05-08
feat(findings): Improve performance on /findings/metadata, /overviews and filters

vicferpoy:
### Context

Fix PRWLR-6427.
Fix PRWLR-6593.

### Description

- Resource data denormalization in findings table (resource_types, services, regions) as lists, speeding /findings with resource filters. This will scale when we start linking multiple resources with the same finding.

- Resource scan summary aggregations table to store the different resource types, services and regions for each scan, speeding `/findings/metadata`.

- Proper indexes for each of these implementations, trying with different variations and implementing ordering in them. In the end, it was not worth given the amount of possibilities we could have with different filters and ordering fields.

- Fallback and backfill mechanism for our current scans when requesting /findings/metadata.

  - Using `filter[scan]`: the previous logic will be used (probably taking a lot of time if the scan is big enough) and a backfill task will be sent. The next time this scan is asked for its metadata, it will use the aggregations table to perform faster.

  - Using `filter[inserted_at]`: similar behavior, but checking if there’s at least one entry of the aggregations table to trigger the mechanism or not.

  - After 48 hours of the deployment, everything should work fine without any kind of mechanism unless the user wants to check a particular older scan.

The aggregations table is scan based. This means that we cannot filter out values from the dynamic filters applying findings filters. In other words, filtering by `filter[delta]` for instance, will not change the values of the `/findings/metadata` endpoint, but `filter[region]` will. This is a tradeoff for performance that the team agreed on.

image

image

image

image

> [!NOTE]
> Some database optimizations have also been applied to `/overviews` endpoints.

### Checklist

#### API

### License


jfagoagas (approved):
Reviewed IRL, this is a huge step forward to improve performance 🚀 

Great job 🏅

PR #7743, opened 2025-05-14
feat(findings): Add /findings/latest and /findings/metadata/latest endpoints

vicferpoy:
### Context

We needed a way to retrieve just the latest findings in the system. The definition of `latest` comes from "findings from the latest scan for each provider".

### Description

Two new endpoints have been added to fulfill this requirement. These endpoints allow the same filters as the original versions except date and scan filters.

Some manual tests have been performed in our DEV environment to validate this, in addition to the automated unit tests.

image

image

### Checklist

#### API

### License


AdriiiPRodri (changes requested):
There are a few lines not covered, mainly some conditions in the views. I’ll leave it up to you whether you want to cover them with a test or if you think it’s good enough to move forward

vicferpoy:
@AdriiiPRodri Logic has not changed. It's just a copy of the original endpoints with a given filter for scan_ids using the latest ones. That's why I only added tests for that. Filter wise, we have a lot of tests in the original endpoints so we should be fine.

That's also why I could refactor the `list` and `latest` endpoints into a single function. We will do the same once we get rid of the backfill logic for `metadata` and `metadata/latest`.

AdriiiPRodri (dismissed):

AdriiiPRodri (approved):

PR #7752, opened 2025-05-15
feat(repository): add new check `repository_has_codeowners_file`

andoniaf:
### Context
Code owners are trusted users responsible for reviewing and managing critical code or configurations. It is highly recommended for organizations to assign code owners to any extremely sensitive code or configuration. By doing so, these trusted individuals will automatically be notified of any changes made to this code and be added as reviewers for pull requests. This ensures that only qualified individuals are reviewing and approving changes to sensitive parts of the codebase, helping to prevent malicious or unintentional alterations that could compromise the security or integrity of the system.

### Description
- Adds check `repository_has_codeowners_file` that ensures that repositories have a CODEOWNERS file.

### Checklist

#### API

### License


MrCloudSec (changes requested):
What do you think about changing the check id to `repository_codeowners_file_present`?

prowler/CHANGELOG.md
MrCloudSec, line 1: Please, update the changelog to the latest version.

prowler/providers/github/services/repository/repository_service.py
```diff
@@ -11,29 +11,44 @@ def __init__(self, provider):
...
+                    codeowners_paths = [
+                        ".github/CODEOWNERS",
+                        "CODEOWNERS",
+                        "docs/CODEOWNERS",
+                    ]
+                    codeowners_exists = False
+                    for path in codeowners_paths:
+                        if self._file_exists(repo, path):
+                            codeowners_exists = True
+                            break
```
MrCloudSec: We are not covering when `self._file_exists is None`.

andoniaf:
> What do you think about changing the check id to `repository_codeowners_file_present`?

No strong opinion here, I like both. But if we change it we will need to update also `repository_public_has_securitymd_file`.

prowler/providers/github/services/repository/repository_service.py
andoniaf: `None` is considered `false`, and `codeowners_exists` default value is `False` so it will keep it false.

MrCloudSec: But remember that we need to keep the `None` value so the check does not run.

MrCloudSec:
> repository_public_has_securitymd_file

I did not remember about the `repository_public_has_securitymd_file ` check, so we can keep this name 😄

prowler/providers/github/services/repository/repository_service.py
andoniaf: You're right! No info, no finding!

MrCloudSec (approved):

PR #7811, opened 2025-05-21
feat(mutelist): make validate_mutelist method static 

pedrooot:
### Context

If fixes an issue please add it with `Fix #XXXX`

### Description

This PR moves the `validate_mutelist` method outside the `Mutelist` class.

### Checklist

#### API

### License


prowler/lib/mutelist/mutelist.py
```diff
@@ -449,3 +437,23 @@ def is_item_matched(matched_items, finding_items, tag=False) -> bool:
...
+        Returns:
+            tuple[bool, dict]: A tuple containing a boolean indicating if the mutelist is valid and the mutelist itself.
+        """
+        try:
+            validate(mutelist, schema=mutelist_schema)
+            return True, mutelist
+        except Exception as error:
+            logger.error(
+                f"{error.__class__.__name__} -- Mutelist YAML is malformed - {error}[{error.__traceback__.tb_lineno}]"
+            )
```
AdriiiPRodri, line 458: I know that actually the sdk has this kind of patten but I think that we should be more specific with the exceptions

prowler/lib/mutelist/mutelist.py
```diff
@@ -449,3 +437,23 @@ def is_item_matched(matched_items, finding_items, tag=False) -> bool:
...
+
+        Args:
+            mutelist (dict): The mutelist to be validated.
+
+        Returns:
+            tuple[bool, dict]: A tuple containing a boolean indicating if the mutelist is valid and the mutelist itself.
+        """
+        try:
+            validate(mutelist, schema=mutelist_schema)
+            return True, mutelist
```
AdriiiPRodri: Why we need to return a boolean and a dict and not only a dict? It is a requirement?

AdriiiPRodri (changes requested):

prowler/lib/mutelist/mutelist.py
pedrooot: Screenshot 2025-06-10 at 17 07 39
this was from this assert, I didn't want to change the logic of the validation at all.

prowler/lib/mutelist/mutelist.py
```diff
@@ -119,7 +118,7 @@ def __init__(
             self._mutelist = mutelist_content
 
         if self._mutelist:
-            self.validate_mutelist()
+            self._mutelist = self.validate_mutelist(self._mutelist)[1]
```
AdriiiPRodri: This way is more Pythonic and we clarify that is a static utility 
```suggestion
            self._mutelist = Mutelist.validate_mutelist(self._mutelist)[1]
```

AdriiiPRodri (changes requested):

prowler/lib/mutelist/mutelist.py
pedrooot: Done ✅

AdriiiPRodri (approved):

PR #7822, opened 2025-05-22
feat(sso): add sso with saml to API

AdriiiPRodri:
### Description

In this PR we have added the SSO with SAML login. **The documentation is in process, it's not definitive because we don't have the UI part. It will be updated when we have everything ready.**

### Checklist

#### API

### License


drewkerrigan (dismissed):
Awesome work @AdriiiPRodri ! The main thing I was looking for was efficient and correct handling of per-tenant IdP configuration, and it looks like this should work perfectly.

Thanks!

api/pyproject.toml
```diff
@@ -8,7 +8,7 @@ dependencies = [
   "celery[pytest] (>=5.4.0,<6.0.0)",
   "dj-rest-auth[with_social,jwt] (==7.0.1)",
   "django==5.1.8",
-  "django-allauth==65.4.1",
+  "django-allauth[saml] (>=65.8.0,<66.0.0)",
```
vicferpoy, line 11: We need to make sure the social login still works as intended because we had hard dependencies between the `dj-rest-auth` library and `django-allauth`.

api/src/backend/api/adapters.py
```diff
@@ -29,33 +38,86 @@ def save_user(self, request, sociallogin, form=None):
...
-                    role=role,
-                    tenant_id=tenant.id,
+
+                with rls_transaction(tenant.id):
+                    role_name = extra.get("userType", ["saml_default_role"])[0].strip()
+
+                    try:
+                        role = Role.objects.using(MainRouter.admin_db).get(
+                            name=role_name, tenant_id=tenant.id
+                        )
```
vicferpoy, line 65: I'm curious about this. How is the SAML role related to our RBAC?

api/src/backend/api/models.py
```diff
@@ -1284,6 +1288,219 @@ class Meta:
         ]
 
 
+class SAMLDomainIndex(models.Model):
+    """
+    Public index of SAML domains. No RLS. Used for fast lookup in SAML login flow.
+    """
+
+    email_domain = models.CharField(max_length=254, unique=True)
```
vicferpoy, line 1374: Why can't we use the `email_domain` from the SAML configurations? With an index it shouldn't take long to look it up.

api/src/backend/api/models.py
```diff
@@ -1284,6 +1288,219 @@ class Meta:
...
+        help_text="Email domain used to identify the tenant, e.g. prowlerdemo.com",
+    )
+    metadata_xml = models.TextField(
+        help_text="Raw IdP metadata XML to configure SingleSignOnService, certificates, etc."
+    )
+    created_at = models.DateTimeField(auto_now_add=True)
+    updated_at = models.DateTimeField(auto_now=True)
+
+    class JSONAPIMeta:
+        resource_name = "saml-configuration"
```
vicferpoy:
```suggestion
        resource_name = "saml-configurations"
```

api/src/backend/api/models.py
```diff
@@ -1284,6 +1288,219 @@ class Meta:
...
+                name="unique_resources_by_email_domain",
+            ),
+            BaseSecurityConstraint(
+                name="statements_on_%(class)s",
+                statements=["SELECT", "INSERT", "UPDATE", "DELETE"],
+            ),
+        ]
+
+
+class SAMLConfigurations(RowLevelSecurityProtectedModel):
```
vicferpoy:
```suggestion
class SAMLConfiguration(RowLevelSecurityProtectedModel):
```

api/src/backend/api/models.py
```diff
@@ -1284,6 +1288,219 @@ class Meta:
...
+                statements=["SELECT", "INSERT", "UPDATE", "DELETE"],
+            ),
+            # 1 config per tenant
+            models.UniqueConstraint(
+                fields=["tenant"],
+                name="unique_samlconfig_per_tenant",
+            ),
+        ]
+
+    def clean(self, old_email_domain=None):
```
vicferpoy, line 1446: Any reason on why we moved from serializer validation to model validation for this particular model?

api/src/backend/api/models.py
```diff
@@ -1284,6 +1288,219 @@ class Meta:
...
+        x509cert = cert.text.strip()
+
+        return {
+            "entity_id": entity_id,
+            "sso_url": sso_url,
+            "slo_url": slo_url,
+            "x509cert": x509cert,
+        }
+
+    def _sync_social_app(self, previous_email_domain=None):
```
vicferpoy, line 1552: Just a question: this is required for the following flow:

Django models -> SocialApp

Because the SAML integration will exist in our models first, then in the Social App so we have to handle it manually here. Did I understand that correctly?

api/src/backend/api/tests/test_models.py
```diff
@@ -1,6 +1,30 @@
...
+from api.db_router import MainRouter
+from api.models import (
+    Resource,
+    ResourceTag,
+    SAMLConfigurations,
+    SAMLDomainIndex,
+    Tenant,
+)
+
+VALID_METADATA = """<?xml version='1.0' encoding='UTF-8'?>
```
vicferpoy: Could we move this within the TestSAMLConfigurationModel class? It should only be used there, right?

api/src/backend/api/v1/views.py
```diff
@@ -377,6 +385,163 @@ def get_response(self):
...
+    def post(self, request, *args, **kwargs):
+        serializer = self.get_serializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        email = serializer.validated_data["email_domain"]
+        domain = email.split("@", 1)[-1].lower()
+
+        try:
+            check = SAMLDomainIndex.objects.get(email_domain=domain)
+            with rls_transaction(str(check.tenant_id)):
+                config = SAMLConfigurations.objects.get(tenant_id=str(check.tenant_id))
```
vicferpoy: Again, I don't get the point on SAMLDomainIndex. RLS protected models create an index on `tenant_id` already so the lookup should be equally fast.

api/src/backend/api/v1/views.py
```diff
@@ -377,6 +385,163 @@ def get_response(self):
...
+    """
+
+    serializer_class = SAMLConfigurationsSerializer
+    required_permissions = [Permissions.MANAGE_INTEGRATIONS]
+    queryset = SAMLConfigurations.objects.all()
+
+    def get_queryset(self):
+        # If called during schema generation, return an empty queryset
+        if getattr(self, "swagger_fake_view", False):
+            return User.objects.none()
```
```suggestion
            return SAMLConfiguration.objects.none()
```

Notice I already removed the 's' here.

api/src/backend/api/v1/views.py
```diff
@@ -377,6 +385,163 @@ def get_response(self):
...
+        return SAMLConfigurations.objects.filter(tenant=self.request.tenant_id)
+
+
+class TenantFinishACSView(FinishACSView):
+    def dispatch(self, request, organization_slug):
```
vicferpoy, line 485: Why overriding the dispatch?

api/src/backend/api/v1/views.py
```diff
@@ -377,6 +385,163 @@ def get_response(self):
...
+            )
+        UserRoleRelationship.objects.using(MainRouter.admin_db).filter(
+            user=user,
+            tenant_id=tenant.id,
+        ).delete()
+        UserRoleRelationship.objects.using(MainRouter.admin_db).create(
+            user=user,
+            role=role,
+            tenant_id=tenant.id,
+        )
```
vicferpoy: Isn't this covered in the adapter?

vicferpoy (changes requested):
I haven't tested this locally yet, but I have some suggestions and concerns. Could you check if the Social Login works after your changes to Allauth settings?

api/src/backend/api/models.py
vicferpoy: https://www.django-antipatterns.com/antipattern/plural-model-class-names.html

api/src/backend/api/v1/urls.py
```diff
@@ -112,6 +116,17 @@
         ),
         name="provider_group-providers-relationship",
     ),
+    # API endpoint to start SAML SSO flow
+    path(
+        "auth/saml/initiate/", SAMLInitiateAPIView.as_view(), name="api_saml_initiate"
+    ),
+    # Allauth SAML endpoints for tenants
+    path("accounts/", include("allauth.urls")),
```
vicferpoy, line 124: "accounts" is too vague in my opinion. It could reference to system accounts and it is not the case. Do you think we could do something about it?

Also, does this include more URLs rather than SAML's? Are they shown in the documentation?

vicferpoy (changes requested):

api/src/backend/api/v1/urls.py
AdriiiPRodri: The account prefix is from the Django Allauth documentation, but we can change it to something else if needed. No, the URLs are not shown in the documentation because these endpoints are not intended for end-user access

api/src/backend/api/v1/views.py
AdriiiPRodri: We need to override this method because the SAML implementation in Allauth doesn’t support multitenancy or role management. Additionally, it doesn’t use the extra fields to populate user information in our tables

api/src/backend/api/v1/views.py
AdriiiPRodri: Aquí tienes la traducción al inglés natural:

Yes, it’s covered, but the adapter is only used when the user is created. With SAML, user attributes (including roles) can be modified from the IdP, so we need to update their permissions and information based on any changes made in the IdP

AdriiiPRodri: What we could do is extract a common part for role creation, but I thought it was more understandable this way, and it doesn’t really save us much code either

api/src/backend/api/v1/views.py
AdriiiPRodri: We need an endpoint where the user provides their email (they’re not authenticated at this point), so we need a table to look up this information. For security reasons, I preferred to keep only the minimal data in the index to avoid having to use admin access to the configuration table and navigate through the different tenants

api/src/backend/api/models.py
AdriiiPRodri: Yes, and also because SocialApp doesn’t support multitenancy

api/src/backend/api/models.py
AdriiiPRodri: https://github.com/prowler-cloud/prowler/pull/7822#discussion_r2107516259

api/src/backend/api/adapters.py
AdriiiPRodri: It is related with the userType, as in the SaaS

api/pyproject.toml
AdriiiPRodri: I checked it at the beginning of the development and it worked and I will check it again to be sure

.env
```diff
@@ -133,3 +133,7 @@ SOCIAL_GOOGLE_OAUTH_CLIENT_SECRET=""
 SOCIAL_GITHUB_OAUTH_CALLBACK_URL="${AUTH_URL}/api/auth/callback/github"
 SOCIAL_GITHUB_OAUTH_CLIENT_ID=""
 SOCIAL_GITHUB_OAUTH_CLIENT_SECRET=""
+
+# Single Sign-On (SSO)
+SAML_PUBLIC_CERT=""
+SAML_PRIVATE_KEY=""
```
jfagoagas, line 143: Why public cert and private key? I think it should be either private/public key or certificate and private key.

.github/workflows/api-pull-request.yml
```diff
@@ -113,6 +113,12 @@ jobs:
           python-version: ${{ matrix.python-version }}
           cache: "poetry"
 
+      - name: Install system dependencies for xmlsec
+        if: steps.are-non-ignored-files-changed.outputs.any_changed == 'true'
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y libxml2-dev libxmlsec1-dev libxmlsec1-openssl pkg-config
```
jfagoagas, line 120: Out of curiosity, why do you need to install all these in the workflow system?

api/Dockerfile
```diff
@@ -6,7 +6,19 @@ ARG POWERSHELL_VERSION=7.5.0
 ENV POWERSHELL_VERSION=${POWERSHELL_VERSION}
 
 # hadolint ignore=DL3008
-RUN apt-get update && apt-get install -y --no-install-recommends wget libicu72 \
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    wget \
```
jfagoagas, line 10: are all these strictly required?

api/src/backend/api/adapters.py
jfagoagas: Are we creating a new role for SAML?

api/src/backend/api/adapters.py
```diff
@@ -29,33 +38,86 @@ def save_user(self, request, sociallogin, form=None):
...
+                    )
+
+                    UserRoleRelationship.objects.using(MainRouter.admin_db).create(
+                        user=user,
+                        role=role,
+                        tenant_id=tenant.id,
+                    )
+
+            else:
+                # Handle other providers (e.g., GitHub, Google)
```
jfagoagas, line 92: I know this was created before but since we are modifying the existing code and it touches critical parts of the system I think we need to work on adding tests for lines 92-117.

api/src/backend/api/migrations/0026_samlconfigurations.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("api", "0025_findings_uid_index_parent"),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name="SAMLDomainIndex",
```
jfagoagas: Why the table has `index` in its name? Why not to have something like `saml_domains` or using the other table for it and have an index there.

jfagoagas: After reading `SAMLConfigurations`, are we creating `SAMLDomainIndex` just to have the domain and the tenant?

api/src/backend/api/migrations/0026_samlconfigurations.py
```diff
@@ -0,0 +1,120 @@
...
+            name="integration_type",
+            field=api.db_utils.IntegrationTypeEnumField(
+                choices=[
+                    ("amazon_s3", "Amazon S3"),
+                    ("aws_security_hub", "AWS Security Hub"),
+                    ("jira", "JIRA"),
+                    ("slack", "Slack"),
+                ]
+            ),
+        ),
```
jfagoagas: Why are we adding new items in the integrations enum?

api/src/backend/api/migrations/0026_samlconfigurations.py
```diff
@@ -0,0 +1,120 @@
...
+                "tenant_id",
+                name="rls_on_samlconfigurations",
+                statements=["SELECT", "INSERT", "UPDATE", "DELETE"],
+            ),
+        ),
+        migrations.AddConstraint(
+            model_name="samlconfigurations",
+            constraint=models.UniqueConstraint(
+                fields=("tenant",), name="unique_samlconfig_per_tenant"
+            ),
```
jfagoagas: If the `email_domain` and `config` need to be unique per tenant, why not to create a unique index with `tenant, email_domain, config`?

api/src/backend/config/django/base.py
```diff
@@ -10,6 +10,7 @@
 SECRET_KEY = env("SECRET_KEY", default="secret")
 DEBUG = env.bool("DJANGO_DEBUG", default=False)
 ALLOWED_HOSTS = ["localhost", "127.0.0.1"]
+SECURE_PROXY_SSL_HEADER = ("HTTP_X_FORWARDED_PROTO", "https")
```
jfagoagas, line 13: Is this required for the app or just for testing?

docs/tutorials/prowler-app-sso.md
jfagoagas, line 1: Review sections headers I think there is a mix of `#` and `##`.

docs/tutorials/prowler-app-sso.md
```diff
@@ -0,0 +1,89 @@
...
+Update this variable to specify which domains Django should accept incoming requests from. This typically includes:
+
+- `localhost` for local development
+- container hostnames (e.g. `prowler-api`)
+- public-facing domains or tunnels (e.g. ngrok)
+
+**Example**:
+
+```env
+DJANGO_ALLOWED_HOSTS=localhost,127.0.0.1,prowler-api,mycompany.prowler
```
jfagoagas, line 20: Does this work if the value is `*`?

docs/tutorials/prowler-app-sso.md
```diff
@@ -0,0 +1,89 @@
...
+
+
+???+ note "API Note"
+    SSO with SAML API documentation.[Prowler API Reference - Upload SAML configuration](https://api.prowler.com/api/v1/docs#tag/SAML/operation/saml_config_create)
+
+# SAML Initiate
+
+### Description
+
+This endpoint receives an email and checks if there is an active SAML configuration for the associated domain (i.e., the part after the @). If a configuration exists and the required certificates are present, it responds with an HTTP 302 redirect to the appropriate saml_login endpoint for the organization.
```
jfagoagas, line 63: We need to always redirect to the SAML endpoint, if not we will allow domain enumeration if responses are different based on existence.

jfagoagas (changes requested):
That's great @AdriiiPRodri 👏 I still need to setup the environment to test it but I did a first pass to the code. 

Thanks!

docs/tutorials/prowler-app-sso.md
AdriiiPRodri: The endpoint from the library itself returns a 404 if the domain doesn’t exist, and a 200 along with the login page if it does. There’s no way to prevent enumeration if we always return a 200, the response body will still be different depending on whether the configuration exists or not

jfagoagas: What's in the response body? We should find a way to prevent or mitigate it.

docs/tutorials/prowler-app-sso.md
AdriiiPRodri: It seems like it’s possible, although not recommended

docs/tutorials/prowler-app-sso.md
AdriiiPRodri: The documentation is only meant to facilitate testing, it’s not the final version since we’re still missing the UI

api/src/backend/config/django/base.py
AdriiiPRodri: It’s required, this propagates the SSL header when Django is behind a proxy, since all IdPs expect that header to be present

docs/tutorials/prowler-app-sso.md
jfagoagas: I mean if SAML will work having `DJANGO_ALLOWED_HOSTS=*` because that's the only way as of today to configure it if running behind an ALB with dynamic IPs.

api/src/backend/config/django/base.py
jfagoagas: Great, good to know, thanks!

api/src/backend/api/migrations/0026_samlconfigurations.py
AdriiiPRodri: This does the same thing, right? What we’re saying here is that the tenant value should be unique, and since each tenant has its own configuration, the result is the same

api/src/backend/api/migrations/0026_samlconfigurations.py
AdriiiPRodri: We’re removing SAML from the integrations, as agreed. To remove it from an enum, we need to redeclare the enum, which is why this is an AlterField

api/src/backend/api/migrations/0026_samlconfigurations.py
AdriiiPRodri: It’s called an index because the idea is similar to an index in a text document. That said, this table exists to support the use of the initiate endpoint from the UI, but I’m considering removing that endpoint, and therefore the table as well

api/src/backend/api/migrations/0026_samlconfigurations.py
jfagoagas: That's totally right, I need more coffee 😅

AdriiiPRodri:
https://github.com/prowler-cloud/prowler/pull/7822#discussion_r2108336184

Yes, as in the SaaS:

> Note: The userType attribute will be used to assign the user's role. If the role does not exist, one will be created with minimal permissions. You can assign permissions to roles on the [Roles](https://app.prowler.pro/app/roles) page.

.env
AdriiiPRodri: You’re right to question the naming, in this context, we actually need the private key for signing SAML requests/responses, and the X.509 certificate (not just the public key) so that the identity provider (IdP) can verify those signatures. The certificate includes the public key along with metadata and a CA signature, which is required by SAML specs

.github/workflows/api-pull-request.yml
AdriiiPRodri: Due to: https://github.com/xmlsec/python-xmlsec/issues/320, we need to reinstall after the other dependencies and it should be the no-binary version because its need to be compilated in our system

api/Dockerfile
AdriiiPRodri: Yes, these packages are necessary for the new Python dependencies

api/src/backend/api/adapters.py
AdriiiPRodri: The change was just wrapping them in an else. If I have time, I’ll write the tests, but right now I need to unblock the development of the integrations

docs/tutorials/prowler-app-sso.md
AdriiiPRodri: An HTML page with a CSRF token the only thing I can think of is to always generate the HTML with a fake CSRF token that’s indistinguishable from a real one. There’s no standard mechanism for this since the responsibility is left to the IdP. If we want to mitigate it on our side, this is the only approach I see as feasible

docs/tutorials/prowler-app-sso.md
AdriiiPRodri: But this isn’t something new introduced by SSO, it already existed and will have to remain as-is in the Cloud environment. Nothing has changed in that regard

.env
jfagoagas: In that case, why not to rename it to `SAML_CERTIFICATE`?

.github/workflows/api-pull-request.yml
jfagoagas: Interesting, can you create a TODO and a ticket to review this in the future once the issue is addressed?

api/Dockerfile
jfagoagas: Thanks!

api/src/backend/api/migrations/0026_samlconfigurations.py
jfagoagas: Interesting, let's talk about it once you decide what to do.


_2025/06/09 - I think this comment is still pending_

api/src/backend/api/migrations/0026_samlconfigurations.py
jfagoagas: Thanks for the clarification, makes sense.

docs/tutorials/prowler-app-sso.md
jfagoagas: Perfect, I was just curious about it.

.env
AdriiiPRodri: We can use whatever name you prefer, it’s called that in other projects. I’m indicating that it’s public so it’s clear that the value is meant to be public and can be accessed, to make sure you shouldn’t put your private key in that variable

.github/workflows/api-pull-request.yml
AdriiiPRodri: Sure!

.env
jfagoagas: I'd call it either `SAML_X509_CERT` or `SAML_CERT`, to me adding public there is confusing.

AdriiiPRodri: Maybe we didn’t understand each other correctly, I’m not quite sure I follow why it’s considered confusing. The certificate is public, that’s its whole purpose by definition. Yes, it’s an X.509, but it’s meant to be publicly accessible

api/src/backend/api/adapters.py
drewkerrigan: You mean are we creating a new permission? In the legacy SaaS we just use the manage_integrations permission to cover SAML configuration.

.env
jfagoagas: It's not confusing but maybe I'm used to call it differently.

api/src/backend/api/adapters.py
AdriiiPRodri: I create a new role just in case the user has an invalid configuration for the userType field, if the role doesn't exists then a new role is get or create for the tenant and assign to that user

api/src/backend/api/models.py
AdriiiPRodri: In this case I though that is more clear that the sanitization should be here but can be cahnged, I can't remember if it is a hard condition for this

api/pyproject.toml
vicferpoy: I will assume this was tested again and everything was correct. Thank you.

api/src/backend/api/adapters.py
vicferpoy: This is still confusing to me, but I'm not familiar with SAML. If you guys believe this is correct, please resolve this conversation. You can ELI5 later on 😄

AdriiiPRodri:
## Review of Reported Cursor Bot Bugs

All the issues reported by the bot have been thoroughly tested and analyzed. After careful validation and test, we conclude that **none of the reported bugs are valid**. Below is a breakdown of each point:

### 1. Bug: SAML Configuration Errors and Inconsistent Parameters

This is not a valid bug. To reach save_user, the SAML login flow must be initiated, which necessarily requires a valid SAML configuration, otherwise, there would be no matching domain or endpoint.
Additionally, the userType field is either None (in which case a default list is assigned), or it is provided by the user and always arrives as a list, even when a single character is passed.

### 2. Bug: TenantFinishACSView SAML Handling Issues

To reach the dispatch method, the SAML login process must have already started, which guarantees that a configuration exists. Otherwise, a 404 would be returned.
Regarding attribute handling: SAML IdPs always send attributes as lists, this is defined by the SAML standard.
As for the reference to tenant vs tenant.id, both are valid and result in the same behavior in the context used.

### 3. Bug: SAML Data Access Assumes List Structure

Again, this follows the SAML specification. Attribute values are always sent as lists.
If an attribute is missing, we fall back to a list containing an empty string. Therefore, an IndexError cannot occur in this logic.

### 4. Bug: Ambiguous Query Causes MultipleObjectsReturned Error

This is prevented by an explicit uniqueness constraint at the database level:
```
{
    "errors": [
        {
            "detail": "saml configuration with this email domain already exists.",
            "status": "400",
            "source": {
                "pointer": "/data/attributes/email_domain"
            },
            "code": "unique"
        }
    ]
}
```

### 5. Bug: SAML Config Handling & Tenant ID Type Mismatch

If execution reaches the relevant logic, it means a valid SAML configuration exists. There is no execution path that allows reaching that point without one.

### 6. Bug: SAML Authentication Completion Bugs

Again, no IndexError can occur. The expected data will always be present. Roles are not duplicated, so using .get() is safe and reliable.

### 7. Bug: SAML Configuration Validation Fails During Updates

Duplicate domains cannot occur due to a uniqueness constraint.
Additionally, the current instance is properly excluded during update operations:

```
# Exclude ourselves in case of update
if self.pk:
    qs = qs.exclude(pk=self.pk)
```

### 8. Bug: SAML Configuration Errors and Data Handling Issues

This is a repetition of the earlier concerns regarding attribute indexing and configuration existence, which have already been addressed.

### 9. Bug: SAML Configuration and Data Handling Errors

Same as above, already covered and invalid.

vicferpoy (approved):

jfagoagas (approved):

PR #7824, opened 2025-05-22
feat(api): support GCP Service Account key

MrCloudSec:
### Description

Support GCP Service Account key in the API.
image

### Checklist

#### API

### License


jfagoagas (changes requested):
We should not remove existing features in the API if we don't want to break existing user's flow. It is great to add new authentication mechanisms but we need to keep compatibility.

@MrCloudSec @vicferpoy @AdriiiPRodri

api/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ All notable changes to the **Prowler API** are documented in this file.
 
 ### Changed
 - Renamed field encrypted_password to password for M365 provider [(#7784)](https://github.com/prowler-cloud/prowler/pull/7784)
+- Support GCP Service Account key. [(#7824)](https://github.com/prowler-cloud/prowler/pull/7824)
```
vicferpoy: Would it make more sense to move this to `Added`? It's a new feature, right?

api/src/backend/api/models.py
```diff
@@ -850,6 +850,7 @@ class ProviderSecret(RowLevelSecurityProtectedModel):
     class TypeChoices(models.TextChoices):
         STATIC = "static", _("Key-value pairs")
         ROLE = "role", _("Role assumption")
+        SERVICE_ACCOUNT = "service_account", _("GCP Service Account Key")
```
vicferpoy, line 853: Missing migrations to update the Postgres enum.

api/src/backend/api/specs/v1.yaml
```diff
@@ -12603,4 +12603,4 @@ tags:
 - name: Integration
   description: Endpoints for managing third-party integrations, including registration,
     configuration, retrieval, and deletion of integrations such as S3, JIRA, or other
-    services.
+    services.
```
vicferpoy:
```suggestion
    services.
 
```

api/src/backend/api/v1/serializers.py
```diff
@@ -1159,6 +1159,8 @@ def validate_secret_based_on_provider(
                 )
         elif secret_type == ProviderSecret.TypeChoices.ROLE:
             serializer = AWSRoleAssumptionProviderSecret(data=secret)
+        elif secret_type == ProviderSecret.TypeChoices.SERVICE_ACCOUNT:
+            serializer = GCPServiceAccountProviderSecret(data=secret)
```
vicferpoy, line 1163: I wonder if this suits the `STATIC` secret type? Is it really necessary to add a new type for this kind of credentials? If it's just a string... Just asking.

api/src/backend/api/specs/v1.yaml
```diff
@@ -12603,4 +12603,4 @@ tags:
 - name: Integration
```
vicferpoy: Missing changes here. Please regenerate the API spec.

vicferpoy (changes requested):
image

Take care of this and check them when applied, please.

api/CHANGELOG.md
MrCloudSec: Yes, good catch!

api/src/backend/api/v1/serializers.py
MrCloudSec: I added a new type since both auth methods in GCP are static, so, if I am not wrong, we need a different type so we can differentiate them.

vicferpoy (approved):

AdriiiPRodri (approved):

PR #7771, opened 2025-05-19
feat(repository): add new check `repository_dependency_scanning_enabled`

andoniaf:
### Context
Implement scanning tools to detect, prevent, and monitor known open-source vulnerabilities in packages used within the organization's projects. Open-source vulnerabilities can exist before a package is adopted, but new vulnerabilities are regularly discovered over time. Monitoring and detecting whether the dependencies in use are affected by new vulnerabilities is critical. By addressing these vulnerabilities early, the organization reduces its attack surface and prevents security flaws from reaching production, which could otherwise lead to a security breach.

### Description
- Adds `repository_dependency_scanning_enabled` check to ensure that dependency/package vulnerability scanning (e.g., Dependabot alerts) is enabled for the repository.

### Checklist

- Are there new checks included in this PR? Yes / No
    - If so, do we need to update permissions for the provider? Please review this carefully.
- [ ] Review if the code is being covered by tests.
- [ ] Review if code is being documented following this specification https://github.com/google/styleguide/blob/gh-pages/pyguide.md#38-comments-and-docstrings
- [ ] Review if backport is needed.
- [ ] Review if is needed to change the [Readme.md](https://github.com/prowler-cloud/prowler/blob/master/README.md)
- [ ] Ensure new entries are added to [CHANGELOG.md](https://github.com/prowler-cloud/prowler/blob/master/prowler/CHANGELOG.md), if applicable.

#### API
- [ ] Verify if API specs need to be regenerated.
- [ ] Check if version updates are required (e.g., specs, Poetry, etc.).
- [ ] Ensure new entries are added to [CHANGELOG.md](https://github.com/prowler-cloud/prowler/blob/master/api/CHANGELOG.md), if applicable.

### License

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #7834, opened 2025-05-26
fix(github): repository `repository_dependency_scanning_enabled` check logic

andoniaf:
### Context
If the repository has dependabot alert enable but 0 results the check gets an `ERROR: IndexError[138]: list index out of range` and it does not return a finding because we do not handle this properly.

This check was added in https://github.com/prowler-cloud/prowler/pull/7771

### Description
- Use `.totalCount` instead, this will fail with 403 if not enabled and return a number (including 0) if enabled.
- Set the retry count for the Github provider to 3 to speed up scans. (Default was 10)

### Checklist

#### API

### License

### DIFF
--- a/prowler/providers/github/github_provider.py
+++ b/prowler/providers/github/github_provider.py
@@ -4,6 +4,7 @@ from typing import Union
 
 from colorama import Fore, Style
 from github import Auth, Github, GithubIntegration
+from github.GithubRetry import GithubRetry
 
 from prowler.config.config import (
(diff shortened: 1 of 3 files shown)


MrCloudSec (approved):
Great catch @andoniaf ⚾ Also, we need to create a flag like in https://docs.prowler.com/projects/prowler-open-source/en/latest/tutorials/aws/boto3-configuration/ to make the retry attempts configurable.

PR #7836, opened 2025-05-26
feat(organization): add new check `organization_members_inactive`

andoniaf:
### Context
Monitor and review user accounts that have been inactive for a significant period. Inactive accounts increase the risk of unauthorized access, particularly if they have elevated privileges. By removing or deactivating these accounts, organizations can reduce their exposure to potential attacks.

### Description
- Add `organization_members_inactive` check to ensure user accounts that have been inactive for a significant period are monitored and reviewed.

### Checklist

#### API

### License


MrCloudSec (changes requested):
@andoniaf I think we should not add this check since it only checks for public activity.

andoniaf:
> @andoniaf I think we should not add this check since it only checks for public activity.

100% agree. Would be worth if we find a different endpoint or an easy way in the future to check the activity info inside and organization.

Closing this.

PR #7848, opened 2025-05-27
feat(lighthouse): Add django endpoints to store config

0xbadshah:
### Context

Part of Lighthouse feature.

### Description

Creates lighthouse config endpoint `/api/v1/lighthouse-config` that allows creation, modification, deletion of lighthouse config.

### Checklist

#### API

API Screenshots

List all Lighthouse configurations

image

Create new Lighthouse config

image

Partially update Lighthouse config

image

Delete a config

image

Check OpenAI key

image

### License


api/src/backend/api/models.py
```diff
@@ -1332,3 +1335,160 @@ class Meta:
...
+    """
+
+    MODEL_CHOICES = [
+        "gpt-4o-2024-11-20",
+        "gpt-4o-2024-08-06",
+        "gpt-4o-2024-05-13",
+        "gpt-4o",
+        "gpt-4o-mini-2024-07-18",
+        "gpt-4o-mini",
+    ]
```
vicferpoy: This requires its own Postgres enum. Check `db_utils.py` for reference.

api/src/backend/api/models.py
```diff
@@ -1332,3 +1335,160 @@ class Meta:
...
+    )
+    api_key = models.BinaryField(
+        blank=False, null=False, help_text="Encrypted API key for the LLM service"
+    )
+    model = models.CharField(
+        max_length=50,
+        blank=False,
+        null=False,
+        help_text="Must be one of the supported model names",
+    )
```
vicferpoy: Change it to use the enum after you create it. It should be a choice.

api/src/backend/api/models.py
```diff
@@ -1332,3 +1335,160 @@ class Meta:
...
+                pointer="/data/attributes/max_tokens",
+            )
+
+        # Validate model name
+        if self.model not in self.MODEL_CHOICES:
+            raise ModelValidationError(
+                detail=f"Model must be one of: {', '.join(self.MODEL_CHOICES)}",
+                code="invalid_model",
+                pointer="/data/attributes/model",
+            )
```
vicferpoy: You can delete this after you configure the Postgres enum.

api/src/backend/api/models.py
```diff
@@ -1332,3 +1335,160 @@ class Meta:
...
+
+        if not self.model:
+            raise ModelValidationError(
+                detail="Model is required",
+                code="missing_model",
+                pointer="/data/attributes/model",
+            )
+
+        self.full_clean()
+        super().save(*args, **kwargs)
```
vicferpoy: You don't need to override `save` to do this. Specify that those fields are required in the field definition.

api/src/backend/api/models.py
```diff
@@ -49,6 +50,8 @@
 # Convert Prowler Severity enum to Django TextChoices
 SeverityChoices = enum_to_choices(Severity)
 
+logger = logging.getLogger(__name__)
```
vicferpoy: We shouldn't need custom logging at model level, but in any case, use the API's custom logger, not `__name__`.

api/src/backend/api/models.py
```diff
@@ -1332,3 +1335,160 @@ class Meta:
                 statements=["SELECT", "INSERT", "UPDATE", "DELETE"],
             ),
         ]
+
+
+class LighthouseConfig(RowLevelSecurityProtectedModel):
```
vicferpoy:
```suggestion
class LighthouseConfiguration(RowLevelSecurityProtectedModel):
```

Explicit is always better in model and table definitions.

api/src/backend/api/models.py
```diff
@@ -1332,3 +1335,160 @@ class Meta:
...
+                detail="Model is required",
+                code="missing_model",
+                pointer="/data/attributes/model",
+            )
+
+        self.full_clean()
+        super().save(*args, **kwargs)
+
+    class Meta(RowLevelSecurityProtectedModel.Meta):
+        db_table = "lighthouse_config"
```
vicferpoy:
```suggestion
        db_table = "lighthouse_configurations"
```

api/src/backend/api/models.py
```diff
@@ -1332,3 +1335,160 @@ class Meta:
...
+            ),
+        ]
+
+        indexes = [
+            models.Index(fields=["name"], name="lighthouse_config_name_idx"),
+            models.Index(fields=["is_active"], name="lighthouse_config_active_idx"),
+        ]
+
+    class JSONAPIMeta:
+        resource_name = "lighthouse-config"
```
vicferpoy:
```suggestion
        resource_name = "lighthouse-configurations"
```

api/src/backend/api/models.py
```diff
@@ -1332,3 +1335,160 @@ class Meta:
...
+            # Add unique constraint for name within a tenant
+            models.UniqueConstraint(
+                fields=["tenant_id"], name="unique_lighthouse_config_per_tenant"
+            ),
+        ]
+
+        indexes = [
+            models.Index(fields=["name"], name="lighthouse_config_name_idx"),
+            models.Index(fields=["is_active"], name="lighthouse_config_active_idx"),
+        ]
```
vicferpoy: I don't think these are really helpful since we are going to have only one configuration per tenant and `tenant_id` is already indexed and applied through filter for every LighthouseConfiguration query.

api/src/backend/api/migrations/0027_lighthouseconfig.py
vicferpoy, line 1: Review this after applying all the suggestions in `models.py`, please.

api/src/backend/api/v1/serializers.py
```diff
@@ -2128,3 +2130,129 @@ def update(self, instance, validated_data):
...
+    class Meta:
+        model = LighthouseConfig
+        fields = [
+            "name",
+            "api_key",
+            "model",
+            "temperature",
+            "max_tokens",
+            "business_context",
+            "is_active",
```
vicferpoy, line 2149: I think for consistency with other serializers, you should define all the expected fields as in the read serializer and then using `extra_kwargs` to define them as `read_only`.

api/src/backend/api/v1/serializers.py
```diff
@@ -2128,3 +2130,129 @@ def update(self, instance, validated_data):
...
+                }
+            )
+        return super().validate(attrs)
+
+    def create(self, validated_data):
+        api_key = validated_data.pop("api_key")
+
+        # Validate API key format before creating the instance
+        if not api_key:
+            raise serializers.ValidationError({"api_key": "API key is required"})
```
vicferpoy:
```python
api_key = serializers.CharField(write_only=True, required=True)
```

Isn't that enough? At that point in the flow the validation should have already kicked in 🤔

api/src/backend/api/v1/serializers.py
```diff
@@ -2128,3 +2130,129 @@ def update(self, instance, validated_data):
...
+        model = LighthouseConfig
+        fields = [
+            "id",
+            "name",
+            "api_key",
+            "model",
+            "temperature",
+            "max_tokens",
+            "business_context",
+            "is_active",
```
vicferpoy, line 2201: Same as the "CreateSerializer".

api/src/backend/api/v1/serializers.py
```diff
@@ -2128,3 +2130,129 @@ def update(self, instance, validated_data):
...
+        extra_kwargs = {
+            "id": {"read_only": True},
+            "name": {"required": False},
+            "model": {"required": False},
+            "temperature": {"required": False},
+            "max_tokens": {"required": False},
+        }
+
+    def update(self, instance, validated_data):
+        api_key = validated_data.pop("api_key", None)
```
vicferpoy, line 2213: Shouldn't we validate the `api_key` format here as well? Maybe you could write a custom writing serializer to DRY this and inherit in both serializers. You have some examples in the module.

api/src/backend/api/v1/urls.py
```diff
@@ -49,6 +50,9 @@
 router.register(r"overviews", OverviewViewSet, basename="overview")
 router.register(r"schedules", ScheduleViewSet, basename="schedule")
 router.register(r"integrations", IntegrationViewSet, basename="integration")
+router.register(
+    r"lighthouse-config", LighthouseConfigViewSet, basename="lighthouseconfig"
```
vicferpoy:
```suggestion
    r"lighthouse-configuration", LighthouseConfigViewSet, basename="lighthouseconfig"
```

api/src/backend/api/v1/views.py
vicferpoy, line 1: Missing tag definition for Lighthouse endpoints in SchemaView.

api/src/backend/api/v1/views.py
```diff
@@ -2900,3 +2905,105 @@ def get_serializer_context(self):
...
+    """
+    API endpoint for managing Lighthouse configuration.
+    """
+
+    filterset_fields = {
+        "name": ["exact", "icontains"],
+        "model": ["exact", "icontains"],
+        "is_active": ["exact"],
+        "inserted_at": ["gte", "lte"],
+    }
```
vicferpoy: Separate this into its own class in `filters.py`, please.

api/src/backend/api/v1/views.py
```diff
@@ -2900,3 +2905,105 @@ def get_serializer_context(self):
...
+    destroy=extend_schema(
+        tags=["Lighthouse"],
+        summary="Delete a Lighthouse configuration",
+        description="Remove a Lighthouse configuration by its ID.",
+    ),
+    check_connection=extend_schema(
+        tags=["Lighthouse"],
+        summary="Check the connection to the OpenAI API",
+        description="Verify the connection to the OpenAI API for a specific Lighthouse configuration.",
+    ),
```
vicferpoy: Define sample responses please, for every handled status code.

vicferpoy: This is required for every endpoint that does not return a response following a serializer, FYI. Otherwise, the plugin generates it.

api/src/backend/api/v1/views.py
```diff
@@ -2900,3 +2905,105 @@ def get_serializer_context(self):
...
+        instance = serializer.save()
+
+        headers = self.get_success_headers(serializer.data)
+        return Response(
+            LighthouseConfigSerializer(
+                instance, context=self.get_serializer_context()
+            ).data,
+            status=status.HTTP_201_CREATED,
+            headers=headers,
+        )
```
vicferpoy: This should not be required. That's the magic of class-based viewsets 😉

api/src/backend/api/v1/views.py
```diff
@@ -2900,3 +2905,105 @@ def get_serializer_context(self):
...
+
+    def get_serializer_class(self):
+        if self.action == "create":
+            return LighthouseConfigCreateSerializer
+        elif self.action == "partial_update":
+            return LighthouseConfigUpdateSerializer
+        return LighthouseConfigSerializer
+
+    @action(detail=True, methods=["get"])
+    def check_connection(self, request, pk=None):
```
vicferpoy: I have doubts regarding this endpoint. Logic is fine, but I don't know if we should use the task module for this, same as Provider's test_connection. Could this check take long in time? If so, it could be blocking for the API and thus delegated to a background task.

What do you think @jfagoagas @AdriiiPRodri  ?

api/src/backend/api/tests/test_views.py
```diff
@@ -5458,3 +5458,331 @@ def test_integrations_filters_invalid(self, authenticated_client, filter_name):
...
+            reverse("lighthouseconfig-list") + "?filter[model]=gpt-4o"
+        )
+        assert response.status_code == status.HTTP_200_OK
+        assert len(response.json()["data"]) == 1
+        # Test is_active filter
+        response = authenticated_client.get(
+            reverse("lighthouseconfig-list") + "?filter[is_active]=true"
+        )
+        assert response.status_code == status.HTTP_200_OK
+        assert len(response.json()["data"]) == 1
```
vicferpoy: Use pytest parametrization here, please. Easier to read, maintain and add new tests if required.

api/src/backend/api/tests/test_views.py
```diff
@@ -5458,3 +5458,331 @@ def test_integrations_filters_invalid(self, authenticated_client, filter_name):
...
+            reverse("lighthouseconfig-list") + "?sort=name"
+        )
+        assert response.status_code == status.HTTP_200_OK
+        assert len(response.json()["data"]) == 1
+        # Test sorting by inserted_at
+        response = authenticated_client.get(
+            reverse("lighthouseconfig-list") + "?sort=-inserted_at"
+        )
+        assert response.status_code == status.HTTP_200_OK
+        assert len(response.json()["data"]) == 1
```
vicferpoy: Use pytest parametrization here, please. Easier to read, maintain and add new tests if required.

vicferpoy (changes requested):
Great job! There are some changes required, but everything seems pretty good. Thank you.

Also, please don't forget about regenerating the API specification and updating the API changelog. It's in the API PR required checks 🫡

vicferpoy (changes requested):
Great job! There are some changes required, but everything seems pretty good. Thank you.

Also, please don't forget about regenerating the API specification and updating the API changelog. It's in the API PR required checks 🫡

api/src/backend/api/v1/serializers.py
```diff
@@ -2128,3 +2130,129 @@ def update(self, instance, validated_data):
...
+
+    def create(self, validated_data):
+        api_key = validated_data.pop("api_key")
+
+        # Validate API key format before creating the instance
+        if not api_key:
+            raise serializers.ValidationError({"api_key": "API key is required"})
+
+        # Validate OpenAI API key format
+        openai_key_pattern = r"^sk-[\w-]+T3BlbkFJ[\w-]+$"
```
AdriiiPRodri: Just a suggestion, there are two occurrences for this regex, I think we can create a constant value, this way it would be easiest to maintain

api/src/backend/api/v1/views.py
AdriiiPRodri: Absolutely, I think any external connection we don’t control should be handled in a background task to avoid blocking the API

api/src/backend/api/models.py
```diff
@@ -1332,3 +1335,160 @@ class Meta:
...
+            api_key_bytes = bytes(self.api_key)
+            decrypted_key = fernet.decrypt(api_key_bytes)
+            return decrypted_key.decode()
+
+        except InvalidToken:
+            logger.warning("Failed to decrypt API key: invalid token.")
+            return None
+        except Exception as e:
+            logger.error(f"Unexpected error while decrypting API key: {e}")
+            return None
```
AdriiiPRodri:
```suggestion
@property
def api_key_decoded(self):
    """Return the decrypted API key, or None if unavailable or invalid."""
    if not self.api_key:
        return None

    try:
        decrypted_key = fernet.decrypt(bytes(self.api_key))
        return decrypted_key.decode()
    except InvalidToken:
        logger.warning("Invalid token while decrypting API key.")
    except Exception as e:
        logger.exception("Unexpected error while decrypting API key.")
```

api/src/backend/api/models.py
AdriiiPRodri: Hmm, I’m not sure, postgres enums aren’t easy to modify, and OpenAI keeps releasing new models, so if we hardcode this in the code or in Postgres, we’ll need to deploy a new version every time a new model comes out. I think this should go in the `.env` file instead

AdriiiPRodri (changes requested):

api/src/backend/api/models.py
vicferpoy: Okay, fair. But in any case, this should be used for a ChoicesField, not a Char.

api/src/backend/api/models.py
vicferpoy: Updated related comment: https://github.com/prowler-cloud/prowler/pull/7848#discussion_r2120913217

api/src/backend/api/models.py
0xbadshah: We wont support all the models from OpenAI. Rather allow user to select from a list of models that we have vetted beforehand and recommend. While there is one model we highly recommend (gpt-4o in this case) there can be other models added here for experimental or cost reduction purposes.

@AdriiiPRodri Let's say we have a new model that we test and will integrate to future version, say 5.10, we can be confident that new LLM model will work with 5.10 only. We cant backport and test every supported Prowler version with new LLM model. So, what do we do? Hardcode (as own Postgres enum) or move it to .env file?

api/src/backend/api/v1/views.py
0xbadshah: This API is not being used in the frontend. We can remove it for now?

jfagoagas:
@Chan9390 @vicferpoy when you get a chance, please review the new endpoints' serializers because type must be `ligthouse-configurations` with the trailing `s`.

Thanks!

api/src/backend/api/models.py
```diff
@@ -1448,3 +1452,129 @@ class Meta:
...
+                statements=["SELECT", "INSERT", "UPDATE", "DELETE"],
+            ),
+            # Add unique constraint for name within a tenant
+            models.UniqueConstraint(
+                fields=["tenant_id"], name="unique_lighthouse_config_per_tenant"
+            ),
+        ]
+
+    class JSONAPIMeta:
+        resource_name = "lighthouse-configuration"
```
jfagoagas:
```suggestion
        resource_name = "lighthouse-configurations"
```

api/src/backend/api/tests/test_views.py
0xbadshah: I have updated the tests to use parametrization wherever possible. Thanks!

api/src/backend/api/tests/test_views.py
0xbadshah: I have updated tests to use parametrization wherever possible. Thanks!

api/src/backend/api/models.py
0xbadshah: Moved this to Django enum (not Postgres).

api/src/backend/api/models.py
0xbadshah: This is removed as its taken care by Django textchoices enum

api/src/backend/api/models.py
0xbadshah: As we're using django enums and not postgres, can this be left as it is?

api/src/backend/api/v1/serializers.py
0xbadshah: I found I was validating OpenAI keys at two places - redundant checks. So removed the one in serializers.

api/src/backend/api/v1/serializers.py
0xbadshah: You're right. Removed redundant check.

api/src/backend/api/v1/serializers.py
0xbadshah: Updated serializer. Thanks!

api/src/backend/api/v1/views.py
0xbadshah: Done. Thanks!

api/src/backend/api/v1/views.py
0xbadshah: Done. Thanks!

api/src/backend/api/models.py
AdriiiPRodri: In my opinion, in this case I would leave it as you have it, since this allows us to add new models as they become available without having to release a new version of Prowler, just by updating the .env. What do you think, @vicferpoy? I believe making this field a choice would force us to release a new version or even add migrations every time new models become available

api/src/backend/api/v1/views.py
AdriiiPRodri: I didn’t understand that last part, is this endpoint not being used? How is the connection being checked?

0xbadshah: @AdriiiPRodri This is fixed. I've moved the connection logic to celery tasks just like provider check.

api/src/backend/api/v1/serializers.py
```diff
@@ -2082,3 +2083,133 @@ class Meta:
...
+        model = LighthouseConfiguration
+        fields = [
+            "id",
+            "name",
+            "api_key",
+            "model",
+            "temperature",
+            "max_tokens",
+            "business_context",
+            "is_active",
```
AdriiiPRodri, line 2107: Should this field be read only?

api/src/backend/api/v1/views.py
```diff
@@ -3373,3 +3384,78 @@ def get_serializer_context(self):
...
+        summary="Check the connection to the OpenAI API",
+        description="Verify the connection to the OpenAI API for a specific Lighthouse configuration.",
+        responses={202: OpenApiResponse(response=TaskSerializer)},
+    ),
+)
+class LighthouseConfigViewSet(BaseRLSViewSet):
+    """
+    API endpoint for managing Lighthouse configuration.
+    """
+
```
AdriiiPRodri, line 3438: This is a suggestion to keep consistency with other views
```suggestion
serializer_class = LighthouseConfigSerializer
```

api/src/backend/api/v1/views.py
```diff
@@ -3373,3 +3384,78 @@ def get_serializer_context(self):
...
+    def get_queryset(self):
+        return LighthouseConfiguration.objects.filter(tenant_id=self.request.tenant_id)
+
+    def get_serializer_class(self):
+        if self.action == "create":
+            return LighthouseConfigCreateSerializer
+        elif self.action == "partial_update":
+            return LighthouseConfigUpdateSerializer
+        return LighthouseConfigSerializer
```
AdriiiPRodri: This is a suggestion to keep consistency with other views, needs to be applied with the previous one 
```suggestion
        return super().get_serializer_class()
```

api/src/backend/api/v1/serializers.py
0xbadshah: Updated

Chan9390:
@cursor-com

api/src/backend/api/models.py
```diff
@@ -1628,3 +1632,129 @@ class Meta:
...
+    Stores configuration and API keys for LLM services.
+    """
+
+    class ModelChoices(models.TextChoices):
+        GPT_4O_2024_11_20 = "gpt-4o-2024-11-20", _("GPT-4o v2024-11-20")
+        GPT_4O_2024_08_06 = "gpt-4o-2024-08-06", _("GPT-4o v2024-08-06")
+        GPT_4O_2024_05_13 = "gpt-4o-2024-05-13", _("GPT-4o v2024-05-13")
+        GPT_4O = "gpt-4o", _("GPT-4o Default")
+        GPT_4O_MINI_2024_07_18 = "gpt-4o-mini-2024-07-18", _("GPT-4o Mini v2024-07-18")
+        GPT_4O_MINI = "gpt-4o-mini", _("GPT-4o Mini Default")
```
AdriiiPRodri, line 1648: Should we add this to the .env file? Because change this cause that we need to release a new version

AdriiiPRodri (changes requested):

api/src/backend/api/models.py
```diff
@@ -1628,3 +1632,129 @@ class Meta:
...
+        max_length=100,
+        validators=[MinLengthValidator(3)],
+        blank=False,
+        null=False,
+        help_text="Name of the configuration",
+    )
+    api_key = models.BinaryField(
+        blank=False, null=False, help_text="Encrypted API key for the LLM service"
+    )
+    model = models.CharField(
```
vicferpoy, line 1664: Do we want to add a default choice here? Or do we want to always force the user to select a model?

api/src/backend/api/migrations/0031_lighthouseconfig.py
vicferpoy, line 1: You didn't update the migration file after changes in `models.py`.

api/CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ All notable changes to the **Prowler API** are documented in this file.
 - Added SSO with SAML support [(#7822)](https://github.com/prowler-cloud/prowler/pull/7822).
 - Support GCP Service Account key. [(#7824)](https://github.com/prowler-cloud/prowler/pull/7824)
 - Added new `GET /compliance-overviews` endpoints to retrieve compliance metadata and specific requirements statuses [(#7877)](https://github.com/prowler-cloud/prowler/pull/7877).
+- Added Lighthouse configuration support [(#7848)](https://github.com/prowler-cloud/prowler/pull/7848)
```
```suggestion
- Lighthouse configuration support [(#7848)](https://github.com/prowler-cloud/prowler/pull/7848)
```

api/src/backend/api/filters.py
```diff
@@ -704,3 +705,16 @@ class Meta:
...
+    inserted_at = DateFilter(field_name="inserted_at", lookup_expr="date")
+
+    class Meta:
+        model = LighthouseConfiguration
+        fields = {
+            "name": ["exact", "icontains"],
+            "model": ["exact", "icontains"],
+            "is_active": ["exact"],
+            "inserted_at": ["gte", "lte"],
+        }
```
vicferpoy: Does this make sense if we can only save one LighthouseConfiguration per tenant? I don't feel like this endpoint should have any filter.

api/src/backend/api/v1/views.py
```diff
@@ -3373,3 +3384,79 @@ def get_serializer_context(self):
...
+    partial_update=extend_schema(
+        tags=["Lighthouse"],
+        summary="Partially update a Lighthouse configuration",
+        description="Update certain fields of an existing Lighthouse configuration.",
+    ),
+    retrieve=extend_schema(
+        tags=["Lighthouse"],
+        summary="Retrieve a Lighthouse configuration",
+        description="Fetch detailed information about a specific Lighthouse configuration by its ID. Add query param `fields[lighthouse-config]=api_key` to get API key.",
+    ),
```
vicferpoy: I feel like this does not make sense, since each tenant will only have one possible configuration. Please remove this and disable the `retrieve` endpoint.

```python
    @extend_schema(exclude=True)
    def retrieve(self, request, *args, **kwargs):
        raise MethodNotAllowed(method="GET")
```

api/src/backend/tasks/jobs/connection.py
```diff
@@ -39,3 +39,48 @@ def check_provider_connection(provider_id: str):
 
     connection_error = f"{connection_result.error}" if connection_result.error else None
     return {"connected": connection_result.is_connected, "error": connection_error}
+
+
+def check_lighthouse_connection(lighthouse_config_id: str):
```
vicferpoy, line 45: Missing unit tests for this new function -> `test_connection.py`

api/src/backend/tasks/jobs/connection.py
```diff
@@ -39,3 +39,48 @@ def check_provider_connection(provider_id: str):
...
+        lighthouse_config.save()
+        return {
+            "connected": False,
+            "error": "API key is invalid or missing.",
+            "available_models": [],
+        }
+
+    try:
+        client = openai.OpenAI(api_key=lighthouse_config.api_key_decoded)
+        models = client.models.list()
```
vicferpoy, line 74: Just curious. Is this needed at all in this task? We are just trying to check if the current configuration is valid and can connect.

Also, we don't allow every possible model in our API that could come from OpenAI, right? I'd probably remove this and `available_models`.

vicferpoy (changes requested):
Please, review those required API PR required checks in the PR description, fill what is missing and check them. Thank you.

api/src/backend/api/models.py
vicferpoy: FYI I'm talking about `model`.

api/src/backend/tasks/jobs/connection.py
0xbadshah: This is to check if the api key is valid. We aren't doing anything with the response as of now.
While the regex prevents from entering any random value, the models list API checks if API key is valid or revoked.

api/src/backend/api/v1/views.py
```diff
@@ -3391,3 +3400,77 @@ def get_serializer_context(self):
...
+    destroy=extend_schema(
+        tags=["Lighthouse"],
+        summary="Delete a Lighthouse configuration",
+        description="Remove a Lighthouse configuration by its ID.",
+    ),
+    connection=extend_schema(
+        tags=["Lighthouse"],
+        summary="Check the connection to the OpenAI API",
+        description="Verify the connection to the OpenAI API for a specific Lighthouse configuration.",
+        responses={202: OpenApiResponse(response=TaskSerializer)},
```
vicferpoy, line 3431:
```suggestion
        request=None,
        responses={202: OpenApiResponse(response=TaskSerializer)},
```

vicferpoy (changes requested):

api/src/backend/api/v1/views.py
```diff
@@ -3391,3 +3400,78 @@ def get_serializer_context(self):
...
+
+    def get_queryset(self):
+        return LighthouseConfiguration.objects.filter(tenant_id=self.request.tenant_id)
+
+    def get_serializer_class(self):
+        if self.action == "create":
+            return LighthouseConfigCreateSerializer
+        elif self.action == "partial_update":
+            return LighthouseConfigUpdateSerializer
+        return super().get_serializer_class()
```
vicferpoy, line 3453:
```suggestion
        elif self.action == "connection":
            return TaskSerializer
        return super().get_serializer_class()
```

vicferpoy (changes requested):

vicferpoy (approved):
LGTM! Really great job 🚀

AdriiiPRodri (approved):
Awesome job, really appreciate it!

PR #7853, opened 2025-05-27
feat: compliance detail view + ENS

alejandrobailo:
### Context

The goal of this epic is to show in the UI the detailed view for each compliance model, we've started with ENS.
In this pull request, we have the whole structure, components, actions, and workflows.

This PR should have been divided into more parts, and the same applies to the task. The next major features should undergo product grooming before creating the Jira ticket.

### Description
246shots_so

- Charts and many other components created.
- SSR wrappers
- Detail view page for compliance
- Server actions for new endpoints
- Custom Dropdown component refactored
- Skeletons for tables, charts, and accordion
- ESLint rule to accept console.error
- New hook to handle loading state on client components and prevent skeleton flickering on streams

### Checklist

#### API

### License


ui/components/compliance/client-accordion-content.tsx
```diff
@@ -0,0 +1,169 @@
...
+        }
+      }
+    }
+
+    loadFindings();
+  }, [requirement, scanId, pageNumber, sort, startLoading, stopLoading]);
+
+  const checks = requirement.check_ids || [];
+  const checksTable = (
+    <div>
```
paabloLC: can be this replace by `<> </>` ?

ui/components/compliance/client-accordion-content.tsx
```diff
@@ -0,0 +1,169 @@
...
+            )}
+            data={expandedFindings || []}
+            metadata={findings?.meta}
+            disableScroll={true}
+          />
+        </div>
+      );
+    }
+
+    return <div>No findings</div>;
```
paabloLC: what about <span></span> ?

ui/lib/ens-compliance.tsx
```diff
@@ -0,0 +1,237 @@
...
+import {
+  Framework,
+  MappedComplianceData,
+  Requirement,
+  RequirementStatus,
+} from "@/types/compliance/compliance";
+
+import { ComplianceAccordionTitle } from "../components/compliance/compliance-accordion-title";
+
+export const translateType = (type: string) => {
```
paabloLC: let's add a new folder inside `/lib/compliance/ens/` or at least `/lib/compliace/here-your-file.tsx` it's up to you = ))

ui/lib/ens-compliance.tsx
```diff
@@ -0,0 +1,237 @@
...
+      return "Reinforcement";
+    case "medida":
+      return "Measure";
+    default:
+      return type;
+  }
+};
+
+export const mapComplianceData = (
+  attributesData: any,
```
paabloLC: let's take a look at these `anys` and refactor them where possible

ui/lib/ens-compliance.tsx
```diff
@@ -0,0 +1,237 @@
...
+  requirementsData: any,
```
paabloLC: same here

ui/lib/ens-compliance.tsx
```diff
@@ -0,0 +1,237 @@
...
+): MappedComplianceData => {
+  const attributes = attributesData?.data || [];
+  const requirements = requirementsData?.data || [];
+
+  // Create a map for quick lookup of requirements by id
+  const requirementsMap = new Map();
+  requirements.forEach((req: any) => {
```
paabloLC: also here

ui/lib/hooks/useLoadingState.ts
```diff
@@ -0,0 +1,73 @@
...
+   * Initial loading state
+   */
+  initialState?: boolean;
+}
+
+/**
+ * A hook to manage loading states with minimum duration and delay options
+ * to prevent flickering for very fast operations
+ */
+export function useLoadingState({
```
paabloLC: I’ve been handling loading states directly via formState.isSubmitting from React Hook Form, or using React’s Suspense, which usually covers our needs across different flows.

That said, this hook introduces delayed and minimum loading durations, which can be useful to avoid UI flicker or when more granular control is needed. It does add some complexity (useEffect, multiple useState calls), so let’s keep an eye on whether that’s strictly necessary for the current use case.

If you think this addresses a specific UX concern or could be reused across components, happy to discuss further and align on the trade-offs.

ui/types/components.ts
```diff
@@ -263,50 +263,19 @@ export interface ApiError {
...
-    pagination: {
-      page: number;
-      pages: number;
-      count: number;
-    };
-    version: string;
-  };
 }
 
 export interface ComplianceOverviewData {
```
paabloLC: This can be moved to the new file inside `types/compliance/`

ui/components/ui/table/data-table.tsx
```diff
@@ -29,12 +29,14 @@ interface DataTableProviderProps<TData, TValue> {
   data: TData[];
   metadata?: MetaDataProps;
   customFilters?: FilterOption[];
+  disableScroll?: boolean;
```
paabloLC, line 32: Just wondering what scenario you're aiming to cover with that one. Curious to understand the use case better = ))

ui/components/ui/custom/custom-dropdown-filter.tsx
```diff
@@ -10,167 +10,183 @@ import {
...
 
-const filterSelectedClass =
-  "inline-flex items-center border py-1 text-xs transition-colors border-transparent bg-default-500 text-secondary-foreground hover:bg-default-500/80 rounded-md px-2 font-normal";
-
-export const CustomDropdownFilter: React.FC<CustomDropdownFilterProps> = ({
+export const CustomDropdownFilter = ({
   filter,
   onFilterChange,
-}) => {
+}: CustomDropdownFilterProps) => {
```
paabloLC, line 24: Big refactor in this component, skipping the review in this first iteration.

ui/components/compliance/skeleton-compliance-accordion.tsx
```diff
@@ -0,0 +1,30 @@
+import { Skeleton } from "@nextui-org/react";
+import React from "react";
+
+interface SkeletonAccordionProps {
+  itemCount?: number;
+  className?: string;
+  isCompact?: boolean;
+}
+
+export const SkeletonAccordion = ({
```
paabloLC: Noticed the new `skeleton` folder under `ui` is the idea to centralize all skeleton components there, or is this one meant to be feature specific?

paabloLC (changes requested):

ui/components/ui/table/data-table.tsx
alejandrobailo: It is a prop that handles the possibility of changing the page and staying at the same point on the Y axis, meaning you won't go up in the page. Is super good for crazy long UI components like an accordion with all the keys opened, to avoid that the user has to go down again to see the table results.

ui/components/compliance/skeleton-compliance-accordion.tsx
alejandrobailo: This one is only for the compliance view, as we have compliance-skeleton-grid here

ui/types/components.ts
alejandrobailo: good catch

ui/lib/hooks/useLoadingState.ts
alejandrobailo: The thing is, we need this in the following scenario: 
A component like the accordion that has a smooth transition while opening, in the last level, we call the findings table, showing the skeleton, but if we show the skeleton with its dimensions instantly without this handler, we are transcending the accordion animation. With this hook, we can delay this loading states.
This is a hook only for special cases, will help us a lot with animated components and loading states, let's remove the hook and if we see we will need some transition for these states, we can cherry pick some commit again.

ui/components/compliance/client-accordion-content.tsx
alejandrobailo: What do you mean here?

ui/components/compliance/client-accordion-content.tsx
alejandrobailo: I just removed the div

ui/app/(prowler)/compliance/[compliancetitle]/page.tsx
```diff
@@ -0,0 +1,269 @@
...
+} from "@/types/compliance";
+
+interface ComplianceDetailSearchParams {
+  complianceId: string;
+  version?: string;
+  scanId?: string;
+  "filter[region__in]"?: string;
+}
+
+const Logo = ({ logoPath }: { logoPath: string }) => {
```
Comment (paabloLC) [inline ui/app/(prowler)/compliance/[compliancetitle]/page.tsx:35]: what do you think about naming this `LogoCompliance`? It helps make things clearer in case someone searches the project for logo.

 I know it’s not exportable and only used on a specific page = ))

paabloLC (changes requested):

ui/app/(prowler)/compliance/[compliancetitle]/page.tsx
Comment (alejandrobailo) [reply ui/app/(prowler)/compliance/[compliancetitle]/page.tsx:35]: I chose ComplianceLogo 😅

Comment (paabloLC) [reply ui/app/(prowler)/compliance/[compliancetitle]/page.tsx:35]: it will be in the next PR. Resolving comment.

paabloLC (approved):
great work!

PR #7854, opened 2025-05-27
fix(ui): Remove Content Layout and move userinfo to global state

chaithanya009:
### Context

Currently, the use(getUserInfo()) call is placed outside the <Suspense> boundary in the ContentLayout component. This causes rendering to be blocked until data is resolved, and the Suspense fallback (<SkeletonContentLayout />) is never shown.
Objective: Centralize page layout management by moving title/icon definitions and navbar functionality from individual pages into the main layout component.

What Changed:
✅ Centralized Metadata: Created PAGE_METADATA array in main-layout.tsx mapping all routes to titles/icons
✅ Unified Layout: Moved navbar rendering into main layout using usePathname() for dynamic content
✅ Simplified Pages: Removed ContentLayout wrapper from 13+ pages - they now render content directly
✅ Cleaner Architecture: Single source of truth for page metadata, reduced code duplication
[](url)
### Description

DRY Principle: No more repeating title/icon on every page
Easier Maintenance: Adding new pages only requires updating the metadata array
Better Performance: Single layout renders navbar once vs. per-page rendering
Cleaner Code: Pages focus purely on content without layout concerns

Fix:
Result: All pages now automatically get proper titles, icons, and navbar through the centralized main layout system.

### Checklist

#### API

### License


alejandrobailo (changes requested):
Thanks, I think it's better to get the user info in `user-nav.tsx` rather than in navbar.tsx to avoid the prop drilling because `navbar.tsx` does not use that data, or keep it in the ContentLayout.

chaithanya009:
> Thanks for the catch, I think it's better to get the user info in `user-nav.tsx` rather than in navbar.tsx to avoid the prop drilling because `navbar.tsx` does not use that data.

For some reason The suggested change is not working . I tried moving the Suspense directly on to `navbar.tsx` but it is showing the loading screen endlessly.

My guess is probably because the `user-nav.tsx` is a client component and the Suspense is in server component (`navbar.tsx` or `content-layout.tsx`).

So, I am reverting this change

alejandrobailo:
> > Thanks for the catch, I think it's better to get the user info in `user-nav.tsx` rather than in navbar.tsx to avoid the prop drilling because `navbar.tsx` does not use that data.
> 
> For some reason The suggested change is not working . I tried moving the Suspense directly on to `navbar.tsx` but it is showing the loading screen endlessly.
> 
> My guess is probably because the `user-nav.tsx` is a client component and the Suspense is in server component (`navbar.tsx` or `content-layout.tsx`).
> 
> So, I am reverting this change

Server actions can be called from client components like:
```typescript
export const UserNav = () => {
  const [user, setUser] = useState<UserProfileProps | null>(null);
  useEffect(() => {
    const fetchUser = async () => {
      const user = await getUserInfo();
      setUser(user);
    };
    fetchUser();
  }, []);

  if (!user || !user.data) return null;
```

We have to choose whether to pass it from the father or call it in the client component.


https://nextjs.org/docs/app/building-your-application/data-fetching/server-actions-and-mutations#behavior

chaithanya009:
> > > Thanks for the catch, I think it's better to get the user info in `user-nav.tsx` rather than in navbar.tsx to avoid the prop drilling because `navbar.tsx` does not use that data.
> > 
> > 
> > For some reason The suggested change is not working . I tried moving the Suspense directly on to `navbar.tsx` but it is showing the loading screen endlessly.
> > My guess is probably because the `user-nav.tsx` is a client component and the Suspense is in server component (`navbar.tsx` or `content-layout.tsx`).
> > So, I am reverting this change
> 
> Server actions can be called from client components like:
> 
> ```ts
> export const UserNav = () => {
>   const [user, setUser] = useState<UserProfileProps | null>(null);
>   useEffect(() => {
>     const fetchUser = async () => {
>       const user = await getUserInfo();
>       setUser(user);
>     };
>     fetchUser();
>   }, []);
> 
>   if (!user || !user.data) return null;
> ```
> 
> We have to choose whether to pass it from the father or call it in the client component.
> 
> https://nextjs.org/docs/app/building-your-application/data-fetching/server-actions-and-mutations#behavior

Thanks for the reference. I will look into it

chaithanya009:
The current approach is better than calling it in the user-nav for mainly two reasons:
1. useEffects are not compatible with Suspense since it happens after page mount, so we would have to handle the loading state manually in the user-nav (defeating the purpose of Suspense)
2. Data fetching would happen on the client, losing server-side rendering advantages."

alejandrobailo:
> The current approach is better than calling it in the user-nav for mainly two reasons:
> 
> 1. useEffects are not compatible with Suspense since it happens after page mount, so we would have to handle the loading state manually in the user-nav (defeating the purpose of Suspense)
> 2. Data fetching would happen on the client, losing server-side rendering advantages."

1. True, Suspense is easier to develop and understand.
2. False, data fetching will happen on the server because getUserInfo is a server action.

The point is that there are a lot of options in NextJS apps. For me, what makes more sense is to get the user on top of the app and save it in a shared state (Zustand, for example) and then consume it where is needed.

ATM, we can get it on the `ContentLayout` and remove the suspense, because we need the user to render the content; otherwise, what content should we render if there is no user? Another option is what you propose, get it on the navbar, this is not a good implementation because a `NavBar` should just render a UI component, not get data.

It is important to be clear in separating concerns and responsibilities. `navbar` shouldn't get user info, should be agnostic to the user.

Then the question is, given all the options we have and the time we have to improve this, what is the better option?

### Option 1: Global State (Zustand) - BEST Long-term

```typescript
// store/user-store.ts
import { create } from 'zustand'

interface UserStore {
  user: UserProfileProps | null
  setUser: (user: UserProfileProps) => void
}

export const useUserStore = create<UserStore>((set) => ({
  user: null,
  setUser: (user) => set({ user }),
}))

// app/(prowler)/layout.tsx
export default async function Layout({ children }) {
  const user = await getUserInfo()
  
  return (
    <UserProvider user={user}>
      {children}
    </UserProvider>
  )
}

// components/ui/user-nav/user-nav.tsx
export const UserNav = () => {
  const user = useUserStore(state => state.user)
  // ...
}
```

**✅ Pros:**
- Perfect separation of concerns
- Single call per session
- Scalable for entire app
- Navbar completely agnostic

**❌ Cons:**
- Requires more initial setup
- Needs provider in layout
- More architectural complexity

### Option 2: ContentLayout async (no Suspense) - BEST for Limited Time

```typescript
// content-layout.tsx
export async function ContentLayout({ title, icon, children }) {
  const user = await getUserInfo() // Blocks until user is available
  
  return (
    <>
      <Navbar title={title} icon={icon} user={user} />
      <div className="px-6 py-4 sm:px-8 xl:px-10">{children}</div>
    </>
  )
}
```

**✅ Pros:**
- Minimal change (just add async)
- Maintains separation of concerns
- Guarantees user is always available
- No regressions

**❌ Cons:**
- Blocks entire page if getUserInfo is slow

### Option 3: Navbar async - Works but Violates Principles

```typescript
// ❌ Violates Single Responsibility
export async function Navbar({ title, icon }) {
  const user = await getUserInfo() // Navbar fetching data
  return <header>...</header>
}
```

**❌ Issues:**
- Navbar should be presentational only
- Violates separation of concerns
- Makes component less reusable

## Recommendation

### For IMMEDIATE Implementation: Option 2 (ContentLayout async)

```typescript
// components/ui/content-layout/content-layout.tsx
import { getUserInfo } from "@/actions/users/users";
import { Navbar } from "../nav-bar/navbar";

interface ContentLayoutProps {
  title: string;
  icon: string;
  children: React.ReactNode;
}

export async function ContentLayout({ title, icon, children }: ContentLayoutProps) {
  const user = await getUserInfo()
  
  return (
    <>
      <Navbar title={title} icon={icon} user={user} />
      <div className="px-6 py-4 sm:px-8 xl:px-10">{children}</div>
    </>
  )
}
```

**Reasons:**
1. ✅ **Minimal time**: Just add `async` and `await`
2. ✅ **Maintains principles**: Navbar remains presentational
3. ✅ **Logical**: Need user to determine what content to render
4. ✅ **No breaking changes**: Doesn't affect existing code

### For FUTURE Improvement: Option 1 (Global State)

What do you think @paabloLC?

chaithanya009:
> > The current approach is better than calling it in the user-nav for mainly two reasons:
> > 
> > 1. useEffects are not compatible with Suspense since it happens after page mount, so we would have to handle the loading state manually in the user-nav (defeating the purpose of Suspense)
> > 2. Data fetching would happen on the client, losing server-side rendering advantages."
> 
> 1. True, Suspense is easier to develop and understand.
> 2. False, data fetching will happen on the server because getUserInfo is a server action.
> 
> The point is that there are a lot of options in NextJS apps. For me, what makes more sense is to get the user on top of the app and save it in a shared state (Zustand, for example) and then consume it where is needed.
> 
> ATM, we can get it on the `ContentLayout` and remove the suspense, because we need the user to render the content; otherwise, what content should we render if there is no user? Another option is what you propose, get it on the navbar, this is not a good implementation because a `NavBar` should just render a UI component, not get data.
> 
> It is important to be clear in separating concerns and responsibilities. `navbar` shouldn't get user info, should be agnostic to the user.
> 
> Then the question is, given all the options we have and the time we have to improve this, what is the better option?
> 
> ### Option 1: Global State (Zustand) - BEST Long-term
> ```ts
> // store/user-store.ts
> import { create } from 'zustand'
> 
> interface UserStore {
>   user: UserProfileProps | null
>   setUser: (user: UserProfileProps) => void
> }
> 
> export const useUserStore = create<UserStore>((set) => ({
>   user: null,
>   setUser: (user) => set({ user }),
> }))
> 
> // app/(prowler)/layout.tsx
> export default async function Layout({ children }) {
>   const user = await getUserInfo()
>   
>   return (
>     <UserProvider user={user}>
>       {children}
>     </UserProvider>
>   )
> }
> 
> // components/ui/user-nav/user-nav.tsx
> export const UserNav = () => {
>   const user = useUserStore(state => state.user)
>   // ...
> }
> ```
> 
> **✅ Pros:**
> 
> * Perfect separation of concerns
> * Single call per session
> * Scalable for entire app
> * Navbar completely agnostic
> 
> **❌ Cons:**
> 
> * Requires more initial setup
> * Needs provider in layout
> * More architectural complexity
> 
> ### Option 2: ContentLayout async (no Suspense) - BEST for Limited Time
> ```ts
> // content-layout.tsx
> export async function ContentLayout({ title, icon, children }) {
>   const user = await getUserInfo() // Blocks until user is available
>   
>   return (
>     <>
>       <Navbar title={title} icon={icon} user={user} />
>       <div className="px-6 py-4 sm:px-8 xl:px-10">{children}</div>
>     </>
>   )
> }
> ```
> 
> **✅ Pros:**
> 
> * Minimal change (just add async)
> * Maintains separation of concerns
> * Guarantees user is always available
> * No regressions
> 
> **❌ Cons:**
> 
> * Blocks entire page if getUserInfo is slow
> 
> ### Option 3: Navbar async - Works but Violates Principles
> ```ts
> // ❌ Violates Single Responsibility
> export async function Navbar({ title, icon }) {
>   const user = await getUserInfo() // Navbar fetching data
>   return <header>...</header>
> }
> ```
> 
> **❌ Issues:**
> 
> * Navbar should be presentational only
> * Violates separation of concerns
> * Makes component less reusable
> 
> ## Recommendation
> ### For IMMEDIATE Implementation: Option 2 (ContentLayout async)
> ```ts
> // components/ui/content-layout/content-layout.tsx
> import { getUserInfo } from "@/actions/users/users";
> import { Navbar } from "../nav-bar/navbar";
> 
> interface ContentLayoutProps {
>   title: string;
>   icon: string;
>   children: React.ReactNode;
> }
> 
> export async function ContentLayout({ title, icon, children }: ContentLayoutProps) {
>   const user = await getUserInfo()
>   
>   return (
>     <>
>       <Navbar title={title} icon={icon} user={user} />
>       <div className="px-6 py-4 sm:px-8 xl:px-10">{children}</div>
>     </>
>   )
> }
> ```
> 
> **Reasons:**
> 
> 1. ✅ **Minimal time**: Just add `async` and `await`
> 2. ✅ **Maintains principles**: Navbar remains presentational
> 3. ✅ **Logical**: Need user to determine what content to render
> 4. ✅ **No breaking changes**: Doesn't affect existing code
> 
> ### For FUTURE Improvement: Option 1 (Global State)
> What do you think @paabloLC?

About the 2nd point what i mean is the api call is initiated after the page is mounted unlike the original approach, data fetching does happen on server in both cases.
And I don't think there is anything wrong with api call in the navbar as there is no degradation in performance at all and we won't be reusing the navbar anyway

alejandrobailo:
> And I don't think there is anything wrong with api call in the navbar as there is no degradation in performance at all and we won't be reusing the navbar anyway

Is not about degradation, is about the separation of concerns, but anyway, there are a few things to change.
1. Content Layout shouldn't exist. In NextJS apps, the Layout component is responsible for these layouts that are repeated across the app.
2. The user should be requested at the top level of the app and saved in a global state, then we can toggle between views in the components regarding the roles of the user.

chaithanya009:
> 1. Content Layout shouldn't exist. In NextJS apps, the Layout component is responsible for these layouts that are repeated across the app.
> 2. The user should be requested at the top level of the app and saved in a global state, then we can toggle between views in the components regarding the roles of the user.

Yes! We can do that. I was just trying to use whatever we had for a quick fix.

alejandrobailo:
> > 1. Content Layout shouldn't exist. In NextJS apps, the Layout component is responsible for these layouts that are repeated across the app.
> > 2. The user should be requested at the top level of the app and saved in a global state, then we can toggle between views in the components regarding the roles of the user.
> 
> Yes! We can do that. I was just trying to use whatever we had for a quick fix.

I appreciate your suggestion, thanks a lot.

chaithanya009:
Hi @alejandrobailo . Implemented the long term fix!

Objective: Centralize page layout management by moving title/icon definitions and navbar functionality from individual pages into the main layout component.

What Changed:
✅ Centralized Metadata: Created PAGE_METADATA array in main-layout.tsx mapping all routes to titles/icons
✅ Unified Layout: Moved navbar rendering into main layout using usePathname() for dynamic content
✅ Simplified Pages: Removed ContentLayout wrapper from 13+ pages - they now render content directly
✅ Cleaner Architecture: Single source of truth for page metadata, reduced code duplication
Key Benefits:
1. DRY Principle: No more repeating title/icon on every page
2. Easier Maintenance: Adding new pages only requires updating the metadata array
3. Better Performance: Single layout renders navbar once vs. per-page rendering
4. Cleaner Code: Pages focus purely on content without layout concerns

Result: All pages now automatically get proper titles, icons, and navbar through the centralized main layout system.

ui/app/(prowler)/findings/page.tsx
```diff
@@ -90,8 +89,8 @@ export default async function Findings({
     completedScans?.map((scan: ScanProps) => scan.id) || [];
 
   return (
-    <ContentLayout title="Findings" icon="carbon:data-view-alt">
-      <FilterControls search date />
```
paabloLC, line 94: What's the reason you're removing the date filter on the Findings page? thanks

ui/app/(prowler)/layout.tsx
```diff
@@ -44,10 +47,12 @@ export default function RootLayout({
           fontSans.variable,
         )}
       >
-        <Providers themeProps={{ attribute: "class", defaultTheme: "dark" }}>
-          <MainLayout>{children}</MainLayout>
-          <Toaster />
-        </Providers>
+        <UserProvider initialUser={user}>
```
paabloLC, line 50: I'm curious, why do you need this?

ui/components/providers/user-provider.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+import { useUser } from "@/hooks/use-user";
+import { UserProfileProps } from "@/types";
+
+interface UserProviderProps {
+  children: React.ReactNode;
+  initialUser?: UserProfileProps;
+}
+
+export function UserProvider({ children, initialUser }: UserProviderProps) {
```
paabloLC, line 13: why do we need this? Are you aware that we're using useSession from next-auth/react? It already contains all the information about the user

ui/components/providers/user-provider.tsx
```diff
@@ -0,0 +1,23 @@
...
+  const { setUser } = useUser();
+
+  useEffect(() => {
```
paabloLC, line 16: I really recommend this article. We try to use `useEffect` only when there is no other option 

https://medium.com/@melikunesi123/why-i-dont-fetch-data-in-useeffect-7e39361d607a

ui/components/ui/user-nav/user-nav.tsx
```diff
@@ -24,11 +24,14 @@ import {
...
 } from "@/components/ui/tooltip/tooltip";
-import { UserProfileProps } from "@/types";
+import { useStore } from "@/hooks/use-store";
+import { useUser } from "@/hooks/use-user";
 
 import { Button } from "../button/button";
 
-export const UserNav = ({ user }: { user?: UserProfileProps }) => {
+export const UserNav = () => {
+  const userStore = useStore(useUser, (x) => x);
```
paabloLC, line 33: I would avoid using useStore. It relies on localStorage to store user info, which I don’t think is a good practice. I don’t really see the need for it if we can just use useSession instead

ui/components/ui/main-layout/main-layout.tsx
```diff
@@ -1,18 +1,101 @@
...
+
 import { useSidebar } from "@/hooks/use-sidebar";
 import { useStore } from "@/hooks/use-store";
 import { cn } from "@/lib/utils";
 
+import { Navbar } from "../nav-bar/navbar";
 import { Sidebar } from "../sidebar/sidebar";
+
+// Page metadata mapping
+const PAGE_METADATA = [
```
paabloLC, line 13: I might agree with extracting this as a shared config, but I’d prefer to handle it directly in the Next.js layout instead of introducing additional client components along the way. 

The function itself could live in a helper file (we’d need to decide which one), and then simply be called from the layout. 

Also, right now the page load doesn’t depend on the user, the user info is independent from the rendered components, is still that way?

ui/hooks/use-user.ts
```diff
@@ -0,0 +1,28 @@
...
+
+import { UserProfileProps } from "@/types";
+
+type UserStore = {
+  user: UserProfileProps | null;
+  setUser: (user: UserProfileProps | null) => void;
+  clearUser: () => void;
+};
+
+export const useUser = create(
```
paabloLC, line 12: Not sure this is the right approach. This stores user info in localStorage, which raises security concerns especially when we already have `useSession` available.

If we need shared access to user data, we could fetch it once in the layout and pass it via context, without persisting anything in the browser.

paabloLC (changes requested):

ui/app/(prowler)/findings/page.tsx
chaithanya009: Thanks for the catch! I will revert it.

PR #7876, opened 2025-05-28
fix: add new get method to avoid race conditions when creating async tasks

AdriiiPRodri:
## Description

In this PR we added a new method to avoid a race condition when we create Celery tasks

### Checklist

#### API

### License

### DIFF
--- a/api/CHANGELOG.md
+++ b/api/CHANGELOG.md
@@ -17,6 +17,7 @@ All notable changes to the **Prowler API** are documented in this file.
 ### Fixed
 - Fixed task lookup to use task_kwargs instead of task_args for scan report resolution. [(#7830)](https://github.com/prowler-cloud/prowler/pull/7830)
 - Fixed Kubernetes UID validation to allow valid context names [(#7871)](https://github.com/prowler-cloud/prowler/pull/7871)
+- Fixed a race condition when creating async tasks [(#7876)](https://github.com/prowler-cloud/prowler/pull/7876).
 
 ---
 
--- a/api/CHANGELOG.md
+++ b/api/CHANGELOG.md
@@ -17,7 +17,7 @@ All notable changes to the **Prowler API** are documented in this file.
 ### Fixed
 - Fixed task lookup to use task_kwargs instead of task_args for scan report resolution. [(#7830)](https://github.com/prowler-cloud/prowler/pull/7830)
 - Fixed Kubernetes UID validation to allow valid context names [(#7871)](https://github.com/prowler-cloud/prowler/pull/7871)
-- Fixed a race condition when creating async tasks [(#7876)](https://github.com/prowler-cloud/prowler/pull/7876).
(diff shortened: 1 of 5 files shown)


api/src/backend/api/models.py
```diff
@@ -352,6 +353,32 @@ class JSONAPIMeta:
         resource_name = "provider_groups-provider"
 
 
+class TaskManager(models.Manager):
+    def get_with_retry(self, id, max_retries=5, delay_seconds=0.5):
```
jfagoagas: could we make those values configurable with environment variables?

api/CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ All notable changes to the **Prowler API** are documented in this file.
 
 ### Fixed
 - Fixed task lookup to use task_kwargs instead of task_args for scan report resolution. [(#7830)](https://github.com/prowler-cloud/prowler/pull/7830)
+- Fixed a race condition when creating async tasks [(#7876)](https://github.com/prowler-cloud/prowler/pull/7876).
```
```suggestion
- Fixed a race condition when creating background tasks [(#7876)](https://github.com/prowler-cloud/prowler/pull/7876).
```

api/src/backend/api/models.py
vicferpoy: Could we reduce the delay to 0.1s? And maybe increment the number of retries. Here or in the views, but 0.5 seconds is too long in a context of sync views.

vicferpoy (dismissed):
Clean solution, great job! I left a few suggestions but it could be merged as is 🚀

api/src/backend/api/models.py
AdriiiPRodri: Sure, no problem!

jfagoagas (commented):
That's a clever and neat solution 👏 🥇 I left just one question about configuration.

api/src/backend/api/models.py
AdriiiPRodri: We can do that, yes, but we’d lose flexibility. I think having these values exposed could be useful in case asynchronous functions take different amounts of time, what do you think @vicferpoy?

jfagoagas: Why'd we lose flexibility? I mean having the same values by default but allow to configure them in whatever environment without needing to rebuild code.

vicferpoy: Maybe we could define default values depending on environment variables and hardcode/override those values in the code when necessary?

`.get_with_retry(id)` -> default value from environment variables.
`.get_with_retry(id, max_retries=10, delay_seconds=0.1)` -> it takes these hardcoded values no matter what the environment variable says.

AdriiiPRodri: Ok thanks guys I added it

jfagoagas (approved):
Great!! My only doubt is if the variables are ok regarding what've talked about including it in Django settings or not. I'm approving the PR and waiting to see what Victor says.

Thanks @AdriiiPRodri 🥇

PR #7877, opened 2025-05-28
feat(compliance): Rework compliance overviews

vicferpoy:
### Description

This PR adds changes on `GET /compliance-overviews`, introducing:
- New model to store compliance IDs, requirement IDs and region with status values.
- Redesign of the logic for `GET /compliance-overviews`.
- New endpoints for compliance attributes and compliance requirements status.
- Separation of the scan compliance data into another background task.
- Performance improvements for findings lookup.
- QoL changes for development experience with endpoints that could return a task and database index creation.

#### Examples

- If the user tries to retrieve the compliance data from a scan that just finished but the compliance tasks is still executing

```json
{
    "data": {
        "type": "tasks",
        "id": "b56fad74-ac2f-4c25-8ac0-3c494562bf4b",
        "attributes": {
            "inserted_at": "2025-05-29T11:45:42.425864Z",
            "completed_at": "2025-05-29T11:45:42.562259Z",
            "name": "scan-compliance-overviews",
            "state": "executing",
            "result": {
                "pid": 857,
                "hostname": "celery@8b57852ded9d"
            },
            "task_args": {
                "scan_id": "019716be-2b20-7e8f-a7b2-5c038176bea5"
            },
            "metadata": {}
        }
    },
    "meta": {
        "version": "v1"
    }
}
```

- If there is compliance overviews available:

```json
{
    "data": [
        {
            "type": "compliance-overviews",
            "id": "aws_account_security_onboarding_aws",
            "attributes": {
                "framework": "AWS-Account-Security-Onboarding",
                "version": "",
                "requirements_passed": 49,
                "requirements_failed": 31,
                "requirements_manual": 0,
                "total_requirements": 80
            }
        },
        {
            "type": "compliance-overviews",
            "id": "aws_audit_manager_control_tower_guardrails_aws",
            "attributes": {
                "framework": "AWS-Audit-Manager-Control-Tower-Guardrails",
                "version": "",
                "requirements_passed": 9,
                "requirements_failed": 5,
                "requirements_manual": 0,
                "total_requirements": 14
            }
        },

(...)

    ]
}
```

image

image

### Checklist

#### API

### License


api/src/backend/api/compliance.py
```diff
@@ -233,3 +236,88 @@ def generate_compliance_overview_template(prowler_compliance: dict):
             provider_compliance[compliance_name] = compliance_dict
 
     return template
+
+
+def aggregate_compliance_requirements(tenant_id: str, scan_id: str):
```
jfagoagas: Is there any reason why this function is not tested? If not, we need to test that this generates data correctly.

api/src/backend/api/migrations/0029_findings_check_index_parent.py
```diff
@@ -0,0 +1,17 @@
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("api", "0028_findings_check_index_partitions"),
+    ]
+
+    operations = [
+        migrations.AddIndex(
```
jfagoagas, line 10: Do you know if Django supports the `ONLY` attribute? To do the following before creating child's indexes
```sql
CREATE INDEX IF NOT EXISTS idx_parent ON ONLY parent_table (column_name);
```

api/src/backend/api/db_utils.py
jfagoagas, line 1: Love this file 👏

api/src/backend/api/migrations/0027_compliance_requirement_overviews.py
```diff
@@ -0,0 +1,125 @@
...
+                        choices=[
+                            ("FAIL", "Fail"),
+                            ("PASS", "Pass"),
+                            ("MANUAL", "Manual"),
+                        ]
+                    ),
+                ),
+                ("passed_checks", models.IntegerField(default=0)),
+                ("failed_checks", models.IntegerField(default=0)),
+                ("manual_checks", models.IntegerField(default=0)),
```
jfagoagas: Is the following correct?
- `passed_checks`: count of checks where all of its findings are `PASS`.
- `failed_checks`: count of checks where at least one of its findings is `FAIL`.
- `manual_checks`: what is a manual check? Prowler has a few manual checks but we are not considering it anymore. The only manual thing is a requirement.

api/src/backend/api/migrations/0027_compliance_requirement_overviews.py
```diff
@@ -0,0 +1,125 @@
...
+                    "tenant",
+                    models.ForeignKey(
+                        on_delete=django.db.models.deletion.CASCADE, to="api.tenant"
+                    ),
+                ),
+            ],
+            options={
+                "db_table": "compliance_requirements_overviews",
+                "abstract": False,
+                "indexes": [
```
jfagoagas, line 70: Do we need all these indexes at first? We should run several performance tests to validate that they are going to be used. If you did that it's fine then.

api/src/backend/api/migrations/0028_findings_check_index_partitions.py
```diff
@@ -0,0 +1,30 @@
...
+    atomic = False
+
+    dependencies = [
+        ("api", "0027_compliance_requirement_overviews"),
+    ]
+
+    operations = [
+        migrations.RunPython(
+            partial(
+                create_index_on_partitions,
```
jfagoagas, line 18: We should run several performance tests to validate that this index is used. If you did that it's fine then.

api/src/backend/api/models.py
```diff
@@ -1143,6 +1147,80 @@ class JSONAPIMeta:
         resource_name = "compliance-overviews"
 
 
+# TODOVICTOR add index to findings -> tenant_id, scan_id, check_id
```
jfagoagas:
```suggestion
```

This is already done, right?

api/src/backend/api/tests/test_db_utils.py
```diff
@@ -138,3 +142,102 @@ def test_batch_delete(self, tenants_fixture, create_test_providers):
         )
         assert Provider.objects.all().count() == 0
         assert summary == {"api.Provider": create_test_providers}
+
+
+class TestShouldCreateIndexOnPartition:
+    @pytest.fixture(scope="function")
+    def fixed_datetime(self, monkeypatch):
```
jfagoagas: Do you know https://github.com/spulec/freezegun?
```python
from freezegun import freeze_time
import datetime

@freeze_time("2012-01-14")
def test():
    assert datetime.datetime.now() == datetime.datetime(2012, 1, 14)
```

api/src/backend/api/v1/serializers.py
```diff
@@ -1679,130 +1679,61 @@ class Meta:
 # Compliance overview
 
 
-class ComplianceOverviewSerializer(RLSSerializer):
+class ComplianceOverviewSerializer(serializers.Serializer):
```
jfagoagas, line 1682: Why the parent class is not `RLSSerializer` anymore?

api/src/backend/api/v1/serializers.py
```diff
@@ -1679,130 +1679,61 @@ class Meta:
...
-            },
-        }
-    )
-    def get_requirements(self, obj):
-        """
-        Returns the detailed structure of requirements.
-        """
-        return obj.requirements
+
+class ComplianceOverviewAttributesSerializer(serializers.Serializer):
```
jfagoagas, line 1722: Same question.

api/src/backend/api/v1/views.py
```diff
@@ -2446,51 +2522,44 @@ def get_queryset(self):
             role, Permissions.UNLIMITED_VISIBILITY.value, False
         )
 
-        if self.action == "retrieve":
```
jfagoagas, line 2449: Do you remember why we had this?

api/src/backend/api/v1/views.py
```diff
@@ -2501,7 +2570,82 @@ def list(self, request, *args, **kwargs):
                     }
                 ]
             )
-        return super().list(request, *args, **kwargs)
+        try:
+            if task := self.get_task_response_if_running(
```
jfagoagas, line 2574: Can you create a ticket to apply this to the `scan-reports` view?

jfagoagas (commented):
What a great job you did 👏 The code is easy to read and follow.

I left several comments / questions, review it when you get a chance. Thanks!

api/src/backend/api/v1/views.py
```diff
@@ -2423,19 +2434,84 @@ def destroy(self, request, *args, **kwargs):
...
     pagination_class = ComplianceOverviewPagination
-    queryset = ComplianceOverview.objects.all()
+    queryset = ComplianceRequirementOverview.objects.all()
     serializer_class = ComplianceOverviewSerializer
     filterset_class = ComplianceOverviewFilter
     http_method_names = ["get"]
     search_fields = ["compliance_id"]
     ordering = ["compliance_id"]
-    ordering_fields = ["inserted_at", "compliance_id", "framework", "region"]
+    ordering_fields = ["compliance_id"]
```
AdriiiPRodri, line 2514: I don’t see why we now only have id as ordering_fields

AdriiiPRodri (changes requested):

api/src/backend/api/compliance.py
vicferpoy: This was the previous logic, isolated. Development was performed in several steps and I missed deleting this in the end.

api/src/backend/api/migrations/0029_findings_check_index_parent.py
vicferpoy: Good point. The time I spent on the `db_utils` is worthless because `AddIndex` will create those indexes non concurrently anyway. There's no way to do this without custom code, and I think it's way out of the scope of this development. What do you think?

api/src/backend/api/migrations/0027_compliance_requirement_overviews.py
vicferpoy: Yes, and we have a few manual checks. Just running a scan on our ECR account:

```
prowler_db=# select count(*) from compliance_requirements_overviews where manual_checks > 0 ;
 count 
-------
    44
(1 row)

prowler_db=# 
```

api/src/backend/api/migrations/0027_compliance_requirement_overviews.py
vicferpoy: We support some filters, that's why some indexes exist with those combinations. But not all of them will be used on a daily basis.

api/src/backend/api/migrations/0028_findings_check_index_partitions.py
vicferpoy: This is going to be used by the UI to retrieve the findings that belong to a particular requirement.

api/src/backend/api/v1/views.py
vicferpoy: Yes. That was the previous detailed endpoint for a particular compliance overview in the database. That logic is not available anymore, since we form aggregations and never return a particular database object.

api/src/backend/api/v1/views.py
vicferpoy: Because it would imply ordering at Python level since we need to retrieve the results ordered by compliance_id to properly do the aggregations and it was not worth it, in my opinion.

We would have to add custom ordering logic for each endpoint and with Python, not SQL. Do you think it's worth @jfagoagas @AdriiiPRodri ?

vicferpoy: In addition to what I previously said about aggregations. This endpoint no longer returns database objects. It returns aggregations/summaries of the stuff we store in the database, so it's not exactly a ModelViewSet per se.

api/src/backend/api/v1/serializers.py
vicferpoy: Because we are not retrieving particular database objects, so this serializer is not a ModelSerializer (the parent class of RLSSerializer). We are "making up" responses based on database data, not retrieving specific data. RLS is applied on a lower level.

api/src/backend/api/v1/serializers.py
vicferpoy: https://github.com/prowler-cloud/prowler/pull/7877#discussion_r2120687413

api/src/backend/api/tests/test_db_utils.py
vicferpoy: Indeed, way better. Thanks.

api/src/backend/api/v1/views.py
vicferpoy: After we merge, yes.

api/src/backend/api/compliance.py
jfagoagas: Awesome, thanks!

api/src/backend/api/v1/views.py
AdriiiPRodri: I see, maybe not in this case, but I think it could be useful in the future. We can tackle it later

api/src/backend/api/migrations/0029_findings_check_index_parent.py
jfagoagas: but it is working somehow because it is not failing when creating the parent indexes even when the child' indexes are already there. 

Now I get what you said about `worthless`.. it is not, we just need to add a helper to handle parent index creation and verify that only child indexes are added to the parent index for the partitions we want.

```sql
CREATE INDEX IF NOT EXISTS idx_parent ON ONLY parent_table (column_name);

CREATE INDEX CONCURRENTLY IF NOT EXISTS idx_child_partition ON child_partition (column_name);

ALTER INDEX idx_parent ATTACH PARTITION idx_child_partition;
```

api/src/backend/api/migrations/0027_compliance_requirement_overviews.py
jfagoagas: I know it, but why do we need that? We are not creating more `manual` checks, so I think the requirements with the `manual` checks should be marked as `manual` requirements. Probably we should talk with @pedrooot and review in which compliances are the affected checks to remove it.

api/src/backend/api/migrations/0027_compliance_requirement_overviews.py
jfagoagas: That's good, I just wanted to make sure that when using those filters the indexes are also selected.

api/src/backend/api/v1/serializers.py
jfagoagas: Thanks for the clarification, good learning!

api/src/backend/api/migrations/0027_compliance_requirement_overviews.py
vicferpoy: Because they are needed for now to know which requirements are MANUAL. Otherwise, how could it know? The API does not have that info. By default, all requirements are PASS.

api/src/backend/api/migrations/0029_findings_check_index_parent.py
vicferpoy: Not sure this approach is maintainable in the future when future partitions are created automatically... I think that's handled by the Django functions.

api/src/backend/api/migrations/0027_compliance_requirement_overviews.py
jfagoagas: I'm missing something.. a `MANUAL` requirement is the one where `checks == []`. That's why I asked why are we counting `finding.status == MANUAL`, it can exist because we have 3-4 checks that returns that but I think we should not take them into account anymore.

> By default, all requirements are PASS.

That can only happen if the `requirement.checks != []`. For `MANUAL` requirements we cannot state their status since it requires manual intervention.

vicferpoy: You are right. I updated it.

api/src/backend/api/migrations/0029_findings_check_index_parent.py
vicferpoy: For now, I disabled the flag for this migration. We will address this issue on another PR.

jfagoagas: We need to create a ticket to verify it before pushing this.

api/src/backend/api/migrations/0027_compliance_requirement_overviews.py
jfagoagas: Reviewed IRL, `manual_checks` is not needed anymore.

api/src/backend/api/migrations/0028_findings_check_index_partitions.py
jfagoagas: We need to create a ticket to verify it before pushing this.

jfagoagas (approved):

AdriiiPRodri (approved):

PR #7890, opened 2025-05-30
fix: set user_id for tenant operations

AdriiiPRodri:
### Description

In this PR, we fixed an issue related to modifying and retrieving tenants in the system. Sometimes the API would return a 500 error indicating that the UUID hadn’t been set, in this case, the user’s UUID for the transaction

### Checklist

#### API

### License

### DIFF
--- a/api/src/backend/api/base_views.py
+++ b/api/src/backend/api/base_views.py
@@ -109,25 +109,11 @@ class BaseTenantViewset(BaseViewSet):
                 pass  # Tenant might not exist, handle gracefully
 
     def initial(self, request, *args, **kwargs):
-        if (
-            request.resolver_match.url_name != "tenant-detail"
-            and request.method != "DELETE"
-        ):
-            user_id = str(request.user.id)
-
-            with rls_transaction(value=user_id, parameter=POSTGRES_USER_VAR):
-                return super().initial(request, *args, **kwargs)
-
-        # TODO: DRY this when we have time
         if request.auth is None:
             raise NotAuthenticated
 
-        tenant_id = request.auth.get("tenant_id")
-        if tenant_id is None:
-            raise NotAuthenticated("Tenant ID is not present in token")
-
-        with rls_transaction(tenant_id):
-            self.request.tenant_id = tenant_id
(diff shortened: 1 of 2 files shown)


jfagoagas:
I think we can also refactor the RLS policies under `tenant` with:

```sql
create POLICY "prowler_user_tenants_select" 
ON tenants 
FOR SELECT / DELETE / UPDATE
TO prowler_user 
USING (
  (EXISTS ( 
    SELECT 1 
    FROM memberships 
    WHERE ((memberships.tenant_id = tenants.id) AND (memberships.user_id = (current_setting('api.user_id'::text, true))::uuid)))))
```

api/src/backend/api/base_views.py
```diff
@@ -109,25 +109,11 @@ def _handle_creation_error(self, error, tenant):
...
             raise NotAuthenticated
 
-        tenant_id = request.auth.get("tenant_id")
-        if tenant_id is None:
-            raise NotAuthenticated("Tenant ID is not present in token")
-
-        with rls_transaction(tenant_id):
-            self.request.tenant_id = tenant_id
+        user_id = str(request.user.id)
+        with rls_transaction(value=user_id, parameter=POSTGRES_USER_VAR):
```
jfagoagas, line 120:
```suggestion
        tenant_id = request.auth.get("tenant_id")
        if tenant_id is None:
            raise NotAuthenticated("Tenant ID is not present in token")

        user_id = str(request.user.id)
        with rls_transaction(value=user_id, parameter=POSTGRES_USER_VAR):
```

AdriiiPRodri: I don't know if we need this because is there is no tenant, then request.auth should be None, if not possible to auth without a tenant. So I think the first condition cover this

jfagoagas: I talked about this with Victor, just in case to have another defensive layer.

AdriiiPRodri: Ok, I will add then

jfagoagas (approved):

PR #7914, opened 2025-06-03
test(performance): Added resources performance tests

sumit-tft:
### Description

This PR includes resources performance test cases for the following endpoints
 - GET /resources – with pagination and default sorting
 - GET /resources – with include=findings,provider
 - GET /resources – with random filters
-  GET /resources – with random filters and include=findings,provider
 
image

### Checklist

#### API

### License


api/tests/performance/scenarios/resources.py
```diff
@@ -0,0 +1,38 @@
...
+        with self.client.get(
+            "/resources", headers=self.headers, name="/resources", catch_response=True
+        ) as response:
+            if response.status_code == 200:
+                json_data = response.json()
+                GLOBAL["resource_ids"] = [
+                    item["id"] for item in json_data.get("data", [])[:10]
+                ]
+            else:
+                response.failure("Failed to load /resources")
```
vicferpoy: This was correctly implemented, but we don't require it. `retrieve` endpoints perform a very fast look up that we know for sure they perform as expected. The related tests add unnecessary load to the scenario. In any case, I appreciate your attention to detail, thank you! :)

vicferpoy: Please remove the helpers used for the random `GET /resources/<id>` tests.

vicferpoy (changes requested):
It's a good baseline! But I'm missing some variety regarding filters and `include` parameter. The idea was to have some automated examples of the requests the UI could do on a daily basis when a user access the Resource view and plays with the filters.

api/tests/performance/scenarios/resources.py
sumit-tft: Got it, thanks for the clarification! I’ve removed the helpers and the related `GET /resources/<id>` tests as per your suggestion.

sumit-tft:
Thanks for the feedback! I've updated the test cases to include more variety by using random filter combinations based on the available filters, and I've also added the include parameter.

PR #7923, opened 2025-06-03
fix(app): change api call for `ftps_state`

danibarranqueroo:
### Context

This PR updates the Azure App Service integration to retrieve the `ftps_state` property for `Function Apps` using the correct Azure SDK API call.
Previously, the code did not fetch the FTPS state properly, resulting in `ftps_state` always being `None`.
Now, the implementation uses `client.web_apps.get_configuration(resource_group_name, app_name)` to obtain the `SiteConfigResource`, from which the `ftps_state` value is accurately extracted and assigned to each `FunctionApp` object.

The issue was identified because the `ftps_state` property was always `None` in the check results, regardless of the actual configuration in Azure after making a fixer.

### Description

Modified the service and added tests.

### Checklist

#### API

### License


prowler/providers/azure/services/app/app_service.py
```diff
@@ -136,6 +136,11 @@ def _get_functions(self):
                             subscription_name, function.resource_group, function.name
                         )
 
+                        web_app_config = client.web_apps.get_configuration(
```
puchy22: Please could you split the SDK functions in separated functions with its own try except

puchy22 (changes requested):

prowler/providers/azure/services/app/app_service.py
danibarranqueroo: Done

puchy22 (dismissed):
Nice catch! 👍

puchy22 (approved):

PR #7967, opened 2025-06-06
fix(iam): check always if root credentials are present

andoniaf:
### Context
When you enable `RootCredentialsManagement` that doesn't remove or deactivate root credentials from your member accounts, so we need to keep checking if those are present and using MFA.

### Description
This PR fixes the logic to always check for root credentials, even when being part of an Organization with `RootCredentialsManagement`. 

Depending on this feature if gives a different recommendation even when the result is PASS. Because even if you're using MFA, if your Organization has `RootCredentialsManagement` enabled the recommendation is to remove root credentials from the member accounts.

### Checklist

#### API

### License


MrCloudSec (changes requested):
Nit: introducing `has_credentials` here means touching every check that reads the credentials report. Let’s keep the original logic in this PR and do a full refactor in a separate task.

prowler/providers/aws/services/iam/iam_no_root_access_key/iam_no_root_access_key.py
```diff
@@ -5,39 +5,70 @@
...
-                        and user["access_key_2_active"] == "true"
-                    ):
-                        report.status = "FAIL"
-                        report.status_extended = (
-                            "Root account has two active access keys."
+                    # Check if root has any credentials at all
+                    has_creds, cred_types = iam_client.has_credentials(user)
+
+                    # Only report if root actually has credentials
+                    if has_creds:
```
MrCloudSec:
```suggestion
                    if user_data["password_enabled"] == "true" or user_data["access_key_1_active"] == "true" or user_data["access_key_2_active"] == "true":
```

MrCloudSec (approved):

PR #7987, opened 2025-06-10
feat(api): handle MitreAttack compliance requirements

pedrooot:
### Description

This PR adds some fields that were needed for the MitreAttack compliance framework on the API side.

Added:
* name
* tactics (inside technique_details)
* subtechniques (inside technique_details)
* platforms (inside technique_details)
* technique_url (inside technique_details)

[API Response]
Screenshot 2025-06-12 at 18 01 29
Screenshot 2025-06-12 at 18 02 13

### Checklist

#### API

### License


api/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ All notable changes to the **Prowler API** are documented in this file.
 - Added SSO with SAML support [(#7822)](https://github.com/prowler-cloud/prowler/pull/7822).
 - Support GCP Service Account key. [(#7824)](https://github.com/prowler-cloud/prowler/pull/7824)
 - Added new `GET /compliance-overviews` endpoints to retrieve compliance metadata and specific requirements statuses [(#7877)](https://github.com/prowler-cloud/prowler/pull/7877).
+- Add attributes for Mitre Attack compliance requirements. [(#7897)](https://github.com/prowler-cloud/prowler/pull/7987)
```
vicferpoy: I don't think this is required. It's part of a new feature that is not yet released.

vicferpoy (changes requested):
Thank you for the help! However, I don't really feel like this workaround is the best. I see two issues:
1. Now every compliance adds information like `subtechniques`, `platforms` and `technique_url` even though they might make no sense.
2. I see this information as part of the attributes of a compliance, but with this approach we would have all of these new fields on a higher level in the response. Directly at `attributes` response object level. This level is used to represent the fields an object has. They are not the same.

Let me know what you think about this and feel free to discuss it. Thank you!

vicferpoy (changes requested):
Also, unit tests are failing and the API PR required checks were not taken into account. Please review that and perform any required changes after the implementation is clear.

api/CHANGELOG.md
pedrooot: Removed ✅

api/src/backend/api/v1/views.py
```diff
@@ -2992,13 +2992,38 @@ def attributes(self, request):
 
             metadata = requirement.get("attributes", [])
 
+            technique_details = {}
```
vicferpoy: If this only makes sense for MITRE, we can use the `compliance_id` as a check to pull data from the requirement. That way we can avoid that if every single time.

api/src/backend/api/v1/views.py
```diff
@@ -2992,13 +2992,38 @@ def attributes(self, request):
...
+                    ),
+                    "name": requirement.get("name", ""),
                     "framework": compliance_framework.get("framework", ""),
                     "version": compliance_framework.get("version", ""),
                     "description": requirement.get("description", ""),
-                    "attributes": {"metadata": metadata, "check_ids": check_ids},
+                    "attributes": {
+                        "metadata": metadata,
+                        "check_ids": check_ids,
+                        "technique_details": technique_details,
```
vicferpoy: Same principle here. If this is not MITRE, let's not add `technique_details`.

api/src/backend/api/v1/views.py
vicferpoy, line 1: Could we add test cases for MITRE to test this implementation? In `test_views.py`.

vicferpoy (changes requested):
```
 Verify if API specs need to be regenerated.
```

Every time we modify serializers, the response changes. Thus, specs must be regenerated.

api/src/backend/api/v1/views.py
pedrooot: You're right, done.

api/src/backend/api/v1/views.py
pedrooot: Done

api/src/backend/api/v1/views.py
pedrooot: Done ✅

api/src/backend/api/v1/views.py
```diff
@@ -2992,13 +2992,31 @@ def attributes(self, request):
 
             metadata = requirement.get("attributes", [])
 
+            base_attributes = {
+                "metadata": metadata,
+                "check_ids": check_ids,
+            }
+
+            # Add technique details for MITRE-ATTACK framework
+            if compliance_framework.get("framework", "") == "MITRE-ATTACK":
```
```suggestion
            if "mitre_attack" in compliance_id:
```

vicferpoy (changes requested):

vicferpoy (approved):
Thank you! 🐐

PR #7992, opened 2025-06-10
feat(m365powershell): add `pwsh` authentication via service principal

HugoPBrito:
### Context

Microsoft is blocking user sign-ins without `MFA`, breaking our authentication flow. 

### Description

This PR implements a solution to authenticate `PowerShell` via `APP credentials`.

### Checklist

#### API

### License


MrCloudSec (dismissed):

AdriiiPRodri (dismissed):

prowler/providers/m365/lib/powershell/m365_powershell.py
```diff
@@ -129,46 +144,151 @@ def test_credentials(self, credentials: M365Credentials) -> bool:
...
-        return True
+    def test_exchange_connection(self) -> bool:
+        """Test Exchange Online API connection and raise exception if it fails."""
+        try:
+            self.execute(
+                '$SecureSecret = ConvertTo-SecureString "$clientSecret" -AsPlainText -Force'
+            )
+            self.execute(
+                '$exchangeToken = Get-MsalToken -clientID "$clientID" -tenantID "$tenantID" -clientSecret $SecureSecret -Scopes "https://outlook.office365.com/.default"'
+            )
```
danibarranqueroo, line 278: I think this line is making the --init-modules fail when you use the --sp-env-auth authentication method, because that code is executed before the initialize_m365_powershell_modules() function and Get-MsalToken is a cmdlet from MSAL.PS module, so if you don't have that module installed it will fail.

The code is the following:
```
if test_session.test_credentials(credentials):
                    if init_modules:
                        initialize_m365_powershell_modules()
                    return credentials
```

So, as I said, the test credentials is executed before installing the modules so it only works with --env-auth (I think this is because before installing modules was only expected if you use --env-auth but now we should install them also if using --sp-env-auth)

MrCloudSec (approved):

AdriiiPRodri (approved):

PR #7993, opened 2025-06-10
feat: Processors API endpoint, implement MuteList

drewkerrigan:
### Context

Users need the ability to store and use a Mutelist (https://docs.prowler.com/projects/prowler-open-source/en/latest/tutorials/mutelist/) using the Prowler API.

### Description

- Introduced a new `/processors` API endpoint.
  - Processors may have a `processor_type`, and at this time the only type supported is `mutelist`
- Once a Mutelist has been created, the next scan will use that Mutelist, and muted findings will have a new `muted_reason` field populated with "Muted by mutelist"
- Mutelist configuration can be uploaded in JSON or YAML formats.

### Checklist

#### API

### License


api/src/backend/api/migrations/0033_processors.py
```diff
@@ -0,0 +1,88 @@
...
+        ),
+        migrations.AddConstraint(
+            model_name="processor",
+            constraint=RowLevelSecurityConstraint(
+                "tenant_id",
+                name="rls_on_processor",
+                statements=["SELECT", "INSERT", "UPDATE", "DELETE"],
+            ),
+        ),
+        migrations.AddField(
```
jfagoagas: What's the reason why we want to add a relationship between scans and processors?

api/src/backend/api/specs/v1.yaml
jfagoagas, line 1: Something's odd with this file, there are changes that should not be.

api/src/backend/api/models.py
```diff
@@ -697,6 +704,9 @@ class DeltaChoices(models.TextChoices):
     check_id = models.CharField(max_length=100, blank=False, null=False)
     check_metadata = models.JSONField(default=dict, null=False)
     muted = models.BooleanField(default=False, null=False)
+    muted_reason = models.TextField(
```
jfagoagas, line 708: Adding this new field creates a new filter in findings. Do we want it? If so, we need to review performance and database indexes because is a text field and using it will be similar as a text search.

api/src/backend/api/specs/v1.yaml
```diff
@@ -3036,7 +3213,7 @@ paths:
             - -pass
         explode: false
       tags:
-      - Overview
+      - findings-overview
```
jfagoagas: This should not be changed.

api/src/backend/api/specs/v1.yaml
```diff
@@ -3181,7 +3358,7 @@ paths:
             - -informational
         explode: false
       tags:
-      - Overview
+      - findings-severity-overview
```
jfagoagas: This should not be changed.

api/src/backend/api/specs/v1.yaml
```diff
@@ -3194,12 +3371,26 @@ paths:
...
+            class MyViewSet(viewsets.ModelViewSet):
+                queryset = Book.objects.all()
+                prefetch_for_includes = {
+                    '__all__': [],
+                    'category.section': ['category']
+                }
+                select_for_includes = {
+                    '__all__': [],
+                    'author': ['author', 'author__authorbio'],
+                }
```
jfagoagas: Is this expected?

api/src/backend/api/specs/v1.yaml
```diff
@@ -3215,7 +3406,7 @@ paths:
           basis by including a fields[TYPE] query parameter.
         explode: false
       tags:
-      - Overview
+      - providers-overview
```
jfagoagas: The same as before, this must be `Overview`.

api/src/backend/api/specs/v1.yaml
```diff
@@ -3356,7 +3547,7 @@ paths:
             - -pass
         explode: false
       tags:
-      - Overview
+      - services-overview
```
jfagoagas: The same as before, this must be `Overview`.

api/src/backend/api/v1/serializer_utils/base.py
```diff
@@ -0,0 +1,23 @@
...
+    def validate(self, data):
+        if hasattr(self, "initial_data"):
+            initial_data = set(self.initial_data.keys()) - {"id", "type"}
+            unknown_keys = initial_data - set(self.fields.keys())
+            if unknown_keys:
+                raise ValidationError(f"Invalid fields: {unknown_keys}")
+        return data
+
+
+class YamlOrJsonField(serializers.JSONField):
```
jfagoagas, line 16: I'm a little bit confused about what this does. I need to continue the review but JSON and YAML are interoperable, so in the SDK the Mutelist is input as YAML but then transformed to JSON to work with it.

Not sure if this comment applies, but just in case.

api/src/backend/api/v1/serializers.py
```diff
@@ -2065,6 +2072,128 @@ def update(self, instance, validated_data):
...
+
+        mutelist_configuration = configuration.get("Mutelist", {})
+
+        if not mutelist_configuration:
+            raise serializers.ValidationError(
+                "Invalid Mutelist configuration: 'Mutelist' is a required property."
+            )
+
+        try:
+            validate_jsonschema(mutelist_configuration, mutelist_schema)
```
jfagoagas: You can get rid of this and use:
```python
from prowler.lib.mutelist.mutelist import Mutelist

Mutelist.validate_mutelist(mutelist_configuration)
```

api/src/backend/api/v1/serializers.py
```diff
@@ -2065,6 +2072,128 @@ def update(self, instance, validated_data):
...
+            return
+        except Exception as error:
+            raise serializers.ValidationError(
+                f"Invalid Mutelist configuration: {error}"
+            )
```
jfagoagas: Can this be moved to somewhere not to repeat it?

jfagoagas (changes requested):
1st pass to the code done! That was easy to follow and review, great one @drewkerrigan 👏 

I left some comments/suggestions, the only blocker is the API spec.

Can you create a new Django fixture for this?

api/src/backend/api/migrations/0033_processors.py
vicferpoy: It was designed so a scan could have (or not) one ore more linked processors. That way users could know if their scan was post-processed in some way.

api/src/backend/api/models.py
vicferpoy: Filters are defined in `filters.py` for each view class. If it was not added explicitly, it should not be available.

jfagoagas: It is present in the API reference though 🤔

api/src/backend/api/specs/v1.yaml
drewkerrigan: I think it's because the lighthouse changes didn't include regeneration the yaml spec, so it got included here maybe?

drewkerrigan: Oh, I see the other changes you're talking about. I'll verify those today

api/src/backend/api/v1/serializers.py
drewkerrigan: I tried but there's a problem with it, it blows up when failing validation because it tries to set an instance property which won't work in a static method.

jfagoagas: I don't get that, either way if the SDK is not providing what we need let's fix\adapt it there.

Looking at the code
```python
@staticmethod
    def validate_mutelist(mutelist: dict) -> dict:
        """
        Validate the mutelist against the schema.

        Args:
            mutelist (dict): The mutelist to be validated.

        Returns:
            dict: The mutelist itself.
        """
        try:
            validate(mutelist, schema=mutelist_schema)
            return mutelist
        except Exception as error:
            logger.error(
                f"{error.__class__.__name__} -- Mutelist YAML is malformed - {error}[{error.__traceback__.tb_lineno}]"
            )
            return {}
```
if it fails nothing will be set.

drewkerrigan: Very weird, I swear I tested it and it was trying to set the `self._mutelist_content` to `{}`.

Either way though, the function needs to change so that we can actually capture the exception, since it will give the user details about where the error / syntax error in their YAML is. I think some more time will need to be given to the validation so that it's user friendly.

jfagoagas: The reason you saw that is because it was then changed https://github.com/prowler-cloud/prowler/pull/7811

You can add the replace the following code to use it within the serializer
```python
@staticmethod
    def validate_mutelist(mutelist: dict, raise_on_exception: bool = False) -> dict:
        """
        Validate the mutelist against the schema.

        Args:
            mutelist (dict): The mutelist to be validated.

        Returns:
            dict: The mutelist itself.
        """
        try:
            validate(mutelist, schema=mutelist_schema)
            return mutelist
        except Exception as error:
            if raise_on_exception:
                raise error
            else:
                logger.error(
                    f"{error.__class__.__name__} -- Mutelist YAML is malformed - {error}[{error.__traceback__.tb_lineno}]"
                )
            return {}
```

api/src/backend/api/specs/v1.yaml
drewkerrigan: yaml spec is updated now, please take another look - thanks!

api/src/backend/api/specs/v1.yaml
drewkerrigan: Should be fixed now.

api/src/backend/api/specs/v1.yaml
drewkerrigan: Fixed

api/src/backend/api/specs/v1.yaml
drewkerrigan: Fixed

api/src/backend/api/specs/v1.yaml
drewkerrigan: Fixed

api/src/backend/api/v1/serializer_utils/base.py
drewkerrigan: My assumption was that it is an explicit way of saying that we can accept processor configurations as YAML or JSON. When the config field comes in as a string, we do a `yaml.safe_load(data)` on it to convert to dict.

api/src/backend/api/v1/serializers.py
drewkerrigan: Done!

api/src/backend/api/v1/serializers.py
drewkerrigan: I spoke with @vicferpoy about this, and he has some plans to create an abstract processor validation class, and this will eventually get moved there. For now though this should be OK as it's only repeated twice, and there are actually slight differences because the requests for create and update accept different fields as of now.

api/src/backend/api/v1/serializer_utils/base.py
jfagoagas: Good point, thanks for clarifying it!

api/src/backend/tasks/jobs/scan.py
```diff
@@ -132,10 +133,24 @@ def perform_prowler_scan(
         scan_instance.started_at = datetime.now(tz=timezone.utc)
         scan_instance.save()
 
+    # Find the mutelist processor if it exists
+    with rls_transaction(tenant_id):
+        try:
+            mutelist_processor = Processor.objects.get(
+                tenant_id=tenant_id, processor_type="mutelist"
```
vicferpoy:
```suggestion
                tenant_id=tenant_id, processor_type=Processor.ProcessorChoices.MUTELIST
```

api/src/backend/conftest.py
```diff
@@ -385,6 +386,18 @@ def providers_fixture(tenants_fixture):
     return provider1, provider2, provider3, provider4, provider5
 
 
+@pytest.fixture
+def processor_fixture(tenants_fixture):
+    tenant, *_ = tenants_fixture
+    processor = Processor.objects.create(
+        tenant_id=tenant.id,
+        processor_type="mutelist",
+        configuration="Mutelist:\n  Accounts:\n    *:\n      Checks:\n        iam_user_hardware_mfa_enabled:\n          Regions:\n            - *\n          Resources:\n            - *",
```
vicferpoy:
```suggestion
        configuration="Mutelist:\n  Accounts:\n    *:\n      Checks:\n        iam_user_hardware_mfa_enabled:\n         "
                      " Regions:\n            - *\n          Resources:\n            - *",
```

PEP8.

api/src/backend/api/v1/views.py
```diff
@@ -3492,3 +3502,54 @@ def connection(self, request, pk=None):
...
+@method_decorator(CACHE_DECORATOR, name="list")
+@method_decorator(CACHE_DECORATOR, name="retrieve")
+class ProcessorViewSet(BaseRLSViewSet):
+    queryset = Processor.objects.all()
+    serializer_class = ProcessorSerializer
+    http_method_names = ["get", "post", "patch", "delete"]
+    filterset_class = ProcessorFilter
+    ordering = ["processor_type", "-inserted_at"]
+    # RBAC required permissions
+    required_permissions = []
```
vicferpoy: Should we include `manage_scans` permissions here? Any other? What do you think @jfagoagas ?

api/src/backend/api/tests/test_views.py
```diff
@@ -5708,6 +5709,157 @@ def test_delete_saml_configuration(self, authenticated_client, saml_setup):
...
+        )
+        assert response.status_code == status.HTTP_200_OK
+
+    def test_create_processor_valid(self, authenticated_client):
+        payload = {
+            "data": {
+                "type": "processors",
+                "attributes": {
+                    "processor_type": "mutelist",
+                    "configuration": "Mutelist:\n  Accounts:\n    '*':\n      Checks:\n        iam_user_hardware_mfa_enabled:\n          Regions:\n            - '*'\n          Resources:\n            - '*'",
```
vicferpoy:
```suggestion
                    "configuration": "Mutelist:\n  Accounts:\n    '*':\n      Checks:\n        "
                                     "iam_user_hardware_mfa_enabled:\n          Regions:\n            - '*'\n          "
                                     "Resources:\n            - '*'",
```

PEP8.

api/src/backend/api/specs/v1.yaml
vicferpoy, line 1: This file is still corrupted 🤔 Did you merge the latest changes from `master` recently?

vicferpoy (changes requested):
Looks really good! I left a few suggestions and questions.

api/src/backend/api/v1/views.py
jfagoagas: Great catch, I think in the SaaS was `manage_account` since the mutelist is a tenant/org configuration.

drewkerrigan: Yes true, we had `manage_account` for this.

api/src/backend/api/specs/v1.yaml
drewkerrigan: Yeah I had just pulled them in before regenerating the file - I can try again

drewkerrigan: I pulled in master and nothing changed. The only real problem I see is that `overviews_providers_retrieve` is messed up. Something about the extended schemas?

vicferpoy: I will review this. If I find the issue and fix it, I will try to push a commit here if you don't mind.

api/src/backend/api/models.py
vicferpoy: I confirm that this is not present anymore.

api/src/backend/api/specs/v1.yaml
vicferpoy: Found and fixed it! I'm closing this conversation.

api/src/backend/api/v1/views.py
jfagoagas: So, we need to add it there.

vicferpoy: Done in https://github.com/prowler-cloud/prowler/pull/7993/commits/58d71ea7c2d23143981673213c99df53fad65d92.

jfagoagas:
ToDo for @jfagoagas 
- [x] Review changelog
- [x] Review migrations
- [x] Merge from master

api/src/backend/api/models.py
```diff
@@ -434,8 +421,28 @@ class TriggerChoices(models.TextChoices):
...
+        blank=True,
+    )
+    processor = models.ForeignKey(
+        "Processor",
+        on_delete=models.SET_NULL,
+        related_name="scans",
+        related_query_name="scan",
+        null=True,
+        blank=True,
+    )
```
vicferpoy, line 452: Would it make more sense to define this as M2M?

api/src/backend/api/v1/serializers.py
```diff
@@ -2064,6 +2070,128 @@ def update(self, instance, validated_data):
...
+            "inserted_at": {"read_only": True},
+            "updated_at": {"read_only": True},
+        }
+        validators = [
+            UniqueTogetherValidator(
+                queryset=Processor.objects.all(),
+                fields=["processor_type", "configuration"],
+                message="A processor with the same type already exists.",
+            )
+        ]
```
vicferpoy: Is `configuration` expected as well?

vicferpoy (changes requested):

api/src/backend/api/v1/serializers.py
jfagoagas: Not sure about this. The uniqueness constraint is processor and tenat right? At least for Mutelist that's the constraint but not sure at this point what'd be for others.

api/src/backend/api/models.py
jfagoagas: Sure, in that case it'll require a new table 🤔

jfagoagas: We discussed it IRL and we decided to leave it as is.

api/src/backend/api/v1/serializers.py
jfagoagas: Removed the `configuration` field from the uniqueness constraint in [`00e9d16` (#7993)](https://github.com/prowler-cloud/prowler/pull/7993/commits/00e9d165574b9a294b437756a68d216963cb2f74)

Screenshot 2025-07-08 at 17 03 14

So, each processor is unique per tenant and type.

vicferpoy (approved):
Great job everyone! 🚀

jfagoagas (approved):

PR #7998, opened 2025-06-11
fix(PRWLR-7380): button nesting hydration error

sumit-tft:
### Context

fix: resolve hydration issue by refactoring compliance-card.

### Description

1. Hydration Issue
Previously, the compliance-card worked as a static UI element `div`, which did not expose any hydration issues.
After recent updates, the compliance-card is using `isPressable`,  which makes element a button. This  exposes a hydration warning due to nested interactive elements (button inside button).

2. @paabloLC As discussed, I have included the fix for the issue below in the same PR
image

3. for below warning
image

`Skipping auto-scroll behavior due to position: sticky or position: fixed css`

To address this issue, we need to wrap the Navbar component in an additional  `div`
For reference, I’ve attached the following link, when you have a moment please take a look @paabloLC (https://stackoverflow.com/questions/78345198/skipping-auto-scroll-behavior-due-to-position-sticky-or-position-fixed-css), If this approach looks good, please let me know, I will proceed with the update, Thanks in advance 

### Checklist

### License


ui/components/compliance/compliance-card.tsx
```diff
@@ -105,6 +105,7 @@ export const ComplianceCard: React.FC<ComplianceCardProps> = ({
       shadow="sm"
       isPressable={isPressable}
       onPress={navigateToDetail}
+      as={"div"}
```
paabloLC: We’ll probably find another way to fix this, but for now, to keep things moving, I’m happy with the changes to SelectScanProvider. Could you remove this change from the PR so we can go ahead and merge it? I’ll look into this part myself. What do you think?

paabloLC (changes requested):

paabloLC (approved):

PR #8035, opened 2025-06-16
chore(metadata): add validator for ResourceType

MrCloudSec:
### Description

Add validator for `ResourceType` attribute in the `CheckMetadata` class.

### Checklist

#### API

### License

### DIFF
--- a/prowler/lib/check/models.py
+++ b/prowler/lib/check/models.py
@@ -96,6 +96,7 @@ class CheckMetadata(BaseModel):
         severity_to_lower(severity): Validator function to convert the severity to lowercase.
         valid_severity(severity): Validator function to validate the severity of the check.
         valid_cli_command(remediation): Validator function to validate the CLI command is not an URL.
+        valid_resource_type(resource_type): Validator function to validate the resource type is not empty.
     """
 
     Provider: str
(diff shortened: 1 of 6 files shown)


prowler/lib/check/models.py
```diff
@@ -646,7 +653,6 @@ def load_check_metadata(metadata_file: str) -> CheckMetadata:
         check_metadata = CheckMetadata.parse_file(metadata_file)
     except ValidationError as error:
         logger.critical(f"Metadata from {metadata_file} is not valid: {error}")
-        # TODO: remove this exit and raise an exception
-        sys.exit(1)
+        raise error
```
jfagoagas, line 656: Why this? Are we sure that this is not going to break anything?

MrCloudSec: Raising the exception is actually safer, it allows errors to be handled or tested, instead of abruptly stopping the whole process with `sys.exit(1)`.

jfagoagas: I knew that, only concerned about this changes because you know what happened in other cases, but better if this fails because something was not right.

Thanks!

jfagoagas (approved):

PR #8043, opened 2025-06-17
fix(m365): avoid user requests in `setup_identity` app context and user auth log enhancement

HugoPBrito:
### Context

After reviewing some logs we identified an error in `setup_identity` within the `M365` provider.

The logic needs to be adjusted to prevent entering that section when using non-delegated authentication methods.

### Description

This PR:

1. Adds the necessary logic fixes to avoid the error specified above. Now that section of the code is only accessed via delegated auth methods.
2. Enhances the logs during user authentication so MsGraph errors are shown instead of a generic `User or Password are not correct` error.

### Checklist

#### API

### License


HugoPBrito:
Example of new log:
image

andoniaf (approved):

PR #8056, opened 2025-06-18
feat(integrations): add s3 integration

AdriiiPRodri:
### Context

This pull request introduces the Amazon S3 integration to the Prowler platform. The motivation behind this change is to enhance Prowler’s integrations by enabling the ability to export scan results and findings directly to an S3 bucket. This integration is part of our ongoing efforts to provide flexible and scalable export options for users, facilitating automated workflows and external reporting.

### Description
- Added a new S3 integration module that allows users to configure S3 buckets as export destinations for scan outputs.
- Implemented secure handling of credentials, ensuring that all sensitive information is stored encrypted and decrypted only when required for connection testing or integration usage.
- The integration is designed to be generic, laying the groundwork for future enhancements such as sending specific findings on demand.
- Updated the API and SDK to support S3 integration, including the necessary endpoints and logic for managing integrations.
- Added validation for the ‎`output_directory` parameter.
- Improved error handling for integration updates, particularly in scenarios where providers are added or removed.
- Updated unit tests to cover the new integration logic and ensure reliability.

### Checklist

#### API

### License


api/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ All notable changes to the **Prowler API** are documented in this file.
 - Support GCP Service Account key [(#7824)](https://github.com/prowler-cloud/prowler/pull/7824)
 - `GET /compliance-overviews` endpoints to retrieve compliance metadata and specific requirements statuses [(#7877)](https://github.com/prowler-cloud/prowler/pull/7877)
 - Lighthouse configuration support [(#7848)](https://github.com/prowler-cloud/prowler/pull/7848)
+- Add S3 integration [(#8056)](https://github.com/prowler-cloud/prowler/pull/8056)
```
```suggestion
- S3 integration [(#8056)](https://github.com/prowler-cloud/prowler/pull/8056)
```

prowler/providers/aws/lib/s3/s3.py
```diff
@@ -218,6 +218,27 @@ def send_to_bucket(
             )
         return uploaded_objects
 
+    def upload_file(self, filename: str, bucket_name: str, key: str) -> None:
```
jfagoagas: Since you created this, could you add it as a replacement in lines 185-192?

Also please add tests to this function using `moto`.

api/src/backend/api/utils.py
```diff
@@ -147,6 +151,47 @@ def prowler_provider_connection_test(provider: Provider) -> Connection:
...
+
+    Args:
+        integration (Integration): The integration object containing the integration type and associated credentials.
+
+    Returns:
+        Connection: A connection object representing the result of the connection test for the specified integration.
+    """
+    if integration.integration_type in [
+        Integration.IntegrationChoices.S3,
+        Integration.IntegrationChoices.AWS_SECURITY_HUB,
```
jfagoagas: Is this expected here?

api/src/backend/api/utils.py
```diff
@@ -147,6 +151,47 @@ def prowler_provider_connection_test(provider: Provider) -> Connection:
...
+    ]:
+        try:
+            session = AwsProvider(**integration.credentials).session.current_session
```
jfagoagas: S3 integration allows you to pass the same credentials as the provider and has a `test_connection` method, so you don't need to setup the AWS provider.
```suggestion
            S3.test_connection(**integration.credentials)
```

I see the `test_connection` does not support passing the raw credentials and it should. I'm going to add support to that.

api/src/backend/api/utils.py
```diff
@@ -147,6 +151,47 @@ def prowler_provider_connection_test(provider: Provider) -> Connection:
...
+    """Test the connection to a Prowler integration based on the given integration type.
+
+    Args:
+        integration (Integration): The integration object containing the integration type and associated credentials.
+
+    Returns:
+        Connection: A connection object representing the result of the connection test for the specified integration.
+    """
+    if integration.integration_type in [
+        Integration.IntegrationChoices.S3,
```
jfagoagas: Can you rename `S3` to `AMAZON_S3`?

api/src/backend/api/v1/views.py
```diff
@@ -3405,6 +3406,32 @@ def get_serializer_context(self):
...
+        tags=["Integration"],
+        summary="Check integration connection",
+        description="Try to verify integration connection",
+        request=None,
+        responses={202: OpenApiResponse(response=TaskSerializer)},
+    )
+    @action(detail=True, methods=["post"], url_name="connection")
+    def connection(self, request, pk=None):
+        get_object_or_404(Integration, pk=pk)
+        with transaction.atomic():
```
jfagoagas, line 3852: Do we need this? Because a view under `BaseRLSViewSet` executes the code within a transaction by default.

api/src/backend/tasks/jobs/integrations.py
```diff
@@ -0,0 +1,123 @@
...
+
+    Returns:
+        tuple[bool, S3 | Connection]: A tuple containing a boolean indicating if the connection was successful and the S3 client or connection object.
+    """
+    session = AwsProvider(**integration.credentials).session.current_session
+    s3 = S3(
+        session=session,
+        bucket_name=integration.configuration["bucket_name"],
+        output_directory=integration.configuration["output_directory"],
+    )
```
jfagoagas:
```suggestion
    s3 = S3(
        **integration.credentials,
        bucket_name=integration.configuration["bucket_name"],
        output_directory=integration.configuration["output_directory"],
    )
```

api/src/backend/tasks/jobs/integrations.py
```diff
@@ -0,0 +1,123 @@
...
+from prowler.providers.aws.aws_provider import AwsProvider
+from prowler.providers.aws.lib.s3.s3 import S3
+from prowler.providers.common.models import Connection
+
+logger = get_task_logger(__name__)
+
+
+def get_s3_client_from_integration(
+    integration: Integration,
+) -> tuple[bool, S3 | Connection]:
```
jfagoagas, line 21: Why are you returning this tuple?

api/src/backend/tasks/jobs/integrations.py
```diff
@@ -0,0 +1,123 @@
...
+    )
+    connection = s3.test_connection(
+        session=session,
+        bucket_name=integration.configuration["bucket_name"],
+    )
+
+    if connection.is_connected:
+        return True, s3
+
+    return False, connection
```
jfagoagas: I think you don't need that because the S3 `__init__` actually does a test connection during the setup.

api/src/backend/tasks/jobs/integrations.py
```diff
@@ -0,0 +1,123 @@
...
+            continue
+
+        if connected:
+            try:
+                for filename in os.listdir(file_path):
+                    local_path = os.path.join(file_path, filename)
+                    if not os.path.isfile(local_path):
+                        continue
+                    file_key = f"{integration_output_directory}/{filename}"
+                    s3.upload_file(
```
jfagoagas: Why are you not using SDK's `send_to_bucket`? We should aim to have all this login in the SDK and to adapt whatever is missing for the API to use it. If not you will need to replicate all the logic in there to handle content-type and file paths.

api/src/backend/tasks/tasks.py
```diff
@@ -414,3 +444,102 @@ def check_lighthouse_connection_task(lighthouse_config_id: str, tenant_id: str =
             - 'available_models' (list): List of available models if connection is successful.
     """
     return check_lighthouse_connection(lighthouse_config_id=lighthouse_config_id)
+
+
+@shared_task(
+    name="integration-check",
+    queue="integrations",
+)
+def check_integrations_task(
```
jfagoagas, line 453: I think something like `run_integrations` would be clearer, what do you think? At first I thought this was just checking the integration status.

api/src/backend/tasks/tasks.py
```diff
@@ -356,17 +380,23 @@ def get_writer(writer_map, name, factory, is_last):
...
-        try:
-            rmtree(Path(compressed).parent, ignore_errors=True)
-        except Exception as e:
-            logger.error(f"Error deleting output files: {e}")
+        # TODO: This should be removed, the s3 integration can't work with this block of code, we need to create a new periodic task to delete the output files
+        # In addition, this task shouldn't be responsible for deleting the output files
+        # try:
+        #     rmtree(Path(compressed).parent, ignore_errors=True)
+        # except Exception as e:
+        #     logger.error(f"Error deleting output files: {e}")
```
jfagoagas: We can't delete this code in the PR. This may cause the container file system to run out of space. I agree with you that this should be handled in a different background task but we need to leave it for now.

If we need to share files between tasks we need to a find a place where the data is accessible from different tasks without relying in temporary filesystems.

jfagoagas (changes requested):
That's great Adri! I did a first pass on the code and left some comments/suggestions.

Some questions:
1. Are the integration's credentials stored encrypted? If so, are they decrypted on the fly for test connection and integration usage?
2. Do not forget to add the new `integrations` to the `start_worker` in the `docker-entrypoint.sh`.
  a. Confirmed during end2end tests 
```
worker-dev-1   | [2025-06-19 16:19:42,998: INFO/MainProcess] Task scan-report[99f728cd-7aae-472f-ac11-25a2b473ed63] received
.......
worker-dev-1   | [2025-06-19 16:19:46,264: INFO/ForkPoolWorker-13] scan-report[99f728cd-7aae-472f-ac11-25a2b473ed63]: Scan outputs at /tmp/prowler_api_output/12646005-9067-4d2a-a098-8bb378604362/019788f8-45a6-7fb2-9c5c-36397c196524/prowler-output-254909274388-20250619160037.zip
No more logs from here.
```
4. We need to adapt SDK's code to add whatever is missing and don't add that code into the API.
5. Why an integration requires a provider? It should but not mandatory, you should be able to remove the relationship.
6. Can you add some fixtures for the new integrations?
7. Are we leaving the S3 integration base prepared to allow sending specific findings? This will happen in the near future, so it is important to take it into account.

In addition to the last point, all the integrations are not going to work in the same way, just leaving this here for context since we are still building the base:
- Amazon S3: Send findings after a scan and also send findings on demand.
- AWS Security Hub: Only for the AWS provider and tied to a scan.
- Jira: Send specific findings after a scan and also send findings on demand.


Also, to improve the review experience:
1. Do not forget to fill the PR context and body, adding also details about how to review and test it.
2. We should share PRs only when all tests pass.

api/src/backend/api/migrations/0031_alter_integration_configuration_and_more.py
jfagoagas, line 1: This file needs to be renamed to `0032` because `0031` is the lighthouse configuration migration.

api/src/backend/api/migrations/0031_alter_integration_configuration_and_more.py
```diff
@@ -0,0 +1,26 @@
+# Generated by Django 5.1.10 on 2025-06-17 16:24
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("api", "0030_samlconfigurations"),
```
jfagoagas:
```suggestion
        ("api", "0031_lighthouseconfiguration"),
```

api/src/backend/tasks/tasks.py
```diff
@@ -414,3 +444,102 @@ def check_lighthouse_connection_task(lighthouse_config_id: str, tenant_id: str =
...
+    output_directory = generate_outputs_result.get("output_directory")
+    with rls_transaction(tenant_id):
+        s3_integrations = list(
+            integrations.filter(integration_type=Integration.IntegrationChoices.S3)
+        )
+
+    if s3_integrations and output_directory:
+        logger.info(f"Found {len(s3_integrations)} S3 integration(s)")
+        integration_tasks.append(
+            s3_integration_task.s(
```
jfagoagas: `s` or `si` as we have in other tasks?

api/src/backend/tasks/tasks.py
```diff
@@ -414,3 +444,102 @@ def check_lighthouse_connection_task(lighthouse_config_id: str, tenant_id: str =
...
+                tenant_id=tenant_id,
+                provider_id=provider_id,
+                output_directory=output_directory,
```
jfagoagas: Since the integration task depends on generate_outputs, I’m wondering what happens if they're not executed on the same worker.

If I recall correctly, we previously discussed how integrations should behave. If we avoid making them dependent on other tasks, the integration task would need to loop through all findings again.

However, in a cloud deployment, this approach could introduce issues—especially if something fails mid-process. To handle this reliably, we either need to persist intermediate results using a storage service or ensure that dependent tasks are executed on the same worker.

api/src/backend/tasks/tasks.py
```diff
@@ -414,3 +444,102 @@ def check_lighthouse_connection_task(lighthouse_config_id: str, tenant_id: str =
...
+        provider_id (str): The provider identifier
+        output_directory (str): Directory containing the output files
+    """
+    logger.info(f"Processing S3 integrations for provider {provider_id}")
+
+    try:
+        result = upload_s3_integration(tenant_id, provider_id, output_directory)
+        if result:
+            logger.info(
+                f"All the S3 integrations completed successfully for provider {provider_id}"
```
jfagoagas: I forgot to point the SDK to this branch and I got this error:
```
worker-dev-1   | [2025-06-19 16:35:02,487: ERROR/ForkPoolWorker-6] integration-s3[fae1321a-ae7c-40c3-af9d-02cf12de8f8e]: S3 output upload failed for integration 8740eeb1-789c-4890-93a9-f41d8c71a742: 'S3' object has no attribute 'upload_file'
worker-dev-1   | [2025-06-19 16:35:02,503: ERROR/ForkPoolWorker-6] integration-s3[fae1321a-ae7c-40c3-af9d-02cf12de8f8e]: S3 compliance upload failed for integration 8740eeb1-789c-4890-93a9-f41d8c71a742: 'S3' object has no attribute 'upload_file'
worker-dev-1   | [2025-06-19 16:35:02,517: INFO/ForkPoolWorker-6] integration-s3[fae1321a-ae7c-40c3-af9d-02cf12de8f8e]: All the S3 integrations completed successfully for provider 1370501e-80a3-4b1b-aad8-bdc4ce26066f
```

I think we should not print this line because there was an error:
```
worker-dev-1   | [2025-06-19 16:35:02,517: INFO/ForkPoolWorker-6] integration-s3[fae1321a-ae7c-40c3-af9d-02cf12de8f8e]: All the S3 integrations completed successfully for provider 1370501e-80a3-4b1b-aad8-bdc4ce26066f
```

api/src/backend/tasks/tasks.py
```diff
@@ -222,6 +242,10 @@ def perform_scheduled_scan_task(self, tenant_id: str, provider_id: str):
         generate_outputs.si(
             scan_id=str(scan_instance.id), provider_id=provider_id, tenant_id=tenant_id
         ),
+        check_integrations_task.s(
```
jfagoagas: `s` or `si` as we have in other tasks?

api/src/backend/tasks/tasks.py
```diff
@@ -109,6 +125,10 @@ def perform_scan_task(
         generate_outputs.si(
             scan_id=scan_id, provider_id=provider_id, tenant_id=tenant_id
         ),
+        check_integrations_task.s(
```
jfagoagas: `s` or `si` as we have in other tasks?

jfagoagas (changes requested):
These comments are from the end2end tests I'm doing:

- I don't know if this comes from this PR but there are views with a different style in the API reference
Screenshot 2025-06-19 at 18 10 03
- I'm being able to add several providers to the same integration but not being able to remove all or just some of them. The API always returns `400 Bad Request` with `This integration already exists`. `cURL command:
```
curl --location --request PATCH 'http://127.0.0.1:8080/api/v1/integrations/8740eeb1-789c-4890-93a9-f41d8c71a742' \
--header 'Content-Type: application/vnd.api+json' \
--header 'Accept: application/vnd.api+json' \
--header 'Authorization: Bearer $TOKEN' \
--data '{
  "data": {
    "type": "integrations",
    "id": "8740eeb1-789c-4890-93a9-f41d8c71a742",
    "attributes": {
      "enabled": false,
      "configuration": {
        "bucket_name": "XXXXX",
        "output_directory": "XXXXX"
      },
      "credentials": {
        "aws_access_key_id": "XXXXX",
        "aws_secret_access_key": "XXXXX"
      }
    },
    "relationships": {
      "providers": {
        "data": [
        ]
      }
    }
  }
}'
```
- End2End tests for a connected integration and an AWS account works well 🎉

api/src/backend/tasks/tasks.py
jfagoagas: In addition to the previous comment, we need to find a unified way to support this both on-prem and cloud. We can't introduce a mandatory dependency between `generate_output` and `integration` tasks because they can run in different workers/services/clusters.

AdriiiPRodri:
> That's great Adri! I did a first pass on the code and left some comments/suggestions.
> 
> Some questions:
> 
> 1. Are the integration's credentials stored encrypted? If so, are they decrypted on the fly for test connection and integration usage?
> 2. Do not forget to add the new `integrations` to the `start_worker` in the `docker-entrypoint.sh`.
>    a. Confirmed during end2end tests
> 
> ```
> worker-dev-1   | [2025-06-19 16:19:42,998: INFO/MainProcess] Task scan-report[99f728cd-7aae-472f-ac11-25a2b473ed63] received
> .......
> worker-dev-1   | [2025-06-19 16:19:46,264: INFO/ForkPoolWorker-13] scan-report[99f728cd-7aae-472f-ac11-25a2b473ed63]: Scan outputs at /tmp/prowler_api_output/12646005-9067-4d2a-a098-8bb378604362/019788f8-45a6-7fb2-9c5c-36397c196524/prowler-output-254909274388-20250619160037.zip
> No more logs from here.
> ```
> 
> 4. We need to adapt SDK's code to add whatever is missing and don't add that code into the API.
> 5. Why an integration requires a provider? It should but not mandatory, you should be able to remove the relationship.
> 6. Can you add some fixtures for the new integrations?
> 7. Are we leaving the S3 integration base prepared to allow sending specific findings? This will happen in the near future, so it is important to take it into account.
> 
> In addition to the last point, all the integrations are not going to work in the same way, just leaving this here for context since we are still building the base:
> 
> * Amazon S3: Send findings after a scan and also send findings on demand.
> * AWS Security Hub: Only for the AWS provider and tied to a scan.
> * Jira: Send specific findings after a scan and also send findings on demand.
> 
> Also, to improve the review experience:
> 
> 1. Do not forget to fill the PR context and body, adding also details about how to review and test it.
> 2. We should share PRs only when all tests pass.

Thanks, these are the answers:

1. Yes, in the same way as we decrypt the provider secret and the api key. Every time we get the `credentials` attribute it will be decrypted.
2. Local change missing, I will add it.
4. I didn’t need to add any code to the SDK beyond the upload_file function. Everything that belongs to the API should stay in the API.
5. You can create and update a integration without provider.
6. Sure
7. The S3 integration it's done in the most generic way, this will require some development, like adding a block of code in the S3 integration.

Review experience:
1. Noted.
2. Regarding the tests, the failure was reported last Wednesday in the SDK channel. It’s due to an unresolved vulnerability in safety. This is preventing API code from being merged. All tests were verified before opening the PR.

api/src/backend/tasks/tasks.py
AdriiiPRodri: It needs to be .s because we require the output of the previous task. We use .si when we don’t need anything from the previous task (within a chain)

api/src/backend/tasks/tasks.py
AdriiiPRodri: It needs to be .s because we require the output of the previous task. We use .si when we don’t need anything from the previous task (within a chain)

api/src/backend/tasks/tasks.py
AdriiiPRodri: Good catch, I’ll push the fix later

api/src/backend/tasks/tasks.py
AdriiiPRodri: It needs to be .s because we require the output of the previous task. We use .si when we don’t need anything from the previous task (within a chain)

jfagoagas:
> > That's great Adri! I did a first pass on the code and left some comments/suggestions.
> > Some questions:
> > 
> > 1. Are the integration's credentials stored encrypted? If so, are they decrypted on the fly for test connection and integration usage?
> > 2. Do not forget to add the new `integrations` to the `start_worker` in the `docker-entrypoint.sh`.
> >    a. Confirmed during end2end tests
> > 
> > ```
> > worker-dev-1   | [2025-06-19 16:19:42,998: INFO/MainProcess] Task scan-report[99f728cd-7aae-472f-ac11-25a2b473ed63] received
> > .......
> > worker-dev-1   | [2025-06-19 16:19:46,264: INFO/ForkPoolWorker-13] scan-report[99f728cd-7aae-472f-ac11-25a2b473ed63]: Scan outputs at /tmp/prowler_api_output/12646005-9067-4d2a-a098-8bb378604362/019788f8-45a6-7fb2-9c5c-36397c196524/prowler-output-254909274388-20250619160037.zip
> > No more logs from here.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > 4. We need to adapt SDK's code to add whatever is missing and don't add that code into the API.
> > 5. Why an integration requires a provider? It should but not mandatory, you should be able to remove the relationship.
> > 6. Can you add some fixtures for the new integrations?
> > 7. Are we leaving the S3 integration base prepared to allow sending specific findings? This will happen in the near future, so it is important to take it into account.
> > 
> > In addition to the last point, all the integrations are not going to work in the same way, just leaving this here for context since we are still building the base:
> > 
> > * Amazon S3: Send findings after a scan and also send findings on demand.
> > * AWS Security Hub: Only for the AWS provider and tied to a scan.
> > * Jira: Send specific findings after a scan and also send findings on demand.
> > 
> > Also, to improve the review experience:
> > 
> > 1. Do not forget to fill the PR context and body, adding also details about how to review and test it.
> > 2. We should share PRs only when all tests pass.
> 
> Thanks, these are the answers:
> 
> 1. Yes, in the same way as we decrypt the provider secret and the api key. Every time we get the `credentials` attribute it will be decrypted.
> 2. Local change missing, I will add it.
> 3. I didn’t need to add any code to the SDK beyond the upload_file function. Everything that belongs to the API should stay in the API.
> 4. You can create and update a integration without provider.
> 5. Sure
> 6. The S3 integration it's done in the most generic way, this will require some development, like adding a block of code in the S3 integration.
> 
> Review experience:
> 
> 1. Noted.
> 2. Regarding the tests, the failure was reported last Wednesday in the SDK channel. It’s due to an unresolved vulnerability in safety. This is preventing API code from being merged. All tests were verified before opening the PR.


Thanks for the detailed answers 🙌

api/src/backend/api/migrations/0031_alter_integration_configuration_and_more.py
AdriiiPRodri: I need to rebase this branch with master due to latests changes

api/src/backend/api/migrations/0031_alter_integration_configuration_and_more.py
AdriiiPRodri: I need to rebase this branch with master due to latests changes

AdriiiPRodri:
> These comments are from the end2end tests I'm doing:
> 
> * I don't know if this comes from this PR but there are views with a different style in the API reference
>   Screenshot 2025-06-19 at 18 10 03
> * I'm being able to add several providers to the same integration but not being able to remove all or just some of them. The API always returns `400 Bad Request` with `This integration already exists`. `cURL command:
> 
> ```
> curl --location --request PATCH 'http://127.0.0.1:8080/api/v1/integrations/8740eeb1-789c-4890-93a9-f41d8c71a742' \
> --header 'Content-Type: application/vnd.api+json' \
> --header 'Accept: application/vnd.api+json' \
> --header 'Authorization: Bearer $TOKEN' \
> --data '{
>   "data": {
>     "type": "integrations",
>     "id": "8740eeb1-789c-4890-93a9-f41d8c71a742",
>     "attributes": {
>       "enabled": false,
>       "configuration": {
>         "bucket_name": "XXXXX",
>         "output_directory": "XXXXX"
>       },
>       "credentials": {
>         "aws_access_key_id": "XXXXX",
>         "aws_secret_access_key": "XXXXX"
>       }
>     },
>     "relationships": {
>       "providers": {
>         "data": [
>         ]
>       }
>     }
>   }
> }'
> ```
> 
> * End2End tests for a connected integration and an AWS account works well 🎉

1. Yes, this happens because some endpoints have the documentation schema manually defined, since the plugin we use wasn’t generating it correctly. In those cases, we manually add the tag (like in the SAML case). So we have two options: either consistently add tags ourselves, or try to mimic the plugin’s style when we do it manually. This has been an ongoing issue for quite a while and has come up in discussions multiple times, the plugin doesn’t work as reliably as we’d like
2. You need to remove the configuration if it's the same, we can add a line to avoid this, I’m not really sure how other APIs behave in this kind of situation, but we can definitely discuss it.

api/src/backend/tasks/tasks.py
AdriiiPRodri: We need to discuss this since it involves significantly modifying the integration by applying a temporary patch, only until we have a proper periodic cleanup task in place. It was done this way because the alternative is still just a workaround, and this responsibility doesn’t really belong to the task

api/src/backend/tasks/tasks.py
AdriiiPRodri: What I proposed is that we could add a cronjob in the container to handle this, I believe it’s a better option than adding code to the S3 integration just to work around a problem that shouldn’t exist. We need to create a proper periodic task for this. We can definitely discuss it, but keeping this code would mean introducing significant changes that will have to be removed anyway once the cleanup task is implemented

api/src/backend/tasks/jobs/integrations.py
AdriiiPRodri: This method was created because the API uses a completely different system to manage outputs. While we could replicate what’s in the CLI, it’s not the most efficient approach for the API. So instead of introducing changes just to mirror the CLI, a new function was created. Regarding the content-type handling, in my tests, I haven’t had to worry about it, files are uploaded correctly, the type is set properly, and everything can be viewed and downloaded without any issues.

We can discuss this in more detail and reach an agreement since it’s an important decision

prowler/providers/aws/lib/s3/s3.py
AdriiiPRodri: It hasn’t been changed because I didn’t see the need to explicitly set the content-type. We can talk about this, as it’s possible I’m missing something or we’re not fully aligned

api/src/backend/api/utils.py
AdriiiPRodri: This won’t be necessary after you mentioned that we can use the S3 class directly and that it already performs the test_connection. I’ll need to run some tests to adjust the code to this change, but this if won’t be needed anymore

api/src/backend/tasks/jobs/integrations.py
jfagoagas: I get that. I think we should first analyse the SDK's status prior starting the work and make all the required changes. 

Regarding the `content-type`, we need to set it for each file because AWS sets only the file type.

api/src/backend/api/utils.py
AdriiiPRodri: This was introduced when the integrations model was added, so it’s already in the database and defined as an enum. I believe we can change it without needing an extra migration since it’s only the name of the constant, but I need to test it first

api/src/backend/tasks/tasks.py
jfagoagas: To me there are two things to discuss:
1. Keep the deletion code -- To me we should not handle this in the current PR, if we want to replace it we need to add a solution at the same time.
2. Dependencies between tasks -- If we want to make tasks dependant we need to analyse how to do that.

api/src/backend/api/v1/views.py
AdriiiPRodri: Replicated from what we have in providers, I think it’s no necessary, but I need to test it

api/src/backend/api/utils.py
AdriiiPRodri: Thanks, let me know when this is available

api/src/backend/tasks/jobs/integrations.py
AdriiiPRodri: After mentioning that the S3 class already works for us, this can be changed

api/src/backend/tasks/jobs/integrations.py
AdriiiPRodri: After mentioning that the S3 class already works for us, this can be changed

api/src/backend/api/utils.py
vicferpoy:
```python
        S3 = "amazon_s3", _("Amazon S3")
```

FYI It's just the variable name. It should not require a new migration.

api/src/backend/tasks/tasks.py
AdriiiPRodri: At least for now, this integration is the only one that depends on another. The thing is that with this piece of code, the S3 integration can’t work

jfagoagas: The problem I see here is that we can't introduce that dependency because this will affect both on-prem and cloud deployments.

prowler/providers/aws/lib/s3/s3.py
jfagoagas: The `Content-Type` needs to be set because in AWS by default is `binary/octet-stream`. That's another reason why I recommended you to use the current SDK's S3 methods. We can adapt whatever is needed to reutilize code.

api/src/backend/api/v1/views.py
jfagoagas: I was just curious about it, nothing really important. Probably @vicferpoy knows about it.

api/src/backend/tasks/tasks.py
jfagoagas: Thanks!

api/src/backend/tasks/jobs/integrations.py
AdriiiPRodri: I do not see that the S3 class performs this step for us. It is true that setting the session does check the connection, but the `test_connection` method in S3 verifies additional aspects, such as whether the necessary permissions on the bucket are present

api/src/backend/tasks/jobs/integrations.py
AdriiiPRodri: This approach allows us to obtain the client directly, and if it is not possible, we receive a connection object containing the error details, which can then be logged for further analysis

jfagoagas (changes requested):
@AdriiiPRodri we need to rename migrations and dependencies because the latest is `0043_github_provider`

.pre-commit-config.yaml
```diff
@@ -115,7 +115,7 @@ repos:
       - id: safety
         name: safety
         description: "Safety is a tool that checks your installed dependencies for known security vulnerabilities"
-        entry: bash -c 'safety check --ignore 70612,66963,74429,76352,76353'
```
jfagoagas, line 118: Can you add the reason why these new two are excluded? For the future.

AdriiiPRodri: Done

jfagoagas (approved):

PR #8082, opened 2025-06-23
fix(scan): set scheduler_task to null when removing periodic tasks

vicferpoy:
### Context

Prior to this change, deleting a periodic task entry from the database would remove all the related scheduled scans.

### Description

From now on, the `scheduled_task` field will be set to `null` instead of removing on cascade.

### Checklist

#### API

### License


jfagoagas (approved):
That's great, thanks for this change. I'm approving it because I assume the lighthouse migration change is required for this to work. Please next time create a separate PR just in case we need to revert the code.

PR #8086, opened 2025-06-23
fix: allow raising exceptions from validate_mutelist

drewkerrigan:
### Context

It is helpful for users to see the actual error returned from validating a Mutelist, since it can give clues to what is wrong with the YAML / JSON.

### Description

- Added a `raise_on_exception` arg to the static `validate_mutelist` method (thanks @jfagoagas !).

### Checklist

#### API

### License

### DIFF
--- a/prowler/lib/mutelist/mutelist.py
+++ b/prowler/lib/mutelist/mutelist.py
@@ -439,12 +439,13 @@ class Mutelist(ABC):
             return False
 
     @staticmethod
-    def validate_mutelist(mutelist: dict) -> dict:
+    def validate_mutelist(mutelist: dict, raise_on_exception: bool = False) -> dict:
         """
         Validate the mutelist against the schema.
 
         Args:
             mutelist (dict): The mutelist to be validated.
+            raise_on_exception (bool): Whether to raise an exception if the mutelist is invalid.
 
         Returns:
             dict: The mutelist itself.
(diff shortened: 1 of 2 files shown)


prowler/lib/mutelist/mutelist.py
```diff
@@ -453,7 +454,10 @@ def validate_mutelist(mutelist: dict) -> dict:
...
-            logger.error(
-                f"{error.__class__.__name__} -- Mutelist YAML is malformed - {error}[{error.__traceback__.tb_lineno}]"
-            )
+            if raise_on_exception:
+                raise error
+            else:
+                logger.error(
+                    f"{error.__class__.__name__} -- Mutelist YAML is malformed - {error}[{error.__traceback__.tb_lineno}]"
+                )
             return {}
```
jfagoagas, line 463:
```suggestion
                return {}
```

I think this is missing one level indentation, right?

prowler/lib/mutelist/mutelist.py
```diff
@@ -453,7 +454,10 @@ def validate_mutelist(mutelist: dict) -> dict:
             validate(mutelist, schema=mutelist_schema)
             return mutelist
         except Exception as error:
-            logger.error(
-                f"{error.__class__.__name__} -- Mutelist YAML is malformed - {error}[{error.__traceback__.tb_lineno}]"
-            )
+            if raise_on_exception:
```
jfagoagas, line 457: Can you add a simple test to validate exceptions are raised when called?

prowler/lib/mutelist/mutelist.py
drewkerrigan: Functionally it's the exact same behavior either way, as long as `raise_on_exception` is false, it will always return `{}` in the case that there is an exception. I prefer it this way just to make that clear that we want to return an empty dict in that case.

prowler/lib/mutelist/mutelist.py
drewkerrigan: Done!

jfagoagas (approved):

PR #8088, opened 2025-06-24
refactor(s3): adapt test_connection to match AwsProvider

AdriiiPRodri:
### Description

In this PR, we’ve adapted the test_connection code for S3 to follow the same logic proposed in `AwsProvider`.

To test this, we just need to execute the method the same way we do for `AwsProvider`.

### Checklist

#### API

### License

### DIFF
--- a/prowler/providers/aws/lib/s3/s3.py
+++ b/prowler/providers/aws/lib/s3/s3.py
@@ -1,12 +1,39 @@
+import os
 import tempfile
 from os import path
 from tempfile import NamedTemporaryFile
 from typing import Optional
 
-from botocore import exceptions
+from boto3.session import Session
+from botocore.exceptions import ClientError, NoCredentialsError, ProfileNotFound
 
 from prowler.lib.logger import logger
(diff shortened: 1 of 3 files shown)


prowler/providers/aws/lib/s3/s3.py
```diff
@@ -220,7 +250,18 @@ def send_to_bucket(
 
     @staticmethod
     def test_connection(
-        session, bucket_name: str, raise_on_exception: bool = True
+        bucket_name: str,
```
jfagoagas, line 253: For compatibility we need to keep `session` as the first argument, and leave the rest in the same order.
```
session, bucket_name: str, raise_on_exception: bool = True
```

but now I'm wondering if something is using this method, because it was created for the SDK 🤔

AdriiiPRodri: I removed it because nothing is using this method at the moment

pedrooot (approved):
Thanks for this!

jfagoagas (approved):

PR #8089, opened 2025-06-24
fix(saml): avoid IndexError when some attributes are not specified

AdriiiPRodri:
### Description

We recently discovered that if an attribute is not defined in the IdP, the API raises a 500 error due to an IndexError. In this PR, we’ve fixed that issue. To test this change, configure SAML and set an IdP attribute like "organization" to None.

### Checklist

#### API

### License

### DIFF
--- a/api/src/backend/api/v1/views.py
+++ b/api/src/backend/api/v1/views.py
@@ -518,7 +518,7 @@ class TenantFinishACSView(FinishACSView):
         )
         user.name = f"{user.first_name} {user.last_name}".strip()
         if user.name == "":
-            user.name = "NoName"
+            user.name = "N/A"
         user.save()
 
         email_domain = user.email.split("@")[-1]
--- a/api/src/backend/api/v1/views.py
+++ b/api/src/backend/api/v1/views.py
@@ -527,7 +527,11 @@ class TenantFinishACSView(FinishACSView):
             .get(email_domain=email_domain)
             .tenant
         )
-        role_name = extra.get("userType", ["saml_default_role"])[0].strip()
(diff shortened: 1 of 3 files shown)


api/src/backend/api/v1/views.py
```diff
@@ -509,10 +509,16 @@ def dispatch(self, request, organization_slug):
...
+        user.first_name = (
+            extra.get("firstName", [""])[0] if extra.get("firstName") else ""
+        )
+        user.last_name = extra.get("lastName", [""])[0] if extra.get("lastName") else ""
+        user.company_name = (
+            extra.get("organization", [""])[0] if extra.get("organization") else ""
+        )
         user.name = f"{user.first_name} {user.last_name}".strip()
+        if user.name == "":
+            user.name = "NoName"
```
vicferpoy: Does `NoName` come from any convention or something?

vicferpoy (changes requested):

api/src/backend/api/v1/views.py
AdriiiPRodri: No, we need a value because the name is mandatory. We can define another value

vicferpoy: Use `N/A` if possible then. For consistency with other wildcards.

AdriiiPRodri: Maybe `Unnamed` sounds better

AdriiiPRodri: Ok, we can change it

vicferpoy (approved):

PR #8091, opened 2025-06-24
fix(organizations): `Key Error: Statement` in check `organizations_scp_deny_regions`

danibarranqueroo:
### Context

This PR introduces a fix for the check `organizations_scp_deny_regions`, we noticed that we were having the following error `organizations_scp_check_deny_regions -- KeyError[38]: 'Statement'`

The problem was that when an organization has a policy without statement, the policy.content.get("Statement") return a None and that makes the rest of the code to fail.

For the unit tests I needed to delete the default policy that Moto creates in order to make the test more realistic.

### Description

Modified the check and added an unit test to cover the case.

### Checklist

#### API

### License

### DIFF
--- a/prowler/providers/aws/services/organizations/organizations_scp_check_deny_regions/organizations_scp_check_deny_regions.py
+++ b/prowler/providers/aws/services/organizations/organizations_scp_check_deny_regions/organizations_scp_check_deny_regions.py
@@ -29,76 +29,69 @@ class organizations_scp_check_deny_regions(Check):
                     report.status_extended = f"AWS Organizations {organizations_client.organization.id} does not have SCP policies."
                     # We use this flag if we find a statement that is restricting regions but not all the configured ones:
                     is_region_restricted_statement = False
-                    has_statements = False
 
                     for policy in organizations_client.organization.policies.get(
                         "SERVICE_CONTROL_POLICY", []
                     ):
                         # Statements are not always list
-                        statements = policy.content.get("Statement")
-                        if statements:
-                            has_statements = True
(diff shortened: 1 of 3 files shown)


prowler/providers/aws/services/organizations/organizations_scp_check_deny_regions/organizations_scp_check_deny_regions.py
```diff
@@ -29,69 +29,76 @@ def execute(self):
                     report.status_extended = f"AWS Organizations {organizations_client.organization.id} does not have SCP policies."
                     # We use this flag if we find a statement that is restricting regions but not all the configured ones:
                     is_region_restricted_statement = False
+                    has_statements = False
 
                     for policy in organizations_client.organization.policies.get(
                         "SERVICE_CONTROL_POLICY", []
                     ):
                         # Statements are not always list
                         statements = policy.content.get("Statement")
```
MrCloudSec: @danibarranqueroo I think it is enough by only changing this line
```suggestion
                        statements = policy.content.get("Statement",[])
```

danibarranqueroo: It was needed to also change this line:

```
if type(policy.content["Statement"]) is not list:
                            statements = [policy.content.get("Statement")]
```

Because `policy.content["Statement"]` is not accessible if there is no Statement, but following your approach is easier the solution. Thanks for the suggestion! 🚀

MrCloudSec (approved):

PR #8095, opened 2025-06-24
feat: update SAML login URL handling and redirect logic

alejandrobailo:
### Context

We need a way of being redirected to UI from API when getting SAML credentials 

### Description

- Fixed an error in the creation of the URL that duplicated the "api/v1" string
- Redirect with tokens implemented

For functional testing, @AdriiiPRodri and I can help you.

### Checklist

#### API

### License


alejandrobailo:
@AdriiiPRodri  would this PR need a `CHANGELOG.md` update?

api/src/backend/api/v1/views.py
```diff
@@ -560,12 +562,15 @@ def dispatch(self, request, organization_slug):
...
-                "type": "saml-social-tokens",
-                "attributes": serializer.validated_data,
-            }
-        )
+
+        token_data = serializer.validated_data
+        access_token = token_data.get("access")
+        refresh_token = token_data.get("refresh")
+        auth_url = os.getenv("AUTH_URL", "http://localhost:3000")
+        callback_url = f"{auth_url}/api/auth/callback/saml"
```
vicferpoy: I think this should come from an environment variable that takes `AUTH_URL` as base. But it shouldn't be hardcoded in the backend since this is a URL from the UI.

vicferpoy (changes requested):
Please take a look at the failing checks (unit tests didn't even get to run and I believe they should fail) and review the API specific PR checks, apply those that could be needed and mark them as done. Thank you!

api/src/backend/api/v1/views.py
alejandrobailo: Got it

AdriiiPRodri: If you’re okay with it, I’ll continue working on this PR myself, focusing on the DB changes since we can’t pass the tokens through the URL, that approach isn’t secure. So if it sounds good to you, I’ll take it from here and update the PR accordingly. For now, this PR should have the `no-merge` label

alejandrobailo: Perfect!

AdriiiPRodri:
Related with PRWLR-7479

api/src/backend/api/models.py
```diff
@@ -1370,6 +1372,26 @@ class Meta:
...
+    expires_at = models.DateTimeField(editable=False)
+    token = models.JSONField(unique=True)
+    user = models.ForeignKey(User, on_delete=models.CASCADE)
+
+    class Meta:
+        db_table = "saml_tokens"
+
+    def save(self, *args, **kwargs):
+        if not self.expires_at:
+            self.expires_at = timezone.now() + timedelta(seconds=15)
```
vicferpoy: To prevent timezone issues, we always use this:

```python
from datetime import datetime, timezone
datetime.now(timezone.utc)
```

api/src/backend/api/models.py
```diff
@@ -1370,6 +1372,26 @@ class Meta:
         ]
 
 
+class SAMLToken(models.Model):
```
vicferpoy, line 1374: No RLS at all?

api/src/backend/api/v1/urls.py
```diff
@@ -128,11 +130,17 @@
...
+    path(
+        "api/v1/accounts/saml/<organization_slug>/login/",
+        CustomSAMLLoginView.as_view(),
+        name="saml_login",
+    ),
     path(
         "api/v1/accounts/saml/<organization_slug>/acs/finish/",
         TenantFinishACSView.as_view(),
         name="saml_finish_acs",
     ),
```
vicferpoy: I just noticed. Is this ok? All of these urls are added to the API router so the `api/v1/` prefix should be added by default.

api/src/backend/api/v1/views.py
```diff
@@ -431,10 +486,14 @@ def post(self, request, *args, **kwargs):
...
             )
 
-        saml_login_url = reverse(
+        # Build the SAML login URL using the configured API host
+        api_host = os.getenv("API_BASE_URL")
+        login_path = reverse(
             "saml_login", kwargs={"organization_slug": config.email_domain}
         )
-        return redirect(f"{saml_login_url}?email={email}")
+        login_url = urljoin(api_host, login_path)
```
vicferpoy: Why can't we use something like:

```python
relative = reverse("saml_login", kwargs={"organization_slug": config.email_domain})
login_url = request.build_absolute_uri(relative)
return redirect(login_url)
```

I find having to load the host from a environment variable too strange since we are running the API ourselves.

vicferpoy (changes requested):
I left some questions and suggestions. Thank you!

api/src/backend/api/v1/views.py
AdriiiPRodri: This is exactly what I meant about the UI + API setup, we need to redirect back to the UI, but since the flow is handled entirely in the API, we have no way to redirect to the UI without this

api/src/backend/api/v1/urls.py
AdriiiPRodri: In this case it works, but if it’s being added by default, it should be safe to remove and test

api/src/backend/api/models.py
AdriiiPRodri: No, we’re not logged in at this point, so we need it to work without RLS

api/src/backend/api/models.py
AdriiiPRodri: Changed

api/src/backend/api/v1/views.py
AdriiiPRodri: We had to add this setting to the Django config:
```
USE_X_FORWARDED_HOST = True
```

This is required when using tools like ngrok with the --host-header="rewrite" option. Without it, request.build_absolute_uri() generates URLs using the internal server’s host (e.g. localhost or the container name), which breaks SAML flows where the IdP expects a fully qualified domain (like the ngrok URL).

By enabling USE_X_FORWARDED_HOST, Django respects the X-Forwarded-Host header passed by the reverse proxy (in this case, ngrok), allowing us to build URLs that point to the correct public-facing domain.

We need this for local SAML testing with:
```
ngrok http 8080 --host-header="rewrite"
```

This way, build_absolute_uri() can generate valid redirect URLs during SAML login that actually work with the IdP

vicferpoy (approved):
🚀

AdriiiPRodri (approved):

PR #8131, opened 2025-06-30
fix: ACS dynamic url and password input visible in sign up

alejandrobailo:
### Context

Hot fix related to the ACS URL and password input field.

### Description

- New public env key, then the client can consume it
- Password input is available in sign-up mode

### Checklist

#### API

### License


.env
```diff
@@ -5,6 +5,7 @@
 #### Prowler UI Configuration ####
 PROWLER_UI_VERSION="stable"
 AUTH_URL=http://localhost:3000
+NEXT_PUBLIC_BACKEND_URL=http://prowler-api:8080
```
jfagoagas: Why not to use `API_BASE_URL`? Everything in the API is behind `/api/v1`.

paabloLC (dismissed):

.env
alejandrobailo: env keys without NEXT_PUBLIC prefix are not available client-side

jfagoagas: You're right, thanks for the clarification!

jfagoagas (changes requested):
We need to add the new `NEXT_PUBLIC` environment variable to the `Dockerfile` and also to the Github Actions where the UI container is built.

paabloLC (dismissed):

paabloLC (dismissed):

jfagoagas (approved):

paabloLC (approved):

PR #8186, opened 2025-07-04
feat: add status filter to /overviews endpoint

paabloLC:
### Summary

  Fixed data inconsistency between the findings page and findings_severity endpoint when using status filters. Previously, filtering by status (e.g., filter[status__in]=FAIL) would return different counts between
  the two endpoints.

 ### Changes Made

  - Updated findings_severity endpoint: Redesigned status filtering logic to properly handle aggregated ScanSummary data by summing appropriate fields (fail, _pass, muted) based on requested status filters
  - Enhanced ScanSummaryFilter: Added custom status filter methods to accept and validate filter[status] and filter[status__in] parameters
  - Improved status handling: Now supports all status combinations (FAIL, PASS, MUTED) and their combinations consistently

 ### Technical Details

  Since ScanSummary model uses separate aggregated fields instead of a single status field, the endpoint now:
  - Parses status filter parameters to determine which fields to sum
  - Uses dynamic Django Sum() expressions based on requested statuses
  - Ensures aggregated counts match individual finding counts when filtered

### Checklist

#### API

### License


AdriiiPRodri:
## Status update

After adding this filter and test it the performance is not affected and the indexes are in use:

```
SELECT "scan_summaries"."severity",
       SUM("scan_summaries"."fail") AS "count"
FROM "scan_summaries"
WHERE ("scan_summaries"."tenant_id" = 12646005-9067-4d2a-a098-8bb378604362
       AND "scan_summaries"."scan_id" IN
         (SELECT DISTINCT ON (U0."provider_id") U0."id"
          FROM "scans" U0
          WHERE (U0."state" = completed
                 AND U0."tenant_id" = 12646005-9067-4d2a-a098-8bb378604362)
          ORDER BY U0."provider_id" ASC,
                   U0."inserted_at" DESC)
       AND "scan_summaries"."tenant_id" = 12646005-9067-4d2a-a098-8bb378604362)
GROUP BY "scan_summaries"."severity"
ORDER BY "scan_summaries"."severity" ASC
5.844ms
0 joins
Query Plan
GroupAggregate  (cost=16.38..16.40 rows=1 width=12)
  Group Key: scan_summaries.severity
  ->  Sort  (cost=16.38..16.39 rows=1 width=8)
        Sort Key: scan_summaries.severity
        ->  Nested Loop Semi Join  (cost=0.29..16.37 rows=1 width=8)
              Join Filter: (scan_summaries.scan_id = u0.id)
              ->  Index Scan using ss_tenant_scan_severity_idx on scan_summaries  (cost=0.14..8.18 rows=1 width=24)
                    Index Cond: (tenant_id = '12646005-9067-4d2a-a098-8bb378604362'::uuid)
                    Filter: CASE WHEN (current_setting('api.tenant_id'::text, true) IS NULL) THEN false ELSE (tenant_id = (current_setting('api.tenant_id'::text))::uuid) END
              ->  Result  (cost=0.14..8.18 rows=1 width=40)
                    ->  Unique  (cost=0.14..8.18 rows=1 width=40)
                          ->  Index Only Scan using scans_prov_ins_desc_idx on scans u0  (cost=0.14..8.17 rows=1 width=40)
                                Index Cond: (tenant_id = '12646005-9067-4d2a-a098-8bb378604362'::uuid)
                                Filter: CASE WHEN (current_setting('api.tenant_id'::text, true) IS NULL) THEN false ELSE (tenant_id = (current_setting('api.tenant_id'::text))::uuid) END
```

api/src/backend/api/specs/v1.yaml
```diff
@@ -3506,6 +3506,27 @@ paths:
         description: A search term.
         schema:
           type: string
+      - in: query
+        name: filter[status]
+        schema:
+          type: string
+          enum:
+          - FAIL
+          - MUTED
```
jfagoagas: Is this correct? Muted is not a valid finding status, is another attribute.

api/src/backend/api/specs/v1.yaml
```diff
@@ -3543,7 +3564,7 @@ paths:
     get:
       operationId: overviews_providers_retrieve
       description: Retrieve an aggregated overview of findings and resources grouped
-        by providers. The response includes the count of passed, failed, and manual
+        by providers. The response includes the count of passed, failed, and muted
```
jfagoagas: Probably manual was right as it is valid status. However it's deprecated and no more checks will generate findings with that status.

api/src/backend/api/specs/v1.yaml
AdriiiPRodri: Got it, removed 👍🏼

api/src/backend/api/specs/v1.yaml
AdriiiPRodri: Reverted

AdriiiPRodri:
## Evidences

Captura de pantalla 2025-09-04 a las 13 10 34
Captura de pantalla 2025-09-04 a las 13 10 20
Captura de pantalla 2025-09-04 a las 13 09 57

jfagoagas (approved):

alejandrobailo (approved):
Well done!

PR #8210, opened 2025-07-08
fix(saml): ensure SocialApp and SAMLDomainIndex are deleted with SAMLConfiguration

AdriiiPRodri:
### Context

When a SAMLConfiguration is deleted and then recreated for the same email domain, a `django.core.exceptions.MultipleObjectsReturned` error occurs. This is because the associated SocialApp and `SAMLDomainIndex` records are not being removed when the SAMLConfiguration is deleted, leading to duplicate entries upon recreation.

### Description

This PR ensures that when a SAMLConfiguration instance is deleted, all its related objects are also removed to maintain data integrity and prevent errors.

The main changes include:
- Overriding the delete method in the SAMLConfiguration model to atomically delete the associated SocialApp and SAMLDomainIndex records.
- Adding a new test case to `api/src/backend/api/tests/test_models.py` to verify that the deletion logic works correctly and cleans up all related objects.

This change resolves the bug and allows for the seamless deletion and recreation of SAML configurations.

### Checklist

#### API

### License

### DIFF
--- a/api/src/backend/api/models.py
+++ b/api/src/backend/api/models.py
@@ -1522,6 +1522,12 @@ class SAMLConfiguration(RowLevelSecurityProtectedModel):
             email_domain=self.email_domain, defaults={"tenant": self.tenant}
         )
 
+    def delete(self, *args, **kwargs):
+        super().delete(*args, **kwargs)
+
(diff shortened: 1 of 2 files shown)


vicferpoy (approved):
Really nice catch!

Issue #8211, opened 2025-07-08
[UI] NEXT_PUBLIC_API_BASE_URL requires build-time availability causing authentication failures


Vombato:
### Steps to Reproduce

# 🔄 Steps to reproduce
1. Build UI Docker image without `NEXT_PUBLIC_API_BASE_URL` build argument (or pull `:5.8.0` from DockerHub):
```bash
   docker build --target prod -t prowler-ui:5.8.0 .
```

2. Run with **runtime environment variable** (this should work but doesn't):
```bash
   docker run -e NEXT_PUBLIC_API_BASE_URL=https://your-prowler-api.com/api/v1 prowler-ui:5.8.0
```
3. **Access the UI** and attempt to login/register

# 🔧 Current Workaround
The UI Docker image must be rebuilt with the correct build arguments:
```bash
docker build --target prod \
  --build-arg NEXT_PUBLIC_API_BASE_URL=https://your-prowler-api.com/api/v1 \
  --build-arg NEXT_PUBLIC_PROWLER_RELEASE_VERSION=v5.8.0 \
  -t prowler-ui:5.8.0-fixed .
```

### Expected behavior

The UI should use the `NEXT_PUBLIC_API_BASE_URL` value provided at runtime and successfully communicate with the API endpoint.

### Actual Result with Screenshots or Logs

- UI attempts to connect to `prowler-api` hostname instead of the configured endpoint

- Login/registration forms fail with network errors

- Logs shows: `getaddrinfo ENOTFOUND prowler-api`

- API container logs show no incoming requests from UI

### How did you install Prowler?

Cloning the repository from github.com (git clone)

### Environment Resource

AWS ECS Fargate (but issue affects all deployment methods)
Architecture: linux/amd64

### OS used

MacOS with Docker (Orbstack) to reproduce locally

### Prowler version

5.8.0

### Pip version

///

### Context

In Prowler v5.8.0, the UI switched from using `process.env.API_BASE_URL` to `process.env.NEXT_PUBLIC_API_BASE_URL` for API endpoint configuration. However, Next.js `NEXT_PUBLIC_*` environment variables must be available at Docker build time to be embedded in the JavaScript bundle, not just at runtime.
When `NEXT_PUBLIC_API_BASE_URL` is not provided during build, the variable becomes `undefined` in the client-side code, causing the UI to fall back to the hardcoded hostname `prowler-api` instead of the configured API endpoint.

Will open a PR for a hotfix just in case.

[...]

PR #8212, opened 2025-07-08
fix: add fallback for apiBaseUrl

Vombato:
This PR solves Issue: #8211
Add fallback support for `process.env.API_BASE_URL` while maintaining backward compatibility with `NEXT_PUBLIC_API_BASE_URL`.


paabloLC:
Thanks for raising this!

You're right, starting from v5.8.0, `process.env.NEXT_PUBLIC_API_BASE_URL` was introduced so that the value is accessible from client-side components. And as you mentioned, since `NEXT_PUBLIC_*` variables need to be available at build time, users should rebuild the Docker image with the correct env value to ensure the UI picks it up properly.

I’m closing the PR for now, as the current behavior is expected and a fix is not required. 

Thanks again for the context and the initiative 🙌

PR #8214, opened 2025-07-08
feat(tasks): create overview queue for summaries and overviews

vicferpoy:
### Description

- Summary and overview tasks now use dedicated queue called `overview`.
- Post-scan tasks are now refactored into a function and the compliance task no longer depends on any other to run asynchronously.

image

### Checklist

#### API

### License

### DIFF
--- a/api/src/backend/tasks/tasks.py
+++ b/api/src/backend/tasks/tasks.py
@@ -50,7 +50,7 @@ def _perform_scan_complete_tasks(tenant_id: str, scan_id: str, provider_id: str)
         kwargs={"tenant_id": tenant_id, "scan_id": scan_id}
     )
     chain(
-        perform_scan_summary_task.si(tenant_id, scan_id),
+        perform_scan_summary_task.si(tenant_id=tenant_id, scan_id=scan_id),
         generate_outputs_task.si(
             scan_id=scan_id, provider_id=provider_id, tenant_id=tenant_id
         ),
--- a/api/src/backend/tasks/tasks.py
+++ b/api/src/backend/tasks/tasks.py
@@ -227,7 +227,7 @@ def perform_scheduled_scan_task(self, tenant_id: str, provider_id: str):
     return result
 
 
-@shared_task(name="scan-summary")
+@shared_task(name="scan-summary", queue="overview")
 def perform_scan_summary_task(tenant_id: str, scan_id: str):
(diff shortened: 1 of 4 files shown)


api/src/backend/tasks/tasks.py
```diff
@@ -37,6 +37,26 @@
...
+    """
+    create_compliance_requirements_task.apply_async(
+        kwargs={"tenant_id": tenant_id, "scan_id": scan_id}
+    )
+    chain(
+        perform_scan_summary_task.si(tenant_id=tenant_id, scan_id=scan_id),
+        generate_outputs_task.si(
+            scan_id=scan_id, provider_id=provider_id, tenant_id=tenant_id
+        ),
+    ).apply_async()
```
AdriiiPRodri, line 57: We have reversed the order of the compliance and summary tasks, could this have any impact?

On another note, compliance is now outside the chain, so any errors from this task will not propagate to the subsequent tasks. In principle, this should not be an issue since the following tasks do not depend on compliance, but I wanted to highlight this in case there are any implications

api/src/backend/tasks/tests/test_tasks.py
```diff
@@ -17,7 +18,7 @@ def test_no_findings_returns_early(self):
         with patch("tasks.tasks.ScanSummary.objects.filter") as mock_filter:
             mock_filter.return_value.exists.return_value = False
 
-            result = generate_outputs(
+            result = generate_outputs_task(
```
jfagoagas, line 21: Why this change? Just curious.

AdriiiPRodri (changes requested):

jfagoagas (approved):

api/src/backend/tasks/tasks.py
vicferpoy: That's part of the required work. Compliance task does not depend on any other task nor provides any value to others. We want that task to run no matter what, the sooner the better.

api/src/backend/tasks/tests/test_tasks.py
vicferpoy: Consistency. Every task had that suffix, except this one.

api/src/backend/tasks/tasks.py
AdriiiPRodri: Understood thanks

AdriiiPRodri (approved):

PR #8243, opened 2025-07-10
fix(outputs): identity type and id for prowler cloud

HugoPBrito:
### Context

We’re not generating the reports for Prowler Cloud correctly, as the `Identity Type` and `Identity ID` fields are not being initialized properly.

### Description

This PR fixes the behavior defined above. Now the fields are successfully initialized.

### Checklist

#### API

### License


prowler/providers/m365/m365_provider.py
```diff
@@ -954,13 +955,20 @@ async def get_m365_identity(identity):
...
             # since that exception is not considered as critical, we keep filling another identity fields
-            identity.identity_id = (
-                getenv("AZURE_CLIENT_ID") or "Unknown user id (Missing AAD permissions)"
-            )
             if sp_env_auth:
                 identity.identity_type = "Service Principal"
+                identity.identity_id = (
+                    getenv("AZURE_CLIENT_ID")
+                    or session.credentials[0]._credential.client_id
+                    or "Unknown user id (Missing AAD permissions)"
```
AdriiiPRodri, line 963: Is it possible that the AZURE_CLIENT_ID is not set and that session.credentials does not exist or is an empty list? If so, we should handle a potential IndexError to prevent unexpected crashes in these cases

prowler/providers/m365/m365_provider.py
```diff
@@ -978,6 +986,10 @@ async def get_m365_identity(identity):
                     logger.error(
                         f"{error.__class__.__name__}[{error.__traceback__.tb_lineno}] -- {error}"
                     )
+            else:
+                # Prowler Cloud Static Credentials
+                identity.identity_type = "Service Principal"
+                identity.identity_id = session._client_id
```
AdriiiPRodri: I do not fully understand the Prowler Cloud section. Should this code actually be included in this repository?

AdriiiPRodri (changes requested):

MrCloudSec (dismissed):

prowler/providers/m365/m365_provider.py
MrCloudSec: This case is for when there are no input flags, right?

prowler/providers/m365/m365_provider.py
HugoPBrito: It should not happen since that would mean an error inside `setup_session` functions, which runs before `setup_identity`.

prowler/providers/m365/m365_provider.py
HugoPBrito: This refers to Prowler APP as well, my bad. I'll let is as Static Credentials so there's no confusion.

prowler/providers/m365/m365_provider.py
AdriiiPRodri: Nice 👌🏼

AdriiiPRodri (approved):

PR #8260, opened 2025-07-14
fix(saml): remove user in case of error

AdriiiPRodri:
### Context

This pull request addresses an issue related to SAML user creation in the authentication flow. Previously, if a configuration or validation error occurred during SAML login, a user could remain partially created in the database, leading to inconsistencies and potential security concerns. The motivation for this change is to ensure that the user is properly removed if any error arises during the SAML authentication process, maintaining data integrity and improving error handling.

### Description

This update introduces logic to rollback and remove a user from the database if a configuration error or validation failure occurs during SAML authentication. Specifically, a helper function ‎`_rollback_saml_user` has been added to handle user deletion when necessary, and it is invoked in key parts of the SAML authentication flow where exceptions or failed authentications may happen. Detailed logging has also been added to facilitate debugging and monitoring of such events. There are no new external dependencies introduced with this change.

### Checklist

#### API

### License

### DIFF
--- a/api/src/backend/api/v1/views.py
+++ b/api/src/backend/api/v1/views.py
@@ -567,10 +567,7 @@ class TenantFinishACSView(FinishACSView):
         """Helper function to rollback SAML user if it was just created and validation fails"""
         saml_user_id = request.session.get("saml_user_created")
         if saml_user_id:
-            try:
-                User.objects.using(MainRouter.admin_db).filter(id=saml_user_id).delete()
-            except User.DoesNotExist:
-                pass
+            User.objects.using(MainRouter.admin_db).filter(id=saml_user_id).delete()
             request.session.pop("saml_user_created", None)
 
     def dispatch(self, request, organization_slug):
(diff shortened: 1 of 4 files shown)


api/src/backend/api/v1/views.py
```diff
@@ -559,10 +563,28 @@ def get_queryset(self):
...
 
 class TenantFinishACSView(FinishACSView):
+    def _rollback_saml_user(self, request):
+        """Helper function to rollback SAML user if it was just created and validation fails"""
+        saml_user_id = request.session.get("saml_user_created")
+        if saml_user_id:
+            try:
+                User.objects.using(MainRouter.admin_db).filter(id=saml_user_id).delete()
+            except User.DoesNotExist:
+                pass
```
vicferpoy: This is not necessary with `filter`.

vicferpoy (changes requested):

api/src/backend/api/v1/views.py
AdriiiPRodri: But I'm doing a deletion with `.delete()`

AdriiiPRodri: Done

vicferpoy (approved):

PR #8272, opened 2025-07-15
fix(users): Add user deletion error handling to return expected respo…

sumit-tft:
### Context

The user deletion API previously returned unstructured error responses, such as plain HTML or raw text, causing issues with frontend error handling and toast notification consistency. This update addresses the issue by returning a structured JSON error response.

### Description

This update improves the user deletion API by:
- Wrapping the deletion logic in a try-except block to handle unexpected errors gracefully.
- Replacing raw or unstructured error responses (e.g., HTML or plain text) with a consistent JSON format.
- Returning structured error fields: `detail`, `status`, and `code`.
- Enabling the frontend to reliably detect failures and display consistent toast notifications to users.

image

#### API

### License

### DIFF
--- a/api/src/backend/api/v1/views.py
+++ b/api/src/backend/api/v1/views.py
@@ -791,7 +791,6 @@ class UserViewSet(BaseUserViewset):
                         "detail": "Failed to delete the user",
                         "status": "400",
                         "code": "delete_failed"
-
                     }
                 ]
             )
--- a/api/src/backend/api/v1/views.py
+++ b/api/src/backend/api/v1/views.py
@@ -783,14 +783,14 @@ class UserViewSet(BaseUserViewset):
         try:
             return super().destroy(request, *args, **kwargs)
         except Exception as e:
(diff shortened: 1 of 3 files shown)


api/src/backend/api/v1/views.py
```diff
@@ -780,7 +780,18 @@ def destroy(self, request, *args, **kwargs):
...
+        except Exception:
+            raise ValidationError(
+                [
+                    {
+                        "detail": "Failed to delete the user",
+                        "status": "400",
+                        "code": "delete_failed",
+                    }
+                ]
+            )
```
AdriiiPRodri, line 794: Actually we don't use the "code" key, and by default this is a 400 error. So, to follow the schema I recommend using this:
```suggestion
            raise ValidationError("Failed to delete the user")
```

AdriiiPRodri (changes requested):
Great job @sumit-tft! 🚀

PR #8274, opened 2025-07-15
fix(aws): make `is_service_role` more restrictive to avoid false positives

kagahd:
### Context

The original `is_service_role()` function in Prowler returned `True` if *any* statement in a role's `AssumeRolePolicyDocument` contained:

- `"Effect": "Allow"`
- An `Action` of `sts:AssumeRole`, `sts:*`, or `*`
- A `"Service"` principal

This could be bypassed simply by adding an additional `"Service"` principal to an otherwise non-service role, which would incorrectly cause several Prowler checks to be skipped (e.g., trust policy or inline policy validation).

### Description

The updated function enforces **stricter matching**:

- All statements must be `Allow`
- All statements must contain only one `Principal`, and it must be of type `"Service"` (i.e., no additional `AWS`, `Federated`, or other principals)
- All statements must permit `sts:AssumeRole` (or `sts:*`, `*`)

This ensures the role is **exclusively** assumable by services, not mixed with cross-account or organization trust.

By ensuring a stricter definition of service roles, we prevent intentional or accidental misclassification of mixed trust policies as service roles — which could otherwise result in important IAM checks being skipped.

### Checklist

#### API

### License


MrCloudSec (dismissed):
Good catch @kagahd , I added tests for this use case.

MrCloudSec (approved):

PR #8281, opened 2025-07-15
fix(users): only list roles and memberships with manage_account

jfagoagas:
### Description

- `UserRoleRelationshipView` requires `MANAGE_ACCOUNT` instead of `MANAGE_USERS`

### Checklist

#### API

### License


josema-xyz (dismissed):
Learning about the `SerializerMethodResourceRelatedField` from `djangorestframework-jsonapi`, thanks!

alejandrobailo (dismissed):

AdriiiPRodri (changes requested):
@alejandrobailo with the `/me` endpoint we can already fetch the user's role using include, so we should update the profile page to use that as well.

Edit: I’ve seen that the issue is due to a request to the endpoint `/users/${userId}/memberships`, which requires `MANAGE_ACCOUNT`. However, this information is already available in the `/me` response. So I will add the memberships in the `/me` include and then the UI needs to be updated

Edit2: Done here https://github.com/prowler-cloud/prowler/pull/8281/commits/ce10aa23f17012edefee64e1738bb377d173d83e

api/CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 All notable changes to the **Prowler API** are documented in this file.
 
+## [1.13.1] (Prowler 5.12.2)
```
vicferpoy: I believe we should release a new minor version for this.

vicferpoy (changes requested):
Small suggestion, other than that LGTM! 🚀

api/CHANGELOG.md
AdriiiPRodri: Sure, thanks!

vicferpoy (approved):

vicferpoy (approved):

alejandrobailo (approved):

AdriiiPRodri (approved):

PR #8308, opened 2025-07-17
feat(api & ui): prwlr-7478 api key

mmuller88:
### Context

Implementing a Prowler API Key accordingly to the [RFC API Key Management for Prowler](https://prowlerpro.atlassian.net/wiki/x/CYBMuw)

Issue [PRWLR-7478](https://prowlerpro.atlassian.net/browse/PRWLR-7478)

### Description

Prowler currently supports authentication through JWT tokens for web UI access and programmatic API access. However, many users and integration scenarios require a more straightforward authentication method for CLI tools, CI/CD pipelines, and third-party integrations.

Current limitations:
* JWT tokens have short lifespans requiring frequent refresh
* No persistent credentials for automated systems
* Complex authentication flow for simple API integrations
* No way to create scoped credentials for specific use cases

Users frequently request the ability to:
* Generate long-lived credentials for automation
* Create multiple keys for different applications
* Revoke access without changing passwords
* Track API usage per integration

HDI ThreadScope would like to use API Keys which provide the token on behalf of the broker

Goals
Security: Implement API keys with industry-standard security practices (hashing, one-time display, audit trails, tenant RBAC managing permissions)

Usability: Provide intuitive UI/UX for creating and managing API keys similar to GitHub/AWS

Flexibility: Support optional expiration, multiple keys per tenant, and granular revocation

Compatibility: Seamlessly integrate with existing authentication system without breaking changes

### Checklist

#### API

### License

[PRWLR-7478]: https://prowlerpro.atlassian.net/browse/PRWLR-7478?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

DOC:
prowlerApiKeyDocs

UI:
ProwlerApiKeyUi_1_list_no_keys
ProwlerApiKeyUi_2_create_key
ProwlerApiKeyUi_3_key_created
ProwlerApiKeyUi_4_see_list_of_keys


api/src/backend/api/authentication.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+class APIKeyAuthentication(authentication.BaseAuthentication):
+    """
+    Simple API key authentication.
+    
+    Clients should authenticate by passing the API key in the "Authorization"
+    HTTP header, prepended with the string "ApiKey ". For example:
+    
+        Authorization: ApiKey pk_abcdef.1234567890abcdef...
```
jfagoagas: Why not to use the same separator instead of `_` and `.`?

api/CHANGELOG.md
```diff
@@ -7,6 +7,22 @@ All notable changes to the **Prowler API** are documented in this file.
 ### Added
 - SSO with SAML support [(#8175)](https://github.com/prowler-cloud/prowler/pull/8175)
 - `GET /resources/metadata`, `GET /resources/metadata/latest` and `GET /resources/latest` to expose resource metadata and latest scan results [(#8112)](https://github.com/prowler-cloud/prowler/pull/8112)
+- **API Key Authentication System**: Comprehensive API key functionality for programmatic access
```
jfagoagas: Changelog entries must be one line with natural language. Also add one for the UI.

api/src/backend/api/authentication.py
```diff
@@ -0,0 +1,83 @@
...
+    keyword = 'ApiKey'
+    
+    def authenticate(self, request):
+        auth_header = authentication.get_authorization_header(request).decode('utf-8')
+        
+        if not auth_header:
+            return None
+            
+        try:
+            auth_type, api_key = auth_header.split(' ', 1)
```
jfagoagas: Is there no other choice to get the Authentication Header API Key value?

api/src/backend/api/middleware.py
```diff
@@ -33,17 +44,105 @@ def __call__(self, request):
...
+                try:
+                    response_size = int(response['Content-Length'])
+                except (ValueError, TypeError):
+                    pass
+            
+            # Convert duration to milliseconds
+            duration_ms = int(duration * 1000) if duration is not None else None
+            
+            # Create the activity record
+            APIKeyActivity.objects.create(
```
jfagoagas: How do you plan to cleanup records in that table?

api/src/backend/api/migrations/0040_add_api_key_model.py
```diff
@@ -0,0 +1,118 @@
...
+                ('tenant_id', models.UUIDField(help_text='Tenant ID for multi-tenancy support')),
+            ],
+            options={
+                'db_table': 'api_keys',
+            },
+        ),
+        migrations.AddIndex(
+            model_name='apikey',
+            index=models.Index(fields=['prefix'], name='api_keys_prefix_idx'),
+        ),
```
jfagoagas: All indexes under RLS must have `tenant_id` because queries are going to use that filter.

api/src/backend/api/migrations/0040_add_api_key_model.py
```diff
@@ -0,0 +1,118 @@
...
+            index=models.Index(fields=['endpoint', '-timestamp'], name='api_key_activity_endpoint_time_idx'),
+        ),
+        migrations.AddIndex(
+            model_name='apikeyactivity',
+            index=models.Index(fields=['status_code', '-timestamp'], name='api_key_activity_status_time_idx'),
+        ),
+        migrations.AddIndex(
+            model_name='apikeyactivity',
+            index=models.Index(fields=['tenant_id', 'api_key', 'source_ip', '-timestamp'], name='api_key_activity_incident_idx'),
+        ),
```
jfagoagas: Do you need all of this? Are we going to support something in the app to check those logs? If not, I think we should not optimize until we know the queries we're going to use. I think for now this is only for auditing purposes getting data directly from the database.

api/src/backend/api/models.py
```diff
@@ -156,6 +156,213 @@ class JSONAPIMeta:
...
+        from django.utils import timezone
+        self.revoked_at = timezone.now()
+        self.save()
+
+    @classmethod
+    def generate_key(cls):
+        """Generate a new API key with a unique prefix."""
+        # Generate a unique prefix to avoid collisions
+        # Use 8 characters for good uniqueness (62^8 = ~218 trillion combinations)
+        prefix = generate_random_token(8)
```
jfagoagas: did you verify that this is using `secrets.token_urlsafe` underneath?

api/src/backend/config/django/base.py
```diff
@@ -247,6 +247,14 @@
 CACHE_MAX_AGE = env.int("DJANGO_CACHE_MAX_AGE", 3600)
 CACHE_STALE_WHILE_REVALIDATE = env.int("DJANGO_STALE_WHILE_REVALIDATE", 60)
 
+# Django Cache Configuration (using same Valkey instance as Celery)
+CACHES = {
```
jfagoagas: What is this about? We don't want to use the same Valkey/Redis we have for the task broker. If we want to start using a cache we should create a new instance but we need to talk.

jfagoagas (commented):
I did a high level pass, I was curious about this new feature. Both API and UI teams will review this.

Thanks for this @mmuller88 👏

api/src/backend/api/authentication.py
mmuller88: Hi @jfagoagas a very good questions :) ! 

The dot provides clear visual separation between the prefix and main key sections. Since the format already uses pk_ as a prefix, using another underscore would create: pk_XXXXXXXX_YYYYYYYY, making it harder to visually distinguish the different parts.

api/CHANGELOG.md
mmuller88: done

api/src/backend/api/migrations/0040_add_api_key_model.py
mmuller88: good point. I removed those

api/src/backend/config/django/base.py
mmuller88: @jfagoagas Fair point. I removed. Let's stay simple for now.

api/src/backend/api/models.py
mmuller88: No I was using `secrets.choice(symbols or _symbols)` . I wasn't aware of `secrets.token_urlsafe` which seems to be a better choice. I switched to token_urlsafe. Thank you so much :)

api/src/backend/api/migrations/0040_add_api_key_model.py
mmuller88: Nice catch thank you so much. adjusted it.

api/src/backend/api/authentication.py
mmuller88: There are other choices like Regular Expression Parsing or Manual String Slicing but the current approach 

* It's simple and reliable
* Follows HTTP authentication header conventions (Type token)
* Used by DRF's built-in authentication classes
* Handles edge cases gracefully with split(' ', 1) (only splits on first space)

api/src/backend/api/middleware.py
mmuller88: @jfagoagas  good point. I didn't thought about that. I implemented now the PostgresPartitioningConfig which gives as an elegant way of partitioning and deleting after 12 months

ui/components/users/profile/api-keys-card.tsx
```diff
@@ -0,0 +1,324 @@
...
+            break;
+          case "30days":
+            date.setDate(date.getDate() + 30);
+            break;
+          case "90days":
+            date.setDate(date.getDate() + 90);
+            break;
+        }
+        expires_at = date.toISOString();
+      }
```
sumit-tft: I would recommend to use a local `enum` for these values something like 
```
[
  { label: 'never', value: 1 },
  { label: '1 Day', value: 2 },
  { label: '7 Days', value: 3 },
  { label: '30 Days', value: 4 },
  { label: '90 Days', value: 5 },
]
```
something like this so it will be more easy to handle numerical key values, and also later we can move this `enum` in some common place to use.

ui/components/users/profile/api-keys-card.tsx
```diff
@@ -0,0 +1,324 @@
...
+                  >
+                    <SelectTrigger id="expiration">
+                      <SelectValue />
+                    </SelectTrigger>
+                    <SelectContent>
+                      <SelectItem value="never">Never</SelectItem>
+                      <SelectItem value="1day">1 day</SelectItem>
+                      <SelectItem value="7days">7 days</SelectItem>
+                      <SelectItem value="30days">30 days</SelectItem>
+                      <SelectItem value="90days">90 days</SelectItem>
```
sumit-tft: After implementing the above `enum` you can use that in these dropdown :)

sumit-tft:
@mmuller88 Thank you for the changes you did, I added a small recommendation, also wanted to ask do we have any confirmation popup when revoking the access key? if yes can you please add a snapshot it also :) Thanks!

ui/components/ui/select/index.ts
alejandrobailo, line 1: We don't need this file

ui/components/ui/label/index.ts
alejandrobailo, line 1: We don't need this file

ui/components/ui/select/index.ts
alejandrobailo, line 1: We don't need this file

ui/components/ui/dialog/index.ts
alejandrobailo, line 1: We don't need this file

ui/components/ui/button/index.ts
alejandrobailo, line 1: We don't need this file

ui/components/users/profile/api-keys-card.tsx
alejandrobailo: Yes, please iterate the enum here.

ui/components/users/profile/api-keys-card.tsx
```diff
@@ -0,0 +1,324 @@
...
+                    </SelectContent>
+                  </Select>
+                </div>
+              </div>
+              <DialogFooter>
+                <Button
```
alejandrobailo: We must use `CustomButtom` across the whole app instead of `Button`

ui/types/users.ts
```diff
@@ -149,3 +149,40 @@ export interface TenantDetailData {
     };
   };
 }
+
+export interface RoleRelationship {
+  id: string;
+  type: "roles";
```
alejandrobailo: Few lines above we have this type:
```ts
export interface RoleData {
  type: "roles";
  id: string;
}
```

alejandrobailo (changes requested):

.github/workflows/api-pull-request.yml
cesararroba, line 1: Why this change? Isn't it out of scope?

.github/workflows/build-documentation-on-pr.yml
cesararroba, line 1: Why this change? Isn't it out of scope?

cesararroba (changes requested):

.github/workflows/api-pull-request.yml
mmuller88: @cesararroba Yes it is! But unlucky I had todo it because the check wouldn't pass as it wasn't fetching the code from my forked repository. I can revert it but than it will fail again.

.github/workflows/build-documentation-on-pr.yml
mmuller88: @cesararroba Yes it is out of scope! But unlucky I had todo it because the check wouldn't pass as it had permission issues when working with my forked repository.

api/src/backend/api/authentication.py
mmuller88: In the meantime, I notices as well that the hash itself can contain `_` as well. One reason more why we need to use `.` for seperating the hash.

ui/components/ui/label/index.ts
mmuller88: removed it

ui/components/ui/select/index.ts
mmuller88: removed it

ui/components/ui/dialog/index.ts
mmuller88: removed it

ui/components/ui/button/index.ts
mmuller88: removed it

ui/components/users/profile/api-keys-card.tsx
mmuller88: @sumit-tft good idea did that

ui/components/users/profile/api-keys-card.tsx
mmuller88: Did that

ui/types/users.ts
mmuller88: @alejandrobailo great catch. Deleted it and use the RoleData interface

ui/components/users/profile/api-keys-card.tsx
mmuller88: Great I changed to CustomButton and aligned the styling with other buttons I found close by

mmuller88:
> confirmation popup when revoking the access key

I added the popup
ProwlerApiKeyRevokeApiKeyPopup

jfagoagas:
@mmuller88 I'm closing this as we decided to create a PR for each iteration towards API Keys support. Closing the PR does not delete the content from Github so everything will be here for us to check if needed.

PR #8319, opened 2025-07-18
feat(ui): add 32-character limit validation for scan name in create a…

sumit-tft:
### Context

This PR addresses consistent validation for the scan name input field during scan creation and scan name editing.

### Description

Added validation for the `scan name` input field in both scan launch and scan name editing.

Enforced the following rules:
- Minimum 3 characters (when not empty)
- Maximum 32 characters
- Allows empty string when scan name is optional

image

image

image

image

### Checklist

### License


alejandrobailo (changes requested):
Nice work, I would ask for a change:

https://github.com/user-attachments/assets/642d262a-f4a7-437c-a186-5489b3f6fa21

As you can see in the video, the input moves up when an error appears; it should be in the same place, no movement, and the error has to appear under it without moving nothing.

I know is not part of the task, but I know you can improve that. Input

sumit-tft:
Thanks for the review @alejandrobailo. I’ve updated the suggested changes. Please let me know if anything else needs to be updated.

https://github.com/user-attachments/assets/4235bb74-2432-48b7-b1b4-08fba66e6ff6

alejandrobailo (approved):
Great job, @sumit-tft !

PR #8330, opened 2025-07-18
fix(aws): avoid false positives in SQS encryption check for ephemeral queues

kagahd:
### Context

Prowler's check `sqs_queues_server_side_encryption_enabled` iterates over all discovered SQS queues and verifies if server-side encryption (SSE) is enabled by checking the presence of a `kms_key_id`. However, in environments where SQS queues are ephemeral (e.g., created per ECS task and deleted shortly after), this check can produce false positives.

The root cause is that the queue may exist at the time of listing (via `list_queues`) but may already be deleted before `get_queue_attributes` or `list_queue_tags` is called. If the queue disappears in this window, AWS raises a `NonExistentQueue` error. In the current implementation, the queue object remains in memory and is treated as unencrypted due to the missing `kms_key_id`, even though the queue no longer exists.

### Description

This PR modifies the `SQS._get_queue_attributes` method to filter out queues that no longer exist when attempting to fetch their attributes. Specifically:

- If a `ClientError` or `Exception` occurs during `get_queue_attributes`, the affected queue is skipped and not added to the list of valid queues.
- This prevents race-condition-induced false positives from being reported.

**Note:** The `_list_queue_tags` method currently does not filter out queues that disappear during the tag-fetching step. While this is acceptable for now (as there are no Prowler checks that depend on SQS tags), it may need to be revisited if any future Prowler checks evaluate SQS tags (e.g., for ownership, environment classification, or cost attribution), then a similar filtering strategy should be applied in `_list_queue_tags` to avoid tag-related false positives.

### Checklist

#### API

### License


prowler/CHANGELOG.md
```diff
@@ -2,10 +2,19 @@
 
 All notable changes to the **Prowler SDK** are documented in this file.
 
-## [v5.9.2] (Prowler v5.9.2) UNRELEASED
+
```
HugoPBrito:
```suggestion
```

prowler/CHANGELOG.md
```diff
@@ -2,10 +2,19 @@
...
+## [v5.9.2] (Prowler UNRELEASED)
 
 ### Fixed
 - Use the correct resource name in `defender_domain_dkim_enabled` check [(#8334)](https://github.com/prowler-cloud/prowler/pull/8334)
+- Avoid false positives in SQS encryption check for ephemeral queues [(#8330)](https://github.com/prowler-cloud/prowler/pull/8330)
```
HugoPBrito:
```suggestion
- False positives in SQS encryption check for ephemeral queues [(#8330)](https://github.com/prowler-cloud/prowler/pull/8330)
```

prowler/providers/aws/services/sqs/sqs_service.py
```diff
@@ -72,6 +73,7 @@ def _get_queue_attributes(self):
                                 == "true"
                             ):
                                 queue.kms_key_id = "SqsManagedSseEnabled"
+                    valid_queues.append(queue)
                 except ClientError as error:
```
HugoPBrito, line 77: Since we know the exact error code is `NonExistentQueue`, we should specifically check for that error. Other exceptions might occur during the process, and it’s better to apply the fix only when we’re certain the queue no longer exists.

HugoPBrito (changes requested):
Please review the comments and let me know what you think.

prowler/providers/aws/services/sqs/sqs_service.py
```diff
@@ -84,10 +86,13 @@ def _get_queue_attributes(self):
                         logger.error(
                             f"{regional_client.region} -- {error.__class__.__name__}[{error.__traceback__.tb_lineno}]: {error}"
                         )
+                        valid_queues.append(queue)
                 except Exception as error:
                     logger.error(
                         f"{regional_client.region} -- {error.__class__.__name__}[{error.__traceback__.tb_lineno}]: {error}"
                     )
+                    valid_queues.append(queue)
```
HugoPBrito, line 94:
```suggestion
logger.error(
    f"{regional_client.region} -- {error.__class__.__name__}[{error.__traceback__.tb_lineno}]: {error}"
)
if error.response['Error']['Code'] != 'AWS.SimpleQueueService.NonExistentQueue':
    valid_queues.append(queue)
```

HugoPBrito: I meant something like this. That way, if we know the error is because the queue no longer exists, we can safely ignore it. If the error is caused by anything else, we continue to take that queue into account, since it still exists.

HugoPBrito (changes requested):

prowler/providers/aws/services/sqs/sqs_service.py
kagahd: > if we know the error is because the queue no longer exists, we can safely ignore it. 

That's what the code actually does. If the error is `NonExistentQueue`, the queue won't be added to `valid_queues`. In all other cases, even error cases such as `ClientError` and `Exception` the queue is added to `valid_queues`.

HugoPBrito (dismissed):
Nice!

HugoPBrito (approved):

MrCloudSec (changes requested):
Good catch @kagahd, why not doing this only in the event of `NonExistentQueue` error?

kagahd:
> Good catch @kagahd, why not doing this only in the event of `NonExistentQueue` error?

@MrCloudSec, because in the event of `NonExistentQueue` the queue should **not** be added to the `valid_queues`.
If you want to inverse the logic, you'd have to **remove** the queue in the event of `NonExistentQueue` but removing an element from a list while being iterated will corrupt the iteration over the list.

MrCloudSec:
> > Good catch @kagahd, why not doing this only in the event of `NonExistentQueue` error?
> 
> @MrCloudSec, because in the event of `NonExistentQueue` the queue should **not** be added to the `valid_queues`. If you want to inverse the logic, you'd have to **remove** the queue in the event of `NonExistentQueue` but removing an element from a list while being iterated will corrupt the iteration over the list.

@kagahd I meant that errors such as `AccessDenied` in `get_queue_attributes` should not make a queue invalid and making it not appearing in the findings. I should make it invalid only in the case you mentioned of `NonExistentQueue` error.

MrCloudSec:
> > > Good catch @kagahd, why not doing this only in the event of `NonExistentQueue` error?
> > 
> > 
> > @MrCloudSec, because in the event of `NonExistentQueue` the queue should **not** be added to the `valid_queues`. If you want to inverse the logic, you'd have to **remove** the queue in the event of `NonExistentQueue` but removing an element from a list while being iterated will corrupt the iteration over the list.
> 
> @kagahd I meant that errors such as `AccessDenied` in `get_queue_attributes` should not make a queue invalid and making it not appearing in the findings. I should make it invalid only in the case you mentioned of `NonExistentQueue` error.

I just saw that you did it like that, you have my approval then ✅

MrCloudSec (approved):
Thank you again @kagahd 🚀

PR #8334, opened 2025-07-21
fix(defender): avoid duplicated findings in check `defender_domain_dkim_enabled`

danibarranqueroo:
### Context

The finding UID generation in Prowler uses the resource_name field as part of the UID construction. When a check evaluates multiple resources (like multiple domains) but uses a generic resource_name, it can result in duplicate UIDs for different findings.
This issue was identified during a comprehensive review of all M365 checks that use generic `resource_name` values. While 37 checks initially appeared suspicious, only `defender_domain_dkim_enabled` actually generates multiple findings with the same UID due to its loop structure.
The fix ensures that each domain evaluation generates a unique finding UID, improving the accuracy of finding tracking and deduplication.

### Description

Update `resource_name` assignment in `defender_domain_dkim_enabled` replacing generic `DKIM Configuration` with the actual domain name.

### Checklist

#### API

### License

### DIFF
--- a/prowler/CHANGELOG.md
+++ b/prowler/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 All notable changes to the **Prowler SDK** are documented in this file.
 
+## [v5.9.2] (Prowler v5.9.2) UNRELEASED
+
+### Fixed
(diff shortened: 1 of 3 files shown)


HugoPBrito (approved):
Nice catch 🔍

PR #8364, opened 2025-07-25
fix(ui): validate email input field in sign-in and sign-up forms

sumit-tft:
### Context

This PR fixes missing email input validation in sign-in and sign-up forms.

### Description

- Applies input validation to enforce correct email formatting.
- Adjusts error message display logic to differentiate between SAML-based and standard authentication.

image

image

image

### Checklist

### License


alejandrobailo (changes requested):
Maybe it's out of scope, but I'm noticing different styles in the error messages. Take a look at your screenshots—font size and icons are missing for sign-up. It seems that our way of rendering messages isn't consistent. Please check, and if you can, provide some context. I appreciate it.

In the e2e test folder, you have a file with the login e2e tests. There is one test commented; uncomment it and make it work with the new validation message. To run the tests, you can use `npm run test:e2e` or `npm run test:e2e:ui` if you want to see the Playwright interface.

sumit-tft:
@alejandrobailo Thanks for the review 
Yes, as you mentioned, we're displaying a `custom error message` on the sign-in page. But for the sign-up page, the error message is coming from `Zod`, which is why you don't see the icon and consistent font styling. 

And, Here's the e2e test result for the commented test case.

Screenshot 2025-07-30 103821


image

sumit-tft:
@alejandrobailo, I’ve updated the error message styles for the `email` input. When you have a moment, please take a look and let me know if anything else needs to be updated.

image

image

image

ui/components/auth/oss/auth-form.tsx
```diff
@@ -232,22 +232,17 @@ export const AuthForm = ({
               <CustomInput
                 control={form.control}
                 name="email"
-                type="email"
```
paabloLC, line 235: @sumit-tft is there any specific reason why you're removing this type?

ui/components/auth/oss/auth-form.tsx
```diff
@@ -232,22 +232,17 @@ export const AuthForm = ({
...
                 <>
                   <CustomInput
                     control={form.control}
                     name="password"
                     password
-                    isInvalid={
-                      !!form.formState.errors.password ||
-                      !!form.formState.errors.email
-                    }
+                    isInvalid={!!form.formState.errors.password}
```
paabloLC, line 245: why? we wanted both, password and email 🤔 am I missing some context?

ui/components/auth/oss/auth-form.tsx
```diff
@@ -318,10 +313,12 @@ export const AuthForm = ({
                   )}
                 </>
               )}
-              {type === "sign-in" && form.formState.errors?.email && (
+              {type === "sign-in" && form.formState.errors?.credentials && (
```
paabloLC, line 316: same question here

paabloLC (changes requested):

ui/components/auth/oss/auth-form.tsx
sumit-tft: Yes @paabloLC when we use `type="email"`, the HTML5 validation prevents the form from being submitted before Zod validation runs. Because of that, error message ('invalid email') doesn’t appear. I removed the type so that Zod validation can handle it as expected.

ui/components/auth/oss/auth-form.tsx
sumit-tft: Earlier, the password field was also checking for the email error. I’ve updated it so that only the password error is considered for this field.

ui/components/auth/oss/auth-form.tsx
sumit-tft: previously, when a wrong email/password was submitted, the error was set directly on the email field. Now, I’ve added a separate `credentials` to display a error message (“Incorrect email or password”) instead of attaching it to the email field.

alejandrobailo:
Done in https://github.com/prowler-cloud/prowler/pull/8698

PR #8365, opened 2025-07-25
feat(integrations): implement AWS Security Hub integration

AdriiiPRodri:
### Context

This pull request introduces the integration of AWS Security Hub within the Prowler API. The integration allows users to send scan findings directly to AWS Security Hub, providing a seamless way to centralize and manage security findings in their AWS environment.

### Description

Important: This pull request must be merged only after #8056, as it depends on the changes introduced there. The current branch is based on that unmerged branch, so please ensure the correct merge order to avoid conflicts.

Summary of changes
- Added models, serializers, and endpoints to manage AWS Security Hub integrations.
- Implemented logic for sending findings to Security Hub using the configured integration and provider credentials.
- Added endpoints for retrieving active Security Hub regions for a given provider.
- Extended the API and worker to support Security Hub connection tests and integration tasks.
- Included comprehensive unit tests for the new integration logic.
- Updated API documentation and OpenAPI specs to reflect the new endpoints and integration type.
- Ensured that all integration tasks are executed asynchronously and logged appropriately.

### Checklist

#### API

### License


api/src/backend/api/specs/v1.yaml
```diff
@@ -8918,6 +8918,19 @@ components:
                     default: output
                 required:
                 - bucket_name
+              - type: object
+                title: AWS Security Hub
+                properties:
+                  send_only_fails:
+                    type: boolean
+                    default: false
+                    description: If true, only findings with status 'FAILED' will
```
jfagoagas:
```suggestion
                    description: If true, only findings with status 'FAIL' will
```

`FAILED` is how AWS Security Hub names failed findings.

jfagoagas: Please, change it in other places.

api/src/backend/api/migrations/0047_remove_integration_unique_configuration_per_tenant.py
```diff
@@ -0,0 +1,16 @@
...
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("api", "0046_lighthouse_gpt5"),
+    ]
+
+    operations = [
+        migrations.RemoveConstraint(
```
jfagoagas, line 12: Why are we removing this? We still need it for S3.

api/src/backend/tasks/jobs/integrations.py
```diff
@@ -154,3 +160,219 @@ def upload_s3_integration(
...
+
+            # Initialize prowler provider for finding transformation
+            prowler_provider = initialize_prowler_provider(provider)
+
+            # Get findings from database and transform them to ASFF format
+            logger.info(f"Transforming findings for scan {scan_id}")
+            asff_findings = []
+
+            # Process findings in batches to avoid memory issues
+            qs = Finding.all_objects.filter(scan_id=scan_id).order_by("uid").iterator()
```
jfagoagas: I think here you need to add `tenant_id` in the filter to take advantage of table indexes. As a rule of thumb, `tenant_id` needs to be passed in all queries as a a filter. RLS includes it implicitly but the planner optimizes if it's set in the filter.

api/src/backend/tasks/jobs/integrations.py
```diff
@@ -154,3 +160,219 @@ def upload_s3_integration(
...
+        integration.credentials if integration.credentials else provider_secret
+    )
+    connection = SecurityHub.test_connection(
+        aws_account_id=provider_uid,
+        raise_on_exception=False,
+        **credentials,
+    )
+
+    if connection.is_connected:
+        # If not saved, calculate them
```
jfagoagas: is this needed? don't know if it comes from a previous iteration.

api/src/backend/tasks/jobs/integrations.py
```diff
@@ -154,3 +160,219 @@ def upload_s3_integration(
...
+                logger.error(
+                    f"No Security Hub integrations found for provider {provider_id}"
+                )
+                return False
+
+            # Get the provider object
+            provider = Provider.objects.get(id=provider_id)
+
+            # Initialize prowler provider for finding transformation
+            prowler_provider = initialize_prowler_provider(provider)
```
jfagoagas, line 267: Should we test connection before?

api/src/backend/tasks/jobs/integrations.py
```diff
@@ -154,3 +160,219 @@ def upload_s3_integration(
...
+                # Transform findings using the same method as the report generation
+                transformed_findings = [
+                    FindingOutput.transform_api_finding(finding, prowler_provider)
+                    for finding in batch
+                ]
+
+                # Convert to ASFF format using the ASFF transformer
+                asff_transformer = ASFF(
+                    findings=transformed_findings,
+                    file_path="",  # Not needed for integration
```
jfagoagas: `file_path` defaults to `None`, just in case you don't want to pass it.

api/CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ All notable changes to the **Prowler API** are documented in this file.
 ### Added
 - Github provider support [(#8271)](https://github.com/prowler-cloud/prowler/pull/8271)
 - Integration with Amazon S3, enabling storage and retrieval of scan data via S3 buckets [(#8056)](https://github.com/prowler-cloud/prowler/pull/8056)
+- Integration with Amazon Security Hub, enabling sending findings to Security Hub [(#8365)](https://github.com/prowler-cloud/prowler/pull/8365)
```
jfagoagas: This needs to be in 1.12.0. Could you please also bump API's version in this PR?

jfagoagas (changes requested):
This is a great one @AdriiiPRodri 👏 I left some comments/suggestions, please check them out when you get a chance. Thanks!

api/src/backend/api/migrations/0047_remove_integration_unique_configuration_per_tenant.py
AdriiiPRodri: No, since S3 is the one that has this requirement, the S3 serializer now takes care of this check

api/src/backend/tasks/jobs/integrations.py
AdriiiPRodri: This is because before each execution, even though we have the regions stored, we check them again since it's a quick process and this way we can detect possible changes

api/src/backend/tasks/jobs/integrations.py
AdriiiPRodri: This is used to calculate the findings, regardless of the provider’s state, if the scan has findings we need the provider for the `transform_api_finding function`:

```
FindingOutput.transform_api_finding(finding, prowler_provider)
```

api/src/backend/tasks/jobs/integrations.py
AdriiiPRodri: I didn’t want to touch the Output class since it affects many components, but it's not correct. If I don't send "" and leave it as None, the following happens:

```
>>> path= None
>>> file_extension = '.json'
>>> f"{path}{file_extension}"
'None.json'
```

api/src/backend/tasks/jobs/integrations.py
jfagoagas: I know but if the provider is not connected, the call will raise an exception. I'm not sure what'd happen then regarding to the integration.

api/src/backend/tasks/jobs/integrations.py
jfagoagas: I meant only the comment.

api/CHANGELOG.md
AdriiiPRodri: Sure!

api/src/backend/tasks/jobs/integrations.py
jfagoagas: Ouch, thanks for clarifying it. Definitely the SDK needs a review on that.

api/src/backend/api/migrations/0047_remove_integration_unique_configuration_per_tenant.py
jfagoagas: Was this included in this PR? I wasn't aware of it.

api/src/backend/tasks/jobs/integrations.py
AdriiiPRodri: Even with this "problem" we can remove this line if you want, the value will be None.json but we don't use it

jfagoagas: It's fine, up to you to leave it or not.

api/src/backend/tasks/jobs/integrations.py
AdriiiPRodri: Good catch I think the same, thanks!

api/src/backend/api/specs/v1.yaml
AdriiiPRodri: Done

api/src/backend/api/migrations/0047_remove_integration_unique_configuration_per_tenant.py
AdriiiPRodri: It is:

```
class BaseWriteIntegrationSerializer(BaseWriteSerializer):
    def validate(self, attrs):
        if (
            attrs.get("integration_type") == Integration.IntegrationChoices.AMAZON_S3
            and Integration.objects.filter(
                configuration=attrs.get("configuration")
            ).exists()
        ):
            raise serializers.ValidationError(
                {"configuration": "This integration already exists."}
            )
```

api/src/backend/tasks/jobs/integrations.py
AdriiiPRodri: Aaa ok I can remove it

api/src/backend/api/migrations/0047_remove_integration_unique_configuration_per_tenant.py
jfagoagas: 🤦 sorry.

api/src/backend/tasks/jobs/integrations.py
AdriiiPRodri: We've never checked it, honestly I don't know what would happen if the provider disconnects right between the scan and this call. I need to investigate it

AdriiiPRodri: I see, if the initialization fails then this exception will be logged and the result of the task will be false:

```
logger.error(
            f"Security Hub integrations failed for provider {provider_id}: {str(e)}"
        )
```

We can discuss whether we want this message to be an error type or maybe info

jfagoagas (approved):

PR #8374, opened 2025-07-27
fix(ec2): add check that protocol is matched in security group checks

### LINKED ISSUE #8373: Incorrect Protocol Validation in EC2 Security Group Checks
## Problem Description

The `check_security_group` function in `prowler/providers/aws/services/ec2/lib/security_groups.py` has a logic issue where it doesn't properly validate protocol matching between ingress rules and the protocol being checked.

## Current Behavior

When checking security groups for specific protocols (e.g., TCP), the function would incorrectly return `True` for ingress rules with different protocols (e.g., UDP) if they have public CIDR ranges, even when the protocols don't match.

## Root Cause

The function was missing a protocol validation check. It would process ingress rules regardless of protocol mismatch, which could lead to incorrect results in security assessments.

### Example Scenario

```python
# Ingress rule with UDP protocol and public CIDR
ingress_rule = {
    "IpProtocol": "udp",
    "FromPort": 0,
    "ToPort": 65535,
    "IpRanges": [{"CidrIp": "0.0.0.0/0"}]
}

# Checking for TCP protocol
result = check_security_group(ingress_rule, "tcp", None, True)
# Current behavior: Returns True (incorrect)
# Expected behavior: Returns False (protocols don't match)
```

-

### Environment Resource

4. EKS

### OS used

Linux (centos)

### Prowler version

5.7.5

### Pip version

25.1.1

### Context

_No response_

aviadlevy:
### Context

Fix #8373 

### Description

Add a protocol validation check that ensures:
- When checking a specific protocol (not "-1" for all protocols)
- The ingress rule's protocol must match the protocol being checked
- If protocols don't match, return `False` immediately

### Checklist

#### API

### License

### DIFF
--- a/prowler/CHANGELOG.md
+++ b/prowler/CHANGELOG.md
@@ -16,6 +16,7 @@ All notable changes to the **Prowler SDK** are documented in this file.
 
 ### Fixed
 - False positives in SQS encryption check for ephemeral queues [(#8330)](https://github.com/prowler-cloud/prowler/pull/8330)
+- Add protocol validation check in security group checks to ensure proper protocol matching [(#8374)](https://github.com/prowler-cloud/prowler/pull/8374)
 
 ---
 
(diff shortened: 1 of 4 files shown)


MrCloudSec (approved):
Good catch @aviadlevy, we needed this logic, thank you.

PR #8375, opened 2025-07-28
feat(ui): show all provider UIDs in scan page filter regardless of co…

sumit-tft:
### Context

This PR updates the scan page's Provider UID filter to show all available providers, regardless of their connection status.

### Description

- Removes the condition that restricted the filter to only connected providers.
- Ensures the filter now lists all providers in the tenant, making all provider UIDs available in the dropdown.

image

image

### Checklist

### License


ui/app/(prowler)/scans/page.tsx
```diff
@@ -31,20 +31,22 @@ export default async function Scans({
...
-      alias: provider.attributes.alias,
-      providerType: provider.attributes.provider,
-      uid: provider.attributes.uid,
-      connected: provider.attributes.connection.connected,
-    })) || [];
+    providersData?.data
+      ?.filter(
+        (provider: ProviderProps) =>
+          provider.attributes.connection.connected === true,
+      )
```
alejandrobailo, line 42: We are removing the filter in the lines above and then adding the filter here?
Maybe I'm losing anything @sumit-tft

sumit-tft: @alejandrobailo previously we used `connectionStatus: true` to make the API call and show the response in both the provider UID filter and the 'Launch Scan' dropdown. Now, I’m fetching all provider details regardless of `connection Status`. In the 'Launch Scan' dropdown, I’m showing only those with `connectionStatus: true`, but in the filter, I’m displaying all provider UIDs. Hope that clarifies it, please let me know if I missed anything.

alejandrobailo (approved):

PR #8409, opened 2025-07-30
fix(s3): use enabled to filter

AdriiiPRodri:
### Context

This pull request addresses the need to handle the enablement and disablement of integrations, specifically for Amazon S3 integrations. The motivation behind this change is to ensure that only enabled integrations are considered during processing, which enhances the accuracy and reliability of integration management.

### Description

This update introduces a filter to include only enabled Amazon S3 integrations in the relevant query within ‎`api/src/backend/tasks/tasks.py`. By adding ‎`enabled=True` to the queryset, the system will now exclude any integrations that are not currently enabled. This change helps prevent unintended processing of disabled integrations and aligns with best practices for integration lifecycle management.

### Checklist

#### API

### License

### DIFF
--- a/api/src/backend/tasks/tasks.py
+++ b/api/src/backend/tasks/tasks.py
@@ -387,6 +387,7 @@ def generate_outputs_task(scan_id: str, provider_id: str, tenant_id: str):
         s3_integrations = Integration.objects.filter(
             integrationproviderrelationship__provider_id=provider_id,
             integration_type=Integration.IntegrationChoices.AMAZON_S3,
+            enabled=True,
         )
 
     if s3_integrations:
--- a/api/src/backend/tasks/tasks.py
+++ b/api/src/backend/tasks/tasks.py
@@ -487,7 +487,8 @@ def check_integrations_task(tenant_id: str, provider_id: str):
     try:
         with rls_transaction(tenant_id):
             integrations = Integration.objects.filter(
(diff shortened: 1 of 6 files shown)


api/src/backend/tasks/tasks.py
```diff
@@ -387,6 +387,7 @@ def get_writer(writer_map, name, factory, is_last):
         s3_integrations = Integration.objects.filter(
             integrationproviderrelationship__provider_id=provider_id,
             integration_type=Integration.IntegrationChoices.AMAZON_S3,
+            enabled=True,
```
jfagoagas, line 390: Please add a test to cover this.

jfagoagas (changes requested):
I think adding `enabled=True` is ok but not enough if only added in `generate_outputs_task`. I think both `check_integration_connection` and `check_integrations_task` have to handle if the integration is enabled to either test the connection or enqueue the integration task.

AdriiiPRodri:
> I think adding `enabled=True` is ok but not enough if only added in `generate_outputs_task`. I think both `check_integration_connection` and `check_integrations_task` have to handle if the integration is enabled to either test the connection or enqueue the integration task.

That sounds good to me. I had considered not limiting the connection check based on whether the integration is enabled or not, but it’s true that it doesn’t make much practical sense to check connections for a disabled integration. I’ll go ahead and introduce these changes

jfagoagas (approved):

PR #8417, opened 2025-07-31
fix: add default values for S3 class

AdriiiPRodri:
### Context

This pull request introduces default values for certain parameters in the S3 class constructor. The motivation behind this change is to enhance the reliability of the class by ensuring that key parameters have sensible defaults, thus reducing the risk of errors when these arguments are not explicitly provided. This update aims to improve usability and maintain consistency across the codebase.

### Description

The following changes were made in the ‎`prowler/providers/aws/lib/s3/s3.py` file:
- The ‎`session_duration` parameter now has a default value of ‎`3600` instead of ‎`None`.
- The ‎`role_session_name` parameter now defaults to ‎`ROLE_SESSION_NAME` instead of ‎`None`.

These modifications ensure that when the S3 class is instantiated without specifying these parameters, the class will use the provided defaults, improving robustness and reducing the need for additional checks elsewhere in the code.

### Checklist

#### API

### License


prowler/providers/aws/lib/s3/s3.py
```diff
@@ -76,9 +76,9 @@ def __init__(
         output_directory: str,
         session: AWSSession = None,
         role_arn: str = None,
-        session_duration: int = None,
+        session_duration: int = 3600,
         external_id: str = None,
-        role_session_name: str = None,
+        role_session_name: str = ROLE_SESSION_NAME,
         mfa: bool = None,
```
pedrooot, line 82:
```suggestion
        mfa: bool = False,
```

pedrooot (changes requested):

MrCloudSec (dismissed):

pedrooot (dismissed):

prowler/providers/aws/aws_provider.py
```diff
@@ -611,16 +613,16 @@ def setup_assumed_session(
                 secret_key=assumed_role_credentials.aws_secret_access_key,
                 token=assumed_role_credentials.aws_session_token,
                 expiry_time=assumed_role_credentials.expiration,
-                refresh_using=self.refresh_credentials,
+                refresh_using=AwsProvider.refresh_credentials,
```
jfagoagas: I think this is not correct. `refresh_credentials` is not an static method.

pedrooot (commented):
I'm a bit afraid of this PR, we have to ensure that everything is working as expected. A "small" change as this one could lead into a broken provider or integration. 

If you need any help with E2E tests just tell me and we can catch up 🤜🤛

prowler/providers/aws/lib/session/aws_set_up_session.py
```diff
@@ -192,10 +192,8 @@ def validate_arguments(
                 "If a role ARN is provided, a session duration, an external ID, and a role session name are required."
             )
     else:
-        if session_duration or external_id or role_session_name:
-            raise ValueError(
-                "If a session duration, an external ID, or a role session name is provided, a role ARN is required."
-            )
+        if external_id:
```
jfagoagas, line 197: Why this change?

MrCloudSec: Because now the role session name and the session duration will have always a value.

prowler/providers/aws/aws_provider.py
MrCloudSec: Please, @pedrooot revert this change.

jfagoagas:
btw I think there is no need to backport this PR because changes are only required for the S3 integration.

prowler/providers/aws/aws_provider.py
```diff
@@ -577,8 +581,9 @@ def setup_session(
             )
```
jfagoagas, line 582: Don't we need `@staticmethod` decorator here?

prowler/providers/aws/aws_provider.py
```diff
@@ -588,6 +593,7 @@ def setup_assumed_session(
         refreshing of the assumed role credentials.
 
         Args:
+            identity (AWSIdentityInfo): The identity information.
             assumed_role_credentials (AWSCredentials): The assumed role credentials.
```
jfagoagas, line 598: Update docstrings here pls.

prowler/providers/aws/aws_provider.py
```diff
@@ -607,20 +613,22 @@ def setup_assumed_session(
...
-                token=assumed_role_credentials.aws_session_token,
-                expiry_time=assumed_role_credentials.expiration,
-                refresh_using=self.refresh_credentials,
+                access_key=assumed_role_configuration.credentials.aws_access_key_id,
+                secret_key=assumed_role_configuration.credentials.aws_secret_access_key,
+                token=assumed_role_configuration.credentials.aws_session_token,
+                expiry_time=assumed_role_configuration.credentials.expiration,
+                refresh_using=lambda: AwsProvider.refresh_credentials(
+                    assumed_role_configuration, session
+                ),
```
jfagoagas, line 624: Why can't we do this?
```suggestion
                refresh_using=lambda: AwsProvider.refresh_credentials,
```

AdriiiPRodri: I think we can't, @pedrooot @MrCloudSec what do you think?

pedrooot: Nope, we can't since the refresh credentials method is now static and we need to pass the `assumed_role_configuration` and `session`

jfagoagas: Great catch, I totally skipped it.

jfagoagas (approved):

pedrooot (approved):

MrCloudSec (approved):

PR #8422, opened 2025-07-31
fix(github): separate --github-app-key and --github-app-key-path CLI

andoniaf:
### Context
Fixes GitHub App authentication inconsistency where `--github-app-key` and `--github-app-key-path` were incorrectly aliased to the same parameter, causing confusion between file paths and key content.  

### Description
- Separate `--github-app-key-path` (file path) and `--github-app-key` (content) parameters                                                                                                                                                                                                              

- Update GithubProvider constructor to accept distinct parameters                                                                                                                                                                                                                                                         

- Add smart environment variable detection for `GITHUB_APP_KEY`                                                                                                                                                                                                                                                            

- Update common provider initialization to pass separated parameters                                                                                                                                                                                                                                                      

- Fix test expectations for new argument count     
                         
### Checklist

#### API

### License


MrCloudSec (changes requested):
Please, leave the flag `--github-app-key` as it is, if not, you are introducing a breaking change. We are not planning on supporting passing the key content in the flag, it is only for the API as in K8s.

andoniaf:
> Please, leave the flag `--github-app-key` as it is, if not, you are introducing a breaking change. We are not planning on supporting passing the key content in the flag, it is only for the API as in K8s.

We should remove or rename the flags then. In k8s we use:
```
Authentication and Configuration:
  --kubeconfig-file [FILE_PATH]
                        Path to the kubeconfig file to use for CLI requests. Not necessary for in-cluster execution.
```

So it's clear that we're expecting a file path.

prowler/providers/github/github_provider.py
```diff
@@ -156,7 +159,9 @@ def __init__(
             self._auth_method = "Personal Access Token"
         elif oauth_app_token:
             self._auth_method = "OAuth App Token"
-        elif github_app_id and (github_app_key or github_app_key_content):
+        elif github_app_id and (
+            github_app_key or github_app_key_path or github_app_key_content
+        ):
             self._auth_method = "GitHub App Token"
```
HugoPBrito:
```suggestion
            self._auth_method = "GitHub App Key and ID"
```

prowler/providers/github/github_provider.py
HugoPBrito:
```suggestion
        elif environ.get("GITHUB_APP_ID", "") and (environ.get("GITHUB_APP_KEY_CONTENT", "") or environ.get("GITHUB_APP_KEY_PATH", "")):
```

prowler/providers/github/github_provider.py
```diff
@@ -303,13 +334,27 @@ def setup_session(
...
-                        app_key = format_rsa_key(environ.get("GITHUB_APP_KEY", ""))
 
-                        if app_id and app_key:
-                            pass
+                        app_key_path = environ.get("GITHUB_APP_KEY_PATH", "")
+                        if app_key_path:
+                            with open(app_key_path, "r") as rsa_key:
+                                app_key = rsa_key.read()
+                        else:
+                            env_key = environ.get("GITHUB_APP_KEY", "")
```
HugoPBrito, line 348:
```suggestion
                            env_key = environ.get("GITHUB_APP_KEY_CONTENT", "")
```

prowler/providers/github/github_provider.py
```diff
@@ -303,13 +334,27 @@ def setup_session(
...
+                            if env_key:
+                                if env_key.startswith("-----BEGIN"):
+                                    app_key = format_rsa_key(env_key)
+                                elif os.path.isfile(env_key):
+                                    with open(env_key, "r") as rsa_key:
+                                        app_key = rsa_key.read()
```
HugoPBrito: This does not make sense anymore.

prowler/providers/github/github_provider.py
```diff
@@ -303,13 +334,27 @@ def setup_session(
...
+                                else:
+                                    raise GithubEnvironmentVariableError(
+                                        file=os.path.basename(__file__),
+                                        message="GITHUB_APP_KEY must contain either RSA key content (starting with -----BEGIN) or a valid file path.",
```
HugoPBrito:
```suggestion
                                        message="GITHUB_APP_KEY_CONTENT must contain either RSA key content (starting with -----BEGIN).",
```

prowler/providers/github/lib/arguments/arguments.py
```diff
@@ -31,11 +31,17 @@ def init_parser(self):
     )
     github_auth_subparser.add_argument(
         "--github-app-key",
+        nargs="?",
+        help="GitHub App Key content (PEM format) to log in against GitHub",
+        default=None,
+        metavar="GITHUB_APP_KEY_CONTENT",
```
HugoPBrito: Either Change this to `GITHUB_APP_KEY` or resolve all other comments regarding the  `GITHUB_APP_KEY` -> `GITHUB_APP_KEY_CONTENT`.

HugoPBrito (changes requested):

prowler/providers/github/github_provider.py
danibarranqueroo: We are not adding GITHUB_APP_KEY_CONTENT as envvar, it was just the metavar of the flag github_app_key, I'll modify there that text.

prowler/providers/github/github_provider.py
danibarranqueroo: Good catch!

prowler/providers/github/lib/arguments/arguments.py
danibarranqueroo: Changed to GITHUB_APP_KEY it wasn't breaking anything because it's just a metavar but good catch so we don't have inconsistence 🔝

HugoPBrito (dismissed):

PR #8428, opened 2025-08-01
Potential fix for code scanning alert no. 55: Server-side request forgery

MrCloudSec:
Potential fix for [https://github.com/prowler-cloud/prowler/security/code-scanning/55](https://github.com/prowler-cloud/prowler/security/code-scanning/55)

To fix the SSRF vulnerability, we should ensure that `scanId` and `complianceId` are validated before being used in the URL. The best approach is to restrict these parameters to a safe pattern, such as allowing only alphanumeric characters, dashes, and underscores (which are typical for IDs). This can be done using regular expressions. If the values do not match the expected pattern, the function should throw an error or return an error response, preventing the request from being made with potentially malicious input.

The changes should be made in the `getComplianceCsv` function in `ui/actions/scans/scans.ts`, specifically before constructing the URL. No new methods are needed, but we will add validation logic for both `scanId` and `complianceId`.

---

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._

### DIFF
--- a/ui/actions/scans/scans.ts
+++ b/ui/actions/scans/scans.ts
@@ -278,6 +278,13 @@ export const getComplianceCsv = async (
   scanId: string,
   complianceId: string,
 ) => {
+  // Only allow alphanumeric, dash, and underscore for IDs
+  const idPattern = /^[a-zA-Z0-9_-]+$/;
+  if (!idPattern.test(scanId) || !idPattern.test(complianceId)) {
+    return {
+      error: "Invalid scan or compliance ID format.",
+    };
+  }
   const headers = await getAuthHeaders({ contentType: false });
 
   const url = new URL(


paabloLC:
Hi @MrCloudSec 

Could you please provide the steps to reproduce the issue? It would help us confirm the vulnerability and verify the fix

Also, please check the GitHub Actions logs, two tests didn’t pass.

Thanks!

paabloLC:
After reviewing our current code, the regex validation proposed in PR `#8428` is NOT necessary. Our code already handles parameters correctly:

  1. `getScan` (line 89): scanId is used directly in the URL but wrapped in a try-catch that handles errors
  2. `getComplianceCsv` (line 277): Both scanId and complianceId are also handled with try-catch blocks

  Our current implementation is cleaner and more functional because:
  - IDs are validated by the backend when the request is made
  - Errors are properly caught and returned appropriately
  - You don't need artificial frontend restrictions if the backend already validates

  The regex validation from the PR could potentially break functionality if your real IDs use characters not included in the `/^[a-zA-Z0-9_-]+$/` pattern.

Closing the PR as the current error handling approach is more robust and follows proper separation of concerns between frontend and backend validation.

PR #8358, opened 2025-07-24
fix(lighthouse): Display errors in Lighthouse & allow resending message

0xbadshah:
### Context

When Lighthouse errors out due to any reason (insufficient access, OpenAI ratelimit, etc), the chat window breaks without any error message. This fix will allow user to see what went wrong.

### Description

This PR adds two things:
1. Highlight error message if anything goes wrong (on langgraph or on Vercel AI SDK stream)
2. If a message errors out, then allow user to retry/edit and retry.

### Checklist

- Are there new checks included in this PR? Yes / No
    - If so, do we need to update permissions for the provider? Please review this carefully.
- [ ] Review if the code is being covered by tests.
- [ ] Review if code is being documented following this specification https://github.com/google/styleguide/blob/gh-pages/pyguide.md#38-comments-and-docstrings
- [ ] Review if backport is needed.
- [ ] Review if is needed to change the [Readme.md](https://github.com/prowler-cloud/prowler/blob/master/README.md)
- [ ] Ensure new entries are added to [CHANGELOG.md](https://github.com/prowler-cloud/prowler/blob/master/prowler/CHANGELOG.md), if applicable.

#### API
- [ ] Verify if API specs need to be regenerated.
- [ ] Check if version updates are required (e.g., specs, Poetry, etc.).
- [ ] Ensure new entries are added to [CHANGELOG.md](https://github.com/prowler-cloud/prowler/blob/master/api/CHANGELOG.md), if applicable.

### Screenshots

Error when OpenAI ratelimits

image

Error when the configured key doesnt have access to the configured model

image

### How to Test

1. Create an OpenAI API key
2. Configure the OpenAI API key to Lighthouse AI
3. In OpenAI dashboard, goto "Settings" -> "Limits"
4. Emulate error 1: Under "Model usage" and select other models like "dall-e-2" under "Allowed models". Ensure gpt-4o and gpt-4o-mini are not selected. Save the configuration. After few minutes try interacting with Lighthouse chat. It will display the error and allow you to edit last message.
5. Emulate error 2: Under "Model usage" -> Ratelimits, edit ratelimit for gpt-4o, gpt-4o-mini and default models as just 1 Token per Minute and 1 Request per minute and save the config. After few minutes try interacting with Lighthouse chat. It will display the error and allow you to edit last message.
6. In both the cases, when there's an error, the last message will be moved to the chatbox for retrying.

### License

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #8431, opened 2025-08-01
fix: clear only last message on error

0xbadshah:
### Context

Fixes feature released in https://github.com/prowler-cloud/prowler/pull/8358. When there's an error and user gets the last sent message in the input form. If user deletes the message, then second last user message (that had no errors from OpenAI) got loaded into the input form. The intended feature is to allow user to delete last message and nothing else.

### Description

The previous PR used useEffects with messages, which had unintended side effects. Once the error occurs, the form input was updated with last message to edit. In case the user deleted the last message content, this modification triggered useEffects and it loaded the second last message - which keeps deleting user messages until all messages are gone.

So, this PR fixes it by using useEffects only when an error message is raised (either by onError or error from Lighthouse with prefix `[LIGHTHOUSE_ANALYST_ERROR]:`

### Checklist

#### API

### License

### DIFF
--- a/ui/components/lighthouse/chat.tsx
+++ b/ui/components/lighthouse/chat.tsx
@@ -41,7 +41,9 @@ export const Chat = ({ hasConfig, isActive }: ChatProps) => {
     experimental_throttle: 100,
     sendExtraMessageFields: true,
     onFinish: (message) => {
-      // Detect error messages sent from backend using specific prefix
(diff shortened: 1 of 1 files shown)


0xbadshah:
@paabloLC @alejandrobailo Please let me know if I need to add a Changelog (this PR fixes issue created in a feature that will be added to latest release).

paabloLC (approved):

PR #8444, opened 2025-08-05
feat(ui): add Mutelist menu item under Configuration

paabloLC:
### Summary

  Adds a "Mutelist" menu item under Configuration that opens the Cloud Providers page with the Mutelist configuration modal automatically opened.

### Changes

 - Move Mutelist-specific logic from Menu.tsx to menu-list.ts configuration
  - Update menu-list to accept store values and callbacks as parameters
  - Make Menu component generic and reusable without item-specific logic
  - Add support for disabled and onClick properties in menu configuration

### Checklist

#### API

### License


ui/CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ All notable changes to the **Prowler UI** are documented in this file.
 - Amazon AWS S3 integration [(#8391)](https://github.com/prowler-cloud/prowler/pull/8391)
 - Github provider support [(#8405)](https://github.com/prowler-cloud/prowler/pull/8405)
 - XML validation for SAML metadata in the UI [(#8429)](https://github.com/prowler-cloud/prowler/pull/8429)
+- Added Mutelist item under Configuration menu for quick access to mutelist configuration [(#8444)](https://github.com/prowler-cloud/prowler/pull/8444)
```
alejandrobailo:
```suggestion
- New menu item under Configuration section for quick access to the Mutelist [(#8444)](https://github.com/prowler-cloud/prowler/pull/8444)
```

ui/components/providers/muted-findings-config-button.tsx
```diff
@@ -1,49 +1,71 @@
...
 
 import { MutedFindingsConfigForm } from "./forms";
 
 interface MutedFindingsConfigButtonProps {
   isDisabled?: boolean;
+  autoOpen?: boolean;
+  hideButton?: boolean;
 }
 
 export const MutedFindingsConfigButton = ({
```
alejandrobailo: You have to emulate the button state; if the button is disabled because there are no providers, the menu button shouldn't appear.
https://github.com/user-attachments/assets/f7c90d5b-10d3-4186-bc16-ebc0334e8745

ui/components/providers/muted-findings-config-button.tsx
```diff
@@ -1,49 +1,71 @@
...
 }: MutedFindingsConfigButtonProps) => {
-  const [isOpen, setIsOpen] = useState(false);
+  const [isOpen, setIsOpen] = useState(autoOpen);
+  const router = useRouter();
+
+  useEffect(() => {
+    if (autoOpen) {
+      setIsOpen(true);
+    }
+  }, [autoOpen]);
```
alejandrobailo: This effect is not needed as `isOpen` is being instantiated with the `autoOpen` value by default; you can remove it to check that it is working as expected.

alejandrobailo (changes requested):
Need some changes, not working as expected:
`Devs 1 - 0 Claude Code`

ui/components/providers/muted-findings-config-button.tsx
paabloLC: that's correct

ui/components/providers/muted-findings-config-button.tsx
paabloLC: done

sumit-tft:
@paabloLC I just pulled the latest changes and checked the functionality.
When I click the `Mutelist` menu, the modal opens as expected. After successfully configuring the mutelist, the modal closes automatically, but the URL still shows http://localhost:3000/providers?mutelist=true.

image

alejandrobailo:
I'm reviewing it again, and I don't like the approach as changing the URL with the Mutelist is causing a re-render of the providers page, meaning pulling the data from the backend again, while this data is still the same.

This is a good example of state management. As far as I know, we have Zustand installed, and maybe using it can help us here. 

We are using URL params everywhere, and this approach is very tough to maintain and scale. I think here we are mixing a component state (Mutelist modal), which is an internal with an external state in the URL.

What do you think?

ui/components/ui/sidebar/menu.tsx
```diff
@@ -180,7 +183,12 @@ export const Menu = ({ isOpen }: { isOpen: boolean }) => {
                       <CollapseMenuButton
                         icon={Icon}
                         label={label}
-                        submenus={submenus}
+                        submenus={submenus.map((submenu) => ({
+                          ...submenu,
+                          disabled:
+                            submenu.label === "Mutelist" &&
+                            hasProviders === false,
+                        }))}
```
alejandrobailo: This is a very customized change within a common component like `Menu.tsx`, it should be agnostic of the item, shouldn't it?

ui/hooks/use-has-providers.ts
```diff
@@ -0,0 +1,34 @@
+"use client";
+
+import { useEffect, useState } from "react";
+
+import { getProviders } from "@/actions/providers";
+
+export const useHasProviders = () => {
```
alejandrobailo: Is this hook needed? Where are we re-using it?

alejandrobailo: Also, a few other important things:
- We are wrapping a server action into a useEffect, making a mismatch between SSR and CSR.
- This calls the server after the render, which is the opposite of what we should do. 
- This is also causing a wrong loading state behavior.

alejandrobailo (changes requested):

ui/components/providers/store-initializer.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+interface StoreInitializerProps {
+  values: {
+    hasProviders?: boolean;
+    // Add more properties here as needed
+    // otherProperty?: string;
+  };
+}
+
+export function StoreInitializer({ values }: StoreInitializerProps) {
```
alejandrobailo: What about moving this file to the store's location? `store/ui`

ui/components/ui/sidebar/collapse-menu-button.tsx
```diff
@@ -94,90 +89,188 @@ export const CollapseMenuButton = ({
...
+                  placement="right"
+                >
+                  <div className="w-full">
+                    <Button
+                      variant={isActive ? "secondary" : "ghost"}
+                      className={cn(
+                        "ml-4 h-8 w-full justify-start",
+                        "cursor-not-allowed opacity-50",
+                      )}
+                      disabled={true}
```
alejandrobailo, line 114: We can extract these lines as a component; we have them too in lines 139-148.

ui/components/providers/store-initializer.tsx
alejandrobailo: I've moved it

ui/components/ui/sidebar/collapse-menu-button.tsx
paabloLC: go for it =D

alejandrobailo: Better to refactor in another task

alejandrobailo (dismissed):

alejandrobailo (approved):

sumit-tft (approved):

PR #8449, opened 2025-08-05
fix(export): remove static timestamp

AdriiiPRodri:
### Context

This pull request addresses the need to ensure that output file paths for exports reflect the actual scan start time, rather than relying on a static timestamp. The previous implementation used a constant value for the timestamp in generated file paths, which could cause confusion and reduce traceability.

### Description

This change removes the use of the static ‎`output_file_timestamp` for naming exported output files and directories. Instead, it retrieves the ‎`started_at` timestamp from the corresponding ‎`Scan` object and formats it appropriately. This ensures that the output directory and file names accurately represent when the scan was initiated.

The implementation updates the ‎`_generate_output_directory` function to fetch and use the real scan start time. The associated tests have also been updated and extended to mock the scan object and verify that the generated paths include the correct timestamp. Additional test cases ensure proper handling of invalid characters in provider names and confirm that the directory structure is created as expected.

### Checklist

#### API

### License

### DIFF
--- a/api/src/backend/tasks/jobs/export.py
+++ b/api/src/backend/tasks/jobs/export.py
@@ -8,11 +8,12 @@ from botocore.exceptions import ClientError, NoCredentialsError, ParamValidation
 from celery.utils.log import get_task_logger
 from django.conf import settings
 
+from api.db_utils import rls_transaction
+from api.models import Scan
 from prowler.config.config import (
     csv_file_suffix,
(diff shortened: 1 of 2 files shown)


alejandrobailo (approved):

cesararroba (approved):

jfagoagas (approved):
Great catch!

PR #8520, opened 2025-08-13
feat(filters): improve provider connection filter UX

paabloLC:
### Description

 - Replace "true/false" with "Connected/Disconnected" labels in connection filter
  - Add CONNECTION_STATUS_MAPPING for reusable connection status labels
  - Simplify isConnectionStatus helper function for better type checking
  - Maintain backward compatibility with boolean URL parameters (connection=true/false)

### Checklist

#### API

### License

### DIFF
--- a/ui/CHANGELOG.md
+++ b/ui/CHANGELOG.md
@@ -12,6 +12,7 @@ All notable changes to the **Prowler UI** are documented in this file.
 ### 🔄 Changed
 
 - Disable `See Compliance` button until scan completes [(#8487)](https://github.com/prowler-cloud/prowler/pull/8487)
+- Provider connection filter now shows "Connected/Disconnected" instead of "true/false" for better UX [(#8520)](https://github.com/prowler-cloud/prowler/pull/8520)
 
 ### 🐞 Fixed
 - DataTable column headers set to single-line [(#8480)](https://github.com/prowler-cloud/prowler/pull/8480)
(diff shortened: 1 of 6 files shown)


sumit-tft:
@paabloLC functionality is working fine, just a couple of suggestions:

1. In ui/lib/helper-filters.ts, for `CONNECTION_STATUS_MAPPING`, I think `as` is not necessary. Is there any particular reason you have used it?
2. In ui/components/ui/custom/custom-dropdown-filter.tsx, instead of `isConnectionStatus`, we could use a more generic name since this check isn’t specific to connection status. In the future, we may reuse it for other dropdowns with value/label pairs. so we could consider names like: `isLabelValuePair` or `isLabeledValue`. If you have other name suggestions, please use them.

paabloLC:
> @paabloLC functionality is working fine, just a couple of suggestions:
> 
> 1. In ui/lib/helper-filters.ts, for `CONNECTION_STATUS_MAPPING`, I think `as` is not necessary. Is there any particular reason you have used it?
> 2. In ui/components/ui/custom/custom-dropdown-filter.tsx, instead of `isConnectionStatus`, we could use a more generic name since this check isn’t specific to connection status. In the future, we may reuse it for other dropdowns with value/label pairs. so we could consider names like: `isLabelValuePair` or `isLabeledValue`. If you have other name suggestions, please use them.

1 -   The `as ProviderConnectionStatus` is necessary because:
  - It forces the specific type within the FilterEntity union type
  - It allows the isConnectionStatus type guard to work correctly
  - It maintains type safety when accessing label and value properties

2 - The `isConnectionStatus` is fine because we also have `isScanEntity` which could also be made more generic, but that would be part of a future refactoring task (if needed), not this current one. Right now, `isConnectionStatus` only handles the connection status of a provider, so the name accurately reflects its current purpose and scope

PR #8535, opened 2025-08-18
fix(organization): list all accessible organizations

HugoPBrito:
### Context

Currently we're facing the following error:
`ERROR: GitHub API error while listing organizations: 403 {"message": "Resource not accessible by integration", "documentation_url": "https://docs.github.com/rest/orgs/orgs#list-organizations-for-the-authenticated-user", "status": "403"}`

### Description

The error has been fixed making the correct calls for each context.

### Checklist

#### API

### License

### DIFF
--- a/prowler/CHANGELOG.md
+++ b/prowler/CHANGELOG.md
@@ -15,7 +15,6 @@ All notable changes to the **Prowler SDK** are documented in this file.
 - Refine kisa isms-p compliance mapping [(#8479)](https://github.com/prowler-cloud/prowler/pull/8479)
 
 ### Fixed
-- Call error in app context for Github organization [(#8535)](https://github.com/prowler-cloud/prowler/pull/8535)
 
 ---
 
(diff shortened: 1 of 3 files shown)


MrCloudSec (changes requested):
I think it is better to catch the error and log it in a warning level.

HugoPBrito:
> I think it is better to catch the error and log it in a warning level.

I don’t agree.

The user call should only be used in a user context; otherwise, we should use the correct method. This isn’t an API issue but a design issue on our side. If no repositories or organizations are provided, we should fetch all available ones from the resources we have, not just from user data when PAT is the selected auth method.

With the current approach, even if we log the error, we still don’t support to fetch the repositories accessible through GitHub Apps when `--organization` / `--repository` are not provided, which means that it also won't work in Prowler Cloud.

MrCloudSec (dismissed):

MrCloudSec (approved):

PR #8544, opened 2025-08-20
fix(m365): only evaluate enabled users in `entra_users_mfa_capable`

HugoPBrito:
### Context

Currently we are evaluating disabled accounts in `entra_users_mfa_capable` check.

### Description

This behavior has been fixed to only evaluate enabled users.

### Checklist

#### API

### License

### DIFF
--- a/prowler/providers/m365/services/entra/entra_service.py
+++ b/prowler/providers/m365/services/entra/entra_service.py
@@ -14,7 +14,10 @@ from prowler.providers.m365.m365_provider import M365Provider
 class Entra(M365Service):
     def __init__(self, provider: M365Provider):
         super().__init__(provider)
+
         if self.powershell:
+            self.powershell.connect_exchange_online()
+            self.user_accounts_status = self.powershell.get_user_account_status()
             self.powershell.close()
 
         loop = get_event_loop()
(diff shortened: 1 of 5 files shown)


prowler/providers/m365/services/entra/entra_service.py
```diff
@@ -585,6 +592,7 @@ class User(BaseModel):
     on_premises_sync_enabled: bool
     directory_roles_ids: List[str] = []
     is_mfa_capable: bool = False
+    account_enabled: bool = True
```
MrCloudSec, line 595:
```suggestion
    account_enabled: bool = True
```
Can we set this to False by default?

MrCloudSec (changes requested):
What about the `entra_user_with_vm_access_has_mfa` check?

prowler/providers/m365/services/entra/entra_service.py
HugoPBrito: I think it’s better to set it to true since most accounts are enabled, and in the worst-case scenario I’d rather have a false positive than a false negative (in this case, a ghost finding).

HugoPBrito:
> entra_user_with_vm_access_has_mfa

That’s an Azure check. I can add the logic to Azure, but I’d need to set up PowerShell for that. We can discuss it, but it should go in a separate PR since it involves many changes and possibly new requirements for Azure provider.

MrCloudSec (approved):

PR #8572, opened 2025-08-26
fix(s3): resource metadata error in `s3_bucket_shadow_resource_vulnerability`

danibarranqueroo:
### Context

The check was failing with error: "Resource metadata could not be converted to dict" due to creating dynamic objects using type() function. This caused worker process failures and prevented the check from completing execution. The fix ensures proper resource serialization while maintaining the intended functionality of detecting shadow resource vulnerabilities.

### Description

Fixed resource metadata conversion error in s3_bucket_shadow_resource_vulnerability check by replacing dynamic object creation with standard dictionary resource.

**Changes Made**

- Replaced type() dynamic object creation with resource={} to fix serialization issues
- Maintained FAIL status for cross-account shadow resources as intended
- Improved test to ensure the check is working as expected

### Steps to review

- Review the changes made to the check and tests code
- Test the check with prowler and with pytest

### Checklist

#### API

### License

### DIFF
--- a/prowler/providers/aws/services/s3/s3_bucket_shadow_resource_vulnerability/s3_bucket_shadow_resource_vulnerability.metadata.json
+++ b/prowler/providers/aws/services/s3/s3_bucket_shadow_resource_vulnerability/s3_bucket_shadow_resource_vulnerability.metadata.json
@@ -3,7 +3,7 @@
   "CheckID": "s3_bucket_shadow_resource_vulnerability",
   "CheckTitle": "Check for S3 buckets vulnerable to Shadow Resource Hijacking (Bucket Monopoly)",
   "CheckType": [
-    ""
+    "Effects/Data Exposure"
   ],
   "ServiceName": "s3",
   "SubServiceName": "",
--- a/prowler/providers/aws/services/s3/s3_bucket_shadow_resource_vulnerability/s3_bucket_shadow_resource_vulnerability.py
+++ b/prowler/providers/aws/services/s3/s3_bucket_shadow_resource_vulnerability/s3_bucket_shadow_resource_vulnerability.py
@@ -70,22 +70,12 @@ class s3_bucket_shadow_resource_vulnerability(Check):
                 )
                 # Check if this bucket exists in another account
                 if s3_client._head_bucket(bucket_name):
-                    # Create a virtual bucket object for reporting
-                    virtual_bucket = type(
-                        "obj",
-                        (object,),
-                        {
-                            "name": bucket_name,
-                            "region": region,
-                            "arn": f"arn:{s3_client.audited_partition}:s3:::{bucket_name}",
-                            "tags": [],
-                        },
-                    )()
-
-                    report = Check_Report_AWS(self.metadata(), resource=virtual_bucket)
+                    report = Check_Report_AWS(self.metadata(), resource={})
                     report.region = region
                     report.resource_id = bucket_name
-                    report.resource_arn = virtual_bucket.arn
+                    report.resource_arn = (
+                        f"arn:{s3_client.audited_partition}:s3:::{bucket_name}"
+                    )
                     report.resource_tags = []
                     report.status = "FAIL"
                     report.status_extended = f"S3 bucket {bucket_name} for service {service} is a known shadow resource that exists and is owned by another account."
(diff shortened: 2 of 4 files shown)


prowler/providers/aws/services/s3/s3_bucket_shadow_resource_vulnerability/s3_bucket_shadow_resource_vulnerability.metadata.json
```diff
@@ -3,7 +3,7 @@
   "CheckID": "s3_bucket_shadow_resource_vulnerability",
   "CheckTitle": "Check for S3 buckets vulnerable to Shadow Resource Hijacking (Bucket Monopoly)",
   "CheckType": [
-    ""
+    "Effects/Data Exposure"
```
jfagoagas, line 6: Is this a supported type?

prowler/providers/aws/services/s3/s3_bucket_shadow_resource_vulnerability/s3_bucket_shadow_resource_vulnerability.py
```diff
@@ -70,22 +70,12 @@ def execute(self):
...
-                        {
-                            "name": bucket_name,
-                            "region": region,
-                            "arn": f"arn:{s3_client.audited_partition}:s3:::{bucket_name}",
-                            "tags": [],
-                        },
-                    )()
-
-                    report = Check_Report_AWS(self.metadata(), resource=virtual_bucket)
+                    report = Check_Report_AWS(self.metadata(), resource={})
```
jfagoagas, line 73: Why you can't add a dict with the previous values?
```
{
                            "name": bucket_name,
                            "region": region,
                            "arn": f"arn:{s3_client.audited_partition}:s3:::{bucket_name}",
                            "tags": [],
                        },
```

Tags must be added too.

jfagoagas (changes requested):

prowler/providers/aws/services/s3/s3_bucket_shadow_resource_vulnerability/s3_bucket_shadow_resource_vulnerability.metadata.json
danibarranqueroo: Yes, we obtain the CheckType for AWS checks from here:

https://docs.aws.amazon.com/securityhub/latest/userguide/asff-required-attributes.html#Types

I picked the one I thought that should be used for this check, but feel free to suggest any other!

prowler/providers/aws/services/s3/s3_bucket_shadow_resource_vulnerability/s3_bucket_shadow_resource_vulnerability.py
danibarranqueroo: All of that is being added to the report below and I used that approach because we don't have the resource object with all the info, it's just mocked info so I felt that it won't add any extra info. To summarize, doing that like a dict or adding those values to the report info will do the same. 

However, if you feel that your approach is better I'll make the changes needed 💯

prowler/providers/aws/services/s3/s3_bucket_shadow_resource_vulnerability/s3_bucket_shadow_resource_vulnerability.metadata.json
jfagoagas: You're totally right. Great one 👏

jfagoagas: Another idea for a `@validator` getting the allowed values from AWS Security Hub documentation.

prowler/providers/aws/services/s3/s3_bucket_shadow_resource_vulnerability/s3_bucket_shadow_resource_vulnerability.py
jfagoagas: That's true, thanks for clarifying it. The only doubt I have is that I can't remember if we are still adding this info in other checks to have something in `.resource` @puchy22 @MrCloudSec.

Regarding `tags` can we add it for the check?

danibarranqueroo: Let's wait them for that doubt, regarding tags we have in line 79 `report.resource_tags = []`, are you expecting something different?

jfagoagas: Sorry for the noise, the code we're reviewing is for non-existent buckets not for the ones we already scanned. I'm resolving this comment.

jfagoagas (approved):
👏 great catch!

PR #2662, opened 2023-08-02
feat(gcp): Improve gcp performance

MrCloudSec:
### Description

Increase GCP performance x8 and check if services are actives.

### License

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #8575, opened 2025-08-26
feat(gcp): Add support for skipping APIs check

tom-intruder:
### Context

In a previous PR (#2662), a check was added to verify if a GCP API is disabled for the project.

GCP only requires the API to be enabled for the project connected to the service account, so in the case where Prowler has been connected to an organisation via an isolated project, the check for an active API in another project will fail when the service account can review these resources. For most services, the API needs to be enabled to create resources; however, there are a few different GCP services, including CloudSQL and API keys, that can be created without enabling the APIs required by Prowler, so detection of these resources fails.

The use case for this flag would be to enable the required APIs once for the service account project and then be able to review every resource the service account is granted access to.

### Description

- Added the `--scan-disabled-apis` flag

### Steps to review
We will create a basic object in a project (an API key) to test that Prowler has the correct access to scan that object.
- Using a GCP organisation, create 2 projects (A and B)
- Create a service account in project A
- Create an API key in project B.
- Enable the API keys service in A.
- Grant the service account access to read API keys across the organisation.
- Run prowler against project B with and without the flag.
- Notice that when the flag is used, Prowler detects the API key. Without the flag, the code skips this check as the API is not enabled on project B

### Checklist

#### API

### License


MrCloudSec (changes requested):
Thanks for the PR @tom-intruder. Would not be better to call the flag `--skip-api-check` or similar?

tom-intruder:
> Thanks for the PR @tom-intruder. Would not be better to call the flag `--skip-api-check` or similar?

Yeah I'm happy to change the flag to that

MrCloudSec (dismissed):
Love this, thanks again @tom-intruder

tests/providers/gcp/gcp_provider_test.py
```diff
@@ -987,3 +987,176 @@ def test_config_import(self):
...
+            )
+
+            from prowler.providers.gcp.lib.service.service import GCPService
+
+            GCPService("testservice", gcp_provider)
+
+            assert gcp_provider.skip_api_check is True
+            mocked_is_api_active.assert_not_called()
+
+    def test_api_check_argugement(self):
```
MrCloudSec: I think this test is duplicated @tom-intruder

tom-intruder: Hi
I originally had the test_no_api_check_argument, but codecov fired an error saying it did not cover the case when the argument was supplied.
Is that an oversight on my end, and could it be implemented in a different test?

tests/providers/gcp/gcp_provider_test.py
```diff
@@ -987,3 +987,176 @@ def test_config_import(self):
...
+                client_secret=None,
+                refresh_token=None,
+                skip_api_check=arguments.skip_api_check,
+            )
+
+            from prowler.providers.gcp.lib.service.service import GCPService
+
+            GCPService("testservice", gcp_provider)
+            assert gcp_provider.skip_api_check is False
+            mocked_is_api_active.assert_called()
```
MrCloudSec:
```suggestion
```

tests/providers/gcp/gcp_provider_test.py
```diff
@@ -987,3 +987,91 @@ def test_config_import(self):
...
+                skip_api_check=arguments.skip_api_check,
+            )
+
+            from prowler.providers.gcp.lib.service.service import GCPService
+
+            GCPService("testservice", gcp_provider)
+
+            assert gcp_provider.skip_api_check is True
+            mocked_is_api_active.assert_not_called()
+
```
MrCloudSec:
```suggestion
```

MrCloudSec (approved):

PR #8587, opened 2025-08-27
fix(github): provider always scans user instead of organization when using provider UID

andoniaf:
### Description
When using Prowler App (UI + API) to scan GitHub providers, creating two providers from the same GitHub account (one for scanning the user, one for scanning an organization) always results in scanning the user, even when the organization name is provided as the UID.

Solution: Add GitHub provider handling in `get_prowler_provider_kwargs`

### Steps to review

1. In Prowler App, create a GitHub provider with a user name as the UID
3. Create another GitHub provider with an organization name as the UID
5. Run scans for both providers
7. Observe that both scan the user's repositories, not the organization's

### Checklist

#### API

### License


HugoPBrito (dismissed):
All good from DR and SDK side.

api/src/backend/api/utils.py
```diff
@@ -120,6 +120,12 @@ def get_prowler_provider_kwargs(
         }
     elif provider.provider == Provider.ProviderChoices.KUBERNETES.value:
         prowler_provider_kwargs = {**prowler_provider_kwargs, "context": provider.uid}
+    elif provider.provider == Provider.ProviderChoices.GITHUB.value:
+        if provider.uid:
+            prowler_provider_kwargs = {
+                **prowler_provider_kwargs,
+                "organizations": [provider.uid],
+            }
```
vicferpoy, line 128: Could you add a test case for this function, please? Check `test_get_prowler_provider_kwargs`

vicferpoy (changes requested):

api/src/backend/api/utils.py
andoniaf: Done, thanks for the heads up!

vicferpoy (approved):
🚀

MrCloudSec (approved):

HugoPBrito (approved):

PR #8590, opened 2025-08-27
feat(models): add `AdditionalUrls` field to check metadata

HugoPBrito:
### Context

Currently, we lack metadata fields to add useful links for the end user, such as explanations, guidance, utilities, etc.

### Description

This enhancement introduces an `AdditionalUrls` field into the metadata structure for checks in Prowler. This allows enriching the metadata with extra documentation, remediation guidance, or reference URLs that go beyond the existing documentation fields.

This feature supports broader automation, traceability, and user education by allowing checks to reference multiple URLs relevant to compliance or cloud provider docs.

### Checklist

#### API

### License


prowler/lib/check/models.py
```diff
@@ -113,6 +115,7 @@ class CheckMetadata(BaseModel):
     Risk: str
     RelatedUrl: str
     Remediation: Remediation
+    AdditionalUrls: list[str] = []
```
jfagoagas: Using `default_factory=list` (e.g., in a dataclass or Pydantic model) is better than using `AdditionalUrls: list[str] = []` because it **avoids a common Python pitfall with mutable default arguments**.

**Explanation:**

- When you write `AdditionalUrls: list[str] = []`, the empty list `[]` is created once when the class is defined. All instances of the class that don’t specify `AdditionalUrls` will share the **same list object**. This means if one instance modifies the list, the change will affect all other instances that didn't set their own value.
- In contrast, `default_factory=list` creates a **new list** for each instance. So each instance gets its own separate list, and changes to one instance’s list won’t affect others.

**Example:**

```python
from dataclasses import dataclass, field

@dataclass
class Example:
    items: list = field(default_factory=list)
```
- Here, every `Example()` instance has its own independent `items` list.

**Why is this important?**
- **Safety:** Prevents bugs caused by shared mutable state.
- **Expected behavior:** Each object should have its own data, not accidentally share it with others.

**Summary:**  
Always use `default_factory=list` for mutable defaults (like lists or dicts) in Python classes. It ensures that each instance gets a fresh, independent object.

prowler/CHANGELOG.md
```diff
@@ -16,10 +16,11 @@ All notable changes to the **Prowler SDK** are documented in this file.
 - ECS privilege escalation patterns (StartTask and RunTask) for AWS provider [(#8541)](https://github.com/prowler-cloud/prowler/pull/8541)
 - Resource Explorer enumeration v2 API actions in `cloudtrail_threat_detection_enumeration` check [(#8557)](https://github.com/prowler-cloud/prowler/pull/8557)
 - GCP `--skip-api-check` command line flag [(#8575)](https://github.com/prowler-cloud/prowler/pull/8575)
+- `AdditionalUrls` field in CheckMetadata Pydantic model [(#8584)](https://github.com/prowler-cloud/prowler/pull/8584)
 
 ### Changed
 - Refine kisa isms-p compliance mapping [(#8479)](https://github.com/prowler-cloud/prowler/pull/8479)
-- Improve AWS Security Hub region check using multiple threads [(#8365)](https://github.com/prowler-cloud/prowler/pull/8365)
+- Improve AWS Security Hub region check using multiple threads [(#8590)](https://github.com/prowler-cloud/prowler/pull/8590)
```
MrCloudSec: Be careful with this change.

prowler/CHANGELOG.md
```diff
@@ -16,10 +16,11 @@ All notable changes to the **Prowler SDK** are documented in this file.
 - ECS privilege escalation patterns (StartTask and RunTask) for AWS provider [(#8541)](https://github.com/prowler-cloud/prowler/pull/8541)
 - Resource Explorer enumeration v2 API actions in `cloudtrail_threat_detection_enumeration` check [(#8557)](https://github.com/prowler-cloud/prowler/pull/8557)
 - GCP `--skip-api-check` command line flag [(#8575)](https://github.com/prowler-cloud/prowler/pull/8575)
+- `AdditionalUrls` field in CheckMetadata Pydantic model [(#8584)](https://github.com/prowler-cloud/prowler/pull/8584)
```
MrCloudSec: The PR number is wrong here.

prowler/CHANGELOG.md
HugoPBrito: Thanks.
Fixed, please review.

prowler/CHANGELOG.md
HugoPBrito: Thanks.
Fixed, please review.

prowler/lib/check/models.py
HugoPBrito: Thanks for that insights. I've followed this comment also for other attributes that were normal lists.

HugoPBrito:
Since this PR is going to be merged after 5.11 release, changelog needs to be updated.

prowler/CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 All notable changes to the **Prowler SDK** are documented in this file.
 
+## [v5.12.0] (Prowler UNRELEASED)
+
+### Added
+- `AdditionalUrls` field in CheckMetadata Pydantic model [(#8590)](https://github.com/prowler-cloud/prowler/pull/8590)
```
jfagoagas:
```suggestion
- `AdditionalUrls` field in CheckMetadata [(#8590)](https://github.com/prowler-cloud/prowler/pull/8590)
```

prowler/lib/check/models.py
```diff
@@ -113,13 +115,14 @@ class CheckMetadata(BaseModel):
     Risk: str
     RelatedUrl: str
     Remediation: Remediation
+    AdditionalUrls: list[str] = Field(default_factory=list)
```
jfagoagas, line 118: I assume this is going to be optional until all checks have it, if not this is going to break execution.

jfagoagas (changes requested):

prowler/lib/check/models.py
HugoPBrito: Nice point.
I also made `RelatedUrl` optional so it doesn't need to be added in new checks.

HugoPBrito: In fact, since` default_factory=list ` sets it as an empty list `[]` it won't break the code when not provided. It's better to not make it optional, and set `RelatedUrl` to a default string for the same reason.

prowler/lib/check/models.py
```diff
@@ -113,9 +113,9 @@ class CheckMetadata(BaseModel):
     ResourceType: str
     Description: str
     Risk: str
-    RelatedUrl: Optional[str] = None
+    RelatedUrl: str = Field(default_factory=str)
```
jfagoagas: Do you need this? As far as I know a `str` in Python is immutable so you should not need the default factory. Maybe it's a good practice to keep.

prowler/lib/check/models.py
jfagoagas: So smart!

prowler/lib/check/models.py
AdriiiPRodri: What Pepe is pointing out is correct, in Pydantic, when you declare a field as str, you don't need default_factory=str because:
- The default value for str when using `default_factory=str` is just "" (empty string).
- `str` in Python is immutable.

We can drop `default_factory=str` unless we want to follow a consistent style where every field uses `default_factory`. But functionally, here it’s not needed

jfagoagas (approved):

PR #8592, opened 2025-08-27
docs(api): fix API response samples

vicferpoy:
### Description

Endpoint categories have been rearranged.

image

Endpoints which return a `task` will now return a proper example.

image

image

### Steps to review

### Checklist

#### API

### License

### DIFF
--- a/api/src/backend/api/specs/v1.yaml
+++ b/api/src/backend/api/specs/v1.yaml
@@ -147,7 +147,22 @@ paths:
           content:
             application/vnd.api+json:
               schema:
-                $ref: '#/components/schemas/PaginatedTaskList'
+                $ref: '#/components/schemas/TaskResponse'
+              examples:
+                Task queued:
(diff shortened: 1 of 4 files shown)


api/src/backend/api/specs/v1.yaml
```diff
@@ -417,6 +470,7 @@ paths:
         name: filter[delta]
         schema:
           type: string
+          x-spec-enum-id: fef6b0e2506af8bc
```
AdriiiPRodri, line 473: This is the only thing that concerns me. This enum ID probably changes every time we build the API container, so we will see differences in the views each time we download the spec. We should take this into account and avoid uploading this kind of change

vicferpoy: Nice catch. It's the first time I notice it. I will keep it mind, thanks!

AdriiiPRodri (approved):

PR #8622, opened 2025-09-01
feat(integrations): Support JIRA integration in the API

vicferpoy:
### Description

image

image

image

image

> [!NOTE]
> `project_key` and `domain` within the integration configuration are used to validate duplicates in the same tenant.

image

### Checklist

#### API

### License

### DIFF
--- a/api/src/backend/api/v1/serializers.py
+++ b/api/src/backend/api/v1/serializers.py
@@ -45,6 +45,8 @@ from api.v1.serializer_utils.integrations import (
     AWSCredentialSerializer,
     IntegrationConfigField,
     IntegrationCredentialField,
+    JiraConfigSerializer,
+    JiraCredentialSerializer,
     S3ConfigSerializer,
     SecurityHubConfigSerializer,
 )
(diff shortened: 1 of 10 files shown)


api/src/backend/api/v1/serializers.py
```diff
@@ -1962,8 +1966,20 @@ def validate(self, attrs):
...
+            and Integration.objects.filter(
+                configuration__contains={
+                    "domain": attrs.get("configuration").get("domain"),
+                    "project_key": attrs.get("configuration").get("project_key"),
+                }
+            ).exists()
+        ):
+            raise serializers.ValidationError(
+                {"configuration": "This integration already exists."}
+            )
```
jfagoagas:
```suggestion
            ```suggestion
            raise ConflictException(
            detail="This integration already exists.",
            pointer="/data/attributes/configuration",
        )
```
```

Comment (jfagoagas) [reply api/src/backend/api/v1/serializers.py]: Can you update it too for S3?

**Review by jfagoagas — COMMENTED** [review]
Well done 👏

File: api/src/backend/api/v1/serializers.py (hunk 1 above)
Comment (AdriiiPRodri) [reply api/src/backend/api/v1/serializers.py]: I believe the pointer should be "/data/attributes/configuration" which is the attribute that causes this exception

Comment (vicferpoy) [reply api/src/backend/api/v1/serializers.py]: For sure @AdriiiPRodri . I added these:

```python
raise ConflictException(
                detail="This integration already exists.",
                pointer="/data/attributes/configuration",
            )
```

And for SH:

```python
raise ConflictException(
                        detail=f"Provider {provider.id} already has a Security Hub integration. Only one "
                               "Security Hub integration is allowed per provider.",
                        pointer="/data/relationships/providers",
                    )
```

jfagoagas: Great catch! I've updated it in the suggestion.

jfagoagas (approved):

AdriiiPRodri (approved):

api/src/backend/api/v1/serializers.py
AdriiiPRodri: Great!! 🚀

PR #8647, opened 2025-09-04
feat(login): add throttling option for the `/api/v1/tokens` endpoint

josema-xyz:
### Context

This prevents brute force attacks when there is no configuration in the reverse proxy for preventing them.

### Description

A throttling configuration has been added as documented in [DRF](https://www.django-rest-framework.org/api-guide/throttling/). Just use the `DJANGO_TOKEN_OBTAIN_THROTTLE` environment variable from the [api/.env.example](api/.env.example) file.

### Steps to review

This can be tested with any tool that can send `POST` requests. For example:
1. Configure `DJANGO_TOKEN_OBTAIN_THROTTLE` to 5 requests per minute, `5/minute`.
2. Call five times to `/api/v1/tokens` as usual to check is working.
3. Call once more just after that, it should not work and return a `429 Too many requests` with the next body:
```json
{
  "errors": [
    {
      "detail": "Request was throttled. Expected available in 57 seconds.",
      "status": "429",
      "source": {
        "pointer": "/data"
      },
      "code": "throttled"
    }
  ]
}
```
4. Wait a minute.
5. Call again to the endpoint and check it's working.

### License

### DIFF
--- a/api/.env.example
+++ b/api/.env.example
@@ -19,6 +19,8 @@ DJANGO_REFRESH_TOKEN_LIFETIME=1440
 DJANGO_CACHE_MAX_AGE=3600
 DJANGO_STALE_WHILE_REVALIDATE=60
 DJANGO_SECRETS_ENCRYPTION_KEY=""
+# Throttle, two options: Empty means no throttle; or if desired use one in DRF format: https://www.django-rest-framework.org/api-guide/throttling/#setting-the-throttling-policy
+DJANGO_TOKEN_OBTAIN_THROTTLE=
 # Decide whether to allow Django manage database table partitions
 DJANGO_MANAGE_DB_PARTITIONS=[True|False]
 DJANGO_CELERY_DEADLOCK_ATTEMPTS=5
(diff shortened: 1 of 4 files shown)


api/.env.example
```diff
@@ -19,6 +19,8 @@ DJANGO_REFRESH_TOKEN_LIFETIME=1440
 DJANGO_CACHE_MAX_AGE=3600
 DJANGO_STALE_WHILE_REVALIDATE=60
 DJANGO_SECRETS_ENCRYPTION_KEY=""
+# Throttle, two options: Empty means no throttle; or if desired use one in DRF format: https://www.django-rest-framework.org/api-guide/throttling/#setting-the-throttling-policy
+DJANGO_TOKEN_OBTAIN_THROTTLE=
```
jfagoagas: I think we need to define something by default. Let's start with `50/m`.

jfagoagas (changes requested):
That's clean 👏 I left just one comment to add a default value.

api/.env.example
josema-xyz: Totally right! Done! 🙌

AdriiiPRodri (dismissed):
Love it! 🚀

jfagoagas (approved):

PR #8655, opened 2025-09-05
feat(config): add generation for JWT keys if missing

josema-xyz:
### Context

This creates the JWT keys, only if they are missing.

### Description

The keys are generated and stored in disk, in `/home/prowler/.keys/prowler-api` inside the container, and `_data/api/keys` outside of it, when executed with `docker-compose-dev.yml`. Then they are set as environment variables and as Django settings.

### Steps to review

Start the application without the `DJANGO_TOKEN_SIGNING_KEY` and `DJANGO_TOKEN_VERIFYING_KEY` environment variables. They should then be created and reused if the application restart. Combinations of missing keys can be tested.

### License


api/src/backend/api/apps.py
```diff
@@ -1,4 +1,27 @@
...
+logger = logging.getLogger(BackendLogger.API)
+
+SIGNING_KEY_ENV = "DJANGO_TOKEN_SIGNING_KEY"
+VERIFYING_KEY_ENV = "DJANGO_TOKEN_VERIFYING_KEY"
+
+SIGNING_KEY_FILE = "jwt_signing_key.pem"
+VERIFYING_KEY_FILE = "jwt_verifying_key.pem"
+
+KEYS_DIRECTORY = (
+    Path.home() / ".keys" / "prowler-api"
```
jfagoagas: I know we talked about it but now I'm wondering if this is redundant as the container is only for the API.

docker-compose-dev.yml
```diff
@@ -14,9 +14,11 @@ services:
     ports:
       - "${DJANGO_PORT:-8080}:${DJANGO_PORT:-8080}"
     volumes:
-      - "./api/src/backend:/home/prowler/backend"
-      - "./api/pyproject.toml:/home/prowler/pyproject.toml"
-      - "outputs:/tmp/prowler_api_output"
+      - ./api/src/backend:/home/prowler/backend
+      - ./api/pyproject.toml:/home/prowler/pyproject.toml
+      - ./api/docker-entrypoint.sh:/home/prowler/docker-entrypoint.sh
```
jfagoagas, line 19: Do you need this?

jfagoagas (changes requested):
That's great 👏  can you add tests for the new code?

jfagoagas:
I've just realised that we need to clear `DJANGO_TOKEN_SIGNING_KEY` and `DJANGO_TOKEN_VERIFYING_KEY` from `./.env` to make this the default behaviour in the self-hosted deployment type.

docker-compose-dev.yml
jfagoagas, line 1: I think this changes needs to be in the production docker-compose too, for the self-hosted installation.

api/src/backend/api/apps.py
```diff
@@ -9,4 +32,123 @@ def ready(self):
...
+        file_path = KEYS_DIRECTORY / file_name
+        return file_path.read_text().strip() if file_path.is_file() else None
+
+    def _write_key_file(self, file_name, content):
+        """
+        Utility method to write content to a file.
+        """
+        file_path = KEYS_DIRECTORY / file_name
+        file_path.parent.mkdir(parents=True, exist_ok=True)
+        file_path.write_text(content)
```
jfagoagas: I'm wondering what'd happen if something fails during writes, could we handle it?

api/src/backend/api/apps.py
```diff
@@ -9,4 +32,123 @@ def ready(self):
...
+        Generate and set RSA key pairs for JWT token operations.
+        """
+        try:
+            from cryptography.hazmat.primitives import serialization
+            from cryptography.hazmat.primitives.asymmetric import rsa
+
+            # Generate RSA key pair
+            private_key = rsa.generate_private_key(
+                public_exponent=65537,
+                key_size=2048,
```
jfagoagas, line 140: In the future we could add another environment variable to increase this value.

jfagoagas (commented):
Two more things I realised:

1. Ensure that files have `0600` permissions on the container.
2. Can we add some notes about the deployment? Something like:
```
If you don’t set DJANGO_TOKEN_SIGNING_KEY/VERIFYING_KEY, the API will generate them at <home>/.keys/prowler-api/… with 0600 perms; back up these files to persist identity across redeploys.”
``

File: api/src/backend/api/apps.py [hunk 5]
```diff
@@ -1,4 +1,27 @@
...
+from config.custom_logging import BackendLogger
+from config.env import env
+
+logger = logging.getLogger(BackendLogger.API)
+
+SIGNING_KEY_ENV = "DJANGO_TOKEN_SIGNING_KEY"
+VERIFYING_KEY_ENV = "DJANGO_TOKEN_VERIFYING_KEY"
+
+SIGNING_KEY_FILE = "jwt_signing_key.pem"
+VERIFYING_KEY_FILE = "jwt_verifying_key.pem"
```
jfagoagas:
```suggestion
PRIVATE_KEY_FILE = "jwt_private.pem"
PUBLIC_KEY_FILE = "jwt_public.pem"
```

Just a suggestion.

jfagoagas (commented):
Now I'm testing it locally and I see the following:

During migrations the code is executed twice:
```
970ceedc0211_prowler-api-dev-1  | Applying database migrations...
970ceedc0211_prowler-api-dev-1  | 2025-09-08 13:47:10 [api] INFO: (apps) [module=apps path=/home/prowler/backend/api/apps.py line=63 function=_ensure_crypto_keys process=7 thread=140737470179136 transaction-id=None] Generating JWT RSA key pair. In production, set 'DJANGO_TOKEN_SIGNING_KEY' and 'DJANGO_TOKEN_VERIFYING_KEY' environment variables.
970ceedc0211_prowler-api-dev-1  | 2025-09-08 13:47:10 [api] INFO: (apps) [module=apps path=/home/prowler/backend/api/apps.py line=101 function=_ensure_jwt_keys process=7 thread=140737470179136 transaction-id=None] JWT keys generated and stored successfully
970ceedc0211_prowler-api-dev-1  | 2025-09-08 13:47:22 [api] INFO: (apps) [module=apps path=/home/prowler/backend/api/apps.py line=63 function=_ensure_crypto_keys process=14 thread=140737470179136 transaction-id=None] Generating JWT RSA key pair. In production, set 'DJANGO_TOKEN_SIGNING_KEY' and 'DJANGO_TOKEN_VERIFYING_KEY' environment variables.
970ceedc0211_prowler-api-dev-1  | 2025-09-08 13:47:22 [api] INFO: (apps) [module=apps path=/home/prowler/backend/api/apps.py line=104 function=_ensure_jwt_keys process=14 thread=140737470179136 transaction-id=None] JWT keys already generated
970ceedc0211_prowler-api-dev-1  | Operations to perform:
970ceedc0211_prowler-api-dev-1  |   Apply all migrations: account, admin, api, auth, authtoken, contenttypes, django_celery_beat, django_celery_results, sessions, sites, socialaccount, token_blacklist
970ceedc0211_prowler-api-dev-1  | Running migrations:
```

Then, for each fixture the code is executed too:
```
Loading api/fixtures/dev/4_dev_resources.json
970ceedc0211_prowler-api-dev-1  | 2025-09-08 13:48:06 [api] INFO: (apps) [module=apps path=/home/prowler/backend/api/apps.py line=63 function=_ensure_crypto_keys process=49 thread=140737470179136 transaction-id=None] Generating JWT RSA key pair. In production, set 'DJANGO_TOKEN_SIGNING_KEY' and 'DJANGO_TOKEN_VERIFYING_KEY' environment variables.
970ceedc0211_prowler-api-dev-1  | 2025-09-08 13:48:06 [api] INFO: (apps) [module=apps path=/home/prowler/backend/api/apps.py line=104 function=_ensure_jwt_keys process=49 thread=140737470179136 transaction-id=None] JWT keys already generated
```

The same happens after managing partitions and one last time during the API startup. 

Could you check if this process can be improved somehow?

api/src/backend/api/apps.py
josema-xyz: But what about the people that starts de API _natively_?

docker-compose-dev.yml
josema-xyz: Just removing the double quotes like in the rest of containers. Also adding the `keys` directory. And a extra, the `docker-entrypoint.sh` as if we want to modify and use it, we need to rebuild the image, and it's useful to just comment the `apply_migrations` and `apply_fixtures` steps. What do you think?

docker-compose-dev.yml
josema-xyz: The `docker-compose.yml` uses the same `.env` file. Do you also mean to add the `keys` volume?

api/src/backend/api/apps.py
josema-xyz: Let's add some handled exceptions.

api/src/backend/api/apps.py
josema-xyz: A perfect suggestion!

api/src/backend/api/apps.py
josema-xyz: Let's left a comment for that.

josema-xyz:
Deployment note added to the `Modify environment variables` section. I think that is the best location.

About the multiple runs problem, it's happening because they are different Python processes. The only way to skip that it's to use external mechanisms, like `lock` and `done` files. But the thing is, how to differentiate what processes need to skip that and not to load the keys? Migration and fixtures of course don't need it, but what about the API process, how can we detect it? I think the only easy and robust solution we have is the current one: generate the files only once and load it once per process. Are you thinking on a different approach?

api/src/backend/api/apps.py
jfagoagas: Do you mean from sources?

docker-compose-dev.yml
jfagoagas: Totally true, I didn't realise that it's for the `dev` one.

docker-compose-dev.yml
jfagoagas: Correct, we need it to make it work, right?

api/src/backend/api/apps.py
josema-xyz: Yes, without using Docker.

.env
```diff
@@ -84,45 +84,6 @@ DJANGO_REFRESH_TOKEN_LIFETIME=1440
 DJANGO_CACHE_MAX_AGE=3600
 DJANGO_STALE_WHILE_REVALIDATE=60
 DJANGO_MANAGE_DB_PARTITIONS=True
-# openssl genrsa -out private.pem 2048
```
jfagoagas, line 87: What if we leave them empty? So users can include the values and we don't miss the comment. Just realised about it.

docker-compose-dev.yml
josema-xyz: Yes, we need it to make it easy to persist them _manually_ by the user. I'll add this change.

.env
josema-xyz: Totally right!

api/src/backend/api/apps.py
jfagoagas: Make sense, thanks for taking care of this!

jfagoagas:
> Deployment note added to the `Modify environment variables` section. I think that is the best location.
> 
> About the multiple runs problem, it's happening because they are different Python processes. The only way to skip that it's to use external mechanisms, like `lock` and `done` files. But the thing is, how to differentiate what processes need to skip that and not to load the keys? Migration and fixtures of course don't need it, but what about the API process, how can we detect it? I think the only easy and robust solution we have is the current one: generate the files only once and load it once per process. Are you thinking on a different approach?

I don't know, I think is tough. Let's see if @AdriiiPRodri and @vicferpoy has some other idea, we have time until this gets security validated.

josema-xyz:
Tests added! 💪

api/CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ All notable changes to the **Prowler API** are documented in this file.
 - Integration with JIRA, enabling sending findings to a JIRA project [(#8622)](https://github.com/prowler-cloud/prowler/pull/8622), [(#8637)](https://github.com/prowler-cloud/prowler/pull/8637)
 - `GET /overviews/findings_severity` now supports `filter[status]` and `filter[status__in]` to aggregate by specific statuses (`FAIL`, `PASS`)[(#8186)](https://github.com/prowler-cloud/prowler/pull/8186)
 - Throttling options for `/api/v1/tokens` using the `DJANGO_THROTTLE_TOKEN_OBTAIN` environment variable [(#8647)](https://github.com/prowler-cloud/prowler/pull/8647)
+- Default JWT keys are generated and stored if they are missing from configuration [(#8655)](https://github.com/prowler-cloud/prowler/pull/8655)
```
vicferpoy: We should move this to a new minor version.

vicferpoy (changes requested):
Small detail about CHANGELOG versioning.

api/CHANGELOG.md
josema-xyz: 🚀

api/src/backend/api/apps.py
```diff
@@ -9,4 +33,137 @@ def ready(self):
         from api import signals  # noqa: F401
         from api.compliance import load_prowler_compliance
 
+        # Generate required cryptographic keys if not present, but only if:
+        #   `"manage.py" not in sys.argv`: If an external server (e.g., Gunicorn) is running the app
+        #   `os.environ.get("RUN_MAIN")`: If the Django development server is running the app
```
vicferpoy:
```suggestion
        #   `os.environ.get("RUN_MAIN")`: If it's not a Django command or using `runserver`, only the main process will do it
```

vicferpoy (changes requested):

vicferpoy (approved):
Awesome job 🚀

api/src/backend/api/apps.py
josema-xyz: 🚀

jfagoagas (approved):

PR #8658, opened 2025-09-08
feat: add APIKey model and migration

mmuller88:
- Add APIKey model with multi-tenancy and RBAC support
- Add APIKeyUser class for API key authentication
- Add APIKeyManager for secure key management
- Add migration 0048_add_api_key_model.py
- Add comprehensive tests for APIKey model

This establishes the foundation for API key functionality.

### Context

If fixes an issue please add it with `Fix #XXXX`

### Description

### Steps to review

### Checklist

#### API

### License


api/src/backend/api/models.py
```diff
@@ -165,6 +172,281 @@ class JSONAPIMeta:
...
+            queryset = queryset.filter(tenant_id=tenant_id)
+        return queryset
+
+    def get_from_key(self, key: str, tenant_id=None):
+        """
+        Get API key from the raw key string.
+        Uses the prefix for efficient lookup while respecting tenant boundaries.
+        """
+        try:
+            prefix, _, _ = key.partition(".")
```
AdriiiPRodri: Here the prefix is "pk_"+<8>, but in the extract_prefix function the prefix is only the <8>, without "pk_", we should fix this inconsistency

api/src/backend/api/models.py
```diff
@@ -165,6 +172,281 @@ class JSONAPIMeta:
...
+        )
+        secret = secrets.token_urlsafe(32)
+        key = f"{prefix}.{secret}"
+
+        # Use the parent manager's assign_key method but override the prefix
+        super().assign_key(obj)  # This sets obj.prefix and obj.hashed_key
+
+        # Override with our custom prefix and key
+        obj.prefix = prefix
+        obj.hashed_key = make_password(key, hasher=Sha512ApiKeyHasher())
```
AdriiiPRodri: Use of Sha512ApiKeyHasher from rest_framework_api_key.crypto, so it's a SHA512 but we expect a pbkdf2_|argon2|bcrypt in the tests

So I'm not sure if verify_key and verify_key are verified with the library hasher or with the django default

api/src/backend/api/models.py
```diff
@@ -165,6 +172,281 @@ class JSONAPIMeta:
...
+
+    def get_usable_keys(self, tenant_id=None):
+        """
+        Get all usable (non-revoked, non-expired) API keys.
+        If tenant_id is provided, filter by tenant.
+        """
+        queryset = self.filter(revoked=False)
+        if tenant_id:
+            queryset = queryset.filter(tenant_id=tenant_id)
+        return queryset
```
AdriiiPRodri: The docstring says non-revoked and non-expired but we are filtering only with revoked, maybe we need to add expired=False (idk if when a key expires, the status change to revoked=True)

api/src/backend/api/models.py
```diff
@@ -165,6 +172,281 @@ class JSONAPIMeta:
...
+        Get API key from the raw key string.
+        Uses the prefix for efficient lookup while respecting tenant boundaries.
+        """
+        try:
+            prefix, _, _ = key.partition(".")
+            if not prefix:
+                raise self.model.DoesNotExist("Invalid key format")
+
+            # Use all keys, not just usable ones, so we can give proper error messages for revoked/expired keys
+            queryset = self.all()
```
AdriiiPRodri, line 248: I believe this doesn't work, we have RLS so if the tenant is not fixed for the session RLS will block the select, maybe we need to use `self.using(MainRouter.admin_db)`. I don't know if we really want this

api/src/backend/api/migrations/0048_add_api_key_model.py
```diff
@@ -0,0 +1,120 @@
+# Generated manually for API Key model with multi-tenancy and RBAC support
```
AdriiiPRodri: Why we can't create this automatically with makemigrations?

api/src/backend/api/migrations/0048_add_api_key_model.py
```diff
@@ -0,0 +1,120 @@
...
+                    models.UUIDField(
+                        default=uuid.uuid4,
+                        editable=False,
+                        unique=True,
+                        help_text="UUID for external API references",
+                    ),
+                ),
+                (
+                    "prefix",
+                    models.CharField(editable=False, max_length=15, unique=True),
```
AdriiiPRodri: max_length should be 15 or 8? I think we can adjust it to our specs

api/src/backend/api/models.py
```diff
@@ -165,6 +172,281 @@ class JSONAPIMeta:
...
+        obj.save()
+
+        return obj, key
+
+    def using_admin_db(self):
+        """
+        Return a manager that uses the admin database to bypass RLS.
+        Useful for authentication where we don't have tenant context yet.
+        """
+        return self.using(MainRouter.admin_db)
```
AdriiiPRodri: We usually don’t need this method, and if we can avoid it, that would be better

AdriiiPRodri (changes requested):

api/src/backend/api/migrations/0048_add_api_key_model.py
```diff
@@ -0,0 +1,120 @@
...
+                    models.CharField(
+                        editable=False,
+                        max_length=150,
+                        primary_key=True,
+                        serialize=False,
+                        unique=True,
+                    ),
+                ),
+                (
+                    "uuid",
```
jfagoagas: Why `id` and `uuid`?

api/src/backend/api/migrations/0048_add_api_key_model.py
jfagoagas: Why not to use `TextField` and the define a validator? Not to create limits in the database, sometimes they are hard to fix.

api/src/backend/api/migrations/0048_add_api_key_model.py
```diff
@@ -0,0 +1,120 @@
...
+                        unique=True,
+                        help_text="UUID for external API references",
+                    ),
+                ),
+                (
+                    "prefix",
+                    models.CharField(editable=False, max_length=15, unique=True),
+                ),
+                ("hashed_key", models.CharField(editable=False, max_length=150)),
+                ("created", models.DateTimeField(auto_now_add=True, db_index=True)),
```
jfagoagas: This should follow the API spec, so `created_at`.

api/src/backend/api/migrations/0048_add_api_key_model.py
```diff
@@ -0,0 +1,120 @@
...
+                (
+                    "name",
+                    models.CharField(
+                        default=None,
+                        help_text="A free-form name for the API key. Need not be unique. 50 characters max.",
+                        max_length=50,
+                    ),
+                ),
+                (
+                    "revoked",
```
jfagoagas: This should follow the API spec, so `revoked_at`.

api/src/backend/api/migrations/0048_add_api_key_model.py
```diff
@@ -0,0 +1,120 @@
...
+                    models.BooleanField(
+                        blank=True,
+                        default=False,
+                        help_text="If the API key is revoked, clients cannot use it anymore. (This cannot be undone.)",
+                    ),
+                ),
+                (
+                    "expiry_date",
```
jfagoagas: Why not `expires_at`?

api/src/backend/api/migrations/0048_add_api_key_model.py
```diff
@@ -0,0 +1,120 @@
...
+                (
+                    "last_used_at",
+                    models.DateTimeField(
+                        blank=True,
+                        help_text="Last time this API key was used",
+                        null=True,
+                    ),
+                ),
+                (
+                    "tenant",
```
jfagoagas: Is `tenant_id`.

api/src/backend/api/models.py
```diff
@@ -165,6 +172,281 @@ class JSONAPIMeta:
         resource_name = "users"
 
 
+class APIKeyUser:
```
jfagoagas, line 174: Why `APIKeyUser`?

api/src/backend/api/models.py
```diff
@@ -165,6 +172,281 @@ class JSONAPIMeta:
...
+
+    @property
+    def is_anonymous(self):
+        """API key users are not anonymous."""
+        return False
+
+    @property
+    def is_active(self):
+        """API key users are considered active by default."""
+        return True
```
jfagoagas, line 204: Is this strictly required?

api/src/backend/api/models.py
```diff
@@ -165,6 +172,281 @@ class JSONAPIMeta:
...
+        return super().get_queryset()
+
+    def get_usable_keys(self, tenant_id=None):
+        """
+        Get all usable (non-revoked, non-expired) API keys.
+        If tenant_id is provided, filter by tenant.
+        """
+        queryset = self.filter(revoked=False)
+        if tenant_id:
+            queryset = queryset.filter(tenant_id=tenant_id)
```
jfagoagas: @AdriiiPRodri @vicferpoy @josemazo is this using RLS correctly?

@mmuller88 why `tenant_id` is a parameter? Everything is bound to tenants. Please review this for the whole class.

api/src/backend/api/models.py
```diff
@@ -165,6 +172,281 @@ class JSONAPIMeta:
...
+        """
+        Check if a key is valid without raising exceptions.
+        """
+        try:
+            api_key = self.get_from_key(key, tenant_id)
+            return not api_key.has_expired
+        except self.model.DoesNotExist:
+            return False
+
+    def create_key(self, tenant_id, role, name, **kwargs):
```
jfagoagas, line 270: If views are not implemented in this PR, why are we creating extra functions? I think it's easier to review once these functions are used.

api/src/backend/api/models.py
```diff
@@ -165,6 +172,281 @@ class JSONAPIMeta:
...
+        """
+        Create a new API key for a specific tenant and role.
+        Returns tuple of (api_key_instance, raw_key_string).
+        """
+        # Remove fields that shouldn't be passed to create
+        kwargs.pop("id", None)
+        kwargs.pop("prefix", None)
+        kwargs.pop("hashed_key", None)
```
jfagoagas, line 278: Is this needed too?

api/src/backend/api/models.py
jfagoagas: I thought we were going to use PBKDF2 as is the same the API uses for hashing.

josema-xyz (changes requested):
Just a simple comment that is going to improve the user experience quality with the API keys: only to use characters that can select the whole API key with a double click.
So, instead of using the dot (`.`) it would be better to use an underscore (`_`) or even nothing, so the two parts of the API key look as one.
Yes, the UI can have a button for copy the API key to the clipboard, but API keys are going to be in multiple places: clients like Postman or Swagger, terminals or code editors.
It's just a small change, and surely people, or even us, are going to request that, so, now it's the moment for doing it.

api/src/backend/api/models.py
```diff
@@ -165,6 +172,281 @@ class JSONAPIMeta:
...
+        """Update the last_used_at timestamp."""
+        self.last_used_at = timezone.now()
+        self.save(update_fields=["last_used_at"])
+
+    def __str__(self):
+        """Return string representation of the API key."""
+        return f"API Key: {self.name}"
+
+    @classmethod
+    def generate_key(cls):
```
jfagoagas: Is this required again? We have a `create_key` using a different code.

api/src/backend/api/models.py
```diff
@@ -165,6 +172,281 @@ class JSONAPIMeta:
...
+        Hash an API key using Django's password hashing with salt.
+
+        Args:
+            key: The raw API key string
+
+        Returns:
+            str: The hashed key
+        """
+        # Use Django's default password hashing which includes salt
+        return make_password(key)
```
jfagoagas, line 409: I need guidance to understand what's the flow, why are we using this here?

jfagoagas (changes requested):

api/src/backend/api/migrations/0048_add_api_key_model.py
mmuller88: @jfagoagas yes good question. The framework rest_framework_api_key requires to have an id as api key value. But that one is not well suited for API CRUD operation where we usually have uuids

jfagoagas: I thought we decided not to use `rest_framework_api_key` as we need something custom to be integrated with our multi-tenancy and RBAC.

Am I right? @AdriiiPRodri @josemazo @vicferpoy

AdriiiPRodri: We decided to use it because, just like with SocialApp (Google, GitHub, and SAML), the library is not designed for multitenancy. However, it’s easy to handle this on our side, and by doing so we can delegate part of the functionality to widely tested libraries

api/src/backend/api/models.py
mmuller88: @jfagoagas Yes I removed it

api/src/backend/api/models.py
mmuller88: Yes I added the expired to the filter

api/src/backend/api/models.py
mmuller88: addressed this. the extract_prefix(cls, key) will always return including pk_

api/src/backend/api/models.py
mmuller88: I removed the Sha512ApiKeyHasher which will default back to PBKDF2

api/src/backend/api/models.py
mmuller88: I added the filter to regard expires

queryset = self.filter(
            models.Q(revoked_at__isnull=True)
            & (models.Q(expires_at__isnull=True) | models.Q(expires_at__gt=now))
        )

api/src/backend/api/migrations/0048_add_api_key_model.py
mmuller88: I ran makemigrations and created the script that way

api/src/backend/api/models.py
mmuller88: removed it

api/src/backend/api/migrations/0048_add_api_key_model.py
mmuller88: renamed it

api/src/backend/api/migrations/0048_add_api_key_model.py
mmuller88: renamed it

api/src/backend/api/migrations/0048_add_api_key_model.py
mmuller88: renamed it

api/src/backend/api/migrations/0048_add_api_key_model.py
mmuller88: the frameworks is predefining it for 8 chars 

https://github.com/florimondmanca/djangorestframework-api-key/blob/02f68e85d2e51f8a2049cb15ee4d5466d38cf1db/docs/security.md?plain=1#L9

api/src/backend/api/models.py
mmuller88: make_password was suggested to use in a previous comment on the previous PR.

jfagoagas:
I'm closing this PR as we are working on in a different branch/PR that will be pushed this week. Thanks!

PR #8690, opened 2025-09-09
feat: Verify that the CheckID is the same as the filename and classname in the Check class

OlmeNav:
### Context

Ensure that the `CheckID`, the class name and the file name where the check is defined all match. This corresponds to a TODO comment in the file `prowler/lib/check/model.py`, specifically in the `__init__` method of the class `Check`.

### Description

Add name consistency control at the end of the method. In case that names do not match, it is logged with 'logger.error'. There are some aspects to consider:

- In some small tests I performed, if the `CheckID` is modified it does not break the execution itself, only the Check ID in the output is changed.  
- If the file name is modified, the corresponding `.json` file is not recognized and the execution aborts.  
 - If the class name is modified, the scan completes but no findings remain. Therefore, it does not appear to be a critical problem.  
- The logger level at that point is 50, so the message is not actually registered. Maybe it should be logged with `logger.critical`, but I am not sure if it is that relevant.

### Checklist

#### API

### License


prowler/lib/check/models.py
```diff
@@ -436,17 +436,31 @@ class Check(ABC, CheckMetadata):
...
+        if check_id != class_name:
+            errors.append(f"CheckID '{check_id}' != class name '{class_name}'")
+        if check_id != file_name:
+            errors.append(f"CheckID '{check_id}' != file name '{file_name}'")
+
+        if errors:
+            logger.error(
+                "Name consistency error. The CheckID in the metadata, the class name "
+                "and the file name must all match.  -- " + " | ".join(errors)
+            )
```
HugoPBrito:
```suggestion
           raise ValidationError(
                "Name consistency error. The CheckID in the metadata, the class name "
                "and the file name must all match.  -- " + " | ".join(errors)
            )
```

HugoPBrito: Please change this to a ValidationError.

HugoPBrito (changes requested):
Hi @OlmeNav,

Thanks for your contribution!
Could you please do the requested changes, merge the master branch and add an entry in the changelog under the _Added_ section for v5.13?

Something like:
- Equality validation for CheckID, filename and classname [#8690](https://github.com/prowler-cloud/prowler/pull/8690)

prowler/lib/check/models.py
OlmeNav: Hi @HugoPBrito,

Thanks for the feedback! 
I've addressed the requested changes, merge the master branch and add an entry in the changelog. 
Please let me know if there’s anything else I should adjust.

jfagoagas (approved):

PR #8702, opened 2025-09-11
fix(defender): change policies rules key

danibarranqueroo:
### Context

Microsoft 365 Defender checks were failing with KeyError when accessing policy rules using policy names. The issue occurred when:

1. Policy names were changed using PowerShell Set-AntiPhishRule -Name "New Name"
2. Microsoft Powershell cmdlet is not working as expected, because while Name and Identity fields updated correctly, the AntiPhishPolicy field retained the original name
3. Code tried to access antiphishing_rules[policy.name] using the new name, but the rule was still stored with the original name in the AntiPhishPolicy field

### Description

Modified defender_service.py to use the Old Policy Name field instead of the Name field when mapping rules.

### Steps to review

Review new code and test it with M365 policies.
### Checklist

#### API

### License

### DIFF
--- a/prowler/CHANGELOG.md
+++ b/prowler/CHANGELOG.md
@@ -7,11 +7,11 @@ All notable changes to the **Prowler SDK** are documented in this file.
 ### Added
 - Support for AdditionalURLs in outputs [(#8651)](https://github.com/prowler-cloud/prowler/pull/8651)
 - Support for markdown metadata fields in Dashboard [(#8667)](https://github.com/prowler-cloud/prowler/pull/8667)
-### Changed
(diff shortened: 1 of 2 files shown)


prowler/CHANGELOG.md
HugoPBrito, line 1: Review the structure of this file, it is not correct.

HugoPBrito (changes requested):

prowler/CHANGELOG.md
danibarranqueroo: It was not correct before, I've fixed it in this PR, look at master branch because there were duplicated "changed" https://github.com/prowler-cloud/prowler/blob/master/prowler/CHANGELOG.md

HugoPBrito (approved):

PR #8729, opened 2025-09-15
feat(auth): add safeguards to prevent self-role removal and enforce MANAGE_ACCOUNT role presence

AdriiiPRodri:
**This PR shouldn't be merged until we merge this one: https://github.com/prowler-cloud/prowler/pull/8281**

### Context

Up until now, users with the `MANAGE_ACCOUNT` permission were allowed to delete roles, even if it was the last remaining role with that permission. This could leave the tenant in an unrecoverable state, as no other user would have the necessary permissions to revert the change.

### Description

This PR introduces safeguards to prevent that scenario. From now on, a role with `MANAGE_ACCOUNT` can only be deleted if there is at least one other role in the system that also has this permission.

### Steps to review

### Checklist

#### API

### License


api/src/backend/api/v1/views.py
```diff
@@ -932,6 +932,43 @@ class UserRoleRelationshipView(RelationshipView, BaseRLSViewSet):
...
+        tenant_id = self.request.tenant_id
+        payload = request.data if isinstance(request.data, dict) else None
+
+        # If an user has more than one role, we will delete the relationship with the roles in the payload
+        data = payload.get("data") if payload else None
+        if data:
+            try:
+                role_ids = [item["id"] for item in data]
+            except Exception:
+                role_ids = []
```
vicferpoy: Could we narrow a little more the exception handling here? Or is it needed because there could be unexpected ones?

api/src/backend/api/v1/views.py
```diff
@@ -932,6 +932,43 @@ class UserRoleRelationshipView(RelationshipView, BaseRLSViewSet):
...
+            return Response(
+                data={
+                    "detail": "Users cannot delete the relationship with their role."
+                },
+                status=status.HTTP_400_BAD_REQUEST,
+            )
+        tenant_id = self.request.tenant_id
+        payload = request.data if isinstance(request.data, dict) else None
+
+        # If an user has more than one role, we will delete the relationship with the roles in the payload
```
```suggestion
        # If a user has more than one role, we will delete the relationship with the roles in the payload
```

api/src/backend/api/v1/serializers.py
```diff
@@ -1741,6 +1769,37 @@ def update(self, instance, validated_data):
...
+            request = self.context.get("request")
+            if request and getattr(request, "user", None):
+                request_user = request.user
+                is_currently_assigned = instance.users.filter(
+                    id=request_user.id
+                ).exists()
+                will_be_assigned = any(u.id == request_user.id for u in users)
+                if is_currently_assigned and not will_be_assigned:
+                    raise serializers.ValidationError(
+                        {"users": "Users cannot remove their own role assignment."}
```
vicferpoy: "role assignment" sounds strange to me. Would "their own role" be enough?

vicferpoy (changes requested):
Could we add some kind of documentation for the affected endpoints with the constraints? In the endpoint description or something.

api/src/backend/api/v1/views.py
AdriiiPRodri: We can check for a `KeyError` thanks! This code is for the future. In this moment we don't allow having more than one role for user

api/src/backend/api/v1/serializers.py
AdriiiPRodri: Sure, is enough

AdriiiPRodri:
> Could we add some kind of documentation for the affected endpoints with the constraints? In the endpoint description or something.

Sure we can, I just added it

vicferpoy (approved):
Great job!

PR #8772, opened 2025-09-26
feat: add lighthouse support for multiple providers

0xbadshah:
### Context

Existing implementation of Lighthouse supports single LLM provider - OpenAI. User has to add their OpenAI credential and start interacting. The models that they can interact were hardcoded. This way of configuration had few issues:
1. Single LLM provider supported - users can't test with other LLM providers
2. Hardcoded model name - users are dependent on Prowler team to modify code to support more models (even from OpenAI)
3. Doesn't support connecting any self hosted OpenAI compatible providers

### Description

This PR introduces 3 new API endpoints:
1. `/api/v1/lighthouse/config` - Singleton URL that allows setting and fetching tenant level configuration (like business_context, default_provider, default_models)
2. `/api/v1/lighthouse/providers` - URL endpoint that allows adding new LLM providers. With this PR, only OpenAI is supported. Additional LLM providers will be added once this PR is merged.
3. `/api/v1/lighthouse/models` - Readonly URL endpoint that lists the models supported by the LLM provider

Two more helper API endpoints:
1. `/api/v1/lighthouse/providers/PROVIDER-UUID/connection` - Verifies if the added LLM provider key is valid and updates the `is_active` field of the providers
2. `/api/v1/lighthouse/providers/PROVIDER-UUID/refresh-models` - Refreshes and fetches model names from the LLM provider.

### Steps to review

This is the intended user flow

**For Existing Lighthouse AI Configuration**

If OpenAI configuration exists, then copy relevant data (API key, business_context, etc) to the new tables of LighthouseTenantConfig and LighthouseProviderConfig. After the first migration, the LLM models are not refreshed - so `/api/v1/lighthouse/models` will be empty even for valid credentials. Triggering `/api/v1/lighthouse/providers/PROVIDER-UUID/refresh-models` will fetch the models that the LLM provider token has access to.

**For New Lighthouse AI Configuration**

1. User add a new provider using the API endpoint `/api/v1/lighthouse/providers`. This encrypts the credential and stores in the DB.
2. By default, the is_active is set to True for every configured LLM provider. User triggers `/api/v1/lighthouse/providers/PROVIDER-UUID/connection` to validate the credentials. This will initiate a celery task. If the creds are valid, then the is_active is left to True, else its updated to False.
3. If is_active is still True, user can trigger `/api/v1/lighthouse/providers/PROVIDER-UUID/refresh-models` to fetch all the supported models and store it in DB for the entire tenant
4. Once the refresh is completed, the models are available at `/api/v1/lighthouse/models`
5. User can set business_context, default_provider and default_models using `/api/v1/lighthouse/config`

### Checklist

#### API

### License


api/src/backend/api/models.py
```diff
@@ -1976,3 +1976,319 @@ class Meta(RowLevelSecurityProtectedModel.Meta):
...
+        blank=False, null=False, help_text="Encrypted JSON credentials for the provider"
+    )
+
+    is_active = models.BooleanField(default=True)
+
+    def __str__(self):
+        return f"{self.get_provider_type_display()} ({self.tenant_id})"
+
+    def clean(self):
+        super().clean()
```
AdriiiPRodri, line 1998: Can we remove this method or we want to do some verifications? Or we want to call full_clean?

```python
def save(self, *args, **kwargs):
    self.full_clean()
    super().save(*args, **kwargs)
```

api/src/backend/api/models.py
```diff
@@ -1976,3 +1976,319 @@ class Meta(RowLevelSecurityProtectedModel.Meta):
...
+    default_provider = models.CharField(max_length=50, blank=True)
+
+    # Mapping of provider -> model id, e.g., {"openai": "gpt-4o", "bedrock": "anthropic.claude-v2"}
+    default_models = models.JSONField(default=dict, blank=True)
+
+    def __str__(self):
+        return f"Lighthouse Tenant Config for {self.tenant_id}"
+
+    def save(self, *args, **kwargs):
+        self.full_clean()
```
AdriiiPRodri: Why we need to do this?

api/src/backend/api/v1/views.py
```diff
@@ -4151,6 +4166,309 @@ def connection(self, request, pk=None):
...
+        try:
+            payload = json.loads(request.body)
+            attributes = payload.get("data", {}).get("attributes", {})
+        except (json.JSONDecodeError, AttributeError):
+            raise ValidationError("Invalid JSON:API payload")
+
+        serializer = self.get_serializer(instance, data=attributes, partial=True)
+        serializer.is_valid(raise_exception=True)
+        serializer.save()
+        return Response(serializer.data)
```
AdriiiPRodri: I think DRF already parses the payload via `request.data`. If we don't use it we can omit some error messages, middlewares, etc.
```suggestion
        instance, created = LighthouseTenantConfiguration.objects.get_or_create(
        tenant_id=self.request.tenant_id,
        defaults={},
    )

    serializer = self.get_serializer(instance, data=request.data, partial=True)
    serializer.is_valid(raise_exception=True)
    serializer.save()

    # Return with read serializer
    read_serializer = LighthouseTenantConfigSerializer(
        instance, context=self.get_serializer_context()
    )
    return Response(read_serializer.data, status=status.HTTP_200_OK)
```

api/src/backend/api/v1/urls.py
```diff
@@ -137,6 +150,14 @@
...
         name="provider_group-providers-relationship",
     ),
+    # Lighthouse tenant config as singleton endpoint
+    path(
+        "lighthouse/config",
+        LighthouseTenantConfigViewSet.as_view(
+            {"get": "retrieve", "post": "create", "patch": "partial_update"}
+        ),
+        name="lighthouse-config",
+    ),
```
AdriiiPRodri, line 160: I think this is not a true singleton, actually we don't need the ID, we ignore it. So I think we can remove retrieve, create and destroy and only keep list (returns single item) and partial_updated

api/src/backend/api/models.py
```diff
@@ -1976,3 +1976,319 @@ class Meta(RowLevelSecurityProtectedModel.Meta):
...
+            models.UniqueConstraint(
+                fields=["tenant_id"], name="unique_tenant_lighthouse_config"
+            ),
+        ]
+
+    class JSONAPIMeta:
+        resource_name = "lighthouse-config"
+
+
+class LighthouseProviderModels(RowLevelSecurityProtectedModel):
```
AdriiiPRodri, line 2097: We don't have a save method so we don't call full_clean and also we don't override the clean method? Is this correct? Maye there is an inconsistency here

Or maybe we don't need to override save method en each class

api/src/backend/api/v1/views.py
```diff
@@ -4151,6 +4166,309 @@ def connection(self, request, pk=None):
...
+        self.check_object_permissions(self.request, obj)
+        return obj
+
+    def retrieve(self, request, *args, **kwargs):
+        """GET endpoint for singleton - no pk required."""
+        instance = self.get_object()
+        serializer = self.get_serializer(instance)
+        return Response(serializer.data)
+
+    def partial_update(self, request, *args, **kwargs):
```
AdriiiPRodri, line 4486: We usually return the read serializer after the update


```
read_serializer = LighthouseTenantConfigSerializer(
        instance, context=self.get_serializer_context()
    )
    return Response(read_serializer.data, status=status.HTTP_200_OK)
```

api/src/backend/api/v1/views.py
```diff
@@ -4151,6 +4166,309 @@ def connection(self, request, pk=None):
...
+    def create(self, request, *args, **kwargs):
+        raise MethodNotAllowed(method="POST")
+
+    @extend_schema(exclude=True)
+    def partial_update(self, request, *args, **kwargs):
+        raise MethodNotAllowed(method="PATCH")
+
+    @extend_schema(exclude=True)
+    def destroy(self, request, *args, **kwargs):
+        raise MethodNotAllowed(method="DELETE")
```
AdriiiPRodri: I believe `http_method_names = ["get"]` already blocks POST/PATCH/DELETE at the router level

api/src/backend/api/v1/serializers.py
```diff
@@ -2836,3 +2841,275 @@ class Meta:
             "inserted_at": {"read_only": True},
             "last_used_at": {"read_only": True},
         }
+
+
+# Lighthouse: Provider configurations
+
+
+class LighthouseProviderConfigSerializer(RLSSerializer):
```
AdriiiPRodri, line 2978: This apply to all serializers, we usually perform the input validation before passing to model. I think that we should the same here, as other serializers do. Check examples like: IntegrationCreateSerializer

api/src/backend/api/v1/serializers.py
```diff
@@ -2836,3 +2841,275 @@ class Meta:
...
+            "url",
+        ]
+        extra_kwargs = {
+            "id": {"read_only": True},
+            "inserted_at": {"read_only": True},
+            "updated_at": {"read_only": True},
+            "is_active": {"read_only": True},
+            "url": {"read_only": True, "view_name": "lighthouse-providers-detail"},
+        }
+
```
AdriiiPRodri, line 3005: Serializer validation **example**:

```
def validate_credentials(self, value):
    """Validate credentials structure at serializer level."""
    if not isinstance(value, dict):
        raise ValidationError("Credentials must be a JSON object")

    # Basic structure check (detailed validation in model)
    if not value:
        raise ValidationError("Credentials cannot be empty")

    return value
```

api/src/backend/tasks/jobs/lighthouse_providers.py
```diff
@@ -0,0 +1,120 @@
...
+
+import openai
+from celery.utils.log import get_task_logger
+
+from api.models import LighthouseProviderConfiguration, LighthouseProviderModels
+
+logger = get_task_logger(__name__)
+
+
+def _extract_openai_api_key(
```
AdriiiPRodri, line 11: Missing docstring

api/src/backend/tasks/jobs/lighthouse_providers.py
```diff
@@ -0,0 +1,120 @@
...
+    creds = provider_cfg.credentials_decoded
+    if not isinstance(creds, dict):
+        return None
+    api_key = creds.get("api_key")
+    if not isinstance(api_key, str) or not api_key:
+        return None
+    return api_key
+
+
+def check_lighthouse_provider_connection(provider_config_id: str) -> Dict:
```
AdriiiPRodri, line 33: Missing params in the docstring

api/src/backend/api/models.py
```diff
@@ -1976,3 +1976,319 @@ class Meta(RowLevelSecurityProtectedModel.Meta):
...
+        api_key = credentials.get("api_key")
+        if not isinstance(api_key, str) or not api_key:
+            raise ModelValidationError(
+                detail="OpenAI credentials must include 'api_key'",
+                code="invalid_openai_credentials",
+                pointer="/data/attributes/credentials/api_key",
+            )
+
+        # Validate OpenAI API key format
+        openai_key_pattern = r"^sk-[\w-]+T3BlbkFJ[\w-]+$"
```
AdriiiPRodri: This is OpenAI in base64? What happens if they change it, can we use something like `r"^sk-[\w-]+$"`?

AdriiiPRodri (changes requested):

api/src/backend/api/models.py
0xbadshah: OpenAI is not base64 but has this pattern `T3BlbkFJ` in every API key. 

> What happens if they change it, can we use something like r"^sk-[\w-]+$"?
If they change it we have to update the code. We can simplify the check to just match the above regex you mentioned.

AdriiiPRodri: I mean, if we decode this `T3BlbkFJ`, the value is OpenAI

api/src/backend/tasks/jobs/lighthouse_providers.py
0xbadshah: Updated. Thanks!

api/src/backend/tasks/jobs/lighthouse_providers.py
0xbadshah: Updated. Thanks!

api/src/backend/api/models.py
0xbadshah: Cool. I've made the regex simpler to just check if the key begins with `sk-`.

api/src/backend/api/v1/serializers.py
0xbadshah: I have moved all validation logic from models to serializers with this https://github.com/prowler-cloud/prowler/pull/8772/commits/8d9e804288bfc9c180a57632ecbe2aada6f38bf4. I've created a util file so we can have all the lighthouse provider logic in same file - https://github.com/prowler-cloud/prowler/pull/8772/commits/7860ba5a6667a269d7988a681ee420a05d8d5426. Thanks!

api/src/backend/api/v1/serializers.py
0xbadshah: I've updated the validation logic. Thanks :)

api/src/backend/api/v1/views.py
0xbadshah: Updated it. Thanks!

api/src/backend/api/v1/views.py
0xbadshah: Correct. Removed it. Thanks!

api/src/backend/api/models.py
0xbadshah: Correct. I removed the JSONAPIMeta resource_name as there was just a single GET endpoint that allows fetching data.

api/src/backend/api/models.py
0xbadshah: Done. Removed save() as there was no additional verification needed. Thanks!

api/src/backend/api/models.py
0xbadshah: Makes sense. Removed unnecessary save methods.

api/src/backend/api/v1/urls.py
0xbadshah: Thanks a lot! Refactored code and made it better.

api/src/backend/api/v1/views.py
```diff
@@ -4176,6 +4190,273 @@ def connection(self, request, pk=None):
...
+            tenant_id=self.request.tenant_id,
+            defaults={},
+        )
+
+        # Extract attributes from JSON:API payload
+        try:
+            payload = json.loads(request.body)
+            attributes = payload.get("data", {}).get("attributes", {})
+        except (json.JSONDecodeError, AttributeError):
+            raise ValidationError("Invalid JSON:API payload")
```
AdriiiPRodri, line 4499: The component in charge of validate if the JSON is valid or not should be the serializer. For the moment is ok, this is only a recomendation, if you can we can change this into, and use request.data:


```
def partial_update(self, request, *args, **kwargs):
    instance, created = LighthouseTenantConfiguration.objects.get_or_create(
        tenant_id=self.request.tenant_id,
        defaults={},
    )
    
    serializer = self.get_serializer(instance, data=request.data, partial=True)
    serializer.save()
    
    read_serializer = LighthouseTenantConfigSerializer(
        instance, context=self.get_serializer_context()
    )
    return Response(read_serializer.data, status=status.HTTP_200_OK)
```

AdriiiPRodri (changes requested):

api/src/backend/api/v1/views.py
0xbadshah: This was done because `LighthouseTenantConfig` is a singleton pattern. If we don't do this, then serializer will expect the ID to be passed in the URL. The existing implementation values the attributes from JSON.

AdriiiPRodri (dismissed):

api/src/backend/api/v1/urls.py
```diff
@@ -137,6 +150,14 @@
         ),
         name="provider_group-providers-relationship",
     ),
+    # Lighthouse tenant config as singleton endpoint
+    path(
+        "lighthouse/config",
```
vicferpoy:
```suggestion
        "lighthouse/configuration",
```

vicferpoy (changes requested):
@Chan9390 We need to deprecate this endpoint: https://api.prowler.com/api/v1/docs#tag/Lighthouse-AI/operation/lighthouse_configurations_list

You have one working example in `views.py`, for Findings: https://api.prowler.com/api/v1/docs#tag/Finding/operation/findings_findings_services_regions_retrieve

api/src/backend/api/migrations/0050_lighthouse_multi_llm.py
```diff
@@ -0,0 +1,273 @@
...
+
+        try:
+            # Use rls_transaction to set tenant context for RLS-protected tables
+            with rls_transaction(tenant_id, using=MainRouter.admin_db):
+                # Query old config for this specific tenant
+                old_config = (
+                    LighthouseConfiguration.objects.using(MainRouter.admin_db)
+                    .filter(tenant=tenant)
+                    .first()
+                )
```
vicferpoy: Why can't we just iterate over the existing LighthouseConfiguration instead of checking all the tenants?

api/src/backend/api/migrations/0050_lighthouse_multi_llm.py
```diff
@@ -0,0 +1,273 @@
...
+
+                if not old_config:
+                    # Skip tenants without lighthouse configuration
+                    continue
+
+                # Create OpenAI provider configuration for this tenant
+                api_key_decrypted = fernet.decrypt(bytes(old_config.api_key)).decode()
+                credentials_encrypted = fernet.encrypt(
+                    json.dumps({"api_key": api_key_decrypted}).encode()
+                )
```
vicferpoy: You don't need to decrypt it, right? You can just copy it to the new table.

api/src/backend/api/migrations/0050_lighthouse_multi_llm.py
```diff
@@ -0,0 +1,273 @@
...
+                    model_id=old_config.model,
+                    model_name=old_config.model,
+                    default_parameters={},
+                )
+
+        except Exception as e:
+            print(
+                f"Failed to migrate lighthouse config for tenant {tenant_id}: {e}",
+                exc_info=True,
+            )
```
vicferpoy: Use the API logger here, please.

api/src/backend/api/models.py
```diff
@@ -1984,3 +1959,210 @@ class Meta(RowLevelSecurityProtectedModel.Meta):
...
+                updated = True
+
+            if tenant_cfg.default_provider == self.provider_type:
+                tenant_cfg.default_provider = ""
+                updated = True
+
+            if updated:
+                tenant_cfg.save()
+
+        return super().delete(*args, **kwargs)
```
vicferpoy: This won't apply if we use `.delete()` after a `.filter()` in a queryset. Just FYI in case we haven't checked it.

api/src/backend/api/models.py
```diff
@@ -1984,3 +1959,210 @@ class Meta(RowLevelSecurityProtectedModel.Meta):
...
+        """
+        Set and encrypt credentials (assumes serializer performed validation).
+        """
+        if not value:
+            raise ModelValidationError(
+                detail="Credentials are required",
+                code="invalid_credentials",
+                pointer="/data/attributes/credentials",
+            )
+        self.credentials = fernet.encrypt(json.dumps(value).encode())
```
vicferpoy, line 2027: Isn't the field called `credentials`? Why `credentials_decoded`? Isn't this the same use case as the provider secrets? Just asking in case I did not understand it.

vicferpoy (changes requested):

api/src/backend/api/models.py
0xbadshah: > Isn't the field called credentials?

Correct. The field is credentials which is a valid JSON object. This JSON objects are encrypted at rest.

> Why credentials_decoded?

This is to get decrypted credentials accessible to NextJS so it can initialize langgraph and connect with LLM providers. We need to decrypt it before returning the creds.

> Isn't this the same use case as the provider secrets?

Not really. The provider secrets are meant to be accessible within the Django server and Celery worker tasks. Hence they don't need an explicit decrypt as Django code can retrieve decrypted creds. But for LLM provider creds, these are required by NextJS langgraphJS code.

api/src/backend/api/migrations/0050_lighthouse_multi_llm.py
0xbadshah: Existing LighthouseConfiguration is tenant specific. There can be multiple tenants in the system with different lighthouse configs. For the migration, it must migrate from old config to new config for all the tenants right who have lighthouse config right?

api/src/backend/api/models.py
0xbadshah: Moved the logic to signals, so bulk delete (if any) will be handled

api/src/backend/api/migrations/0050_lighthouse_multi_llm.py
0xbadshah: @vicferpoy There's a slight change.
Initially we just supported OpenAI so the api_key was always OpenAI key. But in the multi LLM we are supporting multiple LLM providers, and each provider can have different type of API keys (ex OpenAI has API key but Bedrock has Access+Secret key pair). So we store the credentials as a encrypted JSON like `{"api_key": "sk-dummy"}`. Hence decryption to get the key and then re-encrypt after creating the JSON object.

api/src/backend/api/migrations/0050_lighthouse_multi_llm.py
vicferpoy: Yes, but why iterate over every single tenant when most of them will not have a related `LighthouseConfiguration`? Isn't it better to check entries from `LighthouseConfiguration` and retrieve their `tenant_id` from the existing rows? In terms of efficiency I mean.

api/src/backend/api/migrations/0050_lighthouse_multi_llm.py
vicferpoy: Oh I see. Thanks for clarifying.

AdriiiPRodri (dismissed):

vicferpoy (approved):

PR #8787, opened 2025-09-30
fix: move delete user 500 error fix to its right version

josema-xyz:
### Description

Setting the changelog for https://github.com/prowler-cloud/prowler/pull/8731 in its right place.

### License

### DIFF
--- a/api/CHANGELOG.md
+++ b/api/CHANGELOG.md
@@ -12,6 +12,10 @@ All notable changes to the **Prowler API** are documented in this file.
 - Now the MANAGE_ACCOUNT permission is required to modify or read user permissions instead of MANAGE_USERS [(#8281)](https://github.com/prowler-cloud/prowler/pull/8281)
 - Now at least one user with MANAGE_ACCOUNT permission is required in the tenant [(#8729)](https://github.com/prowler-cloud/prowler/pull/8729)
 
+---
+
+## [1.13.2] (Prowler 5.12.3)
+
 ### Fixed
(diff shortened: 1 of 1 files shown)


.python-version
jfagoagas, line 1: Why this?

josema-xyz: Rushing doing `git add -A`. Already deleted.

api/CHANGELOG.md
```diff
@@ -12,6 +12,10 @@ All notable changes to the **Prowler API** are documented in this file.
 - Now the MANAGE_ACCOUNT permission is required to modify or read user permissions instead of MANAGE_USERS [(#8281)](https://github.com/prowler-cloud/prowler/pull/8281)
 - Now at least one user with MANAGE_ACCOUNT permission is required in the tenant [(#8729)](https://github.com/prowler-cloud/prowler/pull/8729)
 
+---
+
+## [1.13.2] (Prowler 5.12.3)
+
 ### Fixed
 - 500 error when deleting user [(#8731)](https://github.com/prowler-cloud/prowler/pull/8731)
```
jfagoagas, line 20: Can we add more context? Seems that we raise a 500 for all `DELETE` but I think it's only if the user has Social Login or SSO, right?

josema-xyz: I tested with the user that are created in the fixtures in the development environment, and they are only user/password (by their fixture definition).

AdriiiPRodri (approved):

PR #8805, opened 2025-10-01
feat(api-keys): Add API Key support for the Prowler API

vicferpoy:
### Description

  This PR adds API Key authentication support to the Prowler API, enabling programmatic access alongside
  JWT tokens. Users can now create, manage, and use API keys for authentication with tenant
  isolation, encryption, RBAC support and lifecycle management.

image

image

image

Core Features

  Authentication System

  - Added TenantAPIKeyAuthentication class extending drf-simple-apikey
  for API key validation
  - Implemented CombinedJWTOrAPIKeyAuthentication that supports both
  JWT Bearer tokens (Bearer

    ) and API keys (Api-Key )
  - API keys use encrypted format: pk_xxxxxxxx.encrypted_key with
  Fernet encryption
  - Automatic last_used_at tracking on authentication

  API Endpoints

  - POST /api-keys - Create new API key (returns plaintext key once)
  - GET /api-keys - List user's API keys
  - GET /api-keys/{id} - Retrieve specific API key details
  - PATCH /api-keys/{id} - Update API key name/settings
  - DELETE /api-keys/{id}/revoke - Revoke API key (soft delete)
  - Supports filtering, sorting, and pagination

  Security & Encryption

  - API keys encrypted using drf-simple-apikey's native Fernet
  encryption
  - Uses DJANGO_SECRETS_ENCRYPTION_KEY for symmetric encryption
  - Payload includes primary key and expiration timestamp for
  validation
  - Automatic prefix generation (pk_*) for key identification

  Logging & Middleware

  - Enhanced APILoggingMiddleware to log API key authentication events
  - Added api_key_prefix to request context for audit trails

  Exception Handling

  - Updated custom_exception_handler in exceptions.py for API key
  errors
  - Custom error messages for invalid/expired/revoked API keys

  Technical Details

  - Key Format: pk_xxxxxxxx.encrypted_payload
  - Prefix: 11-character identifier (pk_ + 8 random chars)
  - Payload: Fernet-encrypted JSON with {"_pk": "", "_exp": }

  - Tenant Isolation: All API keys respect RLS policies; users can only access keys within their tenant context.

  - Dynamic RBAC: All API keys are linked to the user that created
  them. When authenticated, the API key will assume the current RBAC
  roles that said user has in that moment.

  - Lifecycle: Keys can be revoked (soft delete) but not deleted;
  revocation is permanent.

### Checklist

#### API

### License


api/src/backend/api/v1/serializers.py
```diff
@@ -2735,3 +2736,103 @@ def update(self, instance, validated_data):
...
+            "prefix": {"read_only": True},
+            "revoked": {"read_only": True},
+            "entity": {"read_only": True},
+            "inserted_at": {"read_only": True},
+            "last_used_at": {"read_only": True},
+            "api_key": {"read_only": True},
+        }
+
+    def get_api_key(self, obj):
+        """Return the raw API key if it was stored during creation."""
```
AdriiiPRodri, line 2799: I have a question, the api keys are always stored during the creation, right?

api/src/backend/api/tests/test_views.py
```diff
@@ -7405,3 +7406,752 @@ def test_processors_create_another_with_same_type(
...
+        response = authenticated_client.get(reverse("api-key-list"))
+        assert response.status_code == status.HTTP_200_OK
+        data = response.json()["data"]
+        assert len(data) == len(api_keys_fixture)
+        # Verify keys are ordered by -created (newest first)
+        assert data[0]["attributes"]["name"] in [
+            "Test API Key 1",
+            "Test API Key 2",
+            "Revoked API Key",
+        ]
```
AdriiiPRodri: Can we improve this test? I think we are not checking the order

api/src/backend/api/v1/serializers.py
vicferpoy: They are never stored per se. We only store metadata from the API key, such as the prefix, database ID, revocation date, etc. What this method does is retrieving the raw API key on creation from the serializer, because we store it in the temporary serializer attributes to show it to the user ONLY on creation.

api/src/backend/api/authentication.py
```diff
@@ -0,0 +1,76 @@
...
+
+        # Prioritize JWT authentication if both are present
+        if auth_header.startswith("Bearer "):
+            return self.jwt_auth.authenticate(request)
+
+        if auth_header.startswith("Api-Key "):
+            return self.api_key_auth.authenticate(request)
+
+        # Default fallback
+        return self.jwt_auth.authenticate(request)
```
AdriiiPRodri, line 76: I believe we should replace the two `if` checks with an `if-elif` for security. An attacker could craft a header like `NoBearer <valid-JWT>`, if we don't strictly check the scheme, a valid JWT may still be accepted. We must be strict and accept only the `Bearer` or `Api-Key` schemes

AdriiiPRodri (changes requested):

api/src/backend/api/authentication.py
vicferpoy:
```python
# Default fallback
return self.jwt_auth.authenticate(request)
```

This function handles every other situation. We reviewed it live. 🚀

josema-xyz (approved):
Easy as we already saw, and well tested. A shame about using the dot as separator, but that's life.

AdriiiPRodri (approved):

api/src/backend/api/authentication.py
AdriiiPRodri: Thanks!

api/src/backend/api/v1/serializers.py
AdriiiPRodri: Amazing thanks mate!

PR #8849, opened 2025-10-06
fix(tests): Race condition on redundant API unit test

vicferpoy:
### Context

`.first()` was not accurate enough given the default ordering configuration for that endpoint and how we add those rows via fixture.

### Description

Since the default ordering is tested on another, more accurate test, we are removing that assert.

### Checklist

#### API

### License


josema-xyz (approved):
Less is better 😆 🚀

PR #8880, opened 2025-10-09
fix(api): throw errors for all non-ok responses

Alan-TheGentleman:
Simplify error handling by throwing errors for all non-ok HTTP responses
instead of returning error objects for 4XX responses.

### Context

Previously, `handleApiResponse` had inconsistent error handling:

- 5XX errors: threw exceptions ✅
- 4XX errors: returned error objects without throwing ❌

This inconsistency caused server actions to report success even when the
backend returned 4XX errors. For example, when submitting invalid SAML XML
metadata, the API would return 400 Bad Request, but the UI would show
"Configuration saved successfully!" because the error object was not being
caught by try/catch blocks.

This PR ensures all HTTP errors (4XX and 5XX) throw exceptions, making error
handling consistent across all server actions.

**Example scenario this fixes:**

1. User uploads invalid SAML metadata XML (missing required fields)
2. Backend validates and returns 400 with error message
3. Before: UI shows success toast (bug)
4. After: UI correctly shows error message from backend

### Changes

**Modified file:** `ui/lib/server-actions-helper.ts`

- Changed `handleApiResponse` to throw exceptions for all non-ok responses
- Maintained descriptive error messages for 5XX errors
- All existing try/catch blocks in server actions now properly catch other errors

### Testing

All server actions using `handleApiResponse` were verified to have try/catch
blocks, ensuring backward compatibility.

**Manual testing:**

- SAML integration with invalid XML → now shows proper error message
- Other server actions continue to work as expected

Screenshot 2025-10-09 at 10 17 50

### Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented on my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] New and existing unit tests pass locally with my changes
- [x] Verify if new entries are added to [CHANGELOG.md](<https://github.com/prow>
      ler-cloud/prowler/blob/master/prowler/CHANGELOG.md), if applicable.

#### API

- [x] Verify if API specs need to be regenerated. (N/A - UI only change)
- [x] Check if version updates are required (e.g., specs, Poetry, etc.). (N/A)
- [x] Ensure new entries are added to [CHANGELOG.md](<https://github.com/prowler>
      -cloud/prowler/blob/master/api/CHANGELOG.md), if applicable. (N/A)

### License

By submitting this pull request, I confirm that my contribution is made under
the terms of the Apache 2.0 license.


ui/lib/server-actions-helper.ts
```diff
@@ -29,17 +29,15 @@ export const handleApiResponse = async (
...
       throw new Error(
         errorDetail ||
           `Server error (${response.status}): The server encountered an error. Please try again later.`,
       );
     }
 
-    return errorsArray
-      ? { error: errorDetail, errors: errorsArray, status: response.status }
-      : ({ error: errorDetail, status: response.status } as any);
+    throw new Error(errorDetail || `Request error (${response.status})`);
```
alejandrobailo: We don't want the 4XX to be caught by the `Error.tsx` component; this is why we return an object here, that object is handled by the server action, and a Toast shows up.

alejandrobailo (changes requested):
This is the opposite behavior we expect; the ui/lib/server-actions-helper.ts was fine not throwing the error for the 4XX, instead the server actions user de error object to render the toast.

Now, for example, with your changes, forcing a 402 breaks the app:
https://github.com/user-attachments/assets/398a63a5-8f84-486d-b448-d1b7e2835c3e

ui/types/authFormSchema.ts
```diff
@@ -75,11 +75,12 @@ const baseAuthSchema = z.object({
 
 export const signInSchema = baseAuthSchema
   .extend({
-    password: z.string().min(1, { message: "Password is required." }),
+    password: z.string(),
   })
   .refine(
     (data) => {
       // If SAML mode, password is not required
+      console.log(data.isSamlMode);
```
AdriiiPRodri: Do we need this?

alejandrobailo: No

ui/types/authFormSchema.ts
```diff
@@ -75,11 +75,12 @@ const baseAuthSchema = z.object({
 
 export const signInSchema = baseAuthSchema
   .extend({
-    password: z.string().min(1, { message: "Password is required." }),
```
alejandrobailo, line 78: I think this breaks the normal auth flow as password won't show the message

alejandrobailo (changes requested):

alejandrobailo:
The branch has conflicts with the `changelog` then the build won't run in the CI, neither the e2e tests.

ui/lib/server-actions-helper.ts
Alan-TheGentleman: Without this change, no error appears and shows it like a success, because the catch is never executed so no toast appears

ui/types/authFormSchema.ts
Alan-TheGentleman: Tested and it's working, without it, it only works for the normal auth. On this change it always enters on the password condition to chech the SAML

AdriiiPRodri (dismissed):

alejandrobailo (approved):

PR #8925, opened 2025-10-15
feat: Update Lighthouse UI to support multi LLM

0xbadshah:
### Context

Lighthouse supports multiple LLM providers - https://github.com/prowler-cloud/prowler/pull/8772

### Description

This PR updates the frontend to support integrating different LLM providers (OpenAI, Amazon Bedrock, etc). It also replaces react-markdown with [streamdown](https://streamdown.ai/) that allows better rendering of markdown.

### Steps to review

https://github.com/user-attachments/assets/a4662265-996e-4ee6-b425-a8f3810a634d

### Checklist

#### API

### License


ui/actions/lighthouse/lighthouse.ts
```diff
@@ -4,172 +4,466 @@
...
+  const headers = await getAuthHeaders({ contentType: false });
+  const url = new URL(
+    `${apiBaseUrl}/lighthouse/providers/${providerId}/refresh-models`,
+  );
+
+  try {
+    const response = await fetch(url.toString(), {
+      method: "POST",
+      headers,
+    });
```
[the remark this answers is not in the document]
0xbadshah: Looks like a false positive. Will verify and fix/close this finding.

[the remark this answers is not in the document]
0xbadshah: While this is highlighting possible user controlled input to server component, it's not SSRF and the injection doesn't lead to security issues. Reasons are:
1. `url` is created from `${apiBaseUrl}` - which is fully server controlled
2. Any user input like provider ID can change the URL (like `/lighthouse/providers/../../../../sensitive/path/refresh-models`) but can't change the base URL. 

The max security issue is that user can make NextJS server access Django paths that the function is not supposed to. However, NextJS is just frontend for backend and the user can directly interact with Django REST APIs by sending the JWT token to them. Since the function `getAuthHeaders` is used to fetch the token of current user, any urls that attackers can make NextJS to hit will still be limited to the auth/authz assigned to the user.

I feel its a false positive for the above reasons. 
cc @alejandrobailo @jfagoagas

alejandrobailo:
Functional review: 

- Use the same cancel button as the rest of the app.
Screenshot 2025-11-04 at 15 00 54

- Set as default, and delete provider also must have same buttons styles as the other buttons in the app.
Screenshot 2025-11-04 at 15 01 28

- If you add a wrong key, the error displayed is not well formatted, please use "message" key.
Screenshot 2025-11-04 at 15 02 29

- After the error you are not able to add another key, you have to close go to the provider and edit it, this is not a correct behavior.
Screenshot 2025-11-04 at 14 59 59

alejandrobailo (changes requested):
I've added a comment with some functional review changes that need to be addressed.

0xbadshah:
@alejandrobailo Did the changes as follows:

1. Updated buttons similar to those in connect cloud provider flow

image

2. Updated button styles

image

3. Fixing this in the API side - https://github.com/prowler-cloud/prowler/pull/9165

After the PR is merged, the message looks as follows:

image

---

About this image:

image

This error occurs when the workflow errors out. There's no correct way to handle the error and display it as of now (as the error originates inside langgraph supervisor workflow). I'll try to fix it as part of a different PR.

ui/actions/lighthouse/lighthouse.ts
```diff
@@ -4,172 +4,507 @@ import { revalidatePath } from "next/cache";
...
+    const providersConfig = await Promise.all(
+      activeProviders.map(async (provider: any) => {
+        const providerType = provider.attributes.provider_type;
+        const defaultModelId = defaultModels[providerType];
+
+        // Fetch only the default model for this provider if it exists
+        let defaultModel = null;
+        if (defaultModelId) {
+          const headers = await getAuthHeaders({ contentType: false });
+          const url = `${apiBaseUrl}/lighthouse/models?filter[provider_type]=${providerType}&filter[model_id]=${defaultModelId}`;
```
alejandrobailo: No URL encoding for defaultModelId. If model ID contains special characters, this breaks.

```suggestion
          const url = `${apiBaseUrl}/lighthouse/models?filter[provider_type]=${providerType}&filter[model_id]=${defaultModelId}`;const url = `${apiBaseUrl}/lighthouse/models?${new URLSearchParams({
    'filter[provider_type]': providerType,
    'filter[model_id]': defaultModelId
  })}`;
```

ui/actions/lighthouse/lighthouse.ts
```diff
@@ -4,172 +4,507 @@ import { revalidatePath } from "next/cache";
 
 import { apiBaseUrl, getAuthHeaders } from "@/lib/helper";
 
-export const getAIKey = async (): Promise<string> => {
+/**
+ * Create a new lighthouse provider configuration
+ */
+export const createLighthouseProvider = async (config: {
+  provider_type: string;
+  credentials: Record<string, any>;
```
alejandrobailo: We must not use any types, we have to type models.

Try something like:

```ts
type ProviderCredentials =
    | { provider_type: 'openai'; credentials: { api_key: string } }
    | { provider_type: 'bedrock'; credentials: { access_key_id: string; secret_access_key: string; region: string } }
    | { provider_type: 'openai_compatible'; credentials: { api_key: string }; base_url: string };
```

ui/actions/lighthouse/lighthouse.ts
```diff
@@ -4,172 +4,507 @@ import { revalidatePath } from "next/cache";
...
+    const response = await fetch(url.toString(), {
+      method: "POST",
+      headers,
+      body: JSON.stringify(payload),
+    });
+
+    const data = await response.json();
+    return data;
+  } catch (error) {
+    console.error("[Server] Error in createLighthouseProvider:", error);
```
alejandrobailo: This is leaking sever sensitive data, sanitize the error please.

ui/actions/lighthouse/lighthouse.ts
```diff
@@ -4,172 +4,507 @@ import { revalidatePath } from "next/cache";
...
       method: "GET",
       headers,
     });
+
     const data = await response.json();
 
-    // Check if data array exists and has at least one item
     if (data?.data && data.data.length > 0) {
-      return data.data[0].attributes;
+      const provider = data.data[0].attributes;
```
alejandrobailo:
```suggestion
      const provider = data.data[0]?.attributes;
  if (!provider) {
    return { credentials: {} };
  }
```

ui/actions/lighthouse/lighthouse.ts
```diff
@@ -4,172 +4,507 @@ import { revalidatePath } from "next/cache";
...
+  }
+};
 
-    // Check if data array exists and has at least one item
-    if (data?.data && data.data.length > 0) {
-      return data.data[0].attributes.api_key || "";
+/**
+ * Get lighthouse provider models (fetches all pages)
+ */
+export const getLighthouseModels = async (providerType: string) => {
```
alejandrobailo: The getLighthouseModels function fetches all pages of models, but getLighthouseModelIds (line 158) only extracts IDs. For providers with hundreds of models (like Bedrock), this is wasteful.

Add the fields parameter &fields[lighthouse-models]=model_id,model_name to two URL:

1. getLighthouseModels 
2. getLighthouseProvidersConfig

ui/app/(prowler)/lighthouse/config/(connect-llm)/configure/page.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+function ConfigureContent() {
+  const router = useRouter();
+  const searchParams = useSearchParams();
+  const provider = searchParams.get("provider") || "";
+
+  useEffect(() => {
+    if (provider) {
+      // Simply redirect to connect workflow with edit mode
+      router.push(`/lighthouse/config/connect?provider=${provider}&mode=edit`);
```
alejandrobailo: Page loads only to immediately redirect. This adds unnecessary latency. Handle this routing at the server level or combine routes.

alejandrobailo (changes requested):
There are no input validation, credentials are passed as Record<string, any> with no schema validation.
Use Zod to validate:
- API key format/length
- AWS region validity
- Base URL format for OpenAI-compatible providers

ui/app/(prowler)/lighthouse/config/(connect-llm)/configure/page.tsx
0xbadshah: Thanks. Refactored the code and removed this redirection.

alejandrobailo (approved):

PR #8975, opened 2025-10-22
feat(api): enhance provider filtering and pagination capabilities

AdriiiPRodri:
### Context

To power the new Overview filters without adding extra endpoints, we extended the existing `/api/v1/providers` and `/api/v1/overviews/providers` endpoints. This keeps a single source of truth and avoids duplicating production logic.

### Description

- `GET /api/v1/providers` now supports `filter[provider_type]`, `filter[provider_type__in]`, and a `page[disable]=true` flag to fetch the full list without pagination. All other behaviour and schema remain unchanged.
- `GET /api/v1/overviews/providers` accepts `filter[group_by]=provider_type`, which aggregates findings by provider type and adds a count representing the total number of visible providers of that type. Without the filter, the endpoint behaves exactly as before.

#### Performance:
In a test environment containing:
- AWS: 233 providers
- Azure: 111 providers
- GCP: 109 providers
- Kubernetes: 56 providers

Both endpoints consistently return responses in < 100ms on average, with and without filters.

### Steps to review

### Checklist

#### API

### License


josema-xyz (dismissed):
Awesome work! 🚀

AdriiiPRodri:
## Status update

This is the explain analyze of the new endpoint, a very simple one to just get the existing provider_types in the system. In this output we can see that we are using the correct index:

```
SQL Query

SELECT "providers"."provider",
       COUNT("providers"."id") AS "count"
FROM "providers"
WHERE (NOT "providers"."is_deleted"
       AND "providers"."tenant_id" = 12646005-9067-4d2a-a098-8bb378604362)
GROUP BY "providers"."provider"
ORDER BY "providers"."provider" ASC

3.148ms
0 joins
Query Plan
Sort  (cost=57.23..57.24 rows=4 width=12)
  Sort Key: provider
  ->  HashAggregate  (cost=57.15..57.19 rows=4 width=12)
        Group Key: provider
        ->  Bitmap Heap Scan on providers  (cost=12.16..55.88 rows=254 width=20)
              Recheck Cond: (tenant_id = '12646005-9067-4d2a-a098-8bb378604362'::uuid)
              Filter: ((NOT is_deleted) AND CASE WHEN (current_setting('api.tenant_id'::text, true) IS NULL) THEN false ELSE (tenant_id = (current_setting('api.tenant_id'::text))::uuid) END)
              ->  Bitmap Index Scan on providers_tenant_id_b06669a6  (cost=0.00..12.09 rows=509 width=0)
                    Index Cond: (tenant_id = '12646005-9067-4d2a-a098-8bb378604362'::uuid) 
```

vicferpoy (approved):

PR #9053, opened 2025-10-28
fix(api): enhance overview provider aggregation and resource counting

AdriiiPRodri:
### Context

The /api/v1/overviews/providers endpoint is returning an incorrect response structure.

Currently, it returns a separate dictionary for each provider instead of aggregating results by provider type.

This behavior causes the UI to display incomplete data, as it only takes the first dictionary of each provider type (e.g., only the first AWS provider).

### Description

The new behavior is to return one aggregated dictionary per provider type (e.g., one for AWS, one for Azure, etc.), containing the combined data of all providers of that type.

### Steps to review

### Checklist

#### API

### License

### DIFF
--- a/api/CHANGELOG.md
+++ b/api/CHANGELOG.md
@@ -10,6 +10,11 @@ All notable changes to the **Prowler API** are documented in this file.
 - Support for configuring multiple LLM providers [(#8772)](https://github.com/prowler-cloud/prowler/pull/8772)
 - Support for Oracle Cloud Infrastructure (OCI) provider [(#8927)](https://github.com/prowler-cloud/prowler/pull/8927)
 
+## [1.14.1] (Prowler 5.13.1)
+
+### Fixed
+- `/api/v1/overviews/providers` collapses data by provider type so the UI receives a single aggregated record per cloud family even when multiple accounts exist [(#9053)](https://github.com/prowler-cloud/prowler/pull/9053)
+
 ## [1.14.0] (Prowler 5.13.0)
 
 ### Added
--- a/api/src/backend/api/specs/v1.yaml
+++ b/api/src/backend/api/specs/v1.yaml
@@ -5433,7 +5433,7 @@ paths:
         name: filter[provider_type]
         schema:
           type: string
-          x-spec-enum-id: 4c1e219dad1cc0e7
+          x-spec-enum-id: 6f034074d7104650
           enum:
           - aws
           - azure
@@ -5441,6 +5441,7 @@ paths:
           - github
           - kubernetes
           - m365
+          - oci
         description: |-
           * `aws` - AWS
           * `azure` - Azure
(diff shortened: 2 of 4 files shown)


AdriiiPRodri:
## Status update

I reworked the providers overview endpoint so it now aggregates findings and resources at the provider-type level instead of separately per account. That keeps the UI from dropping data when there are multiple providers of the same cloud family. Under the hood, the view uses the existing RLS filter (allowed_providers) and pulls resource totals through a single aggregated query, I also added a regression test that spins up a second AWS account to prove the numbers merge correctly.

For performance, we inspected the SQL Silk captured and saw Postgres doing a parallel sequential scan over the tenant’s entire resources table, basically every row had to be read because so many resources belong to that tenant. There isn’t an obvious missing index, the planner chooses a seq scan even with the (tenant_id, provider_id) index in place. To really speed it up we’d need to materialize/cache counts or restructure the query flow rather than just adding yet another index.

```
Query Plan
Finalize GroupAggregate  (cost=335237.79..335238.80 rows=4 width=12) (actual time=2127.116..2130.502 rows=4 loops=1)
  Group Key: providers.provider
  ->  Gather Merge  (cost=335237.79..335238.72 rows=8 width=12) (actual time=2127.094..2130.477 rows=12 loops=1)
        Workers Planned: 2
        Workers Launched: 2
        ->  Sort  (cost=334237.76..334237.77 rows=4 width=12) (actual time=2115.788..2115.791 rows=4 loops=3)
              Sort Key: providers.provider
              Sort Method: quicksort  Memory: 25kB
              Worker 0:  Sort Method: quicksort  Memory: 25kB
              Worker 1:  Sort Method: quicksort  Memory: 25kB
              ->  Partial HashAggregate  (cost=334237.68..334237.72 rows=4 width=12) (actual time=2115.767..2115.771 rows=4 loops=3)
                    Group Key: providers.provider
                    Batches: 1  Memory Usage: 24kB
                    Worker 0:  Batches: 1  Memory Usage: 24kB
                    Worker 1:  Batches: 1  Memory Usage: 24kB
                    ->  Hash Join  (cost=60.27..333719.16 rows=103705 width=20) (actual time=7.790..2089.418 rows=333350 loops=3)
                          Hash Cond: (resources.provider_id = providers.id)
                          ->  Parallel Seq Scan on resources  (cost=0.00..333112.17 rows=207411 width=32) (actual time=7.057..2052.386 rows=333350 loops=3)
                                Filter: ((tenant_id = '12646005-9067-4d2a-a098-8bb378604362'::uuid) AND CASE WHEN (current_setting('api.tenant_id'::text, true) IS NULL) THEN false ELSE (tenant_id = (current_setting('api.tenant_id'::text))::uuid) END)
                                Rows Removed by Filter: 333340
                          ->  Hash  (cost=53.91..53.91 rows=509 width=20) (actual time=0.703..0.704 rows=510 loops=3)
                                Buckets: 1024  Batches: 1  Memory Usage: 34kB
                                ->  Seq Scan on providers  (cost=0.00..53.91 rows=509 width=20) (actual time=0.285..0.667 rows=510 loops=3)
                                      Filter: CASE WHEN (current_setting('api.tenant_id'::text, true) IS NULL) THEN false ELSE (tenant_id = (current_setting('api.tenant_id'::text))::uuid) END
                                      Rows Removed by Filter: 509
```

image

josema-xyz (approved):
Nice catch, on live!

vicferpoy (approved):

PR #9054, opened 2025-10-29
fix(api): update unique constraint for Provider model to exclude soft…

rakan-pro:
### Context
This change addresses an issue where attempting to soft-delete a provider that was already soft-deleted would result in a duplicate error.

### Description

This change modifies the unique constraint on the `Provider` model to ensure that it only considers non-deleted entries. 
A new migration has been created to reflect this change in the database schema. 
Additionally, tests have been added to verify the behavior of soft deletion and the uniqueness of provider entries.

### Steps to review
1.  **Manual Verification (if possible):** If there's an easy way to spin up the API locally, consider:
    *   Creating a provider.
    *   Soft-deleting that provider.
    *   Attempting to create a new provider with the same unique identifier (`tenant_id`, `provider`, `uid`). This should succeed.
    *   Attempting to create a new provider with the same unique identifier when the original is NOT soft-deleted. This should fail.
    * Soft-deleting that provider. should not fail (it was failing before due to the `UniqueConstraint`) 

### Checklist

#### API

### License


josema-xyz (dismissed):
LGTM! 🚀

josema-xyz (dismissed):
Love the tests!

AdriiiPRodri (dismissed):

josema-xyz (dismissed):

AdriiiPRodri (approved):

josema-xyz (approved):

PR #9064, opened 2025-10-29
feat(rls-transaction): add retry for read replica connections

vicferpoy:
### Context

Read replicas occasionally scale down, causing immediate connection termination and resulting in `OperationalError` failures in Celery tasks that read from replicas. These failures were causing task execution to fail without recovery, requiring manual intervention.

### Description

Implemented automatic retry logic at the database transaction level to handle transient connection failures when Aurora replicas scale down.

  Changes Made

  1. Enhanced `rls_transaction` with Retry Logic (`src/backend/api/db_utils.py`)

  Added exponential backoff retry mechanism to the `rls_transaction()` context manager:

```python
  @contextmanager
  def rls_transaction(tenant_id: str, using: str = "default"):
      """
      Context manager for RLS-protected database transactions with automatic retry.

      Retries on OperationalError (connection failures) with exponential backoff:
      - Max retries: 3
      - Base delay: 0.5s
      - Backoff multiplier: 2x
      - Jitter: ±25% to prevent thundering herd
      """
```

  Configuration:
  - Max retries: 3 attempts
  - Base delay: 0.5 seconds
  - Backoff multiplier: 2x (0.5s → 1s → 2s)
  - Jitter: ±25% random variance to prevent thundering herd
  - Retry trigger: OperationalError (database connection failures)

  2. Comprehensive Test Coverage

  Unit Tests (`src/backend/api/tests/test_db_utils.py`):
  - Validates retry behavior on connection failures
  - Verifies exponential backoff timing
  - Tests retry limit enforcement
  - Confirms successful operations don't trigger retries

  Integration Tests (`src/backend/api/tests/integration/test_rls_transaction.py`):
  - Tests retry logic with real database operations
  - Validates tenant isolation maintained across retries
  - Verifies read replica connection retry scenarios

> [!WARNING]
> This implementation doesn't prevent connection errors during queries after `yield`. Long running transactions will be prone to similar issues but we cannot control that given the nature of the context manager for now.

### Steps to review

### Checklist

#### API

### License


josema-xyz (commented):
Nice, but don't we have a retry external or internal module instead of implementing it for this case?

josema-xyz (dismissed):
Nice retry logic! 🚀

api/src/backend/api/db_utils.py
```diff
@@ -71,24 +87,51 @@ def rls_transaction(
...
+                    cursor.execute(SET_CONFIG_QUERY, [parameter, value])
+                    yield cursor
+            return
+        except OperationalError as e:
+            # If on primary or max attempts reached, raise
+            if not is_replica or attempt == max_attempts:
+                raise
+
+            # Retry with exponential backoff
+            delay = 0.5 * (2 ** (attempt - 1))
```
AdriiiPRodri: This is only a suggestion, imo we need to move these values to the line 93 in order to have the configuration in one place, apart on that I thing we can add a comment explaining why we need to apply the delay and why the 0.5 and 2 value

AdriiiPRodri (changes requested):

josema-xyz (dismissed):
nowyes

api/src/backend/api/db_utils.py
vicferpoy: Solved.

cesararroba (approved):

PR #9080, opened 2025-10-29
fix(api): update database routing logic in MainRouter

AdriiiPRodri:
### Context

Security Hub exports were failing in production with Cannot assign "<Integration …>" and invalid input syntax for type uuid: "". The first came from reading the integration-provider relationship through the read replica; the second was Row Level Security blocking our integration.save() calls because the tenant context was only being set on the replica connection.

### Description

- Allow allow_relation to treat default, admin, and replica as compatible aliases so Django stops rejecting relationships built from replica reads.

- Ensure every Security Hub write (regions update, connection flag) runs inside rls_transaction(..., using=MainRouter.default_db) before saving, so SET CONFIG(api.tenant_id, …) hits the same primary connection that persists the row.

### Steps to review

1. Pull the branch and reproduce a Security Hub export against a tenant with the integration enabled.
2. Confirm logs show the integration completing without Cannot assign or invalid input syntax errors.
3. Verify the integration record keeps the new regions map and connected status as expected.

### Checklist

#### API

### License


AdriiiPRodri:
## Status update - Explanation

When the task blew up with `invalid input syntax for type uuid: ""`, it wasn’t a bad value in our code, it was PostgreSQL’s row-level security telling us we’d lost track of the tenant context. Here’s what was going on:

We iterate findings inside with `rls_transaction(tenant_id, using=READ_REPLICA_ALIAS)`. That context only touches the replica connection, so `SET_CONFIG('api.tenant_id', ...)` is executed on the replica pool. We do that on purpose to offload read traffic.

The moment we call `integration.save()`, Django’s router switches to the write alias (the primary). That connection never saw the `SET_CONFIG`, so `current_setting('api.tenant_id')` is empty and Postgres rejects the update when the RLS policy checks `tenant_id = current_setting('api.tenant_id')::uuid`.

Forcing the RLS transaction with `using=MainRouter.default_db` before each write (saving the regions, toggling connected, etc.) makes sure set_config happens on the same connection that performs the UPDATE. The RLS policy then sees the right tenant and allows the write.

The other piece was in `api/db_router.py`. Before the change, allow_relation only blessed {default, admin}. When we read `IntegrationProviderRelationship` from the replica but had the Integration instance from the primary, the router blocked the relationship and Django raised `Cannot assign "<Integration ...>"`. Adding `self.replica_db` to the allowed set tells Django, "yes, these aliases all point at the same data, it’s fine to relate objects fetched from default/admin/replica even though only default/admin accept writes." That unblocked the lookup so the Security Hub code could actually run.

So the full fix is:

Let the router consider replica a safe partner for relations (reads still stay readonly, writes still go through the primary).
Whenever we mutate the Integration row—like refreshing the regions map or flipping connected, wrap the call in `rls_transaction(tenant_id, using=MainRouter.default_db)` so we reapply the tenant context on the primary connection before saving.
With those two tweaks the job no longer trips over the router, and RLS works fine

josema-xyz (dismissed):
After reading some Django docs... 🚀

api/src/backend/tasks/jobs/integrations.py
```diff
@@ -208,7 +208,7 @@ def get_security_hub_client_from_integration(
             regions_status[region] = region in connection.enabled_regions
 
         # Save regions information in the integration configuration
-        with rls_transaction(tenant_id):
+        with rls_transaction(tenant_id, using=MainRouter.default_db):
```
vicferpoy: Why is this needed?

vicferpoy (dismissed):

vicferpoy (approved):

PR #9101, opened 2025-10-30
feat(aws): CloudTrail timeline for findings

MrCloudSec:
### Description

CloudTrail timeline feature that allows users to see the history of actions performed on AWS resources. When viewing a security finding, users can now see who created, modified, or deleted the resource and when.
The feature queries AWS CloudTrail Event History to retrieve events related to a specific resource, providing visibility into:
- Who performed actions (IAM users, roles, AWS services)
- What actions were taken (CreateInstance, ModifySecurityGroup, etc.)
- When the actions occurred
- From where (source IP, user agent)

**SDK Usage:**
```python
from prowler.providers.aws.lib.cloudtrail_timeline import CloudTrailTimeline
import boto3
# Initialize with a boto3 session
session = boto3.Session(
    region_name="eu-west-1",
    profile_name="production"  # or use IAM role, env vars, etc.
)
# Create timeline client
timeline = CloudTrailTimeline(
    session=session,
    lookback_days=15  # Default: 90, Max: 90 (CloudTrail Event History limit)
)
# Get timeline for a resource
events = timeline.get_resource_timeline(
    resource_id="public",  # Resource name or ID
    resource_arn="arn:aws:elasticloadbalancing:eu-west-1:123456789012:loadbalancer/app/public/abc123",
    region="eu-west-1",
    page_size=10  # Default: 10, Max: 50
)
```

Event structure:
```json
{
    "event_time": datetime(2026, 1, 5, 15, 33, 59, tzinfo=timezone.utc),
    "event_name": "ModifyListener",
    "event_source": "elasticloadbalancing.amazonaws.com",
    "actor": "AWSReservedSSO_AdministratorAccess_ae4271542a0a62e5",
    "actor_type": "AssumedRole",
    "source_ip_address": "81.0.42.211",
    "user_agent": "APN/1.0 HashiCorp/1.0 Terraform/1.6.0 ...",
    "request_parameters": {
        "listenerArn": "arn:aws:elasticloadbalancing:...",
        "port": 443,
        "protocol": "HTTPS",
        # ... full CloudTrail request params
    },
    "response_elements": {
        "listeners": [...]
        # ... full CloudTrail response
    },
    "error_code": None,      # e.g., "AccessDenied" if action failed
    "error_message": None    # e.g., "User is not authorized..."
}
```

**Required IAM Permissions**
```json
{
    Version: 2012-10-17,
    Statement: [
        {
            Effect: Allow,
            Action: cloudtrail:LookupEvents,
            Resource: *
        }
    ]
}
```

### Checklist

#### API

### License


prowler/lib/check/check.py
```diff
@@ -383,6 +383,105 @@ def run_fixer(check_findings: list) -> int:
         )
 
 
+def _enrich_findings_post_scan(findings: list, output_options: Any) -> dict:
```
jfagoagas: Why this name? Is this included in a class?

prowler/lib/scan/scan.py
```diff
@@ -2,6 +2,8 @@
 from types import SimpleNamespace
 from typing import Generator
 
+from alive_progress import alive_bar
```
jfagoagas: We can't add terminal UI things to the SDK. This is a CLI responsibility.

prowler/lib/scan/scan.py
```diff
@@ -41,6 +43,9 @@ class Scan:
     _status: list[str] = None
     _bulk_checks_metadata: dict[str, CheckMetadata]
     _bulk_compliance_frameworks: dict
+    _enrich_findings: bool = False
+    _enrichment_config: dict = None
+    _enricher = None
```
jfagoagas: Not sure about adding this to the `Scan` SDK. We need to think about it to have it as something external, executed afterwards.

We need to talk about the whole logic in this class. We need to move it somewhere else.

prowler/providers/aws/services/cloudtrail/lib/enrichment/models.py
```diff
@@ -0,0 +1,295 @@
...
+
+class TimelineEventType(str, Enum):
+    """Types of timeline events from CloudTrail."""
+
+    # General resource events
+    RESOURCE_CREATED = "resource_created"
+    RESOURCE_MODIFIED = "resource_modified"
+    RESOURCE_DELETED = "resource_deleted"
+
+    # EC2 Instance events
```
jfagoagas: Why not to add classes for each service?

prowler/providers/aws/services/cloudtrail/cloudtrail_service.py
```diff
@@ -207,6 +208,59 @@ def _lookup_events(self, trail, event_name, minutes):
                 f"{error.__class__.__name__}[{error.__traceback__.tb_lineno}]: {error}"
             )
 
+    def lookup_events_for_resource(
```
jfagoagas: If we move the logic out of the Scan, maybe we'll need to create a new class for this part. We can't re-init the CloudTrail class. To me this would fit best in the enricher class.

prowler/providers/common/models.py
```diff
@@ -26,6 +26,9 @@ class ProviderOutputOptions:
     output_filename: str
     only_logs: bool
     unix_timestamp: bool
+    enrich_findings: bool
```
jfagoagas: We need to find a way not to add more data to this object. The idea is to get rid of it as it makes some things dependant in between.

jfagoagas (changes requested):

prowler/CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ All notable changes to the **Prowler SDK** are documented in this file.
 - `codepipeline_project_repo_private` check for AWS provider [(#5915)](https://github.com/prowler-cloud/prowler/pull/5915)
 - `cloudstorage_bucket_versioning_enabled` check for GCP provider [(#9014)](https://github.com/prowler-cloud/prowler/pull/9014)
 - `cloudstorage_bucket_soft_delete_enabled` check for GCP provider [(#9028)](https://github.com/prowler-cloud/prowler/pull/9028)
+- CloudTrail enrichment for AWS provider [(#9101)](https://github.com/prowler-cloud/prowler/pull/9101)
```
jfagoagas: I'd not add this until the feature is implemented as for now is only an SDK thing.

prowler/providers/aws/services/cloudtrail/lib/__init__.py
puchy22, line 1: Is this the best place to put the code? You are not really using nothing for the service and it is not expected to be used by checks or services, for me it should be at provider level like other integrations like S3 or SecurityHub in `prowler/providers/aws/lib`

prowler/providers/aws/services/cloudtrail/lib/enrichment/cloudtrail_enricher.py
```diff
@@ -0,0 +1,1269 @@
...
+                event_details=event_details,
+            )
+
+        except Exception as error:
+            logger.error(
+                f"{error.__class__.__name__}[{error.__traceback__.tb_lineno}]: {error}"
+            )
+            return None
+
+    def _create_enrichment(
```
puchy22: I believe that this method is never used

prowler/CHANGELOG.md
MrCloudSec:
```suggestion
```

MrCloudSec: Done!

tests/providers/aws/services/cloudtrail/lib/enrichment/test_formatters.py
```diff
@@ -0,0 +1,1665 @@
...
+                "keyId": "1234abcd-12ab-34cd-56ef-1234567890ab",
+                "pendingWindowInDays": 30,
+            }
+        }
+
+        message = EventMessageFormatter.format_kms_key_deletion_scheduled(event)
+
+        assert "1234abcd-12ab-34cd-56ef-1234567890ab" in message
+        assert "30" in message
+        assert "⚠️" in message or "DELETION SCHEDULED" in message
```
puchy22: I am not sure if we should add this kind of emojis in the tests

prowler/providers/aws/lib/cloudtrail_enrichment/cloudtrail_enricher.py
```diff
@@ -0,0 +1,1269 @@
+"""CloudTrail enrichment service for Prowler findings.
```
puchy22: I am concerned about the sustainability of this file. I think it should be split into different files for each service and we should try to use another more sustainable way of adding new events that does not involve modifying a file with 1200 lines of code

prowler/providers/aws/lib/cloudtrail_timeline/formatters.py
```diff
@@ -0,0 +1,1882 @@
+"""Event message formatters for CloudTrail events.
```
puchy22: Same as with cloudtrail_enricher.py, it is not sustainable having a class that is 1800 lines of code. It's so difficult to know what is the class exactly doing. Is there any way to divide it into other parts to make it more sustainable?

prowler/providers/aws/services/cloudtrail/lib/enrichment/models.py
```diff
@@ -0,0 +1,427 @@
...
+class EventBridgeEventType(str, Enum):
+    """EventBridge rule events."""
+
+    RULE_CREATED = "eventbridge_rule_created"
+    RULE_DELETED = "eventbridge_rule_deleted"
+    RULE_UPDATED = "eventbridge_rule_updated"
+
+
+# Union type for all event types
+TimelineEventType = (
```
puchy22: Why are generating this union at module level and then only using it in the type hint of the TimelineEvent class. 
Besides, this won't be a type as such, but rather a union. It seems like a strange way to do it, especially since it will only be used in one place.

prowler/providers/aws/services/cloudtrail/lib/enrichment/models.py
```diff
@@ -0,0 +1,427 @@
...
+    | ECREventType
+    | ECSEventType
+    | EKSEventType
+    | BackupEventType
+    | ACMEventType
+    | EventBridgeEventType
+)
+
+
+@dataclass
```
puchy22: Why using dataclass instead of using pydantic BaseModel?

prowler/providers/aws/services/cloudtrail/lib/enrichment/models.py
```diff
@@ -0,0 +1,427 @@
...
+    """
+
+    timeline: list[TimelineEvent] = field(default_factory=list)
+    created_by: str | None = None
+    created_at: datetime | None = None
+    last_modified_by: str | None = None
+    last_modified_at: datetime | None = None
+    related_resources: list[dict[str, str]] = field(default_factory=list)
+
+    def to_dict(self) -> dict[str, Any]:
```
puchy22: Using `BaseModel` saves you this method: https://docs.pydantic.dev/1.10/usage/exporting_models/#modeldict

prowler/providers/aws/services/cloudtrail/lib/enrichment/models.py
```diff
@@ -0,0 +1,427 @@
...
+            ),
+            "related_resources": self.related_resources,
+        }
+
+    def get_age_days(self) -> int | None:
+        """Get the age of the resource in days since creation."""
+        if not self.created_at:
+            return None
+
+        from datetime import timezone
```
puchy22: Why this here and not in the top of the file?

prowler/providers/aws/services/cloudtrail/lib/enrichment/models.py
```diff
@@ -0,0 +1,427 @@
...
+
+    def get_exposure_duration_days(self) -> int | None:
+        """Get how many days the resource has been in its current state.
+
+        This is useful for determining how long a misconfiguration has existed.
+        """
+        if not self.last_modified_at:
+            return None
+
+        from datetime import timezone
```
puchy22: Same here

puchy22 (changes requested):
As general comment the current AWS-specific design would require complete duplication for Azure (Activity Log) and GCP (Audit Logs). I've recommended an abstraction layer with provider-agnostic base classes

prowler/providers/aws/services/cloudtrail/lib/enrichment/models.py
MrCloudSec: Done!

prowler/providers/aws/services/cloudtrail/lib/enrichment/models.py
MrCloudSec: Done, I did it inside the function because we only need to import it if we use this function.

prowler/providers/aws/services/cloudtrail/lib/enrichment/models.py
MrCloudSec: Done, good catch!

prowler/providers/aws/services/cloudtrail/lib/enrichment/models.py
MrCloudSec: Agree, good catch.

prowler/providers/aws/services/cloudtrail/lib/enrichment/models.py
MrCloudSec: 100%, changed!

tests/providers/aws/services/cloudtrail/lib/enrichment/test_formatters.py
MrCloudSec: Removed!

prowler/providers/aws/services/cloudtrail/lib/enrichment/cloudtrail_enricher.py
MrCloudSec: Missed this one when refactoring.

api/src/backend/api/v1/views.py
```diff
@@ -2460,6 +2462,115 @@ def metadata_latest(self, request):
...
+                {"error": "lookback_days must be a valid integer"},
+                status=status.HTTP_400_BAD_REQUEST,
+            )
+
+        try:
+            # Import CloudTrailTimeline here to avoid loading it globally
+            from api.utils import initialize_prowler_provider
+            from prowler.providers.aws.lib.cloudtrail_timeline.cloudtrail_timeline import (
+                CloudTrailTimeline,
+            )
```
AdriiiPRodri: Why are we importing this here? I mean, this code runs on every call to this endpoint. This should be at the top of the file

AdriiiPRodri (changes requested):
You need to add dedicated timeline serializers and hook them into the viewset so the timeline action follows our current response schema, because right now the endpoint isn’t aligned with it.

You also need to clean up the timeline action: validate filter[lookback_days], use the existing provider session initialization, remove the ad-hoc JSON body handling, and return errors through the usual DRF validation flow. Make sure the endpoint is documented with extend_schema, and if there are constants involved, they should live in the view as class-level attributes. Remove the global query-parameter override since it’s no longer needed.

Finally, add the required unittests: provider validation, lookback validation, and the happy path using mocked AWS components

api/src/backend/api/v1/views.py
```diff
@@ -2460,6 +2462,116 @@ def metadata_latest(self, request):
         serializer.is_valid(raise_exception=True)
         return Response(serializer.data)
 
+    @action(
+        detail=True,
+        methods=["get"],
+        url_name="timeline",
+        url_path=r"timeline(?:/(?P<lookback_days>[0-9]+))?",
+    )
+    def timeline(self, request, pk=None, lookback_days=None):
```
jfagoagas: @MrCloudSec based on your tests, how long can this take?

@AdriiiPRodri @vicferpoy @josemazo @rakan-pro I'm not sure about having a synchronous endpoint to retrieve this data. From a design standpoint I think we'd have an async endpoint to enqueue a task to fetch that data. Then, another endpoint to get the status/result. What do you think?

AdriiiPRodri: I don't currently know how long it takes us to retrieve that data, but if it can block, then I completely agree with you, this should be a task, and we should have an endpoint to fetch the results

MrCloudSec: It takes one second or less to retrieve the timelin for each resource.

jfagoagas: Even if the timeline is more than 90 days? Since this depends on CloudTrail and the user's events, we need to make this async.

MrCloudSec: In the API side or in the SDK?

prowler/providers/aws/services/cloudtrail/lib/__init__.py
puchy22: @jfagoagas is this file needed?

prowler/providers/aws/lib/cloudtrail_timeline/cloudtrail_timeline.py
```diff
@@ -0,0 +1,221 @@
...
+                    if self._is_read_only_event(event_name):
+                        continue
+
+                parsed = self._parse_event(raw_event)
+                if parsed:
+                    events.append(parsed)
+
+            logger.debug(
+                f"CloudTrail timeline: found {len(events)} events for {resource_identifier}"
+            )
```
puchy22: Is this intended?

jfagoagas: I can clean this up, good one. It was used for testing.

prowler/providers/aws/services/cloudtrail/lib/__init__.py
jfagoagas: No, that's unnecessary.

api/src/backend/api/v1/views.py
```diff
@@ -2828,6 +2847,212 @@ def metadata_latest(self, request):
...
+            serializer = ResourceEventSerializer(
+                events,
+                many=True,
+                context={
+                    "count": len(events),
+                    "lookback_days": lookback_days,
+                    "include_read_events": include_read_events,
+                    "resource_uid": resource.uid,
+                },
+            )
```
AdriiiPRodri: This should be a task, sync calls to the sdk may block the api

api/src/backend/api/v1/serializers.py
```diff
@@ -3975,3 +3978,42 @@ def get_id(self, obj):
         if getattr(obj, "_aggregated", False):
             return "n/a"
         return str(obj.id)
+
+
+# Resource Events Serializers
+
+
+class ResourceEventSerializer(drf_serializers.Serializer):
```
AdriiiPRodri: This breaks the json:api schema and our rls, we should use RLSSerializer

api/src/backend/api/v1/views.py
```diff
@@ -2828,6 +2847,212 @@ def metadata_latest(self, request):
...
+            raise UpstreamAuthenticationError()
+        except ClientError as e:
+            error_code = e.response.get("Error", {}).get("Code", "")
+            logger.error(
+                f"Provider API error retrieving events: {str(e)}",
+                exc_info=True,
+            )
+            # AccessDenied means the credentials don't have the required permissions
+            if error_code in ("AccessDenied", "AccessDeniedException"):
+                raise UpstreamAccessDeniedError()
```
AdriiiPRodri, line 3058: This is too generic if the user send an invalid parameter he will obtain a 503 error instead of 400

api/src/backend/api/v1/views.py
```diff
@@ -2828,6 +2847,212 @@ def metadata_latest(self, request):
...
+        methods=["get"],
+        url_name="events",
+        filter_backends=[],  # Disable filters - we're calling external API, not filtering queryset
+    )
+    def events(self, request, pk=None):
+        """Get events for a resource."""
+        resource = self.get_object()
+
+        # Validate provider
+        if resource.provider.provider != "aws":
```
AdriiiPRodri: This is literal

AdriiiPRodri (changes requested):

api/src/backend/api/v1/views.py
AdriiiPRodri: We will do this later

AdriiiPRodri (approved):

puchy22 (approved):
LGTM 👍

andoniaf (approved):
Awesome! 👏

jfagoagas (approved):

PR #9148, opened 2025-11-04
feat(threatscore): implement ThreatScoreSnapshot model, filter, serializer, and view for ThreatScore metrics retrieval

AdriiiPRodri:
### Context
- Extend ThreatScore overview so the /api/v1/overviews/threatscore action can serve a tenant-wide aggregated snapshot (weighted averages across providers) while still allowing consumers to fetch the latest provider or scan snapshot.
- Persist ThreatScore metrics after every report generation and add a backfill command that can optionally seed realistic-looking data for UI testing.

### Description
- Added ThreatScoreSnapshotSerializer, filter wiring, and view logic that normalizes filter[...] params, aggregates through SQL annotations, and returns a synthetic snapshot with id = "n/a" plus deterministic section/critical requirement merging.
- Extracted ThreatScore helpers into tasks/jobs/threatscore.py and threatscore_utils.py, updated the report job to log once and create snapshots, and introduced a fake-data mode for the population command (with compliance-driven sections and optional seed).
- Created migration/model (ThreatScoreSnapshot), added tests/adjustments, and reconciled master’s mute-rule serializers with the new ThreatScore serializer.

### Steps to review

### Checklist

#### API

### License


api/src/backend/api/v1/views.py
```diff
@@ -4011,6 +4033,329 @@ def services(self, request):
...
+        Get ThreatScore snapshots.
+
+        Default behavior: Returns the latest snapshot for each provider.
+        With snapshot_id: Returns the specific snapshot requested.
+        """
+        tenant_id = self.request.tenant_id
+        snapshot_id = request.query_params.get("snapshot_id")
+
+        # Base queryset with RLS
+        base_queryset = ThreatScoreSnapshot.objects.filter(tenant_id=tenant_id)
```
josema-xyz, line 4088: What don't you use here the `rls_transaction` context manager?

josema-xyz (changes requested):
Just a little question while I continue reviewing

api/src/backend/api/v1/views.py
AdriiiPRodri: We're building the queryset here, not executing it. Since no data is being fetched or modified at this point, there's no need to wrap it in an `rls_transaction` context

api/src/backend/api/models.py
```diff
@@ -2228,3 +2228,137 @@ class Meta(RowLevelSecurityProtectedModel.Meta):
...
         resource_name = "lighthouse-models"
+
+
+class ThreatScoreSnapshot(RowLevelSecurityProtectedModel):
+    """
+    Stores historical ThreatScore metrics for a given scan.
+    Snapshots are created automatically after each ThreatScore report generation.
+    """
+
+    objects = models.Manager()
```
rakan-pro, line 2250: I think that this is redundant here, since we don't have any custom manager for this model.

AdriiiPRodri: Thanks for flagging this. In this model we expose two managers: objects, which enforces the tenant-scoped RLS filters, and `all_objects`, which we use internally to bypass RLS when needed (background jobs, management commands, etc.). If we drop the explicit `objects = models.Manager()`, Django will make `all_objects` the default manager, so `ThreatScoreSnapshot.objects` disappears. In practice the API relies on `ThreatScoreSnapshot.objects.filter(...)`, and we already saw it blowing up with `AttributeError` as soon as we removed the line. That's why we keep the explicit assignment even though it's the vanilla manager

```
api-dev-1      |   File "/home/prowler/backend/api/v1/views.py", line 4088, in threatscore
api-dev-1      |     base_queryset = ThreatScoreSnapshot.objects.filter(tenant_id=tenant_id)
api-dev-1      |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^
api-dev-1      | AttributeError: type object 'ThreatScoreSnapshot' has no attribute 'objects'
```

josema-xyz (dismissed):
Amazing job with the threatscore snapshots! 👏

josema-xyz (dismissed):

josema-xyz (approved):

PR #9163, opened 2025-11-04
docs: get latest version to install Prowler App

jfagoagas:
### Description

Get Prowler App latest version `docker-compose` and `.env`.

Alternatively, we can use
```
curl -s https://api.github.com/repos/prowler-cloud/prowler/releases/latest | grep '"tag_name":' | sed -E 's/.*"([^"]+)".*/\1/'
```

The rest of changes comes from Mintlify formatter.


StylusFrost (approved):
Working curl sentences! 👍

PR #9165, opened 2025-11-04
fix: return human readable error messages from lighthouse celery tasks

0xbadshah:
### Context

If user adds an invalid API key, the entire JSON error message is returned. 

Celery task output:

image

Rendering the same error message on frontend looks as follows:

image

(As highlighted in the PR review - https://github.com/prowler-cloud/prowler/pull/8925#issuecomment-3486236854)

### Description

This PR returns just the human readable error message if the message exists in the Exception.

### Steps to review

1. Create an OpenAI provider with incorrect API key (sk-xxxxxxx) using the new lighthouse endpoint `/api/v1/lighthouse/providers`
2. The celery task should return just the error message instead of the entire JSON

Example:

image

### Checklist

#### API

### License


0xbadshah:
No changelog needed as multi-llm provider support is not rolled out in any release.

rakan-pro (approved):
LGTM ! note that change log should be adjusted accordingly.

api/src/backend/tasks/jobs/lighthouse_providers.py
```diff
@@ -10,6 +10,39 @@
...
+        str: A user-friendly error message.
+    """
+    # For OpenAI SDK errors (>= v1.0)
+    # OpenAI exceptions have a 'body' attribute with error details
+    if hasattr(e, "body") and isinstance(e.body, dict):
+        if "message" in e.body:
+            return e.body["message"]
+        # Sometimes nested under 'error' key
+        if "error" in e.body and isinstance(e.body["error"], dict):
+            return e.body["error"].get("message", str(e))
```
AdriiiPRodri, line 35: is it possible that the openai error sometimes appears in the message and other times in the error?

api/src/backend/tasks/jobs/lighthouse_providers.py
```diff
@@ -10,6 +10,39 @@
...
+
+    # For boto3 ClientError
+    # Boto3 exceptions have a 'response' attribute with error details
+    if hasattr(e, "response") and isinstance(e.response, dict):
+        error_info = e.response.get("Error", {})
+        if error_info.get("Message"):
+            return error_info["Message"]
```
AdriiiPRodri, line 42: I don't understand this boto3 part inside a function that should handle openai errors, can you provide more information? Thanks

AdriiiPRodri (changes requested):

api/src/backend/tasks/jobs/lighthouse_providers.py
0xbadshah: @AdriiiPRodri Correct. OpenAI doesn't always send the error under error.message. Sometimes it does send in .message. For reference - https://github.com/openai/openai-python/issues/2487

api/src/backend/tasks/jobs/lighthouse_providers.py
0xbadshah: We support multi LLM (including OpenAI and Amazon Bedrock). This function is a wrapper function to handle both. `_extract_error_message` is used by both OpenAI and Amazon Bedrock and the errors are extracted from the response.

AdriiiPRodri (approved):

PR #9213, opened 2025-11-11
fix(aws): false negative in `iam_role_cross_service_confused_deputy_prevention`

snotrauk:
### Context

This PR fixes a false negative in the iam_role_cross_service_confused_deputy_prevention check by updating the "is_policy_public" function in prowler.providers.aws.services.iam.lib.policy.

The issue exists because the current code assumes the service obtained by principal.get("Service", []) is a string. but when the role trusts multiple service principals like the example below, this will be a list and the substring checks will not work and therefore "pass" the role when it could actually be vulnerable.

```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Principal": {
                "Service": [
                    "scheduler.amazonaws.com",
                    "events.amazonaws.com"
                ]
            },
            "Action": "sts:AssumeRole"
        }
    ]
}
```

This normalises the service to a list and checks all entries, to ensure all service principals are properly checked. 

I have tested this against two AWS accounts i have access to.

### License


jfagoagas:
@snotrauk could you please add a new test to validate the fix? 

Thanks for contributing to Prowler 🙌

snotrauk:
> @snotrauk could you please add a new test to validate the fix?
> 
> Thanks for contributing to Prowler 🙌

done... (i think)

MrCloudSec (dismissed):
Good catch @snotrauk! Thanks for your contrib.

MrCloudSec (approved):

PR #9216, opened 2025-11-11
fix(aws): Python mutable default argument in AWS EC2 Security Group lib

leetrout:
### Context

This PR fixes up a subtle Python bug when using an object passed by reference as the default function argument and consolidates the logic for a few boolean return values.

### Steps to review

Nothing special needed.

### Checklist

#### UI

#### API

### License


prowler/providers/aws/services/ec2/lib/security_groups.py
```diff
@@ -71,37 +74,29 @@ def check_security_group(
...
-                            and ingress_rule["IpProtocol"] == protocol
-                        ):
-                            return True
+                for port in ports:
+                    if (
+                        port in ingress_port_range
+                        and ingress_rule["IpProtocol"] == protocol
+                    ):
+                        return True
+
```
MrCloudSec:
```suggestion
```

prowler/providers/aws/services/ec2/lib/security_groups.py
```diff
@@ -71,37 +74,28 @@ def check_security_group(
...
                 # If empty input ports check if all ports are open
-                if len(set(ingress_port_range)) == 65536:
-                    return True
-                # If None input ports check if any port is open
-                if ports is None:
-                    return True
```
MrCloudSec, line 86: Why did you delete these lines?

prowler/providers/aws/services/ec2/lib/security_groups.py
```diff
@@ -71,37 +74,28 @@ def check_security_group(
...
+                        port in ingress_port_range
+                        and ingress_rule["IpProtocol"] == protocol
+                    ):
+                        return True
                 # If empty input ports check if all ports are open
-                if len(set(ingress_port_range)) == 65536:
-                    return True
-                # If None input ports check if any port is open
-                if ports is None:
-                    return True
```
MrCloudSec, line 104: Same here.

leetrout: It was (and still is) unreachable code. Ports was previously a default list.

prowler/providers/aws/services/ec2/lib/security_groups.py
leetrout: It was (and still is) unreachable code. Ports was previously a default list.

MrCloudSec (dismissed):

MrCloudSec:
@leetrout these changes are introducing a false negative in `tests/providers/aws/services/ec2/ec2_securitygroup_allow_ingress_from_internet_to_any_port/ec2_securitygroup_allow_ingress_from_internet_to_any_port_test.py::Test_ec2_securitygroup_allow_ingress_from_internet_to_any_port::test_ec2_non_compliant_default_sg_open_to_one_port`

leetrout:
> @leetrout these changes are introducing a false negative in `tests/providers/aws/services/ec2/ec2_securitygroup_allow_ingress_from_internet_to_any_port/ec2_securitygroup_allow_ingress_from_internet_to_any_port_test.py::Test_ec2_securitygroup_allow_ingress_from_internet_to_any_port::test_ec2_non_compliant_default_sg_open_to_one_port`

Thanks, I'll go look 👍

leetrout:
At a glance it looks like it was relying on the implicit return of `None` to satisfy falsiness. Will push a fix in a bit.

prowler/providers/aws/services/ec2/lib/security_groups.py
```diff
@@ -71,37 +76,43 @@ def check_security_group(
...
+                
+                # We did not find a specific port for the given protocol for 
+                # a public cidr so let's see if all the ports are open
+                all_ports_open = len(set(ingress_port_range)) == 65536
+                
+                # At this point we might have all ports open, return True
+                # otherwise we didn't have any ports to check yet we have a public cidr
+                # so that is the same as all ports open for our purposes
+                ports_are_empty = not ports
+                return all_ports_open or ports_are_empty
```
leetrout: @MrCloudSec tests are passing now locally.

pyproject.toml
```diff
@@ -68,7 +68,9 @@ dependencies = [
   "tzlocal==5.3.1",
   "py-iam-expand==0.1.0",
   "h2 (==4.3.0)",
-  "oci==2.160.3"
+  "oci==2.160.3",
+  "pytest>=8.4.2",
+  "moto>=5.1.16",
```
MrCloudSec:
```suggestion
  "oci==2.160.3"
```

leetrout:
I see the failure. This is very curious to me because the original logic seemed a bit loose and I thought this was going to be an easy "let's just tighten it up a bit".

Now I'm invested in understanding how this is working inside these checks and what assumptions were made by both statements "open to the internet" and "all ports".

leetrout:
The ingress rule of the **currently failed test**:

`test_ec2_non_compliant_default_sg_open_to_one_port`

and ingress rule of **previously failed test**:

`test_ec2_compliant_default_sg_only_open_to_one_port`

are both the same:

```python
{
  "IpProtocol":"tcp",
  "FromPort":80,
  "ToPort":80,
  "UserIdGroupPairs":[],
  "IpRanges":[
    {
      "CidrIp":"0.0.0.0/0"
    }
  ],
  "Ipv6Ranges":[],
  "PrefixListIds":[]
}
```

I was assuming I covered the edge case of the false return value associated with none but I guess the additional guards were necessary to let it fall completely through. I'll send and update after I wrap my head around why the test expectations were different for the same input.

leetrout:
OK, I got it. If `None` was explicitly passed in the code was not unreachable but that was violating the docstring and the type annotation.

I will make a larger change just to see it through and then y'all can decide if you want to adopt the changes or not.

HugoPBrito (approved):
After reviewing this PR, I found that the original description understates the scope of changes. Here's a complete breakdown:

  1. Mutable default argument fix (ports: list = [] → ports: list | None = None)
    - While this is a Python anti-pattern flagged by linters (pylint W0102, ruff B006), the bug wouldn't actually manifest here since ports is never modified in the function. Still, it's a best practice fix.
  2. New all_ports parameter
    - Replaces the confusing ports=None pattern with an explicit all_ports=True flag
    - Before: ports=None meant "match any open port", ports=[] meant "only match if all 65536 ports are open"
    - After: all_ports=True explicitly controls this behavior
  3. Fixed incorrect documentation
    - Old docs incorrectly stated: "If empty, any port will be checked. If None, any port will be checked."
    - The code actually treated [] and None differently
  4. Bug fix in _is_cidr_public (line 119)
    - Added missing return False at the end of the function
    - Previously returned None implicitly when any_address=True and CIDR wasn't 0.0.0.0/0 or ::/0
  5. Code optimization
    - Uses set() and range() instead of manual loop for port range calculation
    - Combines IPv4 and IPv6 checks into a single loop, reducing code duplication
  6. Minor test improvements
    - ex.type == ValueError → ex.type is ValueError (identity check best practice)
    - [str] → list[str] (modern type hint syntax)

  - All 42 tests in security_groups_test.py pass
  - All tests for affected checks pass (DMS, EC2 securitygroup checks, EMR)

Sorry for the delay and thank you very much for your contribution @leetrout! ❤️

PR #9236, opened 2025-11-14
chore(entra): enhance performance for `user_registration_details` and user `mfa` evaluation

HugoPBrito:
### Context

`Azure` scan logs showed the `Entra` service bootstrap issuing thousands of per-user Graph requests, stalling scans.

We now reuse `Microsoft Graph registration summaries` to avoid that bottleneck in both Azure and `M365` Entra services.

We were also using a rather fragile method to check the MFA status in Azure, and the change in how it was retrieved ended up breaking the check. The correct way to do it, instead of counting the authentication methods, is to use the isMfaCapable field.

Logic was updated to use this proper approach, which is also the same method we already use in Microsoft 365.

### Description

- `prowler/providers/azure/services/entra/entra_service.py` now loads user authentication methods via a single `_get_user_registration_details` method, caching methods from the paged `user_registration_details`. `Azure` MFA checks (`entra_privileged_user_has_mfa`, etc.) continue to rely on `user.authentication_methods`, so the method stays in place but uses aggregated data.
- `tests/providers/azure/services/entra/entra_service_test.py` updated to mock the reports endpoint instead of `by_user_id`, asserting the new method population workflow.
- `prowler/providers/m365/services/entra/entra_service.py` mirrors the change by caching `is_mfa_capable` flags from registration details, and `tests/providers/m365/services/entra/microsoft365_entra_service_test.py` now covers both `_get_users` and the new method pagination path.
- Use `isMfaCapable` to check MFA status

> [!IMPORTANT]  
This would incur in a new permission needed for `Azure` provider: `AuditLog.Read.All`

### Steps to review

### Checklist

#### UI

#### API

### License


andoniaf:
## Local test
image

andoniaf (approved):

PR #9272, opened 2025-11-19
fix(outputs): refresh scan timestamps per run

### LINKED ISSUE #8591: Timestamp on generated reports are wrong
### Steps to Reproduce

1. Self host your prowler and deploy every container (don't stop or restart the worker container)
2. Create a cloud provider with scheduled scan
3. Wait few days (without restarting the worker container) for multiple output report to be generated
4. Every report will have their timestamp to the date of container start/first scan (so without container restart, every report will be generated to the same day)

### Expected behavior

Every report should have the `TIMESTAMP` date of the date of generation/scan and not be limited to the date of container start / first scan.

I've found the root cause here:
https://github.com/prowler-cloud/prowler/blob/24364bd73e9db890efcbfa3953afa552922d15d5/prowler/config/config.py#L13-L14

Those definition of `timestamp` and `timestamp_utc` are static. The value will be set during first execution and never updated..
They must be changed to a function to return a dynamic response of the actual current date.

Example of generated report with the wrong date:
Image
In here, my first scan was on Aug 18 but my lasted scan on Aug 27 don't have the correct date and still show the Aug 18 timestamp.

### Environment Resource

Self-hosted docker container

### OS used

Linux

### Prowler version

v5.10.2

### Pip version

N/A

### Context

_No response_

HugoPBrito:
### Context

A user reported an error in timestamps generation.

Fix https://github.com/prowler-cloud/prowler/issues/8591

- CLI findings and compliance outputs embedded the process start time, so long-running containers produced stale timestamps even when the filenames were fresh.
- API exports inherited that stale timestamp value as well, so compliance CSVs and other assets always reported the original deployment time instead of the scan’s started_at.
- This PR reintroduces a mutable timestamp wrapper inside the SDK and refreshes it for every scan, then synchronizes the API exporter to call the same setter so all generated artifacts stay consistent.

### Description

- Prowler SDK (`prowler/config/config.py`, `prowler/providers/common/models.py`)
- Reintroduced the lightweight `_MutableTimestamp` proxy and extended `set_output_timestamp`/`reset_output_timestamp` to keep `timestamp_iso` in sync. It acts as a tiny proxy object that forwards attribute access to the underlying datetime. To keep it drop‑in compatible wherever timestamp was previously used, it defines:
  - __getattr__ so every normal attribute call (`timestamp.strftime`, `timestamp.isoformat`, etc.) transparently hits the wrapped datetime.
  - __str__/__repr__ so anything that prints/logs timestamp sees the same output as before.
  - __eq__ so comparisons like timestamp == other_datetime keep working.
  - A set() helper so the SDK can swap the underlying value without changing every import.

- ProviderOutputOptions now calls `set_output_timestamp(datetime.now())` when instantiated, guaranteeing each CLI run uses a fresh timestamp for findings, HTML, ASFF, etc.
- Prowler API (`api/src/backend/tasks/jobs/export.py`, `tests`)
- `_generate_output_directory` now invokes `set_output_timestamp(started_at)` so compliance/regular exports embed the exact scan start time.
- Updated tests assert that the setter is called for both clean provider names and sanitized ones.

### Steps to review

### Checklist

#### UI

#### API

### License


josema-xyz (approved):
🚀

danibarranqueroo (approved):

PR #9304, opened 2025-11-25
fix(apigateway): retrieve correct `logingLevel` status

bota4go:
### Context

If AWS API Gateway logs were previously enabled and the logging level was set in the stage properties, but then later changed to logging_level = "OFF", Prowler will not detect this in the current logic. This is because it only checks for the presence of the logging property, not its actual value.
Therefore, Prowler needs an additional validation to ensure that logging is truly enabled.

### Description
This PR adds a single additional condition to the existing if logic.

### Steps to review
Review the changes in the modified logic.

### Checklist
Are there new checks included in this PR?
No

### License


HugoPBrito (approved):

HugoPBrito (changes requested):
Changelog needs to be added, and a test case testing the new edge case as well.

HugoPBrito:
Hi @bota4go,

Thanks for this contribution! Feel free to solve the comments.
In case you don't want to I'll handle it myself as soon as possible.

bota4go:
> Hi @bota4go,
> 
> Thanks for this contribution! Feel free to solve the comments. In case you don't want to I'll handle it myself as soon as possible.

Hey, mate. 

1. Changelog added.  Pls have a look. 

2. Do You have any case testing examples handy? I can provide screenshots for API gateway stage with loggingLevel turned off  and  Prowler active alert on that, but I am not sure if it is a right way?

HugoPBrito:
Hi @bota4go,

Sorry for the big delay on the reply.
Yes, the screenshots you're referring to should be enough. In addition to that, we can verify it trough boto3: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/apigateway/client/get_stages.html

Since I didn't reply to you quickly, the changelog is now outdated. I'll update it for you to merge this right away.

Sorry again for the late response.

HugoPBrito (dismissed):

HugoPBrito (approved):

HugoPBrito:
Thank you very much @bota4go for your contribution! 🚀 ❤️

PR #9423, opened 2025-12-03
feat(cloudflare): Add Cloudflare provider with zones service and critical security checks

HugoPBrito:
### Context

This PR introduces the `Cloudflare` provider to Prowler, enabling security assessments for Cloudflare-managed infrastructure. This is part 1 of 4 PRs that will add complete Cloudflare support.

1. This one
2. https://github.com/prowler-cloud/prowler/pull/9424
3. https://github.com/prowler-cloud/prowler/pull/9425
4. https://github.com/prowler-cloud/prowler/pull/9426

  ### Description

  Adds complete Cloudflare provider infrastructure and initial security checks:

  **Provider Infrastructure:**
  - Cloudflare provider with env var API token and API key authentication
  - Zone filtering support via `-f` flag
  - Mutelist support for finding suppression
  - Custom exceptions and error handling
  - Documentation for authentication and getting started

  **Zones Service (5 checks):**

  *TLS/SSL Security:*
  | Check | Description |
  |-------|-------------|
  | `zones_ssl_strict` | Validates SSL mode is set to strict or full_strict |
  | `zones_min_tls_version_secure` | Ensures minimum TLS version is 1.2 or higher |
  | `zones_https_redirect_enabled` | Validates Always Use HTTPS is enabled |
  | `zones_hsts_enabled` | Ensures HSTS is enabled with proper max-age |
  | `zones_dnssec_enabled` | Validates DNSSEC is enabled for DNS security |

  ### Steps to review

  1. Review provider infrastructure in `prowler/providers/cloudflare/`
  2. Review zones service in `prowler/providers/cloudflare/services/zones/`
  3. Review documentation in `docs/user-guide/providers/cloudflare/`
  4. Test locally:
     ```bash
     # With API token
     # With API key
     # Filter specific zones
     prowler cloudflare -f example.com

### Checklist

    - If so, do we need to update permissions for the provider? No - uses standard Cloudflare API token permissions

### License


prowler/config/cloudflare_mutelist_example.yaml
HugoPBrito, line 1: Tested.
Image

prowler/providers/cloudflare/config.py
```diff
@@ -0,0 +1,5 @@
+"""Cloudflare provider configuration defaults."""
```
puchy22: I am not sure if this parameters are configureable parameters at provider level. It looks that in Azure we are using in this way, I believe that this should be a module to configure the provider itself, but here we are using for constants variables for checks, that doesn't look the right place.

I see that Azure is the only other provider where we are using this file for checks directly, I don't think that this should be the way that the code should be organized. Let's think a better place to put this kind of values in other providers.

prowler/providers/cloudflare/models.py
```diff
@@ -0,0 +1,56 @@
+from typing import Any, Optional
+
+from pydantic.v1 import BaseModel, Field
```
puchy22: Any special reason for still using V1 `BaseModel` instead the V2?

prowler/providers/common/provider.py
```diff
@@ -248,6 +248,16 @@ def init_global_provider(arguments: Namespace) -> None:
                         repositories=arguments.repository,
                         organizations=arguments.organization,
                     )
+                elif "cloudflare" in provider_class_name.lower():
```
puchy22, line 251: This is not a comment to fix in this provider, but maybe now that we are adding more providers we should take care of this part of the code. Adding more and more elif statements doesn't seem like the best way to scale the code.

HugoPBrito: That happens in several parts of the codebase. I completely agree with you.

prowler/providers/cloudflare/config.py
HugoPBrito: I moved it to config.yaml like in the other providers.

prowler/providers/cloudflare/models.py
HugoPBrito: Not really. I changed it, thanks.

prowler/providers/cloudflare/config.py
puchy22: I still seeing `CLOUDFLARE_DEFAULT_MIN_TLS`, `CLOUDFLARE_RECOMMENDED_SSL_MODES` and `CLOUDFLARE_RECOMMENDED_SECURITY_LEVELS` defined in this file. That's what I was referring with my comment

HugoPBrito: I did yesterday on https://github.com/prowler-cloud/prowler/pull/9423/commits/e4d1d647c51a806f02500595be74dc52aaa5177a. Please check again, maybe something was cached.

puchy22: Then 2 main concerns:
1. Delete that vars from `config.py` file because they still there and they are unused.
2. I don't think that they should be something configurable by the user.

andoniaf:
Testing locally with python3.9 I got the following error:
`TypeError[154]: unsupported operand type(s) for |: 'types.GenericAlias' and 'NoneType'`
looks like we're using Python 3.10+ union type syntax (X | None) which is incompatible with Python 3.9.

We should start deprecating 3.9 as it's already EOL'd, if we can do it before releasing this shouldn't be a problem. /cc @jfagoagas

andoniaf:
Not sure why Mintlify did not trigger a preview environment, I did it manually (as it's easier to review docs there): https://prowler-prowler-386-add-cloudflare-provider-to-cli.mintlify.app/user-guide/providers/cloudflare/getting-started-cloudflare

prowler/providers/cloudflare/services/zones/zones_dnssec_enabled/zones_dnssec_enabled.metadata.json
```diff
@@ -0,0 +1,35 @@
+{
+  "Provider": "cloudflare",
+  "CheckID": "zones_dnssec_enabled",
+  "CheckTitle": "DNSSEC is enabled to protect DNS responses from tampering",
```
andoniaf: `picky` comment about metadata. Should the tittle be only "DNSSEC is enable", and leave the justification ("to protect DNS responses from tampering") for the description? 

It apply to the other checks. I know it leave them really short, but it matches the pattern from other checks (like: EC2 instance has IMDSv2 enforced)

HugoPBrito: I completely agree. Done.

prowler/providers/cloudflare/config.py
HugoPBrito: Agree. Done.

andoniaf (dismissed):

prowler/providers/cloudflare/services/zones/zones_ssl_strict/zones_ssl_strict.metadata.json
```diff
@@ -0,0 +1,35 @@
+{
+  "Provider": "cloudflare",
+  "CheckID": "zones_ssl_strict",
+  "CheckTitle": "SSL/TLS encryption mode is set to Full Strict",
```
andoniaf: Title says "Full Strict" but code accepts both "strict" AND "full_strict". In the description only says `Full`

HugoPBrito: Done! Only accepting Full (Strict).

andoniaf (approved):

PR #9444, opened 2025-12-03
feat(report): improve the way of reporting and adding reports

pedrooot:
### Context

This PR refactors the PDF compliance report generation system for ThreatScore, ENS RD2022, and NIS2 frameworks. The previous implementation had all report logic in a single large file (`report.py` with ~2800 lines), making it difficult to maintain and extend. Additionally, there were performance issues with memory usage and type conversion bugs.

### Description

**Architecture Refactoring:**
- Extracted report generation into a modular `reports/` package with dedicated modules:
  - `base.py`: Abstract base class implementing Template Method pattern for all reports
  - `config.py`: Centralized configuration (colors, constants, framework configs)
  - `components.py`: Reusable PDF components (tables, badges, info tables)
  - `charts.py`: Chart generation utilities (bar charts, pie charts)
  - `threatscore.py`: ThreatScore-specific report generator
  - `ens.py`: ENS RD2022-specific report generator
  - `nis2.py`: NIS2-specific report generator

**Bug Fixes:**
- Fixed `DJANGO_FINDINGS_BATCH_SIZE` type conversion bug: Changed from `env.str()` to `env.int()` to prevent `'<=' not supported between instances of 'str' and 'int'` errors
- Fixed `LevelOfRisk` and `Weight` type conversion: Added safe `int()` conversion for compliance attributes that may come as strings from different frameworks
- Fixed flake8 E402 errors in `charts.py` by adding `# noqa` comments for imports that must come after `matplotlib.use("Agg")`

**Performance Optimizations:**
- Added memory-efficient findings loading with on-demand queries
- Implemented shared findings cache across multiple reports
- Added `LongTable` support for large datasets (>50 rows)
- Skip alternating row colors for very large tables to reduce memory overhead
- Added garbage collection after heavy sections (charts, findings)

**Code Quality:**
- Added comprehensive docstrings following Google style guide
- Added type hints throughout the codebase
- Created reusable dataclasses (`RequirementData`, `ComplianceData`, `ColumnConfig`, `FrameworkConfig`)
- Improved error logging with full traceback support

### Steps to review

1. Review the new modular architecture in `api/src/backend/tasks/jobs/reports/`
2. Verify the bug fixes in `config/django/base.py` and `reports/threatscore.py`
3. Run the compliance report generation for ThreatScore, ENS, and NIS2 to verify functionality
4. Check memory usage improvements with large scans

### Checklist

#### UI
- N/A (API-only changes)

#### API

### License


api/src/backend/tasks/tests/test_reports_base.py
```diff
@@ -0,0 +1,1177 @@
...
+                return [
+                    Paragraph("Executive Summary", self.styles["h1"]),
+                    Paragraph(
+                        f"Total requirements: {len(data.requirements)}",
+                        self.styles["normal"],
+                    ),
+                ]
+
+            def create_charts_section(self, data):
+                from reportlab.platypus import Image
```
AdriiiPRodri: Let's move this to the top of the file

api/src/backend/tasks/tests/test_reports_threatscore.py
```diff
@@ -0,0 +1,1102 @@
...
+
+
+# Use string status values directly to avoid Django DB initialization
+# These match api.models.StatusChoices values
+class StatusChoices:
+    """Mock StatusChoices to avoid Django DB initialization."""
+
+    PASS = "PASS"
+    FAIL = "FAIL"
+    MANUAL = "MANUAL"
```
AdriiiPRodri: I think we can import this instead of create a new class in the tests

api/src/backend/tasks/jobs/reports/nis2.py
```diff
@@ -0,0 +1,470 @@
...
+        section_scores = defaultdict(lambda: {"passed": 0, "total": 0})
+
+        for req in data.requirements:
+            if req.status == StatusChoices.MANUAL:
+                continue
+
+            req_attrs = data.attributes_by_requirement_id.get(req.id, {})
+            meta = req_attrs.get("attributes", {}).get("req_attributes", [{}])
+            if meta:
+                m = meta[0]
```
AdriiiPRodri: These lines are common in several parts of the code, maybe we can create a helper function to improve the maintainability of this code. Something like `get_requirement_metadata` in base.py

api/src/backend/tasks/jobs/reports/base.py
```diff
@@ -0,0 +1,792 @@
...
+        Args:
+            data: Aggregated compliance data
+            **kwargs: Framework-specific options (e.g., only_failed)
+
+        Returns:
+            List of ReportLab elements
+        """
+        from tasks.jobs.threatscore_utils import _load_findings_for_requirement_checks
+
+        from .components import create_status_badge
```
AdriiiPRodri: Can we move this imports to the top?

api/src/backend/tasks/jobs/reports/ens.py
```diff
@@ -0,0 +1,1003 @@
...
+                        [COLOR_WHITE, colors.Color(1.0, 0.95, 0.95)],
+                    ),
+                ]
+            )
+        )
+        elements.append(table)
+
+        return elements
+
+    def create_detailed_findings(self, data: ComplianceData, **kwargs) -> list:
```
AdriiiPRodri, line 746: Are we taking into account the parameter `include_manual`?

api/src/backend/tasks/jobs/reports/base.py
```diff
@@ -0,0 +1,792 @@
...
+
+        elements = []
+        only_failed = kwargs.get("only_failed", True)
+
+        # Filter requirements if needed
+        requirements = data.requirements
+        if only_failed:
+            requirements = [r for r in requirements if r.status == StatusChoices.FAIL]
```
AdriiiPRodri: What about the `include_manual` parameter?

api/src/backend/tasks/jobs/reports/base.py
```diff
@@ -0,0 +1,792 @@
...
+            compliance_id: Compliance framework identifier
+            provider_id: Provider identifier
+            provider_obj: Optional pre-fetched Provider
+            requirement_statistics: Optional pre-aggregated statistics
+            findings_cache: Optional pre-loaded findings
+
+        Returns:
+            Aggregated ComplianceData object
+        """
+        from tasks.jobs.threatscore_utils import (
```
AdriiiPRodri: Same here, we need to move this to the top if possible

api/src/backend/tasks/jobs/report.py
```diff
@@ -3486,43 +220,36 @@ def generate_compliance_reports(
...
     )
     requirement_statistics = _aggregate_requirement_statistics_from_database(
         tenant_id, scan_id
     )
 
-    # Create shared findings cache (major optimization for findings queries)
+    # Create shared findings cache
     findings_cache = {}
-    logger.info("Created shared findings cache for both reports")
+    logger.info("Created shared findings cache for all reports")
```
AdriiiPRodri, line 251: This is important, I think something tricky is happening. In Python, **an empty dictionary evaluates to False**:

```
bool({})
False
```

So I think the following is happening, and I'd like us to verify whether this is the case, since it could potentially break the cache system. I'll descrbe what I think is going on:

- `generate_compliance_reports` creates a single cache to be shared across the three PDFs: `findings_cache = {}`, and passes it to each `generate_*_report.` Given the idea of this cache, this part is correct
- Inside each generator, `_load_compliance_data` builds the `ComplianceData` like this: `findings_by_check_id = findings_cache or {}` (line 664). **Since {} evaluates to False, a new dict is created, losing the reference to the shared cache**
- Then create_detailed_findings calls `_load_findings_for_requirement_checks(…, data.findings_by_check_id, …)`. That helper fills the dict it receives, but now it is the new dict, not the one created in `generate_compliance_reports`. **This is where the problem is**
- Result: after generating the first report, the original cache remains empty and the second and third reports query the database again for the same `check_id`. The query and memory savings of the shared findings cache are lost, and for large scans the DB load and generation time increase significantly
- To fix this, we should keep the same reference even if it is empty: `findings_by_check_id = findings_cache if findings_cache is not None else {}`. That way, the nitial dict is populated during the first report and the rest can reuse it without querying again

I have not tested this, but I believe this is what is happening, so I'd like you to check it when you can. If that is the case, I think the solution described here should work

AdriiiPRodri (changes requested):

api/src/backend/tasks/tests/test_reports_base.py
pedrooot: Done

api/src/backend/tasks/tests/test_reports_threatscore.py
pedrooot: Totally

api/src/backend/tasks/jobs/reports/nis2.py
pedrooot: Done

api/src/backend/tasks/jobs/reports/base.py
pedrooot: Done

api/src/backend/tasks/jobs/reports/ens.py
pedrooot: Now yes, thanks for the heads up!

api/src/backend/tasks/jobs/reports/base.py
pedrooot: Added, thanks!

api/src/backend/tasks/jobs/reports/base.py
pedrooot: Done

api/src/backend/tasks/jobs/report.py
pedrooot: Totally, good point thanks for the insight

AdriiiPRodri: No problem!

api/src/backend/tasks/jobs/report.py
```diff
@@ -3682,12 +360,13 @@ def generate_compliance_reports(
                         else ""
                     )
                     logger.info(
-                        f"ThreatScore snapshot created with ID {snapshot.id} "
-                        f"(score: {snapshot.overall_score}%{delta_msg})"
+                        "ThreatScore snapshot created with ID %s (score: %s%%%s)",
```
AdriiiPRodri: I think there is one extra "%"

AdriiiPRodri (changes requested):

api/src/backend/tasks/jobs/report.py
AdriiiPRodri: Taking advantage of this change, we are going to use Python f-strings for messages, just like in the rest of the code

api/src/backend/tasks/jobs/threatscore.py
```diff
@@ -131,9 +131,18 @@ def compute_threatscore_metrics(
...
 
         m = metadata[0]
-        risk_level = getattr(m, "LevelOfRisk", 0)
-        weight = getattr(m, "Weight", 0)
+        risk_level_raw = getattr(m, "LevelOfRisk", 0)
+        weight_raw = getattr(m, "Weight", 0)
         section = getattr(m, "Section", "Unknown")
+        # Ensure numeric types for calculations (compliance data may have str)
+        try:
+            risk_level = int(risk_level_raw) if risk_level_raw else 0
```
AdriiiPRodri: This does not need to be done, it is just a recommendation. In these cases, being strict and letting the exception be raised is convenient, because it would mean that we are not using our code correctly. If a class expects an integer and we pass another data type, masking this and “fixing” it can hide incorrect uses of our own code

AdriiiPRodri (changes requested):

api/src/backend/tasks/tests/test_reports.py
```diff
@@ -1,56 +1,32 @@
-import io
 import uuid
-from unittest.mock import MagicMock, Mock, patch
+from unittest.mock import Mock, patch
```
AdriiiPRodri, line 2: Let's change the filename, with s or without it

AdriiiPRodri (changes requested):

api/src/backend/tasks/jobs/report.py
pedrooot: fixed, done ✅

api/src/backend/tasks/jobs/threatscore.py
pedrooot: totally, done ✅

api/src/backend/tasks/tests/test_reports.py
pedrooot: done ✅

AdriiiPRodri (approved):

PR #9517, opened 2025-12-11
feat(s3): new check s3_bucket_object_public to detect public objects

### LINKED ISSUE #3463: Add a new s3 check to verify if objects inside the bucket are public
### New feature motivation

The s3_bucket_public_access checks for public access at the bucket level, but objects inside of it might be public

### Solution Proposed

Its not feasible to check every object in the bucket. My proposal is to use a function that will select a user-defined (via config options) number of random objects in the bucket, and check if they are public. What I am seeing on my current assessment is that there are buckets that arnt public, but every object in the buckets are public, so this check would catch this type of misconfig. 

Risk is mitigated (when compared to a full-blown public bucket) as you cant simply list the objects in the bucket, as the bucket is not publicly accessible.

Here is some pseduo-code that could be modified and used
```
import boto3
import random

def list_and_randomly_select_s3_objects(bucket_name, number_of_objects=3):
    # Initialize a boto3 client
    s3 = boto3.client('s3')
    
    # Retrieve the list of objects in the bucket
    try:
        response = s3.list_objects_v2(Bucket=bucket_name)
        objects = response.get('Contents', [])
        
        # Check if the bucket is empty
        if not objects:
            print("The bucket is empty.")
            return []
        
        # Extract object keys
        object_keys = [obj['Key'] for obj in objects]
        
        # Randomly select the user-defined number of objects, default is 3
        selected_keys = random.sample(object_keys, min(len(object_keys), number_of_objects))
        
        print(f"Randomly selected object keys: {selected_keys}")
        return selected_keys
    except Exception as e:
        print(f"An error occurred: {e}")
        return []

# Example usage
bucket_name = 'your-bucket-name'
selected_objects = list_and_randomly_select_s3_objects(bucket_name, 3)  # You can change 3 to any number you prefer
```

 

None

### Additional context

_No response_

Synchx00:
### Context

The current S3 checks in Prowler verify public access at the bucket level (via Bucket Policies or ACLs). However, a bucket can be private while individual objects inside it are public (e.g., via Object ACLs). This PR addresses this gap by adding a check to detect public objects within buckets.

Fix #3463

### Description

I have added a new check: `s3_bucket_object_public`.

**Logic implemented:**
1. Iterates through all S3 buckets.
2. Lists objects in the bucket (up to 100).
3. Randomly samples 3 objects (as suggested in the issue to avoid performance penalties on large buckets).
4. Inspects the ACL of these sampled objects.
5. Returns **FAIL** if an object grants permissions to the `AllUsers` group (URI: `http://acs.amazonaws.com/groups/global/AllUsers`).
6. Returns **PASS** if the sampled objects are private or if the bucket is empty.

**Dependencies:**
- No new external dependencies. Uses standard `boto3` methods (`list_objects_v2`, `get_object_acl`).

### Steps to review

To verify this check manually:

1. **Create a test S3 bucket** and upload two files:
   - One private file.
   - One public file (e.g., `aws s3api put-object-acl --bucket <bucket> --key public.txt --acl public-read`).
2. **Run the specific check** against that bucket:
   ```bash
   python prowler-cli.py aws --check s3_bucket_object_public --resource-arn arn:aws:s3:::<your-bucket-name>

### Checklist

#### UI

#### API

### License

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **New Features**
  * Added an S3 control to spot-check bucket objects for public ACL grants (AllUsers/AuthenticatedUsers) and generate per-bucket PASS/FAIL findings (disabled by default).
  * Added configuration to enable the check and tune sampling limits (max objects and sample size).
* **Bug Fixes**
  * Improved reporting when spot-checking is blocked or errors occur by producing clearer MANUAL outcomes and messages.
* **Tests**
  * Added coverage for the new S3 spot-check behavior, including empty buckets, public ACL detection, and error handling.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->


jfagoagas:
Hi @Synchx00 thanks for this contribution. I apologise for the delay. We'll review it and get back to you as soon as possible.

Thanks for using Prowler 🚀

andoniaf:
Hi @Synchx00, thank you for working on this! We appreciate your contribution to Prowler. 🙌

After reviewing the implementation, we have some concerns about the sampling approach, but we think we can address them with some adjustments.

**Our Concerns**

1. `ListObjectsV2` returns objects in lexicographical order, not randomly distributed across the bucket. This means the check only examines objects whose keys sort early alphabetically. Public objects named `z-sensitive-file.txt` or `uploads/2024/file.txt` might never be examined.
2. **Sampling has inherent limitations for detecting sparse violations.** For a bucket with 10,000 objects where only a handful are public, a small sample may not catch them. This could give users a false sense of security when they see "PASS".
3. There's already a root-cause check - `s3_bucket_acl_prohibited` verifies that buckets have `BucketOwnerEnforced` Object Ownership enabled, which prevents public object ACLs entirely. This is AWS's recommended approach since April 2023.

**Suggested Path Forward**
We'd like to propose making this check configurable and disabled by default. This way:

- Users who understand the limitations can opt-in
- It won't create false assurance for users who don't realize it's sampling
- Your work is preserved for those who find value in it (e.g., spot-checking during assessments)

Specifically, we'd suggest:

1. Add configuration options in the [Prowler config](https://docs.prowler.com/developer-guide/configurable-checks#configurable-checks-in-prowler) file for:

- `s3_bucket_object_public_enabled`: false (disabled by default)
- `s3_bucket_object_public_max_objects`: 100 (how many objects to list)
- `s3_bucket_object_public_sample_size`: 3 (how many to check ACLs on)

2. Update the metadata to clarify the limitations:

- Severity could be `low` given the sampling limitations
- Description/Notes should clearly state this is a spot-check, not comprehensive
- Perhaps add a recommendation to use `s3_bucket_acl_prohibited` for complete coverage


3. Consider the status messaging - When passing, the `status_extended` could say something like: `"No public objects detected in spot-check sample of {n} objects in bucket {name}. For complete assurance, ensure ACLs are disabled via Object Ownership settings."`

What do you think about this approach? We're happy to provide more guidance on implementing the configuration if needed.

Thanks again for your work on this!

Synchx00:
Hi @andoniaf, thanks for the detailed feedback!

I completely agree with your concerns regarding the lexicographical order of `ListObjectsV2` and the potential for creating a false sense of security with the current sampling method.

The proposed path forward, making this check configurable (disabled by default), adding sample size controls, and updating the metadata to explicitly state these limitations, sounds like the perfect solution.

I will start working on these changes right away!

jfagoagas:
Hi @Synchx00 are you still planning on work on this PR?

Synchx00:
Hi @jfagoagas, yes. I will get this wrapped up this weekend.

Synchx00:
Thanks for the patience @andoniaf, @jfagoagas — pushed the follow-up in a812249c4.

  Mapping back to @andoniaf's feedback:

  1. Configurable + disabled by default — three keys added to prowler/config/config.yaml:
  s3_bucket_object_public_enabled: False
  s3_bucket_object_public_max_objects: 100
  s3_bucket_object_public_sample_size: 3
  1. The check early-returns with no findings when _enabled is False.
  2. Metadata disclosure — Severity is now low. The Description, Notes, and Recommendation all state explicitly that this is a sampling spot-check, not a comprehensive audit, and point users to
  s3_bucket_acl_prohibited for full coverage (added to RelatedTo).
  3. Status messaging — PASS now uses your suggested wording:
"No public objects detected in spot-check sample of {n} objects in bucket {name}. For complete assurance, ensure ACLs are disabled via Object Ownership settings."

  3. FAIL names the offending keys and reports the sample size.

  Other small cleanups: switched to Check_Report_AWS, and the AllUsers grantee check now matches the URI exactly (http://acs.amazonaws.com/groups/global/AllUsers) instead of substring-matching "AllUsers".

  Added 6 unit tests in tests/providers/aws/services/s3/s3_bucket_object_public/ covering: disabled-by-default, empty bucket, all-private objects, public-object FAIL, AccessDenied → MANUAL, and other
  ClientError → MANUAL. All green locally via poetry run pytest -n auto -vvv -s -x tests/providers/aws/services/s3/s3_bucket_object_public/.

danibarranqueroo (approved):
Really good job! Thanks for this and for all the effort ❤️

PR #9523, opened 2025-12-12
feat(lighthouse): filter out non-compatible OpenAI models

0xbadshah:
### Context

OpenAI provides lot of non-compatible LLM models. For a model to work with Lighthouse it must support text input, text output and tool calls.

### Description

This PR filters out non-compatible models at the Django task level. The filtering is based on the model names. 

### Steps to review

1. Add OpenAI key to Lighthouse
2. In the models section, you'll not find models that don't support text input, text output or tool calling (ex: tts-1, sora-2, etc)

### Checklist

#### UI

#### API

### License


josema-xyz (changes requested):
When returning a dictionary with the key equals to the value instead of a list?

0xbadshah:
@josemazo OpenAI is the only provider that returns model ID and not model name. They take care of having shorter model ID. But comparing this to other providers (like Bedrock and OpenRouter, etc), they have model IDs that are not very straightforward (has versions, region info, etc) but have model names that can be displayed for the end user.

So because we use <Model Name>:<Model ID> dict for other providers, for OpenAI we just pass <Model ID>:<Model ID>.

AdriiiPRodri (approved):

PR #9574, opened 2025-12-16
feat(report): add account id, alias and provider to PDF report

pedrooot:
### Description

This PR adds the account name, alias and provider name to the PDF report table.

### Steps to review

### Checklist

#### UI

#### API

### License

### DIFF
--- a/api/src/backend/tasks/jobs/report.py
+++ b/api/src/backend/tasks/jobs/report.py
@@ -1103,11 +1103,15 @@ def generate_threatscore_report(
         elements.append(Spacer(1, 0.5 * inch))
 
         # Add compliance information table
+        provider_alias = provider_obj.alias or "N/A"
         info_data = [
             ["Framework:", compliance_framework],
             ["ID:", compliance_id],
             ["Name:", Paragraph(compliance_name, normal_center)],
             ["Version:", compliance_version],
+            ["Provider:", provider_type.upper()],
+            ["Account ID:", provider_obj.uid],
+            ["Alias:", provider_alias],
             ["Scan ID:", scan_id],
             ["Description:", Paragraph(compliance_description, normal_center)],
         ]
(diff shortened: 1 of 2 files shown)


api/src/backend/tasks/jobs/report.py
```diff
@@ -2059,25 +2063,28 @@ def generate_ens_report(
...
-                    ("FONTNAME", (0, 0), (0, 6), "FiraCode"),
-                    ("BACKGROUND", (1, 0), (1, 6), colors.Color(0.95, 0.97, 1.0)),
-                    ("TEXTCOLOR", (1, 0), (1, 6), colors.Color(0.2, 0.2, 0.2)),
-                    ("FONTNAME", (1, 0), (1, 6), "PlusJakartaSans"),
+                    ("BACKGROUND", (0, 0), (0, -1), colors.Color(0.2, 0.4, 0.6)),
+                    ("TEXTCOLOR", (0, 0), (0, -1), colors.white),
+                    ("FONTNAME", (0, 0), (0, -1), "FiraCode"),
+                    ("BACKGROUND", (1, 0), (1, -1), colors.Color(0.95, 0.97, 1.0)),
+                    ("TEXTCOLOR", (1, 0), (1, -1), colors.Color(0.2, 0.2, 0.2)),
+                    ("FONTNAME", (1, 0), (1, -1), "PlusJakartaSans"),
```
AdriiiPRodri, line 2123: Do we need to make this change? What does changing the 6 to -1 imply? IMO we should add some comments here explaining how this works

AdriiiPRodri (changes requested):

api/src/backend/tasks/jobs/report.py
pedrooot: In ReportLab, table indices work like this:
	•	(col, row) – cell coordinates
	•	-1 means “up to the last row/column” (just like Python slicing)
	•	6 literally means row 6 (hard-coded)

The problem with hard-coding 6:
If your table has 7 rows (indices 0–6), it works perfectly. But if the table has more or fewer rows, the style either breaks or is not applied correctly.

Why -1 is better:
-1 is dynamic. It always applies the style up to the last row, regardless of how many rows the table has. It’s equivalent to doing list[0:-1] in Python.

Regarding the comments on the code, totally. I'll add them

AdriiiPRodri: Great! Thanks for the explanation 👌🏼

AdriiiPRodri (approved):

PR #9793, opened 2026-01-14
feat(attack-paths): add Bedrock and AttachRolePolicy privilege escalation queries

andoniaf:
## Summary
- Add two new AWS privilege escalation queries for Attack Paths feature:
  - [**iam:AttachRolePolicy + sts:AssumeRole**](https://pathfinding.cloud/paths/iam-014): Detects principals who can both attach policies to roles AND assume those roles, enabling a two-step attack to gain elevated access 
  - [**Bedrock Code Interpreter with PassRole**](https://pathfinding.cloud/paths/bedrock-001): Detects principals that can escalate privileges by passing a role to a Bedrock AgentCore Code Interpreter to execute code with elevated credentials

## Test plan
- [ ] Verify queries execute successfully in the Attack Paths graph view
- [ ] Test with AWS accounts that have principals with the relevant permissions
- [ ] Verify virtual nodes and relationships render correctly in the graph visualization

### DIFF
--- a/api/src/backend/api/attack_paths/query_definitions.py
+++ b/api/src/backend/api/attack_paths/query_definitions.py
@@ -336,6 +336,185 @@ _QUERY_DEFINITIONS: dict[str, list[AttackPathsQueryDefinition]] = {
                 ),
             ],
         ),
+        AttackPathsQueryDefinition(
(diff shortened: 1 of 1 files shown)


josema-xyz (approved):
🚀

PR #9826, opened 2026-01-19
feat: upgrade to Next.js 16.1.3

alejandrobailo:
### Context

Upgrade Prowler UI to Next.js 16.1.3 to leverage the latest features, performance improvements, and stay current with the framework.

### Description

- Upgrade Next.js from 15.5.9 to 16.1.3 and React from 19.2.2 to 19.2.3
- Migrate ESLint from v8 legacy config to v9 flat config format (`eslint.config.mjs`)
- Migrate `middleware.ts` to `proxy.ts` (new Next.js 16 convention)
- Update `revalidateTag` API calls to new required signature with second argument
- Move `reactCompiler` from experimental to stable config (now v1.0.0)
- Remove `--turbopack` flag from dev script (now default in Next.js 16)
- Replace `rss-parser` with `@extractus/feed-extractor` for RSS feed parsing

### Steps to review

1. **Verify build passes:** Run `pnpm run build` - should complete without errors
2. **Verify dev server:** Run `pnpm run dev` - should start without deprecation warnings
3. **Check TypeScript:** Run `pnpm run typecheck` - should pass
4. **Check ESLint:** Run `pnpm run lint:check` - should pass with zero errors
5. **Review `proxy.ts`:** Verify auth logic is preserved from old `middleware.ts`
6. **Review `next.config.js`:** Confirm `reactCompiler: true` moved to top level
7. **Review `actions/api-keys/api-keys.ts`:** Verify `revalidateTag` calls use `"max"` profile
8. **Review `actions/feeds/feeds.ts`:** Verify RSS feed parsing works with new library
9. **Test auth flow:** Sign in/out works correctly with the new proxy

### Checklist

<details>

</details>

#### SDK/CLI

#### UI

#### API

### License


Alan-TheGentleman:
Hey\! Did a review of the migration. Found a few things worth discussing:

**1. proxy.ts export signature**

Looking at the current code:
```typescript
export default auth((req: NextRequest & { auth: any }) => {
```

According to Next.js 16 docs, the function needs to be named `proxy` - either as a named export or a default with the name. The anonymous default export might not work as expected. Should be something like:
```typescript
export const proxy = auth((req: NextRequest & { auth: any }) => {
```

Can you double check this works correctly? I'm seeing conflicting info about whether next-auth's wrapper handles this automatically.

**2. Was middleware.ts deleted?**

I see proxy.ts was added but couldn't confirm middleware.ts was removed from the diff. If both files exist, Next.js might have unexpected behavior. Just want to make sure we don't have both sitting there.

**3. RSS feed data loss**

The switch from rss-parser to @extractus/feed-extractor looks clean, but noticed:
```typescript
author: undefined,  // was: item.creator || item.author
categories: [],     // was: item.categories || []
```

Is this intentional? If we're not displaying author/categories anywhere in the UI then no worries, just wanted to flag it in case we need that data.

**4. Exported constants**

Noticed several constants that were private are now exported (STEP_COLORS, RISK_LEVELS, NAVIGATION_TYPE, INFO_FIELD_VARIANTS, etc). Was this intentional or a side effect of some eslint rule? If they're not used outside their modules, might be better to keep them private.

Let me know if any of these need clarification\!

alejandrobailo:
Hey Alan! Thanks for the thorough review. Let me address each point:

**1. proxy.ts export signature**

The current implementation is correct according to Next.js 16 docs. The official documentation states:

> *"The file must export a single function, either as a default export or named `proxy`."*

Our code uses a default export (`export default auth(...)`), which is one of the two valid options. NextAuth's `auth()` wrapper returns a function that Next.js recognizes as the proxy handler. No changes needed here.

**2. middleware.ts deleted**

Confirmed! `middleware.ts` was properly removed. In Next.js 16, the `middleware.ts` file convention has been deprecated and renamed to `proxy.ts` - this is why we migrated from one to the other. The only `middleware.ts` files that exist now are inside `node_modules/next-auth/`, which is expected. You can verify with `find ui -name "middleware.ts" -not -path "*/node_modules/*"` - it returns nothing.

**3. RSS feed data loss**

Good catch. This is intentional - `@extractus/feed-extractor` has a different API than `rss-parser` and doesn't expose `creator`/`author` or `categories` fields in the same way. I verified that our UI doesn't display author or categories anywhere in the feed components, so there's no functional impact. The switch was made to resolve compatibility issues with the new Next.js version.

**4. Exported constants**

You're right - this was a side effect of fixing ESLint errors. The `@typescript-eslint/no-unused-vars` rule was flagging constants like `INFO_FIELD_VARIANTS` as unused because they're only referenced for type derivation (`typeof X[keyof typeof X]`). Exporting them was a pragmatic fix to silence ESLint while keeping the codebase error-free.

It's not ideal but has no negative impact - tree-shaking will remove them if unused externally. We could revisit this with `// eslint-disable-next-line` comments if preferred, but I'd consider that a separate cleanup task.

Let me know if you have any other questions!

Alan-TheGentleman (dismissed):
LGTM! Thanks for the clarifications. The Next.js 16 migration looks solid.

PR #9829, opened 2026-01-19
fix(scans): scheduled scans duplicates

vicferpoy:
### Context

When scans are running for more than 24 hours, more than one scheduled scan can be created for the same provider, leading to unexpected behaviors and errors.

### Description

With this approach, no more than one scheduled scan can be run in the same day for the same provider.
Also, when the scheduled scan task is triggered, it will take care of duplicates to clean the tenant.

> [!NOTE]
> Locally, more than one scheduled scans were forced under the same provider. After triggering the perform scheduled scan task, the duplicated scheduled scan was deleted and the scan was properly run.

### Steps to review

### Checklist

<details>

</details>

#### SDK/CLI

#### UI

#### API

### License


api/src/backend/tasks/tasks.py
```diff
@@ -275,44 +278,38 @@ def perform_scheduled_scan_task(self, tenant_id: str, provider_id: str):
...
+                f"Scheduled scan already executing for provider {provider_id}. Skipping."
+            )
+            return ScanTaskSerializer(instance=executing_scan).data
 
         executed_scan = Scan.objects.filter(
             tenant_id=tenant_id,
             provider_id=provider_id,
             task__task_runner_task__task_id=task_id,
-        ).order_by("completed_at")
+        ).first()
```
jfagoagas, line 301: Don't you need to order it by completion time? I assume not but I can't see why.

api/src/backend/tasks/tasks.py
```diff
@@ -275,44 +278,38 @@ def perform_scheduled_scan_task(self, tenant_id: str, provider_id: str):
...
-                trigger=Scan.TriggerChoices.SCHEDULED,
-                state=StateChoices.EXECUTING,
-                scheduler_task_id=periodic_task_instance.id,
-                scheduled_at__date=datetime.now(timezone.utc).date(),
-            ).exists()
-            or executed_scan.exists()
-        ):
-            # Duplicated task execution due to visibility timeout or scan is already running
+        if executed_scan:
+            # Duplicated task execution due to visibility timeout
```
jfagoagas, line 304: 1. Don't we need to increase it for this kind of tasks? 
2. Is the message being re-queued?

api/src/backend/tasks/tasks.py
vicferpoy: The `.order_by("completed_at")` is unnecessary because the query filters by `task__task_runner_task__task_id=task_id`, which is the current Celery task's unique ID. A given Celery task ID maps to at most one scan, so the queryset will always return 0 or 1 results. Ordering is meaningless.

api/src/backend/tasks/tasks.py
vicferpoy: 1. No. Since `acks_late=False`, the message is acked instantly upon receipt. The visibility timeout is effectively irrelevant for re-queuing because the message is already acked before the scan runs. Increasing it wouldn't change behavior. I would even say that we could reduce it.
2. Under normal operation, no. The immediate ack removes the message from the unacked queue right away. The only scenario where re-queuing could happen is if the worker crashes or dies between receiving the message and sending the ack). The duplicate detection code is a defensive safety net, not a routine path.

api/src/backend/tasks/tasks.py
jfagoagas: Great one, thanks for the clarification.

jfagoagas (dismissed):

AdriiiPRodri (approved):

PR #9998, opened 2026-02-09
fix(aws): sanitise zip extraction to prevent Zip Slip vulnerability

sandiyochristan:
## Description

Fixed a **Zip Slip** vulnerability in `awslambda_function_no_secrets_in_code.py`.
Previously, `zipfile.extractall` was used on downloaded Lambda code, which could allow malicious zip files to overwrite arbitrary files via directory traversal (e.g., `../../evil.sh`).

### Changes
- Added `safe_extract_zip` utility in `prowler/lib/utils/utils.py`.
- Updated `awslambda_function_no_secrets_in_code` to use `safe_extract_zip`.

## Checklist
- [x] Verified with unit tests (`tests/test_safe_extract.py`)
- [x] Validated with reproduction script (`tests/reproduce_zip_slip.py`)

### DIFF
--- a/prowler/lib/utils/utils.py
+++ b/prowler/lib/utils/utils.py
@@ -11,11 +11,12 @@ except ImportError:
 import re
 import sys
 import tempfile
+import zipfile
 from datetime import datetime
 from hashlib import sha512
 from io import TextIOWrapper
 from ipaddress import ip_address
-from os.path import exists
+from os.path import abspath, exists, join, normpath
 from time import mktime
 from typing import Any, Optional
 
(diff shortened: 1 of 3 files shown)


danibarranqueroo:
Hello @sandiyochristan, this PR contains changes that you added in #9997. Please remove those changes from here and after that we'll review it once we have time. Thanks for this! 🚀

sandiyochristan:
Done.

sandiyochristan:
Added entry to `prowler/CHANGELOG.md` to fix the `check-changelog` workflow failure. CI should pass now.

jfagoagas:
@sandiyochristan please review all the issues from linters. Thanks!

HugoPBrito:
Hi @sandiyochristan, thanks for the contribution. We are closing this PR because there is no real vulnerability to fix.

Python's standard library already prevents Zip Slip: `ZipFile.extract` and `extractall` strip leading slashes, drive letters, and all `".."` components from member names before writing, so a member cannot escape the target directory ([docs](https://docs.python.org/3/library/zipfile.html#zipfile.ZipFile.extract)).

Confirmed on Python 3.12 (Prowler supports `>=3.10,<3.13`):

```python
import io, os, zipfile, tempfile
buf = io.BytesIO()
with zipfile.ZipFile(buf, "w") as zf:
    zf.writestr("../pwned.txt", "x")
buf.seek(0)
with tempfile.TemporaryDirectory() as d:
    with zipfile.ZipFile(buf) as zf:
        zf.extractall(d)
    print(os.path.exists(os.path.join(d, "..", "pwned.txt")))  # -> False
```

`zipfile` also does not restore symlinks, so the `tarfile` style escape does not apply either. The original `extractall` is safe as is.

Thanks again!

PR #10003, opened 2026-02-10
feat(api): add OpenStack provider support

danibarranqueroo:
### Context

Add OpenStack as a supported provider in the API, allowing users to register and scan OpenStack projects.

### Description

This PR adds full OpenStack provider support to the API:

- SDK layer: Added clouds_yaml_content parameter to OpenstackProvider.__init__(), setup_session(), and test_connection(), enabling the API to pass clouds.yaml content as a string (instead of a file path)
- Model layer: Added OPENSTACK to ProviderChoices enum with UID validation (UUID or alphanumeric project name up to 255 characters)
- PostgreSQL migration: Added ALTER TYPE provider ADD VALUE to extend the native enum
- SDK wiring: Connected to OpenstackProvider with project_id kwarg for explicit credentials, or clouds_yaml_content/clouds_yaml_cloud for clouds.yaml authentication
- Secret handling: Added serializers for both authentication methods:
  - clouds.yaml content + cloud name
  - Explicit credentials (auth_url, username, password, region_name + optional fields)
- Connection testing: Special-cased OpenStack in prowler_provider_connection_test() since it uses project_id instead of the generic provider_id parameter
- Test coverage: Valid/invalid UID validation, secret creation for both auth methods, provider return and kwargs tests
- OpenAPI specs: Updated v1.yaml with OpenStack enum values and credential schemas

### Steps to review

1. Review the migration file 0076_openstack_provider.py follows the pattern of previous provider migrations
2. Verify UID validation regex ^[a-zA-Z0-9][a-zA-Z0-9._-]{0,254}$ matches OpenStack project ID format
3. Check secret serializers handle both authentication methods correctly
4. Review prowler_provider_connection_test() special handling for OpenStack (uses project_id instead of provider_id)
5. Run the test suite: pytest api/src/backend/api/tests/test_views.py -k openstack -v

### Checklist

<details>

</details>

#### SDK/CLI

#### UI

#### API

### License


api/src/backend/config/django/testing.py
```diff
@@ -18,6 +18,7 @@
 
 DATABASE_ROUTERS = []
 TESTING = True
+REST_FRAMEWORK["PAGE_SIZE"] = 50  # noqa: F405
```
josema-xyz: This could mask pagination errors in the future. I think you can override it for the tests you need.

api/src/backend/api/utils.py
```diff
@@ -208,6 +215,8 @@ def get_prowler_provider_kwargs(
             **prowler_provider_kwargs,
             "filter_accounts": [provider.uid],
         }
+    elif provider.provider == Provider.ProviderChoices.OPENSTACK.value:
+        pass
```
josema-xyz, line 223: I think a comment would be nice here.

prowler/providers/openstack/openstack_provider.py
```diff
@@ -203,6 +213,39 @@ def setup_session(
...
+        Args:
+            clouds_yaml_content: The full YAML content of a clouds.yaml file.
+            clouds_yaml_cloud: Cloud name to use from the clouds.yaml content.
+
+        Returns:
+            OpenStackSession configured from the provided clouds.yaml content.
+        """
+        tmp_file = None
+        try:
+            tmp_file = tempfile.NamedTemporaryFile(
```
josema-xyz: Maybe this file could have harder permissions, like `600`. Check the `api/src/backend/api/apps.py:_write_key_file`.

josema-xyz (commented):
Just some little things!

api/src/backend/config/django/testing.py
danibarranqueroo: Good catch, however I've tried per-test overrides via settings.REST_FRAMEWORK["PAGE_SIZE"] but DRF caches the page size at startup, so runtime mutations don't propagate to the pagination class. We can set it to exactly 11 (matching the fixture size) instead of 50, with a comment to update when adding providers.

api/src/backend/api/utils.py
danibarranqueroo: Added!

prowler/providers/openstack/openstack_provider.py
danibarranqueroo: Good point. Instead of hardening permissions, I've now removed the temp file entirely. Now the YAML content is parsed directly with yaml.safe_load(), following the same approach as the Kubernetes provider.

api/src/backend/config/django/testing.py
josema-xyz: You are right with that, so maybe we can set it to more round number, 20, or 15? Keeping the comment, of course.

josema-xyz (approved):

HugoPBrito (approved):

PR #10116, opened 2026-02-18
fix(api): handle provider deletion race condition in attack paths scan

josema-xyz:
### Description

Fixes the race condition where a provider is deleted while an attack Paths scan is running.                                   

Applied the `@handle_provider_deletion` decorator (already used by every other scan task) to the Attack Paths task. Broadened the decorator's except clause from `IntegrityError` to `DatabaseError`, since `save(update_fields=...)` on a CASCADE-deleted row raises the latter (`IntegrityError` is a subclass of `DatabaseError`, so existing behavior is preserved). Protected the scan error handler from secondary failures that could mask the original exception.

### Steps to review

Launch the tests.

### Checklist

#### API

### License

### DIFF
--- a/api/src/backend/tasks/jobs/attack_paths/scan.py
+++ b/api/src/backend/tasks/jobs/attack_paths/scan.py
@@ -204,8 +204,8 @@ def run(tenant_id: str, scan_id: str, task_id: str) -> dict[str, Any]:
         return ingestion_exceptions
 
     except Exception as e:
-        exception_message = utils.stringify_exception(e, "Cartography failed")
-        logger.error(exception_message)
+        exception_message = utils.stringify_exception(e, "Attack Paths scan failed")
+        logger.exception(exception_message)
         ingestion_exceptions["global_error"] = exception_message
 
         # Handling databases changes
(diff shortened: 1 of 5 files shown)


api/src/backend/tasks/jobs/attack_paths/scan.py
```diff
@@ -204,20 +204,26 @@ def run(tenant_id: str, scan_id: str, task_id: str) -> dict[str, Any]:
...
+        logger.exception(exception_message)
         ingestion_exceptions["global_error"] = exception_message
 
         # Handling databases changes
         try:
             graph_database.drop_database(tmp_cartography_config.neo4j_database)
 
         except Exception:
-            logger.exception(
+            logger.error(
```
AdriiiPRodri, line 216: With this change we lose the stacktrace, is this intentional?

AdriiiPRodri (changes requested):

api/src/backend/tasks/jobs/attack_paths/scan.py
josema-xyz: Yes, is an error on a cleaning operation, a cleaning after the real problem.

AdriiiPRodri (approved):

PR #10118, opened 2026-02-19
feat(api): filter Attack Paths query results by provider_id

josema-xyz:
### Description

Nodes and relationships returned by Attack Paths queries are now filtered by `provider_id`, so results from other providers in the same tenant database are excluded. Virtual nodes and relationships created by `apoc` are also stamped with `provider_id` for consistency.

### Steps to review

Run tests, a local scan and check the queries.

### Checklist

#### API

### License

### DIFF
--- a/api/src/backend/api/attack_paths/views_helpers.py
+++ b/api/src/backend/api/attack_paths/views_helpers.py
@@ -35,6 +35,7 @@ def prepare_query_parameters(
     definition: AttackPathsQueryDefinition,
     provided_parameters: dict[str, Any],
     provider_uid: str,
+    provider_id: str,
 ) -> dict[str, Any]:
     parameters = dict(provided_parameters or {})
     expected_names = {parameter.name for parameter in definition.parameters}
@@ -56,6 +57,7 @@ def prepare_query_parameters(
 
     clean_parameters = {
         "provider_uid": str(provider_uid),
+        "provider_id": str(provider_id),
     }
 
     for definition_parameter in definition.parameters:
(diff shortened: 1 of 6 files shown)


api/src/backend/api/tests/test_attack_paths.py
```diff
@@ -169,7 +179,37 @@ def __exit__(self, exc_type, exc, tb):
...
+        nodes=[node_keep, node_drop],
+        relationships=[rel_keep, rel_drop_by_provider],
+    )
+
+    result = views_helpers._serialize_graph(graph, provider_id)
+
+    assert len(result["nodes"]) == 1
+    assert result["nodes"][0]["id"] == "n1"
+    assert len(result["relationships"]) == 1
+    assert result["relationships"][0]["id"] == "r1"
```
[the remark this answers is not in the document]
josema-xyz: Good catch! Added a test case with a relationship that has the matching provider_id but connects to a filtered-out node, exercising the orphan check.

api/src/backend/api/attack_paths/views_helpers.py
```diff
@@ -95,9 +98,14 @@ def execute_attack_paths_query(
         )
 
 
-def _serialize_graph(graph):
+def _serialize_graph(graph, provider_id: str):
```
AdriiiPRodri, line 101: `_serialize_graph` only checks `provider_id`, but real nodes/relationships now carry `_provider_id`. That can drop legit nodes and return an empty graph (only the virtual Internet node has `provider_id` set in the query). Maybe accept both during the transition, e.g `props.get("_provider_id")` or `props.get("provider_id")`, or set `_provider_id` on the virtual nodes too

AdriiiPRodri (changes requested):

api/src/backend/api/attack_paths/views_helpers.py
josema-xyz: This is planed. `provider_id` in the current and future graphs. When we remove `provider_id` beacuse `_provider_id` is in every graph, we'll that change also here.

AdriiiPRodri (approved):

PR #9984, opened 2026-02-06
feat(image): add container image provider for CLI scanning

andoniaf:
### Context

Adds a new `image` provider to Prowler that enables container image scanning via Trivy, directly from the CLI.

### Description

New Image provider that wraps Trivy for vulnerability, secret, and misconfiguration scanning of container images. 

- **Provider class** (`ImageProvider`) with core scanning functionality
- **Trivy integration** for vuln/secret/misconfig detection with JSON output parsing
- **Local Docker auth** — scans images available in the local Docker daemon
- **CLI support** — `prowler image -I <image>` with options for severity filters, timeout, ignore-unfixed, and image-list-file
- **CheckReportImage model** for image-specific findings
- **Custom exceptions** (9000–9005) with clear remediation messages
- **Error handling** for Trivy failures (non-zero exit, binary not found, image pull errors)
- **Comprehensive test coverage** (26 tests)

### Steps to review

1. Review the new provider at `prowler/providers/image/image_provider.py`
2. Check CLI argument definitions in `prowler/providers/image/lib/arguments/arguments.py`
3. Verify integration points in `prowler/__main__.py`, `prowler/lib/cli/parser.py`, `prowler/lib/outputs/finding.py`, `prowler/lib/outputs/summary_table.py`, and `prowler/providers/common/provider.py`
4. Run tests: `pytest -vv tests/providers/image/`
5. Manual test with a public image: `python prowler-cli.py image -I alpine:3.18`

Docs are in this other PR: https://github.com/prowler-cloud/prowler/pull/9986

### Checklist

<details>

<summary><b>Community Checklist</b></summary>

- [x] This feature/issue is listed in [here](https://github.com/prowler-cloud/prowler/issues?q=sort%3Aupdated-desc+is%3Aissue+is%3Aopen) or roadmap.prowler.com
- [x] Is it assigned to me, if not, request it via the issue/feature in [here](https://github.com/prowler-cloud/prowler/issues?q=sort%3Aupdated-desc+is%3Aissue+is%3Aopen) or [Prowler Community Slack](goto.prowler.com/slack)

</details>

- [x] Review if the code is being covered by tests.
- [x] Review if code is being documented following this specification https://github.com/google/styleguide/blob/gh-pages/pyguide.md#38-comments-and-docstrings
- [ ] Review if backport is needed.
- [ ] Review if is needed to change the [Readme.md](https://github.com/prowler-cloud/prowler/blob/master/README.md)
[...]

PR #10128, opened 2026-02-20
feat(api): add Image provider support for container image scanning

andoniaf:
### Context

Adds full Image provider support to the Prowler API, enabling container image vulnerability scanning via 

### Description

**API:**
- Database migration adding `image` to the provider enum (`0081_image_provider.py`)
- Provider serializer updates to accept `image` as a valid provider type
- Automatic detection of registry URL vs image reference in the UID field
- Connection test and scan task support for the Image provider
- Mutelist/exclusion support for Image provider findings
- Trivy version bump to 0.69.1 in Dockerfile

### Steps to review

1. Review API integration in `api/src/backend/api/` (migration, models, serializers, utils)
2. Review test coverage in `tests/providers/image/` and `api/src/backend/api/tests/`

### Checklist

<details>

</details>

- Are there new checks included in this PR? No (vulnerability findings come from Trivy, not custom checks)
    - If so, do we need to update permissions for the provider? N/A

#### SDK/CLI
    - If so, do we need to update permissions for the provider? N/A

#### API (if applicable)
- [x] Endpoint response output (if applicable) — see E2E results in Jira
- [x] Any other relevant evidence of the implementation (if applicable) — E2E Jira comment
- [x] Ensure new entries are added to api/CHANGELOG.md

### License


andoniaf:
Adding `no-changelog` label, as SDK changelog was already update in https://github.com/prowler-cloud/prowler/pull/9984

api/src/backend/api/utils.py
```diff
@@ -221,11 +227,41 @@ def get_prowler_provider_kwargs(
...
+        image_uid = provider.uid
+        registry_host = ImageProvider._extract_registry(image_uid)
+        if registry_host:
+            repo_and_tag = image_uid[len(registry_host) + 1 :]
+        else:
+            repo_and_tag = image_uid
+
+        is_registry_url = (
+            registry_host and "/" not in repo_and_tag and ":" not in repo_and_tag
+        )
```
josema-xyz: The same lines are also in `ImageProvider`. Couldn't be extracted into a method there?

prowler/providers/image/lib/registry/base.py
```diff
@@ -104,21 +119,23 @@ def _request_with_retry(self, method: str, url: str, **kwargs) -> requests.Respo
...
+    @staticmethod
+    def _next_page_url(resp: requests.Response) -> str | None:
         link_header = resp.headers.get("Link", "")
         if not link_header:
             return None
         match = re.search(r'<([^>]+)>;\s*rel="next"', link_header)
-        if not match:
-            return None
-        url = match.group(1)
-        if url.startswith("/"):
```
josema-xyz, line 121: From our friend: This removes relative URL handling, and some registries is what they use.

prowler/providers/image/image_provider.py
```diff
@@ -453,6 +469,29 @@ def run(self) -> list[CheckReportImage]:
...
+                try:
+                    image_findings = []
+                    for batch in self._scan_single_image(image):
+                        image_findings.extend(batch)
+                    yield (image, image_findings)
+                except (ImageScanError, ImageTrivyBinaryNotFoundError):
+                    raise
+                except Exception as error:
+                    logger.error(f"Error scanning image {image}: {error}")
+                    continue
```
josema-xyz: This should yield something, right? Like the `bad:image` text in the tests?

api/src/backend/api/v1/serializers.py
```diff
@@ -1702,6 +1704,15 @@ class Meta:
         resource_name = "provider-secrets"
 
 
+class ImageProviderSecret(serializers.Serializer):
```
josema-xyz, line 1707: We need a validator here, something like:
- If all three are missing: public image
- Else if token: Use always token
- If not token, both username and password needs to be here

josema-xyz (commented):
4 simple comments!

api/src/backend/api/utils.py
andoniaf: Good catch! Extracted a `_is_registry_url(image_uid)` staticmethod in `ImageProvider` that encapsulates the heuristic. Both `test_connection()` and `get_prowler_provider_kwargs()` now call it instead of duplicating the logic. Added tests in `TestIsRegistryUrl`.

prowler/providers/image/lib/registry/base.py
andoniaf: Fixed — restored the relative URL reconstruction. When the `Link` header contains a path-only URL (e.g. `</v2/...>`), we now rebuild the full URL using `resp.url`'s scheme and netloc. Added a test (`test_link_header_relative_url`) to cover this case.

prowler/providers/image/image_provider.py
andoniaf: Right, replaced `continue` with `yield (image, [])` so the caller always sees the failed image. The existing `test_skips_generic_error` test now passes.

api/src/backend/api/v1/serializers.py
andoniaf: Added a `validate()` method to `ImageProviderSecret` that enforces these rules:
- All three missing → valid (public registry)
- Token present → valid (token auth, username/password ignored)
- No token + both username and password → valid (basic auth)
- No token + only one of username/password → invalid

Added 6 tests covering valid and invalid combinations.

josema-xyz (dismissed):
You 🪨

josema-xyz (approved):

PR #10144, opened 2026-02-24
fix(api): harden security hub retries

AdriiiPRodri:
## Description

Security Hub integrations were failing intermittently with the error `generator didn't stop after throw()`.
The root cause was an invalid retry path in `rls_transaction`: when a Postgres replica cancels a long read
(`OperationalError: conflict with recovery`) inside the `with` block, the generator-based context manager
tries to continue and `yield` again, which violates the context manager protocol.

Fix implemented:

- Added `retry_on_replica` flag to `rls_transaction` (default `True`) to keep existing behavior intact.
- In Security Hub export, disabled internal retry (`retry_on_replica=False`) and moved retries to the job
  level with exponential backoff and final fallback to primary.

This avoids the generator error while still masking transient replica conflicts from end users.

## Acceptance Criteria

- Security Hub export does not fail with `generator didn't stop after throw()`.
- Transient replica recovery conflicts are retried and do not surface as user-visible failures.
- No behavioral change for other callers of `rls_transaction` (default retry remains enabled).
- Security Hub export falls back to primary on the final retry attempt.
- Unit tests cover: disabling replica retry in `rls_transaction` and retry behavior in Security Hub export.

## Testing

- `poetry run pytest api/src/backend/api/tests/test_db_utils.py -k retry_on_replica -v`
- `poetry run pytest api/src/backend/tasks/tests/test_integrations.py -k retries_on_operational_error -v`

## Checklist

- [ ] PR - Local environment
- [ ] Add evidences and how to test in in the PR
- [ ] After merge in prowler - dev at https://prowler-ui.dev.prowler.com / https://prowler-api.dev.prowler.com
- [ ] Add evidences to this ticket
- [ ] After merge in prowler-cloud - dev at https://cloud.dev.prowler.com / https://api.dev.prowler.com
- [ ] Add evidences to this ticket


jfagoagas (approved):
Great catch!!

PR #10164, opened 2026-02-24
fix(api): [security] check all user roles in HasPermissions, not only the first

sandiyochristan:
## Description

`HasPermissions.has_permission` evaluated each required permission only against `user_roles[0]` — the first role in the queryset. A user with multiple roles could be incorrectly denied access even when one of their other roles grants the necessary permission.

**Root cause:**
```python
for perm in required_permissions:
    if not getattr(user_roles[0], perm.value, False):  # only index 0
        return False
```

**Fix:**
```python
for perm in required_permissions:
    if not any(getattr(role, perm.value, False) for role in user_roles):
        return False
```

## Changes

- `api/src/backend/api/rbac/permissions.py` — one-line fix
- `api/src/backend/api/tests/test_rbac.py` — regression test for multi-role user

## Checklist

- [x] No new dependencies
- [x] Minimal, focused diff

### DIFF
--- a/api/CHANGELOG.md
+++ b/api/CHANGELOG.md
@@ -39,6 +39,7 @@ All notable changes to the **Prowler API** are documented in this file.
 
 - Bump `Pillow` to 12.1.1 (CVE-2021-25289) [(#10027)](https://github.com/prowler-cloud/prowler/pull/10027)
 - Remove safety ignore for CVE-2026-21226 (84420), fixed via `azure-core` 1.38.x [(#10110)](https://github.com/prowler-cloud/prowler/pull/10110)
+- Fix RBAC permission check to evaluate all user roles instead of only the first, preventing silent permission denial for valid multi-role users [(#10164)](https://github.com/prowler-cloud/prowler/pull/10164)
 
 ---
 
--- a/api/src/backend/api/rbac/permissions.py
+++ b/api/src/backend/api/rbac/permissions.py
@@ -36,7 +36,7 @@ class HasPermissions(BasePermission):
             return False
 
         for perm in required_permissions:
-            if not getattr(user_roles[0], perm.value, False):
+            if not any(getattr(role, perm.value, False) for role in user_roles):
                 return False
 
         return True
--- a/api/pyproject.toml
+++ b/api/pyproject.toml
@@ -24,7 +24,7 @@ dependencies = [
   "drf-spectacular-jsonapi==0.5.1",
   "gunicorn==23.0.0",
   "lxml==5.3.2",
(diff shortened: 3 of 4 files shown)


sandiyochristan:
## CI Failure Note

The `api-code-quality`, `api-security-scans`, and `api-tests` jobs may fail due to a structural limitation of fork-based PRs.

```
```

During CI, this is rewritten via `sed` to use the head branch name (`fix/rbac-check-all-roles`). Since that branch only exists in my fork (`sandiyochristan/prowler`), not in `prowler-cloud/prowler`, Poetry fails to resolve the dependency.

**This PR only touches API code** — `api/src/backend/api/rbac/permissions.py` — and does not require any SDK changes. The fix and its regression test are correct; the CI infrastructure failure is unrelated to the change itself.

sandiyochristan:

```yaml
```

api/CHANGELOG.md
```diff
@@ -40,6 +40,7 @@ All notable changes to the **Prowler API** are documented in this file.
 
 - Bump `Pillow` to 12.1.1 (CVE-2021-25289) [(#10027)](https://github.com/prowler-cloud/prowler/pull/10027)
 - Remove safety ignore for CVE-2026-21226 (84420), fixed via `azure-core` 1.38.x [(#10110)](https://github.com/prowler-cloud/prowler/pull/10110)
+- Fix RBAC permission check to evaluate all user roles instead of only the first, preventing silent permission denial for valid multi-role users [(#10164)](https://github.com/prowler-cloud/prowler/pull/10164)
```
josema-xyz, line 42: This changelog entry is deprecated, please get the latest changes from Prowlers' master branch.

sandiyochristan: Done — rebased on latest upstream master. The CHANGELOG now includes all current entries from master with our fix entry on top.

josema-xyz (approved):

api/src/backend/api/rbac/permissions.py
```diff
@@ -36,7 +36,7 @@ def has_permission(self, request, view):
             return False
 
         for perm in required_permissions:
-            if not getattr(user_roles[0], perm.value, False):
+            if not any(getattr(role, perm.value, False) for role in user_roles):
```
AdriiiPRodri, line 39: Hi Sandiyo, this code snippet is written this way intentionally. Right now we only allow one role per user, but if for some reason someone has more than one we will only use the first one. This is because at the moment we don’t have a way to merge permissions, so for now we can’t merge this PR. We really appreciate the time you invested, but for now we’re going to close this PR.

PR #10165, opened 2026-02-24
fix(api): [security] use defusedxml to prevent XML bomb DoS in SAML metadata parsing

sandiyochristan:
## Description

`SAMLConfiguration._parse_metadata` used `xml.etree.ElementTree.fromstring` to parse IdP-supplied metadata XML.

While Python's stdlib ET blocks classic XXE (external-entity file reads since Python 3.2), it does **not** protect against entity-expansion attacks such as the [Billion Laughs](https://en.wikipedia.org/wiki/Billion_laughs_attack) XML bomb. A small crafted payload with nested entity definitions can exhaust server memory and CPU, causing a denial-of-service.

**Before:**
```python
import xml.etree.ElementTree as ET
...
root = ET.fromstring(self.metadata_xml)  # vulnerable to XML bomb
```

**After:**
```python
from defusedxml import ElementTree as ET
...
root = ET.fromstring(self.metadata_xml)  # entity expansion blocked
```

`defusedxml` is already a declared dependency in `pyproject.toml`.

A broad `except Exception` clause is added alongside the existing `ET.ParseError` catch so that `defusedxml`-specific exceptions surface as a `ValidationError` with a descriptive message.

## Changes

- `api/src/backend/api/models.py` — replace stdlib `xml.etree.ElementTree` with `defusedxml.ElementTree`; expand exception handling
- `api/src/backend/api/tests/test_models.py` — adds regression test that verifies a four-level entity-expansion bomb is rejected

## How to review

1. Confirm the single import swap in `models.py`
2. Run `test_xml_bomb_rejected` — should pass (defusedxml raises an exception before any expansion occurs)

## Checklist

- [x] `defusedxml` already in project deps — no new dependencies
- [x] Minimal, focused diff


sandiyochristan:
## CI Failure Note

The `api-code-quality`, `api-security-scans`, and `api-tests` jobs may fail due to a structural limitation of fork-based PRs.

```
```

During CI, this is rewritten via `sed` to use the head branch name (`fix/api-saml-xml-bomb`). Since that branch only exists in my fork (`sandiyochristan/prowler`), not in `prowler-cloud/prowler`, Poetry fails to resolve the dependency.

**This PR only touches API code** — `api/src/backend/api/models.py` — and does not require any SDK changes. The fix (replacing stdlib `xml.etree.ElementTree` with `defusedxml`) and its test are correct; the CI infrastructure failure is unrelated to the change itself. `defusedxml` is already a declared dependency in `api/pyproject.toml`.

sandiyochristan:

```yaml
```

api/src/backend/api/models.py
```diff
@@ -2048,6 +2048,8 @@ def _parse_metadata(self):
             root = ET.fromstring(self.metadata_xml)
         except ET.ParseError as e:
             raise ValidationError({"metadata_xml": f"Invalid XML: {e}"})
+        except Exception as e:
```
josema-xyz: Using except `Exception` is too broad: if there's a bug in the code (e.g. `AttributeError`, `TypeError`), it would be swallowed and resented as "malformed XML" when it's actually a programming error. Probably `defusedxml` defines a set of exception that this case can use.

api/CHANGELOG.md
```diff
@@ -40,6 +40,7 @@ All notable changes to the **Prowler API** are documented in this file.
 
 - Bump `Pillow` to 12.1.1 (CVE-2021-25289) [(#10027)](https://github.com/prowler-cloud/prowler/pull/10027)
 - Remove safety ignore for CVE-2026-21226 (84420), fixed via `azure-core` 1.38.x [(#10110)](https://github.com/prowler-cloud/prowler/pull/10110)
+- Replace stdlib XML parser with `defusedxml` in SAML metadata parsing to prevent XML bomb (billion laughs) DoS attacks [(#10165)](https://github.com/prowler-cloud/prowler/pull/10165)
```
josema-xyz: This changelog entry is deprecated, please get the latest changes from Prowlers' master branch.

api/pyproject.toml
```diff
@@ -24,7 +24,7 @@ dependencies = [
   "drf-spectacular-jsonapi==0.5.1",
   "gunicorn==23.0.0",
   "lxml==5.3.2",
-  "prowler @ git+https://github.com/prowler-cloud/prowler.git@master",
+  "prowler @ git+https://github.com/prowler-cloud/prowler.git@2115344de80744239a856ef4f31d6037103db7a4",
```
AdriiiPRodri: We need to keep the reference to master here:

```suggestion
  "prowler @ git+https://github.com/prowler-cloud/prowler.git@master",
```

AdriiiPRodri: I understand the issue you had with this now. In that case, we will probably create a branch and cherry-pick your work so we don’t lose the authorship, since we must avoid merging this change into the pyproject

sandiyochristan (commented):
Please Review and Merge this change

api/pyproject.toml
sandiyochristan: Got it

api/src/backend/api/models.py
josema-xyz: Checking a little bit more about this exception, it should catch `DefusedXmlException`, the parent on any Exception from `defusedxml`.

api/CHANGELOG.md
josema-xyz: About this, it should go into the next release changelog, that as today, Mar 17 2026, is under 1.22.0.

sandiyochristan (commented):
LGTM! Good to go from my end

api/src/backend/api/models.py
AdriiiPRodri: @sandiyochristan once you do this we can merge this PR 🚀

api/CHANGELOG.md
josema-xyz: Thank you @AdriiiPRodri

api/src/backend/api/models.py
jfagoagas: @AdriiiPRodri @josema-xyz in this PR maintainers have permission to make edits. Please go ahead with the change in order to merge. Thanks!

sandiyochristan: Fixed — replaced `except Exception` with `except defusedxml.DefusedXmlException`, the parent class for all defusedxml-specific exceptions (DTDForbidden, EntitiesForbidden, ExternalReferenceForbidden, NotSupportedYet). Programming errors like AttributeError/TypeError will now propagate normally instead of being masked.

sandiyochristan: Done, catching `defusedxml.DefusedXmlException` now.

AdriiiPRodri (approved):

api/src/backend/api/models.py
AdriiiPRodri: Thanks!

PR #10166, opened 2026-02-24
fix(api): [security] use psycopg2.sql to safely compose DDL in PostgresEnumMigration

sandiyochristan:
## Description

`PostgresEnumMigration.create_enum_type` and `drop_enum_type` were building raw PostgreSQL DDL statements via Python f-strings and passing them directly to `cursor.execute`.

Although the enum name and values are hardcoded in Django migration files today, the pattern is **inherently unsafe** — any future caller that derives these arguments from external input would be vulnerable to SQL injection.

**Before:**
```python
def create_enum_type(self, apps, schema_editor):
    string_enum_values = ", ".join([f"'{value}'" for value in self.enum_values])
    cursor.execute(
        f"CREATE TYPE {self.enum_name} AS ENUM ({string_enum_values});"  # unsafe
    )

def drop_enum_type(self, apps, schema_editor):
    cursor.execute(f"DROP TYPE {self.enum_name};")  # unsafe
```

**After:**
```python
def create_enum_type(self, apps, schema_editor):
    cursor.execute(
        psycopg2_sql.SQL("CREATE TYPE {} AS ENUM ({})").format(
            psycopg2_sql.Identifier(self.enum_name),
            psycopg2_sql.SQL(", ").join(
                psycopg2_sql.Literal(v) for v in self.enum_values
            ),
        )
    )

def drop_enum_type(self, apps, schema_editor):
    cursor.execute(
        psycopg2_sql.SQL("DROP TYPE {}").format(
            psycopg2_sql.Identifier(self.enum_name)
        )
    )
```

`psycopg2.sql.Identifier` handles identifier quoting; `psycopg2.sql.Literal` handles value escaping. The `psycopg2` package is already a direct dependency used elsewhere in this module.

## Changes

- `api/src/backend/api/db_utils.py` — import `psycopg2.sql`; rewrite `create_enum_type` and `drop_enum_type`
- `api/src/backend/api/tests/test_db_utils.py` — adds `TestPostgresEnumMigration` with two tests that assert `cursor.execute` receives a `psycopg2.sql.Composable` object, not a plain string

## How to review

1. Confirm that `psycopg2_sql` is imported from the existing dep
2. Run the new tests in `TestPostgresEnumMigration` — both should pass
3. Verify no existing migration tests break

## Checklist

- [x] No new dependencies — `psycopg2` already used in this module
- [x] Minimal, focused diff


sandiyochristan:
## CI Failure Note

The `api-code-quality`, `api-security-scans`, and `api-tests` jobs are failing due to a structural limitation of fork-based PRs.

```
```

During CI, this is rewritten via `sed` to use the head branch name:
```
prowler @ git+https://github.com/prowler-cloud/prowler.git@fix/api-db-enum-sql-injection
```

Since this branch only exists in my fork (`sandiyochristan/prowler`), not in `prowler-cloud/prowler`, Poetry fails to resolve the dependency.

**This PR only touches API code** — `api/src/backend/api/db_utils.py` — and does not require any SDK changes. The fix and its test are correct; the CI infrastructure failure is unrelated to the change itself.

sandiyochristan:

```yaml
```

api/src/backend/api/tests/test_db_utils.py
```diff
@@ -848,3 +850,48 @@ def test_rls_transaction_cursor_yielded_correctly(self, tenants_fixture):
...
+    def test_create_uses_parameterized_identifier(self):
+        """create_enum_type passes a psycopg2_sql.Composable to cursor.execute."""
+        migration = PostgresEnumMigration("my_enum", ("val_a", "val_b"))
+        schema_editor, mock_cursor = self._make_mock_schema_editor()
+
+        migration.create_enum_type(apps=None, schema_editor=schema_editor)
+
+        mock_cursor.execute.assert_called_once()
+        query_arg = mock_cursor.execute.call_args[0][0]
+        assert isinstance(query_arg, psycopg2_sql.Composable), (
```
josema-xyz: These tests verify that `psycopg2.sql` is used, but not that the generated SQL is correct. You could render the SQL with `.as_string(connection)` and assert the actual output, like C`REATE TYPE "my_enum" AS ENUM ('val_a', 'val_b')`. That way the test covers both: the right library is used and the query is what you expect.

api/CHANGELOG.md
```diff
@@ -40,6 +40,7 @@ All notable changes to the **Prowler API** are documented in this file.
 
 - Bump `Pillow` to 12.1.1 (CVE-2021-25289) [(#10027)](https://github.com/prowler-cloud/prowler/pull/10027)
 - Remove safety ignore for CVE-2026-21226 (84420), fixed via `azure-core` 1.38.x [(#10110)](https://github.com/prowler-cloud/prowler/pull/10110)
+- Use `psycopg2.sql` to safely compose DDL in `PostgresEnumMigration`, preventing SQL injection via f-string interpolation [(#10166)](https://github.com/prowler-cloud/prowler/pull/10166)
```
josema-xyz: This changelog entry is deprecated, please get the latest changes from Prowlers' master branch.

api/src/backend/api/tests/test_db_utils.py
sandiyochristan: Good call — updated. The tests now inspect the composed SQL structure directly via `.seq`, `.strings`, and `._wrapped`:

```python
parts = query_arg.seq
assert parts[0] == psycopg2_sql.SQL("CREATE TYPE ")
assert isinstance(parts[1], psycopg2_sql.Identifier)
assert parts[1].strings == ("my_enum",)
# ...
enum_literals = [p for p in parts[3].seq if isinstance(p, psycopg2_sql.Literal)]
assert [lit._wrapped for lit in enum_literals] == ["val_a", "val_b"]
```

Note: I couldn't use `.as_string(connection)` because `psycopg2.sql.Identifier.as_string()` requires a real libpq connection (C extension), not a mock. Inspecting `.seq`/`.strings`/`._wrapped` achieves the same goal — verifying both that psycopg2.sql is used **and** that the generated query is correct.

api/CHANGELOG.md
sandiyochristan: Done — rebased on latest upstream master. The CHANGELOG now includes all current entries from master with our security fix entry on top.

api/pyproject.toml
```diff
@@ -24,7 +24,7 @@ dependencies = [
   "drf-spectacular-jsonapi==0.5.1",
   "gunicorn==23.0.0",
   "lxml==5.3.2",
-  "prowler @ git+https://github.com/prowler-cloud/prowler.git@master",
+  "prowler @ git+https://github.com/prowler-cloud/prowler.git@e3e2408717de9e93aff60ad51f46811f40020e2a",
```
AdriiiPRodri: We need to keep the reference to master here

```suggestion
  "prowler @ git+https://github.com/prowler-cloud/prowler.git@master",
```

AdriiiPRodri: I understand the issue you had with this now. In that case, we will probably create a branch and cherry-pick your work so we don’t lose the authorship, since we must avoid merging this change into the pyproject

josema-xyz (dismissed):

jfagoagas:
@AdriiiPRodri please review it and reconcile changelog. Thanks!

sandiyochristan (commented):
LGTM! Good to go from my end

AdriiiPRodri (approved):

josema-xyz (approved):

jfagoagas (approved):

PR #10167, opened 2026-02-25
feat(image): add image provider to UI

andoniaf:
### Context

Adds full UI support for the Image (Container Registry) provider, enabling users to connect and scan container images from registries like Docker Hub, Amazon ECR, and other OCI-compliant registries directly from the Prowler UI.

### Description

**UI** (`ui/`)
- New `ImageCredentialsForm` component for registry authentication (registry URL, repository, tag, username/password)
- Provider badge, icon, logo, and radio selector entries for the Image provider
- Credentials builder, field mappings, and Zod form schema with cross-field validation (e.g. username requires password)
- Integration into existing provider workflows: connect-account, base-credentials, custom-provider-inputs, and provider-type-selector
- Fix syntax errors in the credentials form schema ternary chain for all providers

**SDK** (`prowler/`)
- Enable misconfiguration scanner by default for image scans
- Route bare registry hostnames (e.g. `docker.io`) to OCI catalog in `test_connection`
- Isolate Trivy cache directory per provider instance to prevent conflicts in concurrent scans

### Steps to review

1. **UI flow**: Navigate to *Providers → Add Provider* and select "Container Image". Verify the credential form renders with fields for registry URL, repository, tag, and optional username/password.
2. **Form validation**: Submit the form with missing required fields and confirm validation errors appear. Verify that entering a username without a password (and vice versa) triggers cross-field validation.
3. **SDK changes**: Run an image scan and confirm Trivy cache isolation works correctly — each provider instance should use its own cache dir under `/tmp/`.
4. **Test connection**: Verify `test_connection` correctly resolves bare hostnames like `docker.io` via the OCI catalog endpoint.

### Checklist

<details>

</details>

#### SDK/CLI

#### UI

### License


andoniaf:
Adding `no-changelog` label — the SDK changes in this PR (Trivy cache isolation, misconfig scanner default, OCI catalog routing) are incremental improvements to the `image` provider which is still unreleased (tracked under the existing 5.19.0 UNRELEASED entries in `prowler/CHANGELOG.md`).

alejandrobailo (commented):
Hey! Nice work adding the Image provider to the UI — looks great overall. Left a couple of comments on things I noticed while reviewing.

prowler/providers/image/image_provider.py
```diff
@@ -946,32 +964,31 @@ def test_connection(
...
-                adapter.list_repositories()
-                return Connection(is_connected=True)
 
-            # Image reference mode — verify the specific tag exists
+            # Image reference → verify tag exists via registry API
             registry_host = ImageProvider._extract_registry(image)
-            repo_and_tag = image[len(registry_host) + 1 :] if registry_host else image
-            if ":" in repo_and_tag:
-                repository, tag = repo_and_tag.rsplit(":", 1)
+            is_dockerhub = registry_host is None
```
alejandrobailo: Hey! I think this might be a regression — the old code checked for explicit Docker Hub hostnames:

```python
is_dockerhub = not registry_host or registry_host in (
    "docker.io",
    "registry-1.docker.io",
)
```

With `registry_host is None`, images like `docker.io/alpine:3.18` or `registry-1.docker.io/nginx` will have a non-None `registry_host`, so `is_dockerhub` becomes `False`. That means the `library/` prefix won't be added for official images, and the tag lookup will likely 404.

Would it make sense to restore the explicit hostname check? Something like:

```python
is_dockerhub = registry_host is None or registry_host in ("docker.io", "registry-1.docker.io")
```

alejandrobailo:
One more thing — it looks like `ui/hooks/use-credentials-form.ts` is missing a `case "image"` in the `getDefaultValues()` switch ([line 96](https://github.com/prowler-cloud/prowler/blob/a2076048af773cfe8fe103080fab64ca215e5b70/ui/hooks/use-credentials-form.ts#L96)). Without it, the new credential fields (`registry_username`, `registry_password`, `registry_token`, `image_filter`, `tag_filter`) will initialize as `undefined` instead of `""`, which can cause React's controlled/uncontrolled input warnings and potentially break validation.

Every other provider has a case there, so adding one for `image` should do the trick!

ui/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ All notable changes to the **Prowler UI** are documented in this file.
 
 ### 🚀 Added
 
+- Image (Container Registry) provider support in UI: badge icon, credentials form, and provider-type filtering [(#10167)](https://github.com/prowler-cloud/prowler/pull/10167)
```
jfagoagas, line 9: Watch out, this is for v5.20, not v5.19.

prowler/providers/image/image_provider.py
andoniaf: Fixed, restored the explicit hostname check for "docker.io" and "registry-1.docker.io" in e8865a56d9.

ui/CHANGELOG.md
andoniaf: Moved to a new [1.21.0] (Prowler UNRELEASED) section in e8865a56d9.

alejandrobailo (changes requested):
### Code review

Found 2 issues:

1. New `ImageCredentialsForm` uses legacy `CustomInput` (HeroUI) instead of `WizardInputField` (shadcn). `ui/AGENTS.md` says "New/Existing UI? → shadcn/ui + Tailwind (NEVER HeroUI for new code)" and "`components/ui/` → HeroUI (LEGACY - do not add here)". Every other credential form in the same directory (`iac-credentials-form.tsx`, `k8s-credentials-form.tsx`, `mongodbatlas-credentials-form.tsx`, `openstack-credentials-form.tsx`) was migrated to `WizardInputField` in `b1c5fa4c4`. This new form re-introduces the legacy pattern.

https://github.com/prowler-cloud/prowler/blob/a44137fbf108589315374b8019fb3fb3b1b3e9e8/ui/components/providers/workflow/forms/via-credentials/image-credentials-form.tsx#L2-L4

2. `providerUid` schema for the image provider uses bare `z.string()` without `.trim().min(1)` validation. Every other provider with a user-supplied UID (AWS, Azure, GCP, Kubernetes, GitHub, IaC, OracleCloud, MongoDBAtlas, AlibabaCloud) uses `z.string().trim().min(1, "Provider ID is required")`. Since the Registry URL is a required field (placeholder text shows `"e.g. 123456789012.dkr.ecr.us-east-1.amazonaws.com"`), this allows submitting an empty string without client-side validation.

https://github.com/prowler-cloud/prowler/blob/a44137fbf108589315374b8019fb3fb3b1b3e9e8/ui/types/formSchemas.ts#L118-L122

Alan-TheGentleman (changes requested):
One real issue to fix:

1. **Cross-field validation is one-directional** — `formSchemas.ts` blocks submitting a password without a username, but the reverse isn't enforced: a user can submit `registry_username` without `registry_password` and the form passes validation. The PR description says \"username requires password\", so both directions should be guarded. Add the symmetric check inside the same `if (providerType === \"image\")` superRefine block.

Minor note: `tempfile.mkdtemp()` in `__init__` creates a dir that's only guaranteed to be cleaned up if `run_scan`/`run_scan_list` is called (both have `finally` -> `cleanup()`). If a provider instance is created and abandoned before any scan (e.g., in `test_connection`-only flows), the `/tmp/prowler-trivy-cache-*` dir leaks silently. Not blocking, but worth a `weakref.finalize` or `TemporaryDirectory` follow-up.

The SDK work is solid — Trivy cache isolation is correctly implemented, `_test_registry_connection` is well-extracted and testable, and the new tests cover the important paths. The `PROVIDER_BADGE_BY_NAME` rename is a good clarification too.

andoniaf:
> Minor note: `tempfile.mkdtemp()` in `__init__` creates a dir that's only guaranteed to be cleaned up if `run_scan`/`run_scan_list` is called (both have `finally` -> `cleanup()`). If a provider instance is created and abandoned before any scan (e.g., in `test_connection`-only flows), the `/tmp/prowler-trivy-cache-*` dir leaks silently. Not blocking, but worth a `weakref.finalize` or `TemporaryDirectory` follow-up.

Good catch! Switched `tempfile.mkdtemp()` to `tempfile.TemporaryDirectory` (which uses `weakref.finalize` internally) so the cache dir auto-cleans on GC even in test_connection-only flows. Also removed the now-unused `shutil` import.

alejandrobailo (approved):
LGTM!

PR #10298, opened 2026-03-10
feat(sdk): add universal compliance schema models and loaders

pedrooot:
## Summary
- Add all Pydantic v1 models for the universal compliance schema (`ComplianceFramework`, `UniversalComplianceRequirement`, `AttributeMetadata`, `TableConfig`, `PDFConfig`, `OutputsConfig`, etc.)
- Add loader functions: `load_compliance_framework_universal`, `get_bulk_compliance_frameworks_universal`, `adapt_legacy_to_universal`

## Test plan
- [ ] `poetry run pytest tests/lib/check/universal_compliance_models_test.py -v`

> Part 1/4 of PROWLER-683 chained PRs

### DIFF
--- a/prowler/lib/check/compliance_models.py
+++ b/prowler/lib/check/compliance_models.py
@@ -657,6 +657,14 @@ class ComplianceFramework(BaseModel):
                         f"Requirement '{req.id}': missing required attribute '{key}'"
                     )
 
+            # Unknown keys — anything outside the declared schema is a typo or drift
+            unknown_keys = set(attrs) - valid_keys
+            for key in sorted(unknown_keys):
+                errors.append(
+                    f"Requirement '{req.id}': unknown attribute '{key}' "
+                    f"(not declared in attributes_metadata)"
+                )
+
             # Enum validation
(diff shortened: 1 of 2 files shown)


Alan-TheGentleman (commented):
One important gap in the validator:

1. **Unknown attribute keys still pass schema validation.** `validate_attributes_against_metadata()` checks required keys, enums, and primitive types, but it never rejects `req.Attributes` keys that are not declared in `AttributesMetadata`. That means a typo like `Sectoin` will silently load, and downstream CSV/OCSF rendering will just miss the field. Add a check for `set(attrs) - valid_keys` and fail fast, otherwise the metadata contract is not actually protecting the JSONs.

The rest is heading in the right direction — especially the legacy adapter and the provider-aware bulk loader. But this validator needs to be strict, otherwise the whole universal schema becomes too forgiving to trust.

pedrooot:
thanks for the feedback @Alan-TheGentleman changes applied ✅

jfagoagas (approved):

PR #10299, opened 2026-03-10
feat(sdk): add universal compliance output modules (CSV, OCSF, table)

pedrooot:
## Summary
- `UniversalComplianceOutput` — CSV output with dynamic columns from `AttributeMetadata`
- `OCSFComplianceOutput` — OCSF `ComplianceFinding` (class_uid 2003) JSON output
- `get_universal_table` — CLI table renderer with configurable grouping via `TableConfig`
- Tests for all three modules (~1,359 lines)

## Test plan
- [ ] `poetry run pytest tests/lib/outputs/compliance/universal/ -v`

> Part 2/4 of PROWLER-683 chained PRs (base: PR 1)


Alan-TheGentleman (commented):
One thing here will bite us as soon as outputs are written in more than one batch:

1. **`OCSFComplianceOutput.batch_write_data_to_file()` is not append-safe.** The first call writes a valid JSON array and closes it with `]`, but a later call on the same file reopens in append mode and writes more objects after that closing bracket. The result is malformed JSON. If this class is meant to support batched/streaming writes, it needs newline-delimited JSON, a true incremental array writer, or a single writer instance reused across the whole run. Right now the method name says 'batch', but the file format only works reliably as single-shot.

The CSV/table/OCSF split is clean, and the metadata-driven columns are a strong improvement. But the file-writing contract needs to be nailed down before other PRs wire this into the pipeline.

jfagoagas (approved):

PR #10300, opened 2026-03-10
feat(sdk): add multi-provider compliance framework JSONs

pedrooot:
## Summary
- Add `prowler/compliance/csa_ccm_4.0.json` — CSA CCM 4.0 as multi-provider universal framework
- Update `get_available_compliance_frameworks()` to scan top-level `prowler/compliance/` directory

## Test plan
- [ ] `poetry run python prowler-cli.py aws --list-compliance | grep csa_ccm_4.0`
- [ ] Verify JSON schema matches `ComplianceFramework` model

> Part 3/4 of PROWLER-683 chained PRs (base: PR 2)


Alan-TheGentleman (commented):
One thing I would change before merging:\n\n1. ** should not live in the production compliance directory.** Because it sits under , it becomes discoverable through  and looks like a real runnable framework to end users. Templates/examples are useful, but they should live under docs, fixtures, or a dedicated examples path — not next to supported frameworks. Otherwise we ship internal scaffolding as product surface area.\n\nThe CSA CCM data itself looks fine as the real deliverable. I would just move the generic template out of the runtime discovery path so the CLI stays clean and intentional.

Alan-TheGentleman (commented):
One thing I would change before merging:

1. **`generic_compliance.json` should not live in the production compliance directory.** Because it sits under `prowler/compliance/`, it becomes discoverable through `--list-compliance` and looks like a real runnable framework to end users. Templates/examples are useful, but they should live under docs, fixtures, or a dedicated examples path — not next to supported frameworks. Otherwise we ship internal scaffolding as product surface area.

The CSA CCM data itself looks fine as the real deliverable. I would just move the generic template out of the runtime discovery path so the CLI stays clean and intentional.

tests/lib/outputs/compliance/universal/fixtures/generic_compliance.json
jfagoagas, line 1: If this is for testing, let's move it under `tests` for compliance.

pedrooot: Done!

jfagoagas (dismissed):

Alan-TheGentleman:
I pushed the fix locally on my side: top-level universal compliance JSONs are now filtered by real provider support before they get exposed through --list-compliance, and I added a targeted test so  shows up for AWS but not for Kubernetes. I also added the missing SDK changelog entry. One blocker on my side: the commit is currently being stopped by unrelated repo hooks in , so the branch update is not pushed yet.

Alan-TheGentleman:
I applied the fix locally on my side. Top-level universal compliance JSONs are now filtered by real provider support before they get exposed through `--list-compliance`, and I added a targeted test so `csa_ccm_4.0` shows up for AWS but not for Kubernetes. I also added the missing SDK changelog entry.

One blocker on my side: the commit is currently being stopped by unrelated repo hooks under `.claude/worktrees`, so the branch update is not pushed yet.

jfagoagas (approved):
Neat!

PR #10301, opened 2026-03-10
feat(sdk): integrate universal compliance into CLI pipeline

pedrooot:
## Summary
- Load universal frameworks at startup, pass to `load_checks_to_execute()`
- Fix `--list-compliance` / `--list-compliance-requirements` to show universal frameworks
- Extract `process_universal_compliance_frameworks()` — replaces 9 identical per-provider blocks (~260 lines removed)
- Always generate OCSF for universal frameworks (no `--output-formats` gate)
- Fix `display_compliance_table` dispatch: `in` → `startswith` (prevents false matches like `cisa` → `cis`)
- Handle dict-style `Checks` and `Provider=None` in requirement listing
- Tests for dispatch logic (245 lines) and process function (601 lines)

## Test plan
- [ ] `poetry run pytest tests/lib/outputs/compliance/display_compliance_table_test.py -v`
- [ ] `poetry run pytest tests/lib/outputs/compliance/process_universal_test.py -v`
- [ ] `poetry run python prowler-cli.p--list-compliance | grep csa_ccm`
- [ ] `poetry run python prowler-cli.py aws --compliance csa_ccm_4.0 --list-compliance-requirements`

> Part 4/4 of PROWLER-683 chained PRs (base: PR 3)


Alan-TheGentleman (commented):
Two things to fix in the integration layer:

1. **`process_universal_compliance_frameworks()` is called inside `streaming_callback`, so it creates new CSV/OCSF output objects on every batch.** That means the same framework gets appended to `generated_outputs['compliance']` multiple times during a single run, which is not what the rest of the pipeline expects. Initialize or reuse one output writer per framework for the whole execution instead of recreating them per batch.

2. **This PR makes the OCSF append bug reachable in real runs.** `OCSFComplianceOutput` writes a closed JSON array each time, and a second batch appends more objects after the closing bracket. So once a scan emits multiple batches, the generated `.ocsf.json` becomes invalid. Even if the writer bug is fixed in the previous PR, this integration needs to be structured around persistent writers, not one-shot writers per callback.

The `startswith` dispatch cleanup is a very good catch — that part is definitely better than the old substring matching. But the streaming integration needs one more pass so outputs stay valid on real, multi-batch scans.

prowler/__main__.py
```diff
@@ -633,6 +644,19 @@ def streaming_callback(findings_batch):
     input_compliance_frameworks = set(output_options.output_modes).intersection(
         bulk_compliance_frameworks.keys()
     )
```
josema-xyz, line 647: I think this drops the new universal-only frameworks. `csa_ccm_4.0` only lives in `universal_frameworks`, not in `bulk_compliance_frameworks`, so the intersection comes back empty for it and process_universal_compliance_frameworks never sees the name.

Could we union both dicts here? `set(bulk_compliance_frameworks.keys()) | set(universal_frameworks.keys())`.                                           

Also, the comment right above ends mid-sentence, `using it as the source produced`, worth finishing it once the filter is fixed.

pedrooot: Totally, fixed thanks for the review

josema-xyz (approved):

HugoPBrito (approved):

PR #10332, opened 2026-03-12
fix(azure): create distinct report per key/secret in keyvault checks

kagahd:
### Context

6 Azure Key Vault checks share the same bug: a single `Check_Report_Azure` object is created per vault and reused across all keys/secrets/diagnostic settings in the inner loop. Since Python appends by reference, all findings for a vault end up pointing to the same object — showing only the **last** item's status and name. This produces duplicate/incorrect findings in reports and hides real security issues.

Additionally, `keyvault_key_rotation_enabled` only checked `lifetime_actions[0]` instead of scanning all lifetime actions for a `Rotate` policy.

### Description

**Bug 1 — Shared report object (6 checks):**
Each of the following checks created one `Check_Report_Azure` per vault and overwrote its `status`/`status_extended` in the inner loop, then appended the same reference multiple times to `findings`:
- `keyvault_key_rotation_enabled`
- `keyvault_key_expiration_set_in_non_rbac`
- `keyvault_rbac_key_expiration_set`
- `keyvault_non_rbac_secret_expiration_set`
- `keyvault_rbac_secret_expiration_set`
- `keyvault_logging_enabled`

**Fix:** A new `Check_Report_Azure` is now created **per item** (key, secret, or diagnostic setting) inside the inner loop.

**Bug 2 — `[0]` index access (`keyvault_key_rotation_enabled`):**
Only `lifetime_actions[0].action` was checked. If the `Rotate` action was at index 1 or later, the check incorrectly reported FAIL.

**Fix:** Replaced with `any(action.action == "Rotate" for action in ...)`.

**Tests:**
- Existing unit tests updated to match the new per-item reporting semantics.
- New multi-item unit tests added for all 6 checks, verifying distinct objects, correct count, and correct PASS/FAIL per item.

### Steps to review

1. Review each of the 6 check files — the core change is moving `report = Check_Report_Azure(...)` from the vault loop into the inner item loop.
2. For `keyvault_key_rotation_enabled`, also verify the `any()` change in the rotation policy condition.
3. Review the test files — each check now has a multi-item test case (e.g. `test_multiple_keys_mixed_expiration`, `test_multiple_diagnostic_settings_per_vault`).
4. Run `pytest tests/providers/azure/services/keyvault/ -v` — all 45 tests should pass.

### Checklist

<details>

</details>

#### SDK/CLI

#### UI
- N/A (no UI changes)

#### API
- N/A (no API changes)

### License


prowler/providers/azure/services/keyvault/keyvault_rbac_secret_expiration_set/keyvault_rbac_secret_expiration_set.py
```diff
@@ -5,21 +5,19 @@
...
-
         for subscription, key_vaults in keyvault_client.key_vaults.items():
             for keyvault in key_vaults:
-                if keyvault.properties.enable_rbac_authorization and keyvault.secrets:
-                    for secret in keyvault.secrets:
+                if keyvault.properties.enable_rbac_authorization:
+                    for secret in keyvault.secrets or []:
                         report = Check_Report_Azure(
-                            metadata=self.metadata(), resource=secret
+                            metadata=self.metadata(), resource=keyvault
```
HugoPBrito: Now that this emits one finding per secret, I think the report should stay bound to `secret`, not `keyvault`. Otherwise `resource_name` / `resource_id` will identify the vault for every finding even though the audited entity is the secret.

```suggestion
                        report = Check_Report_Azure(
                            metadata=self.metadata(), resource=secret
                        )
```

prowler/providers/azure/services/keyvault/keyvault_non_rbac_secret_expiration_set/keyvault_non_rbac_secret_expiration_set.py
```diff
@@ -7,23 +7,17 @@ def execute(self) -> Check_Report_Azure:
...
-                    )
-                    report.subscription = subscription
-                    report.status = "PASS"
-                    report.status_extended = f"Keyvault {keyvault.name} from subscription {subscription} has all the secrets with expiration date set."
-                    has_secret_without_expiration = False
-                    for secret in keyvault.secrets:
+                if not keyvault.properties.enable_rbac_authorization:
+                    for secret in keyvault.secrets or []:
+                        report = Check_Report_Azure(
+                            metadata=self.metadata(), resource=keyvault
```
HugoPBrito: Same here: once the finding becomes secret-level, I think `Check_Report_Azure` should be instantiated with `resource=secret` so the structured identity matches the audited object.

```suggestion
                        report = Check_Report_Azure(
                            metadata=self.metadata(), resource=secret
                        )
```

prowler/providers/azure/services/keyvault/keyvault_rbac_key_expiration_set/keyvault_rbac_key_expiration_set.py
```diff
@@ -7,20 +7,17 @@ def execute(self) -> Check_Report_Azure:
...
-                    )
-                    report.subscription = subscription
-                    report.status = "PASS"
-                    report.status_extended = f"Keyvault {keyvault.name} from subscription {subscription} has all the keys with expiration date set."
-                    has_key_without_expiration = False
-                    for key in keyvault.keys:
+                if keyvault.properties.enable_rbac_authorization:
+                    for key in keyvault.keys or []:
+                        report = Check_Report_Azure(
+                            metadata=self.metadata(), resource=keyvault
```
HugoPBrito: This is now key-level reporting, so I would bind the report to `key` instead of `keyvault`. That keeps `resource_name` / `resource_id` aligned with the actual audited entity.

```suggestion
                        report = Check_Report_Azure(
                            metadata=self.metadata(), resource=key
                        )
```

prowler/providers/azure/services/keyvault/keyvault_key_expiration_set_in_non_rbac/keyvault_key_expiration_set_in_non_rbac.py
```diff
@@ -7,20 +7,17 @@ def execute(self) -> Check_Report_Azure:
...
-                    )
-                    report.subscription = subscription
-                    report.status = "PASS"
-                    report.status_extended = f"Keyvault {keyvault.name} from subscription {subscription} has all the keys with expiration date set."
-                    has_key_without_expiration = False
-                    for key in keyvault.keys:
+                if not keyvault.properties.enable_rbac_authorization:
+                    for key in keyvault.keys or []:
+                        report = Check_Report_Azure(
+                            metadata=self.metadata(), resource=keyvault
```
HugoPBrito: Since this now produces one finding per key, I think the report should be created with `resource=key` rather than `resource=keyvault`, so the structured fields identify the audited key.

```suggestion
                        report = Check_Report_Azure(
                            metadata=self.metadata(), resource=key
                        )
```

prowler/providers/azure/services/keyvault/keyvault_key_rotation_enabled/keyvault_key_rotation_enabled.py
```diff
@@ -7,23 +7,23 @@ def execute(self) -> Check_Report_Azure:
         findings = []
         for subscription, key_vaults in keyvault_client.key_vaults.items():
             for keyvault in key_vaults:
-                if keyvault.keys:
+                for key in keyvault.keys or []:
                     report = Check_Report_Azure(
                         metadata=self.metadata(), resource=keyvault
```
HugoPBrito: The `any(... == "Rotate")` fix looks good. I would just make the report key-level as well by binding it to `key`, otherwise multiple findings from the same vault still share the same structured identity.

```suggestion
                    report = Check_Report_Azure(
                        metadata=self.metadata(), resource=key
                    )
```

prowler/providers/azure/services/keyvault/keyvault_logging_enabled/keyvault_logging_enabled.py
```diff
@@ -8,14 +8,20 @@ def execute(self) -> Check_Report_Azure:
...
+                        metadata=self.metadata(), resource=keyvault
+                    )
+                    report.subscription = subscription_name
                     report.status = "FAIL"
                     report.status_extended = f"There are no diagnostic settings capturing audit logs for Key Vault {keyvault.name} in subscription {subscription_name}."
                     findings.append(report)
                 else:
                     for diagnostic_setting in keyvault.monitor_diagnostic_settings:
+                        report = Check_Report_Azure(
+                            metadata=self.metadata(), resource=keyvault
```
HugoPBrito: Creating a fresh report per diagnostic setting fixes the shared-object bug. Since the finding is now effectively diagnostic-setting-level, I think it would be worth checking whether the structured resource identity, metadata/resource type, and `status_extended` wording should also consistently reflect that audited object instead of keeping the vault as the primary subject.

HugoPBrito (changes requested):
Thanks for working on this. The per-item reporting direction is the right one, but I think this PR still needs one more consistency pass before it can be merged.

This change is denser than it looks, because it is not only moving report creation into the inner loop. By switching these checks from vault-level findings to key/secret/diagnostic-setting-level findings, the PR also changes the effective audited entity, and that has consequences beyond the Python logic:

- Metadata should be reviewed so titles, descriptions, remediation text, and resource types align with the actual subject of the finding.
- `status_extended` should follow the same model: once a finding is item-level, the audited object should become the grammatical subject, with the Key Vault kept as context.
- Tests should validate that behavior consistently at the structured finding level, not only through the human-readable message.

I would address those consistency updates across the whole PR together with the inline resource-binding fixes below.

HugoPBrito (changes requested):
Thanks @kagahd!

I found what I believe will be the last bug before approval.

Disabled `keys/secrets` should be **skipped** entirely instead of being reported as `PASS`.

Right now these checks only fail when the item is both missing an expiration and `enabled=True`, but everything else falls into the else branch and gets reported as `PASS`. That means a disabled `key/secret` with no expiration still produces a finding saying it "has an expiration date set", which is not accurate.

  Since disabled items are effectively out of scope for this validation, they should not generate findings at all. Please update all corresponding key/secret expiration checks so that, before building/appending a report, they first check whether the `key/secret` is disabled and, if so, continue to the next item.

  In other words:

1. `enabled=False` -> skip, no finding
2.  `enabled=True` and no expiration -> `FAIL`
3.  `enabled=True` and expiration set -> `PASS`

  Please apply this consistently across all affected checks, not just one of them.

Thanks again for all the effort put into this!

kagahd:
Hi @HugoPBrito, thanks for the feedback. I’ve implemented your suggestions. Please take a look and let me know if anything is missing or needs adjustment.

HugoPBrito:
Hi @kagahd,

The PR looks good overall. All the other requested changes appear to be addressed.

What still looks inconsistent is `keyvault_logging_enabled`.

Its metadata and check title still describe a vault-level control, but the implementation now emits one finding per diagnostic setting. That means a vault with one valid diagnostic setting and one invalid one can still produce a `FAIL`, even though the current control wording says the vault should pass if it has a diagnostic setting capturing audit logs.

I believe this should be resolved keeping the **vault-level** approach, since it's defined that way on the check's mapped CIS controls:

  - Stay vault-level, it should emit a single finding per Key Vault:
    - `PASS` when at least one diagnostic setting is compliant
    - `FAIL` otherwise

 From my side, this is the only remaining blocker I see before approval.

HugoPBrito:
Thank you very much for your hard work @kagahd! It's a pleasure to review your contributions.

HugoPBrito (approved):

PR #10335, opened 2026-03-13
feat(ec2): add check for SG ingress from public IPs to any port

### LINKED ISSUE #6231: New Check - ec2_securitygroup_allow_ingress_from_internet_to_any_port_from_ip
### New feature motivation

Due to a trend where developers add their personal home IP addresses to SGs for various connectivity, I want to know which Security Groups have these explicit holes in them. Especially if the developer no longer should have access to the resource.
Also, if an attacker gets hold of AWS credentials and wants to open ingress to an EC2 instance (or publicly accessible RDS or other resource), they are likely to use a single/small range of IPs, potentially on an obscure port.

### Solution Proposed

An expansion of the `ec2_securitygroup_allow_ingress_from_internet_to_any_port` check to include any non-internal IP addresses, that way getting the SG itself checked and monitored.
Or could build off the ec2_instance_port_*_exposed_to_internet rule, check for _any_ port exposed to an internet IP address. 
Due to the nature of these kind of rules, it would likely be a medium or low severity finding.

N/A?

### Additional context

_No response_

raajheshkannaa:
Closes #6231

## Summary

- New check: `ec2_securitygroup_allow_ingress_from_internet_to_any_port_from_ip`
- Flags security groups with ingress rules from any globally routable IP to any port/protocol
- Catches cases where developers add specific public IPs directly (not just 0.0.0.0/0 or ::/0)
- Uses `check_security_group(any_address=False, all_ports=True)` with `ipaddress.is_global`
- Emits PASS with explanation when the all_ports check already failed (matches existing pattern)

## Test plan

- [x] 7 test cases covering: default SGs, specific public IPv4, private IP, specific port, all_ports dedup, unused SGs, public IPv6
- [x] Follows same structure as port-22/port-3389/port-3306 checks
- [x] Metadata follows prowler conventions


danibarranqueroo (approved):
Thanks for that! Really good job on this new check 🚀

PR #10347, opened 2026-03-16
fix(oci): use home region for audit configuration API call

### LINKED ISSUE #10209: [OCI] Audit log retention check fails with 404 error due to incorrect API path/endpoint
### Issue search

Prowler API

Not applicable

### Steps to Reproduce

I am encountering a FAILED result for the check: Tenancy audit log retention period is 365 days or greater. Upon checking the Prowler logs, it appears that Prowler is attempting to call an incorrect OCI API endpoint, resulting in a 404 Not Found error. This causes the check to fail even when the actual OCI configuration is set correctly (365 days or more).

To Reproduce:

1. Run Prowler against an OCI environment.

2. Observe the check: audit_log_retention_period_365_days.

3. Check the debug logs to see the API request path.

4. Prowler hits a non-existent path/endpoint, returning a 404 error.

### Expected behavior

Prowler should correctly call the Audit Configuration API (e.g., GET /audit/configuration) using the valid regional endpoint or the home region to retrieve the retention-period-days value.

I use Prowler opensource on enterprise environment, so I couldn't offer result with screenshots or Logs.

### Environment Resource

1. EC2 instance 
2. I search the log prowler gets with Splunk - Then I found the error 404. I guess the path prowler goes is incorrect. 

### OS used

1. Amazon Linux 2023

### Prowler version

5.13.1

### Python version

3.12

### Pip version

pip3

### Context

_No response_

Ker102:
## Summary

Fixes #10209

The OCI audit log retention check (`audit_log_retention_period_365_days`) fails with a **404 error** because the `AuditClient` is created without specifying the home region. The OCI Audit Configuration API (`get_configuration`) is a tenancy-level resource that must be queried from the home region endpoint.

## Root Cause

In `audit_service.py`, `_create_oci_client(oci.audit.AuditClient)` was called without any `config_overrides`, so the client used the session's default region. If that region's endpoint doesn't serve the audit configuration API, the call returns 404.

## Fix

Pass `config_overrides={"region": home_region}` to `_create_oci_client()`, where `home_region` is sourced from `self.provider.identity.region`. This is the same home region source that the Identity service already uses for its tenancy-level API calls.

### Changes

| File | Change |
|------|--------|
| `prowler/providers/oraclecloud/services/audit/audit_service.py` | Pass home region via `config_overrides` when creating `AuditClient` |
| `tests/providers/oraclecloud/services/audit/audit_service_test.py` | Add tests verifying home region is used and error handling still defaults to 90 days |

### Before
```python
audit_client = self._create_oci_client(oci.audit.AuditClient)
```

### After
```python
home_region = self.provider.identity.region
audit_client = self._create_oci_client(
    oci.audit.AuditClient,
    config_overrides={"region": home_region},
)
```

## Testing

- ✅ New test: `test_get_configuration_uses_home_region` — verifies `_create_oci_client` is called with the home region config override
- ✅ New test: `test_get_configuration_handles_api_error` — verifies error handling still defaults to 90-day retention
- ✅ Existing test preserved: `test_service`


Ker102:
Hi @jfagoagas  👋 Just checking in on this — I noticed the status/waiting-for-revision label was added last week, but I don't see any review comments or requested changes. Is there something specific you'd like me to update, or is this still in the review queue? Happy to make any adjustments needed. Thanks!

jfagoagas:
> Hi @jfagoagas 👋 Just checking in on this — I noticed the status/waiting-for-revision label was added last week, but I don't see any review comments or requested changes. Is there something specific you'd like me to update, or is this still in the review queue? Happy to make any adjustments needed. Thanks!

Hi @Ker102 We'll review it as soon as possible. Thanks!

HugoPBrito (changes requested):
Thanks for the fix, but I believe the root issue is still there: `provider.identity.region` is just the configured session region, not the tenancy home region. 

`OraclecloudProvider.set_identity` sets that value from the current config (`prowler/providers/oraclecloud/oraclecloud_provider.py#L447`), so when you run Prowler from a non‑home region the override still points at that wrong endpoint and `get_configuration` keeps hitting a `404`.

The only reliable source of the home region is from `identity.list_region_subscriptions(...).is_home_region`, so we need to capture the region where `is_home_region=True` (or persist it on the provider) and use that when creating `AuditClient`.

The new test should cover `configured region ≠ home region` to prove it actually fixes the reported failure.

Please let me know what you think, and thanks again for contributing.

Ker102:
Thanks for the catch — you were right that my first pass was still using the configured session region.

I updated this so it now resolves the actual tenancy home region from `list_region_subscriptions(...).is_home_region` and stores it separately from the configured region. The audit service uses that home region when creating `AuditClient`, and I added a regression test for the case where the configured region and home region are different.

I also reran the Oracle Cloud tests after the change.

HugoPBrito:
Hi @Ker102,

Sorry, I didn't catch up with this.

prowler/providers/oraclecloud/oraclecloud_provider.py
```diff
@@ -495,11 +495,30 @@ def set_identity(
...
+            home_region = next(
+                (
+                    region_subscription.region_name
+                    for region_subscription in region_subscriptions
+                    if getattr(region_subscription, "is_home_region", False)
+                ),
+                region,
+            )
+        except Exception as error:
+            logger.warning(f"Could not determine tenancy home region: {error}")
```
HugoPBrito: If `list_region_subscriptions` raises here, `home_region` silently falls back to the configured `region` — which is exactly the broken state we are fixing. Either fail fast with a clear error, or log at `error`/`warning` with a message that explicitly mentions the audit configuration check will likely 404, so it's debuggable in the wild.

prowler/providers/oraclecloud/oraclecloud_provider.py
```diff
@@ -495,11 +495,30 @@ def set_identity(
...
+            ).data
+            home_region = next(
+                (
+                    region_subscription.region_name
+                    for region_subscription in region_subscriptions
+                    if getattr(region_subscription, "is_home_region", False)
+                ),
+                region,
+            )
+        except Exception as error:
```
HugoPBrito: `except Exception` is too broad. Please narrow to `oci.exceptions.ServiceError` (and `ClientError` if applicable) so unexpected errors aren't swallowed.

prowler/providers/oraclecloud/oraclecloud_provider.py
```diff
@@ -495,11 +495,30 @@ def set_identity(
         logger.info(f"OCI User ID: {user_id}")
         logger.info(f"OCI Region: {region}")
 
+        home_region = region
+        try:
+            region_subscriptions = identity_client.list_region_subscriptions(
```
HugoPBrito: `list_region_subscriptions` is already called in `get_regions_to_audit`, which also reads `is_home_region`. We're now paying for two round trips per run. Could you cache the subscription list on `self._identity` (or the session) and reuse it across both code paths?

tests/providers/oraclecloud/services/audit/audit_service_test.py
```diff
@@ -26,3 +29,67 @@ def test_service(self):
...
+            return_value=None,
+        ):
+            from prowler.providers.oraclecloud.services.audit.audit_service import Audit
+
+            audit = Audit(oraclecloud_provider)
+            audit.provider = oraclecloud_provider
+            audit.audited_tenancy = OCI_TENANCY_ID
+            audit.session_config = oraclecloud_provider.session.config
+            audit.session_signer = oraclecloud_provider.session.signer
+            audit.configuration = None
```
HugoPBrito: Patching `Audit.__init__` to a no-op and then manually setting `audited_tenancy`, `session_config`, `session_signer`, `configuration`, etc. is brittle — any new field added to the constructor will silently break or, worse, mask a real regression. Please let the real `__init__` run and patch only `_create_oci_client` / the underlying SDK call so the test stays honest about the constructor contract.

prowler/providers/oraclecloud/services/audit/audit_service.py
```diff
@@ -19,9 +19,15 @@ def __init__(self, provider):
     def __get_configuration__(self):
         """Get Audit configuration."""
         try:
-            audit_client = self._create_oci_client(oci.audit.AuditClient)
+            home_region = self.provider.identity.home_region
```
HugoPBrito: The `Tools: Check Changelog` job is failing — please add an entry under the appropriate unreleased section in `prowler/CHANGELOG.md` for this fix so the check passes.

HugoPBrito (changes requested):
Please review my comments.

HugoPBrito:
Hi @Ker102,

Do you plan to continue this development?

Ker102:
@HugoPBrito Sorry didnt see the reply, will take this on tomorrow, this PR been open for a while now.

HugoPBrito:
> @HugoPBrito Sorry didnt see the reply, will take this on tomorrow, this PR been open for a while now.

No hurries, I just wanted to know the current status. Thanks!

prowler/providers/oraclecloud/oraclecloud_provider.py
Ker102: Fixed in `bc8d5d557`. The fallback warning now calls out the audit config 404 risk instead of hiding it.

prowler/providers/oraclecloud/oraclecloud_provider.py
Ker102: Fixed in `bc8d5d557`. This now only catches OCI `ServiceError` / `ClientError` there.

prowler/providers/oraclecloud/oraclecloud_provider.py
Ker102: Fixed in `bc8d5d557`. Region subscriptions are cached on the identity and reused by `get_regions_to_audit`.

tests/providers/oraclecloud/services/audit/audit_service_test.py
Ker102: Fixed in `bc8d5d557`. The audit tests now let the real constructor run and patch `_create_oci_client` only.

prowler/providers/oraclecloud/services/audit/audit_service.py
Ker102: Added in `bc8d5d557`.

Ker102:
Pushed the review fixes and merged current master so the branch is no longer conflicting.\n\nLocal check: pytest tests/providers/oraclecloud -q -> 199 passed, 2 skipped.

HugoPBrito:
Thanks @Ker102! Could you please update the branch and the changelog? It should be for `v5.27.2`.

HugoPBrito:
Hi @Ker102, thanks again for this fix and for addressing all the earlier feedback.
To help this land in the next release, I pushed two commits.

Thanks for all the effort you've put into this! 🚀

HugoPBrito (dismissed):

HugoPBrito:
I slimmed this PR down to the minimum still needed.

Since this was opened, #10529 merged and already added a `home_region` property to `OraclecloudProvider` (derived from `is_home_region`), and updated the identity service to use it. That overlaps with the provider-side changes here (the `home_region`/`region_subscriptions` fields on `OCIIdentityInfo`, the `set_identity` lookup, the `region_subscriptions` caching and the `__init__` rework).

To avoid two parallel home-region mechanisms, I reverted those files back to `master` and kept only the part that #10529 did **not** cover and that actually fixes #10209: the Audit service now creates `AuditClient` against the tenancy home region via the existing `provider.home_region` property and `config_overrides`.

Net change is now just `audit_service.py`, its tests, and the CHANGELOG entry (moved to `5.28.0`). OCI tests pass locally (195 passed, 2 skipped). Thanks again @Ker102.

HugoPBrito:
Quick follow-up to my previous comment: I had to keep one small provider-side change after all.

#10529's `home_region` is derived from the `--region`-filtered audit set, so when scanning a region that isn't the tenancy home region, `home_region` resolved to that filtered (non-home) region. Since the Audit configuration is a tenancy-level resource that must be queried from the home region, this made the call `404` and fall back to 90 days, which flipped the retention check result.

Fix: `home_region` is now resolved from the **full** subscription list (independent of the `--region` filter), so tenancy-level APIs always target the real home region. Added `test_home_region_uses_full_subscription_list_not_region_filter` to cover the non-home-region scan.

Net change is now `audit_service.py` + `oraclecloud_provider.py` (home region resolution) + their tests + the CHANGELOG entry. OCI tests pass locally (196 passed, 2 skipped).

HugoPBrito (approved):

Ker102:
Thank you alot for helping finish up this PR, I appreciate the collaboration, I am currently abroad so apologize for the delay!

PR #10361, opened 2026-03-17
fix(api): make CORS_ALLOWED_ORIGINS configurable via env variable

### LINKED ISSUE #10344: Make CORS_ALLOWED_ORIGINS configurable via environment variable
### Feature search

Prowler API

Not provider-specific

### New feature motivation

When deploying Prowler with the UI and API on separate hosts (e.g., `prowler.example.com` for UI and `prowler-api.example.com` for API), browser requests from the UI are blocked by CORS. This is because `CORS_ALLOWED_ORIGINS` is **hardcoded** to `["http://localhost", "http://127.0.0.1"]` in [`config/django/base.py`](https://github.com/prowler-cloud/prowler/blob/e9860f700293b4fcfbf0e401abb186ee42a8bc38/api/src/backend/config/django/base.py#L65) and is never overridden in `production.py`.

There is currently **no way to configure allowed CORS origins** without modifying the application source code. This is a common deployment scenario — the Helm chart creates separate Ingress resources for the UI and API, and Kubernetes deployments typically use different hostnames or subdomains for each service.

Related: #8516 — describes the same root cause affecting reverse proxy deployments.

### Solution Proposed

Add an environment variable override for `CORS_ALLOWED_ORIGINS` in `config/django/production.py`, following the same pattern already used for [`ALLOWED_HOSTS`](https://github.com/prowler-cloud/prowler/blob/e9860f700293b4fcfbf0e401abb186ee42a8bc38/api/src/backend/config/django/production.py#L5):

```python
CORS_ALLOWED_ORIGINS = env.list(
    "DJANGO_CORS_ALLOWED_ORIGINS",
    default=["http://localhost", "http://127.0.0.1"]
)
```

This is a one-line change. `django-environ`'s `env.list()` parses comma-separated values, so users can specify multiple origins: `DJANGO_CORS_ALLOWED_ORIGINS=https://prowler.example.com,https://prowler.internal.local`.

The [Helm chart](https://artifacthub.io/packages/helm/prowler-app/prowler) can then automatically set this value in the API ConfigMap based on the UI ingress host, similar to how `AUTH_URL` is already derived from `ui.ingress.hosts`.

### Use case and benefits

- **Kubernetes/Helm deployments** where UI and API are exposed via separate Ingress resources on different hostnames
- **Reverse proxy setups** (nginx, Traefik, Caddy) where the UI and API are behind different virtual hosts
- **Any deployment** where the UI origin differs from the API origin — which is the default when using the official Helm chart with both ingresses enabled

Without this fix, users must either work around CORS at the ingress/proxy level (adding CORS headers via nginx annotations) or deploy a custom nginx container as described in #8516.

1. **Ingress-level CORS headers** — Adding `nginx.ingress.kubernetes.io/enable-cors` annotations to the API Ingress. This works but is a workaround that depends on the specific ingress controller and may conflict with Django's CORS middleware (duplicate headers).
2. **Single-origin ingress with path-based routing** — Serving both UI and API from the same host (`/api` → API, `/` → UI). Eliminates CORS entirely but limits deployment flexibility and doesn't match the Helm chart's current two-ingress design.
3. **Setting `CORS_ALLOW_ALL_ORIGINS = True`** — Insecure, not suitable for production.

### Additional context

The pattern for the fix already exists in the codebase. `ALLOWED_HOSTS` is configured via environment variable in `production.py`:

```python
ALLOWED_HOSTS = env.list("DJANGO_ALLOWED_HOSTS", default=["localhost", "127.0.0.1"])
```

The proposed change for `CORS_ALLOWED_ORIGINS` follows the exact same pattern. The [Helm chart](https://artifacthub.io/packages/helm/prowler-app/prowler) already derives `NEXT_PUBLIC_API_BASE_URL` from the API ingress host, so adding `DJANGO_CORS_ALLOWED_ORIGINS` to the API ConfigMap based on `ui.ingress.hosts` would be a natural extension.

raajheshkannaa:
Closes #10344

## Summary

- Added `DJANGO_CORS_ALLOWED_ORIGINS` env var support in `production.py`
- Uses `env.list()` with same defaults as `base.py` (`http://localhost`, `http://127.0.0.1`)
- Follows the existing pattern used by `DJANGO_ALLOWED_HOSTS` on the line above

## Usage

```bash
DJANGO_CORS_ALLOWED_ORIGINS=https://prowler.example.com,https://prowler.internal.local
```

## Test plan

- [ ] Verify default behavior unchanged when env var is not set
- [ ] Set `DJANGO_CORS_ALLOWED_ORIGINS` to a custom origin, confirm CORS headers are returned
- [ ] Confirm `ALLOWED_HOSTS` and `CORS_ALLOWED_ORIGINS` can be configured independently

### DIFF
--- a/api/src/backend/config/django/production.py
+++ b/api/src/backend/config/django/production.py
@@ -3,6 +3,10 @@ from config.env import env
 
 DEBUG = env.bool("DJANGO_DEBUG", default=False)
 ALLOWED_HOSTS = env.list("DJANGO_ALLOWED_HOSTS", default=["localhost", "127.0.0.1"])
+CORS_ALLOWED_ORIGINS = env.list(
+    "DJANGO_CORS_ALLOWED_ORIGINS",
+    default=["http://localhost", "http://127.0.0.1"],
+)
 
 # Database
 # TODO Use Django database routers https://docs.djangoproject.com/en/5.0/topics/db/multi-db/#automatic-database-routing


jfagoagas:
Hi @raajheshkannaa the issue https://github.com/prowler-cloud/prowler/issues/10344 was closed this morning as implemented in https://github.com/prowler-cloud/prowler/pull/10355. 

Thank you for the contribution but it was done by another community member. I'm closing this PR for that reason.

PR #10374, opened 2026-03-18
fix(api): rewrite rls_transaction to retry mid-query replica failures with primary fallback

josema-xyz:
_**EDIT**_: Superseded by #10379. Closing this in favor of a simpler approach that fixes the same bug without changing any call sites. The new PR uses Django's `execute_wrapper` API inside `rls_transaction` to intercept mid-query `OperationalError` on the replica, retry with backoff, and fall back to primary. Same `with rls_transaction(...)` interface, ~80 lines changed in one file instead of 2400 across 35.

### Context

When the read replica dies mid-query, the retry and primary fallback logic in `rls_transaction` never executes. The function is a `@contextmanager` generator that can only yield once. After yielding the cursor to the caller, any error hits a guard that re-raises immediately — the retry loop is unreachable.

Also, `REPLICA_MAX_ATTEMPTS=3` only gives 2 replica tries because the primary fallback consumes one.

### Description

Rewrites `rls_transaction` from a `@contextmanager` to a class with an iterator protocol. All call sites change from with `rls_transaction(...)` to `for attempt in rls_transaction(...): with attempt:`. The for loop re-executes the entire body on failure, so both connection-setup errors and mid-query errors are retried.

#### Changes:
- `rls_transaction` is now a class that yields `_RLSAttempt` context managers
- `_RLSAttempt.__enter__` retries connection-setup failures inline
- `_RLSAttempt.__exit__` catches mid-query failures and lets the loop continue
- `REPLICA_MAX_ATTEMPTS=3` now means 3 replica tries + 1 primary fallback
- Deleted the hand-rolled retry loop in `integrations.py`
- Migrated all call sites and test mocks to the new `for/with` pattern, over 100 of each

### Steps to review

Deep look at:
- Core rewrite: `api/src/backend/api/db_utils.py`
- Special cases: `renderers.py` (conditional), `integrations.py` (deleted manual retry), `scan.py` (deadlock loop)
- New tests in `test_db_utils.py` for mid-body retry and max attempts semantics

Run tests and use the application locally.

### Checklist

#### API

### License

### DIFF
--- a/api/CHANGELOG.md
+++ b/api/CHANGELOG.md
@@ -17,6 +17,7 @@ All notable changes to the **Prowler API** are documented in this file.
 ### 🐞 Fixed
 
 - Attack Paths: Recover `graph_data_ready` flag when scan fails during graph swap, preventing query endpoints from staying blocked until the next successful scan [(#10354)](https://github.com/prowler-cloud/prowler/pull/10354)
+- Rewrite `rls_transaction` to retry mid-query read replica failures with primary DB fallback, fixing scan crashes during RDS replica recovery [(#10374)](https://github.com/prowler-cloud/prowler/pull/10374)
 
 ### 🔐 Security
 
--- a/api/src/backend/api/adapters.py
+++ b/api/src/backend/api/adapters.py
@@ -45,26 +45,27 @@ class ProwlerSocialAccountAdapter(DefaultSocialAccountAdapter):
                 tenant = Tenant.objects.using(MainRouter.admin_db).create(
                     name=f"{user.email.split('@')[0]} default tenant"
                 )
-                with rls_transaction(str(tenant.id)):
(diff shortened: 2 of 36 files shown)


josema-xyz:
Closing this in favor of #10379. After reviewing the approach I realized the `for`/`with` rewrite at every call site is overkill for what's fundamentally a query-level retry problem.

The replacement PR keeps `rls_transaction` as a context manager and uses `connection.execute_wrappers` to catch `OperationalError` during `cursor.execute()` on the replica. It retries on replica with backoff, then falls back to primary, transparently, no call-site changes. Also fixes the off-by-one in `REPLICA_MAX_ATTEMPTS` and closes stale connections between retries.

The `.iterator()` limitation (server-side cursor fetches via `fetchmany()`) exists in both approaches, neither can retry mid-iteration without risking duplicate rows.

PR #10388, opened 2026-03-19
feat(ui): add batch apply pattern to Findings filters

Alan-TheGentleman:
### Context

https://github.com/user-attachments/assets/c1b9b6f5-375e-4910-bbcc-28fa6d037cf4

Implements [PROWLER-1228](https://prowlerpro.atlassian.net/browse/PROWLER-1228) — Findings Filters Batch Apply with Selection Summary.

Currently, every filter selection in the Findings view triggers an immediate API call to the backend. This causes excessive API calls when configuring multiple filters, poor UX as the table reloads on every change, and unnecessary backend load for intermediate filter states.

### Description

Introduces a **batch apply** pattern for the Findings view filters: users select all desired filters freely without triggering API calls, see a summary of pending selections, and explicitly apply them with a single "Apply Filters" button.

**New components:**
- `useFilterBatch` hook — manages two-state model (pending vs applied/URL) with `setPending`, `applyAll`, `discardAll`, `clearAll`, `hasChanges`, `changeCount`, `getFilterValue`
- `FilterSummaryStrip` — horizontal strip below filter bar showing pending selections as removable chips with human-readable labels
- `ApplyFiltersButton` — Apply/Discard button pair with change detection

**Modified components (backward compatible):**
- `DataTableFilterCustom` — new `mode` prop (`"instant"` | `"batch"`, default `"instant"`) for opt-in batch behavior. All other views (Scans, Providers, Compliance, Users) continue using instant mode unchanged
- `ProviderTypeSelector` / `AccountsSelector` — new optional `onBatchChange` + `selectedValues` props for batch participation. Overview page keeps instant-apply
- `ClearFiltersButton` — new `onClear` + `pendingCount` props for batch-aware clearing
- `CustomDatePicker` / `CustomCheckboxMutedFindings` — batch support via optional props

**Removed:**
- Cross-filter dependency logic between provider type and accounts (auto-clearing)
- Scan option narrowing via `useRelatedFilters` in Findings view
- All filters now show their full option set independently

**Tests:** 47 new unit tests (175 total, all passing)

### Steps to review

1. Navigate to Findings view
2. Select multiple filters (severity, status, provider, etc.) — verify NO API calls are made
3. Verify the summary strip shows pending selections with readable labels (e.g., `Provider: AWS`, `Severity: Critical`)
4. Click "Apply Filters" — verify single API call with all filters
5. Verify Apply button is disabled when no changes are pending
6. Click "Clear Filters" — verify all pending selections are cleared (including provider/account)
7. Click Discard (X) — verify pending state resets to last applied
8. Reload page — verify filters are preserved from URL params
9. Navigate to other views (Scans, Providers, Compliance) — verify they still use instant-apply

### Checklist

<details>

</details>

#### UI (if applicable)

### License

[PROWLER-1228]: https://prowlerpro.atlassian.net/browse/PROWLER-1228?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ui/app/(prowler)/_overview/_components/accounts-selector.tsx
```diff
@@ -48,58 +48,51 @@ const PROVIDER_ICON: Record<ProviderType, ReactNode> = {
...
   providers: ProviderProps[];
+  /**
+   * When provided, called instead of navigating immediately.
+   * Use this on pages that batch filter changes (e.g. Findings).
+   * The Overview page omits this prop to keep instant-apply behavior.
+   *
+   * @param filterKey - The raw filter key without "filter[]" wrapper, e.g. "provider_id__in"
+   * @param values - The selected values array
+   */
+  onBatchChange?: (filterKey: string, values: string[]) => void;
```
pfe-nazaries: You could improve the API making a discriminated union between the "two modes" of usage (batched or not) to clarify which props are really needed an avoid coginitive overload and DX.

The same could be applied for other functions like this one with batch mode

ui/components/filters/apply-filters-button.tsx
```diff
@@ -0,0 +1,67 @@
...
+ * - Uses Prowler's shadcn `Button` component.
+ */
+export const ApplyFiltersButton = ({
+  hasChanges,
+  changeCount,
+  onApply,
+  onDiscard,
+  className,
+}: ApplyFiltersButtonProps) => {
+  const label = hasChanges
```
pfe-nazaries: You can just leave as 

```typescript 
 const label = changeCount > 0
      ? `Apply Filters (${changeCount})`
      : "Apply Filters";
```

ui/components/filters/custom-date-picker.tsx
```diff
@@ -14,22 +14,70 @@ import {
...
-    } catch {
-      return undefined;
+  // In instant mode, we need local state to track the selected date so the
+  // calendar stays in sync when URL params change externally (e.g. Clear Filters).
+  // In batch mode, `valueProp` is the source of truth — derive date directly.
+  const [localDate, setLocalDate] = useState<Date | undefined>(() =>
+    parseDate(searchParams.get("filter[inserted_at]")),
+  );
+
+  // In batch mode: derive the displayed date from the controlled prop.
```
pfe-nazaries: localDate is derived state — it mirrors searchParams.get("filter[inserted_at]") and
  needs a useEffect to stay in sync. This is a classic React anti-pattern ("state that syncs
   with props/external source via effect").

Consider deriving the date directly:

```typescript
const date = valueProp !== undefined
  ▎     ? parseDate(valueProp)
  ▎     : parseDate(searchParams.get("filter[inserted_at]"));
```

ui/components/filters/index.ts
```diff
@@ -1,6 +1,8 @@
+export * from "./apply-filters-button";
```
pfe-nazaries, line 1: FYI: In general i would avoid barrel files. They avoid proper tree shaking and make extra changes needed to work

ui/components/findings/findings-filters.tsx
```diff
@@ -32,10 +35,78 @@ interface FindingsFiltersProps {
   uniqueGroups: string[];
 }
 
+/**
+ * Maps raw filter param keys (e.g. "filter[severity__in]") to human-readable labels.
+ * Used to render chips in the FilterSummaryStrip.
+ */
+const FILTER_KEY_LABELS: Record<string, string> = {
```
pfe-nazaries: You can improve the key type of the object with two approaches depending on how hard you want to restrict the types:

```typescript
type FilterKey = `filter[${string}]`;
```

```typescript
type FilterParam =
    | "provider_type__in"
    | "provider_id__in"
    | "severity__in"
    | "status__in"
    | "delta__in"
    | "region__in"
    | "service__in"
    | "resource_type__in"
    | "category__in"
    | "resource_groups__in"
    | "scan__in"
    | "inserted_at"
    | "muted";

  type FilterKey = `filter[${FilterParam}]`;
```

ui/app/(prowler)/_overview/_components/accounts-selector.tsx
alejandrobailo: Agreed. Right now `onBatchChange` and `selectedValues` are independent optional props, but they're semantically coupled (you always need both or neither). A discriminated union would enforce that at the type level and make the API self-documenting. This same pattern should apply to `ProviderTypeSelector`, `CustomDatePicker`, and `CustomCheckboxMutedFindings` all of them have the same `onBatchChange` + controlled-value pair that should be all-or-nothing.

ui/components/filters/apply-filters-button.tsx
alejandrobailo: Agreed, the simpler ternary is cleaner and more readable here. No need for the extra indirection.

ui/components/filters/custom-date-picker.tsx
alejandrobailo: Good catch on the derived state pattern. In batch mode this is partially mitigated since `valueProp` acts as the controlled source of truth from the parent. But for instant mode the `useEffect` sync is still there and is indeed the classic anti-pattern.

Deriving the date directly (as suggested) would be the cleanest approach for the instant-mode path, and in batch mode the `valueProp` already takes precedence, so both cases would be covered.

ui/components/filters/index.ts
alejandrobailo: Fair point in general, though worth noting the barrel file already existed before this PR. The change here just adds 2 new exports to the existing pattern. Probably not something to address in this PR specifically, but good to keep in mind for future refactors.

ui/components/findings/findings-filters.tsx
alejandrobailo: Agreed, the second approach with the explicit `FilterParam` union is especially nice. It would catch typos at compile time and make the supported filter set self-documenting. This aligns well with the AGENTS.md typing guidelines too.

alejandrobailo (changes requested):
### Code Review

Nice work on the overall architecture. The `useFilterBatch` hook design with pending/applied separation is clean, the tests are thorough (588 lines for the hook alone with great edge case coverage), and backward compatibility with instant mode is well preserved.

That said, there are a few issues that should be addressed before merging:

---

#### 1. `PROVIDER_TYPE_DISPLAY` duplicates the existing `PROVIDER_DISPLAY_NAMES`

`findings-filters.tsx` introduces a new `PROVIDER_TYPE_DISPLAY` mapping, but [`ui/types/providers.ts`](https://github.com/prowler-cloud/prowler/blob/957439030c9773b407fa5c4eaca9b1a429067b22/ui/types/providers.ts#L20-L35) already exports `PROVIDER_DISPLAY_NAMES` and `getProviderDisplayName()`.

Worse, the values **diverge**:
- PR: `gcp: "GCP"` / `iac: "IaC"`
- Existing: `gcp: "Google Cloud"` / `iac: "Infrastructure as Code"`

This will cause the filter summary chips to show different names than the rest of the app (provider selectors, badges, etc.).

**Fix:** Use `getProviderDisplayName()` from `@/types/providers` instead of the local map.

https://github.com/prowler-cloud/prowler/blob/957439030c9773b407fa5c4eaca9b1a429067b22/ui/components/findings/findings-filters.tsx#L62-L81

---

#### 2. `STATUS_DISPLAY` duplicates existing status patterns

`findings-filters.tsx` introduces a `STATUS_DISPLAY` map (`PASS → "Pass"`, etc.), but status values and display logic already exist in:
- [`FINDING_STATUS`](https://github.com/prowler-cloud/prowler/blob/957439030c9773b407fa5c4eaca9b1a429067b22/ui/types/components.ts#L90-L96) in `ui/types/components.ts`
- [`FindingStatusValues`](https://github.com/prowler-cloud/prowler/blob/957439030c9773b407fa5c4eaca9b1a429067b22/ui/components/ui/table/status-finding-badge.tsx#L5-L10) in `ui/components/ui/table/status-finding-badge.tsx`

For consistency, a centralized `STATUS_DISPLAY_NAMES` should be created (similar to how [`SEVERITY_DISPLAY_NAMES`](https://github.com/prowler-cloud/prowler/blob/957439030c9773b407fa5c4eaca9b1a429067b22/ui/types/severities.ts#L11-L17) works in `ui/types/severities.ts`) and reused here.

https://github.com/prowler-cloud/prowler/blob/957439030c9773b407fa5c4eaca9b1a429067b22/ui/components/findings/findings-filters.tsx#L55-L59

---

#### 3. `formatFilterValue` ignores existing formatting utilities

The `formatFilterValue` function does a generic "capitalize first letter" fallback, but this produces incorrect results for many filter values (e.g., `"iam"` → `"Iam"` instead of `"IAM"`).

The codebase already has proper formatting for each filter type:
- `getProviderDisplayName()` for provider types (`@/types/providers`)
- `SEVERITY_DISPLAY_NAMES` for severities (`@/types/severities`) — **not handled at all currently**
- `getCategoryLabel()` / `getGroupLabel()` for categories and groups (`@/lib/categories`)
- `formatLabel()` for generic delimiter-based formatting (`@/lib/categories`)

The function should dispatch to these existing utilities per filter key, and only fall back to `formatLabel()` for unknown keys.

https://github.com/prowler-cloud/prowler/blob/957439030c9773b407fa5c4eaca9b1a429067b22/ui/components/findings/findings-filters.tsx#L88-L100

---

#### 4. AccountsSelector loses contextual information (UX regression)

The PR removes the dynamic description and empty-state messages:

```diff
-  const filterDescription =
-    selectedTypesList.length > 0
-      ? `Showing accounts for ${selectedTypesList.join(", ")} providers`
-      : "All connected cloud provider accounts";
+  const filterDescription = "All connected cloud provider accounts";
```

```diff
-  {selectedTypesList.length > 0
-    ? "No accounts available for selected providers"
-    : "No connected accounts available"}
+  No connected accounts available
```

Even in batch mode, the selector has access to `selectedValues` — it could derive which provider types are selected and show context-appropriate messages. Users losing this feedback is a UX regression.

https://github.com/prowler-cloud/prowler/blob/957439030c9773b407fa5c4eaca9b1a429067b22/ui/app/(prowler)/_overview/_components/accounts-selector.tsx#L108-L111

---

#### 5. Misleading comment in `applyAll` about `filter[muted]`

In `use-filter-batch.ts`, the comment in `applyAll` says:

> "Start from the current URL params to preserve non-batch params (search, muted, etc.)"

But `filter[muted]` **is** now batch-managed (it's not in `EXCLUDED_FROM_BATCH`). The "etc." is also misleading since only `filter[search]` is excluded. This will confuse future developers about which filters are batch-managed vs excluded.

https://github.com/prowler-cloud/prowler/blob/957439030c9773b407fa5c4eaca9b1a429067b22/ui/hooks/use-filter-batch.ts#L161

---

#### 6. Minor: `FilterSummaryStrip` should use `aria-live` for dynamic updates

The `role="region"` is good, but adding `aria-live="polite"` would ensure screen readers announce when chips are added or removed. Without it, the dynamic chip changes are invisible to assistive technology.

https://github.com/prowler-cloud/prowler/blob/957439030c9773b407fa5c4eaca9b1a429067b22/ui/components/filters/filter-summary-strip.tsx#L48-L51

Alan-TheGentleman:
@alejandrobailo @pfe-nazaries I addressed all requested review points in the latest commits ( and ).\n\n✅ Done:\n- Replaced local provider/status label maps with shared utilities\n- Updated filter value formatting to use existing helpers (, , category/group formatters)\n- Restored contextual account selector messaging without reintroducing dependent filtering\n- Fixed misleading  comment about excluded params\n- Added  to summary strip for accessibility\n\nAlso reran lint/tests for touched areas and everything is green.\n\nCould you please take another look? Thanks!

Alan-TheGentleman:
@alejandrobailo @pfe-nazaries quick follow-up (previous comment had formatting issues from shell escaping):\n\nI addressed all requested review points in the latest commits.\n\nDone:\n- Replaced local provider/status label maps with shared utilities\n- Updated filter value formatting to use existing helpers from shared modules\n- Restored contextual account selector messaging without reintroducing dependent filtering\n- Fixed misleading applyAll comment about excluded params\n- Added aria-live="polite" to summary strip for accessibility\n\nI also reran lint/tests for touched areas and everything is green.\n\nCould you please re-review when you have a moment? Thanks!

Alan-TheGentleman:
@pfe-nazaries totally agree in general — I also avoid global barrel files because they can hurt tree-shaking and make ownership less explicit. In this PR I kept the existing domain-scoped barrel under ui/components/filters (it already existed before this change) and only added exports for the new filter components.\n\nSo my approach is: use barrel files only where they are truly needed and make sense (shared domain folders), and avoid broad/global barrels. If you want, we can align on a stricter project-wide rule in a separate refactor.

Alan-TheGentleman:
@alejandrobailo @pfe-nazaries quick update: I pushed another pass addressing the remaining review-thread items. Done in latest changes: 

- Enforced all-or-nothing batch props with discriminated unions in AccountsSelector, ProviderTypeSelector, CustomDatePicker, and CustomCheckboxMutedFindings
- Tightened findings filter key typing with an explicit FilterParam union and strict key record typing
- Simplified ApplyFiltersButton label logic
- Clarified apply

All comment about excluded params 

Also handled minor UX polish:
- hide default muted false chip
- exclude default muted from pending count
- keep inserted_at chip display as ISO date

Lint and tests are green on touched areas.

Could you please re-review when you have a moment? Thanks!

ui/app/(prowler)/_overview/_components/accounts-selector.tsx
Alan-TheGentleman: Great point — fixed in commit 8604215b9. I updated AccountsSelector, ProviderTypeSelector, CustomDatePicker, and CustomCheckboxMutedFindings to use discriminated unions so batch props are all-or-nothing (onBatchChange + controlled value together, or neither in instant mode).

ui/components/filters/apply-filters-button.tsx
Alan-TheGentleman: Applied, thanks. Simplified label generation as suggested to a single ternary in commit 8604215b9: changeCount > 0 ? Apply Filters (N) : Apply Filters.

ui/components/filters/custom-date-picker.tsx
Alan-TheGentleman: Implemented in commit 8604215b9. CustomDatePicker now uses discriminated props for batch/instant mode, and the instant-mode path was simplified while keeping URL-sync behavior intact. Appreciate the callout.

ui/components/filters/index.ts
Alan-TheGentleman: Makes sense. I kept the existing domain-level barrel in this PR for consistency (it already existed), and I documented the preference in the updated skills to avoid broad/global barrels going forward. Thanks for the note.

ui/components/findings/findings-filters.tsx
Alan-TheGentleman: Implemented as requested in commit 8604215b9. I introduced an explicit FilterParam union in ui/types/filters.ts and typed FILTER_KEY_LABELS as Record<FilterParam, string> in findings-filters.tsx to enforce compile-time key safety.

Alan-TheGentleman:
All inline review threads have now been addressed and resolved with follow-up replies (including the remaining type-safety and API-shape items). Ready for final re-review when you have a moment 🙌

alejandrobailo (changes requested):
### Code review (re-review)

10 of 11 previous review points are resolved. Nice work on the discriminated unions, centralized display maps, and `FilterParam` typing.

1 item remains partially unresolved:

1. `CustomDatePicker` instant-mode derived-state anti-pattern (Pablo's original flag) — batch mode correctly derives date from `valueProp`, but instant mode still uses `localDate` state synced via `useEffect([searchParams])`. The date could be derived directly from `searchParams.get("filter[inserted_at]")` without local state or effect.

https://github.com/prowler-cloud/prowler/blob/cca5c52a15fea4947d3b40bd0b290a89b9f46a86/ui/components/filters/custom-date-picker.tsx#L69-L80

pfe-nazaries (approved):

pfe-nazaries (approved):

PR #10398, opened 2026-03-19
fix(azure): normalize MySQL configuration values to uppercase

tejas0077:
Azure MySQL Flexible Server configuration values returned from the API 
can be mixed case (e.g. "ON", "on", "On"). This causes security checks 
to return incorrect results when comparing configuration values against 
expected strings like "ON" or "OFF".

Added `.upper()` normalization to configuration.value in mysql_service.py.
None values are handled safely with a conditional check.

Before:
value=configuration.value,

After:
value=configuration.value.upper() if configuration.value else configuration.value,

No new dependencies required.
Are there new checks included? No

1. Review the one-line change in:
   prowler/providers/azure/services/mysql/mysql_service.py

2. Run existing tests to verify:
   python -m pytest tests/providers/azure/services/mysql/ -v

3. All 26 existing tests pass with the fix applied.

### DIFF
--- a/prowler/CHANGELOG.md
+++ b/prowler/CHANGELOG.md
@@ -959,3 +959,4 @@ All notable changes to the **Prowler SDK** are documented in this file.
 - Restore packages location in PyProject [(#7510)](https://github.com/prowler-cloud/prowler/pull/7510)
 
 ---
+- fix(azure): normalize MySQL configuration values to uppercase in mysql_service.py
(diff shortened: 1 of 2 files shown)


HugoPBrito:
As this PR was not linked with the required issue, this was not seen until now.
Closing as duplicate in favor of already merged https://github.com/prowler-cloud/prowler/pull/10396.

PR #10419, opened 2026-03-23
fix(api): align finding-group latest aggregation

AdriiiPRodri:
### Context

The `finding-groups/latest` endpoint was aggregating by the latest timestamp per check globally, which could undercount impacted resources when different providers had different latest snapshots for the same check.

### Description

- Select the latest summary row per `(check_id, provider)` before re-aggregating in `GET /api/v1/finding-groups/latest`.
- Keep aggregation by `check_id` so totals reflect all affected providers at their latest snapshot.
- Add regression coverage to verify cross-provider totals (`20 + 7 = 27`) for the same check.
- Keep mute historical processing chained with finding-group summary re-aggregation after rule creation.

### Steps to review

1. Run `poetry run pytest src/backend/api/tests/test_views.py -k \"finding_groups_latest and not resources\" -v` from `api/`.
2. Verify the new test `test_finding_groups_latest_aggregates_latest_per_provider` passes.
3. Validate `GET /api/v1/finding-groups/latest?filter[check_id]=...` now reflects latest-per-provider aggregation.

### Checklist

<details>

</details>

#### API (if applicable)

### License

### DIFF
--- a/api/src/backend/api/tests/test_views.py
+++ b/api/src/backend/api/tests/test_views.py
@@ -45,6 +45,7 @@ from api.models import (
     ComplianceRequirementOverview,
     DailySeveritySummary,
     Finding,
+    FindingGroupDailySummary,
     Integration,
     Invitation,
     LighthouseProviderConfiguration,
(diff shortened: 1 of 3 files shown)


api/src/backend/api/v1/views.py
```diff
@@ -6725,10 +6727,25 @@ def create(self, request, *args, **kwargs):
         )
 
         # Launch background task for historical muting
-        with transaction.atomic():
-            mute_historical_findings_task.apply_async(
-                kwargs={"tenant_id": tenant_id, "mute_rule_id": str(mute_rule.id)}
-            )
+        latest_scan_id = (
```
josema-xyz, line 6730: What about there are no scans? `latest_scan_id` will be `None` then. Is that a real case here?

api/src/backend/api/tests/test_views.py
```diff
@@ -14730,8 +14737,14 @@ def test_mute_rules_create_valid(
...
+        # Verify background task chain was called
+        mock_mute_signature.assert_called_once()
+        mock_aggregate_signature.assert_called_once()
+        mock_chain.assert_called_once_with(
+            mock_mute_signature.return_value,
+            mock_aggregate_signature.return_value,
+        )
+        mock_chain.return_value.apply_async.assert_called_once()
 
     @patch("tasks.tasks.mute_historical_findings_task.apply_async")
```
josema-xyz, line 14749: ~`tasks.tasks.mute_historical_findings_task.apply_async` mocks are dead code now.~

api/src/backend/api/v1/views.py
AdriiiPRodri: This is already handled in the aggregation task, and at this point there must be a scan since there are findings

josema-xyz (approved):

PR #10428, opened 2026-03-23
fix(api): support finding-group aggregated filters

AdriiiPRodri:
### Context

`/api/v1/finding-groups` and `/api/v1/finding-groups/latest` were not consistently applying several filters users expect at group level (especially computed and finding-level filters). This caused mismatches between requested filters and returned groups.

### Description

- Keep group aggregation on `FindingGroupDailySummary` for performance and apply computed filters (`status`, `status__in`, `severity`, `severity__in`) on aggregated rows.
- Add finding-level prefiltering by `check_id` for advanced dimensions (`delta`, `scan`, `region`, `service`, `category`, `resource_groups`, `resource_type`) so only matching groups are returned.
- Align latest group aggregation to use the latest row per `(check_id, provider)` and then aggregate by `check_id`.
- Add `check_title` support for sorting and for finding-level filtering paths.
- Extend tests in `TestFindingGroupViewSet` to cover latest/list behavior and the filter matrix.

### Steps to review

1. Run:
   - `cd api`
   - `poetry run pytest src/backend/api/tests/test_views.py -k "TestFindingGroupViewSet and not resources and not latest_resources" -v`
2. Verify filter behavior in:
   - `GET /api/v1/finding-groups`
   - `GET /api/v1/finding-groups/latest`
3. Check examples:
   - `filter[status]=FAIL`
   - `filter[severity]=critical`
   - `filter[region]=eu-west-1`
   - `filter[service]=s3`
   - `filter[scan]=<scan_uuid>`
   - `sort=check_title`

### Checklist

<details>

</details>

#### API (if applicable)

### License


api/src/backend/api/v1/views.py
```diff
@@ -7166,6 +7239,17 @@ def list(self, request, *args, **kwargs):
 
         # Re-aggregate daily summaries across the date range
         aggregated_queryset = self._aggregate_daily_summaries(filtered_queryset)
+        aggregated_queryset = self._apply_aggregated_computed_filters(
```
josema-xyz: I think there is a filter-order issue here. We apply status/severity before narrowing by finding-level filters (region/service/delta), so combined filters can return aggregated status that does not represent the filtered subset.
I say this because the result of the aggregated metrics could change, but if not, we can pass of this.

api/src/backend/api/tests/test_views.py
```diff
@@ -15245,6 +15245,85 @@ def test_finding_groups_status_muted_all(
         # rds_encryption has all muted findings
         assert data[0]["attributes"]["status"] == "MUTED"
 
+    def test_finding_groups_status_filter(
```
josema-xyz, line 15270: I would also add combined-filter tests (region+status, delta+severity, etc.) and error-validation tests (invalid status/severity, invalid __in). That would harden the most fragile part of this change.

josema-xyz (commented):
Two small and quick comments.

api/src/backend/api/v1/views.py
AdriiiPRodri: Yes, I’m aware of this, but for a first phase I thought it was fine since fixing it would require either modifying the aggregation table or querying findings and hurting performance. Thinking about it again, now that we have the apply filters button in the UI, I think we can safely query findings without issues, so I’ll change it. Thanks!

AdriiiPRodri:
> ```python
> check_title__icontains = CharFilter(field_name="check_title", lookup_expr="icontains")
> ```

That's not correct. I'm updating the PR, but we don’t always use CheckTitle written that way, and that's also a "known issue" we have in findings. It can be solved with this new table, but we'll need a hybrid approach between findings and the new aggregation table

josema-xyz (approved):

PR #10471, opened 2026-03-25
chore(ui): add pnpm supply chain security protections

pfe-nazaries:
## Summary

- Add `pnpm-workspace.yaml` with layered supply chain protections
  per https://pnpm.io/supply-chain-security
- Bump pnpm from 10.24.0 to 10.33.0 (required for
  `blockExoticSubdeps` and `trustPolicy`)
- Fix CI workflows to read exact pnpm version from
  `ui/package.json` instead of floating `version: 10`

## Supply chain protections added

| Protection | Setting | What it does |
  |---|---|---|
| Minimum release age | `minimumReleaseAge: 1440` | Blocks
  packages published < 24h ago |
| Build script allow-list | `onlyBuiltDependencies` | Only 6
  explicitly trusted packages may run postinstall scripts |
| Trust policy | `trustPolicy: no-downgrade` | Fails install if a
  package's trust evidence degrades (e.g., account takeover) |
| Block exotic subdeps | `blockExoticSubdeps: true` | Prevents
  transitive deps from using git/tarball URLs |

## CI fix

`pnpm/action-setup` was configured with `version: 10`, which
installs the latest 10.x ignoring the pinned version in
`packageManager`. Since `package.json` lives in `ui/` (not repo
root), the action never found it. Now uses `package_json_file:
  ui/package.json` to read the exact SHA512-verified version.

## Test plan

- [x] Clean install (`rm -rf node_modules && pnpm install`) — no
  trust violations or exotic subdep issues
- [x] All 6 allowed packages run build scripts successfully
- [x] 123/123 unit tests pass
- [x] Healthcheck (typecheck + ESLint) passes
- [x] Production build succeeds
- [ ] CI workflows pass with `package_json_file: ui/package.json`

### DIFF
--- a/.github/workflows/ui-e2e-tests-v2.yml
+++ b/.github/workflows/ui-e2e-tests-v2.yml
@@ -165,7 +165,7 @@ jobs:
       - name: Setup pnpm
         uses: pnpm/action-setup@41ff72655975bd51cab0327fa583b6e92b6d3061 # v4.2.0
         with:
-          version: 10
+          package_json_file: ui/package.json
           run_install: false
 
       - name: Get pnpm store directory
--- a/.github/workflows/ui-tests.yml
+++ b/.github/workflows/ui-tests.yml
@@ -98,7 +98,7 @@ jobs:
         if: steps.check-changes.outputs.any_changed == 'true'
         uses: pnpm/action-setup@41ff72655975bd51cab0327fa583b6e92b6d3061 # v4.2.0
         with:
-          version: 10
+          package_json_file: ui/package.json
           run_install: false
 
       - name: Get pnpm store directory
--- a/ui/Dockerfile
+++ b/ui/Dockerfile
@@ -13,7 +13,7 @@ RUN apk add --no-cache libc6-compat
 WORKDIR /app
 
 # Install dependencies based on the preferred package manager
-COPY package.json pnpm-lock.yaml .npmrc ./
+COPY package.json pnpm-lock.yaml pnpm-workspace.yaml .npmrc ./
 COPY scripts ./scripts
 RUN pnpm install --frozen-lockfile
 
--- a/ui/Dockerfile
+++ b/ui/Dockerfile
@@ -3,7 +3,7 @@ FROM node:24.13.0-alpine@sha256:cd6fb7efa6490f039f3471a189214d5f548c11df1ff9e5b1
 LABEL maintainer="https://github.com/prowler-cloud"
 
 # Enable corepack for pnpm
-RUN corepack enable && corepack prepare pnpm@10 --activate
+RUN corepack enable
 
 # Install dependencies only when needed
 FROM base AS deps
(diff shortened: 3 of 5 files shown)


alejandrobailo (dismissed):
LGTM!

jfagoagas (dismissed):

Alan-TheGentleman (dismissed):
Clean, well-scoped security hardening. The layered approach (release age + build allowlist + trust policy + exotic subdep blocking) is solid, and pinning pnpm via `packageManager` + `package_json_file` in CI is the right fix — floating `version: 10` was quietly defeating the SHA512 pin.

One thing worth a follow-up (not blocking): the `onlyBuiltDependencies` allowlist has `@heroui/shared-utils` and `msw` — both feel unusual for needing postinstall scripts. Worth confirming those are actually required, not just leftover from a previous audit.

Ship it.

Alan-TheGentleman (commented):
Quick question on the `onlyBuiltDependencies` allowlist: `@heroui/shared-utils` and `msw` look unusual at first glance. Do you know why those two need postinstall scripts?

pfe-nazaries:
> Quick question on the `onlyBuiltDependencies` allowlist: `@heroui/shared-utils` and `msw` look unusual at first glance. Do you know why those two need postinstall scripts?

Good point there. I will take a look at heroui (i think we can remove it). Msw is expected to be there

pfe-nazaries:
> Quick question on the `onlyBuiltDependencies` allowlist: `@heroui/shared-utils` and `msw` look unusual at first glance. Do you know why those two need postinstall scripts?

Apparently, HeroUI needs to inject different scripts depending on the React and Next.js version during build time. This is annoying from their side but necessary until we get rid of HeroUI

Alan-TheGentleman (commented):
Three things to fix before merge:

1. **The builder stage doesn't use the pinned pnpm.** `corepack install` runs in `deps`, but `builder` inherits from `base` — which after this PR only has `corepack enable`. So `pnpm run build` resolves pnpm from corepack's base shim, not the `package.json` pin. Version pinning breaks at the build stage.

2. **`onlyBuiltDependencies` has entries that don't belong.** `msw` isn't in `dependencies` or `devDependencies` — dead entry. For the transitive ones (`@heroui/shared-utils`, `unrs-resolver`, `@sentry/cli`), add a comment per entry explaining why it needs a build script. I verified `@heroui/shared-utils` is legit (detects React/Next version, copies demi bundles), but the others aren't obvious from the code alone.

3. **No escape hatch for emergency patches.** `minimumReleaseAge: 1440` blocks packages published in the last 24h — including critical CVE patches. If a zero-day drops in `next` or `react`, the team can't install the fix until the next day. Add `minimumReleaseAgeExclude` with at least the most critical packages.

alejandrobailo (approved):
### Code review

Alan's three blocking items are addressed in the latest commits:
1. Builder stage now runs `corepack install` — pinned pnpm in all build stages.
2. `msw` removed from `onlyBuiltDependencies`, each entry has an explanatory comment.
3. `minimumReleaseAgeExclude` documented as opt-in mechanism — reasonable operational choice.

Alan-TheGentleman (approved):
All three blocking items from my previous review are resolved:

1. Builder stage now runs `corepack install` — pinned pnpm across all Docker stages.
2. `msw` removed from `onlyBuiltDependencies`, each entry has an explanatory comment.
3. `minimumReleaseAgeExclude` documented as opt-in escape hatch for emergency patches.

Clean supply chain hardening, 4 layered protections, CI fixed to read exact pnpm version. Ship it.

pedrooot (approved):

PR #10477, opened 2026-03-25
fix(api): fix finding groups muted filter, counters and reaggregation

AdriiiPRodri:
## Summary

- **Muted filter on fast path**: Added `muted` as a computed filter in `FindingGroupAggregatedComputedFilter`, so `filter[muted]=true/false` no longer forces the slow Finding-table path. `muted=true` shows all finding groups (including fully-muted ones), `muted=false` hides groups where all findings have zero fail, pass, and muted counts.
- **Fixed check_description always empty in summaries**: The `aggregate_finding_group_summaries` function was looking for `metadata["Description"]` but the actual key is lowercase `"description"`. Now tries lowercase first, falls back to capitalized.
- **Fixed quoted strings in slow path**: Replaced `Cast(F("check_metadata__..."), CharField())` with `KeyTextTransform` which uses PostgreSQL's `->>` operator (returns clean text instead of JSON-quoted strings).
- **Reaggregate all providers on mute rule creation**: Previously, mute rule creation only reaggregated summaries for a single scan. Now uses a dedicated Celery task that reaggregates the latest completed scan for every provider.
- **Fixed inflated resources_total**: Summary and finding-level aggregation included stale data from old scans. Now both paths use the actual latest scan per provider (from the Scan table for findings, Window MAX for summaries) to match the `/resources` endpoint counts.
- **Optimized summary deduplication**: Replaced correlated Subquery with `Window(Max("inserted_at"))` for single-pass deduplication by provider.

## Changes

| File | Change |
|------|--------|
| `api/filters.py` | Add `muted` BooleanFilter to `FindingGroupAggregatedComputedFilter` |
| `api/v1/views.py` | Add `"muted"` to `computed_keys`, use `KeyTextTransform` for JSON extraction, simplify mute rule task dispatch, fix `resources_total` inflation on both fast/slow paths, optimize summary dedup with Window MAX |
| `tasks/jobs/scan.py` | Fix `check_description` metadata key case |
| `tasks/tasks.py` | Add `reaggregate_all_finding_group_summaries_task` with RLS context |
| `api/tests/test_views.py` | Update `test_mute_rules_create_valid` to match new task pattern |

## Test plan

- [x] All 119 finding_group + mute_rule tests pass
- [x] Deploy to dev and verify:
  - `filter[muted]=true` returns all finding groups including fully-muted ones (fast path)
  - `filter[muted]=false` hides fully-muted finding groups
  - `check_description` is populated correctly
  - `check_title` has no surrounding quotes
  - Creating a mute rule triggers reaggregation for all providers
  - `resources_total` matches `/resources` endpoint count
  - `/latest` endpoint responds in <2s

### DIFF
--- a/api/src/backend/api/filters.py
+++ b/api/src/backend/api/filters.py
@@ -1065,7 +1065,7 @@ class LatestFindingGroupSummaryFilter(FilterSet):
 
 
 class FindingGroupAggregatedComputedFilter(FilterSet):
-    """Filter aggregated finding-group rows by computed status/severity."""
+    """Filter aggregated finding-group rows by computed status/severity/muted."""
 
     STATUS_CHOICES = (
         ("FAIL", "Fail"),
@@ -1077,6 +1077,7 @@ class FindingGroupAggregatedComputedFilter(FilterSet):
     status__in = CharInFilter(method="filter_status_in", lookup_expr="in")
     severity = ChoiceFilter(method="filter_severity", choices=SeverityChoices)
     severity__in = CharInFilter(method="filter_severity_in", lookup_expr="in")
+    muted = BooleanFilter(method="filter_muted")
 
     def filter_status(self, queryset, name, value):
         return queryset.filter(aggregated_status=value)
@@ -1147,6 +1148,15 @@ class FindingGroupAggregatedComputedFilter(FilterSet):
             return queryset
 
         return queryset.filter(severity_order__in=orders)
+
+    def filter_muted(self, queryset, name, value):
+        if value is True:
+            # Show all finding groups, including fully-muted ones
+            return queryset
+        # muted=false: hide finding groups where all findings are muted
+        return queryset.filter(
+            Q(fail_count__gt=0) | Q(pass_count__gt=0) | Q(muted_count__gt=0)
+        )
 
 
 class ProviderSecretFilter(FilterSet):
--- a/api/src/backend/api/filters.py
+++ b/api/src/backend/api/filters.py
@@ -1151,12 +1151,10 @@ class FindingGroupAggregatedComputedFilter(FilterSet):
 
     def filter_muted(self, queryset, name, value):
         if value is True:
-            # Show all finding groups, including fully-muted ones
-            return queryset
-        # muted=false: hide finding groups where all findings are muted
-        return queryset.filter(
-            Q(fail_count__gt=0) | Q(pass_count__gt=0) | Q(muted_count__gt=0)
-        )
+            # Only fully-muted groups: all findings muted, none passing/failing
+            return queryset.filter(fail_count=0, pass_count=0, muted_count__gt=0)
+        # muted=false: groups with at least one non-muted finding
+        return queryset.filter(Q(fail_count__gt=0) | Q(pass_count__gt=0))
 
 
 class ProviderSecretFilter(FilterSet):
--- a/api/src/backend/api/filters.py
+++ b/api/src/backend/api/filters.py
@@ -1077,7 +1077,7 @@ class FindingGroupAggregatedComputedFilter(FilterSet):
     status__in = CharInFilter(method="filter_status_in", lookup_expr="in")
     severity = ChoiceFilter(method="filter_severity", choices=SeverityChoices)
     severity__in = CharInFilter(method="filter_severity_in", lookup_expr="in")
-    muted = BooleanFilter(method="filter_muted")
+    include_muted = BooleanFilter(method="filter_include_muted")
 
     def filter_status(self, queryset, name, value):
         return queryset.filter(aggregated_status=value)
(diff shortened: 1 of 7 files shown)


api/src/backend/api/filters.py
```diff
@@ -1148,6 +1149,15 @@ def filter_severity_in(self, queryset, name, value):
...
         return queryset.filter(severity_order__in=orders)
 
+    def filter_muted(self, queryset, name, value):
+        if value is True:
+            # Show all finding groups, including fully-muted ones
+            return queryset
+        # muted=false: hide finding groups where all findings are muted
+        return queryset.filter(
+            Q(fail_count__gt=0) | Q(pass_count__gt=0) | Q(muted_count__gt=0)
+        )
```
josema-xyz: The `Q(muted_count__gt=0)` in the `OR` lets a fully-muted group (`fail=0`, `pass=0`, `muted=5`) through, because `muted_count > 0`. Should be just `Q(fail_count__gt=0) | Q(pass_count__gt=0)`?

api/src/backend/api/filters.py
```diff
@@ -1148,6 +1149,15 @@ def filter_severity_in(self, queryset, name, value):
 
         return queryset.filter(severity_order__in=orders)
 
+    def filter_muted(self, queryset, name, value):
+        if value is True:
```
josema-xyz, line 1153: `muted=true` returns the `queryset` as-is, same as not passing the filter. An API consumer would expect `muted=true` to return only fully-muted groups. If the intent is "include muted groups in results", maybe the parameter should be called `include_muted` instead of `muted`, or `muted=true` could filter by `fail_count=0`, `pass_count=0`.

api/src/backend/tasks/tasks.py
```diff
@@ -760,6 +760,28 @@ def aggregate_finding_group_summaries_task(tenant_id: str, scan_id: str):
     return aggregate_finding_group_summaries(tenant_id=tenant_id, scan_id=scan_id)
 
 
+@shared_task(
+    base=RLSTask, name="reaggregate-all-finding-group-summaries", queue="overview"
+)
+@set_tenant(keep_tenant=True)
+def reaggregate_all_finding_group_summaries_task(tenant_id: str):
```
josema-xyz, line 767: This task it's only verified with a mock that checks it gets called. Worth adding at least one test for the internal logic.

josema-xyz, line 767: Does this task need `@handle_provider_deletion`?

api/src/backend/tasks/tasks.py
```diff
@@ -760,6 +760,28 @@ def aggregate_finding_group_summaries_task(tenant_id: str, scan_id: str):
...
+    """Reaggregate finding group summaries for all providers' latest completed scans."""
+    latest_scan_ids = list(
+        Scan.objects.filter(tenant_id=tenant_id, state=StateChoices.COMPLETED)
+        .order_by("provider_id", "-completed_at", "-inserted_at")
+        .distinct("provider_id")
+        .values_list("id", flat=True)
+    )
+    if latest_scan_ids:
```
josema-xyz, line 775: the `group().apply_async()` is fire-and-forget. A `logger.info` with the scan IDs being reaggregated would help debug failures.

api/src/backend/api/filters.py
AdriiiPRodri: This is just a safety mechanism. If a `check_id` has a record in this table, it means there is at least one finding. In this case, we will not show check_ids that are completely muted by default. The `Q(muted_count__gt=0)` check is only to ensure that it is actually muted. In other words, although it shouldn't happen, if a check_id were to have 0 findings overall, that would be an error, not a sign that it is muted

api/src/backend/tasks/tasks.py
AdriiiPRodri: It doesn't need it. `@handle_provider_deletion` is for tasks that receive a specific `provider_id` and could fail if that provider is deleted during execution

This task does not receive a provider_id. It queries all existing providers with `Scan.objects.filter(state=COMPLETED).distinct(“provider_id”)`. If a provider is deleted, it simply won't appear in the query

api/src/backend/api/filters.py
AdriiiPRodri: Totally, I'll rename it to make it clearer 👌🏼

api/src/backend/tasks/tasks.py
AdriiiPRodri: Added

api/src/backend/tasks/tasks.py
AdriiiPRodri: Added

josema-xyz (approved):

PR #10483, opened 2026-03-26
fix(kubernetes): use cluster name as provider_uid in OCSF output

andoniaf:
## Context

When running Kubernetes scans in in-cluster mode with the `--cluster-name` parameter, the OCSF output's `provider_uid` field was always set to `"In-Cluster"` (the context name) instead of the actual cluster name specified by the user. This caused Prowler Cloud to display the provider UID as "In-Cluster" regardless of the `--cluster-name` value.

## Description of Changes

The `provider_uid` for Kubernetes findings was mapped to `provider.identity.context`, which is always `"In-Cluster"` in in-cluster mode. Changed it to use `provider.identity.cluster`, which correctly reflects the `--cluster-name` CLI parameter (or defaults to `"in-cluster"` when not specified).

### Changes
- `prowler/lib/outputs/finding.py`: Map `provider_uid` to `provider.identity.cluster` instead of `provider.identity.context`
- `tests/lib/outputs/finding_test.py`: Update test assertion to match new `provider_uid` value

## Steps to Review

1. Verify the change in `prowler/lib/outputs/finding.py:256`
2. Run K8s finding output tests: `pytest -vv tests/lib/outputs/finding_test.py -k kubernetes`
3. Run OCSF output tests: `pytest -vv tests/lib/outputs/ocsf/ocsf_test.py -k kubernetes`

## Checklist

- [x] Code follows project conventions
- [x] Tests updated and passing
- [x] Pre-commit hooks passing


prowler/lib/outputs/finding.py
```diff
@@ -253,7 +253,7 @@ def generate_output(
                 output_data["account_uid"] = get_nested_attribute(
                     provider, "identity.cluster"
                 )
-                output_data["provider_uid"] = provider.identity.context
```
HugoPBrito, line 256: As far as I understand we need to add a verification for the cluster mode or not, so context ones keep the same. Right?

HugoPBrito (changes requested):

prowler/lib/outputs/finding.py
jfagoagas: I'm not sure of this change. If we're not executing `In-Cluster` `provider_uid` must be the `context`. Otherwise I think we can use the cluster.

HugoPBrito: That's what I tried to say, but done properly.

HugoPBrito (approved):

prowler/CHANGELOG.md
```diff
@@ -12,6 +12,10 @@ All notable changes to the **Prowler SDK** are documented in this file.
 
 - Azure Network Watcher flow log checks now require workspace-backed Traffic Analytics for `network_flow_log_captured_sent` and align metadata with VNet-compatible flow log guidance [(#10645)](https://github.com/prowler-cloud/prowler/pull/10645)
 
+### 🐞 Fixed
+
+- Kubernetes OCSF `provider_uid` now uses the cluster name in in-cluster mode (so `--cluster-name` is correctly reflected in findings) and keeps the kubeconfig context in kubeconfig mode [(#10483)](https://github.com/prowler-cloud/prowler/pull/10483)
```
HugoPBrito: Please move this entry to `v5.25.1`.

HugoPBrito (changes requested):

jfagoagas (approved):

HugoPBrito (approved):

PR #10486, opened 2026-03-26
fix(api): resolve check_title filter to check_id for consistent finding-group counts

AdriiiPRodri:
### Context

Finding-groups endpoint returns inconsistent resource/muted counts when filtering by `check_title__icontains` vs without it. Caused by providers scanned with older Prowler versions having different `checktitle` values in `check_metadata` for the same `check_id`. Also adds `resource_group` to `/resources` response and fixes `name__icontains` filter recognition.

### Description

- **check_title resolution**: `filter_check_title_icontains` now resolves matching titles to `check_id` first, then filters by `check_id__in`, ensuring all provider findings are included regardless of title variant. Same strategy already used by `_CheckTitleToCheckIdMixin` on the summary path.
- **resource_group field**: Adds `resource_group` (from `check_metadata.resourcegroup` via `KeyTextTransform`) to finding-group `/resources` endpoint inside the existing `resource` object.
- **name__icontains filter**: Dynamically builds resource filter fields from `LatestResourceFilter.Meta.fields` so `filter[name__icontains]` is recognized on `/resources`.

### Steps to review

1. Review `filter_check_title_icontains` in `filters.py` — now resolves to `check_id__in` instead of filtering JSON directly.
2. Review `_build_resource_aggregation` and `_post_process_resources` in `views.py` — new `resource_group` annotation and passthrough.
3. Review `FindingGroupResourceSerializer.get_resource` in `serializers.py` — new `resource_group` field in response.
4. Review `_RESOURCE_FILTER_FIELDS` and `_split_resource_filters` in `views.py` — dynamic filter field recognition.

### Checklist

<details>

</details>

#### API

### License


api/src/backend/api/v1/views.py
alejandrobailo, line 1: If resource_group comes from resource, is it better to call it "group"?

AdriiiPRodri: Well, I'm not entirely sure, since even if it comes in resource, the field in metadata is named that way and it helps us easily identify it if we need to debug or for anyone who wants to work on the API

AdriiiPRodri:
## Explanation of the findings calls fix

The `list()` materialization fix, explained clearly

### The problem

The findings/latest endpoint filters findings using the most recent completed scan per provider:

```
latest_scan_ids = (
    Scan.all_objects.filter(tenant_id=tenant_id, state=StateChoices.COMPLETED)
    .order_by("provider_id", "-inserted_at")
    .distinct("provider_id")
    .values_list("id", flat=True)
)

filtered_queryset = filtered_queryset.filter(scan_id__in=latest_scan_ids)
```

If you don’t wrap that queryset in list(), Django turns it into a SQL subquery:

```
SELECT ...
FROM findings
WHERE scan_id IN (
    SELECT DISTINCT ON (provider_id) id
    FROM scans
    WHERE tenant_id = 'a3b26b14-...' AND state = 'completed'
    ORDER BY provider_id, inserted_at DESC
)
AND tenant_id = current_setting('api.tenant_id')::uuid
```

What `EXPLAIN ANALYZE` showed

With **prowler_user (RLS enabled)** and no list(), PostgreSQL can’t properly estimate how many rows the subquery returns. The RLS condition (current_setting(...)) behaves like a black box for the planner.

Because of that, it assumes the subquery is large and chooses a Nested Loop Semi Join. That means: for every row in findings, it re-runs the subquery to check if there’s a match.

The execution plan looks roughly like this:

```
Nested Loop Semi Join
  -> Append
       -> Seq Scan on findings_p2026_03 (≈500k rows)
       -> Seq Scan on findings_p2026_02 (≈400k rows)
  -> Materialize
       -> Unique (DISTINCT ON provider_id)
            -> Sort
                 -> Seq Scan on scans (filtered by tenant_id and state)
```

The key issue is that the subquery is effectively re-evaluated many times. With ~900k rows, that becomes very expensive. Around 3.9 seconds, and much worse with larger datasets or cold cache.

With postgres (BYPASSRLS) and still no list(), the planner has full visibility. It correctly estimates the subquery returns only a handful of rows and switches to a Hash Semi Join. The subquery runs once, builds a hash table, and each row just probes it.

**Execution time drops to about 250 ms.**

### The fix

latest_scan_ids = list(
    Scan.all_objects.filter(...)
    .order_by("provider_id", "-inserted_at")
    .distinct("provider_id")
    .values_list("id", flat=True)
)

Calling `list()` forces evaluation in Python. The subquery runs once and returns a small set of UUIDs.

Django then generates something like:

```
WHERE scan_id IN ('019558a1-...', '019558a2-...', '019558a3-...')
```

Now PostgreSQL sees literal values instead of a subquery, so it can plan efficiently even with RLS enabled. It typically uses a Hash Semi Join or an indexed ANY(...).

The plan becomes:

```
Append
  -> Index Scan on findings_p2026_03 using scan_id index
       Filter: scan_id = ANY('{019558a1-..., 019558a2-..., ...}')
  -> Index Scan on findings_p2026_02 using scan_id index
```

With `prowler_user`, execution time drops to ~260 ms, basically the same as the superuser case.

### Root cause

The PostgreSQL planner treats RLS predicates as opaque. It can’t incorporate them properly when estimating subquery size, especially in correlated contexts. That leads to bad decisions like choosing Nested Loop instead of Hash joins.

This issue had been there since RLS was added to the findings table. It didn’t show up in development because the connection usually runs as a superuser, which bypasses RLS and gives the planner full information. It only becomes visible in production where RLS is actually enforced.

api/src/backend/api/tests/test_views.py
```diff
@@ -15851,6 +15851,25 @@ def test_resources_fields(self, authenticated_client, finding_groups_fixture):
...
+        )
+        assert response.status_code == status.HTTP_200_OK
+        data = response.json()["data"]
+        assert len(data) == 2
+        for item in data:
+            resource = item["attributes"]["resource"]
+            assert (
+                resource["resource_group"] == "storage"
+            ), "resource_group must be 'storage'"
+
```
josema-xyz, line 15916: Missing a test for the main fix: two providers with different `checktitle` values for the same `check_id`, so this can't regress silently.

api/src/backend/api/v1/views.py
```diff
@@ -7237,6 +7249,9 @@ def _build_resource_aggregation(
                 # applied per-check so all findings for the same resource
                 # share the same muted_reason in practice.
                 muted_reason=Max("finding__muted_reason"),
+                resource_group=Max(
```
josema-xyz, line 7251: Maybe an unified comment for `muted_reason` and `check_metadata`?

api/src/backend/api/tests/test_views.py
josema-xyz, line 15916: Maybe a `name__icontains` test on `/resources`?

josema-xyz: Fixed in 1474a8694484011a2ced1647dfb79670310ce7f2.

api/src/backend/api/v1/views.py
josema-xyz: Fixed in 1474a8694484011a2ced1647dfb79670310ce7f2.

api/src/backend/api/tests/test_views.py
josema-xyz: Fixed in 1474a8694484011a2ced1647dfb79670310ce7f2.

josema-xyz (approved):

PR #10491, opened 2026-03-27
feat(ui): Add tenant management

pfe-nazaries:
## Summary

Multi-tenant organization management: users can now create, switch, edit, and delete organizations from the profile page.

Each organization is displayed as a membership card with role, join date, and action buttons based on ownership. Switching between orgs updates the session tokens and reloads the page. Creating a new org automatically switches to it.

Deleting the active organization prompts the user to select a target org to switch to. The switch happens first (while the token is still valid), then the delete uses the new token — all in a single server action (`switchThenDeleteTenant`) to avoid the backend rejecting an invalidated JWT.

On the API side, RBAC role lookup is now filtered by `tenant_id` to prevent cross-tenant privilege leaks, and tenant creation assigns the admin role automatically.

### Browser tests

The profile organizations page is covered by 58 browser tests running in real Chromium via Playwright instead of jsdom — faster, more reliable, with real form submissions via `useActionState` and API calls intercepted by MSW. If the team prefers, they can be split into a separate PR.

`window.location.reload()` was extracted to `lib/navigation.ts` so it can be mocked in browser tests (real Chromium doesn't allow redefining `window.location`).


api/src/backend/api/base_views.py
```diff
@@ -144,6 +147,11 @@ def _create_admin_role(self, tenant_id):
             manage_scans=True,
             unlimited_visibility=True,
         )
+        UserRoleRelationship.objects.using(MainRouter.admin_db).create(
+            user=self.request.user,
+            role=admin_role,
+            tenant_id=tenant_id,
+        )
```
josema-xyz, line 146: These two creates should be wrapped in one `transaction.atomic`.

api/src/backend/api/v1/views.py
```diff
@@ -1417,7 +1422,7 @@ def set_required_permissions(self):
             self.required_permissions = [Permissions.MANAGE_PROVIDERS]
 
     def get_queryset(self):
-        user_roles = get_role(self.request.user)
+        user_roles = get_role(self.request.user, self.request.tenant_id)
```
josema-xyz, line 1449: All the `get_role` calls should now return an object with the name in singular.

api/src/backend/api/base_views.py
Davidm4r: Fixed in the last commit

api/src/backend/api/v1/views.py
Davidm4r: In the past they were always returning a list and picking the first one.
It should be imposible to have more than one role, but I don't know if returning a single object it's going to break something. To ensure backward compatibility, I'd keep this as it is.

josema-xyz: Yep, the original function returned only one `Role` or `None`. Let's keep like that because it was like that before.

josema-xyz (dismissed):
API approved!

.github/workflows/api-security.yml
```diff
@@ -69,9 +69,10 @@ jobs:
 
       - name: Safety
         if: steps.check-changes.outputs.any_changed == 'true'
-        run: poetry run safety check --ignore 79023,79027,86217
+        run: poetry run safety check --ignore 79023,79027,86217,SFTY-20260218-01424
         # TODO: 79023 & 79027 knack ReDoS until `azure-cli-core` (via `cartography`) allows `knack` >=0.13.0
         # TODO: 86217 because `alibabacloud-tea-openapi == 0.4.3` don't let us upgrade `cryptography >= 46.0.0`
+        # TODO: SFTY-20260218-01424 nltk 3.9.2 CVE-2025-14009 — nltk is a transitive dependency of `safety` itself, pinned to <=3.9.2; we have no control over upgrading it until safety releases a fix
```
AdriiiPRodri: I think it's better to update our lock file rather than add this exception, since we're actually using 3.9.2

AdriiiPRodri (changes requested):

api/src/backend/api/base_views.py
```diff
@@ -133,17 +136,23 @@ def dispatch(self, request, *args, **kwargs):
             self.db_alias = MainRouter.default_db
```
AdriiiPRodri: I think we can improve this, actually we create the tenant and the role in you different transactions this is why we need a `handle_creation_error` that tries to remove the tenant but if this fail then we will have an orphan tenant, we can simplify and improve this this way:


```suggestion
with transaction.atomic(using=self.db_alias):
    tenant = super().dispatch(request, *args, **kwargs)

    if (
        request.method == "POST"
        and isinstance(tenant, Response)
        and tenant.status_code == 201
    ):
        self._create_admin_role(tenant.data["id"])

    return tenant
finally:
    if alias_token is not None:
        reset_read_db_alias(alias_token)
    self.db_alias = MainRouter.default_db
```

josema-xyz: Totally aggree.

alejandrobailo (changes requested):
### Code review

Found 4 issues, plus 4 suggestions:

**Issues:**

1. **`AlertDialog` and `AlertModal` are unnecessary — use the existing `Modal`**. The PR introduces `alert-dialog.tsx` (141 lines) + `alert-modal/alert-modal.tsx` (67 lines) as new Radix `AlertDialog` primitives. `AlertDialog` is designed for cases where the user **must** choose an action (e.g., "Save before leaving?" where dismissing would cause data loss). In this PR, both use cases (switch and delete confirmation) have no such requirement — dismissing the modal is functionally identical to clicking Cancel: no action is taken, no data is lost, no state changes. The existing `Modal` component (wrapping `Dialog`) already handles this perfectly. `AlertModal` is also a near-exact copy of `Modal` — identical `SIZE_CLASSES`, identical props interface, identical `cn()` base classes, identical conditional header rendering. The only difference is swapping `Dialog*` for `AlertDialog*`. This adds 208 lines of duplicated code, a new `@radix-ui/react-alert-dialog` dependency, and two new files to maintain. Replace all `AlertModal` usages with `Modal` in `membership-item.tsx`, then delete both files.

https://github.com/prowler-cloud/prowler/blob/6270e8437714c17055b3f1571ea61f710dd20345/ui/components/shadcn/alert-dialog.tsx#L1-L141

https://github.com/prowler-cloud/prowler/blob/6270e8437714c17055b3f1571ea61f710dd20345/ui/components/shadcn/alert-modal/alert-modal.tsx#L1-L67

https://github.com/prowler-cloud/prowler/blob/6270e8437714c17055b3f1571ea61f710dd20345/ui/components/users/profile/membership-item.tsx#L48-L75

2. `availableTenants: Array<{ id: string; name: string }>` appears in both `DeleteTenantFormProps` and `MembershipItem` props as an inline nested object (AGENTS.md: "NEVER: Inline nested objects" + "Used 2+ places -> extract to `types/`"). Should extract to a named interface.

https://github.com/prowler-cloud/prowler/blob/6270e8437714c17055b3f1571ea61f710dd20345/ui/components/users/forms/delete-tenant-form.tsx#L27-L31

https://github.com/prowler-cloud/prowler/blob/6270e8437714c17055b3f1571ea61f710dd20345/ui/components/users/profile/membership-item.tsx#L27-L28

3. `delete-tenant-form.tsx` uses a raw `<input>` with manual Tailwind classes instead of the shadcn `Input` component (exported from `@/components/shadcn`). The manual classes use different focus tokens (`ring-offset-background`, `focus-visible:ring-ring`) than `Input`'s CVA variants (`focus:ring-inset`, `focus:ring-border-input-primary-press`), creating visual inconsistency. `CreateTenantForm` correctly uses `CustomServerInput`.

https://github.com/prowler-cloud/prowler/blob/6270e8437714c17055b3f1571ea61f710dd20345/ui/components/users/forms/delete-tenant-form.tsx#L115-L124

4. `alert-dialog.tsx` uses `import * as React from "react"` (AGENTS.md: "NEVER: `import React`, `import * as React`, `import React as *`"). The existing `dialog.tsx` uses the global `React` namespace without an explicit import. This issue goes away if the file is removed per issue #1.

https://github.com/prowler-cloud/prowler/blob/6270e8437714c17055b3f1571ea61f710dd20345/ui/components/shadcn/alert-dialog.tsx#L3-L5

**Suggestions (non-blocking):**

5. `hasManageAccount` is declared in `MembershipsCardClientProps` and threaded from `page.tsx` -> `MembershipsCard` -> `MembershipsCardClient`, but never destructured or used. The "Create organization" button renders unconditionally. If intentional ("any user can create"), the prop should be removed from the interface and call sites to avoid confusion.

https://github.com/prowler-cloud/prowler/blob/6270e8437714c17055b3f1571ea61f710dd20345/ui/components/users/profile/memberships-card-client.tsx#L24-L32

6. Fallback URLs are inconsistent between MSW handlers (`"https://some-api-server/api/v1"`) and browser tests (`"https://api.test.prowler.com/api/v1"`). Both work today because `vitest.config.ts` defines the env var, but if the `define` is ever removed, tests will silently fail. Consider extracting to a shared constant in `testing/constants.ts`.

https://github.com/prowler-cloud/prowler/blob/6270e8437714c17055b3f1571ea61f710dd20345/ui/testing/msw/handlers.ts#L3-L4

https://github.com/prowler-cloud/prowler/blob/6270e8437714c17055b3f1571ea61f710dd20345/ui/app/(prowler)/profile/profile-organizations.browser.test.tsx#L831-L833

7. The delete confirmation `<input>` has no `<label>` element or `aria-label`. The "Type X to confirm" text is in a `<div>`, not accessible as a form label. Screen readers will announce it as an unlabeled text input.

https://github.com/prowler-cloud/prowler/blob/6270e8437714c17055b3f1571ea61f710dd20345/ui/components/users/forms/delete-tenant-form.tsx#L115-L137

8. If issue #1 is addressed, the browser test harness queries for `alertdialog` role in several locators (`alertDialog()`, `hiddenTenantInput()`) that should be updated to `dialog`.

https://github.com/prowler-cloud/prowler/blob/6270e8437714c17055b3f1571ea61f710dd20345/ui/app/(prowler)/profile/profile-organizations.harness.ts#L59-L60

.github/workflows/api-security.yml
AdriiiPRodri: We can remove this once this PR is merged: https://github.com/prowler-cloud/prowler/pull/10492

Alan-TheGentleman (changes requested):
Three things to fix before this can merge, plus a few smaller items:

1. **`get_role()` hits the default DB, not `admin_db`**. `HasPermissions` correctly uses `.using(MainRouter.admin_db)` to bypass RLS for role lookups, but `get_role()` goes through the default DB where RLS is active. Same user, two different code paths, potentially different results. Add `.using(MainRouter.admin_db)` to the `user.roles.filter()` call in `get_role()`.

https://github.com/prowler-cloud/prowler/blob/6270e8437714c17055b3f1571ea61f710dd20345/api/src/backend/api/rbac/permissions.py#L56-L61

2. **Tenant + role creation isn't atomic**. The tenant commits first, then `_create_admin_role()` runs in a separate transaction on `admin_db`. If role creation fails and the compensating delete also fails, you end up with an orphan tenant. AdriiiPRodri already suggested wrapping both in a single atomic block. +1 on that.

https://github.com/prowler-cloud/prowler/blob/6270e8437714c17055b3f1571ea61f710dd20345/api/src/backend/api/base_views.py#L117-L136

3. **The security fix has no regression test**. You're fixing a cross-tenant privilege leak, which is great. But there's no test that proves it works: create a user with a role only in tenant A, mint a token for tenant B, hit an RBAC-protected endpoint, assert denial. Without that, this can regress silently.

Smaller stuff (not blocking, but worth addressing):

4. **`get_serializer_context()` swallows `PermissionDenied`** and returns `role=None`. If a user has no role in this tenant, that's a real signal. Catching it and continuing with `None` can crash downstream serializers or let unauthorized data through.

https://github.com/prowler-cloud/prowler/blob/6270e8437714c17055b3f1571ea61f710dd20345/api/src/backend/api/v1/views.py#L950-L956

5. **`CreateTenantState`, `DeleteTenantState`, `SwitchThenDeleteTenantState` use loose optionals**. `SwitchTenantState` already does the right thing with a discriminated union. The other three should match so you can't accidentally end up with `{ success: true, error: "something" }`.

https://github.com/prowler-cloud/prowler/blob/6270e8437714c17055b3f1571ea61f710dd20345/ui/actions/users/tenants.ts#L155-L160

6. **`hasManageAccount` is passed through 3 components but never read**. The Create button shows unconditionally. If that's intentional, drop the prop. If not, wire it up.

https://github.com/prowler-cloud/prowler/blob/6270e8437714c17055b3f1571ea61f710dd20345/ui/components/users/profile/memberships-card-client.tsx#L24-L32

7. **`as any` in MSW handlers**. Type as `unknown` instead.

https://github.com/prowler-cloud/prowler/blob/6270e8437714c17055b3f1571ea61f710dd20345/ui/testing/msw/handlers.ts#L28

8. **`process.env: {}` in vitest browser config nukes all env vars**, not just the one you need. Any code reading `process.env.NODE_ENV` or similar will silently get `undefined`. Scope it to the specific vars.

9. **`_prevState: any` in `updateTenantName`**. Pre-existing, but your new actions in the same file type it correctly. Good time to fix the old one too.

https://github.com/prowler-cloud/prowler/blob/6270e8437714c17055b3f1571ea61f710dd20345/ui/actions/users/tenants.ts#L41

10. **`import React` in `page.tsx` and `import * as React` in `alert-dialog.tsx`**. Both violate our import convention. The `alert-dialog.tsx` one goes away if you adopt alejandrobailo's `Modal` suggestion. The `page.tsx` one is a quick fix.

https://github.com/prowler-cloud/prowler/blob/6270e8437714c17055b3f1571ea61f710dd20345/ui/app/(prowler)/profile/page.tsx#L1

Non-blocking suggestions:

11. MSW handler fallback URL and test fallback URL are different strings for the same thing. One shared constant would prevent them from drifting.

12. Harness queries `role="alertdialog"`. If you switch to `Modal` per alejandrobailo's review, update to `role="dialog"`.

13. No test checks that `reloadPage()` is actually called after switch/create/delete succeeds.

14. `create-tenant-form.tsx` has an async `useEffect` with no unmount cleanup. If the user navigates away mid-flight, `update()` and `reloadPage()` still fire.

I agree with alejandrobailo's four issues and both of AdriiiPRodri's requests. The RBAC fix is solid work and the 58 browser tests are impressive. Fix the three critical items and this is good to go.

Davidm4r:
> 2\. Tenant + role creation isn't atomic

I have already reviewed and corrected all the API changes.

.github/workflows/api-security.yml
Davidm4r: Done

api/src/backend/api/base_views.py
Davidm4r: I already fixed this. Thanks

pfe-nazaries:
@Alan-TheGentleman @alejandrobailo Fixed comments

AdriiiPRodri (dismissed):
API side LGTM!

alejandrobailo (changes requested):
Hey! Quick question about the Vitest Browser Mode setup. Was this something discussed beforehand? I don't recall us aligning on introducing a third testing layer.

We already have Playwright for E2E and jsdom for unit tests. What does browser mode give us that we can't cover with those two? Just want to understand the reasoning before we set a precedent for future tests.

Alan-TheGentleman (approved):

alejandrobailo (approved):

jfagoagas (approved):

josema-xyz (approved):

PR #10499, opened 2026-03-27
fix: replace bare except with except Exception in prowler-wrapper

harshadkhetpal:
## Summary

Replace bare `except:` with `except Exception:` in `contrib/wazuh/prowler-wrapper.py`.

PEP 8 recommends against bare `except:` as it catches system-exiting exceptions. Using `except Exception:` catches all program errors while letting `KeyboardInterrupt` and `SystemExit` propagate normally.

**Change:**
```python
# Before
except:

# After
except Exception:
```

## Testing
No behavior change — style/correctness fix only.


jfagoagas (approved):
Thanks for this one!

PR #10501, opened 2026-03-28
feat(github): add --repo-list-file flag for GitHub scanning

### LINKED ISSUE #8708: Add Github scan based on customer provided repo-list file.
### New feature motivation

In my use case, I would like to scan only the repo that I care about for my team or my organization, not just one or two repos. 

I suggest to create a new argument in prowler CLI tool at least, to allow customer provide their own repo-list file to include all the repo they would like to scan each round. In my case these repos are under the same organization. 

This allow Security Team that governing the entire organization to monitor and find out potential problem. 

### Solution Proposed

By using the current Prowler CLI v5.12.0 to scan github repo, I find it is more from developer point of view to scan their local saved repos. As a security engineer, I would like to have a tool to scan organization or team level repos to find out code venerability and remind the developer who usually do not scan their code or do not have such ideas

I do not have other alternatives right now. 

### Additional context

_No response_

raajheshkannaa:
Closes #8708

## Summary

Adds a `--repo-list-file` CLI flag to the GitHub provider that accepts a path to a file containing a list of repositories to scan, one per line in `owner/repo-name` format. Lines starting with `#` are treated as comments and ignored.

This allows security teams governing an organization to maintain a curated list of repos for scanning instead of passing them all as individual `--repository` arguments or scanning entire organizations.

## Changes

- Added `--repo-list-file` argument to the GitHub provider's CLI parser (`prowler/providers/github/lib/arguments/arguments.py`)
- Added file reading logic in `prowler/providers/common/provider.py` where GitHub CLI args are mapped to the provider constructor, merging repos from the file with any repos passed via `--repository`
- Added unit and integration tests for the new argument

## Usage

```bash
# repos.txt (one repo per line, # comments supported)
# Team Alpha repos
myorg/service-api
myorg/service-web
# Team Beta repos
myorg/data-pipeline

prowler github --personal-access-token ghp_xxx --repo-list-file repos.txt
```

The flag can be combined with `--repository` and `--organization` for flexible scoping. Duplicate repos are automatically deduplicated.


prowler/providers/github/github_provider.py
```diff
@@ -256,6 +266,44 @@ def organizations(self) -> list:
...
+                    if len(line) > self.MAX_REPO_NAME_LENGTH:
+                        logger.warning(
+                            f"Skipping repo name exceeding {self.MAX_REPO_NAME_LENGTH} chars at line {line_count} in {file_path}"
+                        )
+                        continue
+                    self._repositories.append(line)
+            logger.info(
+                f"Loaded {len(self._repositories)} repositories from {file_path}"
+            )
+        except OSError:
```
andoniaf: `OSError` is too broad here — it catches `PermissionError`, `IsADirectoryError`, etc., all of which would incorrectly raise `GithubRepoListFileNotFoundError`. Should be `except FileNotFoundError:` to match the `ImageProvider._load_images_from_file` pattern and let other OS errors fall through to the generic `except Exception` handler.

prowler/providers/github/github_provider.py
```diff
@@ -147,6 +153,10 @@ def __init__(
         else:
             self._repositories = list(repositories)
 
+        # Load repos from file if provided
+        if repo_list_file:
+            self._load_repos_from_file(repo_list_file)
```
andoniaf, line 158: The PR description mentions automatic deduplication, but repos from the file are appended to `self._repositories` without dedup. If a repo appears in both `--repository` and the file, it'll be scanned twice. Consider adding `self._repositories = list(dict.fromkeys(self._repositories))` after this call.

prowler/providers/github/github_provider.py
```diff
@@ -256,6 +266,44 @@ def organizations(self) -> list:
         """
         return self._organizations
 
+    def _load_repos_from_file(self, file_path: str) -> None:
+        """Load repository names from a file (one per line)."""
+        try:
+            line_count = 0
+            with open(file_path, "r") as f:
+                for line in f:
+                    line_count += 1
```
andoniaf: `line_count` increments before blank/comment lines are skipped, so a file with 10k comment lines and 1 real repo would hit `MAX_REPO_LIST_LINES`. If the intent is to cap *repository entries* rather than total file lines, move the increment after the `if not line or line.startswith("#"): continue` block.

prowler/providers/github/github_provider.py
```diff
@@ -256,6 +266,44 @@ def organizations(self) -> list:
...
+                        )
+                    line = line.strip()
+                    if not line or line.startswith("#"):
+                        continue
+                    if len(line) > self.MAX_REPO_NAME_LENGTH:
+                        logger.warning(
+                            f"Skipping repo name exceeding {self.MAX_REPO_NAME_LENGTH} chars at line {line_count} in {file_path}"
+                        )
+                        continue
+                    self._repositories.append(line)
```
andoniaf, line 290: `len(self._repositories)` includes repos that were already present from `--repository` before the file was loaded. Consider capturing the count before the loop and logging only the delta, e.g.:
```python
before = len(self._repositories)
# ... loop ...
logger.info(f"Loaded {len(self._repositories) - before} repositories from {file_path}")
```

andoniaf (commented):
Thanks for the contribution @raajheshkannaa! The feature looks good, I left a few inline comments. Please take a look when you get a chance.

prowler/providers/github/github_provider.py
raajheshkannaa: Fixed -- changed to `except FileNotFoundError` so `PermissionError`, `IsADirectoryError`, etc. fall through to the generic `except Exception` handler.

prowler/providers/github/github_provider.py
raajheshkannaa: Good catch. Added `self._repositories = list(dict.fromkeys(self._repositories))` after the file load loop, which deduplicates while preserving order. Repos from `--repository` that also appear in the file won't be scanned twice.

prowler/providers/github/github_provider.py
raajheshkannaa: Fixed. Moved the increment after the blank/comment skip so the cap applies only to real repository entries, not total file lines. Renamed to `repo_count` for clarity.

prowler/providers/github/github_provider.py
raajheshkannaa: Fixed. Capture `before = len(self._repositories)` before the loop and log `len(self._repositories) - before` so the message reflects only repos added from the file.

raajheshkannaa:
Hi @andoniaf, the four inline review comments are all addressed (FileNotFoundError narrowing, dedup, repo_count vs line_count, delta-count logging). CI is green aside from codecov coverage gates. Would appreciate another look when you have time. Thanks!

andoniaf (approved):

PR #10510, opened 2026-03-30
fix(api): false 404 and sorting on finding group resources endpoints

AdriiiPRodri:
### Context

The finding group resources endpoints (`/resources` and `/latest/.../resources`) had two bugs:
1. Applying a status filter (e.g. `filter[status]=PASS`) on a valid finding group that only has FAIL findings returned a **404 instead of an empty list**.
2. The `sort` query parameter was **completely ignored** — results were always ordered by `resource_id`.

### Description

**False 404 fix:** The existence check for the finding group was performed *after* applying user filters. If the filters eliminated all findings (e.g. filtering PASS on a check with only FAIL), the endpoint incorrectly concluded the finding group didn't exist. The fix moves the existence check *before* applying user filters.

**Sort support:** Added a `_RESOURCE_SORT_MAP` that maps JSON:API sort fields (`status`, `severity`, `first_seen_at`, `last_seen_at`, `resource.uid`, `resource.name`, `resource.region`, `resource.service`, `resource.type`, `provider.uid`, `provider.alias`) to the aggregation annotation names. Reuses the existing `_sorted_paginated_response` method (now parameterized) to avoid code duplication.

Both `resources` and `latest_resources` endpoints are fixed.

### Steps to review

1. Review the existence check move in both `resources()` and `latest_resources()` methods.
2. Review the `_RESOURCE_SORT_MAP` and the parameterization of `_sorted_paginated_response` / `_validate_sort_fields`.
3. Verify the 12 new tests cover all scenarios.
4. Test manually:
   - `GET /api/v1/finding-groups/latest/<check_id>/resources?filter[status]=PASS` on a check with only FAIL → should return `{"data": []}` (200), not 404.
   - `GET /api/v1/finding-groups/latest/<check_id>/resources?sort=status` → PASS first, FAIL last.
   - `GET /api/v1/finding-groups/latest/<check_id>/resources?sort=-status` → FAIL first, PASS last.
   - `GET /api/v1/finding-groups/latest/<check_id>/resources?sort=invalid` → 400.

### Checklist

#### API

### License

### DIFF
--- a/api/src/backend/api/tests/test_views.py
+++ b/api/src/backend/api/tests/test_views.py
@@ -16011,6 +16011,108 @@ class TestFindingGroupViewSet:
         # Should still return the 2 resources within the date range
         assert len(response.json()["data"]) == 2
 
+    def test_resources_status_filter_returns_empty_not_404(
+        self, authenticated_client, finding_groups_fixture
+    ):
+        """Test that filtering by status on a valid check returns empty list, not 404."""
+        # s3_bucket_public_access has only FAIL findings, filtering by PASS should return []
+        response = authenticated_client.get(
+            reverse(
+                "finding-group-resources", kwargs={"pk": "s3_bucket_public_access"}
+            ),
+            {"filter[inserted_at]": TODAY, "filter[status]": "PASS"},
+        )
+        assert response.status_code == status.HTTP_200_OK
+        assert response.json()["data"] == []
+
+    def test_resources_nonexistent_check_still_404(
+        self, authenticated_client, finding_groups_fixture
+    ):
+        """Test that a truly nonexistent check_id still returns 404."""
+        response = authenticated_client.get(
+            reverse("finding-group-resources", kwargs={"pk": "totally_fake_check"}),
+            {"filter[inserted_at]": TODAY},
+        )
+        assert response.status_code == status.HTTP_404_NOT_FOUND
+
+    def test_resources_sort_by_status_ascending(
+        self, authenticated_client, finding_groups_fixture
+    ):
+        """Test sort=status returns PASS before FAIL."""
+        # ec2_instance_public_ip has 1 PASS (resource1) and 1 FAIL (resource2)
+        response = authenticated_client.get(
+            reverse(
+                "finding-group-resources",
+                kwargs={"pk": "ec2_instance_public_ip"},
+            ),
+            {"filter[inserted_at]": TODAY, "sort": "status"},
+        )
+        assert response.status_code == status.HTTP_200_OK
+        data = response.json()["data"]
+        assert len(data) == 2
+        assert data[0]["attributes"]["status"] == "PASS"
+        assert data[1]["attributes"]["status"] == "FAIL"
+
+    def test_resources_sort_by_status_descending(
+        self, authenticated_client, finding_groups_fixture
+    ):
+        """Test sort=-status returns FAIL before PASS."""
+        response = authenticated_client.get(
+            reverse(
+                "finding-group-resources",
+                kwargs={"pk": "ec2_instance_public_ip"},
+            ),
+            {"filter[inserted_at]": TODAY, "sort": "-status"},
+        )
+        assert response.status_code == status.HTTP_200_OK
+        data = response.json()["data"]
+        assert len(data) == 2
+        assert data[0]["attributes"]["status"] == "FAIL"
+        assert data[1]["attributes"]["status"] == "PASS"
+
+    def test_resources_sort_invalid_field_returns_400(
+        self, authenticated_client, finding_groups_fixture
+    ):
+        """Test that an invalid sort field returns 400."""
+        response = authenticated_client.get(
+            reverse(
+                "finding-group-resources", kwargs={"pk": "s3_bucket_public_access"}
+            ),
+            {"filter[inserted_at]": TODAY, "sort": "invalid_field"},
+        )
+        assert response.status_code == status.HTTP_400_BAD_REQUEST
+
+    def test_latest_resources_status_filter_returns_empty_not_404(
+        self, authenticated_client, finding_groups_fixture
+    ):
+        """Test latest resources with status filter on valid check returns empty, not 404."""
+        response = authenticated_client.get(
+            reverse(
+                "finding-group-latest_resources",
+                kwargs={"check_id": "s3_bucket_public_access"},
+            ),
+            {"filter[status]": "PASS"},
+        )
+        assert response.status_code == status.HTTP_200_OK
+        assert response.json()["data"] == []
+
+    def test_latest_resources_sort_by_status(
+        self, authenticated_client, finding_groups_fixture
+    ):
+        """Test latest resources sort=status returns PASS before FAIL."""
+        response = authenticated_client.get(
+            reverse(
+                "finding-group-latest_resources",
+                kwargs={"check_id": "ec2_instance_public_ip"},
+            ),
+            {"sort": "status"},
+        )
+        assert response.status_code == status.HTTP_200_OK
+        data = response.json()["data"]
+        assert len(data) == 2
+        assert data[0]["attributes"]["status"] == "PASS"
+        assert data[1]["attributes"]["status"] == "FAIL"
+
     # Test provider_id filter actually filters data
     def test_finding_groups_provider_id_filter_actually_filters(
         self, authenticated_client, finding_groups_fixture, providers_fixture
--- a/api/src/backend/api/tests/test_views.py
+++ b/api/src/backend/api/tests/test_views.py
@@ -16113,6 +16113,89 @@ class TestFindingGroupViewSet:
         assert data[0]["attributes"]["status"] == "PASS"
         assert data[1]["attributes"]["status"] == "FAIL"
 
+    def test_resources_nonexistent_check_missing_date_returns_400(
+        self, authenticated_client, finding_groups_fixture
+    ):
+        """Nonexistent check_id with missing required date filter returns 400, not 404."""
+        response = authenticated_client.get(
+            reverse("finding-group-resources", kwargs={"pk": "totally_fake_check"}),
+        )
+        # FindingGroupFilter requires inserted_at — validation fires before existence check
+        assert response.status_code == status.HTTP_400_BAD_REQUEST
+
+    def test_resources_nonexistent_check_invalid_sort_returns_400(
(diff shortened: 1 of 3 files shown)


Alan-TheGentleman (dismissed):
Clean fix. The false 404 is properly solved by moving the existence check before user filters, and the sort refactor is well-structured. A few things worth noting:

1. **Sort is validated twice per request.** `_validate_resource_sort()` runs in `resources()` / `latest_resources()` and discards the result, then `_paginated_resource_response()` calls `_validate_sort_fields()` again to get the ordering list. Either return the parsed ordering from the first call and thread it through, or drop the pre-validation method entirely.

2. **No cross-tenant test for the modified endpoints.** The existence check uses `_get_finding_queryset()` which filters by `tenant_id`, so isolation should work. But there's no test proving a check_id from tenant A returns 404 when requested from tenant B. Our convention says always test this.

3. **No multi-field or dot-notation sort tests.** The sort map supports 11 fields and comma-separated multi-sort, but all tests use a single field. `sort=status,severity` and `sort=resource.uid` are completely untested. The dot-notation fields go through a different parsing path than filters, so this is a real gap.

4. **`status_order` / `severity_order` Case/When is duplicated.** Same expressions exist in `_build_resource_aggregation()` and `_RESOURCE_SORT_ANNOTATIONS`. If a new status or severity is added, both need updating manually. Extract into a shared factory.

5. **`source.pointer` should be `source.parameter` in the sort ValidationError.** JSON:API spec says `source.pointer` is for request body paths, `source.parameter` for query params. Pre-existing pattern but now propagated to the new resource sort errors.

Non-blocking:

6. PR description says "7 new tests" but the diff adds 12. Minor traceability thing.

7. `sort=,` and `sort= ` silently fall back to default order instead of returning 400. Documented in tests as intentional, but a client sending those probably has a bug.

8. `@extend_schema` doesn't expose the `sort` parameter in OpenAPI. The 11 valid fields are undiscoverable via Swagger or generated SDKs.

Items 2 and 3 are worth a quick follow-up. The rest can be tackled whenever. Good work on the test coverage and the reusable sort infrastructure.

AdriiiPRodri:
> Clean fix. The false 404 is properly solved by moving the existence check before user filters, and the sort refactor is well-structured. A few things worth noting:
> 
>     1. **Sort is validated twice per request.** `_validate_resource_sort()` runs in `resources()` / `latest_resources()` and discards the result, then `_paginated_resource_response()` calls `_validate_sort_fields()` again to get the ordering list. Either return the parsed ordering from the first call and thread it through, or drop the pre-validation method entirely.
> 
>     2. **No cross-tenant test for the modified endpoints.** The existence check uses `_get_finding_queryset()` which filters by `tenant_id`, so isolation should work. But there's no test proving a check_id from tenant A returns 404 when requested from tenant B. Our convention says always test this.
> 
>     3. **No multi-field or dot-notation sort tests.** The sort map supports 11 fields and comma-separated multi-sort, but all tests use a single field. `sort=status,severity` and `sort=resource.uid` are completely untested. The dot-notation fields go through a different parsing path than filters, so this is a real gap.
> 
>     4. **`status_order` / `severity_order` Case/When is duplicated.** Same expressions exist in `_build_resource_aggregation()` and `_RESOURCE_SORT_ANNOTATIONS`. If a new status or severity is added, both need updating manually. Extract into a shared factory.
> 
>     5. **`source.pointer` should be `source.parameter` in the sort ValidationError.** JSON:API spec says `source.pointer` is for request body paths, `source.parameter` for query params. Pre-existing pattern but now propagated to the new resource sort errors.
> 
> 
> Non-blocking:
> 
>     6. PR description says "7 new tests" but the diff adds 12. Minor traceability thing.
> 
>     7. `sort=,` and `sort= ` silently fall back to default order instead of returning 400. Documented in tests as intentional, but a client sending those probably has a bug.
> 
>     8. `@extend_schema` doesn't expose the `sort` parameter in OpenAPI. The 11 valid fields are undiscoverable via Swagger or generated SDKs.
> 
> 
> Items 2 and 3 are worth a quick follow-up. The rest can be tackled whenever. Good work on the test coverage and the reusable sort infrastructure.

# Finding Groups Resources Review Response

Thanks for the review. For this patch, we would keep the scope focused and only consider adding the missing tests that materially improve confidence in the change:

* cross-tenant isolation coverage for `resources` and `latest_resources`
* one representative multi-sort test
* one representative dot-notation sort test

For the remaining comments, my view is:

## False `404` Framing

I would frame this a bit differently. The fix is not simply "move the existence check before user filters."

The intended behavior is:

1. validate request parameters first
2. check existence against the base tenant- and RBAC-scoped queryset, while ignoring user filters that may legitimately return an empty result
3. apply the drill-down filters and return either a filtered result or an empty list

That distinction matters because malformed requests must still return `400` before `404`.

## Double Sort Validation

This is intentional.

The first validation enforces the `400`-before-`404` contract. The second call turns the sort fields into the `ordering` used by the execution path. That is duplicate in-memory work, but it does not add database cost or meaningfully affect performance.

We are keeping it this way for clarity: validate first, then execute.

## Duplicated `status_order` / `severity_order`

This duplication is also intentional.

The ordering-only path and the full aggregation path have slightly different responsibilities, and keeping the expressions local makes each path easier to read in place. We could extract helpers, but that would trade a small amount of duplication for more indirection in an already dense area of the code.

For this change, the explicit duplication feels like the better clarity trade-off.

## `source.pointer` vs `source.parameter`

This is technically fair from a JSON:API point of view, but it looks like a broader API-wide pattern rather than something introduced by this patch.

Changing it here only would make this endpoint less consistent with the rest of the API. If we want to improve it, that should be done as a broader cleanup.

## OpenAPI Sort Discoverability

This also looks broader than this patch.

The schema is generated through `drf_spectacular_jsonapi`, and the same discoverability gap exists in other endpoints as well. If we want to expose supported sort fields better in Swagger or generated SDKs, that is worth improving, but it should be handled as a general API schema enhancement rather than inside this fix alone.

## Summary

So, from our side:

* yes to adding the most useful missing tests
* no to expanding this fix into broader refactors or API-wide cleanup items

That keeps the patch focused on the concrete regression it fixes, while still improving confidence where it matters most.

To keep moving, I would prefer to leave the implementation as it is after adding the most relevant missing tests, and then open a follow-up ticket for broader filter-combination coverage.

That additional test work is worth doing, but it is not limited to these two resource endpoints. The same kind of filter-combination coverage is relevant across several finding-group endpoints, so it makes more sense to track it as a separate testing task rather than folding it into this fix.

api/src/backend/api/v1/views.py
```diff
@@ -7336,26 +7403,127 @@ def _build_aggregated_queryset(self, finding_params, latest=False):
...
+        """Paginate and return resources.
+
+        Without sort: paginate lightweight resource IDs first, aggregate only the page.
+        With sort: build a lightweight ordering subquery (resource_id + sort keys),
+        paginate that, then aggregate full details only for the page.
+        """
+        sort_param = request.query_params.get("sort")
+
+        if sort_param:
+            ordering = self._validate_sort_fields(sort_param, self._RESOURCE_SORT_MAP)
```
josema-xyz, line 7452: Sort is parsed twice: `_validate_resource_sort()` validates but discards the result, then `_paginated_resource_response()` calls  `_validate_sort_fields()` again. The first one could be cached.

api/src/backend/api/tests/test_views.py
```diff
@@ -16011,6 +16011,191 @@ def test_resources_filters_applied(
...
+        """Nonexistent check_id with invalid sort on latest returns 400, not 404."""
+        response = authenticated_client.get(
+            reverse(
+                "finding-group-latest_resources",
+                kwargs={"check_id": "totally_fake_check"},
+            ),
+            {"sort": "invalid_field"},
+        )
+        assert response.status_code == status.HTTP_400_BAD_REQUEST
+
```
josema-xyz, line 16198: Missing cross-tenant test for `resources` and `latest_resources`. Need to verify that a `check_id` with findings only in another tenant returns `404`. The pattern is already in `test_finding_groups_rls_isolation`.

api/src/backend/api/v1/views.py
```diff
@@ -7253,7 +7269,58 @@ def _build_resource_aggregation(
                 ),
             )
             .filter(resource_id__isnull=False)
-            .order_by("resource_id")
+        )
+
+    # Annotations needed for each sort field (lightweight versions for ordering)
+    _RESOURCE_SORT_ANNOTATIONS = {
```
josema-xyz, line 7275: Case/When expressions for `status_order` and `severity_order` are duplicated between `_build_resource_aggregation` and `_RESOURCE_SORT_ANNOTATIONS`. Maybe they need to be extracted into shared factories?

api/src/backend/api/v1/views.py
```diff
@@ -7336,26 +7403,127 @@ def _build_aggregated_queryset(self, finding_params, latest=False):
...
 
-    def _sorted_paginated_response(self, request, aggregated_queryset):
+    def _sorted_paginated_response(
+        self,
+        request,
+        aggregated_queryset,
+        sort_field_map=None,
+        default_ordering=("-fail_count", "-severity_order", "check_id"),
+        post_processor=None,
+        serializer_class=None,
```
josema-xyz: The new params on `_sorted_paginated_response` (`sort_field_map`, `default_ordering`, `post_processor`, `serializer_class`) have no callers. Both call sites only pass `request` and `aggregated_qs`. Resources use `_paginated_resource_response`. Can the unused params be removed?

api/src/backend/api/v1/views.py
josema-xyz: Answered by @AdriiiPRodri in https://github.com/prowler-cloud/prowler/pull/10510#issuecomment-4153724356.

api/src/backend/api/tests/test_views.py
josema-xyz: Answered by @AdriiiPRodri in https://github.com/prowler-cloud/prowler/pull/10510#issuecomment-4153724356.

api/src/backend/api/v1/views.py
josema-xyz: Answered by @AdriiiPRodri in https://github.com/prowler-cloud/prowler/pull/10510#issuecomment-4153724356.

api/src/backend/api/v1/views.py
AdriiiPRodri: Done 👌🏼

PR #10525, opened 2026-03-31
feat(aws): add iam_policy_no_wildcard_marketplace_subscribe and iam_inline_policy_no_wildcard_marketplace_subscribe checks

danibarranqueroo:
### Context

Granting `aws-marketplace:Subscribe` on all resources (`*`) allows any principal with the policy to subscribe to any AWS Marketplace product, including expensive Amazon Bedrock foundation models, without governance controls. This can lead to uncontrolled costs from unauthorized model subscriptions, shadow AI usage outside compliance boundaries, and unapproved model deployments.

### Description

Two new IAM checks that evaluate custom and inline IAM policies for overly broad `aws-marketplace:Subscribe` permissions granted on `Resource: *`:

- **`iam_policy_no_wildcard_marketplace_subscribe`** — Inspects customer-managed IAM policies (follows the same pattern as `iam_policy_no_full_access_to_kms`)
- **`iam_inline_policy_no_wildcard_marketplace_subscribe`** — Inspects inline IAM policies on users/roles/groups (follows the same pattern as `iam_inline_policy_no_full_access_to_kms`)

Both checks:
- Match actions via wildcard patterns (`aws-marketplace:Subscribe`, `aws-marketplace:*`, `*`)
- Handle case-insensitive action matching
- Account for explicit Deny statements on `Resource: *` that negate the finding
- Correctly ignore Deny statements scoped to specific resources (they don't negate a wildcard Allow)

Compliance framework mappings added to 17 frameworks (AWS Well-Architected, NIST 800-53, FedRAMP, HIPAA, SOC2, ISO 27001, etc.) under least-privilege controls.

### Steps to review

1. Review the check implementations:
   - `prowler/providers/aws/services/iam/iam_policy_no_wildcard_marketplace_subscribe/`
   - `prowler/providers/aws/services/iam/iam_inline_policy_no_wildcard_marketplace_subscribe/`
2. Review the metadata files for correct severity, remediation, and compliance mappings
3. Review compliance framework mappings in `prowler/compliance/aws/`
4. Run the check tests:
   ```bash
   poetry run pytest tests/providers/aws/services/iam/iam_policy_no_wildcard_marketplace_subscribe/ tests/providers/aws/services/iam/iam_inline_policy_no_wildcard_marketplace_subscribe/ -v
   ```
5. **Run the checks against a real environment** (if possible):
   ```bash
   prowler aws --check iam_policy_no_wildcard_marketplace_subscribe iam_inline_policy_no_wildcard_marketplace_subscribe
   ```

### Checklist

<details>

</details>

#### SDK/CLI
    - If so, do we need to update permissions for the provider? **No** — uses existing IAM permissions (`iam:ListPolicies`, `iam:GetPolicyVersion`, `iam:ListRolePolicies`, `iam:GetRolePolicy`, etc.) already required by other IAM checks.

### License


HugoPBrito (commented):
The new helper does not handle `NotAction`, so it can miss policies that also allow `aws-marketplace:Subscribe` through `Allow` + `NotAction` on `Resource: "*"`. In `prowler/providers/aws/services/iam/lib/policy.py`, the logic only inspects `Action` and `Resource`, but not the equivalent `NotAction` case that `check_full_service_access` already supports.

Example that would slip through: `{"Effect":"Allow","NotAction":"ec2:Describe*","Resource":"*"}`.

I would recommend reusing the shared `check_full_service_access` logic or extending this helper to cover `NotAction`, and adding tests for that scenario in both checks.

HugoPBrito (changes requested):
Please review the comment and let me know.

prowler/providers/aws/services/iam/lib/policy.py
```diff
@@ -997,6 +997,85 @@ def is_codebuild_using_allowed_github_org(
         return False, None
 
 
+def policy_allows_marketplace_subscribe_on_all_resources(
```
HugoPBrito, line 1000: The helper treats any Deny on `Resource: "*"` as if it globally cancels the finding, but it does not evaluate Condition. That leads to false `PASS` results. A real example is a policy with `Allow aws-marketplace:Subscribe on *` plus `Deny aws-marketplace:Subscribe on *` only under a specific condition.

In AWS, that deny does not remove the permission in every context, but this implementation does, because it adds the action to `denied_on_all` unconditionally and then computes `allowed_on_all` - `denied_on_all`.

If the goal is to model effective permissions, this is too aggressive and can hide genuinely risky policies.

HugoPBrito (approved):

PR #10529, opened 2026-03-31
fix(oci): scan identity in known valid region

### LINKED ISSUE #10528: [OCI] Identity should scan known region
### Issue search

Prowler CLI/SDK

Not applicable

### Steps to Reproduce

prowler oraclecloud --region us-phoenix-1 us-chicago-1 --service identity ...

### Expected behavior

The script would be expected to detect identity domains within all regions and check identity compliance

The script defaults to us-ashburn-1 and only scans that region's identity domains

### Environment Resource

Container instance

### OS used

Ubi9 Redhat

### Prowler version

5.23.0

### Python version

3.11

### Pip version

26.0.1

### Context

Scanning the home region which defaults to us-ashburn-1 breaks use cases where us-ashburn-1 is not supported and possibly ignores identity domains in other regions

rchotacode:
### Context

Fix #10528 

### Description

Scans found region for identity misconfigurations instead of generally defaulting to us-ashburn-1

### Steps to review

Scanning any tenancy with an identity domain outside us-ashburn-1 or not supporting us-ashburn-1

### Checklist

<details>

- [X] This feature/issue is listed in [here](https://github.com/prowler-cloud/prowler/issues?q=sort%3Aupdated-desc+is%3Aissue+is%3Aopen) or roadmap.prowler.com

</details>

#### SDK/CLI

#### UI
- [X] All issue/task requirements work as expected on the UI

#### API
- [X] All issue/task requirements work as expected on the API

### License


prowler/providers/oraclecloud/services/identity/identity_service.py
```diff
@@ -61,8 +61,8 @@ def __list_users__(self, regional_client):
             regional_client: Regional OCI client
         """
         try:
-            # Identity is a global service, use home region
-            if regional_client.region not in self.provider.identity.region:
+            # Only use one region for global users
+            if regional_client.region is not self.provider.regions[0].key:
```
HugoPBrito: Bug: `is not` checks **object identity**, not value equality. Two strings with the same content are not guaranteed to be the same object in Python, so this guard can silently fail — either never triggering (method runs in every region) or always triggering (method never runs).

This same issue affects `__list_groups__`, `__list_policies__`, `__list_dynamic_groups__`, and `__get_password_policy__`.

```suggestion
            if regional_client.region != self.provider.regions[0].key:
```

prowler/providers/oraclecloud/services/identity/identity_service.py
```diff
@@ -551,7 +553,11 @@ def __list_domain_password_policies__(self, regional_client):
     def __get_password_policy__(self, regional_client):
         """Get the password policy for the tenancy."""
         try:
-            if regional_client.region not in self.provider.identity.region:
+            # Only use one region for global password policies
+            if regional_client.region is not self.provider.regions[0].key:
```
HugoPBrito: `self.provider.regions[0]` is an `OCIRegion` dataclass, not a string. This will log something like `OCIRegion(key='us-phoenix-1', name='...', is_home_region=True)` instead of just the region name.

Also minor: "its" → "it's".

```suggestion
                    f"Skipping {regional_client.region} as it's not {self.provider.regions[0].key}"
```

prowler/providers/oraclecloud/services/identity/identity_service.py
```diff
@@ -578,10 +584,6 @@ def __get_password_policy__(self, regional_client):
     def __search_root_compartment_resources__(self, regional_client):
         """Search for resources in the root compartment using OCI Resource Search."""
         try:
-            # Search is a global service, use home region
-            if regional_client.region not in self.provider.identity.region:
-                return
```
HugoPBrito, line 583: The region guard was removed entirely from `__search_root_compartment_resources__` (and `__search_active_non_root_compartments__` below) without a replacement. Resource Search is a global service — without a guard, this will execute in every audited region, causing redundant API calls and potentially duplicate results.

Needs a guard similar to the other methods:
```python
if regional_client.region != self.provider.regions[0].key:
    return
```

prowler/providers/oraclecloud/oraclecloud_provider.py
```diff
@@ -127,10 +127,8 @@ def __init__(
 
         logger.info("Initializing OCI provider ...")
 
-        # Check if the configuration is scanning a single region
-        single_region = None
-        if region:
-            single_region = list(region)[0] if len(region) == 1 else None
+        # Set initial region if specified
+        single_region = list(region)[0] if region else None
```
HugoPBrito: The original code only picked a single region when exactly one was provided (`len(region) == 1`), and passed `None` otherwise (falling back to the tenancy's home region for session setup). This change always picks the first region even when multiple are specified (e.g. `--region us-phoenix-1 us-chicago-1`), which forces the OCI SDK session to that region instead of the home region.

```suggestion
        single_region = list(region)[0] if region and len(region) == 1 else None
```

prowler/CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ All notable changes to the **Prowler SDK** are documented in this file.
 
 - Added `internet-exposed` category to 13 AWS checks (CloudFront, CodeArtifact, EC2, EFS, RDS, SageMaker, Shield, VPC) [(#10502)](https://github.com/prowler-cloud/prowler/pull/10502)
 - Minimum Python version from 3.9 to 3.10 and updated classifiers to reflect supported versions (3.10, 3.11, 3.12) [(#10464)](https://github.com/prowler-cloud/prowler/pull/10464)
+- Oracle cloud identity scans now scan known or supplied regions to better support non ashburn tenancys [(#10528)](https://github.com/prowler-cloud/prowler/pull/10529)
```
HugoPBrito: This is a bugfix, so the entry should go under `### 🐞 Fixed` instead of the general changes section. Also minor typo: "tenancys" → "tenancies".

HugoPBrito (changes requested):
Thanks for tackling this — the issue is real and the approach (using the provider's known regions instead of hardcoding `us-ashburn-1`) makes sense. However, there are a few issues that need addressing before this can be merged:

1. **`is not` vs `!=`** — all 5 string comparisons use identity check instead of equality. This is the most critical issue.
2. **Removed guards on search methods** — `__search_root_compartment_resources__` and `__search_active_non_root_compartments__` lost their region guard entirely, which will cause redundant calls and potentially duplicate results.
3. **Provider change breaks multi-region** — `single_region` now always picks the first region even when multiple are specified.
4. **Missing test updates** — Codecov shows 41% patch coverage with 7 uncovered lines.

prowler/providers/oraclecloud/oraclecloud_provider.py
rchotacode: The oci session initially is created using the default of us-ashburn-1, not the home region in many cases. This makes the session in a known region which is getting scanned instead of the default us-ashburn-1 if regions are specified 
https://github.com/prowler-cloud/prowler/blob/abaacd7dbf4e7fed579b19f89c2b4d5d02d0ac38/prowler/providers/oraclecloud/oraclecloud_provider.py#L455-L457

rchotacode: This also will not break multi region, the regional clients still scan multiple regions. The initial session region is changed, but regional clients are still made. Before, in multi region scans, this value was for all intents and purposes set to us-ashburn-1.

rchotacode:
> 4. **Missing test updates** — Codecov shows 41% patch coverage with 7 uncovered lines.
I've never used codecov before, are there any specific changes I should implement to resolve this?

HugoPBrito:
> > 4. **Missing test updates** — Codecov shows 41% patch coverage with 7 uncovered lines.
> >    I've never used codecov before, are there any specific changes I should implement to resolve this?

You should be able to see the uncovered lines [here](https://app.codecov.io/gh/prowler-cloud/prowler/pull/10529?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=prowler-cloud).

Based on that, you can try to add some tests to cover them.

rchotacode:
Im not sure how to test these functions since they dont seem to be called in any tests as the __init__ is mocked for the identity service

HugoPBrito:
Thanks for the updates. I checked Oracle’s official docs to verify the region behavior before asking for more changes.

For classic IAM global resources (`users`, `groups`, `policies`, `dynamic groups`, etc.), Oracle documents them as global tenancy resources that are available across subscribed regions, even though their master definition lives in the home region. So replacing the hardcoded `us-ashburn-1` is directionally correct.

  The remaining logic issue is with **identity domains**.

  Oracle documents that domains visible in a region can be either:
  - homed in that region, or
  - replicated to that region

  Because of that, this check is too restrictive:

  ```python
  if regional_client.region != domain.home_region:
      continue
```

  A domain can be valid in the current region even when that region is not its home_region.

  So I think the logic should be:

  - use a known valid region for global IAM resources, ideally the tenancy home region when available
  - do not rely on `provider.regions[0]` as a proxy for “correct global region”
  - do not filter domains/password policies only by `home_region`
  - instead, trust the domains returned by the current regional endpoint

About your testing question:

> Im not sure how to test these functions since they dont seem to be called in any tests as the init is mocked for the identity service

The approach you started using is fine. You can keep mocking `Identity.__init__`, set the needed attributes manually, mock `__get_client__` / OCI SDK calls, and call `__list_domains__` or `__list_domain_password_policies__` directly.

What’s still missing is a positive test that proves the intended behavior, not just skip behavior. For example:

  - a domain returned by the current region is processed even if `domain.home_region != regional_client.region`
  - password policies are fetched for domains available in the current region
  - global IAM resources still run once in the chosen valid region

So from my side the main remaining issue is the identity-domain region logic, and the formatting, since the sdk test are failing.

Happy to help with anything if needed, and thanks for all the effort you've put into this.

**Comment by rchotacode** [comment]
Thanks for the feedback Hugo! I went in and switched to adding a new property in the oraclecloud provider for the home region. It grabs the first home region in _regions which has is_home_region = True. I also changed the code for filtering identity domains to only remove duplicate identity domains outside of their home region. This will ensure that all identity domains are scanned without having duplicate findings for identity domains replicated, and that the findings for those identity domains will be tagged with their home regions if possible. All the logic for filtering has been moved to list domains so there's no longer a need for testing the password policy listing. I created a test case which uses 3 identity domains, all in us-ashburn-1. One is replicated from a domain in ashburn and another would not otherwise be included as its replicated from chicago. The test makes sure that the chicago and ashburn domains are kept and the phoenix is discarded as it's already included.

**Comment by HugoPBrito** [comment]
Thanks for the iteration — the direction is right, but I think there are a few blocking issues left, plus a couple of regressions the new test doesn't exercise. Splitting them out for clarity in separate comments below.

**Critical issue 1 — `__list_domains__`: indices into the wrong list**

```python
to_remove = []
for i, domain in enumerate(domains):           # `domains` = SDK response for one compartment
    if any(d.id == domain.id and d is not domain for d in domains):
        if domain.home_region != regional_client.region:
            to_remove += [i]
for i in sorted(to_remove, reverse=True):
    logger.info("Skipping replicated domain in non-home region: %s", domain.id)
    del self.domains[i]                        # `self.domains` is accumulated across compartments AND regions
```

- The indices come from `enumerate(domains)` (per-compartment SDK response), but `del self.domains[i]` deletes from the accumulated list — they don't line up as soon as more than one compartment or region is involved, so this deletes arbitrary items.
- `domain.id` in the log uses the loop variable left over from the previous `for`, so every "skipping" log line points at the last enumerated domain.
- Duplicate detection only looks **within a single SDK response**. The real case (a domain replicated to another region, returned by `list_domains` in both `us-ashburn-1` and `us-phoenix-1`) is never detected here, because each `__list_domains__` call only sees its own region's response.

HugoPBrito:
**Critical issue 2 — `__list_domain_password_policies__` runs once per region with no dedupe**

The `if regional_client.region not in self.provider.identity.region: return` guard was removed, but `__threading_call__` invokes the method once per regional client. For each region we now iterate `self.domains` and `domain.password_policies.append(...)` — every domain ends up with its policies duplicated N times (N = number of audited regions). Anything downstream iterating `password_policies` will see N copies.

HugoPBrito:
**Critical issue 3 — `_home_region` fallback returns an `OCIRegion`, not a string**

```python
self._home_region = next(
    (region.key for region in self._regions if region.is_home_region),
    self._regions[0],     # OCIRegion object, not a string
)
```

When no element of `_regions` has `is_home_region=True` (e.g. user passes `--region us-phoenix-1` on a tenancy whose home lives elsewhere), `_home_region` becomes a dataclass instance. Every `regional_client.region != self.provider.home_region` then evaluates `True` (string vs dataclass), so **every identity listing silently returns** and identity findings come back empty with no error. Should be `self._regions[0].key`.

As a side effect, the `logger.info(f"Home region is: {self._home_region}")` line will render the `OCIRegion` repr in that fallback path.

HugoPBrito:
**Test coverage**

`test_list_domains_skipped_outside_home` only calls `__list_domains__` once with `regional_client.region="us-ashburn-1"`, so:

- It doesn't exercise the scenario this PR is fixing (multi-region scan on a non-Ashburn tenancy).
- It can't catch the indices issue, because with `self.domains` starting empty and a single call, the SDK indices happen to line up with `self.domains` indices.
- It doesn't cover the `home_region` fallback.
- It doesn't cover password-policy duplication.

It would help to invoke `__list_domains__` (and then `__list_domain_password_policies__`) **twice** with different `regional_client.region` values, then assert one entry per `domain.id` in `self.domains` and no duplicates in `domain.password_policies`.

HugoPBrito:
**Suggested approach**

Dedupe **at insert time** instead of fixing up indices after the fact:

```python
for domain in domains:
    existing = next((d for d in self.domains if d.id == domain.id), None)
    if existing is not None:
        # Prefer the entry from the domain's home region
        if domain.home_region == regional_client.region:
            self.domains.remove(existing)
        else:
            continue
    self.domains.append(IdentityDomain(...))
```

For password policies, either restore a guard so they're fetched once per domain, or make the call idempotent (clear `domain.password_policies` before populating, or skip if already populated).

HugoPBrito:
**Minor**

- The new imports in `oraclecloud_provider_test.py` aren't alphabetical (`OCISession, OCIIdentityInfo, OCIRegion`) — `ruff/isort` will flag it.
- Changelog typo: `tenancys` → `tenancies`.

rchotacode:
Thanks for the feedback Hugo! I've implemented the changes requested so that domains are filtered at insertion with the provided code, the test now tests both password policies and list_domains with multiple regional clients. Home region's default is fixed along with the other issues listed. Let me know if there's anything else I should look into!

HugoPBrito (changes requested):
Found a regression while testing this branch. With an empty `_regions` list the new initializer at `oraclecloud_provider.py:164-167` crashes the entire scan:

```
INFO: Found 0 subscribed regions
CRITICAL: IndexError[347]: list index out of range
```

Root cause: the default value passed to `next()` is evaluated eagerly, so `self._regions[0].key` raises `IndexError` whenever `_regions` is empty (e.g. when `--region`/`-f` filters don't match any subscribed region):

```python
self._home_region = next(
    (region.key for region in self._regions if region.is_home_region),
    self._regions[0].key,   # IndexError when self._regions == []
)
```

Pre-PR this same scenario completed silently with 0 findings; now it aborts the whole run. Could you guard against the empty-list case and surface a clear error (e.g. "no subscribed regions match the requested filter") instead of letting it crash here?

HugoPBrito (commented):
Could you also share evidence that the fix behaves as expected in a tenancy whose home region is **not** `us-ashburn-1`? Ideally a full `--log-level INFO` run of `--service identity` showing the new `Home region is: <non-ashburn-region>` log line and identity findings being produced where the previous behavior would have returned none. We aren't able to validate the original scenario end-to-end on our side, so concrete evidence from your test environment would be very helpful.

rchotacode:
Thanks for the feedback again Hugo! I have fixed the edge case described, but am unfortunately not comfortable showing test results for any of our tenancies. I setup a free tier personal OCI account with a home region of us-phoenix-1.  This image shows that was detected. 
image

rchotacode:
image
Here is evidence of a home region being detected in a multi region environment

prowler/CHANGELOG.md
```diff
@@ -103,6 +103,7 @@ All notable changes to the **Prowler SDK** are documented in this file.
 
 - OCI mutelist support: pass `tenancy_id` to `is_finding_muted` and update `oraclecloud_mutelist_example.yaml` to use `Accounts` key [(#10566)](https://github.com/prowler-cloud/prowler/pull/10566)
 - `return` statements in `finally` blocks replaced across IAM, Organizations, GCP provider, and custom checks metadata to stop silently swallowing exceptions [(#10102)](https://github.com/prowler-cloud/prowler/pull/10102)
+- Oracle cloud identity scans now scan known or supplied regions to better support non ashburn tenancies [(#10528)](https://github.com/prowler-cloud/prowler/pull/10529)
```
HugoPBrito: Please, move this entry to a new `5.25.3 (Prowler UNRELEASED)` section in the changelog.

HugoPBrito (changes requested):
That will do the job, thanks!

Just take into account this last comment an it will be ready to be merged.

Thank you for the effort put into this.

rchotacode:
Changelog moved, thanks Hugo!

HugoPBrito (dismissed):

HugoPBrito (approved):

Issue #10567, opened 2026-04-02
[OCI] Add CIS 3.1 1.15 compliance scans


rchotacode:
### Feature search

- [x] I have searched the existing issues and this feature has not been requested yet or is already in our [Public Roadmap](https://roadmap.prowler.com/roadmap)

### Which component would this feature affect?

Prowler CLI/SDK

### Related to specific cloud provider?

Not provider-specific

### New feature motivation

CIS 3.1 Standard has a check 1.15 which is currently not implemented

### Solution Proposed

A new check should be created to detect file-storage service level admins that have not been scoped properly

### Use case and benefits

Clients trying to comply with CIS 3.1 standards will be able to scan for these misconfigurations

### Describe alternatives you've considered

Manually going through policies is a viable alternative

HugoPBrito:
Hi @rchotacode,

Thanks for opening this and for your contribution! We'll review it as soon as possible.

PR #10568, opened 2026-04-02
feat(oci): add `identity_storage_service_level_admins_scoped` check for CIS 3.1 1.15 compliance

rchotacode:
### Context

CIS 3.1 include 1.15 which requires service level storage admins to be scoped to remove deletion permissions. 
Feat #10567 

### Description

Policies such as ```manage file-systems resources``` enable deletion at the service level. According to OCI CIS 3.1 1.15, these policies should not be service level or have the condition ```where request.permission!='*SERVICE*_DELETE'```. 

### Steps to review

Running prowler oraclecloud --check identity_storage_service_level_admins_scoped

### Checklist

<details>

- [X] This feature/issue is listed in [here](https://github.com/prowler-cloud/prowler/issues?q=sort%3Aupdated-desc+is%3Aissue+is%3Aopen) or roadmap.prowler.com

</details>

#### SDK/CLI

#### UI
- [X] All issue/task requirements work as expected on the UI

#### API
- [X] All issue/task requirements work as expected on the API

### License


prowler/providers/oraclecloud/services/identity/identity_storage_service_level_admins_scoped/identity_storage_service_level_admins_scoped.py
```diff
@@ -0,0 +1,109 @@
...
+                # Only check groups
+                if not statement_upper.startswith("ALLOW GROUP"):
+                    continue
+
+                # Check for "allow group ... to manage *file-service* resources" without restriction (not restricted to certain buckets or non delete)
+                if any(
+                    f"MANAGE {global_storage_policy}" in statement_upper
+                    for global_storage_policy in storage_policies
+                ):
+                    if "WHERE" not in statement_upper:
```
HugoPBrito, line 52: This logic is still too permissive for CIS 3.1 control 1.15.

The control does not require any `WHERE` clause; it requires a condition that excludes delete access for the storage resource being managed. Right now a statement such as `where request.region="iad"` or `where target.bucket.name="x"` would pass, even though it still allows deletion.

Please update the check so it only treats a statement as compliant when it explicitly excludes the relevant delete permission for that resource, for example `request.permission!='VOLUME_DELETE'`, `request.permission!='FILE_SYSTEM_DELETE'`, `request.permission!='OBJECT_DELETE'`, etc. This is the core requirement of CIS 1.15: service-level admins may manage the service, but they must not be able to delete the resources they manage.

prowler/providers/oraclecloud/services/identity/identity_storage_service_level_admins_scoped/identity_storage_service_level_admins_scoped.py
```diff
@@ -0,0 +1,109 @@
...
+                        break
+                if any(
+                    f"MANAGE {base_storage_policy}" in statement_upper
+                    for base_storage_policy in all_base_policies
+                ):
+                    if "WHERE" not in statement_upper:
+                        has_violation = True
+                        offending_statements.append(statement)
+                        break
+                if "MANAGE ALL-RESOURCES" in statement_upper:
```
HugoPBrito, line 64: I would also revisit the `manage all-resources` handling here.

For CIS 1.15, broad statements should not pass just because they contain a `WHERE`. The implementation should only pass when it can prove that delete is excluded for the storage resources covered by the statement. Otherwise this can create false PASS results for policies that remain non-compliant with the benchmark.

prowler/providers/oraclecloud/services/identity/identity_storage_service_level_admins_scoped/identity_storage_service_level_admins_scoped.py
```diff
@@ -0,0 +1,109 @@
...
+                resource=policy,
+                region=region,
+                resource_id=policy.id,
+                resource_name=policy.name,
+                compartment_id=policy.compartment_id,
+            )
+
+            if has_violation:
+                report.status = "FAIL"
+                report.status_extended = f"Policy '{policy.name}' grants 'manage' permissions with delete. Service-level storage administrators should be created with delete permissions.\n{offending_statements}"
```
HugoPBrito: The FAIL message is currently inverted relative to CIS 1.15.

The control requires storage service-level admins to be able to manage resources without delete permissions. This message says they should be created with delete permissions, which is the opposite of the benchmark requirement.

Please update the wording so the remediation/result text clearly says delete must be excluded, not granted.

prowler/providers/oraclecloud/services/identity/identity_storage_service_level_admins_scoped/identity_storage_service_level_admins_scoped.metadata.json
```diff
@@ -0,0 +1,37 @@
...
+    "Description": "**OCI IAM policies** are reviewed for **delete permissions**, specifically statements granting `manage` to storage services without scoping to avoid deletion. Only **active policies** are considered; the default tenant admin policy is excluded.",
+    "Risk": "Creating service-level administrators without the ability to delete the resource they are managing helps in tightly controlling access to Oracle Cloud Infrastructure (OCI) services by implementing the separation of duties security principle.",
+    "RelatedUrl": "",
+    "AdditionalURLs": [
+        "https://docs.oracle.com/en-us/iaas/Content/Identity/Concepts/policygetstarted.htm",
+        "https://docs.oracle.com/en-us/iaas/Content/Identity/getstarted/identity-domains.htm"
+    ],
+    "Remediation": {
+        "Code": {
+            "CLI": "oci iam policy update --policy-id <policy-ocid> --statements '[\"Allow group <GroupName> to manage objects where request.permission!='OBJECT_DELETE'\"]'",
```
HugoPBrito: Please align the metadata remediation examples with the CIS 1.15 wording and examples.

A couple of things to fix here:
- the examples should be framed around excluding delete permissions for the managed storage resource
- `request.permission` should be used consistently
- the Terraform example currently uses `request.PERMISSIONS`, which does not match the benchmark examples
- if possible, include examples that mirror the control more closely for the different storage resources (`VOLUME_DELETE`, `FILE_SYSTEM_DELETE`, `OBJECT_DELETE`, etc.), not just a single generic example

Since this metadata is what users will read for remediation, it should map directly to the benchmark language.

tests/providers/oraclecloud/services/identity/identity_storage_service_level_admins_scoped/identity_storage_service_level_admins_scoped_test.py
HugoPBrito, line 1: This new check still needs real unit test coverage.

At the moment the test file is empty, so we are not validating the CIS-specific behavior of the parser. Please add tests for at least:
- FAIL: `manage <storage-resource>` with no delete exclusion
- PASS: `manage <storage-resource>` with the correct `request.permission!='*_DELETE'` condition
- FAIL: a statement with an unrelated `WHERE` clause that does not exclude delete
- inactive policies
- `Tenant Admin Policy`

This is especially important here because the correctness of the check depends on parsing OCI policy statements exactly as required by CIS 1.15.

prowler/providers/oraclecloud/services/identity/identity_storage_service_level_admins_scoped/identity_storage_service_level_admins_scoped.py
rchotacode: I originally implemented this behavior, but looking at the CIS 1.15 examples it seems service level admins refer to unrestricted manage access to storage services. Other CIS benchmarks also seem to flag it as such. I would think that ```target.bucket.name="x"``` would be a resource level of control, not a service. 

These are the example policies adjusted to fit within the control within its definition document:

Example policies for global/tenant level for block volume service-administrators: 
```
Allow group VolumeUsers to manage volumes in tenancy where request.permission!='VOLUME_DELETE' 
Allow group VolumeUsers to manage volume-backups in tenancy where request.permission!='VOLUME_BACKUP_DELETE
```

Example policies for global/tenant level for file storage system service-administrators:
```
Allow group FileUsers to manage file-systems in tenancy where request.permission!='FILE_SYSTEM_DELETE' 
Allow group FileUsers to manage mount-targets in tenancy where request.permission!='MOUNT_TARGET_DELETE' 
Allow group FileUsers to manage export-sets in tenancy where request.permission!='EXPORT_SET_DELETE'
```

Example policies for global/tenant level for object storage system service-administrators: 
```
Allow group BucketUsers to manage objects in tenancy where request.permission!='OBJECT_DELETE' 
Allow group BucketUsers to manage buckets in tenancy where request.permission!='BUCKET_DELETE'
```

Each of these allows manage for a file service storage, but doesn't restrict the permissions at all without the request.permission!='X_DELETE'. In addition, the reference here https://docs.oracle.com/en/solutions/oci-best-practices/protect-data-rest1.html#GUID-EBE9212C-78B0-4795-9FB5-59DB0E91E106 linked in the control says that the permissions should be give "to only the users who need these privileges," which implies that some users should still be able to have delete permissions. I interpreted this as meaning that the CIS benchmark was to limit unrestricted delete permissions given to admins over all storage services.

PR #10574, opened 2026-04-06
chore(ui): unset GIT_WORK_TREE in pre-commit hook

pfe-nazaries:
## Summary

- Fixes pre-commit hook path resolution when Python `pre-commit` framework sets `GIT_WORK_TREE`, `GIT_DIR`, and `GIT_INDEX_FILE` to a temp staging area
- `git rev-parse --show-toplevel` was returning `prowler/ui` instead of `prowler`, causing `cd "$GIT_ROOT/ui"` to resolve to `prowler/ui/ui` (non-existent)
- `git -C "$GIT_ROOT" diff --cached` was reading the wrong index, so quality gates could silently pass against the wrong staged file set
- Unsetting all `GIT_*` env vars (`GIT_WORK_TREE`, `GIT_DIR`, `GIT_INDEX_FILE`, `GIT_PREFIX`, `GIT_COMMON_DIR`, `GIT_OBJECT_DIRECTORY`) after `set -e` restores correct behavior without side effects

## Test plan

- [ ] Reproduce the full failure mode the Python pre-commit framework creates:
  ```bash
  env GIT_WORK_TREE=/tmp/fake GIT_DIR=/tmp/fake/.git GIT_INDEX_FILE=/tmp/fake/index \
    bash ui/.husky/pre-commit
  ```
- [ ] Or run the actual framework hook: `pre-commit run ui-checks --verbose --all-files`
- [ ] Commit a UI file change and verify the husky pre-commit hook completes successfully
- [ ] Verify healthcheck, tests, and build steps still run from the correct directory

### DIFF
--- a/ui/.husky/pre-commit
+++ b/ui/.husky/pre-commit
@@ -6,6 +6,9 @@
 
 set -e
 
+# Ensure git resolves the real repo root, not a pre-commit framework override
+unset GIT_WORK_TREE
+
 # Colors
 RED='\033[0;31m'
 GREEN='\033[0;32m'
--- a/ui/.husky/pre-commit
+++ b/ui/.husky/pre-commit
@@ -6,8 +6,11 @@
 
 set -e
 
-# Ensure git resolves the real repo root, not a pre-commit framework override
-unset GIT_WORK_TREE
+# The Python pre-commit framework (see .pre-commit-config.yaml, hook "ui-checks")
+# exports GIT_WORK_TREE, GIT_DIR, and GIT_INDEX_FILE pointing to its temp staging
+# area. Unset them so git commands below resolve against the real repo and index.
+# See: https://github.com/prowler-cloud/prowler/pull/10574
+unset GIT_WORK_TREE GIT_DIR GIT_INDEX_FILE GIT_PREFIX GIT_COMMON_DIR GIT_OBJECT_DIRECTORY
 
 # Colors
 RED='\033[0;31m'
--- a/ui/CHANGELOG.md
+++ b/ui/CHANGELOG.md
@@ -21,7 +21,6 @@ All notable changes to the **Prowler UI** are documented in this file.
 - Deleting the active organization now switches to the target org before deleting, preventing JWT rejection from the backend [(#10491)](https://github.com/prowler-cloud/prowler/pull/10491)
 - Clear Filters now resets all filters including muted findings and auto-applies, Clear all in pills only removes pill-visible sub-filters, and the discard icon is now an Undo text button [(#10446)](https://github.com/prowler-cloud/prowler/pull/10446)
 - Send to Jira modal now dynamically fetches and displays available issue types per project instead of hardcoding `"Task"`, fixing failures on non-English Jira instances [(#10534)](https://github.com/prowler-cloud/prowler/pull/10534)
-- Pre-commit hook now unsets `GIT_WORK_TREE` to fix path resolution when the variable is inherited from parent processes [(#10574)](https://github.com/prowler-cloud/prowler/pull/10574)
 
 ---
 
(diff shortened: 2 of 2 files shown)


Alan-TheGentleman (changes requested):
Hey, ran a deep review on this. The fix is in the right direction but it does not finish the job, and I think it should not merge as-is.

## The actual blocker

The Python `pre-commit` framework does not only export `GIT_WORK_TREE`. It also exports `GIT_DIR` and `GIT_INDEX_FILE` (and sometimes `GIT_PREFIX`). Unsetting just `GIT_WORK_TREE` leaves `GIT_INDEX_FILE` pointing at the framework's temp staging index.

That matters because lines 42, 168 and 171 do:

```bash
git -C "$GIT_ROOT" diff --cached --name-only ...
```

`git -C` only changes the working directory of the git invocation. It does NOT clear `GIT_DIR` or `GIT_INDEX_FILE` from the environment, and those env vars take precedence over auto-discovery. So `--cached` can still read the wrong index, and the hook will report on the wrong staged file set. Worst case: a quality gate that thinks it ran but ran against the wrong data, and silently passes a commit that should fail.

Suggested fix:

```bash
# Python pre-commit framework exports GIT_WORK_TREE/GIT_DIR/GIT_INDEX_FILE pointing
# to its temp staging area. Unset them so 'git rev-parse --show-toplevel' and every
# 'git -C "$GIT_ROOT" diff --cached' below resolve against the real repo and index.
unset GIT_WORK_TREE GIT_DIR GIT_INDEX_FILE GIT_PREFIX GIT_COMMON_DIR GIT_OBJECT_DIRECTORY
```

## Other things worth fixing before merge

1. **Changelog entry is in the wrong place.** This is dev tooling (a husky hook), not a user-facing UI fix. `ui/CHANGELOG.md` tracks shipped UI behavior that end users care about. End users do not run our husky hook. Either drop the entry and add the `no-changelog` label, or move it to an internal/tooling section if one exists.

2. **The test plan does not reproduce the real failure mode.** Running `GIT_WORK_TREE=$(pwd)/ui git rev-parse --show-toplevel` only sets one var, but the actual bug happens when the Python pre-commit framework sets all three. A proper repro should set `GIT_DIR`, `GIT_INDEX_FILE` and `GIT_WORK_TREE` together, or invoke `pre-commit run` against the husky hook directly. As written, the test plan will pass even if the underlying issue is not fully fixed.

3. **Comment will age poorly.** The current comment does not say which framework, why it sets the var, or link this PR. Six months from now someone will think it looks redundant and remove it. Worth being explicit that it is the Python `pre-commit` framework and linking #10574.

## Nice-to-haves (not blocking)

- Commit scope: `fix(ui)` implies a user-facing UI fix. This is dev tooling, so `chore(ui)` or `build(ui)` would be more accurate and avoid polluting release notes.
- Architectural: husky and Python pre-commit are both wired to `.git/hooks/pre-commit` in this repo. They are fighting over the same slot, and this PR papers over the symptom. At some point it would be worth consolidating into one framework (e.g. invoke the husky logic as a `local` hook from `.pre-commit-config.yaml`) so this class of bug stops coming back.

Once `GIT_DIR`/`GIT_INDEX_FILE` are unset and the changelog entry is sorted out, happy to re-review. Good catch on the root cause, just needs to go one step further.

Alan-TheGentleman (approved):
Re-reviewed. The technical blocker from my previous review is fully addressed — `unset` now covers `GIT_WORK_TREE`, `GIT_DIR`, `GIT_INDEX_FILE`, `GIT_PREFIX`, `GIT_COMMON_DIR` and `GIT_OBJECT_DIRECTORY`, so the `git -C "$GIT_ROOT" diff --cached` calls on lines 42, 168 and 171 will resolve against the real index instead of the Python pre-commit framework's temp staging. The inline comment now explicitly names the framework and links this PR, so it will age well. Changelog routing is sorted via the `no-changelog` label. ✅

Approving — two non-blocking nits I'd still suggest before merge:

**Nit 1 — Test plan does not reproduce the real failure mode.** Running `GIT_WORK_TREE=$(pwd)/ui git rev-parse --show-toplevel` only sets one env var. The actual bug requires `GIT_DIR` and `GIT_INDEX_FILE` to also be set (which is exactly what the Python pre-commit framework does). As written, the test plan would pass even if the fix were incomplete. A more faithful repro:

```bash
# Reproduce the env the Python pre-commit framework creates
env GIT_WORK_TREE=/tmp/fake GIT_DIR=/tmp/fake/.git GIT_INDEX_FILE=/tmp/fake/index \
  bash ui/.husky/pre-commit
# Or better: run the actual framework hook
pre-commit run ui-checks --verbose --all-files
```

**Nit 2 — Commit scope.** `fix(ui):` implies a user-facing UI fix. This is dev tooling (a husky hook), not shipped UI behavior. `chore(ui):` or `build(ui):` would be more accurate and keeps release notes clean. Low-cost rename before merge.

Neither of these blocks merge. Good iteration on the root cause — the `GIT_INDEX_FILE` part was subtle and it's now correct.

PR #10590, opened 2026-04-07
feat(ui): improvements over findings groups feature

alejandrobailo:
### Context

Finding groups drill-down was limited to FAIL-only resources, preventing users from seeing PASS resources in a group. Additionally, the scans page showed raw scan IDs instead of a proper selector, and muting UX in the resource drawer lacked optimistic feedback.

### Description

- **Remove FAIL-only filter**: Stop hardcoding `filter[status]=FAIL` on finding group resource endpoints so PASS resources also appear in the drill-down
- **Allow expanding PASS-only groups**: Use `resourcesTotal > 0` instead of `resourcesFail > 0` for expand eligibility
- **Normalize status filters**: Add `normalizeFindingGroupResourceFilters` to collapse single-value `status__in` into `status` for API compatibility
- **Extract selection logic**: Create `canMuteFindingGroup` and `canMuteFindingResource` modules — disable mute checkbox for PASS-only groups and PASS resources
- **Optimistic mute in drawer**: Add `isMutedOptimistic` state to `OtherFindingRow` for instant UI feedback after muting
- **Scan selector on /scans page**: Replace raw scan ID badge with the `ScanSelector` dropdown (reused from compliance) when navigating via "View Scan"
- **Scan date filter resolution**: New `resolveFindingScanDateFilters` utility that translates `filter[scan__in]` into required `inserted_at` date filters
- **Filter display improvements**: Extract `getFindingsFilterDisplayValue` utility, resolve provider account aliases and scan labels in filter badges
- **Default sort change**: Prioritize `fail_count` over `severity` in default finding groups sort (`-fail_count,-severity,-last_seen_at`)
- **DateWithTime truncation**: Remove `flex-wrap` from inline mode, add `truncate` + tooltip to prevent overflow in compact layouts
- **Drawer label fix**: Rename "Impacted Resource X of Y" to "Resource X of Y" since both PASS and FAIL resources are now shown

### Steps to review

1. **Drill-down with PASS groups**: Navigate to `/findings`, find a group with 0 impacted / N total resources, click to expand — PASS resources should now appear
2. **Mute selection**: Verify PASS-only groups have disabled checkboxes, PASS resources inside drill-down are not selectable for mute
3. **Optimistic mute**: In the resource detail drawer "Other findings" section, mute a finding — row should immediately show "Muted" without full refresh
4. **Scan selector**: Click "View Scan" from a resource drawer → `/scans?filter[id__in]=...` should show the `ScanSelector` dropdown, not a raw ID chip
5. **Filter badges**: On `/findings`, apply provider/scan filters and verify badges show human-readable aliases
6. **DateWithTime**: Check inline dates in the scan selector and resource drawer — should truncate with tooltip on hover

### Checklist

#### UI

### DIFF
--- a/ui/actions/finding-groups/finding-groups.adapter.test.ts
+++ b/ui/actions/finding-groups/finding-groups.adapter.test.ts
@@ -163,6 +163,7 @@ describe("adaptFindingGroupResourcesResponse — malformed input", () => {
               alias: "production",
             },
             status: "FAIL",
+            delta: "new",
             severity: "critical",
             first_seen_at: null,
             last_seen_at: "2024-01-01T00:00:00Z",
@@ -178,5 +179,6 @@ describe("adaptFindingGroupResourcesResponse — malformed input", () => {
     expect(result).toHaveLength(1);
     expect(result[0].checkId).toBe("s3_check");
     expect(result[0].resourceName).toBe("my-bucket");
+    expect(result[0].delta).toBe("new");
   });
 });
--- a/ui/actions/finding-groups/finding-groups.adapter.ts
+++ b/ui/actions/finding-groups/finding-groups.adapter.ts
@@ -98,6 +98,7 @@ interface FindingGroupResourceAttributes {
   resource: ResourceInfo;
   provider: ProviderInfo;
   status: string;
+  delta?: string | null;
   severity: string;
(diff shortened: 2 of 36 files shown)


pfe-nazaries (dismissed):
LGTM

PR #10627, opened 2026-04-09
chore(skills): improve prowler-compliance

pedrooot:
### Description

This pull request introduces a suite of new Python utilities to support and streamline the mapping, validation, and auditing of Prowler compliance checks against cloud security frameworks. The tools enable building and querying a check inventory, reviewing framework mappings, and encoding audit decisions with pre-validation. These scripts are intended to make the compliance mapping process more robust, transparent, and auditable.

The most important changes are:

**New utilities for compliance mapping and auditing:**

* Added `audit_framework_template.py`, a template script for encoding and validating explicit mappings between compliance framework requirements and Prowler checks, including strong pre-validation and clear workflow documentation.
* Added `build_inventory.py`, a script to scan Prowler check metadata and build a per-provider inventory of available checks, outputting JSON files used by other utilities for validation and lookup.

**Support tools for review and lookup:**

* Added `dump_section.py`, a utility to dump all requirements and their check mappings for a given section/prefix across providers, aiding in review and bulk mapping.
* Added `query_checks.py`, a command-line tool for keyword, service, or ID-based lookup of Prowler checks within the generated inventory, supporting both full-text and metadata queries.

### Steps to review

### Checklist

<details>

</details>

#### SDK/CLI

#### UI

#### API

### License


skills/prowler-compliance/assets/build_inventory.py
```diff
@@ -0,0 +1,82 @@
...
+    /tmp/checks_azure.json (~167 checks)
+    /tmp/checks_gcp.json   (~102 checks)
+"""
+from __future__ import annotations
+
+import json
+import sys
+from pathlib import Path
+
+DEFAULT_PROVIDERS = ["aws", "azure", "gcp", "kubernetes", "m365", "github",
```
HugoPBrito: Should `DEFAULT_PROVIDERS` include every provider that currently has `prowler/providers/<provider>/services` in the repo?

Right now this list seems to skip at least `openstack` and `vercel`, so running the script without arguments would not build a complete inventory for all providers present in the repo. Is that omission intentional, or should the default be derived from the filesystem to avoid stale lists over time?

skills/prowler-compliance/assets/sync_framework.py
```diff
@@ -0,0 +1,478 @@
...
+        if name not in attrs:
+            return None
+        value = attrs[name]
+        if value in ("", None, [], {}):
+            return None
+        parts.append(_freeze(value))
+    return tuple(parts)
+
+
+def load_legacy_check_maps(
```
HugoPBrito, line 201: Could this preservation strategy silently mix mappings for different CCC requirements?

`load_legacy_check_maps()` indexes legacy checks first by `Id` and then by fallback keys from `Attributes[0]` (for CCC, `(Section, Applicability)`), and `build_provider_json()` reuses the first match it finds. When I checked the current `master` CCC catalogs, those keys did not look unique:
- duplicate `Id`s already exist for some CCC requirements
- `(Section, Applicability)` also collides multiple times across `aws`, `azure`, and `gcp`

If that is the case, wouldn’t a sync end up merging `Checks` from different requirements instead of preserving them faithfully? Should this fail fast when the preservation key is ambiguous, or use a stronger key for CCC?

skills/prowler-compliance/SKILL.md
```diff
@@ -455,14 +539,453 @@ Prowler ThreatScore is a custom security scoring framework developed by Prowler
...
+```bash
+python assets/query_checks.py aws encryption transit    # keyword AND-search
+python assets/query_checks.py aws --service iam         # all iam checks
+python assets/query_checks.py aws --id kms_cmk_rotation_enabled  # full metadata
+```
+
+**Step 3 — Dump a framework section with current mappings** — see [assets/dump_section.py](assets/dump_section.py):
+
+```bash
+python assets/dump_section.py "CCC.Core."      # all Core ARs across 3 providers
```
HugoPBrito: Should the `dump_section.py` examples include the `framework_key` argument as well?

In the new workflow docs, the examples show:

```bash
python assets/dump_section.py "CCC.Core."
python assets/dump_section.py "CCC.AuditLog."
```

But the script usage in `dump_section.py` expects two positional arguments: `framework_key` and `id_prefix`. Should these examples be something like `python assets/dump_section.py ccc "CCC.Core."` instead?

skills/prowler-compliance/assets/build_inventory.py
pedrooot: Done!

skills/prowler-compliance/assets/sync_framework.py
pedrooot: Fixed, added some logic to distinct the ID with some fallback fields too

skills/prowler-compliance/SKILL.md
pedrooot: Fixed, thanks for the heads up!

skills/prowler-compliance/assets/parsers/finos_ccc.py
```diff
@@ -0,0 +1,219 @@
...
+    seen_ids: set[str] = set()
+
+    for filename in catalog_files:
+        path = upstream_dir / filename
+        if not path.exists():
+            # The runner handles fatal errors; a missing optional catalog
+            # file is surfaced as a warning via print to stderr.
+            import sys
+
+            print(f"warn: missing upstream file {filename}", file=sys.stderr)
```
HugoPBrito: Question on the intended behavior here: should missing configured catalog files be treated as a warning, or as a hard failure?

Right now `parse_upstream()` prints `warn: missing upstream file ...` and continues. I can see the argument for best-effort output, but I am not sure that matches the canonical-sync framing in the docs/config, where `ccc.yaml` declares a closed set of upstream catalog files expected to make up the framework.

My concern is that with the current behavior an incomplete upstream cache still produces valid-looking JSON, just with silently missing sections of the framework.

If this is intentionally a best-effort helper, maybe that is fine. But if this runner is meant to produce a canonical sync from the upstream source of truth, I wonder whether one of these would be safer:

- fail at the end if any configured catalog was missing
- add an explicit partial/incomplete-sync warning in the output
- make optional catalogs explicit in config instead of treating every miss the same way

Mostly asking because the current middle ground can look like a successful sync even when part of the framework was skipped.

skills/prowler-compliance/assets/sync_framework.py
HugoPBrito: I re-checked this after the latest changes and I think the ambiguity issue is still present, just in a different form.

The runner no longer silently merges colliding fallback buckets, but with the current CCC config it now seems to fail hard on valid legacy data instead:

- `skills/prowler-compliance/assets/configs/ccc.yaml` still uses `fallback_keys: [[Section, Applicability]]`
- in the current `ccc_{provider}.json` files, that tuple does not appear to be unique

I verified this against the current repo state and found multiple collisions per provider (22 in `aws`, `azure`, and `gcp`). A few concrete examples:

- `CCC.AuditLog.CN01.AR01` and `CCC.AuditLog.CN01.AR02`
- `CCC.Logging.CN01.AR01` and `CCC.Logging.CN01.AR02`
- `CCC.LB.CN01.AR01` and `CCC.LB.CN01.AR02`

So unless I am missing something, the documented CCC sync flow would still fail in `load_legacy_check_maps()` before writing output.

Should CCC use a stronger fallback key here, or a more framework-specific preservation strategy?

skills/prowler-compliance/assets/parsers/finos_ccc.py
pedrooot: Fixed! Thanks

skills/prowler-compliance/assets/dump_section.py
```diff
@@ -0,0 +1,91 @@
...
+    for ar_id in sorted(by_id):
+        rows = by_id[ar_id]
+        sample = next(iter(rows.values()))
+        print(f"\n### {ar_id}")
+        print(f"  desc: {sample['desc']}")
+        if sample["sec"]:
+            print(f"  sec : {sample['sec']}")
+        if sample["obj"]:
+            print(f"  obj : {sample['obj']}")
+        for prov in ["aws", "azure", "gcp", "kubernetes", "m365", "github",
```
HugoPBrito: The output here looks incomplete because discovery is dynamic but printing is not.

The script finds matching framework files by scanning , but the final print loop uses a hard-coded provider list. That means valid matches can be silently omitted from the output.

For example,  is discoverable, but  is never printed here. Could this iterate over the discovered provider keys instead of a fixed list?

skills/prowler-compliance/assets/audit_framework_template.py
```diff
@@ -0,0 +1,188 @@
...
+        touched += 1
+
+    if unknown:
+        print(f"\n!! {provider} — UNKNOWN CHECK IDS (typos?):", file=sys.stderr)
+        for rid, c in unknown:
+            print(f"   {rid} -> {c}", file=sys.stderr)
+        print("\nAborting: fix the check ids above and re-run.", file=sys.stderr)
+        sys.exit(2)
+
+    with open(path, "w") as f:
```
HugoPBrito: I think this still allows partial writes even though the script says it pre-validates before writing.

apply_for_provider validates and writes one provider at a time. If aws passes but azure contains an unknown check id, the script exits after aws has already been rewritten, which leaves the repo in a partially updated state.

It would be safer to validate all target providers first and only start writing once the full validation pass succeeds.

skills/prowler-compliance/assets/sync_framework.py
```diff
@@ -0,0 +1,522 @@
...
+
+    Returns ``(json_dict, counts)`` where ``counts`` tracks how each
+    requirement's checks were resolved (primary, fallback, or none).
+    """
+    framework = config["framework"]
+    preservation = config["post_processing"]["check_preservation"]
+    primary_key = preservation["primary_key"]
+    fallback_keys = preservation.get("fallback_keys") or []
+
+    legacy_path = resolve_output_path(
```
HugoPBrito, line 357: I do not think the preservation logic is fully generic for versioned frameworks yet.

load_legacy_check_maps reads from the same path template that the runner uses for output. That works for unversioned frameworks like CCC, but for versioned ones such as cis_<version>_<provider>.json, a sync to a new version will look for the new file rather than the previous legacy file whose mappings we want to preserve.

As written, a version bump would fall back to an empty preservation map and lose all legacy Checks. This probably needs either a separate legacy input path or template in config, or a framework-specific migration strategy.

skills/prowler-compliance/assets/dump_section.py
HugoPBrito: The output here looks incomplete because discovery is dynamic but printing is not.

The script finds matching framework files by scanning prowler/compliance/*/*.json, but the final print loop uses a hard-coded provider list. That means valid matches can be silently omitted from the output.

For example, iso27001_2022_nhn.json is discoverable, but nhn is never printed here. Could this iterate over the discovered provider keys instead of a fixed list?

pedrooot: we were adding them in this way because some providers doesn't support compliance frameworks, applying changes to remove the logic

skills/prowler-compliance/assets/audit_framework_template.py
pedrooot: validation added

skills/prowler-compliance/assets/sync_framework.py
pedrooot: fixed in the latest commit.

I went with the "separate legacy input template" option since it keeps the runner generic and doesn't require any framework-specific migration code. Added an optional `post_processing.check_preservation.legacy_path_template` to the config:

- **Unversioned frameworks (CCC, etc.)**: omit the field. The runner falls back to `output.path_template`, so the current behavior is preserved so no config change needed.
- **Versioned frameworks (CIS, etc.)**: point it at the previous version's file so legacy Checks are read from there, e.g.:
  ```yaml
  check_preservation:
    primary_key: Id
    legacy_path_template: "prowler/compliance/{provider}/cis_5.0_{provider}.json"
    fallback_keys:
      - [Section, Profile]
  ```

The template supports the same `{provider}`, `{framework}`, `{version}` placeholders as `output.path_template`, but the previous version can also just be hardcoded as shown above, whichever is clearer per framework.

The runner docstring and `configs/ccc.yaml` now document this explicitly so the next person adding a versioned framework hits the right path on first try.

skills/prowler-compliance/assets/dump_section.py
pedrooot: Fixed too, thanks!

skills/prowler-compliance/assets/sync_framework.py
pedrooot: fixed in the latest commit.

I went with the "separate legacy input template" option since it keeps the runner generic and doesn't require any framework-specific migration code. Added an optional `post_processing.check_preservation.legacy_path_template` to the config:

- **Unversioned frameworks (CCC, etc.)**: omit the field. The runner falls back to `output.path_template`, so the current behavior is preserved so no config change needed.
- **Versioned frameworks (CIS, etc.)**: point it at the previous version's file so legacy Checks are read from there, e.g.:
  ```yaml
  check_preservation:
    primary_key: Id
    legacy_path_template: "prowler/compliance/{provider}/cis_5.0_{provider}.json"
    fallback_keys:
      - [Section, Profile]
  ```

The template supports the same `{provider}`, `{framework}`, `{version}` placeholders as `output.path_template`, but the previous version can also just be hardcoded as shown above, whichever is clearer per framework.

The runner docstring and `configs/ccc.yaml` now document this explicitly so the next person adding a versioned framework hits the right path on first try.

HugoPBrito (approved):

PR #10630, opened 2026-04-09
refactor(api): split finding-groups status from muted state

AdriiiPRodri:
### Context

While reviewing the finding-groups endpoints we identified several inconsistencies in how aggregated status, mute state, and per-status counters are exposed:

1. The aggregated `status` field could return `MUTED`, mixing two orthogonal concepts: the underlying check outcome (FAIL/PASS/MANUAL) and whether all the findings of the group are muted. The status enum should expose only FAIL/PASS/MANUAL and `muted` should live on each row as its own boolean.
2. `MANUAL` findings were silently dropped from the daily summary counts (the aggregator only counted `status=PASS` and `status=FAIL`), so any check producing manual-review findings was invisible in `/finding-groups`.
3. After a mute rule was created, `mute_historical_findings_task` rewrote `Finding.muted` across the entire history with no time limit, but `reaggregate_all_finding_group_summaries_task` only refreshed the latest scan per provider. The result was a permanent drift between `findings.muted=True` and the daily summary rows of older days.
4. Clients using `/finding-groups/{check_id}/resources` had no way to navigate from a resource row back to the actual underlying `Finding`, since the endpoint only returned aggregated resource data.
5. The `new_count`/`changed_count` deltas exposed totals but no breakdown by status or mute state, so a client could see "5 new findings" without knowing how many were FAIL vs PASS vs MANUAL.

This PR rewrites the model and the surrounding aggregation pipeline to fix all five issues at once across `/finding-groups`, `/finding-groups/latest` and `/finding-groups/{check_id}/resources`.

This PR supersedes #10628 (`feat(api): support sort=status on finding-groups endpoints`): the `sort=status` capability is delivered here from day 1 with the corrected priority `FAIL > PASS > MANUAL`, so #10628 should be closed without merging.

#### Model — `FindingGroupDailySummary` gains 17 new columns

Grouped by purpose:

**Status counters (4)**
- `manual_count` — non-muted MANUAL findings, parallel to `pass_count`/`fail_count`.
- `pass_muted_count`, `fail_muted_count`, `manual_muted_count` — muted half of each status, so clients can derive "actionable PASS" as `pass_count - pass_muted_count` and so on. Invariant: `muted_count == pass_muted_count + fail_muted_count + manual_muted_count`.

**Group-level mute flag (1)**
- `muted: BooleanField(default=False)` — `True` only when every finding for that `(provider, check, day)` is muted. Orthogonal to `status`, exposed as a separate field on the API.

**Delta breakdown (12)**

The `new_*`/`changed_*` deltas are now broken down by `(status, muted)` cross-product:
- `new_fail_count`, `new_fail_muted_count`, `new_pass_count`, `new_pass_muted_count`, `new_manual_count`, `new_manual_muted_count`
- `changed_fail_count`, `changed_fail_muted_count`, `changed_pass_count`, `changed_pass_muted_count`, `changed_manual_count`, `changed_manual_muted_count`

The legacy `new_count`/`changed_count` are kept as non-muted totals for back-compat and the existing `sort=delta` lexicographic ordering. Invariant: `new_count == new_fail_count + new_pass_count + new_manual_count` (and the same for `changed_*`).

`pass_count`/`fail_count` are now stored **inclusive** of muted findings, so the aggregated `status` reflects the underlying check outcome even for fully muted groups.

#### Migrations (consolidated to 2 files)

- `0088_finding_group_status_muted_fields.py` — single schema migration with **17 `AddField`** operations and safe defaults (`0` / `False`).
- `0089_backfill_finding_group_status_muted.py` — `RunPython` that dispatches `backfill_finding_group_summaries_task.delay(tenant_id, days=10)` per tenant. Same pattern as `0082`. The actual recompute runs asynchronously in Celery workers; in a fresh test DB the dispatch is a no-op.

#### Aggregator — single writer (`tasks/jobs/scan.py`)

`aggregate_finding_group_summaries` is the only writer to `finding_group_daily_summaries`. Rewritten to compute all 17 new fields plus the existing ones in a single GROUP BY pass: now ~23 filtered `Count(...)` annotations + a `BoolAnd("muted")`. The constructor and `update_fields` list cover every column, so each upsert refreshes the row entirely — no partial updates.

#### Views — both aggregation paths (`api/v1/views.py`)

- `_aggregate_daily_summaries` (summary path): adds `Sum(...)` for the 17 new columns and `BoolAnd("muted")`. Cross-provider aggregate reports `muted=True` only when every contributing daily row is itself fully muted.
- `_aggregate_findings` (finding-level path, used when finding-level filters force live re-aggregation): re-counts inclusively from raw findings, derives `muted` via a chained `Case` annotation against a helper `nonmuted_count`, and computes the same 12 delta breakdown counters and 3 status-muted siblings.
- `_post_process_aggregation`: cascade is now `FAIL > PASS > MANUAL`; `muted` is already on the row from the SQL aggregation.
- `_apply_aggregated_computed_filters`: the `aggregated_status` `Case` expression switches from `MUTED` default to `MANUAL` default. The default fully-muted exclusion is now `exclude(muted=True)` instead of the legacy 3-way count check, and is skipped when the caller passes either `filter[include_muted]` or `filter[muted]`.

#### Resources endpoint — `/finding-groups/{check_id}/resources` and `/latest/.../resources`

- `_build_resource_aggregation`, `_RESOURCE_SORT_ANNOTATIONS` and `_post_process_resources`: the resource-level `status_order` `Case` considers all findings (drops the `finding__muted=False` predicate) and a new `muted = BoolAnd("finding__muted")` annotation is exposed in the response.
- **New `finding_id` field on every row**, pointing to the most recent matching `Finding` for that resource. Computed as `Max(Cast("finding__id", CharField()))`: PostgreSQL has no built-in `max(uuid)` aggregate, so we cast to text first; on UUIDv7's canonical lowercase form a lexicographic `Max()` still resolves to the latest snapshot in time. Lets clients link from a resource row directly to the underlying finding.

#### Sort

- `_FINDING_GROUP_SORT_MAP` adds `status`, `muted` and `manual_count`.
- `_sorted_paginated_response` annotates `status_order` on demand with the new priority `FAIL=3 / PASS=2 / MANUAL=1`, mirroring the resource path. The 12 delta breakdown counters and 3 status-muted siblings are exposed on the serializer but **not** added as sortable fields on purpose, to keep the API surface focused.

#### Filters (`api/filters.py` + `views.py`)

- `FindingGroupAggregatedComputedFilter.STATUS_CHOICES` becomes `(FAIL, PASS, MANUAL)`. `filter[status]=MUTED` is rejected with 400 (pre-production breaking change, intentional).
- New `muted = BooleanFilter(field_name="muted")` targeting the new column.
- `filter_include_muted` now uses `queryset.exclude(muted=True)`; the legacy fallback against `muted_count` is gone.
- `_split_computed_aggregate_filters`: `muted` routed to computed (group-level) instead of finding-level.

#### Serializers (`api/v1/serializers.py`)

- `FindingGroupSerializer` exposes the 17 new fields plus `status` (now FAIL/PASS/MANUAL).
- `FindingGroupResourceSerializer` exposes `muted: BooleanField` and `finding_id: UUIDField`.

#### `reaggregate_all_finding_group_summaries_task` — fixed to match `mute_historical_findings_task`'s scope (`tasks/tasks.py`)

Until now, after a mute rule was created the chain was:

1. `mute_historical_findings_task` updated **every** `Finding` row whose UID matched the rule, with no time limit.
2. `reaggregate_all_finding_group_summaries_task` re-ran the aggregator only on **the latest scan per provider**.

The mismatch left the daily summaries of every previous day stale: the underlying `findings.muted` was correct, but the per-day `finding_group_daily_summaries` rows still showed pre-mute counters. Querying `filter[inserted_at]=<old date>` returned wrong data.

The task is now rewritten to mirror `mute_historical_findings_task`'s unbounded scope:

```python
completed_scans = list(
    Scan.objects.filter(
        tenant_id=tenant_id,
        state=StateChoices.COMPLETED,
        completed_at__isnull=False,
    )
    .order_by("-completed_at")
    .values("id", "completed_at", "provider_id")
)

# Dedupe in Python by (provider, day): keep the most recent scan per pair.
latest_scans: dict[tuple, str] = {}
for scan in completed_scans:
    key = (scan["provider_id"], scan["completed_at"].date())
    if key not in latest_scans:
        latest_scans[key] = str(scan["id"])

group(
    aggregate_finding_group_summaries_task.si(tenant_id=tenant_id, scan_id=scan_id)
    for scan_id in latest_scans.values()
).apply_async()
```

Same Celery `group(...).apply_async()` parallelism — the wallclock scales with the worker pool, not with the number of `(provider, day)` pairs. After this fix the daily summaries are eventually consistent with `findings.muted` for every day that exists in the database.

#### Tests (`api/tests/test_views.py`, `api/tasks/tests/test_tasks.py`)

- `test_finding_groups_status_muted_all` rewritten as `test_finding_groups_fully_muted_group_reflects_underlying_status`: rds_encryption (2 muted FAIL findings) now reports `status=FAIL`, `muted=True`, `fail_count=2`, `fail_muted_count=2`, `muted_count=2` and asserts the sum invariant `muted_count == pass_muted_count + fail_muted_count + manual_muted_count`.
- `test_check_title_icontains_includes_all_title_variants[finding_level_path]`: routing parameter switched from `filter[muted]=include` (no longer finding-level) to `filter[delta]=new`.
- `test_dispatches_subtasks_for_each_provider` → `test_dispatches_subtasks_for_each_provider_per_day`: validates the new `(provider, day)` deduplication of `reaggregate_all_finding_group_summaries_task`.
- New `test_dedupes_scans_to_latest_per_provider_per_day`: when two scans run on the same day for the same provider, only the latest is dispatched.
- New finding-groups tests, parametrized over `finding-group-list` and `finding-group-latest`:
  - `test_finding_groups_status_choices_no_muted` — invariant `status ⊆ {FAIL, PASS, MANUAL}`.
  - `test_finding_groups_serializer_exposes_muted_and_manual_count` — also asserts the new `*_muted_count` siblings are present.
  - `test_finding_groups_filter_status_muted_is_rejected` — verifies `filter[status]=MUTED` returns 400.
  - `test_finding_groups_filter_muted_true` / `test_finding_groups_filter_muted_false`.
  - `test_finding_groups_sort_by_status` (priority FAIL > PASS > MANUAL).
  - `test_finding_groups_sort_by_muted`.
  - `test_finding_groups_delta_status_breakdown` — exercises the 12 delta counters end-to-end against `s3_bucket_public_access` (1 new FAIL + 1 changed FAIL) and validates the `new_count == sum(new_*)` invariant.
- Resources endpoint tests:
  - `test_finding_groups_resources_serializer_exposes_muted` — verifies `rds_encryption` resources expose `muted=True`, `status=FAIL`.
  - `test_finding_groups_resources_exposes_finding_id` — verifies `finding_id` is present on `/finding-groups/rds_encryption/resources` and matches one of the actual fixture finding IDs.
  - `test_finding_groups_latest_resources_exposes_finding_id` — same check on `/finding-groups/latest/rds_encryption/resources`.

#### Breaking changes (intentional, pre-production)

- `status` field on `/finding-groups`, `/finding-groups/latest` and `/finding-groups/{id}/resources` no longer returns `MUTED`. Clients must read the new `muted: bool` field.
- `filter[status]=MUTED` is rejected with 400.
- `pass_count`/`fail_count` semantics changed from non-muted to inclusive of muted findings. The "actionable" non-muted counts are now derivable as `pass_count - pass_muted_count` (and friends).
- `reaggregate_all_finding_group_summaries_task` semantics changed from "latest scan per provider" to "latest scan per `(provider, day)` across the whole history". External callers (none today) would notice more parallel work being dispatched per invocation.
- New `finding_id` field on the response of `/finding-groups/{id}/resources` and `/latest/.../resources`. Additive, no client breakage.
- The UI relies on the existing `include_muted` toggle, which keeps working unchanged: by default fully-muted groups are hidden, the toggle re-enables them.

### Steps to review

1. Apply the migrations:
   ```bash
   cd api
   poetry run python src/backend/manage.py migrate
   ```
   The backfill is dispatched to Celery for the last 10 days. In a fresh test DB the dispatch is a no-op.

2. Run the focused test suite:
   ```bash
   cd api
   poetry run pytest src/backend/api/tests/test_views.py::TestFindingGroupViewSet -x --tb=short
   poetry run pytest src/backend/tasks/tests/test_tasks.py::TestReaggregateAllFindingGroupSummaries -x --tb=short
   ```

3. Smoke-test the API:
   ```bash
   # status enum no longer contains MUTED
   curl '.../api/v1/finding-groups?filter[inserted_at]=YYYY-MM-DD' \
     | jq '.data[].attributes.status' | sort -u
   # → must be a subset of {"FAIL", "PASS", "MANUAL"}

   # muted, manual_count and per-status muted siblings are exposed
   curl '.../api/v1/finding-groups?filter[inserted_at]=YYYY-MM-DD' \
     | jq '.data[0].attributes | {status, muted, pass_count, pass_muted_count, fail_count, fail_muted_count, manual_count, manual_muted_count, muted_count}'

   # Delta breakdown is exposed (non-muted + muted halves)
   curl '.../api/v1/finding-groups?filter[inserted_at]=YYYY-MM-DD' \
     | jq '.data[0].attributes | {new_count, new_fail_count, new_fail_muted_count, new_pass_count, new_pass_muted_count, new_manual_count, new_manual_muted_count}'

   # filter[muted] works in both directions and bypasses the default exclusion
   curl '.../api/v1/finding-groups?filter[inserted_at]=YYYY-MM-DD&filter[muted]=true'
   curl '.../api/v1/finding-groups?filter[inserted_at]=YYYY-MM-DD&filter[muted]=false'

   # filter[status]=MUTED returns 400
   curl -i '.../api/v1/finding-groups?filter[inserted_at]=YYYY-MM-DD&filter[status]=MUTED'

   # sort=status (FAIL > PASS > MANUAL)
   curl '.../api/v1/finding-groups?filter[inserted_at]=YYYY-MM-DD&sort=-status' \
     | jq '.data[].attributes.status'

   # Resources endpoint exposes muted, status (FAIL/PASS/MANUAL only) and finding_id
   curl '.../api/v1/finding-groups/<check_id>/resources?filter[inserted_at]=YYYY-MM-DD' \
     | jq '.data[0].attributes | {status, muted, finding_id}'
   curl '.../api/v1/finding-groups/latest/<check_id>/resources' \
     | jq '.data[0].attributes | {status, muted, finding_id}'
   ```

4. Confirm `/finding-groups/latest` and `/finding-groups/latest/<check_id>/resources` behave the same way (the same code path is exercised).

5. (Optional, requires a populated tenant) Mute a finding via `POST /api/v1/mute-rules`, then query a daily summary from a previous day with `filter[inserted_at]=<old_date>` and confirm the mute is reflected after the background reaggregation completes. Before this PR the older day would still show pre-mute counters.

### Checklist

<details>

</details>

#### SDK/CLI

#### UI

#### API

### License


Alan-TheGentleman (commented):
Three things to look at. The rest is solid work.

1. **Reaggregation task now dispatches one subtask per (provider, day) across the full scan history.** The old behavior was bounded to one scan per provider (latest only). A tenant with 10 providers and a year of daily scans fires ~3,650 tasks in a single `group().apply_async()`, which can saturate the queue and starve other work. The unbounded scope is needed for correctness (fixing the mute drift on old days), but it should have a safety valve — either a configurable lookback (`completed_at__gte=now()-timedelta(days=N)`) defaulting to something reasonable, or chunked dispatch with rate limiting. `mute_historical_findings_task` having the same pattern is known debt, not a justification.

2. **The 23 Count annotations in `_aggregate_findings` (fallback path) could be expensive on large tenants.** The summary path reads pre-aggregated rows and is fine. But the fallback path (triggered by finding-level filters like `delta`, `region`, `service`) runs 23 conditional aggregations against the raw partitioned findings table. Consider whether the 12 delta-muted breakdown fields are actually needed in the fallback path — defaulting them to 0 with a documented limitation would be simpler and safer. The EXPLAIN ANALYZE checkbox is unchecked in the PR template.

3. **`finding_id = Max(Cast("finding__id", CharField()))` relies on UUIDv7 lexicographic ordering via text cast.** This is correct — UUIDv7 high bits are timestamp and hex sort preserves the ordering. But the result is a string that DRF's UUIDField silently parses back. A one-line comment explaining why this works ("UUIDv7 canonical hex is monotonic, so text Max resolves to the latest finding") prevents someone from "fixing" the Cast later without understanding the invariant.

Two smaller things worth a follow-up:

4. **No database-level constraint enforces the muted-count invariant.** The aggregator keeps `pass_muted_count + fail_muted_count + manual_muted_count == muted_count` consistent, but a `CheckConstraint` on the model would catch bugs at write time.

5. **`muted_count` is now redundant** with the three per-status muted counts. Keeping it for backward compat is fine but it is a denormalization risk — worth documenting as computed/derived so nobody updates it independently.

What's good: migration 0088 is textbook zero-downtime (all AddField with safe defaults, no table rewrite). Backfill via Celery dispatch follows the established convention (0062, 0082). `BoolAnd` for the muted flag is elegant and correct. Test coverage is thorough — invariant checks on muted-count sum, filter[status]=MUTED rejection, sort by status priority, delta breakdown end-to-end, resources endpoint exposing finding_id, and reaggregation dedup logic. The status cascade FAIL > PASS > MANUAL is the right priority order. The PR description is one of the best I have seen in this repo — explains the why behind every change, documents breaking changes explicitly, and includes concrete smoke-test commands. Clean work.

PR #10639, opened 2026-04-09
feat(codebuild): use batched API calls to prevent throttling and false positives

DannyLyubenov:
### Context

When scanning AWS accounts with many CodeBuild projects (~100+), the `codebuild_service` dispatches individual API calls per project through a `ThreadPoolExecutor` with 10 workers. This overwhelms the CodeBuild API rate limits, causing `ThrottlingException` errors on `BatchGetProjects`, `BatchGetBuilds`, and `ListBuildsForProject`.

When `BatchGetProjects` is throttled, `project_visibility` remains None, causing the `codebuild_project_not_publicly_accessible` check to report false positive FAIL findings — marking private projects as publicly accessible.

### Description

Refactored `_batch_get_projects` and `_batch_get_builds` in `codebuild_service.py` to operate per-region instead of per-project, and tightened error handling in _list_builds_for_project:

- `_batch_get_projects`: collects all project names for a region and calls
- `_batch_get_builds`: collects all build IDs for a region and calls batch_get_builds once per 100-ID batch, instead of one call per project
- `_list_builds_for_project`: no batch API exists for this operation, so it continues to issue per-project calls in parallel via the shared thread pool, preserving the previous wall-clock behaviour. Per-project failures are now caught and logged individually so a single error no longer aborts the rest of the region.

### Steps to review

- Review `codebuild_service.py` — the three refactored methods (`_batch_get_projects, _batch_get_builds, _list_builds_for_project`) now accept `regional_client` instead of `project`, and the `__init__` calls no longer pass `self.projects.values()` as an iterator for these methods

- Verify that `_batch_get_projects` and `_batch_get_builds` correctly batch in groups of 100 and map responses back to project objects using `arn` and `build id` respectively

- Check `_list_builds_for_project` handles throttling with a simple retry per project within the sequential loop

- Confirm the test mock responses in `codebuild_service_test.py` include the `arn` and `id` fields needed by the new lookup logic

### Checklist

<details>

</details>

#### SDK/CLI

#### UI

#### API

### License


DannyLyubenov:
Hey @MrCloudSec is it possible to review my PR please? Do you prefer me to create an issue and link it to the PR or is this ok

DannyLyubenov:
Hey @AdriiiPRodri could I please ask you to review my changes or if possible point me to someone else?

prowler/providers/aws/services/codebuild/codebuild_service.py
```diff
@@ -44,42 +46,102 @@ def _list_projects(self, regional_client):
...
+            for project in self.projects.values():
+                if project.region != regional_client.region:
+                    continue
+                try:
+                    build_ids = regional_client.list_builds_for_project(
+                        projectName=project.name
+                    ).get("ids", [])
+                    if len(build_ids) > 0:
+                        project.last_build = Build(id=build_ids[0])
+                except ClientError as error:
```
danibarranqueroo: This inner handler is narrower than the previous behavior here. A non-API exception from one project call (for example a connection/read timeout) would fall through to the outer `except` and stop processing the rest of the projects in this region. Could we broaden this inner handler so per-project failures stay isolated, as they were before?

prowler/providers/aws/services/codebuild/codebuild_service.py
```diff
@@ -44,42 +46,102 @@ def _list_projects(self, regional_client):
...
+                    if error.response["Error"]["Code"] == "ThrottlingException":
+                        time.sleep(1)
```
danibarranqueroo: Prowler already configures botocore standard retries at the provider level, so this fixed `sleep(1)` retry only runs after the SDK has already exhausted its own backoff. If there is a known reason the SDK retries are not sufficient for this API specifically, could you add a short comment explaining that? Otherwise it may be simpler to rely on the shared retry policy alone.

prowler/providers/aws/services/codebuild/codebuild_service.py
```diff
@@ -44,42 +46,102 @@ def _list_projects(self, regional_client):
                 f"{regional_client.region} -- {error.__class__.__name__}[{error.__traceback__.tb_lineno}]: {error}"
             )
 
-    def _list_builds_for_project(self, project):
+    def _list_builds_for_project(self, regional_client):
```
danibarranqueroo, line 60: This threading change makes sense for the real batch APIs, but `ListBuildsForProject` still has no batch variant, so the call count here is still `N`. Previously this could use the shared worker pool across projects; now it is one serial loop per region. Have you measured the wall-clock impact on a large single-region account? If it is noticeable, it may be worth keeping per-project threading just for this call.

tests/providers/aws/services/codebuild/codebuild_service_test.py
danibarranqueroo, line 1: The current test still covers only the one-project happy path, so the main behavior change in this PR is not exercised yet. Could you extend this with a `>=101` projects case to verify chunking, and an out-of-order batch response case to verify the `id`/`arn` mapping back to projects? If the manual retry stays, a stateful retry-path test would also help.

danibarranqueroo (changes requested):
Hello @DannyLyubenov! Thanks for this, I've done a first review, please read my comments and answer them when you can.

Some general comments that I wanted to tell you too:

- The batching of `BatchGetBuilds` and `BatchGetProjects` is a real improvement and worth keeping. One framing note though: the previous implementation was already bounded by the shared worker pool, so the strongest benefit here is reduced total API call volume rather than removal of unbounded concurrency. It may be worth updating the PR description message to reflect that more precisely.

- It's needed to add this changes into the changelog.

prowler/providers/aws/services/codebuild/codebuild_service.py
DannyLyubenov: good call, i widened the exception

prowler/providers/aws/services/codebuild/codebuild_service.py
DannyLyubenov: the gains a very minimal so i will revert to the standard

tests/providers/aws/services/codebuild/codebuild_service_test.py
DannyLyubenov: I've created a new test which checks 100 + 50

DannyLyubenov:
Hey @danibarranqueroo thank you so much for taking the time and reviewing my PR. I tried addressing all your pointers and as far as i know you tweaked the name of my PR title. I hope I havent missed something

danibarranqueroo:
Thanks for this @DannyLyubenov! I'll review everything again as soon as possible 🔝

danibarranqueroo (approved):
Thanks again for this @DannyLyubenov! I've tested using our AWS accounts but we have less than 100 projects so it's working as expected but there is no improvements. Please, feel free to test with your account and open an issue if the problem still exists.

I've also updated the code with flake8 and black so now tests passed and I can merge this.

Thanks again, really good job! 🚀

PR #10653, opened 2026-04-10
chore(ui): bump serialize-javascript override from 7.0.3 to 7.0.5

pfe-nazaries:
## Summary

- Bump `serialize-javascript` pnpm override from 7.0.3 to 7.0.5
- Version 7.0.3 was published manually by the maintainer instead of through the GitHub Actions OIDC trusted publishing pipeline established in 7.0.2, causing pnpm's `trustPolicy: no-downgrade` to block installation with `ERR_PNPM_TRUST_DOWNGRADE` (trust evidence regressed from "trusted publisher" to "no trust evidence")
- Version 7.0.5 restores trusted publishing via GitHub Actions and patches **CVE-2026-34043** (CPU exhaustion DoS via crafted array-like objects)

## Test plan

- [x] `pnpm install` succeeds with no `ERR_PNPM_TRUST_DOWNGRADE` errors
- [x] Lockfile correctly resolves all `serialize-javascript` references to 7.0.5

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #10654, opened 2026-04-10
chore(ui): add knip for dead code detection

pfe-nazaries:
## Summary

- Install [knip](https://knip.dev) and create initial configuration tailored to the Next.js 16 / React 19 project structure
- Add `lint:knip` and `lint:knip:fix` scripts to `package.json`
- Configure entry points for Next.js app conventions, Auth.js, Sentry instrumentation, and build scripts
- Ignore `@heroui/*` sub-package imports (resolved transitively via `@heroui/react`)

## Baseline report

Running `pnpm lint:knip` surfaces ~504 issues across the UI codebase:

| Category | Count |
|----------|-------|
| Unused files | 37 |
| Unused dependencies | 13 |
| Unused devDependencies | 9 |
| Unlisted dependencies | 1 |
| Unresolved imports | 2 |
| Unused exports | 399 |
| Unused exported types | 37 |
| Unused exported enum members | 5 |
| Duplicate exports | 1 |

These findings will be addressed in follow-up PRs (dependency cleanup, dead export removal, config refinement).

## Roadmap

This is the first step in a series of incremental PRs to reduce dead code in the UI:

1. **Base knip setup** ← this PR
2. **Remove unused npm dependencies** — drop packages flagged as unused by knip
3. **Remove unused exports** — clean up barrel re-exports and dead utility functions
4. **Refine config for Next.js 16** — tighten entry points for route groups, metadata exports, instrumentation
5. **Enforce issue ceiling** — lock the clean state with `--max-issues` in healthcheck
6. **CI integration & docs** — add dedicated knip step to CI and contributor guidance in AGENTS.md

## Dependencies

> **Note:** This PR requires #10653

## Test plan

- [x] `pnpm typecheck` passes
- [x] `pnpm lint:check` passes
- [x] `pnpm build` passes
- [x] `pnpm lint:knip` runs and produces the baseline report (exits non-zero as expected)
- [x] No changes to existing code — config and tooling only


Alan-TheGentleman (changes requested):
Three things to address:

1. **`lint:knip` exits non-zero with 504 issues.** Anyone running it will think something is broken. Add a note in the PR description or a comment in `knip.config.ts` explaining the baseline, or consider adding `--max-issues 504` so it only fails on regressions. The roadmap mentions this for step 5, but even a code comment now would save confusion.

2. **Missing `instrumentation.ts` entry point.** Next.js auto-loads this file for Sentry setup. The config already lists `sentry/sentry.server.config.ts` and `sentry/sentry.edge.config.ts` but not the root `instrumentation.ts` that imports them. Knip will likely flag it as unused.

3. **Missing `middleware.ts` entry point.** Same idea. Next.js loads it automatically but knip does not know that unless it is listed explicitly. Add it to the entry array to avoid a false positive.

Alan-TheGentleman (dismissed):
All three points from the previous review are addressed — verified against the latest commit.

1. **Baseline threshold.** `lint:knip` and `lint:knip:fix` now run with `--max-issues 504`, and `knip.config.ts` has a clear header comment explaining the baseline and pointing at step 5 of the roadmap for reducing it. Exactly what I was after.

2. **`instrumentation.ts`.** Not needed in the explicit `entry` array — knip's built-in Next.js plugin already includes `{instrumentation,instrumentation-client,middleware,proxy}.{js,jsx,ts,tsx}` in its default entry patterns, and since this config only overrides root-level `entry` (not `plugins.next.entry`), the plugin defaults still apply. Good call documenting it inline.

3. **`middleware.ts`.** My original comment was outdated — this repo is on Next.js 16, where `middleware.ts` was renamed to `proxy.ts`. The file exists at `ui/proxy.ts` and is covered by the same Next.js plugin default pattern.

LGTM. Thanks for the clean follow-up and the explanatory comments in the config — they make the intent obvious for the next person who opens the file.

Alan-TheGentleman (approved):

Issue #10137, opened 2026-02-23
Azure - filtering scans by Resource Groups


vwkE67JR6M:
### Feature search

- [x] I have searched the existing issues and this feature has not been requested yet or is already in our [Public Roadmap](https://roadmap.prowler.com/roadmap)

### Which component would this feature affect?

Prowler CLI/SDK

### Related to specific cloud provider?

Azure

### New feature motivation

At first thank you for the wonderful tool! I'd like to ask if it's possible to implement a simple switch for Azure scans that narrows down scans to a certain Resource Group. I often audit customers' Azure environments, but most often they request only one (or two) RGs to be validated, corresponding to the environment of some application - not entire Azure tenant/sub.

### Solution Proposed

Adding a CLI arg for Azure scans like `--resource-group group1 group2` that narrow searches to given RGs. 

### Use case and benefits

1. Scans are faster, especially for large subscriptions.
2. Scans consume less memory and disk, especially with Cloud Shell for large subscriptions.
3. Scans are compliant with customer's scope ("Please audit only RG 'foo', 'bar' is out of scope").

### Describe alternatives you've considered

For now I'm parsing JSON output with jq:

```
jq '.[] | select(.resources[].data.metadata.id|test("some-resource-group")?)' prowler-output-xxx.ocsf.json
```

Then load into my own simple HTML viewer. It does not, however, solve the problems described above in "use cases and benefits".

jfagoagas:
Hi @vwkE67JR6M this would be a great addition for Prowler's CLI. Currently this is not in our roadmap but we are looking for new contributors hence contributions. If you are interested in working on this let us know. We/you can create an implementation plan, validate it and the assign the issue to you. 

Thanks for using Prowler!

Legin-ML:
Hi, I would like to work on this issue if this is not being worked on currently!

jfagoagas:
Hi @Legin-ML, that's great! Please, share with us a high level plan if you can!

Legin-ML:
@jfagoagas Here's my initial analysis. Please let me know if there is anything I've missed

- My inital thought is to replicate how subscription_ids are handled. Adding resource groups as an argument, passing it to the `azure_provider.py`.
- Now that the RGs can be accessed via the services's own .py file, the next step is to replace the `.list()` (or `.list_all()` ) function with `.list_by_resource_group()` . 

Based on my initial research, It seems *most* of the services support listing by resource group, but I'll give a concrete answer once I'm sure.

[...]

PR #10479, opened 2026-03-25
feat(cli): add --resource-group flag to filter checks by resource group

HugoPBrito:
### Context

The `ResourceGroup` metadata field is already populated across all 1,160+ checks in all 13 providers, and is available via the API and UI. However, there was no CLI flag to filter checks by resource group. This PR adds that capability.

### Description

- Add `--resource-group` CLI argument to filter checks by their `ResourceGroup` metadata field (mutually exclusive with `--check`, `--service`, `--compliance`, `--category`)
- Add `--list-resource-groups` CLI argument to list all available resource groups for a provider
- Add `list_resource_groups()` and `print_resource_groups()` helper functions
- Add resource group filtering logic in `load_checks_to_execute()`
- Update `CHANGELOG.md`

### Steps to review

1. List available resource groups: `prowler aws --list-resource-groups`
2. Filter by a single resource group: `prowler m365 --resource-group security`
3. Filter by multiple resource groups: `prowler aws --resource-group compute storage`
4. Verify mutual exclusivity: `prowler aws --resource-group compute --category logging` should error
5. Verify invalid resource group: `prowler aws --resource-group invalid_group` should exit with error
6. Run tests: `pytest tests/lib/check/check_loader_test.py tests/lib/cli/parser_test.py -v`

### Checklist

- Are there new checks included in this PR? No
- [x] Review if the code is being covered by tests.
- [x] Review if code is being documented following https://github.com/google/styleguide/blob/gh-pages/pyguide.md#38-comments-and-docstrings
- [ ] Review if backport is needed.
- [ ] Review if is needed to change the Readme.md
- [x] Ensure new entries are added to CHANGELOG.md, if applicable.

#### SDK/CLI
- Are there new checks included in this PR? No

### License

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #10657, opened 2026-04-13
feat(azure): filtering scans at resource group level

Legin-ML:
### Context

This PR resolves the issue #10137. This is partially related to the PR #10479.

### Description

Adds `--azure-resource-group/--azure-resource-groups` CLI flag to the Azure provider to narrow scans to specific resource groups. When provided, Prowler validates each resource group against all accessible subscriptions via ResourceManagementClient and filters service calls accordingly. Missing resource groups are logged as warnings.

### Steps to review

1. Scan with a single resource group: `prowler azure --az-cli-auth --azure-resource-group rg-prod`
2. Scan with multiple resource groups: `prowler azure --az-cli-auth --azure-resource-group rg-prod1 rg-prod2`
3. Verify fallback to full scan when flag is omitted: `prowler azure --az-cli-auth`
4. Verify warning on invalid resource group(s): `prowler azure --az-cli-auth --azure-resource-group invalid-rg` should log a warning and skip it
6. Run tests: (22/22 services covered under test currently)

### Checklist

<details>

- [x] This feature/issue is listed in [here](https://github.com/prowler-cloud/prowler/issues?q=sort%3Aupdated-desc+is%3Aissue+is%3Aopen)
- [x] Is it assigned to me.

</details>

#### SDK/CLI
- Are there new checks included in this PR? : No

## Summary by CodeRabbit

* **New Features**
  * Azure scans can now be limited to one or more resource groups using a new scan option.
  * Supported Azure services now honor resource-group scoping when listing resources, improving targeted scans.
* **Documentation**
  * Added guidance for configuring and using resource-group scoped Azure scans.
  * Updated the Azure provider navigation and changelog entry.


prowler/providers/azure/services/network/network_service.py
```diff
@@ -63,8 +85,29 @@ def _get_network_watchers(self):
         network_watchers = {}
         for subscription, client in self.clients.items():
             try:
+                network_watchers_list = []
+
+                if self.resource_groups:
```
HugoPBrito, line 90: This looks like one of the complex edge cases that was already anticipated in the issue discussion: https://github.com/prowler-cloud/prowler/issues/10137#issuecomment-4192541579

With `--azure-resource-group`, `network_watchers` are now filtered by the resource group that contains the watcher resource itself. The problem is that the downstream checks still evaluate subscription-wide coverage (`network_watcher_enabled`, flow log checks), and in Azure the watcher commonly lives in `NetworkWatcherRG`, not in the audited application resource group. That means a scoped scan can now return false `FAIL` results even when the subscription is correctly configured.

I would either skip these subscription-scoped checks when RG filtering is enabled, with a warning, or redesign them to evaluate the target resources instead of the watcher's own resource group.

prowler/providers/azure/services/monitor/monitor_service.py
```diff
@@ -71,8 +76,30 @@ def get_alert_rules(self):
         for subscription, client in self.clients.items():
             try:
                 alert_rules.update({subscription: []})
-                rules = client.activity_log_alerts.list_by_subscription_id()
-                for rule in rules:
+                rules_list = []
+                if self.resource_groups:
```
HugoPBrito, line 80: This seems to hit the same category of issue discussed here: https://github.com/prowler-cloud/prowler/issues/10137#issuecomment-4192541579

`activity_log_alerts` are now limited to the selected resource groups, but the checks that consume them still report compliance at subscription level. If an alert rule lives in a shared operations resource group and protects resources inside the audited RG, this implementation will return a false `FAIL`.

For ambiguous subscription-scoped checks like these, I would prefer `skip + warning` while RG filtering is enabled rather than returning an incorrect result.

prowler/providers/azure/azure_provider.py
```diff
@@ -1071,6 +1088,37 @@ def get_regions(self, subscription_ids: Union[list[str], None] = None) -> set:
...
+        if not resource_groups:
+            return {}
+
+        rg_map = {sub: [] for sub in self._identity.subscriptions}
+        credentials = self.session
+
+        for display_name, subscription_id in self._identity.subscriptions.items():
+            rg_client = ResourceManagementClient(credentials, subscription_id)
+
+            existing_rgs = {rg.name for rg in rg_client.resource_groups.list()}
```
HugoPBrito, line 1101: This validation flow is aligned with the behavior agreed in the issue discussion (`warning` for invalid RGs and no findings when none are valid): https://github.com/prowler-cloud/prowler/issues/10137#issuecomment-4192111460

The problem is that this comparison is case-sensitive, while Azure resource group names are case-insensitive. That means a valid input such as `RG-Prod` can be treated as missing if Azure returns `rg-prod`.

I would normalize both sides before comparing, and update the new docs page as well because it currently states that resource group names are case-sensitive.

HugoPBrito (commented):
Thanks for the implementation. The overall direction matches the issue discussion, but I found three cases where `--azure-resource-group` can still return incorrect results, plus one validation/docs mismatch.

Main issues:
- `network` can produce false `FAIL` results because `Network Watcher` resources are filtered by the watcher's own resource group while the downstream checks still evaluate subscription-wide coverage.
- `monitor` can produce false `FAIL` results for activity log alert checks for the same reason.
- Resource group validation is currently case-sensitive even though Azure resource group names are case-insensitive.

Suggested remediation plan:
1. Classify Azure checks as RG-scoped, subscription/tenant-scoped, or cross-scope.
2. For subscription-scoped or cross-scope checks that become ambiguous under RG filtering, skip them with a warning instead of returning a `FAIL`.
3. Make RG validation case-insensitive and update the docs accordingly.
4. Add targeted tests for `NetworkWatcherRG`, activity log alerts living in a different resource group, and valid RG names provided with different casing.

Legin-ML:
Thanks for the review @HugoPBrito. I'll make the changes requested.

Regarding test files, could you let me know if any changes are needed? If not I'll write test files for all of them.

prowler/providers/azure/services/network/network_service.py
Legin-ML: @HugoPBrito, I'm skipping it at the individual check's level with MANUAL status. I'm open to suggestions if there is a better way.

HugoPBrito:
Thanks @Legin-ML! Regarding tests:

Yes, please add coverage for all 22 services — the current 1/22 is what's making Codecov fail (`patch` at 29.03%). For each service test, the key cases to exercise are:

1. **No RG filter** → service returns the full set of resources (current behavior, regression guard).
2. **Single RG filter** → only resources in that RG are returned.
3. **Multiple RG filter** → union across the provided RGs.
4. **Mixed-case RG** → matches Azure's case-insensitive semantics (once you apply the normalization fix in `azure_provider.py`).

On top of that, please add the targeted cases for the ambiguous subscription-scoped checks:

- `network_watcher_*` checks return `MANUAL` when `NetworkWatcherRG` is outside the filtered RGs.
- `monitor.activity_log_alerts` checks return `MANUAL` when the alert lives in a different RG than the one being scanned.

No need to rewrite the existing tests — just extend them with the parametrized RG-filter cases. If a helper for "build a mocked Azure provider with `azure_resource_groups=[...]`" doesn't exist yet, feel free to add one under `tests/providers/azure/azure_fixtures.py` so the 22 service tests stay DRY.

Ping me once it's ready and out of Draft and I'll re-review. If not, I'll be happy to take care of it myself.

Legin-ML:
I have added a new class for 3 kinds of tests. with one RG, 2 RGs and mixed case RG. I have also set a `.lower()` guard while comparing the RGs @HugoPBrito. 3 checks which consume `network_watcher` and 11 checks which consume `activity_log_alerts` have been set to manual.

prowler/providers/azure/azure_provider.py
```diff
@@ -103,6 +104,7 @@ class AzureProvider(Provider):
     _region_config: AzureRegionConfig
     _locations: dict
     _mutelist: AzureMutelist
+    _resource_groups: list
```
HugoPBrito: **Blocking — type annotation mismatch.**

The annotation says `list`, but `validate_resource_groups` returns `dict[str, list[str]]` (a `subscription_name -> [rg_names]` map) and every consumer in the `*_service.py` files treats it as a dict (`self.resource_groups.get(sub, [])`). Change the annotation to `dict[str, list[str]]` (or `dict`) to avoid misleading future maintainers and to keep type checkers happy.

prowler/providers/azure/azure_provider.py
```diff
@@ -333,6 +344,11 @@ def mutelist(self) -> AzureMutelist:
         """Mutelist object associated with this Azure provider."""
         return self._mutelist
 
+    @property
+    def resource_groups(self) -> list:
+        """Returns the list of resource groups to be scanned."""
+        return self._resource_groups
```
HugoPBrito, line 354: **Blocking — return type mismatch.**

Same issue as `:107`: this property returns a dict, not a list. Update both the signature (`-> dict[str, list[str]]`) and the docstring (e.g. *"Mapping of subscription name to the list of resource groups to scan within it"*).

prowler/providers/azure/azure_provider.py
```diff
@@ -448,6 +464,7 @@ def print_credentials(self):
             f"Azure Tenant Domain: {Fore.YELLOW}{self._identity.tenant_domain}{Style.RESET_ALL} Azure Tenant ID: {Fore.YELLOW}{self._identity.tenant_ids[0]}{Style.RESET_ALL}",
             f"Azure Region: {Fore.YELLOW}{self.region_config.name}{Style.RESET_ALL}",
             f"Azure Subscriptions: {Fore.YELLOW}{printed_subscriptions}{Style.RESET_ALL}",
+            f"Azure Resource Groups: {Fore.YELLOW}{sorted({rg for rgs in self._resource_groups.values() for rg in rgs}) if self._resource_groups else 'ALL'}{Style.RESET_ALL}",
```
HugoPBrito: **Blocking — misleading output when no RG matches.**

When the user passes a RG that doesn't exist in any subscription, `_resource_groups` ends up as `{sub1: [], sub2: []}` — a non-empty dict, so the `'ALL'` branch is skipped and the literal `[]` is printed. That gives the user a false hint that the scan is correctly scoped. Use `any(rgs for rgs in self._resource_groups.values())` as the condition, or print something explicit like `"NONE (no matching resource groups found)"` so it's obvious.

prowler/providers/azure/azure_provider.py
```diff
@@ -1071,6 +1088,37 @@ def get_regions(self, subscription_ids: Union[list[str], None] = None) -> set:
 
         return set(chain.from_iterable(locations.values()))
 
+    def validate_resource_groups(self, resource_groups: list) -> dict:
```
HugoPBrito: **Blocking — three issues here:**

**1. No try/except per subscription.** If a single sub returns 403 (Reader without `Microsoft.Resources/subscriptions/resourcegroups/read`), throttling, or a network blip, the whole provider initialization explodes and the scan aborts for **every** sub. That's exactly the scenario where someone reaches for `--azure-resource-group`: scoped permissions. Wrap the loop body in a try/except per-sub and emit a warning, the same way services do.

**2. Loss of the RG's "real case".** You compare with `rg.lower() in existing_rgs` but you accumulate the `rg` exactly as the user typed it. If Azure has `mygroup` and the user passes `MyGroup`, the SDK later receives `MyGroup` in `resource_group_name=`. Most ARM SDK endpoints accept any case, but some (NRP/Network, MySQL flexible) are case-sensitive. Store the real `name`:
```python
existing_rgs = {rg.name.lower(): rg.name for rg in rg_client.resource_groups.list()}
for rg in resource_groups:
    if rg.lower() in existing_rgs:
        rg_map[display_name].append(existing_rgs[rg.lower()])
```

**3. Input validation.** Empty strings (`--azure-resource-group ""`) make it through and the SDK then rejects them with an opaque error. Filter `[r for r in resource_groups if r and r.strip()]` at the top.

prowler/providers/azure/lib/service/service.py
```diff
@@ -20,6 +20,7 @@ def __init__(
         )
 
         self.subscriptions = provider.identity.subscriptions
+        self.resource_groups = provider.resource_groups
```
HugoPBrito, line 29: **Important — extract the shared helper here.**

Flagging on this line because it's the natural place for it. The block
```python
if self.resource_groups:
    rgs = self.resource_groups.get(sub, [])
    if not rgs: warn
    else: for rg in rgs: try: list += list_by_rg(rg) except: warn
else: list = list_all()
```
is copy-pasted across ~20 services. Suggested:
```python
def list_with_rg_scope(self, sub_name, list_all_fn, list_by_rg_fn):
    if not self.resource_groups:
        return list(list_all_fn())
    rgs = self.resource_groups.get(sub_name, [])
    if not rgs:
        return []
    out = []
    for rg in rgs:
        try:
            out += list(list_by_rg_fn(resource_group_name=rg))
        except Exception as error:
            logger.warning(...)
    return out
```
Removes ~600 lines and locks in consistency (also fixes the inconsistency flagged in `aisearch_service.py:25`). If you'd rather defer this to a follow-up, let's open an issue.

prowler/providers/azure/services/aisearch/aisearch_service.py
```diff
@@ -16,8 +16,29 @@ def _get_aisearch_services(self):
...
         for subscription, client in self.clients.items():
             try:
+                aisearch_services_list = []
                 aisearch_services.update({subscription: {}})
-                aisearch_services_list = client.services.list_by_subscription()
+                if self.resource_groups:
+                    rgs = self.resource_groups.get(subscription, [])
+                    if not rgs:
+                        logger.warning(
+                            f"No valid resource groups for subscription {subscription}"
```
HugoPBrito: **Important — two issues:**

**1. Skip inconsistency.** Here you don't `continue` and you leave an empty dict. In `keyvault_service.py:44` and `mysql_service.py` you *do* `continue`. Pick a single behavior (I'd leave the dict empty without `continue`, so the sub key is preserved).

**2. Duplicate warning.** `validate_resource_groups` already warns for each RG not found and for the "none found at all" case. Re-warning here for each of the ~20 services produces noise (N×services warnings). I'd drop this to `logger.info` or remove it.

prowler/providers/azure/services/monitor/monitor_service.py
```diff
@@ -71,8 +76,9 @@ def get_alert_rules(self):
         for subscription, client in self.clients.items():
             try:
                 alert_rules.update({subscription: []})
-                rules = client.activity_log_alerts.list_by_subscription_id()
-                for rule in rules:
+                rules_list = client.activity_log_alerts.list_by_subscription_id()
```
HugoPBrito: **Important — unnecessary API call when `--azure-resource-group` is active.**

When `--azure-resource-group` is active, every monitor check short-circuits to `MANUAL` before touching `monitor_client.alert_rules`. We're still calling `list_by_subscription_id()` here though, which is subscription-scoped and requires permissions a user scoped to a single RG likely doesn't have — service initialization will fail with 403 before checks even run. Add the guard:
```python
if self.resource_groups:
    return alert_rules  # checks return MANUAL anyway
```

prowler/providers/azure/services/entra/entra_service.py
```diff
@@ -66,6 +66,11 @@ def __init__(self, provider: AzureProvider):
     async def _get_users(self):
         logger.info("Entra - Getting users...")
         users = {}
+        if self.resource_groups:
+            logger.warning(
+                "Entra users are tenant-scoped and cannot be filtered by resource group. Skipping."
```
HugoPBrito: **Blocking — silent skips lose findings.**

Same pattern across the 7 Entra methods (`:155`, `:235`, `:273`, `:301`, `:338`, `:375`), in Defender (`defender_service.py:39`, `:85`, `:119`, `:151`, `:195`) and in IAM (`iam_service.py:29`, `:75`).

**Problem:** returning `{}` means downstream checks simply produce no findings (nothing to iterate over). The user sees no trace in the scan output. For a check like `entra_admin_no_mfa_enabled` or `defender_ensure_iot_hub_defender_is_on`, this is a real risk: someone using `--azure-resource-group` to scope a scan will assume their tenant is clean.

Network and monitor checks already emit a `MANUAL` finding with `status_extended` telling the user to re-run without the flag — homogenize that pattern here. Cleanest approach: mark these checks/services as tenant/subscription-scoped and emit the `MANUAL` from the check itself, not from the service. If you keep the service-level guard, at least make sure each check produces a `MANUAL` finding when `data == {}` and `self.resource_groups` is active.

HugoPBrito (changes requested):
Thanks for the PR — the feature is valuable and the test coverage is solid. Before approving, I'd like the following points addressed. I'm flagging as **blocking** anything that affects correctness or security (a user could silently lose findings) and as **important** anything that affects maintainability/UX.

**Blocking**
1. `validate_resource_groups` can take down provider initialization if a single subscription fails (permissions/throttling). [comment on `azure_provider.py:1091`]
2. Type annotations declare `list` for what is actually `dict[str, list]` — misleads readers and breaks type checkers. [comments on `azure_provider.py:107` and `:348`]
3. `print_credentials` shows `[]` (instead of `ALL` or a clearer message) when no RG matches — the user sees an apparently "empty" scan with no signal. [comment on `azure_provider.py:467`]
4. Entra (all methods), Defender (`pricings`/`auto_provisioning`/`assessments`/`settings`/`security_contacts`) and IAM (`roles`/`role_assignments`) **silently skip** without producing any findings when `--azure-resource-group` is used. Monitor and Network checks do emit a `MANUAL` finding with `status_extended`. We need to homogenize so the user never loses visibility. [comments on `entra_service.py:71`, `defender_service.py:39`, `iam_service.py:29`]

**Important**
5. Heavy duplication of the `if self.resource_groups: ...` block across ~20 services. Should be extracted into a helper in `lib/service/service.py`. [comment on `lib/service/service.py:23`]
6. Inconsistent behavior when `rgs` is empty for a subscription (some `continue`, some don't), and the *"No valid resource groups for subscription X"* warning duplicates the one already emitted in `validate_resource_groups`. [comment on `aisearch_service.py:25`]
7. `monitor_service.get_alert_rules` still calls `list_by_subscription_id()` even though all monitor checks return `MANUAL` — wastes an API call and requires permissions the user may not have. [comment on `monitor_service.py:79`]

Happy to approve once these are sorted. **I'm glad to help with whatever you need** — pair on the shared helper, sketch the `MANUAL`-finding pattern for the tenant/subscription-scoped checks, review intermediate pushes, or take any of these items off your plate. Just ping me.

prowler/providers/azure/lib/service/service.py
Legin-ML: This has been added.

prowler/providers/azure/azure_provider.py
Legin-ML: Updated!

prowler/providers/azure/azure_provider.py
Legin-ML: wrapped in try-except, trimmed the input and preserve the case

Legin-ML:
Really thankful for the in-depth review @HugoPBrito. I have made the changes you requested. Please, take your time in reviewing it. I really appreciate the support here 😊

HugoPBrito (changes requested):
Thanks for working through everything in the previous round, @Legin-ML. The shared helper, the case-insensitive RG resolution with the real Azure name preserved, the MANUAL findings in Entra / Defender / IAM, the per-subscription `try/except`, the input trim, the `print_credentials` fallback, and the type annotations all look good. Reviewed at `e2edd481`.

I found two new issues introduced in `feat: add manual to checks` that need addressing before approval, plus a couple of minor follow-ups.

**Blocking**

1. **All 11 monitor `activity_log_alerts` checks now produce zero findings when `--azure-resource-group` is active.**
   `monitor_service.get_alert_rules` early-returns `{}` (`monitor_service.py:76-77`), and every monitor alert check iterates `monitor_client.alert_rules.items()` — e.g. `monitor_alert_create_update_nsg.py:13`. With an empty dict, the `for` loop never executes, so the MANUAL branch added inside it is unreachable. This re-introduces the silent-skip problem the previous round fixed in Entra / Defender / IAM.
   Two ways to fix it:
   - Keep the API-call skip but pre-populate the dict, e.g. `alert_rules = {sub: [] for sub in self.clients}` before returning early; **or**
   - Have the checks iterate `monitor_client.subscriptions` instead of `monitor_client.alert_rules`, mirroring the Defender / IAM pattern.
   I'd lean toward the second option since it's the same shape used elsewhere and keeps the service free of "fake" entries.

2. **`tests/providers/azure/azure_provider_test.py::test_validate_resource_groups_mixed_case` is failing.**
   ```
   AssertionError: assert ['MyGroup'] == ['mygroup']
   ```
   The implementation now correctly returns Azure's real casing (`existing_rgs.get(rg.lower())`), which is what we agreed on, but the test still asserts the user-typed casing. Update the assertion to expect `["MyGroup"]`. While you're there, also assert the `existing_rgs.list()` mock was called only once per subscription (regression guard).

**Important**

3. `validate_resource_groups` is still annotated `-> dict:` at `azure_provider.py:1091`. The property is now `dict[str, list[str]]`; the function should match for consistency and to lock in the contract.

4. The new docs page (`docs/user-guide/providers/azure/resource-groups.mdx`) no longer claims case-sensitivity (good), but it also doesn't tell users it's case-insensitive. One sentence under "How It Works" would close the loop, e.g. *"Resource group names are matched case-insensitively, mirroring Azure's own behavior."*

Once the monitor checks emit MANUAL findings under RG filtering and the test is green I'm happy to approve. Ping me when ready.

Legin-ML:
Hi @HugoPBrito,

1. I have modified the monitor checks and tested that they produce a MANUAL output instead of exiting silently.
2. Fixed the mixed case test case (It seems I only ran the tests under tests/providers/azure/services before 😅 )
3. Modified the annotation and docs.

HugoPBrito:
Hi @Legin-ML!

Thank you very much for your hard work. At first glance, I do not see any blocking points anymore. Nevertheless, I still need to do some testing before approving.

In the meantime, please make sure all CI/CD checks pass. Please ignore the SDK Duplicate one for the moment.

HugoPBrito:
A big PR was just merged. @Legin-ML, please let me know if you need help resolving the conflicts.

Legin-ML:
@HugoPBrito I have merged the master branch and have run tests with 100% success, Could you run the CI/CD to enable me to do any remaining fixes required?

HugoPBrito:
Regarding `Duplicate Test Names`, please rename Azure `iam_service_test.py` to `azure_iam_service_test.py`.

Legin-ML:
Done @HugoPBrito!

prowler/providers/azure/services/monitor/monitor_service.py
```diff
@@ -21,6 +21,11 @@ def _get_diagnostics_settings(self):
         diagnostics_settings = {}
         for subscription, client in self.clients.items():
             try:
+                if self.resource_groups:
```
HugoPBrito: When `--azure-resource-group` is active, this `continue` runs before the `diagnostics_settings.update(...)` further down, so `self.diagnostics_settings` ends up as an empty `{}`. The four checks that iterate `monitor_client.diagnostics_settings.items()` then never enter the loop and produce zero findings silently:

- `monitor_diagnostic_settings_exists`
- `monitor_diagnostic_setting_with_appropriate_categories`
- `monitor_storage_account_with_activity_logs_is_private`
- `monitor_storage_account_with_activity_logs_cmk_encrypted`

Diagnostic settings are subscription-scoped, so please apply the same pattern you used for the alert checks: have these four checks iterate `monitor_client.subscriptions` and emit a MANUAL finding when `resource_groups` is set, so the user never loses visibility.

prowler/providers/azure/services/network/network_watcher_enabled/network_watcher_enabled.py
```diff
@@ -9,6 +9,21 @@ def execute(self) -> list[Check_Report_Azure]:
             subscription_name = network_client.subscriptions.get(
                 subscription, subscription
             )
+            if network_client.resource_groups:
```
HugoPBrito: The monitor alert checks iterate `subscriptions` directly, which guarantees one MANUAL finding per subscription. Here the loop above iterates `network_watchers.items()` instead. If RG filtering leaves a subscription without an entry in that dict, the MANUAL finding is not emitted for it. Consider iterating `network_client.subscriptions` for consistency and to keep the MANUAL guarantee uniform across services.

docs/user-guide/providers/azure/resource-groups.mdx
```diff
@@ -0,0 +1,47 @@
...
+```
+
+## How It Works
+
+When `--azure-resource-group` is provided, Prowler validates each specified resource group against all accessible subscriptions. A resource group is included in the scan if it exists in **at least one** subscription.
+
+- If a resource group is found in one or more subscriptions, it will be scanned in those subscriptions only.
+- If a resource group is **not found in any** subscription, Prowler logs a warning and skips it.
+- If **none** of the provided resource groups are found across any subscription, Prowler logs a warning and no resource group scoped checks will run.
+- Resource group names are matched case-insensitively — `MyGroup` and `mygroup` are treated as the same group, mirroring Azure's own behavior.
```
HugoPBrito: Small style nit: this line uses an em dash. Please rephrase without it to match the docs style, for example: "Resource group names are matched case-insensitively, so `MyGroup` and `mygroup` are treated as the same group, mirroring Azure's own behavior."

HugoPBrito (changes requested):
Thanks for the great work here, @Legin-ML! There are still a couple of things pending, please check the inline comments.

Legin-ML:
Sorry for the delay @HugoPBrito, I have made the changes to monitor, network and docs

HugoPBrito:
Hi @Legin-ML,

First of all, an apology. I am reversing part of something we agreed earlier, and that lands on you after you already implemented it. That is on me. Let me explain the reasoning so this is not just churn.

Earlier I asked you to skip the checks that are not resource group scoped, and during review that turned into the MANUAL findings. After thinking it through, I believe that was the wrong call:

1. **A MANUAL is not necessary, because the verdict would not be a lie.** These controls are subscription wide or tenant wide by nature. Queried globally (`list_by_subscription_id`, `list_all`, tenant scope), the verdict is accurate: "No NSG activity log alert exists in the subscription" is a true FAIL, not a false one. The false FAIL risk only appears if we filter the data by resource group, which we do not need to do here.
2. **A silent loss of coverage is worse than an out of scope finding.** For a security tool, a false negative that quietly disappears is more dangerous than a finding the user can review and mute. MANUAL leaves the control unevaluated, while a real PASS or FAIL keeps it actionable.

So the proposal is: every check that currently returns MANUAL under the flag should instead evaluate globally and emit the same real PASS or FAIL the unfiltered scan returns today. The resource group filtering you built for the resource level checks stays exactly as it is. The one conscious tradeoff is that a resource group scoped scan will surface subscription level findings, which is intentional and true, and anything a user considers out of scope can always be muted through the mutelist.

This is a small and well contained change, so you can absolutely handle it yourself without any problem. If at any point you would rather I take it off your plate, just say the word and I will push it on top of your branch. Whatever suits you best.

Thanks for your patience with the back and forth.

Legin-ML:
No issues @HugoPBrito, I'll take this up.

Legin-ML:
@HugoPBrito I have reverted the checks with MANUAL findings to have the current (I am taking master as the source of truth) code. Defender's JIT and IOT have the resource group filtering enabled, as do some others, If the service as a whole needs to be reverted, let me know about that as well.

prowler/providers/azure/azure_provider.py
```diff
@@ -115,6 +117,7 @@ def __init__(
         tenant_id: str = None,
         region: str = "AzureCloud",
         subscription_ids: list = [],
+        resource_groups: list = [],
```
HugoPBrito: This changes the existing SDK constructor contract.

`resource_groups` was added before `config_path`, `config_content`, and `fixer_config`. That means existing programmatic callers using the old positional argument order can now pass a `config_path` value that gets interpreted as `resource_groups`.

The CLI path passes this by keyword, so it avoids the issue, but the SDK constructor is still a public API surface. Please move `resource_groups` after the existing parameters, or make it keyword-only, and add a regression test for the old positional `config_path` call.

prowler/providers/azure/azure_provider.py
```diff
@@ -1082,6 +1099,49 @@ def get_regions(self, subscription_ids: Union[list[str], None] = None) -> set:
 
         return set(chain.from_iterable(locations.values()))
 
+    def validate_resource_groups(self, resource_groups: list) -> dict[str, list[str]]:
+        resource_groups = [r for r in resource_groups if r and r.strip()]
```
HugoPBrito: This still leaves valid resource group input unnormalized.

Empty values are filtered out, but values with surrounding whitespace are kept as-is. For example, `--azure-resource-group " rg-prod "` will not match `rg-prod`, even though the user provided the right resource group name.

Please strip the input values before matching, while still preserving Azure's real resource group casing in the returned map.

docs/user-guide/providers/azure/resource-groups.mdx
```diff
@@ -0,0 +1,47 @@
...
+
+When `--azure-resource-group` is provided, Prowler validates each specified resource group against all accessible subscriptions. A resource group is included in the scan if it exists in **at least one** subscription.
+
+- If a resource group is found in one or more subscriptions, it will be scanned in those subscriptions only.
+- If a resource group is **not found in any** subscription, Prowler logs a warning and skips it.
+- If **none** of the provided resource groups are found across any subscription, Prowler logs a warning and no resource group scoped checks will run.
+- Resource group names are matched case-insensitively, so `MyGroup` and `mygroup` are treated as the same group, mirroring Azure's own behavior.
+
+<Warning>
+Even if no resource groups are found, passing the --azure-resource-group flag disables subscription-only scoped scans.
```
HugoPBrito: This warning is misleading with the current implementation.

Resource-level services are scoped to the matching resource groups, but tenant/subscription-wide checks are not simply disabled globally. Some still run at their natural scope, and ambiguous cases need explicit handling so users do not lose visibility.

Please rephrase this to explain that split behavior instead of saying the flag disables subscription-only scoped scans.

HugoPBrito (changes requested):
Thanks for the continued work here. The resource group filtering flow is much closer now, but I left a few inline comments that need to be addressed before this is ready.

The constructor signature issue is the one that keeps this as CHANGES_REQUESTED for me, since it can break existing SDK callers. Please also refresh the PR body/checklist and handle the remaining PR hygiene items before re-requesting review.

Legin-ML:
Hi @HugoPBrito, sorry for the long delay. I have updated RG to be in the last position, updated whitespace handling, fixed docs and added entry to changelog.md. I have updated the PR checklist as well, with what I am confident about, A workflow run is required to see if any changes are needed.

prowler/providers/azure/services/policy/policy_service.py
```diff
@@ -18,8 +18,12 @@ def _get_policy_assigments(self):
 
         for subscription_id, client in self.clients.items():
             try:
```
HugoPBrito, line 20: This should keep listing policy assignments at subscription scope.

The latest agreed behavior is that tenant/subscription-level checks keep running in their natural scope when `--azure-resource-group` is used. `policy_ensure_asc_enforcement_enabled` depends on the subscription-level `SecurityCenterBuiltIn` assignment, so routing this through `list_with_rg_scope()` can make the check disappear under RG filtering.

Please use `client.policy_assignments.list()` here instead of `list_with_rg_scope()`.

prowler/providers/azure/azure_provider.py
```diff
@@ -1101,6 +1118,49 @@ def get_regions(self, subscription_ids: Union[list[str], None] = None) -> set:
...
+            return {}
+
+        rg_map = {
+            subscription_id: [] for subscription_id in self._identity.subscriptions
+        }
+        credentials = self.session
+
+        for subscription_id, display_name in self._identity.subscriptions.items():
+            try:
+                rg_client = ResourceManagementClient(credentials, subscription_id)
```
HugoPBrito: Please pass the Azure region configuration here too.

The rest of the Azure SDK clients use `base_url` and `credential_scopes`, but this `ResourceManagementClient` uses the default public cloud endpoint. That can make resource group validation fail for `AzureUSGovernment` or `AzureChinaCloud` even when the RG exists.

HugoPBrito (changes requested):
No worries about the delay. Quite the opposite, thank you so much for continuing to work on this despite all the review rounds.

I know this PR has gone through several reviews already, but there are still a couple of things we need to fix before approval. Also, remember that at any point you can delegate the work to me while I keep asking for the remaining changes.

Please check the inline comments I left.

Legin-ML:
Hi @HugoPBrito, I have made both of your requested changes, as well as merged main. One file had a conflict, the recovery_services test file, so I made sure to keep both my tests and the incoming tests.

HugoPBrito (changes requested):
Hi @Legin-ML,

Thanks again for the continued work here. Most of the previous review comments are addressed now, but I found two remaining issues before approval.

Please check the inline comments.

prowler/providers/azure/services/network/network_service.py
HugoPBrito, line 1: I think Virtual Networks should still be scoped by resource group here.

This is different from the Network Watcher path we discussed earlier. Network Watcher is tied to subscription/global checks, so keeping it evaluated globally matches the decision to avoid losing coverage for subscription-level controls.

Virtual Networks are concrete resource-group resources, and the Azure SDK supports both scopes: `virtual_networks.list(resource_group_name)` lists VNets in one resource group, while `virtual_networks.list_all()` lists VNets across the subscription. The checks consuming this collection evaluate specific VNets and subnets, so an RG-scoped scan can currently report findings for VNets outside the selected resource groups.

Please use the RG-scoped helper here when `self.resource_groups` is set, and add the corresponding tests.

tests/providers/azure/services/policy/policy_service_test.py
HugoPBrito, line 1: These tests look stale after the final scope decision.

The runtime now uses `client.policy_assignments.list()`, which matches the agreed behavior for subscription-level policy checks: they should keep running in their natural scope under `--azure-resource-group`.

Please update these tests so they assert the subscription-scope behavior instead of expecting `list_for_resource_group()`.

Legin-ML: Updated @HugoPBrito

prowler/providers/azure/services/network/network_service.py
Legin-ML: Updated @HugoPBrito

HugoPBrito:
Hi @Legin-ML!

Thanks for your awesome work on this and the effort you've put through many reviews and months. Please fix the conflicts and move the entry of the changelog to the v5.32 section and this will be ready to be merged.

I'll do another functional test in the meantime.

Legin-ML:
Glad to hear that @HugoPBrito, The changes have been made and tested. Thanks to you as well for the patience and testing.

HugoPBrito:
Hi @Legin-ML,

Could you please fix the test suite? This is finally the last step before merging. Once all required checks pass we're good to go.

Legin-ML:
Hi @HugoPBrito, I have ran the tests again and all pass locally, could you retrigger the CI and ensure the same?

HugoPBrito (approved):

HugoPBrito:
Thank you very much @Legin-ML! It's an awesome feature, you should be proud ❤️

Legin-ML:
Glad I had the opportunity to contribute! 😄

PR #10663, opened 2026-04-13
chore(vercel): add disclaimer for checks depending on billing plan

danibarranqueroo:
### Context

Several Vercel checks test features that are only available on Pro/Enterprise plans (WAF, SAML SSO, password protection, production deployment protection, SCIM, skew protection, managed WAF rulesets). When Prowler scans a Hobby account, these checks FAIL without explaining why — users see a failure they cannot remediate without upgrading their Vercel
plan, which looks like a false positive.

### Description

Surfaces the plan requirement directly in the check output for the 11 plan-gated Vercel checks:

- Appends "This feature is only available on Vercel Pro/Enterprise plans." (or Enterprise plans. for Enterprise-only features) to the FAIL status_extended messages.
- Sets Notes in each check's metadata to "Required billing plan: Pro." or "Required billing plan: Enterprise.".
- Updates all affected unit tests to match the new status_extended strings.

Checks touched: project_password_protection_enabled, project_production_deployment_protection_enabled, project_skew_protection_enabled, security_waf_enabled, security_custom_rules_configured, security_ip_blocking_rules_configured, security_rate_limiting_configured, security_managed_rulesets_enabled, team_saml_sso_enabled, team_saml_sso_enforced, team_directory_sync_enabled.

### Checklist

<details>

</details>

#### SDK/CLI

#### UI

#### API

### License


prowler/providers/vercel/lib/billing.py
```diff
@@ -0,0 +1,59 @@
...
+    if not isinstance(data, dict):
+        return None
+    billing = data.get("billing")
+    if not isinstance(billing, dict):
+        return None
+    plan = billing.get("plan")
+    return plan.lower() if isinstance(plan, str) else None
+
+
+def resolve_scope_billing_plan(scope_id: Optional[str] = None) -> Optional[str]:
```
HugoPBrito: `resolve_scope_billing_plan` exists because `VercelProject` does not carry `billing_plan`, whereas `VercelTeam` does. If `VercelProject.billing_plan` is populated during the service fetch (mirroring `team_service._fetch_single_team`), project checks can read it directly and this helper can be removed, eliminating the dependency on `Provider.get_global_provider()`.

prowler/providers/vercel/services/security/security_managed_rulesets_enabled/security_managed_rulesets_enabled.py
```diff
@@ -31,8 +31,9 @@ def execute(self) -> List[CheckReportVercel]:
                 report.status = "MANUAL"
                 report.status_extended = (
                     f"Project {config.project_name} ({config.project_id}) "
-                    f"could not be assessed for managed rulesets. "
-                    f"Enterprise plan required to access this feature."
+                    f"could not be assessed for managed rulesets because the "
```
HugoPBrito, line 35: The new MANUAL message no longer communicates the plan dependency:

- Previous: `"...Enterprise plan required to access this feature."`
- Current:  `"...firewall configuration endpoint was not accessible. Manual verification is required."`

The `Notes` metadata documents the per-ruleset availability, but `status_extended` is what surfaces in findings. Consider appending a conditional suffix via `plan_reason_suffix` so non-Enterprise scopes still receive the rationale.

prowler/providers/vercel/services/project/project_password_protection_enabled/project_password_protection_enabled.py
```diff
@@ -35,9 +39,11 @@ def execute(self) -> List[CheckReportVercel]:
                 )
             else:
                 report.status = "FAIL"
+                billing_plan = resolve_scope_billing_plan(project.team_id)
                 report.status_extended = (
                     f"Project {project.name} does not have password protection "
                     f"configured for deployments."
+                    f"{plan_reason_suffix(billing_plan, {'hobby'}, 'password protection is not available on the Vercel Hobby plan.')}"
```
HugoPBrito: The suffix only triggers for `hobby`, but the metadata `Notes` states *"Enterprise, or as a paid add-on for Pro plans"*. A Pro user without the add-on will see a FAIL without context. This is acceptable given the API does not expose add-on status, but it should be documented in the docs table to prevent future regressions being filed as bugs.

prowler/providers/vercel/services/team/team_service.py
```diff
@@ -67,6 +68,7 @@ def _fetch_single_team(self, team_id: str):
                 id=team_data.get("id", team_id),
                 name=team_data.get("name", ""),
                 slug=team_data.get("slug", ""),
+                billing_plan=extract_billing_plan(team_data),
```
HugoPBrito, line 71: `extract_billing_plan` is wired into `_fetch_single_team` only. If any other code path constructs `VercelTeam` (for example, a bulk listing), those instances will have `billing_plan=None` and the plan-aware suffix will no-op silently. Please confirm all construction sites are covered.

tests/providers/vercel/vercel_fixtures.py
```diff
@@ -46,10 +47,12 @@ def set_mocked_vercel_provider(
         user_id=USER_ID,
         username=USERNAME,
         email=USER_EMAIL,
+        billing_plan=billing_plan,
         team=VercelTeamInfo(
             id=TEAM_ID,
             name=TEAM_NAME,
             slug=TEAM_SLUG,
+            billing_plan=billing_plan,
```
HugoPBrito: Consider exposing a separate `team_billing_plan` parameter. As written, `billing_plan` sets both the identity's user-level plan and the team-level plan to the same value, which prevents tests from exercising the case where the user and team plans differ.

prowler/providers/vercel/lib/service/service.py
```diff
@@ -84,10 +84,10 @@ def _get(self, path: str, params: dict = None) -> dict:
                     )
 
                 if response.status_code == 403:
-                    # Plan limitation or permission error — return None for graceful handling
+                    # Endpoint unavailable for this token/scope; let checks handle it gracefully
                     logger.warning(
                         f"{self.service} - Access denied for {path} (403). "
-                        "This may be a plan limitation."
+                        "This may be caused by plan or permission restrictions."
```
HugoPBrito, line 90: Now that the firewall check no longer treats 403 as implying an Enterprise requirement, consider lowering this log level from `warning` to `info`. A 403 is an expected condition for some token scopes and should not be surfaced as a warning.

HugoPBrito (changes requested):
The conditional plan-aware suffix is the right design: it avoids asserting a plan limitation when the API has not confirmed one. The following items should be addressed before merge.

1. Scope mismatch between description and diff. The description lists 11 checks and specifies the suffix text as `"This feature is only available on Vercel Pro/Enterprise plans."`, but the implementation modifies 8 files (6 with plan-aware logic, 2 message-only) and uses `"This may be expected because ..."`. `security_custom_rules_configured`, `security_ip_blocking_rules_configured`, and `security_rate_limiting_configured` are referenced in the description but not modified. Please either extend the implementation to cover them or update the description to reflect the actual scope.

2. Inconsistent resolution pattern. Project checks resolve the billing plan via `resolve_scope_billing_plan(project.team_id)`, which introduces coupling with `Provider.get_global_provider()`. Team checks read `team.billing_plan` directly. Hydrating `VercelProject.billing_plan` at fetch time would unify both paths and remove the need for the global provider lookup.

3. Loss of actionable signal in `security_managed_rulesets_enabled`. The previous MANUAL message indicated that an Enterprise plan was required. The new message reports only that the endpoint was inaccessible. Since `status_extended` is the user-facing field, consider retaining a plan hint (conditional on the scope's plan) while keeping the neutral wording as the default.

Test coverage is adequate: the `billing_plan` fixture parameter is well-scoped and each modified check has hobby and pro plan cases.

prowler/providers/vercel/lib/billing.py
danibarranqueroo: Agreed.

HugoPBrito (commented):
Follow-up on a few items from the previous review round — non-blocking, but worth considering before merge.

prowler/providers/vercel/services/project/project_service.py
```diff
@@ -96,6 +99,34 @@ def _list_projects(self):
                 f"{error.__class__.__name__}[{error.__traceback__.tb_lineno}]: {error}"
             )
 
+    def _resolve_billing_plan(self, scope_id: Optional[str] = None) -> Optional[str]:
```
HugoPBrito: Wouldn't it be cleaner to unify billing-plan resolution in a single place? `team_service._fetch_single_team` already hydrates `VercelTeam.billing_plan` via `extract_billing_plan(team_data)`, but here we re-derive the plan by walking `provider.identity`. We end up with two code paths reading from two different sources of truth for the same value.

What about moving the logic onto `VercelIdentityInfo` (e.g. `get_billing_plan_for(scope_id)`) and having both services call it? That would also remove the defensive `getattr(self.provider, "identity", None)` and leave a single source of truth.

prowler/providers/vercel/services/project/project_service.py
```diff
@@ -96,6 +99,34 @@ def _list_projects(self):
...
+
+            if identity.user_id == scope_id:
+                return identity.billing_plan
+
+            return None
+
+        if identity.team and identity.team.billing_plan:
+            return identity.team.billing_plan
+
+        return identity.billing_plan
```
HugoPBrito: When `scope_id` is `None`, aren't we attributing the first team's plan to the project even though there's no team context? For tokens that span multiple teams this can end up reporting the wrong plan.

Wouldn't it be better to either always require `scope_id` (assert/raise if missing) or, at minimum, document the `None` semantics explicitly? As it stands, it looks easy for a future caller to assume the fallback is safe when it isn't.

prowler/providers/vercel/lib/service/service.py
```diff
@@ -84,10 +84,10 @@ def _get(self, path: str, params: dict = None) -> dict:
                     )
 
                 if response.status_code == 403:
-                    # Plan limitation or permission error — return None for graceful handling
-                    logger.warning(
+                    # Endpoint unavailable for this token/scope; let checks handle it gracefully
+                    logger.info(
```
HugoPBrito, line 88: Downgrading every 403 to `info` lumps two very different cases together:
- A Hobby account hitting a Pro-only endpoint (expected — shouldn't warn).
- A token lacking permissions on a Pro/Enterprise account (real misconfiguration — should warn).

Wouldn't it be better to keep `warning` by default and only drop to `info` when `billing_plan in unsupported_plans`? That way operators still get a signal when there's a genuine permissions problem, instead of having to dig through `info` logs to find it.

prowler/providers/vercel/lib/billing.py
```diff
@@ -0,0 +1,27 @@
+from typing import Optional
+
+
+def extract_billing_plan(data: Optional[dict]) -> Optional[str]:
```
HugoPBrito, line 4: The docstring already acknowledges that `data["billing"]["plan"]` is not part of Vercel's public OpenAPI schema. If Vercel renames or relocates the field, the feature degrades silently — the suffix simply stops appearing, with no error.

Wouldn't it be safer to log once (at `warning` or `info`) if `billing_plan` comes back `None` for the primary team during identity setup? That gives operators a signal that the heuristic broke, instead of having to notice that findings stopped including plan context.

danibarranqueroo: Makes sense from an observability standpoint. I’m leaving that out of this PR because it is additive telemetry around an undocumented field rather than a correctness issue, and the current behavior still degrades safely by omitting the plan-aware suffix.

prowler/providers/vercel/lib/service/service.py
danibarranqueroo: Agreed on the distinction. I’m not taking that change in this PR because the generic _get() layer does not have enough context to reliably separate expected plan-gated 403s from genuine permission misconfiguration.

HugoPBrito (approved):

PR #10673, opened 2026-04-13
feat(ui): new resources side drawer with redesigned detail panel

alejandrobailo:
### Context

Fix PROWLER-1296

Redesigns the resource detail experience: client-side drawer, new detail panel layout with tabs, AWS organization hierarchy, unified filter UX, and component cleanup.

### Description

- **Client-side drawer with row click**: Replaced server-side Sheet with a client-side Drawer; `DataTable` now supports `onRowClick` so clicking any row opens the detail panel
- **Detail panel redesign**: Compact 4-column info grid with `EntityInfo` components, region flags, tabs reorganized to Findings / Metadata / Tags / Events (default: Findings)
- **AWS organization hierarchy**: For cloud AWS providers, shows Organization → Account hierarchy using `EntityInfo` with a `CornerDownRight` visual connector (cloud-only, gated by `NEXT_PUBLIC_IS_CLOUD_ENV`)
- **Filter improvements**: Added resource type filter (with correct `type__in` API key), prefixed all filter placeholders with "All" for consistency, hidden "Select All" button when no selections are active
- **Component cleanup**: Extracted reusable `InfoTooltip` from `InfoField`, migrated all consumers from legacy HeroUI `InfoField` to shadcn version, deleted legacy component
- **Type safety**: Typed `FINDING_GROUP_ONLY_FILTERS` with `FilterParam` union to prevent typo errors

### Steps to review

1. Navigate to `/resources`, click any table row → detail drawer should open from the right
2. Check the compact info grid: provider, resource name, service, region (with flag), type, group, partition, dates
3. For an AWS provider with an organization: verify the org → account hierarchy with the corner arrow connector
4. Expand "More Filters" → confirm Region, Service, Type, and Group filters all show "All [name]" when empty
5. Select a filter value → "Select All" button appears; deselect all → button disappears
6. Verify Findings tab shows an info tooltip ("This table also includes muted findings")
7. Check Metadata, Tags, and Events tabs render correctly (including empty states)
8. Deep-link: copy a resource URL (link icon), open in new tab → drawer should reopen automatically
9. Verify non-AWS resources do NOT show the organization hierarchy

### Checklist

#### UI

### License


ui/components/resources/table/resources-table-with-selection.tsx
```diff
@@ -1,28 +1,87 @@
...
+        const provider = {
+          data: providerDict[resource.relationships?.provider?.data?.id],
+        };
+        setSelectedResource({
+          ...resource,
+          relationships: { ...resource.relationships, provider },
+        } as ResourceProps);
+        setDrawerOpen(true);
+      },
+    );
```
pfe-nazaries: This promise doesn't have a `.catch()`. If the resource no longer exists (shared link with a deleted `resourceId`) or there's a network error, it produces a silent unhandled promise rejection and the drawer simply doesn't open without any feedback to the user.

Suggestion:

```typescript
getResourceById(initialResourceId, { include: ["provider"] })
  .then((response) => {
    if (!response?.data) return;
    // ... existing code
  })
  .catch((error) => {
    console.error("Failed to fetch resource:", error);
  });
```

alejandrobailo: Good catch! You're absolutely right — if the resource has been deleted or there's a network hiccup, this would silently fail without any feedback. I've added a `.catch()` to handle that case. Thanks for the review!

Alan-TheGentleman (commented):
Two things on the resource fetch approach:

1. **The `Promise.all` in `page.tsx` is fine.** Two independent server-side calls, parallel execution. Nothing better for this case.

2. **The `useEffect` + `.then()` fetch in `resources-table-with-selection.tsx` can be simplified.** Right now the resource-by-ID fetch moved to the client with a manual effect, a ref guard, and no loading state. This introduces a client-side waterfall on deep-links and skips error/loading UX.

Since the server component already has the `initialResourceId` and runs `Promise.all`, the simplest fix is to resolve the resource server-side and pass it as a prop:

```tsx
// page.tsx
const [metadataInfoData, providersData, resourceByIdData] = await Promise.all([
  (hasDateOrScan ? getMetadataInfo : getLatestMetadataInfo)({ query, filters, sort }),
  getProviders({ pageSize: 50 }),
  initialResourceId
    ? getResourceById(initialResourceId, { include: ["provider"] })
    : Promise.resolve(null),
]);

// process + pass down
<ResourcesTableWithSelection
  data={expandedResources}
  metadata={resourcesData?.meta}
  initialResource={processedResource}
/>
```

Then the client component just opens the drawer with the prop on mount. No effect, no ref, no `.then()` chain, no race conditions. The deep-link resource loads in parallel with metadata and providers instead of waiting for hydration.

pfe-nazaries (approved):

PR #10675, opened 2026-04-13
fix(ui): keep update credentials wizard open

Alan-TheGentleman:
### Context

Updating provider credentials in the providers table wizard was closing the modal after a successful connection test, which prevented users from reaching the final launch step and starting a scan from the same flow.

### Description

- keep update flows inside the wizard after a successful connection test so they advance to the launch step
- stop revalidating `/providers` during wizard connection checks to avoid remounting the row-level modal in Next 16
- add regression coverage for update credentials and check connection wizard actions

### Steps to review

1. Open `/providers`
2. Pick a provider with existing credentials and click `Update Credentials`
3. Submit valid updated credentials
4. In `Check connection`, click `Check connection`
5. Verify the wizard stays open and advances to the final `Launch scan` step
6. Launch a scan from the modal
7. Run `pnpm run healthcheck`
8. Run `pnpm test:run actions/providers/providers.test.ts`
9. Run `pnpm test:run components/providers/wizard/hooks/use-provider-wizard-controller.test.tsx`

### Checklist

https://github.com/user-attachments/assets/c6007467-0cd7-4684-93a7-8a072b1c41d9

<details>

</details>

    - If so, do we need to update permissions for the provider?

#### SDK/CLI

#### UI (if applicable)
- [ ] Ensure new entries are added to ui/CHANGELOG.md

#### API (if applicable)

### License

### DIFF
--- a/ui/actions/providers/providers.ts
+++ b/ui/actions/providers/providers.ts
@@ -293,6 +293,7 @@ export const updateCredentialsProvider = async (
       }),
     });
 
+    // No revalidation — keeps the wizard modal open until checkConnectionProvider revalidates
     return handleApiResponse(response);
   } catch (error) {
     return handleApiError(error);
--- a/ui/actions/providers/providers.ts
+++ b/ui/actions/providers/providers.ts
@@ -293,8 +293,7 @@ export const updateCredentialsProvider = async (
       }),
     });
 
-    // No revalidation — keeps the wizard modal open until checkConnectionProvider revalidates
-    return handleApiResponse(response);
+    return handleApiResponse(response, "/providers");
   } catch (error) {
     return handleApiError(error);
   }
@@ -309,7 +308,7 @@ export const checkConnectionProvider = async (formData: FormData) => {
     const response = await fetch(url.toString(), { method: "POST", headers });
     await wait(2000);
 
-    return handleApiResponse(response);
+    return handleApiResponse(response, "/providers");
   } catch (error) {
     return handleApiError(error);
   }
--- a/ui/actions/providers/providers.ts
+++ b/ui/actions/providers/providers.ts
@@ -309,7 +309,7 @@ export const checkConnectionProvider = async (formData: FormData) => {
     const response = await fetch(url.toString(), { method: "POST", headers });
     await wait(2000);
 
-    return handleApiResponse(response, "/providers");
+    return handleApiResponse(response);
   } catch (error) {
     return handleApiError(error);
   }
--- a/ui/actions/providers/providers.ts
+++ b/ui/actions/providers/providers.ts
@@ -293,7 +293,7 @@ export const updateCredentialsProvider = async (
       }),
     });
 
-    return handleApiResponse(response, "/providers");
+    return handleApiResponse(response);
   } catch (error) {
     return handleApiError(error);
   }
--- a/ui/CHANGELOG.md
+++ b/ui/CHANGELOG.md
@@ -44,6 +44,7 @@ All notable changes to the **Prowler UI** are documented in this file.
 - Clear Filters now resets all filters including muted findings and auto-applies, Clear all in pills only removes pill-visible sub-filters, and the discard icon is now an Undo text button [(#10446)](https://github.com/prowler-cloud/prowler/pull/10446)
 - Send to Jira modal now dynamically fetches and displays available issue types per project instead of hardcoding `"Task"`, fixing failures on non-English Jira instances [(#10534)](https://github.com/prowler-cloud/prowler/pull/10534)
 - Exclude service filter from finding group resources endpoint to prevent empty results when a service filter is active [(#10652)](https://github.com/prowler-cloud/prowler/pull/10652)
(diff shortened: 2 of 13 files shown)


alejandrobailo (changes requested):
Hey!

I spotted one thing that I think introduces a subtle regression though:

## `/providers` is never revalidated in the update flow

After this PR, the full update-credentials chain looks like:

1. `updateCredentialsProvider` → `handleApiResponse(response)` — no revalidation
2. `checkConnectionProvider` → `handleApiResponse(response)` — no revalidation
3. `LaunchStep` → calls `scanOnDemand` / `scheduleDaily` → revalidates `/scans`, **not** `/providers`
4. `handleClose()` → resets wizard state, closes modal — no revalidation

So the `/providers` page data is **never refreshed** after the user updates credentials through the wizard. Any changes that should be reflected in the providers table (connection status, credential type, etc.) will stay stale until a manual page refresh.

The PR description says _"revalidation deferred to the connection test step"_, but `checkConnectionProvider` also had its `"/providers"` removed — so the revalidation isn't deferred, it's gone entirely.

## Suggested fix

The controller fix (removing the early-return in `handleTestSuccess`) is what actually solves the modal-closing bug. I think we can safely **restore** `revalidatePath` in `checkConnectionProvider`:

```ts
// checkConnectionProvider
return handleApiResponse(response, "/providers");
```

Since the modal no longer remounts on revalidation (the early-close is gone), this should keep the wizard open *and* ensure the providers table shows fresh data when the user eventually closes it.

The removal in `updateCredentialsProvider` is fine to keep — revalidating at connection-test time is enough.

Let me know what you think!

ui/actions/providers/providers.ts
```diff
@@ -308,7 +309,7 @@ export const checkConnectionProvider = async (formData: FormData) => {
     const response = await fetch(url.toString(), { method: "POST", headers });
     await wait(2000);
 
-    return handleApiResponse(response, "/providers");
+    return handleApiResponse(response);
```
pfe-nazaries: **Warning — stale data after update-credentials flow**

Neither `updateCredentialsProvider` nor `checkConnectionProvider` revalidate `/providers` anymore. The inline comment on line 296 says *"until checkConnectionProvider revalidates"*, but this function no longer does either.

The wizard `handleClose` and the LAUNCH step also don't trigger any revalidation of `/providers` (LAUNCH only revalidates `/scans`). This means after completing the update-credentials wizard, the `/providers` page will show stale data until the user navigates away and back or does a manual refresh.

Consider adding a `revalidatePath("/providers")` call either:
- Here in `checkConnectionProvider` (after the wizard advances to LAUNCH), or
- In the wizard's `handleClose` callback, or
- At the end of the LAUNCH step

Also the comment on line 296 should be updated since it references behavior that no longer exists.

ui/actions/providers/providers.ts
```diff
@@ -293,7 +293,8 @@ export const updateCredentialsProvider = async (
       }),
     });
 
-    return handleApiResponse(response, "/providers");
+    // No revalidation — keeps the wizard modal open until checkConnectionProvider revalidates
```
pfe-nazaries: **Nit** — This comment is misleading: `checkConnectionProvider` no longer revalidates either (see line 312). Consider updating to something like:
```ts
// No revalidation — the wizard stays open; /providers is revalidated when the wizard completes
```
(once you decide where to place the revalidation)

Alan-TheGentleman:
Fixed the wizard flow end to end.

The root cause was that the provider wizard modal was mounted inside table rows, so every refresh of the providers page could remount the row and close the modal.

I moved the wizard to a stable page-level host and wired row actions and Add Provider to open that shared modal instead.

With that change in place, the providers table now refreshes after:
1. linking the provider
2. authenticating credentials
3. checking the connection

and the modal stays open through each intermediate step until the flow actually finishes.

I also fixed the action label so rows without a secret show Add Credentials instead of Update Credentials.

Tests updated and passing locally.

@alejandrobailo @pfe-nazaries could you please take another look?

ui/actions/providers/providers.ts
Alan-TheGentleman: Fixed this by moving the wizard out of the table rows and restoring `/providers` refreshes in the flow. The table now refreshes after link, authenticate, and check connection, and the modal stays open because it no longer depends on row-level state.

ui/actions/providers/providers.ts
Alan-TheGentleman: Yep, that comment was wrong. I removed the misleading wording and reworked the flow so the refresh behavior is explicit in the actions instead of being implied by that comment.

pfe-nazaries (dismissed):
LGTM

alejandrobailo (approved):

alejandrobailo (changes requested):
The `changelog` must be for the .24

Alan-TheGentleman:
Fixed. Moved the changelog entry out of the released 1.23.0 block into a new `[1.24.0] (Prowler UNRELEASED)` section. Also updated the entry text to reflect the full scope of the change (page-level wizard host).

PR #10683, opened 2026-04-14
feat(googleworkspace): add Gmail service with first batch of checks

lydiavilchez:
### Context

Expanding Google Workspace check coverage with the Gmail service, first of two PRs. This PR adds the service infrastructure and 9 boolean-based checks. A second PR will follow with the 8 consequence-based checks (attachment safety and spoofing/authentication).

### Description

Implements 9 CIS Google Workspace Gmail controls using the Cloud Identity Policy API v1:
 
**User Settings:**
- **3.1.3.1.1** — Ensure users cannot delegate access to their mailbox: PASS if `enableMailDelegation` is false
 
**Links and external images:**
- **3.1.3.4.2.1** — Ensure link identification behind shortened URLs is enabled: PASS if `enableShortenerScanning` is true
- **3.1.3.4.2.2** — Ensure scan linked images for malicious content is enabled: PASS if `enableExternalImageScanning` is true
- **3.1.3.4.2.3** — Ensure warning prompt for clicks on untrusted domain links is enabled: PASS if `enableAggressiveWarningsOnUntrustedLinks` is true
 
**End User Access:**
- **3.1.3.5.1** — Ensure POP and IMAP access is disabled: PASS if both `enablePopAccess` and `enableImapAccess` are false
- **3.1.3.5.2** — Ensure automatic forwarding is disabled: PASS if `enableAutoForwarding` is false
- **3.1.3.5.3** — Ensure per-user outbound gateways is disabled: PASS if `allowUsersToUseExternalSmtpServers` is false
 
**Spam, Phishing and Malware:**
- **3.1.3.6.1** — Ensure enhanced pre-delivery message scanning is enabled: PASS if `enableImprovedSuspiciousContentDetection` is true
 
**Compliance:**
- **3.1.3.7.1** — Ensure comprehensive mail storage is enabled: PASS if `ruleId` is present

5 controls are not automatable (offline Gmail, quarantine notifications, external recipient warnings, spam bypass for internal senders, TLS compliance)

Changes:
- New Gmail service (`gmail_service.py`) fetching policies from 10 Policy API setting types
- Gmail client (`gmail_client.py`)
- 9 new checks with metadata and unit tests
- `_is_customer_level_policy()` filter applied from the start
- `policies_fetched` guard on all checks

### Checklist

<details>

</details>

#### SDK/CLI

#### UI

#### API

### License

### DIFF
--- a/prowler/providers/googleworkspace/lib/service/service.py
+++ b/prowler/providers/googleworkspace/lib/service/service.py
@@ -44,7 +44,8 @@ class GoogleWorkspaceService:
     def _is_customer_level_policy(self, policy: dict) -> bool:
         """Check if a policy applies at the customer (domain-wide) level.
 
-        The Cloud Identity Policy API scopes all policies to an OU.
+        The Cloud Identity Policy API typically scopes all policies to an OU;
+        absence of orgUnit is treated as customer-level as a safety net.
         The root OU is equivalent to customer-level. This method accepts
         policies with no orgUnit or policies targeting the root OU,
         and rejects group-targeted and sub-OU policies.
--- a/prowler/providers/googleworkspace/lib/service/service.py
+++ b/prowler/providers/googleworkspace/lib/service/service.py
@@ -41,21 +41,25 @@ class GoogleWorkspaceService:
             )
             return None
 
-    @staticmethod
-    def _is_customer_level_policy(policy: dict) -> bool:
+    def _is_customer_level_policy(self, policy: dict) -> bool:
         """Check if a policy applies at the customer (domain-wide) level.
 
-        The Cloud Identity Policy API returns policies at multiple
-        organizational levels (customer, OU, group). Customer-level
-        policies have no group targeting and no sub-OU targeting in
-        their policyQuery.
+        The Cloud Identity Policy API scopes all policies to an OU.
+        The root OU is equivalent to customer-level. This method accepts
+        policies with no orgUnit or policies targeting the root OU,
+        and rejects group-targeted and sub-OU policies.
         """
         policy_query = policy.get("policyQuery", {})
         if policy_query.get("group"):
             return False
-        if policy_query.get("orgUnit"):
-            return False
-        return True
+        org_unit = policy_query.get("orgUnit")
+        if not org_unit:
+            return True
+        # Accept root OU as customer-level
+        root_id = getattr(self.provider.identity, "root_org_unit_id", None)
+        if root_id and org_unit == f"orgUnits/{root_id}":
+            return True
+        return False
 
     def _handle_api_error(self, error, context: str, resource_name: str = ""):
         """
--- a/prowler/providers/googleworkspace/services/gmail/gmail_service.py
+++ b/prowler/providers/googleworkspace/services/gmail/gmail_service.py
@@ -49,10 +49,7 @@ class Gmail(GoogleWorkspaceService):
                             self.policies.enable_mail_delegation = value.get(
                                 "enableMailDelegation"
                             )
-                            logger.debug(
-                                "Gmail mail delegation setting fetched: "
-                                f"{self.policies.enable_mail_delegation}"
-                            )
+                            logger.debug("Gmail mail delegation setting fetched.")
 
                         elif setting_type == "gmail.email_attachment_safety":
                             self.policies.encrypted_attachment_protection_consequence = value.get(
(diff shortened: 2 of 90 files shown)


prowler/providers/googleworkspace/services/gmail/gmail_service.py
```diff
@@ -0,0 +1,219 @@
...
+                except Exception as error:
+                    self._handle_api_error(
+                        error,
+                        "fetching Gmail policies",
+                        self.provider.identity.customer_id,
+                    )
+                    fetch_succeeded = False
+                    break
+
+            self.policies_fetched = fetch_succeeded or self.policies != GmailPolicies()
```
HugoPBrito: Blocking: `policies_fetched` becomes `True` even when pagination fails after only a subset of settings has been read.

Because the request is not scoped to a fixed setting list, a later `429` or API error can leave some Gmail settings unread while checks still execute as if the fetch were complete. At that point `None` stops meaning "unknown" and starts being treated as a real result, which produces false PASS/FAIL findings depending on the check.

I think we should either:
1. keep `policies_fetched = False` on any pagination failure, or
2. track completeness per setting and make checks skip/report UNKNOWN when their specific setting was not fetched.

prowler/providers/googleworkspace/services/gmail/gmail_service.py
```diff
@@ -0,0 +1,219 @@
...
+
+            request = service.policies().list(pageSize=100)
+            fetch_succeeded = True
+            customer_level_settings = set()
+
+            while request is not None:
+                try:
+                    response = request.execute()
+
+                    for policy in response.get("policies", []):
```
HugoPBrito, line 43: Blocking: this now accepts any OU-level policy as a fallback for a domain-wide check.

All Gmail checks here emit a single finding for the whole customer (`resource_id = customer_id`, `location = "global"`), so using an `orgUnit` policy as the source of truth can turn one OU configuration into a domain-wide PASS/FAIL. If different OUs have different values, the result becomes order-dependent and no longer matches the scope of the finding.

Unless these checks are redesigned to report per-OU findings, I think we should only use customer-level policies here.

HugoPBrito (changes requested):
The new Gmail service is well structured, but I found two blocking issues that can produce incorrect domain-wide findings: partial fetches are treated as complete, and OU-level policies are used as fallback for global checks. Please address the inline comments before merge.

prowler/providers/googleworkspace/services/gmail/gmail_service.py
```diff
@@ -0,0 +1,212 @@
...
+                except Exception as error:
+                    self._handle_api_error(
+                        error,
+                        "fetching Gmail policies",
+                        self.provider.identity.customer_id,
+                    )
+                    fetch_succeeded = False
+                    break
+
+            self.policies_fetched = fetch_succeeded
```
HugoPBrito, line 161: Is this loss of partial-fetch tolerance intentional?

Previously this was `fetch_succeeded or self.policies != GmailPolicies()`, so a transient error (e.g. a 429 on page 2 of N) would still mark `policies_fetched=True` as long as page 1 had stored data. With the new strict assignment, any pagination error now invalidates everything already collected and the `policies_fetched` guard skips every Gmail check.

If this is the desired contract, could you add a regression test pinning the new strict semantics? The previous `test_gmail_fetch_policies_partial_fetch_still_marks_fetched` was removed and nothing replaces it, so the behavior is currently only implicit.

prowler/providers/googleworkspace/lib/service/service.py
```diff
@@ -41,21 +41,25 @@ def _build_service(self, api_name: str, api_version: str):
...
+        and rejects group-targeted and sub-OU policies.
         """
         policy_query = policy.get("policyQuery", {})
         if policy_query.get("group"):
             return False
-        if policy_query.get("orgUnit"):
-            return False
-        return True
+        org_unit = policy_query.get("orgUnit")
+        if not org_unit:
```
HugoPBrito, line 57: The docstring states *"The Cloud Identity Policy API scopes all policies to an OU"*, but the implementation still treats a missing `orgUnit` as customer-level:

```python
org_unit = policy_query.get("orgUnit")
if not org_unit:
    return True
```

If the API guarantee really holds, this branch is dead code and can be removed. If you're keeping it as a safety net, consider tweaking the docstring to reflect that — e.g. *"…scopes all policies to an OU; absence is treated as customer-level for safety"* — so a future reader doesn't see the docstring and the code disagree.

prowler/providers/googleworkspace/services/gmail/gmail_service.py
lydiavilchez: Yes, this is intentional, due to your earlier review feedback about partial data producing incorrect results. With the Gmail API filter now returning ~10 policies in a single page, rate limits shouldn't be an issue. I've added `test_gmail_fetch_policies_strict_fetch_failure`

prowler/providers/googleworkspace/lib/service/service.py
lydiavilchez: Updated the docstring to clarify it's a safety net.

HugoPBrito (approved):

jfagoagas (approved):

PR #10685, opened 2026-04-14
feat(ui): improve attack paths scan selection UX

Alan-TheGentleman:
## Summary

- Contextual button labels based on `graph_data_ready` instead of scan state (e.g. "Scheduled", "Running...", "Failed" instead of generic "Waiting...")
- Tooltip on disabled select button explaining why the scan cannot be selected
- Green dot indicator next to the account name on rows where graph data is available
- Warning banner when the selected scan is in a non-completed state (scheduled/executing) but graph data is available from a previous cycle, informing the user they are viewing older data
- Enable "Select" on scheduled/executing scans when graph data is already available from a previous cycle

## Behavior matrix

| State | graph_data_ready | Button label | Enabled | Tooltip |
|-------|-----------------|--------------|---------|---------|
| scheduled | false | Scheduled | No | "Graph will be available once this scan runs and completes." |
| scheduled | true | Select | Yes | — |
| available | false | Queued | No | "This scan is queued. Graph will be available once it completes." |
| executing | false | Running... | No | "Scan is running. Graph will be available once it completes." |
| executing | true | Select | Yes | — |
| completed | true | Select | Yes | — |
| failed | false | Failed | No | "This scan failed. No graph data is available." |
| failed | true | Select | Yes | — |
| (any) | (any) | Selected | No | — (already selected) |

## Test plan

- [x] All 8 scan-list-table tests pass (5 new)
- [x] All 23 attack paths tests pass
- [x] `pnpm run typecheck` — verified via local build

### DIFF
--- a/.opencode/package-lock.json
+++ b/.opencode/package-lock.json
@@ -1,0 +1,115 @@
+{
+  "name": ".opencode",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "dependencies": {
+        "@opencode-ai/plugin": "1.3.17"
+      }
+    },
+    "node_modules/@opencode-ai/plugin": {
+      "version": "1.3.17",
+      "resolved": "https://registry.npmjs.org/@opencode-ai/plugin/-/plugin-1.3.17.tgz",
+      "integrity": "sha512-N5lckFtYvEu2R8K1um//MIOTHsJHniF2kHoPIWPCrxKG5Jpismt1ISGzIiU3aKI2ht/9VgcqKPC5oZFLdmpxPw==",
+      "license": "MIT",
+      "dependencies": {
+        "@opencode-ai/sdk": "1.3.17",
+        "zod": "4.1.8"
+      },
+      "peerDependencies": {
+        "@opentui/core": ">=0.1.96",
+        "@opentui/solid": ">=0.1.96"
+      },
+      "peerDependenciesMeta": {
+        "@opentui/core": {
+          "optional": true
+        },
+        "@opentui/solid": {
+          "optional": true
+        }
+      }
+    },
+    "node_modules/@opencode-ai/sdk": {
+      "version": "1.3.17",
+      "resolved": "https://registry.npmjs.org/@opencode-ai/sdk/-/sdk-1.3.17.tgz",
+      "integrity": "sha512-2+MGgu7wynqTBwxezR01VAGhILXlpcHDY/pF7SWB87WOgLt3kD55HjKHNj6PWxyY8n575AZolR95VUC3gtwfmA==",
+      "license": "MIT",
+      "dependencies": {
+        "cross-spawn": "7.0.6"
+      }
+    },
+    "node_modules/cross-spawn": {
+      "version": "7.0.6",
+      "resolved": "https://registry.npmjs.org/cross-spawn/-/cross-spawn-7.0.6.tgz",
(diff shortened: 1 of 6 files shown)


alejandrobailo:
### Code review

Found 3 issues:

1. Changelog entry added under `## [1.23.1] (Prowler UNRELEASED)`, but the current active unreleased block on `master` is `## [1.24.0] (Prowler UNRELEASED)` (bumped in `22a9a7e58`). The entry should be moved to the `1.24.0` block.

https://github.com/prowler-cloud/prowler/blob/fcbabca2c1793515535f063b8a793db1ab3e5fd1/ui/CHANGELOG.md#L4-L10

2. `Alert` uses `className="border-border-warning-secondary bg-bg-warning-secondary"`, but no `--color-border-warning-secondary` CSS variable is defined in `ui/styles/globals.css` (only `--color-bg-warning-secondary` and `--color-text-warning-primary` exist). With Tailwind 4 utilities generated from `--color-*` vars, the border class resolves to nothing and the banner renders without a border. The shared `Alert` component already exposes a `warning` variant with the correct tokens — use `<Alert variant=\"warning\">` instead of hand-rolled classes.

https://github.com/prowler-cloud/prowler/blob/fcbabca2c1793515535f063b8a793db1ab3e5fd1/ui/app/%28prowler%29/attack-paths/%28workflow%29/query-builder/attack-paths-page.tsx#L383-L399

3. The banner description interpolates `selectedScan.attributes.state` directly. The condition `state !== COMPLETED && graph_data_ready` fires for states `scheduled`, `available`, `executing`, and `failed`, but only `EXECUTING` is formatted explicitly. For the other branches the raw enum value is rendered to the user — e.g. *\"This scan is currently available.\"* (technical jargon) and *\"This scan is currently failed.\"* (grammatically broken). Map each state to a user-friendly label (e.g. `scheduled → \"scheduled to run\"`, `available → \"queued\"`, `failed → \"failed\"`).

https://github.com/prowler-cloud/prowler/blob/fcbabca2c1793515535f063b8a793db1ab3e5fd1/ui/app/%28prowler%29/attack-paths/%28workflow%29/query-builder/attack-paths-page.tsx#L390-L397

josema-xyz:
- Scan list now polls every 5s while the page is visible, including when a scan is selected. Fixes the case where the table went stale after picking a scan or when a teammate started a new one.
- Dropped the duplicate graph-ready dot next to the Account column; the status badge already shows it.
- Switched the "previous cycle" banner from warning (yellow) to info (blue) since the orange clashed with the failed/in-progress state colors.
- Removed two obsolete tests that asserted on the old dot.

Alan-TheGentleman (commented):
~~Previous review was based on a stale diff from `gh pr diff` that showed changes not present in the actual branch (AutoRefresh removal, polling useEffect). Disregard.~~

Actual review after checking out the branch:

1. **`bg-green-500` hardcoded in the graph-ready dot indicator** (scan-list-table.tsx:148). The rest of the project uses semantic color tokens (`bg-bg-pass-secondary`, `text-text-success-primary`). Should use a semantic token for consistency — if `bg-bg-pass-secondary` is too light for a dot, consider `bg-bg-pass-primary` or create a dedicated token.

Everything else is solid: contextual labels match the behavior matrix, tooltip wrapper with `<span tabIndex={0}>` is the correct pattern for disabled buttons, `isViewingPreviousCycleData` is clean derived state with no useState mirror, `cn()` used correctly, 5 new tests cover the key states.

josema-xyz (approved):

Issue #10690, opened 2026-04-14
`prowler image --registry-list` crashes with `AttributeError: 'NoneType' object has no attribute 'print_credentials'`


eblume:
### Issue search

- [x] I have searched the existing issues and this bug has not been reported yet

### Which component is affected?

Prowler CLI/SDK

### Cloud Provider (if applicable)

Not applicable

### Steps to Reproduce

```bash
prowler image --registry https://any-registry.example.com --registry-list
```

### Expected behavior

lists repositories and tags, then exits cleanly.

### Actual Result with Screenshots or Logs

When run as `prowler image --registry https://registry.ops.eblu.me --image-filter "^blumeops/" --registry-list --no-color`:
```
                         _
 _ __  _ __ _____      _| | ___ _ __
| '_ \| '__/ _ \ \ /\ / / |/ _ \ '__|
| |_) | | | (_) \ V  V /| |  __/ |
| .__/|_|  \___/ \_/\_/ |_|\___|_|v5.23.0
|_| Get the most at https://cloud.prowler.com 

New! Send findings from Prowler CLI to Prowler Cloud
More details here: goto.prowler.com/import-findings

Date: 2026-04-14 14:31:50


Registry: https://registry.ops.eblu.me (36 repositories, 147 images)

  blumeops/alloy (6 tags)
    v1.14.0-613f05d, v1.14.0-613f05d-nix, v1.14.0-61f02a0, v1.14.0-61f02a0-nix, v1.14.0-fd0bebb, v1.14.0-fd0bebb-nix
  blumeops/alpine (1 tags)
    latest
  [[ ... Removed many images here for this issue ... ]]

Traceback (most recent call last):
  File "/Users/eblume/code/3rd/prowler/.venv/bin/prowler", line 10, in <module>
    sys.exit(prowler())
             ^^^^^^^^^
  File "/Users/eblume/code/3rd/prowler/prowler/__main__.py", line 294, in prowler
    global_provider.print_credentials()
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'print_credentials'
---EXIT CODE: 1

```

### How did you install Prowler?

Cloning the repository from github.com (git clone)

### Environment Resource

I've reproduced this issue when run as a CronJob in my homelab minikube cluster, as a local docker container, and built from source from a github clone.

### OS used

MacOS (local)
Debian Bookworm (container)

### Prowler version

[...]

PR #10470, opened 2026-03-25
fix(image): pass registry arguments through init_global_provider

eblume:
This PR resoles the corresponding issue:  https://github.com/prowler-cloud/prowler/issues/10457

PR #9985 added registry scan support with --registry, --image-filter, --tag-filter, --max-images, --registry-insecure, and --registry-list-images flags. PR #10128 accidentally removed these kwargs from the init_global_provider call, so the CLI parsed them but never forwarded them to the ImageProvider constructor.

This restores the six missing kwargs and adds a regression test.

**Please note, this is my first time contributing to Prowler**, I have done my best to follow the guidelines from CONTRIBUTING.md.

### Context

While deploying Prowler 5.22.0 as a weekly CronJob to scan my private OCI registry ([blumeops](https://github.com/eblume/blumeops)), I discovered that `prowler image --registry <url>` always fails with `ImageNoImagesProvidedError[11000]: No images provided for scanning`, regardless of registry configuration.

Root cause: PR #9985 (merged 2026-02-19) correctly added registry kwargs (`registry`, `image_filter`, `tag_filter`, `max_images`, `registry_insecure`, `registry_list_images`) to the `init_global_provider` call. PR #10128 (merged 2026-02-24) accidentally removed them — likely a merge conflict resolution that took the older version of the `ImageProvider` constructor call.

My current workaround is an [init container that enumerates images via the registry's `/v2/_catalog` API](https://forge.eblu.me/eblume/blumeops/src/commit/75fd5b029d682e2f2a1337519bc6f3c5a1a210fc/argocd/manifests/prowler/cronjob-image-scan.yaml) and generates an `--image-list` file, bypassing `--registry` entirely.

### Description

Restores the six missing registry-related kwargs to the `init_global_provider` call for the image provider in `prowler/providers/common/provider.py`:

- `registry`
- `image_filter`
- `tag_filter`
- `max_images`
- `registry_insecure`
- `registry_list_images`

These are parsed by the CLI argument parser and accepted by `ImageProvider.__init__`, but were not being forwarded between the two.

Adds a regression test that exercises the full CLI-to-registry-enumeration path: it passes `--registry` and `--image-filter` arguments through `init_global_provider`, mocks the registry adapter, and asserts that matching images are discovered and filtered correctly. This reproduces the exact `ImageNoImagesProvidedError` when the fix is reverted.

### Steps to review

[...]

PR #10691, opened 2026-04-14
fix(image): --registry-list crashes with AttributeError on global_provider

eblume:
### Context

Fix for #10690

Discovered while testing the v5.23.0 release against a private OCI registry ([blumeops](https://github.com/eblume/blumeops)). The `--registry` fix from PR #10470 works correctly for scanning, but `--registry-list` (added in the same feature set) crashes immediately after printing the listing.

**Note on base branch:** This PR is based on `master` rather than the `5.23.0` tag where I originally developed and tested the fix. The fix was straightforward to rebase but I was unable to run the full test suite locally due to missing transitive dependencies (`markdown_it` via `jira.py`). The image provider tests (221 tests) all pass. If there are issues with the broader suite, happy to work through them.

### Description
`ImageProvider.__init__` returns early in listing-only mode (line 167) before reaching `Provider.set_global_provider(self)` at the bottom of the constructor. This leaves `Provider._global` as `None`, so `__main__.py:294` crashes when it calls `global_provider.print_credentials()`.

This PR refactors the early return into a guard clause and adds a clean exit in `__main__.py` for listing-only mode.

**Design note:** The minimal fix would be to add `Provider.set_global_provider(self)` before the early return:

```python
if self._listing_only:
    Provider.set_global_provider(self)
    return
```
I opted instead to remove the early return and use a guard (if not self._listing_only:) so that `set_global_provider` is only called once, at the bottom of `__init__`, matching every other provider. Happy to go with the simpler approach if you prefer.

### Steps to review

1. Read the regression test in `tests/providers/image/image_provider_test.py::TestRegistryListMode` which reproduces the exact crash sequence from `__main__.py` (init_global_provider → get_global_provider → print_credentials).
2. Revert the `image_provider.py` change and run the test to see it fail with `AttributeError: 'NoneType' object has no attribute 'print_credentials'`.
3. Restore the change and confirm all 221 image provider tests pass: `pytest tests/providers/image/ -xvs`
4. Review the `__main__.py` change: after `print_credentials`, check for `_listing_only` and exit. This matches the pattern used by other `--list-*` flags (lines 253-286).
5. Optional end-to-end: `prowler image --registry <any-registry> --registry-list` should print the listing and exit cleanly instead of crashing.

### Checklist

<details>

</details>

#### SDK/CLI
- Are there new checks included in this PR?  **No**

#### UI
N/A

#### API
N/A

### License


andoniaf:
Merged master to resolve the changelog conflict and added defensive initialization of _audit_config, _fixer_config, _mutelist, and audit_metadata before the if not self._listing_only: branch. Without these defaults, any future code that accesses those attributes in listing-only mode would hit the same class of AttributeError this PR fixes.

And tested it locally:
image

Thanks @eblume 🙌

andoniaf (dismissed):

andoniaf (approved):

Issue #6240, opened 2024-12-18
Environment-Specific Log Group Retention Policy Validation


prit2477:
### New feature motivation

When scanning multiple accounts, we often have environment-specific log group retention policies:

Sandbox: 7 days
Staging: 30 days
Production: 365 days
Currently, these policies may be flagged as non-compliant unless explicitly configured for each environment. For example, a Sandbox log group with a retention policy of exactly 7 days should pass the compliance check, while any configuration below 7 days should be flagged as non-compliant. Similarly, Staging and Production environments should align with their respective policies.
Introduce functionality to allow environment-specific log group retention policies to be validated dynamically based on predefined rules.

Expected Behavior:

Ability to define retention policies per environment (e.g., Sandbox, Staging, Production).
Ensure that compliance checks only flag log groups as non-compliant when their configuration deviates from the defined environment-specific rules.
Support for handling multiple accounts and environments seamlessly.

### Solution Proposed

Add a configuration file or parameter to specify retention rules for each environment.
Update compliance logic to check against these rules during scans.
Provide detailed output for any non-compliance based on environment-specific thresholds

### Describe alternatives you've considered

This feature would be useful for organizations managing multiple cloud accounts with distinct compliance requirements for different environments. It ensures accurate reporting without false positives and better adherence to organizational policies.

pedrooot:
Hi @prit2477 , this is a great idea, thanks for opening the issue!

What comes to mind is making the check configurable using the config.yaml file that we currently have in Prowler. This way, the value for log retention days can change depending on the environment you’re working in.

We are reviewing this feature with the team and will let you know as soon as we have a solution. Thank you so much!!!

Dharshukutti:
Hi! I'd like to work on this issue. I'll start by implementing environment-specific CloudWatch Log Group retention validation using `config.yaml`, and I'll share a PR once it's ready.

PR #10700, opened 2026-04-15
feat(sdk): Enhance dynamic provider loading and compliance framework

StylusFrost:
### Context

Enables external and custom providers to be discovered and integrated through Python
entry points, without modifying Prowler's core codebase.

### Description

External and custom providers can now register themselves through Python entry points and
run through the same flows as built-in providers — output formatting, compliance,
mutelist, and check resolution — with no changes to Prowler's core.

- External providers, their checks, and their compliance frameworks are discovered via
  entry points and merged with the built-ins.
- A provider contract on the `Provider` base class lets a plug-in supply its own CLI
  parsing, output, compliance, and mutelist behavior; anything it does not implement falls
  back to sensible defaults.
- Built-in providers always take precedence on a name collision: a plug-in registered
  under a built-in's name is ignored with a warning and never executed.
- `prowler --help` lists installed external providers alongside the built-ins.

No behavior change for built-in providers.

Deferred follow-ups: surfacing plug-in override events; a fixer-module contract for
external providers.

### Steps to review

1. Built-in providers behave exactly as before: `prowler aws --list-checks`,
   `prowler aws --list-compliance`.
2. `prowler --help` lists any installed external providers in the usage/epilog.
3. A plug-in registered under a built-in name (e.g. `aws`) is ignored with a warning and
   never executed.

### Checklist

<details>

</details>

#### SDK/CLI

#### UI

#### API

### License


HugoPBrito (changes requested):
Request changes — see inline comments for the specifics. One cross-cutting issue that doesn't anchor to any single line in the diff:

### YAML namespace unwrap only knows about 5 providers

`prowler/config/config.py:244-253` only sniffs `aws`, `gcp`, `azure`, `kubernetes` and `m365`. Prowler ships 16+ built-ins, so twelve are outside the list (`nhn`, `github`, `googleworkspace`, `cloudflare`, `iac`, `llm`, `image`, `mongodbatlas`, `oraclecloud`, `openstack`, `alibabacloud`, `vercel`).

A `github` user writing a namespaced config:

```yaml
github:
  token: abc
```

The sniff finds none of the five keys, falls into the `else` at line 250, `github` isn't in the safeguard at line 252, and the function returns `{'github': {'token': 'abc'}}` whole. `github_provider` receives the wrapped dict instead of the inner block.

This is a latent bug in master for 12 built-ins; the PR inherits it and additionally blocks any external plug-in from using the namespaced format. One change covers both:

```python
if provider in config_file:
    config = config_file.get(provider, {})
else:
    config = config_file if config_file else {}
```

prowler/providers/common/provider.py
```diff
@@ -136,6 +137,99 @@ def get_checks_to_execute_by_audit_resources(self) -> set:
...
+        External providers can override this to display a detailed
+        compliance table (e.g., per-section breakdown). Return True
+        if the table was rendered, False to fall back to the generic table.
+        """
+        raise NotImplementedError(
+            f"{self.__class__.__name__} has not implemented display_compliance_table()"
+        )
+
+    @property
+    def is_external_tool_provider(self) -> bool:
```
HugoPBrito: This property is meant to replace the `EXTERNAL_TOOL_PROVIDERS` frozenset, but the runtime never asks for it — it still consults the frozenset directly:

- `prowler/__main__.py:223` (skip compliance loading)
- `prowler/__main__.py:281` (skip service/check loading)
- `prowler/__main__.py:397` (route to the tool-wrapper branch)
- `prowler/lib/check/models.py:258,287,303,313,326,332,345,371,380` (relax metadata schema validations)

`iac`, `llm` and `image` are in the frozenset and keep working. An external plug-in that declares itself a tool wrapper through this property never lands in those branches — Prowler tries to load compliance and services for it as a regular provider, and the scan ends up empty or crashes on the wrong path.

In each of those spots the check needs to combine the frozenset with this property on the provider class.

prowler/providers/common/provider.py
```diff
@@ -398,6 +500,9 @@ def init_global_provider(arguments: Namespace) -> None:
                         mutelist_path=arguments.mutelist_file,
                         fixer_config=fixer_config,
                     )
+                else:
+                    # Dynamic fallback: any external/custom provider
+                    provider_class.from_cli_args(arguments, fixer_config)
```
HugoPBrito: The contract above declares `from_cli_args(...) -> "Provider"`, but here the return value is discarded. Built-ins survive because all of them call `Provider.set_global_provider(self)` from inside `__init__` (mongodbatlas, openstack, vercel, azure, oraclecloud, googleworkspace, github, iac, image, kubernetes, …). The side-effect carries the wiring.

A provider that implements the literal contract (`return cls(...)` with no side-effect in `__init__`) leaves `Provider._global = None` and crashes in `__main__.py` at `global_provider.print_credentials()`. Nothing in the docstring documents the side-effect requirement.

Honoring the type hint at the call site avoids the ambiguity:

```python
provider_instance = provider_class.from_cli_args(arguments, fixer_config)
if provider_instance:
    Provider.set_global_provider(provider_instance)
```

prowler/lib/check/utils.py
```diff
@@ -19,24 +41,34 @@ def recover_checks_from_provider(
...
+                if (
+                    check_module_name.count(".") == 6
+                    and ".lib." not in check_module_name
+                    and (not check_module_name.endswith("_fixer") or include_fixers)
+                ):
+                    check_path = module_name.module_finder.path
+                    check_name = check_module_name.split(".")[-1]
+                    check_info = (check_name, check_path)
+                    checks.append(check_info)
+        except ModuleNotFoundError:
```
HugoPBrito: Two issues here that compound:

1. The `sys.exit(1)` runs before checking entry points. `prowler dockerdesktop --service container` raises `ModuleNotFoundError` (no built-in `prowler.providers.dockerdesktop.services.container`), service is set, and we exit before asking entry points.
2. The `_recover_ep_checks(provider)` call below at line 70 is gated by `if not service`, so even if we didn't exit here, entry points would only be loaded when there's no `--service`.

This breaks both:
- Full external providers with `--service`.
- External checks added to a built-in provider with `--service some_new_service`.

Fix: teach `_recover_ep_checks` to accept and filter by service (the entry-point path already embeds it: `prowler_artifacts_dockerdesktop.services.container.{check}.{check}`), try that lookup before exiting, and drop the `if not service` gate at line 70 so entry points are always considered (passing `None` as filter when there's no `--service`).

prowler/providers/common/arguments.py
```diff
@@ -23,6 +23,16 @@ def init_providers_parser(self):
                 ),
                 init_provider_arguments_function,
             )(self)
+        except ImportError:
```
HugoPBrito: This bare `except ImportError:` swallows transitive import errors in built-ins. When a built-in's `lib.arguments.arguments` module exists but its imports fail at runtime (e.g. `boto3` not installed in the environment), the except treats it as "must be external, try entry points". `_load_ep_provider("aws")` returns `None` because AWS isn't an entry-point provider, the `if cls and hasattr(...)` is `False`, and the code silently moves on. AWS ends up with no parser registered.

What the user sees:

```
$ prowler aws --profile myprofile
prowler: error: argument provider: invalid choice: 'aws' (choose from ...)
```

No mention of `boto3`, no `ImportError`, no signal of what's actually missing.

`ImportError.name` already tells you the exact module that failed, which is enough to tell the two cases apart:

```python
except ImportError as e:
    expected_prefix = f"{providers_path}.{provider}"
    if e.name and (e.name == expected_prefix or e.name.startswith(expected_prefix + ".")):
        # Provider's own module path is missing → external provider, try entry points
        cls = Provider._load_ep_provider(provider)
        if cls and hasattr(cls, "init_parser"):
            try: cls.init_parser(self)
            except Exception as error: logger.warning(...)
    else:
        # Transitive dependency missing — built-in is there, something it imports isn't
        logger.critical(
            f"Failed to load arguments for provider '{provider}': "
            f"missing dependency '{e.name}'. "
            f"Make sure all required dependencies are installed (try 'pip install prowler[{provider}]')."
        )
        sys.exit(1)
```

Same shape applies at `provider.py:264`.

prowler/providers/common/provider.py
```diff
@@ -159,13 +253,21 @@ def set_global_provider(global_provider: "Provider") -> None:
...
-            )
+            # Try built-in provider first, fall back to entry point
+            provider_class = None
+            try:
+                provider_class_path = f"{providers_path}.{arguments.provider}.{arguments.provider}_provider"
+                provider_class_name = f"{arguments.provider.capitalize()}Provider"
+                provider_class = getattr(
+                    import_module(provider_class_path), provider_class_name
+                )
+            except (ImportError, AttributeError):
```
HugoPBrito: Same root cause as the inline comment on `arguments.py:26` — this `(ImportError, AttributeError)` catches transitive import failures in built-ins and re-raises with a misleading message ("not found as built-in or entry point") when the real reason is a missing dependency. Apply the same `e.name`-based discriminator here so the user sees the actual missing module.

prowler/providers/common/provider.py
StylusFrost: Thanks @HugoPBrito !  

Addressed in [`60e76570`](https://github.com/prowler-cloud/prowler/pull/10700/commits/60e7657086fd45c4e2b52a80b83fd96d1dbf8ecb):

new `Provider.is_tool_wrapper_provider(provider)` helper combines the frozenset with the per-provider flag, and replaces all 12 call sites.

Note: changed the flag from `@property` to class attribute so the `CheckMetadata` validators can read it from the class (they run before any instance exists). Happy to revert if you prefer keeping it as a property.

prowler/providers/common/provider.py
StylusFrost: Addressed in [`cf70d1f9`](https://github.com/prowler-cloud/prowler/pull/10700/commits/cf70d1f9f). 

Call site now uses the returned instance and wires it via `set_global_provider`.Tolerated `None` returns for providers that already set the global from `__init__`, so existing implementations keep working.
                                                                                                                                                                                             
Docstring updated to document both patterns. Test added covering a pure-contract provider (returns instance, no side-effect in `__init__`).

prowler/lib/check/utils.py
StylusFrost: Thanks @HugoPBrito !  
 
Fixed in [`0883baad`](https://github.com/prowler-cloud/prowler/pull/10700/commits/0883baad7):

  - `_recover_ep_checks(provider, service=None)` filters by `.services.{service}.` in `ep.value` (the dotted path the plugin declares in `pyproject.toml`) — same convention as built-in discovery, just applied to the entry-point path.
  - `recover_checks_from_provider` no longer exits early in the `except ModuleNotFoundError`; entry points are consulted in both paths, and the gate `if not service:` before `_recover_ep_checks` is gone. If the requested service matches nothing in either source, we fail with a clear message that mentions both built-ins and entry points.                       
   
5 new tests covering: service-filtered entry-point discovery, external provider with `--service`, built-in with a new external service, and typo-service clean failure.

prowler/providers/common/arguments.py
StylusFrost: Fixed in [`907166d8`](https://github.com/prowler-cloud/prowler/pull/10700/commits/907166d88) — took a different angle than `e.name` matching:                                              
  
 Added `Provider.is_builtin(provider)` helper using `importlib.util.find_spec` to discriminate built-in vs external **upfront**, before the import attempt. In both call sites (`arguments.py:init_providers_parser` and `provider.py:init_global_provider`) the flow is now `if Provider.is_builtin(provider): <builtin path> else: <external path>`. Inside the built-in path, any `ImportError` is treated as a transitive dependency failure — we log it with the missing module name and `sys.exit(1)` with `exc_info=True` preserving the traceback.           
                                                                              
Chose `find_spec` over `e.name` matching because `ImportError.name` is optional (can be `None` on manually-raised ImportError or in some edge cases) and relies on post-hoc inspection of  the exception, whereas `find_spec` is Python's canonical API for module existence checking.
                                                                                                                                                                                             
  Tests added: `TestIsBuiltinProvider` (helper), `TestInitProvidersParserBuiltinDependencyFailure`, `TestInitGlobalProviderBuiltinDependencyFailure` — covering built-in with missing dep fails loudly, external provider takes the entry-point path, and find_spec edge cases.

prowler/providers/common/provider.py
StylusFrost: Same `find_spec` approach applied after [907166d8](https://github.com/prowler-cloud/prowler/pull/10700/commits/907166d88)

prowler/lib/check/utils.py
```diff
@@ -19,24 +51,44 @@ def recover_checks_from_provider(
             return []
 
         checks = []
```
HugoPBrito, line 60: `Provider.is_tool_wrapper_provider()` was added in [`60e76570`](https://github.com/prowler-cloud/prowler/pull/10700/commits/60e7657086fd45c4e2b52a80b83fd96d1dbf8ecb) as the single source of truth combining `EXTERNAL_TOOL_PROVIDERS` with the per-provider class attribute, and it's wired into `__main__.py` (3 sites) and the 9 `CheckMetadata` validators. Two sites in this file were missed, and one of them silently narrows the existing behavior:

- `utils.py:53` (`recover_checks_from_provider`) — master had `provider in EXTERNAL_TOOL_PROVIDERS`, the PR replaced it with the literal tuple `("iac", "llm", "image")`. Same set, but the source of truth is duplicated again — exactly the thing the helper was meant to consolidate.
- `utils.py:119` (`recover_checks_from_service`) — master had `provider in EXTERNAL_TOOL_PROVIDERS`, the PR shrunk it to `provider == "iac"`. LLM and Image stop bypassing here. Today only `aws_provider.py:944` calls this function so the runtime impact is nil, but it's a behavior change unrelated to this PR's scope and breaks symmetry with `:53`.

Knock-on effect for external plug-ins: a provider that declares `is_external_tool_provider = True` bypasses in `__main__` and the validators but not here, so the contract leaks.

Fix in both spots:

```python
from prowler.providers.common.provider import Provider
...
if Provider.is_tool_wrapper_provider(provider):
    return []  # or set() at :119
```

and drop the literal tuple / `== "iac"` line and the comment that lists the providers.

prowler/providers/common/provider.py
```diff
@@ -398,6 +530,18 @@ def init_global_provider(arguments: Namespace) -> None:
                         mutelist_path=arguments.mutelist_file,
                         fixer_config=fixer_config,
                     )
+                else:
```
HugoPBrito, line 606: This `else` is the right shape, but the chain above it (`if "aws" in provider_class_name.lower(): ... elif "azure" in ...: ... elif "github" in ...:`) is `substring`, not equality. Pre-existing, but this PR turns it from theoretical to active footgun: external providers register through entry points and there's nothing stopping a plug-in name from colliding with a built-in substring.

| Plug-in name        | Branch taken | Outcome                                                  |
|---------------------|--------------|----------------------------------------------------------|
| `aws_lite`          | aws          | `TypeError` on `retries_max_attempts=...`                |
| `awsx`              | aws          | `TypeError`                                              |
| `azure_gov`         | azure        | `TypeError` on `az_cli_auth=...`                         |
| `gcp_org`           | gcp          | `TypeError`                                              |
| `github_enterprise` | github       | `TypeError` on `personal_access_token=...`               |
| `iac_v2`            | iac          | `TypeError`                                              |

And the `except TypeError` at the bottom of the function swallows the traceback and prints a generic *"provider configuration error"* — the user has no signal that their plug-in is being routed to the wrong branch.

Two ways out:

- Cheap: switch to equality on the bare provider name — `provider_class_name.lower().removesuffix("provider") == "aws"` and friends. Stops the collision.
- Right: every built-in implements `from_cli_args` and the dispatch becomes a single `provider_class.from_cli_args(arguments, fixer_config)` call — this `else` stops being a fallback and becomes the path. The 12-method contract you defined points this way already.

If neither lands in this PR, at least add a test that registers a plug-in named `awsx` and asserts the error surfaces clearly instead of dying inside the AWS branch.

prowler/lib/cli/parser.py
```diff
@@ -20,19 +20,58 @@
...
+            if extra_lines:
+                extra_providers_text = "\n" + "\n".join(extra_lines)
+
         # CLI Arguments
         self.parser = argparse.ArgumentParser(
             prog="prowler",
             formatter_class=RawTextHelpFormatter,
-            usage="prowler [-h] [--version] {aws,azure,gcp,kubernetes,m365,github,googleworkspace,nhn,mongodbatlas,oraclecloud,alibabacloud,cloudflare,openstack,vercel,dashboard,iac,image} ...",
-            epilog="""
+            usage=f"prowler [-h] [--version] {{aws,azure,gcp,kubernetes,m365,github,googleworkspace,nhn,mongodbatlas,oraclecloud,alibabacloud,cloudflare,openstack,vercel,dashboard,iac,image{extra_providers_csv}}} ...",
```
HugoPBrito: The static `usage` line dropped `llm`:

```diff
- usage="prowler [-h] [--version] {aws,azure,gcp,kubernetes,m365,github,googleworkspace,nhn,mongodbatlas,oraclecloud,alibabacloud,cloudflare,openstack,vercel,dashboard,iac,image,llm} ...",
+ usage=f"prowler [-h] [--version] {{...,vercel,dashboard,iac,image{extra_providers_csv}}} ...",
```

The epilog at L75 still lists it, so `prowler --help` ends up self-contradictory: `llm` shows up in the detailed provider list but not in the usage line that argparse prints first. Just put `,llm` back in the usage string.

HugoPBrito (changes requested):
Three more issues — see inline comments.

prowler/lib/cli/parser.py
StylusFrost: Thanks Hugo!, good catch!,                                                                                                                                                                              
Fixed in a5de6608a.

prowler/providers/common/provider.py
StylusFrost: Thanks Hugo! 👍 

Fixed in 52f6653cc. 

Went with the cheap path — equality on arguments.provider — so external plug-ins whose names contain a built-in substring (awsx, aws_lite, azure_gov, gcp_org, github_enterprise, iac_v2) fall through to the dynamic else and use from_cli_args instead of being silently captured by the wrong built-in branch. 

Added a parametrized regression test covering all six cases. Kept provider_class_name in scope (as a noqa: F841) so downstream forks that extend the dispatch with their own custom built-in branches don't break on merge.

prowler/lib/check/utils.py
StylusFrost: Fixed in 7836905. 

Both sites now consult Provider.is_tool_wrapper_provider(provider) — the single source of truth that combines EXTERNAL_TOOL_PROVIDERS with the per-provider is_external_tool_provider class attribute.

Extended in 79f12f361: the helper now lives in a leaf module prowler.lib.check.tool_wrapper, and Provider.is_tool_wrapper_provider delegates to it. This breaks the import cycle CodeQL flagged on the lazy Provider imports, and lets us migrate prowler/lib/check/checks_loader.py:29 to the same source of truth — the last remaining site that was still consulting EXTERNAL_TOOL_PROVIDERS directly. External plug-ins declaring is_external_tool_provider = True now bypass consistently across __main__, the CheckMetadata validators, check discovery, and the checks loader.

prowler/lib/check/check.py
```diff
@@ -362,6 +362,29 @@ def import_check(check_path: str) -> ModuleType:
...
 
 
+def _resolve_check_module(
+    provider_type: str, service: str, check_name: str
+) -> ModuleType:
+    """Resolve and import a check module — tries built-in path first, then entry points."""
+    # Built-in path
+    builtin_path = f"prowler.providers.{provider_type}.services.{service}.{check_name}.{check_name}"
+    try:
+        return import_check(builtin_path)
```
HugoPBrito: When a plugin re-registers an existing built-in `CheckID` (an edit over Prowler's own check), the executed module and the loaded metadata come from different sources:

- `_resolve_check_module` tries the **built-in** path first and returns it. Entry points are only consulted if the built-in import fails, so the executed module is the built-in one.
- `CheckMetadata.get_bulk()` (`models.py:419-432`) walks the loader output (built-ins first, entry points appended last) and does `bulk[check_id] = metadata`. Last write wins, so the **plugin's** metadata replaces the built-in's for the same `CheckID`.

Net effect: built-in code runs with the plugin's `Severity`, `Description`, `Remediation`, etc. A finding generated by code that expected `LOW` is published with the plugin's `HIGH`. Audit assumptions silently break.

Pick precedence once and apply it to **both** module and metadata — never split the source. Suggested: plugin wins (registering a duplicate is most likely an intentional edit, not a coincidence), with a `logger.warning` naming the overridden built-in.

Test: register an entry-point check sharing a built-in `CheckID`, run the check, and assert both `lib` and `check.metadata` come from the same source.

prowler/lib/check/check.py
```diff
@@ -362,6 +362,29 @@ def import_check(check_path: str) -> ModuleType:
...
+    except ModuleNotFoundError:
```
HugoPBrito: Same masking pattern as `recover_checks_from_provider` in `utils.py`, in a different file:

```python
try:
    return import_check(builtin_path)
except ModuleNotFoundError:
    pass  # falls through to entry points
```

If a built-in check exists but its module fails to import for an unrelated reason (broken transitive dep, misconfigured client, etc.), the `ModuleNotFoundError` is caught silently and we resolve via entry points. If a plugin happens to register the same name, **it silently runs in place of the built-in**. This is worse than just hiding errors: a plugin can hijack a built-in check by causing its import to fail.

Apply the same fix:

```python
if importlib.util.find_spec(builtin_path) is not None:
    return import_check(builtin_path)  # any error here is real, surface it
# Only fall through to entry points if the built-in module truly doesn't exist
```

Test: patch a built-in check so its module raises `ImportError("missing: foo")` and assert `_resolve_check_module` re-raises that error instead of falling back to the entry-point loader.

prowler/lib/outputs/outputs.py
```diff
@@ -7,39 +7,46 @@
 from prowler.lib.outputs.finding import Finding
 
 
-def stdout_report(finding, color, verbose, status, fix):
+def stdout_report(finding, color, verbose, status, fix, provider=None):
```
HugoPBrito, line 10: `stdout_report()` was extended to accept `provider=None`, but the call site in `report()` (line 88) doesn't pass it:

```python
stdout_report(finding, color, verbose, status, fixer)
```

So for any external/custom provider, the `else` branch inside `stdout_report` falls back to `Provider.get_global_provider()`. That defeats the purpose of the new parameter and re-introduces the dependency on the global singleton this PR set out to remove.

Pass the local `provider` through:

```python
stdout_report(finding, color, verbose, status, fixer, provider=provider)
```

Test: invoke `report()` after setting `Provider._global_provider = None` (or to a different provider) and assert the output uses the `provider` argument.

prowler/lib/check/utils.py
```diff
@@ -15,29 +47,53 @@ def recover_checks_from_provider(
...
+                if (
+                    check_module_name.count(".") == 6
+                    and ".lib." not in check_module_name
+                    and (not check_module_name.endswith("_fixer") or include_fixers)
+                ):
+                    check_path = module_name.module_finder.path
+                    check_name = check_module_name.split(".")[-1]
+                    check_info = (check_name, check_path)
+                    checks.append(check_info)
+        except ModuleNotFoundError:
```
HugoPBrito: `list_modules` calls `walk_packages`, which imports each package to walk it. Catching `ModuleNotFoundError` broadly here masks two distinct conditions:

```python
try:
    modules = list_modules(provider, service)
    ...
except ModuleNotFoundError:
    pass  # falls through to entry points
```

If a built-in service exists but one of its modules fails because of a broken transitive dependency, `walk_packages` raises `ModuleNotFoundError` with a different `name` than the service path. We swallow it and silently fall through to entry points. The user either never sees the real error or ends up running an entry-point check that happens to share the same name (see the duplicate-CheckID comment).

Distinguish "service doesn't exist" from "service exists but failed to import":

```python
service_path = (
    f"prowler.providers.{provider}.services.{service}"
    if service else f"prowler.providers.{provider}.services"
)
if importlib.util.find_spec(service_path) is not None:
    modules = list_modules(provider, service)  # any error here is real, surface it
    ...
# Only fall through to entry points if the built-in service truly doesn't exist
```

Test: patch a built-in service so one of its modules raises `ImportError("missing: foo")` and assert `recover_checks_from_provider` re-raises instead of silently falling back.

prowler/__main__.py
```diff
@@ -236,7 +235,7 @@ def prowler():
     logger.debug("Loading compliance frameworks from .json files")
 
     # Skip compliance frameworks for external-tool providers
-    if provider not in EXTERNAL_TOOL_PROVIDERS:
+    if not Provider.is_tool_wrapper_provider(provider):
```
HugoPBrito, line 248: Nice work introducing `Provider.is_tool_wrapper_provider()` as the single source of truth here (and at `:303`, `:426`). But line 202 still uses the hardcoded list:

```python
elif default_execution and provider not in ["iac", "llm"]:
    args.output_formats.extend(get_available_compliance_frameworks(provider))
```

Any new tool-wrapper provider (built-in or external) silently misses this branch unless someone also remembers to edit this list — exactly the bug the helper was meant to prevent.

```python
elif default_execution and not Provider.is_tool_wrapper_provider(provider):
    args.output_formats.extend(get_available_compliance_frameworks(provider))
```

A grep for `["iac", "llm"]` / `"iac"` / `"llm"` would also help confirm no other call site duplicates the same decision.

HugoPBrito (changes requested):

prowler/__main__.py
StylusFrost: Fixed in 5e876579f. 

The line now uses Provider.is_tool_wrapper_provider(provider) so the compliance-framework branch is skipped for any tool wrapper - built-in (iac, llm, image) or external plug-in declaring is_external_tool_provider = True. No behavior change for the built-ins (image already returned an empty list of frameworks, so extend([]) was a no-op).

Per your suggestion I grepped for ["iac", "llm"] / "iac" / "llm" / "image" across prowler/. The remaining hits fall into two buckets:

  - scan.py:97 and scan.py:164 use provider.type in ("iac", "image") - note the asymmetry: llm is intentionally excluded because it has no synthetic llm_scan check. Consolidating these to is_tool_wrapper_provider would route llm into a branch that tries to execute a non-existent check. I'm leaving these as-is; happy to revisit if you'd prefer a separate refactor that introduces a per-tool-wrapper "synthetic check name" registry.
  
  - __main__.py:392-396, 428 and the dispatch chain in provider.py are per-provider branches with different logic per tool wrapper (each one needs distinct kwargs / setup), so the binary helper doesn't apply there.

If you'd rather see scan.py consolidated too, let me know and I'll open a separate PR with the synthetic-check-registry approach

prowler/lib/check/check.py
StylusFrost: Fixed in [`82132a93`](https://github.com/prowler-cloud/prowler/pull/10700/commits/82132a93).                                                                                               
                                                                                                                                                                                             
  `_resolve_check_module` now uses `importlib.util.find_spec` to check if the built-in module exists before importing - same pattern already applied in `init_global_provider` via `Provider.is_builtin()`. If the built-in is absent, we fall through to entry points cleanly. If the built-in exists but its import fails (broken transitive dep, etc.), the error propagates instead of being silently swallowed and replaced by an entry-point plug-in that happens to share the check name. 

Added a regression guard `test_resolve_check_module_surfaces_error_when_builtin_import_fails` that registers a same-named entry-point check and asserts it does NOT take over when the built-in raises `ImportError`.

prowler/lib/check/utils.py
StylusFrost: Fixed in [`82132a93`](https://github.com/prowler-cloud/prowler/pull/10700/commits/82132a93).                                                                                               
                                                                                                                                                                                             
 `recover_checks_from_provider` now distinguishes "service doesn't exist" from "service exists but failed to import" via `importlib.util.find_spec(service_path)`, exactly as you suggested. Removed the `except ModuleNotFoundError: pass` that was masking real import errors. Built-in services that fail to import now surface the error through the function's global exception handler (logs the original `ImportError` class and message at CRITICAL, then `sys.exit(1)`) instead of silently routing to entry points.                                                   

Added a regression guard `test_recover_checks_surfaces_error_when_builtin_service_import_fails` that mocks a broken built-in service and asserts the import error wins over a same-named entry-point plug-in.

prowler/lib/outputs/outputs.py
StylusFrost: Fixed in [`e7f23bb1`](https://github.com/prowler-cloud/prowler/pull/10700/commits/e7f23bb1).

`report()` now passes its local `provider` through to `stdout_report`, so the dynamic else inside `stdout_report` no longer needs to fall back to the global singleton for external providers. Wires the new parameter the way the change was intended.

Added a regression guard `test_report_propagates_provider_to_stdout_report` that clears `Provider._global` and asserts `report()` still renders the finding using the local provider's `get_stdout_detail()`, proving the argument is being used instead of the global.

prowler/lib/check/check.py
StylusFrost: Fixed in [`c7aa5368`](https://github.com/prowler-cloud/prowler/pull/10700/commits/c7aa5368).

Picked precedence and applied it consistently across the three lookups so the executed module and the loaded metadata always come from the same source. The chosen direction is **built-in wins** on collision, with a `logger.warning` naming the shadowed plug-in:

1. `_resolve_check_module` (check.py): built-in first, entry point only when the built-in is absent.
2. `CheckMetadata.get_bulk` (models.py): first-write-wins on `CheckID`, plug-in duplicate is ignored and a warning is emitted naming the metadata file.
3. `Provider.init_global_provider` (provider.py): same pattern extended to the provider name itself, since the same divorce is possible there (built-in provider class running plug-in checks). When a plug-in registers a name colliding with a built-in, the built-in is loaded and the plug-in is announced as ignored.

Reasoning for built-in wins instead of plug-in wins: a security tool benefits more from fail-loud predictability than from permissive overrides. A plug-in lowering a check's severity, or shadowing a provider's session setup, would silently change the user's security posture. With built-in wins, plug-ins remain first-class extenders (new providers, new services, new checks under new CheckIDs all keep working) but cannot replace existing built-ins. Override is still possible by renaming, and the warning tells the user how.

Tests:
- `test_resolve_check_module_builtin_wins_over_entry_point`: module resolution with both sources present, asserts built-in wins and the plug-in is not loaded.
- `test_get_bulk_builtin_wins_on_check_id_collision`: metadata loader with a duplicate CheckID, asserts built-in metadata stays and a warning is emitted naming the plug-in path.
- `test_init_global_provider_warns_when_plugin_shadowed_by_builtin`: provider name collision, asserts the warning fires before dispatch.

StylusFrost:
### Follow-up fix on `find_spec` for external providers

#### Problem

`importlib.util.find_spec` on a dotted submodule path **imports the parent package** to resolve the child. For an external provider — whose package does not live under `prowler.providers.{provider}` — that parent import raises `ModuleNotFoundError` and **propagates**; it does not return `None`.

Reproducer:

```python
>>> importlib.util.find_spec("prowler.providers.<external>.services")
ModuleNotFoundError: No module named 'prowler.providers.<external>'
```

Two earlier call sites in this PR used `find_spec` directly on `prowler.providers.{provider}.services...`:

- `recover_checks_from_provider` in `prowler/lib/check/utils.py`
- `_resolve_check_module` in `prowler/lib/check/check.py`

For external providers, the propagated `ModuleNotFoundError` was caught by the broad `except Exception` in `recover_checks_from_provider` → `logger.critical(...)` + `sys.exit(1)`. Entry points were never consulted, so plug-ins that registered checks correctly via `prowler.checks.{provider}` were unreachable.

A third call site, `Scan.scan` in `prowler/lib/scan/scan.py`, was still using a hard-coded `f"prowler.providers.{type}.services..."` + `import_check(...)` and bypassed the unified resolver entirely.

#### Solution

The canonical guard for "is this provider shipped with the SDK?" is `Provider.is_builtin`, which already wraps `find_spec` in `try/except (ImportError, ValueError)`. Two changes:

1. **Extract the predicate to a leaf module** — `prowler/providers/common/builtin.py::is_builtin_provider`, depending only on `importlib.util`. `Provider.is_builtin` delegates to it. This is the same pattern applied earlier in the PR for `is_tool_wrapper_provider`, and avoids the import cycle (`prowler.lib.check.* → prowler.providers.common.provider → prowler.config.config → prowler.lib.check.*`) that CodeQL flagged.
2. **Gate the built-in branch on `is_builtin_provider(provider)`** at all three call sites, so external providers fall through to the entry-point lookup as intended:
   - `prowler/lib/check/utils.py::recover_checks_from_provider`
   - `prowler/lib/check/check.py::_resolve_check_module`
   - `prowler/lib/scan/scan.py::Scan.scan` — also rewired to delegate to `_resolve_check_module(...)` instead of the hard-coded path, restoring the single source of truth for resolving check modules.

`tests/lib/scan/scan_test.py` updated to patch `_resolve_check_module` instead of `import_check` to follow the new code path.

#### Why this matters

`is_builtin` already existed in this PR; the bug was simply that the call sites duplicated `find_spec` logic instead of consuming it. The fix removes that duplication and makes the entry-point path actually reachable for installed external providers.

prowler/lib/check/models.py
```diff
@@ -313,8 +317,8 @@ def valid_check_id(cls, check_id, values):  # noqa: F841
         return check_id
 
     @validator("CheckTitle", pre=True, always=True)
-    def validate_check_title(cls, check_title, values):  # noqa: F841
-        if values.get("Provider") not in EXTERNAL_TOOL_PROVIDERS:
+    def validate_check_title(cls, check_title, values):
```
HugoPBrito: CodeQL is flagging five of the new `CheckMetadata` validators for having `cls` as the first parameter without `@classmethod` (alerts #6240–#6244 on lines 320, 333, 341, 380, 389). Pydantic v1 happily accepts stacking `@classmethod` underneath `@validator(...)`, so we can silence the alerts without changing any behavior.

Would something like this work?

```python
@validator("CheckTitle", pre=True, always=True)
@classmethod
def validate_check_title(cls, check_title, values):
    ...
```

Same shape applies to `validate_related_url`, `validate_recommendation_url`, `validate_description`, and `validate_risk`. Open to another approach if you prefer (renaming to `_cls`, etc.) — this one just felt closest to the existing convention.

HugoPBrito (commented):
Thanks for the back-and-forth on this — really appreciate the iterations. No blockers from my side for shipping the functionality. That said, a handful of small details worth tightening at some point, mostly around stability, consistency, and the experience for plug-in authors picking up the new contract. Have a look through the comments and let me know what you think.

Two cross-cutting observations that don't anchor cleanly to the diff:

### run_fixer still hand-builds the built-in path (prowler/lib/check/check.py:452)

`run_fixer` constructs `prowler.providers.{provider}.services.{service}.{check}.{check}_fixer` directly instead of going through `_resolve_check_module`. For external plug-ins this always raises `ModuleNotFoundError` and falls into the silent `"Fixer method not implemented for check {check}"` log.

Definitely not a priority for this PR — fixers aren't part of the new dynamic contract (no `get_fixer_module` among the 12 methods) and built-ins keep working untouched. Just leaving the thread here so it doesn't get lost. Two reasonable directions whenever:

- A follow-up that adds `get_fixer_module` to the contract and routes through `_resolve_check_module`, or
- A short comment in `run_fixer` calling out that fixers are built-in only for now.

Whichever feels lighter — happy to defer entirely if you'd rather punt to a separate ticket.

### except ImageBaseException catches more than its name suggests (prowler/__main__.py:454)

Now that the `else` under `Provider.is_tool_wrapper_provider(provider)` covers `iac`, `image`, AND any external tool-wrapper that isn't `llm`, the `except ImageBaseException` here is catching more than its name suggests. The outer `except Exception` higher up still backstops anything else, so nothing actually breaks — just that the name will confuse a plug-in author reading this file.

Would either of these feel cleaner?

- Move the `ImageBaseException` catch inside an explicit `image` branch so the name matches what's caught.
- Or broaden it to `except Exception as error` here to reflect the actual scope.

No strong preference, both work. Flagging because the contract feels off, not the behavior.

prowler/config/config.py
```diff
@@ -114,6 +141,18 @@ def get_available_compliance_frameworks(provider=None):
...
+    # Multi-provider support for external plug-ins is tracked in PROWLER-1444.
+    ep_dirs = _get_ep_compliance_dirs()
+    for prov, path in ep_dirs.items():
+        if provider and prov != provider:
+            continue
+        if os.path.isdir(path):
+            for file in os.scandir(path):
+                if file.is_file() and file.name.endswith(".json"):
+                    available_compliance_frameworks.append(
+                        file.name.removesuffix(".json")
```
HugoPBrito: Small inconsistency: this entry-point loop appends compliance framework names without dedup, while two lines above (universal frameworks) we do dedupe, and `Compliance.get_bulk` dedupes too. Listing-only impact — the loader still respects "built-in wins" — but `--list-compliance` and `--list-compliance-requirements` would print the same name twice on a collision.

Would this be cleaner?

```python
name = file.name.removesuffix(".json")
if name not in available_compliance_frameworks:
    available_compliance_frameworks.append(name)
```

Could also be a small follow-up if you'd rather keep this PR focused — your call.

prowler/providers/common/provider.py
```diff
@@ -415,17 +581,96 @@ def init_global_provider(arguments: Namespace) -> None:
...
+            if ep.name == name:
+                try:
+                    cls = ep.load()
+                    Provider._ep_providers[name] = cls
+                    return cls
+                except Exception as error:
+                    logger.warning(
+                        f"{error.__class__.__name__}[{error.__traceback__.tb_lineno}]: {error}"
+                    )
+        return None
```
HugoPBrito, line 694: Noticed `_load_ep_provider` doesn't cache misses: every unknown-name lookup re-iterates `entry_points(group="prowler.providers")`. Meanwhile `tool_wrapper._ep_class_cache` does cache `None` for the same kind of lookup, so the two caches end up with different invariants.

Would it be worth aligning them? Something like:

```python
Provider._ep_providers[name] = None
return None
```

at the end of the function — the existing `if name in Provider._ep_providers` already returns the cached `None` correctly. Microperf, so happy to defer if you'd rather not touch it now.

prowler/lib/check/models.py
StylusFrost: Done in [0203888](https://github.com/prowler-cloud/prowler/pull/10700/commits/020388824e1105a7622bc2a9ac2e6948b758612a): stacked `@classmethod` under `@validator` for `validate_check_title`, `validate_related_url`, `validate_recommendation_url`, `validate_description`, and `validate_risk`. Pydantic v1 was already treating them as classmethods via signature introspection, so runtime behavior is unchanged — alerts #6240–#6244 should clear on the next CodeQL run.

One small note: vulture (which the project runs at `--min-confidence 100`) still flags `cls` as unused in those signatures. To stay consistent with the rest of the file (`valid_category`, `validate_service_name`, `valid_check_id`, `validate_check_type`, `validate_resource_group`, `validate_additional_urls` all use `# noqa: F841`), I added the same suppression to the five new ones. Happy to switch to the `_cls` approach you mentioned if you prefer that direction — would need a follow-up to remove the existing `# noqa` markers and the TODO at L250 to keep the file consistent.

prowler/config/config.py
StylusFrost: Done in https://github.com/prowler-cloud/prowler/pull/10700/commits/e5b9fee942259c462f0bd654dea41b73db03ea80: applied the same dedup pattern as the universal frameworks loop a few lines above. The listing function now matches the loader (`Compliance.get_bulk`) and the rest of the PR's built-in-wins policy across providers and checks (`c7aa53689`).

Kept the dedup silent for symmetry with the loader. Surfacing override events to plugin authors uniformly across providers, checks, and compliance is a separate follow-up — captured as a Jira task linked to the parent epic.

Included a regression test in `tests/config/config_test.py` that registers a fake entry point colliding with `cis_2.0_aws` and asserts the framework appears exactly once in the resulting list.

prowler/providers/common/provider.py
StylusFrost: Done in [4fb14bb](https://github.com/prowler-cloud/prowler/pull/10700/commits/4fb14bbb21cc595fb19d308570889135a8a0d044): cached the miss before `return None` so repeated lookups for unknown names short-circuit on the existing `if name in Provider._ep_providers` guard. `Provider._ep_providers` now matches `tool_wrapper._ep_class_cache` on the miss invariant.

Also underscore-prefixed the remaining unused parameters on the abstract `Provider` stubs (`get_output_options`, `get_stdout_detail`, `generate_compliance_output`, `display_compliance_table`) since vulture flagged them once the file landed in the diff — same incidental cleanup pattern from `bbe3a7dbf`.

Included a regression test in `tests/providers/external/test_dynamic_provider_loading.py` that mocks `importlib.metadata.entry_points`, calls `_load_ep_provider("nonexistent")` twice, and asserts `entry_points` is invoked exactly once.

StylusFrost:
Closing the loop on the two cross-cutting observations from your [review](https://github.com/prowler-cloud/prowler/pull/10700#pullrequestreview-4237079875):

**`except ImageBaseException` scoping** → Done in [b13baa9](https://github.com/prowler-cloud/prowler/pull/10700/commits/b13baa90768344b214ba5b25a4aeea44c5e8c115)`. Moved the catch into an explicit `if provider == "image":` branch so the name matches what it catches. iac and external tool-wrapper providers fall through to the outer `except Exception` backstop, which was already the effective behavior since they don't raise `ImageBaseException`. No behavior change.

**`run_fixer` hard-coding the built-in path** → Tracked as [PROWLER-1520](https://prowlerpro.atlassian.net/browse/PROWLER-1520) under parent epic PROWLER-1390. Plan follows your suggestion: add `get_fixer_module` to the Provider contract and route fixer resolution through `_resolve_check_module` rather than hand-building the path. Keeps consistency with the 12-method contract pattern this PR established for the rest of the provider surface. Built-ins continue to work unchanged via a default implementation that derives the fixer path from the resolved check module.

[PROWLER-1520]: https://prowlerpro.atlassian.net/browse/PROWLER-1520?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

HugoPBrito (dismissed):
Thanks for the quick turnaround on the four follow-ups!

Approving on my side.

Alan-TheGentleman (changes requested):
I'm requesting changes for two contract/security issues before merge:

1. Shadowed built-in provider entry points are still loaded via `ep.load()` just to emit the “ignored” warning. That means a plug-in named `aws`, `gcp`, etc. can execute during a built-in provider run, even though built-in precedence is intended.
2. `Compliance.get_bulk()` still uses substring provider matching. That can break external compliance loading for provider names that overlap built-ins, for example `cloud` matching `cloudflare`.

Please fix those before merge. Also, `codecov/project` is failing, so this should be updated from current `master` and rerun before final review.

prowler/lib/check/tool_wrapper.py
```diff
@@ -0,0 +1,57 @@
...
+    Caches the result in `_ep_class_cache`. Errors during entry-point loading
+    are swallowed (returning None) so a broken plug-in never crashes the
+    is-tool-wrapper check; it just falls through to "not a tool wrapper".
+    """
+    if provider in _ep_class_cache:
+        return _ep_class_cache[provider]
+    for ep in importlib.metadata.entry_points(group="prowler.providers"):
+        if ep.name == provider:
+            try:
+                cls = ep.load()
```
josema-xyz, line 37: Same `ep.load()` issue @Alan-TheGentleman raised, but a different spot from the warning he pointed at. This one runs during metadata validation, on every check. A plugin sharing a built-in's name gets its code run here too. Worth fixing alongside the warning one.

HugoPBrito (changes requested):
One more place where the built-in guard should be applied is `is_tool_wrapper_provider()`.

This is not about overriding the built-in provider class directly. The issue is classification: for any provider not in `EXTERNAL_TOOL_PROVIDERS`, the helper currently looks up `prowler.providers` entry points by name and calls `ep.load()`. So a same-name plug-in registered as `aws` with `is_external_tool_provider = True` can make `Provider.is_tool_wrapper_provider("aws")` return `True`.

That can change the built-in AWS execution path before provider initialization, for example by skipping compliance/check loading paths that are gated behind `not Provider.is_tool_wrapper_provider(provider)`. Built-ins should short-circuit here too:

```python
if provider in EXTERNAL_TOOL_PROVIDERS:
    return True
if is_builtin_provider(provider):
    return False
# only then inspect external entry points
```

This keeps the tool-wrapper contract separate from provider override semantics, while still preventing same-name plug-ins from affecting built-in behavior.

prowler/lib/check/tool_wrapper.py
StylusFrost: Fixed in [`5070ce39c`](https://github.com/prowler-cloud/prowler/commit/5070ce39c).

`is_tool_wrapper_provider` now short circuits on `is_builtin_provider(provider)` before it reaches `_load_ep_class`, so any plugin registered under a builtin's name is classified without its module being imported. That closes this spot during metadata validation, which runs on every check, exactly as you flagged. `_load_ep_class` has a single caller, so there is no other path to that `ep.load()` for a builtin name.

Loading entry point code stays reserved for genuine external providers whose name does not collide with a builtin, where importing the class to read `is_external_tool_provider` is intended.

Added `test_builtin_name_shortcircuits_before_loading_same_name_plugin`: registers a plugin named `aws` with `is_external_tool_provider = True` and asserts the classification stays `False` and `ep.load` is never called.

StylusFrost:
@HugoPBrito re: applying the built-in guard in `is_tool_wrapper_provider`

Fixed in [`5070ce39c`](https://github.com/prowler-cloud/prowler/commit/5070ce39c).

`is_tool_wrapper_provider` now short-circuits on `is_builtin_provider(provider)` before consulting any entry point, exactly as you laid out:

```python
if provider in EXTERNAL_TOOL_PROVIDERS:
    return True
if is_builtin_provider(provider):
    return False
cls = _load_ep_class(provider)
return bool(cls and getattr(cls, "is_external_tool_provider", False))
```

A same-name plug-in registered as `aws` with `is_external_tool_provider = True` can no longer make `is_tool_wrapper_provider("aws")` return `True`, so the built-in execution path stays intact. The guard delegates to the leaf `prowler.providers.common.builtin.is_builtin_provider`, so no import cycle is introduced.

Regression test `test_builtin_name_shortcircuits_before_loading_same_name_plugin` asserts the classification stays `False` and the shadowing plug-in is never loaded.

StylusFrost:
@Alan-TheGentleman re: the two contract/security issues

Both addressed.

**1. Shadowed built-in entry points loaded via `ep.load()`** → Fixed in [`64e82682b`](https://github.com/prowler-cloud/prowler/commit/64e82682b). The "ignored" warning in `init_global_provider` now detects a shadowing plug-in by entry-point name only, instead of calling `_load_ep_provider` (which would `ep.load()` the module):

```python
if any(
    ep.name == arguments.provider
    for ep in importlib.metadata.entry_points(group="prowler.providers")
):
    logger.warning(...)
```

A plug-in named `aws`/`gcp`/etc. no longer executes during a built-in run. The same guard covers the classification path in `is_tool_wrapper_provider` (commit [`5070ce39c`](https://github.com/prowler-cloud/prowler/commit/5070ce39c), per @HugoPBrito's comment). The regression test asserts `ep.load` is never called.

**2. `Compliance.get_bulk()` substring matching** → Fixed in [`9c7afd64c`](https://github.com/prowler-cloud/prowler/commit/9c7afd64c). Switched `provider in compliance_framework.name` to an exact match on the last dotted segment:

```python
if compliance_framework.name.split(".")[-1] == provider:
```

Verified against all 16 real compliance modules: no change for any built-in. The bug was broader than `cloud`/`cloudflare` - an external provider named `cloud` matched `alibabacloud`, `cloudflare` and `oraclecloud` at once. Parametrized regression test covers `cloud`, `git`, `work` and `open`.

josema-xyz (dismissed):

prowler/__main__.py
```diff
@@ -436,6 +436,9 @@ def prowler():
         output_options = ScalewayOutputOptions(
             args, bulk_checks_metadata, global_provider.identity
         )
+    else:
+        # Dynamic fallback: any external/custom provider
+        output_options = global_provider.get_output_options(args, bulk_checks_metadata)
```
HugoPBrito: Reproduced by running an external provider whose `get_output_options` is not implemented: this line raises an **uncaught `NotImplementedError` and aborts the entire run** (no try/except here, unlike the HTML/compliance fallbacks). The PR description says unimplemented contract methods "fall back to sensible defaults", but this one has no default. Could we either document `get_output_options` as **required**, or wrap it with a graceful error/fallback? A minimal provider fully crashes today.

prowler/lib/outputs/summary_table.py
```diff
@@ -121,6 +121,9 @@ def display_summary_table(
         elif provider.type == "scaleway":
             entity_type = "Organization"
             audited_entities = provider.identity.organization_id
+        else:
+            # Dynamic fallback: any external/custom provider
+            entity_type, audited_entities = provider.get_summary_entity()
```
HugoPBrito, line 126: Tested with an external provider missing `get_summary_entity`: it raises `NotImplementedError`, which `display_summary_table` catches and logs as `CRITICAL`, **silently dropping the summary table**. Same "no sensible default" gap. A real default (e.g. `(self.type, getattr(self.identity, "account_id", ""))`) or marking it required would avoid the silent loss.

prowler/__main__.py
```diff
@@ -1293,6 +1303,30 @@ def streaming_callback(findings_batch):
...
+        # Dynamic fallback: any external/custom provider
+        try:
+            global_provider.generate_compliance_output(
+                finding_outputs,
+                bulk_compliance_frameworks,
+                input_compliance_frameworks,
+                output_options,
+                generated_outputs,
+            )
+        except NotImplementedError:
```
HugoPBrito, line 1327: Confirmed by running `--compliance` with a CIS-schema framework on a provider that does **not** override `generate_compliance_output`: this `except` only catches `NotImplementedError`, but the last-resort `GenericCompliance.transform` reads `attribute.SubGroup/.Service/.Type/.Comment` directly and raises `AttributeError: 'CIS_Requirement_Attribute' object has no attribute 'SubGroup'` -> no CSV written. Both example providers hide this with a `_SchemaTolerantCompliance` override that uses `getattr`. Suggestion: lift that defensive `getattr` into core `GenericCompliance` (or also catch `AttributeError` here with an actionable message), so the documented "last resort" holds for provider-specific schemas.

HugoPBrito (changes requested):
Functionally tested this branch in an isolated venv (PR head): built-in no-regression, external provider E2E (`local-template`), external checks on `aws`, built-in-name precedence, and the PR test suite (235 passed). Two confirmed issues below, both reproduced by running prowler — the rest of the contract works as expected.

prowler/__main__.py
StylusFrost: Fixed in [efa3283a2](https://github.com/prowler-cloud/prowler/commit/efa3283a2).

`Provider.get_output_options` now returns a generic `ProviderOutputOptions` instead of raising, so an external provider that doesn't override it still produces output. The filename honors `arguments.output_filename` and otherwise falls back to `prowler-output-{provider.type}-{timestamp}`. Built-ins are unaffected (they never reach the `else` branch).

Added regression tests for both the default and the explicit-filename path.

prowler/lib/outputs/summary_table.py
StylusFrost: Fixed in [356e6e2bb](https://github.com/prowler-cloud/prowler/commit/356e6e2bb).

`Provider.get_summary_entity` now returns `(self.type, getattr(self.identity, "account_id", ""))` instead of raising, so an external provider that doesn't override it keeps its summary table instead of having it silently dropped. Used the exact default you suggested. Built-ins are unaffected (they never reach the `else` branch).

Added regression tests for both the account_id case and the empty-string fallback.

prowler/__main__.py
StylusFrost: Update: moved the default to the call site ([29825f9a2](https://github.com/prowler-cloud/prowler/commit/29825f9a2)) to avoid a provider <-> models import cycle CodeQL flagged. `get_output_options` stays override-only and `__main__` falls back to a `default_output_options` helper when a provider doesn't implement it. Same behavior, no cycle.

prowler/__main__.py
StylusFrost: Fixed in [f9682c135](https://github.com/prowler-cloud/prowler/commit/f9682c135).

Lifted the defensive read into core `GenericCompliance.transform`: all six universal attribute fields (Section, SubSection, SubGroup, Service, Type, Comment) are now read with `getattr(attribute, ..., None)`, so the documented last-resort renderer no longer raises `AttributeError` on a provider-specific schema and the CSV is written.

Applied it to all six rather than only the four CIS lacks, because ENS and ISO27001 also omit `Section`, and five schemas omit `SubSection`. Also deduped the finding and manual rows into a single helper. Output for schemas that do declare the fields is unchanged (the exact-CSV test still passes).

Added a regression test that runs the transform over an ISO27001 attribute schema (declares none of the six) and asserts the missing fields render as None instead of aborting.

Follow-up in the template repo: with core now schema-tolerant, the example providers' `_SchemaTolerantCompliance` override can be dropped.

pedrooot (approved):
Good job! 👏🏼

Alan-TheGentleman (approved):

danibarranqueroo (approved):

josema-xyz (approved):

PR #10704, opened 2026-04-15
fix(ui): improve attack paths scan table UX and fix info banner variant

Alan-TheGentleman:
### Context

Follow-up to #10685. The attack paths scan selection table had several UX issues: the "In Progress" badge used warning colors (ugly brown/orange in dark mode), the info banner used warning variant instead of info, the status badge mixed graph availability concerns, and the selection affordance (buttons) was heavy.

### Description

**Scan table redesign:**
- Replace action buttons with radio buttons for scan selection (first column) — semantically correct for single-select
- Entire row is clickable to select a scan (via `onRowClick`)
- Add dedicated "Graph" column with check/minus icons
- Remove redundant "Progress" column (already shown in status badge)

**Status badge simplification:**
- Remove `graphDataReady` prop and tooltip logic — graph status is now its own column
- Change "In Progress" from warning (orange) to info (blue) colors
- Remove unused `showGraphDot` config

**Info banner fix:**
- Change "Viewing data from a previous scan" from `variant="default"` with warning classes to `variant="info"`
- Replace `TriangleAlert` icon with `Info` icon

**Other fixes:**
- Fix Navbar Suspense missing `key` prop (React console warning)
- Replace magic strings with `SCAN_STATES` constants
- Remove redundant `dark:text-text-neutral-secondary` classes
- Add `.opencode/` to `.gitignore`
- Fix duplicate `### Changed` section in CHANGELOG

### Steps to review

1. Go to Attack Paths → verify scan table shows radio buttons in first column, Graph column with check/minus icons
2. Click any row (not just the radio) to select a scan → should work if graph data is available
3. Verify disabled radios (no graph data) look visually distinct from enabled ones
4. Select a running/failed scan with graph data → verify info banner is blue, not orange
5. Check "In Progress" badge is blue, not brown/orange
6. Open browser console → verify no React key warning from Navbar

### Checklist

- [x] Review if code is being documented

#### UI (if applicable)

### License


pfe-nazaries (dismissed):
LGTM

alejandrobailo (changes requested):
Thanks for the cleanup, the radio button lives in the right place (`ui/components/shadcn/radio-group/radio-group.tsx`, consistent with the shadcn/HeroUI split in `ui/CLAUDE.md`) and the simplifications to `ScanStatusBadge` and the info banner are clearly better. A few changes I'd like before merging:

### Radio group usage
- **Use `onValueChange` on `<RadioGroup>` instead of `onClick` on each `<RadioGroupItem>`** (`scan-list-table.tsx:116-120`). That's the idiomatic Radix API and unifies pointer + keyboard activation (space/arrow keys with roving tabindex). Something like:
  ```tsx
  <RadioGroup value={selectedScanId ?? ""} onValueChange={handleSelectScan}>
    ...
    <RadioGroupItem value={row.original.id} disabled={!canSelect} />
  ```
  Also removes the manual `canSelect && !isSelected` guard inside the item.
- **Avoid double-dispatch between radio and row click.** Clicking the radio bubbles up to `onRowClick` and fires `handleSelectScan` twice. It's idempotent today but conceptually messy — either `stopPropagation` on the radio or rely solely on `onValueChange` and let `onRowClick` cover the rest of the row.
- **Flatten the `className` ternary** (`scan-list-table.tsx:121-128`):
  ```tsx
  className={cn(
    "size-5",
    canSelect && !isSelected && "border-text-neutral-secondary cursor-pointer",
    !canSelect && "disabled:opacity-70",
  )}
  ```
  `cn()` already drops falsy values.

### Accessibility — Graph column
The `<Check />` / `<Minus />` icons have no accessible text. Please add `aria-label` (e.g. `"Graph available"` / `"Graph not available"`) or a visually hidden label so screen readers can convey the state.

### Tests
`scan-list-table.test.tsx:278, 300` assert the Graph column via `document.querySelector(".lucide-check")` / `.lucide-minus`. Those classes depend on `lucide-react`'s internals and will silently break on a dependency bump. Please switch to `getByTestId` or query by the new accessible label above.

### Disabled tooltip coverage
If `getDisabledTooltip` ever returns `null` for a disabled state, the radio ends up disabled without explanation — inconsistent UX. Either guarantee a tooltip for every disabled path or render a generic fallback.

### Navbar `Suspense key="feeds"`
The commit silences a React warning without explaining the root cause. A missing `key` warning only happens when the element is part of a sibling list — worth checking whether `Navbar` is being rendered inside a `.map` somewhere and fixing that at the source, rather than pinning a static key here.

### Changelog
The new "Attack Paths scan table: replace action buttons…" entry in `### 🔄 Changed` is missing the `[(#10704)]` reference, the surrounding entries include it.

Non-blocking but worth addressing before merge.

alejandrobailo (approved):

PR #10718, opened 2026-04-15
fix(sdk): scan every Azure subscription when display names collide

### LINKED ISSUE #10719: Azure: subscriptions sharing a display name collapse into one; only one is scanned
### Issue search

Prowler CLI/SDK

Azure

### Steps to Reproduce

1. Use an Azure tenant that contains several subscriptions sharing the same `display_name` (for example, auto-generated names under certain billing account types — all subscriptions show the same name in `az account list`). A Service Principal with `Reader` assigned at a management group that covers all of them can see every subscription:
   ```
   az account list --query "[].{name:name, id:id, state:state}" -o table
   Name              State
   ----------------  -------
   <same-name>       Enabled
   <same-name>       Enabled
   <same-name>       Enabled
   <same-name>       Enabled
   ```
2. Run any Azure scan, either letting Prowler auto-discover the subscriptions:
   ```
   prowler azure --sp-env-auth
   ```
   or passing the IDs explicitly:
   ```
   prowler azure --sp-env-auth --subscription-ids <ID1> <ID2> <ID3> <ID4>
   ```
3. Observe Prowler's banner and the resulting CSV.

### Expected behavior

All four subscriptions are scanned, and each one is represented in `identity.subscriptions` and in the output CSV (one row per finding per subscription).

Only a single subscription is scanned regardless of how many were requested. Prowler's banner line lists just one subscription:

```
· Azure Subscriptions: ['<same-name>: <last-subscription-id>']
```

The output CSV contains findings only for that last subscription; the other subscription IDs do not appear anywhere in the file.

**Root cause** (in `prowler/providers/azure/azure_provider.py`, `AzureProvider.setup_identity`):

```python
if not subscription_ids:
    for subscription in subscriptions_client.subscriptions.list():
        identity.subscriptions.update(
            {subscription.display_name: subscription.subscription_id}
        )
else:
    for id in subscription_ids:
        subscription = subscriptions_client.subscriptions.get(subscription_id=id)
        identity.subscriptions.update({subscription.display_name: id})
```

`identity.subscriptions` is a `dict` keyed by `display_name`. Azure allows multiple subscriptions to share the same `display_name`, so each `update()` call overwrites the previous entry and only the last subscription survives. The dictionary is later consumed by every downstream service and by output/reporting code, so only that surviving subscription is scanned.

The argparse definition for `--subscription-ids` (`prowler/providers/azure/lib/arguments/arguments.py`) uses `nargs="+"` and parses the list correctly — the list arrives intact at `setup_identity`. The bug is purely in how the dictionary is built.

From brew (brew install prowler)

### Environment Resource

Workstation

### OS used

macOS

### Prowler version

5.23.0 (also confirmed against the 5.18.3 container image — same code path is unchanged)

### Python version

3.12

### Pip version

Not applicable (installed via `brew`)

### Context

Workaround until this is fixed: invoke Prowler once per subscription, passing a single `--subscription-ids <ID>` each time. With one ID per process no dictionary collision can occur, so every subscription is scanned independently.

Fix proposed in #10718.

ivan-necheporenko:
### Context

Fix #10719

`AzureProvider.setup_identity` stores the scanned subscriptions in `identity.subscriptions`, a `dict` keyed by the subscription's `display_name`. Azure allows several subscriptions to share the same `display_name`, which is the norm for auto-generated names under some billing account types. When that happens, the later `identity.subscriptions.update({display_name: id})` calls silently overwrite the earlier entries, so the dictionary ends up with a single survivor and the downstream services only ever see one subscription.

The bug is present in both branches of the method:

- **Auto-discovery** (no `--subscription-ids` passed): iterates `subscriptions_client.subscriptions.list()` and collapses duplicates.
- **Explicit filter** (`--subscription-ids ID1 ID2 …`): iterates the user-supplied list and collapses duplicates.

The CLI output even reflects this — the banner line `Azure Subscriptions: […]` shows only the last subscription that survived the collapse, and the scan runs against that one alone.

### Description

Build the subscription index with unique keys so every subscription is retained regardless of how Azure named it. When a `display_name` is unique across the resolved subscriptions it is used unchanged (preserving existing behavior). When a name is shared by multiple subscriptions, every colliding key is disambiguated by appending its subscription ID: `"<display_name> (<subscription_id>)"`.

The change is confined to `AzureProvider.setup_identity` — the dict still maps strings to subscription IDs, so every downstream consumer (service clients that look up by `self.subscriptions[subscription_name]`, reporting, and output formatters) continues to work unchanged.

Regression tests cover both branches:
- Unique display names → keys are left as-is.
- Duplicated display names → every subscription is preserved under a disambiguated key.

### Steps to review

1. Read the minimal diff in `prowler/providers/azure/azure_provider.py`: both branches feed a shared `(display_name, subscription_id)` list, then a small counter-based loop writes the final dict.
2. Verify the new tests in `tests/providers/azure/azure_provider_test.py` — four cases covering auto-discovery × explicit filter × unique × duplicated names.
3. Run `pytest tests/providers/azure/azure_provider_test.py` (fast) and, if you want the belt-and-braces check, the full Azure suite `pytest tests/providers/azure/`.

### Checklist

<details>

</details>

#### SDK/CLI
    - If so, do we need to update permissions for the provider? Not applicable.

### License


puchy22 (changes requested):
Hi @ivan-necheporenko, thanks for your contribution. But it is one major issue that concerns me about the fix. 

This solution is invalid and is failing, at the very least, the [`app_function_identity_without_admin_privileges`](https://github.com/prowler-cloud/prowler/blob/master/prowler/providers/azure/services/app/app_function_identity_without_admin_privileges/app_function_identity_without_admin_privileges.py) check, which uses the subscription name extracted from the key of this dict to determine whether a user in that subscription has a privileged role.

The way to resolve the issue you mentioned is not to add the subscription ID at the end of the name, but rather to reverse the keys in the dictionary. In other words, the key should become the ID and the value should be the subscription display name. Once that is done, you must ensure that this new format is used correctly in all places where that attribute is used.

ivan-necheporenko:
@puchy22 Thanks for the review! I've reworked the approach as suggested:

- Reversed `identity.subscriptions` so keys are subscription IDs and values are display names — this eliminates the duplicate name problem entirely since IDs are guaranteed unique, no name-mangling needed.
- Updated all consumers across services, checks, outputs, and tests (103 files) to use the new `{subscription_id: display_name}` format.
- Renamed `subscription_name` loop variables to `subscription_id` and fixed log labels accordingly for consistency.

All unit tests pass (866 Azure service + 217 finding/check tests).

I also ran a full scan (`prowler azure --sp-env-auth`) against our Azure tenant with 4 subscriptions — 166 checks executed successfully with no errors. However, I wasn't able to test the actual duplicate display name scenario end-to-end because we had already renamed our subscriptions to have unique names as a workaround for the original issue. The duplicate-name path is covered by unit tests that assert `{first_id: shared_name, second_id: shared_name}` is produced correctly.

puchy22:
Hi again @ivan-necheporenko . 

Thank you very much for your contribution. Since this involves a large number of files and includes critical changes—such as findings now being identified by the subscription ID instead of the name, and the `status_extended` values for many checks also changing, I need to review this more carefully with the team. If any changes need to be made, we’ll likely handle them ourselves at this stage. I’ll keep you updated on how the fix is progressing.

Thank you very much for your input 🚀

prowler/lib/check/check.py
```diff
@@ -749,9 +749,7 @@ def execute(
                 if global_provider.type == "cloudflare":
                     is_finding_muted_args["account_id"] = finding.account_id
                 if global_provider.type == "azure":
-                    is_finding_muted_args["subscription_id"] = (
-                        global_provider.identity.subscriptions.get(finding.subscription)
-                    )
+                    is_finding_muted_args["subscription_id"] = finding.subscription
```
HugoPBrito, line 752: Origin of the mutelist issue above. Pass both:

```python
if global_provider.type == "azure":
    is_finding_muted_args["subscription_id"] = finding.subscription
    is_finding_muted_args["subscription_name"] = (
        global_provider.identity.subscriptions.get(
            finding.subscription, finding.subscription
        )
    )
```

Open to other shapes.

prowler/providers/azure/services/apim/apim_threat_detection_llm_jacking/apim_threat_detection_llm_jacking.py
```diff
@@ -50,17 +50,20 @@ def execute(self):
             ],
         )
 
-        # 1. Aggregate logs from all APIM instances first
-        all_llm_logs: List[LogsQueryLogEntry] = []
         for subscription, instances in apim_client.instances.items():
+            subscription_name = apim_client.subscriptions.get(
+                subscription, subscription
+            )
+            all_llm_logs: List[LogsQueryLogEntry] = []
```
HugoPBrito, line 57: `all_llm_logs = []` moved inside the per-subscription loop here. Real fix (master accumulates across iterations and misattributes findings on multi-sub scans), but a bit out of scope for this one. Mind splitting into its own PR with a changelog entry?

HugoPBrito (changes requested):
Thanks for working through @puchy22's feedback. Couple of things still worth a look before this lands.

### Mutelist by Subscription Name silently stops matching

`prowler/providers/azure/lib/mutelist/mutelist.py` (not in this diff) has two branches: match by ID, match by `finding.subscription`. After the dict reversal `finding.subscription` is the ID, so both branches collapse onto the same value at `prowler/lib/check/check.py:752` and the "by Name" form goes dead. Docs at `docs/user-guide/cli/tutorials/mutelist.mdx:63` and `docs/user-guide/tutorials/prowler-app-mute-findings.mdx:145` recommend that form, so existing YAMLs would silently stop muting.

CI misses it because `test_is_finding_muted_subscription_name` sets `finding.subscription` by hand. Could `is_finding_muted` take a `subscription_name`, wired from the dispatch (inline)?

prowler/providers/azure/services/apim/apim_threat_detection_llm_jacking/apim_threat_detection_llm_jacking.py
HugoPBrito: Nevermind, lets let it here.

HugoPBrito (approved):

puchy22 (approved):

PR #10724, opened 2026-04-16
fix(api): detect silent failures in ResourceFindingMapping

pedrooot:
## Summary

Fix silent data loss in `ResourceFindingMapping.bulk_create` that leaves findings orphaned (no mapping to their Resource) when the underlying `INSERT ... ON CONFLICT DO NOTHING` drops rows without raising.

## Context

Some from an IAC scan had no `ResourceFindingMapping` rows, despite:
- The Resource existing (created in the same transaction).
- The finding's `resource_services` / `resource_types` / `resource_regions` being correctly populated (written *after* the mapping bulk_create in the same `rls_transaction`).

That combination proves the transaction committed end-to-end without exception, yet `ResourceFindingMapping.objects.bulk_create(...)` inserted zero rows. The scan completed successfully and the data inconsistency went unnoticed.

## Root cause

`api/src/backend/tasks/jobs/scan.py` was calling:

```python
ResourceFindingMapping.objects.bulk_create(
    mappings_to_create,
    batch_size=SCAN_DB_BATCH_SIZE,
    ignore_conflicts=True,
)
```

Three problems compound here:

1. `ignore_conflicts=True` with no `unique_fields` generates a generic `ON CONFLICT DO NOTHING` with no explicit conflict target. On a partitioned parent table (`ResourceFindingMapping` is partitioned by `finding_id`) with a default partition, this can silently drop rows if the unique index isn't fully propagated across partitions.
2. The return value of `bulk_create` (the list of created objects) was discarded, so there was no way to detect partial insertion.
3. No assertion on expected vs. actual inserted row count.

Result: a failure mode in which N rows silently become 0 and the task reports success.

## Fix

- Pass `unique_fields=["tenant_id", "resource_id", "finding_id"]` to `bulk_create` so the `ON CONFLICT` targets the actual unique constraint explicitly.
- Capture the return value and count rows where `pk` was assigned (i.e., actually inserted).
- If `inserted != len(mappings_to_create)`, log the mismatch with the scan id and raise `RuntimeError`. The enclosing `rls_transaction` rolls back the findings `bulk_create` and denormalized array updates, so the scan fails visibly instead of committing inconsistent state.

## Impact

- Scans where mapping insertion silently fails will now fail loudly and be marked `FAILED` instead of `COMPLETED`.
- No change to the happy path: when all mappings insert correctly, behavior is identical.
- Existing orphaned findings from past scans are not touched by this change.

### Steps to review

### Checklist

<details>

</details>

#### SDK/CLI

#### UI

#### API

### License


api/src/backend/tasks/jobs/scan.py
```diff
@@ -752,11 +752,19 @@ def _process_finding_micro_batch(
...
-            ResourceFindingMapping.objects.bulk_create(
+            created_mappings = ResourceFindingMapping.objects.bulk_create(
                 mappings_to_create,
                 batch_size=SCAN_DB_BATCH_SIZE,
                 ignore_conflicts=True,
+                unique_fields=["tenant_id", "resource_id", "finding_id"],
             )
+            inserted = sum(1 for m in created_mappings if m.pk)
+            if inserted != len(mappings_to_create):
+                logger.error(
```
jfagoagas, line 763: How are you rolling back?

jfagoagas (approved):

PR #10729, opened 2026-04-16
perf(attack-paths): cleanup task prioritization, restore default batch sizes to 1000, upgrade Cartography to 0.135.0

josema-xyz:
### Description

- Restore `SYNC_BATCH_SIZE` and `FINDINGS_BATCH_SIZE` defaults from 250/500 back to 1000. The OOM risk that motivated the reduction was fixed by the sequential single-session cursor architecture, not the batch size.
- Upgrade Cartography from 0.132.0 to 0.135.0.
- Enable Celery queue priority strategy and set default task priority, with migration to give the cleanup periodic task higher priority than scans.
- Rewrite `INSERT_FINDING_TEMPLATE` to use direct index lookups on `_AWSResource` instead of traversing all outgoing relationships from the `AWSAccount` node. Removes `$provider_uid` parameter and account `MATCH` since the temp database is single-account.
- Remove `aws_post_ingestion_principals_cleanup` job. It deletes `AWSPrincipal` nodes where `lastupdated` differs from the current `UPDATE_TAG`, but the temp database is ephemeral and all nodes share the same tag, so it always matches zero nodes.
- Add timing logs to scan orchestrator (total scan, ingestion, findings, sync, drop subgraph), per-sync-function completion lines in AWS ingestion, and cumulative timing in node/relationship sync batches.

### Steps to review

Run tests and a local Attack Paths scan.

### Checklist

#### API

### License


api/pyproject.toml
```diff
@@ -38,7 +38,7 @@ dependencies = [
   "matplotlib (==3.10.8)",
   "reportlab (==4.4.10)",
   "neo4j (==6.1.0)",
-  "cartography (==0.132.0)",
+  "cartography (==0.135.0)",
```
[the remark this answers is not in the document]
josema-xyz: Working on that, this is a draft.

api/src/backend/api/migrations/0090_attack_paths_cleanup_priority.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+def unset_cleanup_priority(apps, schema_editor):
+    PeriodicTask = apps.get_model("django_celery_beat", "PeriodicTask")
+    PeriodicTask.objects.filter(name=TASK_NAME).update(priority=None)
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("api", "0089_backfill_finding_group_status_muted"),
```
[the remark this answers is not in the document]
josema-xyz: Well, no, because 89 depends on 88, that one on 87, and finally that on 86. 86 depends on 19 as is the one that created that Celery beat task.

api/src/backend/api/migrations/0090_attack_paths_cleanup_priority.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+def set_cleanup_priority(apps, schema_editor):
+    PeriodicTask = apps.get_model("django_celery_beat", "PeriodicTask")
+    PeriodicTask.objects.filter(name=TASK_NAME).update(priority=0)
+
+
+def unset_cleanup_priority(apps, schema_editor):
+    PeriodicTask = apps.get_model("django_celery_beat", "PeriodicTask")
+    PeriodicTask.objects.filter(name=TASK_NAME).update(priority=None)
```
AdriiiPRodri, line 13: Don't we need to use `MainRouter.admin_db`?

api/src/backend/config/celery.py
```diff
@@ -17,8 +17,10 @@
 celery_app.conf.update(result_extended=True, result_expires=None)
 
 celery_app.conf.broker_transport_options = {
-    "visibility_timeout": BROKER_VISIBILITY_TIMEOUT
+    "visibility_timeout": BROKER_VISIBILITY_TIMEOUT,
+    "queue_order_strategy": "priority",
 }
+celery_app.conf.task_default_priority = 6
```
AdriiiPRodri, line 23: Just to know, why we need this change?

AdriiiPRodri (changes requested):

api/src/backend/config/celery.py
josema-xyz: We have a task that is important to run ASAP, the Attack Paths scans task clenaup. So for not stacking it up on the queue, we gave it a priority, so when a worker look for new tasks, even if there are ton of Attack Paths scans, if there is a cleanup, is going to take it, as it has more priority.

api/src/backend/api/migrations/0090_attack_paths_cleanup_priority.py
josema-xyz: Not needed really, because the router already sends `django_*` tables to `admin_db`. Same as the migration `0086` where we create the task without `.using()`.

AdriiiPRodri (approved):

PR #10736, opened 2026-04-16
fix(ui): correct IaC findings counters

Alan-TheGentleman:
### Context

Fixes PROWLER-1402.

IaC providers can return findings without associated resources, which made the findings group counter show misleading totals when `resources_total` was `0`.

### Description

- add a shared findings-group counter helper for impacted totals
- fall back to `pass_count + fail_count` when no resources exist
- include muted pass/fail counters in the denominator for muted groups
- reuse the same logic in the grouped findings table and drill-down header
- add UI/unit test coverage and a UI changelog entry

### Steps to review

- Open findings grouped view with an IaC provider finding that has no associated resources
- Verify the impacted counter shows `fail_count / (pass_count + fail_count)` when `resources_total = 0`
- Verify muted groups show `fail_count / (pass_count + fail_count + pass_muted_count + fail_muted_count)`
- Confirm standard resource-backed groups still show `resourcesFail / resourcesTotal`
- Automated validation run during commit hooks:
  - `pnpm run typecheck`
  - `pnpm run lint:check`
  - `pnpm run test:run`

### Checklist

<details>

</details>

#### SDK/CLI

#### UI (if applicable)

#### API (if applicable)

### License


AdriiiPRodri (dismissed):

ui/lib/findings-groups.ts
```diff
@@ -23,7 +18,32 @@ type FindingGroupDeltaState = Pick<
...
 
-export function isFindingGroupMuted(group: FindingGroupMutedState): boolean {
+const FINDING_GROUP_DELTA = {
+  CHANGED: "changed",
+  NEW: "new",
+  NONE: "none",
+} as const;
+
+type FindingGroupDelta =
+  (typeof FINDING_GROUP_DELTA)[keyof typeof FINDING_GROUP_DELTA];
```
alejandrobailo: `FINDING_GROUP_DELTA` / `FindingGroupDelta` are exact duplicates of `DeltaValues` / `DeltaType` already exported from [`notification-indicator.tsx`](https://github.com/prowler-cloud/prowler/blob/master/ui/components/findings/table/notification-indicator.tsx#L21-L27):

```typescript
export const DeltaValues = { NEW: "new", CHANGED: "changed", NONE: "none" } as const;
export type DeltaType = (typeof DeltaValues)[keyof typeof DeltaValues];
```

Since `DeltaValues` is already imported by 3+ files (`column-finding-groups`, `findings-group-drill-down`, `column-standalone-findings`), it meets the scope rule for shared code. Could we:

1. Move `DeltaValues` / `DeltaType` to a shared location (e.g. `lib/` or `types/`)
2. Import and reuse it here instead of re-declaring
3. Use `DeltaType` as the return type for `getFindingGroupDelta` / `getFilteredFindingGroupDelta`

This would also eliminate the identity ternary in the consumers — right now `column-finding-groups.tsx` and `findings-group-drill-down.tsx` map `"new"` → `DeltaValues.NEW` (which is `"new"` again). If the functions returned `DeltaType` directly, that mapping becomes unnecessary.

ui/lib/findings-groups.ts
```diff
@@ -23,7 +18,32 @@ type FindingGroupDeltaState = Pick<
...
+
+const FINDING_GROUP_STATUS = {
+  FAIL: "FAIL",
+  MANUAL: "MANUAL",
+  PASS: "PASS",
+} as const;
+
+type FindingGroupStatus =
+  (typeof FINDING_GROUP_STATUS)[keyof typeof FINDING_GROUP_STATUS];
```
alejandrobailo: `FINDING_GROUP_STATUS` / `FindingGroupStatus` overlap significantly with `FindingStatusValues` / `FindingStatus` already exported from [`status-finding-badge.tsx`](https://github.com/prowler-cloud/prowler/blob/master/ui/components/ui/table/status-finding-badge.tsx#L3-L11):

```typescript
export const FindingStatusValues = { FAIL: "FAIL", PASS: "PASS", MANUAL: "MANUAL", MUTED: "MUTED" } as const;
export type FindingStatus = (typeof FindingStatusValues)[keyof typeof FindingStatusValues];
```

`FindingGroupStatus` is the same set minus `MUTED`. Rather than a parallel const map, we could derive it:

```typescript
type FindingGroupStatus = Exclude<FindingStatus, "MUTED">;
```

Same idea as above — `FindingStatusValues` could also move to `lib/` or `types/` if it isn't already shared enough, and `FINDING_GROUP_STATUSES` could be derived from it.

alejandrobailo (changes requested):
Nice work on the core fix — `getFindingGroupImpactedCounts` with the fallback to `passCount + failCount` for IaC providers is clean and well-tested. The three code paths (resources available, no resources, no resources + muted) are all covered.

However, I noticed some type duplication that I think we should address before merging. See the inline comments for details — the gist is that the new const maps already exist elsewhere in the codebase, and consolidating them would reduce surface area and remove some unnecessary mapping code downstream.

ui/lib/findings-groups.ts
Alan-TheGentleman: Fixed in 9efca286a. Grouped findings now derive their status type from the shared finding status definitions instead of keeping a parallel const map.

ui/actions/findings/findings.ts
```diff
@@ -141,12 +141,25 @@
...
+  options?: GetFindingByIdOptions,
+) => {
   const headers = await getAuthHeaders({ contentType: false });
 
   const url = new URL(`${apiBaseUrl}/findings/${findingId}`);
   if (include) url.searchParams.append("include", include);
 
+  if (options?.source === "resource-detail-drawer") {
+    // eslint-disable-next-line no-console
+    console.info(`[resource-detail-drawer] GET ${url.pathname}${url.search}`);
```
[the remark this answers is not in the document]
Alan-TheGentleman: Fixed in 9efca286a. I removed the drawer request logs from the server actions so we are no longer logging user-influenced URL data.

ui/lib/findings-groups.ts
Alan-TheGentleman: Fixed in 9efca286a. I moved grouped findings to the shared delta definitions in `@/types`, added the missing `none` state there, and removed the identity mapping in the consumers.

Alan-TheGentleman:
Pushed the last drawer loading UX pass in bb6afba64.

The drawer now stays structurally stable while moving through carousel resources. We keep the shell, table headers, and scan labels visible, and only skeletonize the values/rows that actually change. This avoids the full-drawer reload feel and reduces layout shift between items.

alejandrobailo (changes requested):
The `changelog` must belong to the 24.1

alejandrobailo (approved):

PR #10757, opened 2026-04-16
fix(ui): improve Resource Inventory cards light mode

HugoPBrito:
### Context

The Resource Inventory cards in the Overview were hard to read in light mode (low-contrast backgrounds inherited from a dark-only palette) and the card heights were not aligned with the loading skeleton, producing layout shifts when data arrived.

### Description

- Refactor `ResourcesInventoryCardItem` to use neutral surfaces (`bg-bg-neutral-secondary` / `border-border-neutral-secondary`) with a colored top accent bar (fail/pass) so cards render consistently in both light and dark modes.
- Apply unified hover/focus states (lift + shadow + border-color transition) and a focus ring on the wrapping `<Link>` for keyboard navigation.
- Format `resourceCount` as a localized `"N Resources"` string and use `ShieldCheck` (pass) / `TriangleAlert` (fail) badges based on `failedFindings`.
- Align `ResourcesInventorySkeleton` with the new card surface and add a top accent skeleton so the loading state matches the final layout (no shift on data arrival).
- Add unit tests for `ResourcesInventoryCardItem` covering link/filter behavior, neutral container styling, and the empty state.

### Steps to review

1. Run the new tests:
   ```
   cd ui && npm run test -- resources-inventory-card-item
   ```
2. Start the UI (`npm run dev` in `ui/`) and open Overview → Resource Inventory:
   - Toggle to **light mode**: cards should use neutral surfaces, top accent bar visible (red when there are failed findings, green otherwise), hover/focus lifts the card, focus ring appears with keyboard `Tab`.
   - Toggle to **dark mode**: no regressions in surfaces, accent bar, or hover/focus.
   - Trigger the loading state (slow network / refresh): skeleton matches the final card surface and accent strip — no layout shift on hydration.
   - Cards with `totalResources === 0` render the empty state without a link.

### Checklist

<details>

</details>

#### UI
- [ ] Screenshots/Video of the functionality flow (if applicable) - Tablet (640px > X < 1024px)

### License


ui/CHANGELOG.md
```diff
@@ -2,12 +2,21 @@
 
 All notable changes to the **Prowler UI** are documented in this file.
 
+## [1.25.0] (Prowler UNRELEASED)
+
+### 🐞 Fixed
+
+- Resource Inventory cards in light mode now use neutral surfaces, consistent hover/focus states, and aligned loading placeholders [(#00000)](https://github.com/prowler-cloud/prowler/pull/00000)
```
pfe-nazaries: update with the PR number

alejandrobailo (changes requested):
Nice work on catching the light-mode contrast issue, Hugo — the bug is real, light mode on these cards did look washed out. I had a closer look at the approach and I'd like to suggest a different path before merging. I'll go ahead and push the change myself so you don't have to juggle it.

### The root cause lives in the design system

The low-contrast look in light mode isn't coming from `ResourcesInventoryCardItem` — it's coming from `ui/components/shadcn/card/resource-stats-card/resource-stats-card.tsx` (lines 17–27), where `CardVariant.fail | pass | warning | info` are hardcoded with literal `rgba()` values pulled from the Figma dark palette:

```tsx
[CardVariant.fail]:
  "border-[rgba(67,34,50,0.5)] bg-[rgba(67,34,50,0.2)] dark:border-[rgba(67,34,50,0.7)] dark:bg-[rgba(67,34,50,0.3)]",
```

Those colors were picked to read well on a dark background, so in light mode they render as low-opacity burgundy/forest tones on white — hence the muddy look. That's the thing to fix.

### Why patching it in the feature is risky

In this PR the card is opted out of the design system (`variant={CardVariant.default}` + a locally-built `before:bg-bg-fail-primary` strip via `cn()`), which does solve the visual for this screen, but it leaves the shared `ResourceStatsCard` untouched. So anywhere else in the app that renders `<ResourceStatsCard variant={CardVariant.fail} />` keeps the old washed-out look, and we end up with two "fail" card styles living side by side. The accent-bar pattern is nice — it should live inside the shared component so every future consumer inherits it for free.

### What I'll do

I'll take this over and:

1. Add an `accent?: CardVariant` prop to `ResourceStatsCard` that renders a neutral surface + a colored top accent bar using design tokens (`bg-bg-fail-primary`, `bg-bg-pass-primary`, …), no hardcoded hex/rgba.
2. Rewrite `ResourcesInventoryCardItem` to use that new prop instead of inlined `cn()` classes, so the feature file goes back to being declarative.
3. Loosen the unit test so it asserts behavior (link href, empty state, accent presence) rather than exact Tailwind class names, which will keep it resilient across theme refactors.
4. Keep your skeleton + focus-ring changes — those are genuine wins and I want them in.

### Smaller notes while I'm in there

- `resourceCount: \`${item.totalResources.toLocaleString()} Resources\`` hardcodes the English label. Low priority, but worth parking a TODO so it picks up whatever i18n strategy lands.
- `rounded-[12px]` on the `<Link>` duplicates the Card's internal radius — I'll switch it to `rounded-xl` so it tracks the token.
- The badge count still shows `failedFindings` (i.e. `0`) next to the `ShieldCheck` for healthy groups. Worth a quick design check on whether we want a "0" there or a different label.
- Thanks for the `focus-visible` ring on the `<Link>` — genuine a11y win.

Again, really good catch on the light-mode issue. I'll have the follow-up pushed shortly. 🙌

alejandrobailo (approved):

PR #10777, opened 2026-04-19
Review Azure new checks for ISMS-P alignment

Shin-723:
## Summary
This PR opens the review branch for Azure new checks and includes two follow-up fixes found during an ISMS-P-oriented review.

## What changed
- tightened `entra_user_with_recent_sign_in` to avoid false positives when sign-in telemetry is unavailable for all enabled users in a tenant
- updated `recovery_vault_backup_policy_retention_adequate` to report a `FAIL` when a Recovery Vault has no backup policies configured
- added/updated tests for both behaviors

## Why
During review of the new Azure checks, two areas looked too optimistic from an ISMS-P / control-validation perspective:

1. Entra stale-user check
`last_sign_in = None` does not always mean "the user has never signed in".
If sign-in activity is unavailable tenant-wide, the safer interpretation is a telemetry/licensing/permission gap rather than a user-level inactivity finding.

2. Recovery backup policy check
A vault with no backup policies configured should not be skipped silently.
From a control assurance perspective, this is a meaningful failure condition and should surface explicitly.

## Validation
Validated in a local Python 3.12 test environment with targeted tests:

- `tests/providers/azure/services/entra/entra_user_with_recent_sign_in/entra_user_with_recent_sign_in_test.py`
- `tests/providers/azure/services/recovery/recovery_vault_backup_policy_retention_adequate/recovery_vault_backup_policy_retention_adequate_test.py`

## Review focus
Please review whether these two adjustments better reflect conservative control interpretation for Azure:

- `entra_user_with_recent_sign_in`
- `recovery_vault_backup_policy_retention_adequate`

### DIFF
--- a/prowler/compliance/azure/cis_2.0_azure.json
+++ b/prowler/compliance/azure/cis_2.0_azure.json
@@ -3303,7 +3303,9 @@
     {
       "Id": "9.11",
       "Description": "Ensure Azure Key Vaults are Used to Store Secrets",
-      "Checks": [],
+      "Checks": [
+        "entra_app_registration_credential_not_expired"
+      ],
       "Attributes": [
         {
           "Section": "9 AppService",
--- a/prowler/compliance/azure/cis_2.0_azure.json
+++ b/prowler/compliance/azure/cis_2.0_azure.json
@@ -248,7 +248,9 @@
     {
       "Id": "1.4",
       "Description": "Ensure Access Review is Set Up for External Users in Azure AD Privileged Identity Management",
-      "Checks": [],
+      "Checks": [
+        "entra_user_with_recent_sign_in"
+      ],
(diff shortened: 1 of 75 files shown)


jfagoagas:
@Shin-723 Thanks for this contribution 🙌

In order for us to be able to review the PR and merge it as fast as possible we need you to split it creating one PR by check. Please, let us know if you need help for that and the team will work with you.

danibarranqueroo:
Hello @Shin-723! Thanks for this, after reviewing the PR I've seen that all of this checks were the same as the ones done here https://github.com/prowler-cloud/prowler/pull/10809. Since all of the splitted checks from that PR are already released I can close this.

Thanks again for your contribution! 🙌

Issue #9724, opened 2026-01-06
SAML ACS URL uses internal hostname prowler-api:8080 instead of external domain


jobigaila:
### Issue search

- [x] I have searched the existing issues and this bug has not been reported yet

### Which component is affected?

Prowler UI

### Cloud Provider (if applicable)

Not applicable

## Issue Description
  When configuring SAML SSO for Prowler behind an nginx reverse proxy, the ACS URL is generated using the
  internal hostname `http://prowler-api:8080/api/v1/accounts/saml/mydomain.lt/acs/` instead of the external domain
  URL.

  ## Environment
  - Prowler version: v5.14.0 (docker)
  - Deployment: Docker Compose with nginx reverse proxy
  - External URL: https://prw.mydomain.lt:8445
  - IdP: Azure AD

 ## Steps to Reproduce
  1. Deploy Prowler using docker-compose behind nginx reverse proxy
  2. Set environment variables:
     - `API_BASE_URL=https://prw.mydomain.lt:8445/api/v1`
     - `AUTH_URL=https://prw.mydomain.lt:8445`
     - `NEXT_PUBLIC_API_BASE_URL=https://prw.mydomain.lt:8445/api/v1`
     - `DJANGO_ALLOWED_HOSTS=...,prw.mydomain.lt,...`
  3. Create SAML configuration for domain (e.g., `mydomain.lt`)
  4. ACS URL is generated as: `http://prowler-api:8080/api/v1/accounts/saml/mydomain.lt/acs/`
  5. Expected ACS URL: `https://prw.mydomain.lt:8445/api/v1/accounts/saml/mydomain.lt/acs/`

  ## Workarounds Attempted
  - Set `API_BASE_URL`, `AUTH_URL`, `NEXT_PUBLIC_API_BASE_URL` to external URLs
  - Added external domain to `DJANGO_ALLOWED_HOSTS`
  - Restarted all containers
  - ACS URL still uses internal hostname

### Expected behavior

Expected ACS URL: `https://prw.mydomain.lt:8445/api/v1/accounts/saml/mydomain.lt/acs/

### Actual Result with Screenshots or Logs

ACS URL still uses internal hostname. when trying to login with SAML SSO

Sign in
Sorry, but we’re having trouble signing you in.

[...]

PR #10779, opened 2026-04-19
feat: add SAML_ACS_BASE_URL for reverse proxy deployments

### LINKED ISSUE #10533: Add env var to dynamically configure SAML ACS URL
### Feature search

Prowler API, Prowler UI

Not provider-specific

### New feature motivation

Currently, configuring the correct ACS URL for SAML SSO in self-hosted deployments requires building a custom frontend container, as the value is hardcoded to `http://prowler-api:8080/`. This is impractical and introduces unnecessary complexity for users running Prowler behind a proxy or load balancer. See related discussion: #10498 and issue #9724.

### Solution Proposed

Introduce a new environment variable (e.g. SAML_ACS_URL) that allows users to dynamically set the ACS URL at runtime via the .env file, without needing to rebuild or customize any container image.

### Use case and benefits

This would benefit any team running a self-hosted Prowler deployment behind a reverse proxy or load balancer who needs to configure SAML SSO with an external URL. It removes the need to maintain a custom container build just to change a URL, significantly simplifying the deployment process.

Building a custom frontend container to override the hardcoded ACS URL, as described in the [current documentation](https://docs.prowler.com/troubleshooting#saml/oauth-acs-url-incorrect-when-running-behind-a-proxy-or-load-balancer). This is not practical for some self-hosted users.

### Additional context

Originally raised by the community in discussion [#10498](https://github.com/prowler-cloud/prowler/discussions/10498).

boonchuan:
Fixes #10533
Related: #9724, #10498

## Problem
When Prowler runs behind a reverse proxy (nginx, Traefik, etc.), SAML ACS URLs are generated using the internal container hostname (`http://prowler-api:8080/...`) instead of the external domain. This breaks SSO authentication because the IdP's reply URL doesn't match.

The current workaround requires rebuilding the UI container with custom `NEXT_PUBLIC_API_BASE_URL` — impractical for most deployments.

## Solution
Added a `SAML_ACS_BASE_URL` environment variable and a lightweight middleware (`SAMLACSURLMiddleware`) that overrides request metadata for `/accounts/saml/*` paths so `request.build_absolute_uri()` produces the correct external URL.

### Usage
```env
SAML_ACS_BASE_URL=https://prowler.example.com
```

### How it works
- Middleware parses the env var at startup (zero overhead when not set)
- For SAML endpoints only, it overwrites `HTTP_HOST`, `SERVER_NAME`, `SERVER_PORT`, and `X-Forwarded-Proto`
- Django's `build_absolute_uri()` then generates correct external ACS URLs
- No-op when env var is empty — zero impact on existing deployments

### Files changed
- `api/src/backend/api/middleware.py` — added `SAMLACSURLMiddleware`
- `api/src/backend/config/django/base.py` — registered middleware
- `.env` — added `SAML_ACS_BASE_URL` with documentation

### DIFF
--- a/.env
+++ b/.env
@@ -173,3 +173,9 @@ LANGCHAIN_PROJECT=""
 RSS_FEED_SOURCES='[{"id":"prowler-releases","name":"Prowler Releases","type":"github_releases","url":"https://github.com/prowler-cloud/prowler/releases.atom","enabled":true}]'
 # Example with multiple sources (no trailing comma after last item):
 # RSS_FEED_SOURCES='[{"id":"prowler-releases","name":"Prowler Releases","type":"github_releases","url":"https://github.com/prowler-cloud/prowler/releases.atom","enabled":true},{"id":"prowler-blog","name":"Prowler Blog","type":"blog","url":"https://prowler.com/blog/rss","enabled":false}]'
+
+# SAML ACS URL Override
+# Set this to your external Prowler URL when running behind a reverse proxy.
+# This ensures SAML ACS URLs use your public domain instead of the internal container hostname.
+# Example: https://prowler.example.com
+# SAML_ACS_BASE_URL=
--- a/api/src/backend/api/middleware.py
+++ b/api/src/backend/api/middleware.py
@@ -53,3 +53,62 @@ class APILoggingMiddleware:
         )
 
         return response
+
+
+class SAMLACSURLMiddleware:
+    """
+    Middleware to override the host used for SAML ACS URL generation.
+
+    When Prowler runs behind a reverse proxy or load balancer, Django's
+    request.build_absolute_uri() uses the internal container hostname
+    (e.g., prowler-api:8080) instead of the external domain. This causes
+    SAML ACS URLs to be incorrect, breaking SSO authentication.
+
+    Set SAML_ACS_BASE_URL in your environment (e.g., https://prowler.example.com)
+    to override the host for SAML-related endpoints.
+
+    Fixes: https://github.com/prowler-cloud/prowler/issues/10533
+    """
+
+    SAML_PATH_PREFIX = "/accounts/saml/"
+
+    def __init__(self, get_response):
+        self.get_response = get_response
+        self.saml_acs_base_url = None
+
+        from config.env import env
+
+        url = env("SAML_ACS_BASE_URL", default="")
+        if url:
+            # Parse the URL to extract scheme, host, and port
+            from urllib.parse import urlparse
+
+            parsed = urlparse(url.rstrip("/"))
+            if parsed.scheme and parsed.hostname:
+                self.saml_acs_base_url = parsed
+                logging.getLogger("prowler").info(
+                    f"SAML ACS URL override active: {url}"
+                )
+
+    def __call__(self, request):
+        if (
+            self.saml_acs_base_url
+            and self.SAML_PATH_PREFIX in request.path
+        ):
+            parsed = self.saml_acs_base_url
+
+            # Override the host so build_absolute_uri() uses the external URL
+            host = parsed.hostname
+            if parsed.port and parsed.port not in (80, 443):
+                host = f"{host}:{parsed.port}"
+
+            request.META["HTTP_HOST"] = host
+            request.META["SERVER_NAME"] = parsed.hostname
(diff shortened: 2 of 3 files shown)


jfagoagas:
Hi @boonchuan thanks for this contribution. We'll look into this once we get a chance.

jfagoagas:
Hi @boonchuan we changed it as `not-planned` as we don't have enough bandwidth as of now. We know this is helpful for you as well as for the community, so if you want you can make this to work from the `/contrib` folder which is not maintained by us.

Thanks for using Prowler 🚀

boonchuan:
Hi @jfagoagas — thanks for the response and the suggestion. Happy to look at moving this.

One quick clarification before I rework: the current change is Django middleware (`api/src/backend/api/middleware.py`) that runs as part of the Prowler API server's request handling. Looking at the existing `/contrib/` folder, everything there appears to be standalone scripts or external configs — I'm not sure how middleware would run from `/contrib/` without being registered in Django's `MIDDLEWARE` setting in core.

Two ways I could see this working — would either fit what you have in mind?

1. **A `/contrib/reverse-proxy/` folder with a sample nginx + Docker Compose configuration** that achieves the same reverse-proxy SAML ACS rewriting at the proxy layer (no code change to Prowler).

2. **Documentation in `/contrib/reverse-proxy/README.md`** that walks through self-applying the middleware patch as a downstream change, for users who want it.

Or something else entirely? I want to make sure I rework this in a shape you'd actually accept rather than guessing. Thanks!

jfagoagas:
> Hi @jfagoagas — thanks for the response and the suggestion. Happy to look at moving this.
> 
> One quick clarification before I rework: the current change is Django middleware (`api/src/backend/api/middleware.py`) that runs as part of the Prowler API server's request handling. Looking at the existing `/contrib/` folder, everything there appears to be standalone scripts or external configs — I'm not sure how middleware would run from `/contrib/` without being registered in Django's `MIDDLEWARE` setting in core.
> 
> Two ways I could see this working — would either fit what you have in mind?
> 
> 1. **A `/contrib/reverse-proxy/` folder with a sample nginx + Docker Compose configuration** that achieves the same reverse-proxy SAML ACS rewriting at the proxy layer (no code change to Prowler).
> 2. **Documentation in `/contrib/reverse-proxy/README.md`** that walks through self-applying the middleware patch as a downstream change, for users who want it.
> 
> Or something else entirely? I want to make sure I rework this in a shape you'd actually accept rather than guessing. Thanks!

Both makes sense. Option may be 1 easier, what do you think?

boonchuan:
Sounds good — going with option 1. I'll put together a `/contrib/reverse-proxy/` folder with a sample nginx config and docker-compose example showing how to handle SAML ACS rewriting at the proxy layer.

Will open a fresh PR (or push to this branch, whichever you prefer) once it's drafted. Probably within the next 1-2 weeks. Thanks for the steer.

jfagoagas:
> Sounds good — going with option 1. I'll put together a `/contrib/reverse-proxy/` folder with a sample nginx config and docker-compose example showing how to handle SAML ACS rewriting at the proxy layer.
> 
> Will open a fresh PR (or push to this branch, whichever you prefer) once it's drafted. Probably within the next 1-2 weeks. Thanks for the steer.

You can use this branch, whatever is best for you. Thanks!

boonchuan:
Hi @jfagoagas — quick update on this. Looking at it again, I realised PR #10780 (which you merged into contrib/reverse-proxy/ last month) already handles the SAML ACS URL case at the nginx layer: it routes /accounts/saml/* to the API and sets X-Forwarded-Host / X-Forwarded-Proto so Django's build_absolute_uri() produces the correct external URLs without any code change.

Given that, would you prefer I:

1. Close this PR since the merged contrib/reverse-proxy/ already solves the underlying problem
2. Keep it open as the in-core option for users who can't / don't want to add an nginx container
3. Refresh this PR's description to point users at contrib/reverse-proxy/ first and keep this as a fallback

Happy to do whichever fits. Thanks again for merging #10780.

jfagoagas:
Thanks for the heads up @boonchuan! As you correctly pointed out `contrib/reverse-proxy/` solves the issue to use SAML with a reverse proxy. I'll go ahead and close this PR as the solution for the issue.

Thanks for contributing and using Prowler 🚀

PR #10780, opened 2026-04-20
feat: add nginx reverse proxy configuration (#8516)

### LINKED ISSUE #8516: Reverse Proxy Needs
### New feature motivation

_TLDR: I like to use reverse proxies, the base container setup doesn't seem to be reverse proxy friendly. Am I missing something?_

I just got back from DefCon and attended the talk put on by you guys. Was great, although the 2-Talk room was kind of bizarre (Literally had 2 talks happening in the same room). We're really excited to try this tool out in our environments for ourselves and our clients.

While I was in the talk I very quickly stood this up using docker compose on one of our dev servers at our office, but immediately ran into issues when trying to authenticate.

After looking through the container setup and some of the documentation, the containerized setup looks like it's assuming connectivity to either http://localhost:xxxx/service-paths, or connecting to something local to you in your network possibly via hostname?

For our use case and how I have our web apps set up, I like using reverse proxies to allow connectivity over the internet (don't worry, I have protections in place ;) but I found this setup doesn't work very well with reverse proxies since it has 2 different containers that provide services web services, which causes CORS issues when doing a one-for-one proxy

### Solution Proposed

My solution that I'm going to set up for our access (**and can help contribute to if it would be helpful**), will be using an NGinx container that functions as the intermediary between the outside reverse proxy connections to the prowler ui and api containers.

Our reverse proxy connection will proxy port 80 of the nginx container, which will internally be pointing to the prowler ui port 3000 container for most url requests, and then to prowler-api port 8080 for any requests for the /api/ path.

Unless there is an already existing solution I'm missing, I'll be moving forward with this and will document my work as I go along.

I considered mapping my reverse proxy system (pangolin), to two separate addresses for the ui and api containers, but it's not a very elegant solution and would cause CORS issues.

I also considered mapping this out manually within Traefik, but since our setup involves wireguard tunnels and mapping within a new system I'm not 100% familiar with yet, this idea felt like it would take more time and patience than I have right now.

boonchuan:
Fixes #8516

## Problem
Prowler's default Docker setup exposes UI (port 3000) and API (port 8080) as separate services. This causes CORS issues and authentication failures when accessed through an external reverse proxy.

Multiple users have hit this — running behind nginx, Azure App Gateway, Pangolin, and other proxies.

## Solution
Added a ready-to-use nginx reverse proxy config in `contrib/reverse-proxy/` that unifies both services behind a single port:

- `/api/*` and `/accounts/saml/*` → prowler-api:8080
- `/*` → prowler-ui:3000

### Usage

    docker compose -f docker-compose.yml \
      -f contrib/reverse-proxy/docker-compose.reverse-proxy.yml \
      up -d

### Files added
- `contrib/reverse-proxy/nginx.conf` — nginx routing config
- `contrib/reverse-proxy/docker-compose.reverse-proxy.yml` — compose override
- `contrib/reverse-proxy/README.md` — setup guide with Traefik and Caddy examples

### Architecture

    Internet → Reverse Proxy → prowler-nginx:80
                                  ├── /api/*          → prowler-api:8080
                                  ├── /accounts/saml/ → prowler-api:8080
                                  └── /*              → prowler-ui:3000

Pairs with the SAML ACS URL fix in #10533 for complete reverse proxy + SSO support.

### DIFF
--- a/contrib/reverse-proxy/README.md
+++ b/contrib/reverse-proxy/README.md
@@ -1,0 +1,64 @@
+# Prowler Reverse Proxy Configuration
+
+Ready-to-use nginx configuration for running Prowler behind a reverse proxy.
+
+## Problem
+
+Prowler's default Docker setup exposes two separate services:
+- **UI** on port 3000
+- **API** on port 8080
+
+This causes CORS issues and authentication failures (especially SAML SSO) when accessed through an external reverse proxy, since the proxy typically exposes a single domain.
+
+## Solution
+
+This adds an nginx container that unifies both services behind a single port, correctly forwarding headers so that Django generates proper URLs for SAML ACS callbacks and API responses.
+
+## Quick Start
+
+From the prowler root directory:
+
+    docker compose -f docker-compose.yml \
+      -f contrib/reverse-proxy/docker-compose.reverse-proxy.yml \
+      up -d
+
+Access Prowler at http://localhost (port 80).
+
+## With an External Reverse Proxy
+
+Point your external reverse proxy to the prowler-nginx container on port 80.
+
+### Environment Variables
+
+| Variable | Default | Description |
+|----------|---------|-------------|
+| PROWLER_PROXY_PORT | 80 | Port exposed by the nginx proxy |
+
+### Example: Traefik
+
+    services:
+      nginx:
+        labels:
+          - "traefik.enable=true"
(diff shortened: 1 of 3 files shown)


jfagoagas (approved):
Thanks for this contribution @boonchuan!

PR #10783, opened 2026-04-20
feat(m365): add entra_conditional_access_policy_p1_license_utilization security check

HugoPBrito:
### Context

When more users consume Conditional Access than the organization has Azure AD Premium P1 licenses for, it creates a licensing compliance gap that Microsoft may flag during audits. This can lead to retroactive charges, service restrictions, or — worse — loss of Conditional Access protection for unlicensed users if Microsoft enforces entitlements. Attackers who compromise unlicensed accounts may bypass Conditional Access controls that administrators assumed were in place.

### Description

This check queries the Microsoft Graph beta `reports/azureADPremiumLicenseInsight` endpoint to retrieve the total number of entitled P1 licenses and the number of users actively utilizing Conditional Access. It passes when the P1 license count meets or exceeds the Conditional Access user count, and fails otherwise. The recommended remediation is to purchase additional P1 licenses or adjust Conditional Access policy scope to match entitlements.

### Steps to review

1. Review the check implementation at `prowler/providers/m365/services/entra/entra_conditional_access_policy_p1_license_utilization/`
3. Review compliance framework mappings in `prowler/compliance/m365/` to ensure the check is correctly mapped to relevant requirements
4. Run the check tests: `poetry run pytest tests/providers/m365/services/entra/entra_conditional_access_policy_p1_license_utilization/ -v`
   ```bash
   prowler m365 --check entra_conditional_access_policy_p1_license_utilization
   ```

### Checklist

<details>

</details>

#### SDK/CLI

### License


HugoPBrito:
After more investigation into how to avoid the Microsoft Graph `/beta` dependency, I'm going to close this PR.

The check relies on `reports/azureADPremiumLicenseInsight`, which Microsoft only exposes in `/beta` and explicitly states is not supported for production use. The alternatives I evaluated don't solve it:

- **PowerShell**: `Get-MgBetaReportAzureAdPremiumLicenseInsight` is a wrapper over the same beta endpoint, so it doesn't remove the dependency. The v1.0 cmdlets (`Get-MgSubscribedSku`, `Get-MgIdentityConditionalAccessPolicy`) don't expose this aggregated insight, and MSOnline / AzureAD are deprecated.
- **Microsoft Graph v1.0**: `azureADPremiumLicenseInsight` has no v1.0 equivalent. Its `userCount` measures users who *historically triggered* CA in sign-in logs (computed by Microsoft internally) and is not reproducible from v1.0 endpoints. The closest derivable signal — "users in scope of enabled CA policies" via `subscribedSkus` + `conditionalAccess/policies` + transitive group/role expansion — is an upper bound, not the same number, and would diverge from what admins see in the Entra portal, creating permanent confusion in findings.

Since exact coverage isn't possible with stable APIs today, the cleanest path is to drop this for now and revisit once Microsoft promotes the endpoint (or an equivalent) to v1.0.

PR #10787, opened 2026-04-20
feat(ui): Add user expulsion from tenants with JWT authentication fix

Davidm4r:
### Context

  This PR implements a comprehensive user and membership management system that allows tenant owners to expel users from their organizations. During implementation, a critical authentication bug was discovered and fixed where JWT token parsing was failing to extract user IDs, preventing proper role-based access control.

  ### Description

  **Key Changes:**
  1. **User Expulsion Functionality**: Complete implementation allowing tenant owners to remove members from organizations
  2. **Critical JWT Authentication Fix**: Fixed JWT claim mapping from non-existent `user_id` to standard `sub` field
  3. **Smart Account Management**: Automatic account deletion when expelled user has no other memberships
  4. **Security Enhancements**: Token revocation and proper authorization checks
  5. **Modern React Patterns**: Migration from manual `useState` to React 19's `useTransition`

  **Technical Improvements:**
  - **UI**: Expel user modal with clear UX about consequences (account deletion vs retention)
  - **API**: Enhanced `TenantMembersViewSet.destroy()` with token blacklisting and atomic transactions
  - **Auth**: Fixed `applyDecodedClaims()` to use JWT `sub` field as fallback for `user_id`
  - **Types**: Derived TypeScript types from Zod schemas for better type safety
  - **Tests**: Comprehensive coverage including edge cases and token blacklisting scenarios

  **Dependencies:**
  - No new dependencies required
  - Backward-compatible JWT authentication fix

  ### Steps to review

  1. **Authentication Fix Verification**:
     - Check JWT token parsing in `auth.config.ts` - verify `sub` fallback logic
     - Test login/logout flows to ensure no regression
     - Verify session.userId is properly populated across the app

  2. **User Expulsion Flow**:
     - Navigate to `/users` page as a tenant owner
     - Verify "Expel from organization" appears in user row actions (not for own row)
     - Test expulsion of user with multiple memberships (account retained)
     - Test expulsion of user with single membership (account deleted)
     - Verify token blacklisting works (expelled user loses access immediately)

  3. **Security & Edge Cases**:
     - Verify owners cannot expel themselves if they're the last owner
     - Test that members cannot access expel functionality
     - Confirm proper tenant isolation (can only expel from own tenant)

  4. **Documentation & Tests**:
     - Review updated CHANGELOGs across UI, API, and docs
     - Run test suite to verify all expulsion scenarios pass
     - Check documentation screenshots and flow descriptions

  ### Checklist

  <details>

  </details>

  #### SDK/CLI

  #### UI
  - [x] Screenshots/Video of the functionality flow (if applicable) - Mobile (X < 640px)
  - [x] Screenshots/Video of the functionality flow (if applicable) - Table (640px > X < 1024px)
  - [x] Screenshots/Video of the functionality flow (if applicable) - Desktop (X > 1024px)

  #### API

  ### License


api/src/backend/api/v1/views.py
```diff
@@ -1398,7 +1403,54 @@ def destroy(self, request, *args, **kwargs):
...
+                    ]
+                )
+
+            has_other_memberships = (
+                Membership.objects.using(MainRouter.admin_db)
+                .filter(user_id=user_to_check_id)
+                .exists()
+            )
+            if not has_other_memberships:
+                User.objects.using(MainRouter.admin_db).filter(
```
pfe-nazaries, line 1484: **Orphaned role grants after membership deletion**

This transaction deletes the `Membership` (and the `User` row when no other memberships remain), but it never deletes the `UserRoleRelationship` rows scoped to `membership_to_delete.tenant_id`.

`UserRoleRelationship.user` uses `on_delete=CASCADE`, so the grants are only cleaned up in the branch where the `User` row is deleted (line 1449-1451). When the expelled user keeps other memberships, the `User` row stays and their role grants for this tenant remain in `role_user_relationship`.

Consequences:
- `get_role(user, tenant_id=...)` in `api/rbac/permissions.py:54-64` will keep returning a role for a tenant the user no longer belongs to.
- If the same user is re-invited to the tenant later, they inherit the previous role silently instead of getting the one from the new invitation.
- Any access token issued before expulsion that is still within `ACCESS_TOKEN_LIFETIME` will still resolve permissions via the stale grant (refresh tokens are already blacklisted, good — but the permission check path is independent).

Suggested fix — add inside the same `transaction.atomic(using=MainRouter.admin_db)` block, before or right after the `Membership` delete:

```python
UserRoleRelationship.objects.using(MainRouter.admin_db).filter(
    user_id=user_to_check_id,
    tenant_id=membership_to_delete.tenant_id,
).delete()
```

Please also add a regression test in `test_views.py` covering: user with two memberships → delete one → assert `UserRoleRelationship` for that tenant is gone while the grant in the other tenant is preserved.

pfe-nazaries (changes requested):
Found one blocker via CodeRabbit review. Leaving an inline comment — happy to push the fix + tests once confirmed.

api/CHANGELOG.md
```diff
@@ -61,6 +68,7 @@ All notable changes to the **Prowler API** are documented in this file.
 - Attack Paths: Periodic cleanup of stale scans with dead-worker detection via Celery inspect, marking orphaned `EXECUTING` scans as `FAILED` and recovering `graph_data_ready` [(#10387)](https://github.com/prowler-cloud/prowler/pull/10387)
 - Attack Paths: Replace `_provider_id` property with `_Provider_{uuid}` label for provider isolation, add regex-based label injection for custom queries [(#10402)](https://github.com/prowler-cloud/prowler/pull/10402)
 
+
```
AdriiiPRodri:
```suggestion
```

api/src/backend/api/v1/views.py
AdriiiPRodri: This doesn't remove the role, we need to remove orphans roles too.

Already checked by David, resolved

api/src/backend/api/v1/views.py
```diff
@@ -1398,7 +1403,54 @@ def destroy(self, request, *args, **kwargs):
...
                 )
 
-        membership_to_delete.delete()
+        user_to_check_id = membership_to_delete.user_id
+        # All writes run on the admin connection so that the uncommitted
+        # membership delete is visible to the subsequent "other memberships"
+        # check. Splitting the delete and the check across the default
+        # (prowler_user, RLS) and admin connections caused the admin side to
+        # miss the just-deleted row and leave the User row orphaned.
+        with transaction.atomic(using=MainRouter.admin_db):
```
AdriiiPRodri, line 1416: Can we remove the `using(MainRouter.admin_db)` from every `objects.using` if this is already set at a top level?

AdriiiPRodri (changes requested):

api/src/backend/api/v1/views.py
```diff
@@ -1398,7 +1403,54 @@ def destroy(self, request, *args, **kwargs):
...
+                        BlacklistedToken(token_id=token_id)
+                        for token_id in outstanding_token_ids
+                        if token_id not in already_blacklisted
+                    ]
+                )
+
+            has_other_memberships = (
+                Membership.objects.using(MainRouter.admin_db)
+                .filter(user_id=user_to_check_id)
+                .exists()
```
pfe-nazaries, line 1481: **Note on account deletion:** The `User` row is only removed when the expelled membership was the user's last one. In practice, users onboarded via the invitation + sign-up flow always retain a personal tenant (created during sign-up), so this deletion branch is a defensive safeguard rather than the common path hit from the UI. It is exercised by the new pytest cases (`test_expel_user_deletes_account_if_last_membership`, `test_expel_user_keeps_account_if_has_other_memberships`).

ui/CHANGELOG.md
```diff
@@ -4,10 +4,15 @@ All notable changes to the **Prowler UI** are documented in this file.
...
 
+### 🔄 Changed
+
+- Allows tenant owners to expel users from their organizations  [(#10787)](https://github.com/prowler-cloud/prowler/pull/10787)
+
 ### ❌ Removed
 
 - Backward-compatibility middleware redirect from `/sign-up?invitation_token=…` to `/invitation/accept?invitation_token=…`; new invitation emails use `/invitation/accept` directly
 
+
```
josema-xyz, line 15: Extra white line.

api/CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 All notable changes to the **Prowler API** are documented in this file.
 
+## [1.26.0] (Prowler UNRELEASED)
+
+### 🔄 Changed
+
+- Allows tenant owners to expel users from their organizations  [(#10787)](https://github.com/prowler-cloud/prowler/pull/10787)
+
```
josema-xyz, line 10: No `---` line here.

api/src/backend/api/v1/views.py
```diff
@@ -1398,7 +1403,54 @@ def destroy(self, request, *args, **kwargs):
...
+            )
+            if not has_other_memberships:
```
josema-xyz, line 1483: This also runs when a user leaves on their own, they'd lose their whole account silently. Is that the desired behaviour?

api/src/backend/api/v1/views.py
```diff
@@ -1398,7 +1403,54 @@ def destroy(self, request, *args, **kwargs):
...
+                .filter(user_id=user_to_check_id)
+                .values_list("id", flat=True)
+            )
+            if outstanding_token_ids:
+                already_blacklisted = set(
+                    BlacklistedToken.objects.using(MainRouter.admin_db)
+                    .filter(token_id__in=outstanding_token_ids)
+                    .values_list("token_id", flat=True)
+                )
+                BlacklistedToken.objects.using(MainRouter.admin_db).bulk_create(
```
josema-xyz, line 1470: Small race between the pre-query and the insert. `bulk_create([...], ignore_conflicts=True)` can avoid it and the "already blacklisted" query can be dropped.

api/src/backend/api/v1/views.py
josema-xyz: This description is out of date. It doesn't mention token blacklisting or the potential account deletion.

ui/auth.config.ts
```diff
@@ -90,7 +90,8 @@ const applyDecodedClaims = (
     target.accessTokenExpires = decodedToken.exp
       ? decodedToken.exp * 1000
       : target.accessTokenExpires;
-    target.user_id = decodedToken.user_id ?? target.user_id;
+    // Map standard JWT "sub" field to user_id if user_id is not present
+    target.user_id = decodedToken.user_id ?? decodedToken.sub ?? target.user_id;
```
josema-xyz: Doesn't the token only carry `sub`? If so, `decodedToken.user_id` is always `undefined` and `decodedToken.sub` is enough on its own. Also `CustomJwtPayload` has `user_id: string` as required, which doesn't match.

api/src/backend/api/v1/views.py
Davidm4r: transaction.atomic(using=db) only sets the transaction boundary on that database, but doesn't
  change Django's ORM query routing.
  
  Each .objects.filter() call still uses the default database unless explicitly told to use a
  different one with .using(db).

ui/auth.config.ts
Davidm4r: I dont0 understand why user_id was there, because it was never used.
The commit it's from 14 months ago.

I made the user_id optional and add the sub. This is the safest option.

api/src/backend/api/v1/views.py
Davidm4r: Is the desired behaviour. But we should update the UI to tell the user that he is going to get their account deleted.
We can do it in another PR.

ui/CHANGELOG.md
josema-xyz: 🚀

api/CHANGELOG.md
josema-xyz: 🚀

api/src/backend/api/v1/views.py
josema-xyz: 🚀

api/src/backend/api/v1/views.py
josema-xyz: 🚀

josema-xyz (dismissed):

pfe-nazaries (dismissed):

AdriiiPRodri (dismissed):

AdriiiPRodri (approved):

jfagoagas (approved):

PR #10842, opened 2026-04-22
chore(pre-commit): add priority tiers to .pre-commit-config.yaml

adriantomas:
### Context

The project already uses prek (a Rust-based pre-commit replacement) but still relies on `.pre-commit-config.yaml` without priority annotations. Adding priority tiers unlocks prek's concurrent hook execution while preserving Dependabot support for automatic hook updates.

### Description

* Add priority tiers to all hooks in `.pre-commit-config.yaml` so prek can run hooks concurrently within each tier.
* Sets `default_install_hook_types` to include `pre-push` so the trufflehog hook triggers on push as intended.

### Steps to review

1. Verify priority assignments are sensible (fixers < validators < formatters < linters < security < deps).
2. Run `prek install -f` and `prek run --all-files` to confirm all hooks execute correctly.

### Checklist

* Review if backport is needed. -> Not needed                                                                                                                                              
* Review if is needed to change the Readme.md -> Not needed
* Ensure new entries are added to CHANGELOG.md, if applicable.

#### SDK/CLI

* Are there new checks included in this PR? No                                                                                                                                            

### License


prek.toml
```diff
@@ -0,0 +1,227 @@
...
+    files = '.*\.py',
+    exclude = { glob = ["{contrib,skills}/**", "**/.venv/**", "**/*_test.py"] },
+    priority = 40,
+  },
+  # TODO: Botocore needs urllib3 1.X so we need to ignore these vulnerabilities 77744,77745. Remove this once we upgrade to urllib3 2.X
+  # TODO: 79023 & 79027 knack ReDoS until `azure-cli-core` (via `cartography`) allows `knack` >=0.13.0
+  # TODO: 86217 because `alibabacloud-tea-openapi == 0.4.3` don't let us upgrade `cryptography >= 46.0.0`
+  # TODO: 71600 CVE-2024-1135 false positive - fixed in gunicorn 22.0.0, project uses 23.0.0
+  {
+    id = "safety",
```
jfagoagas: We need to get changes from https://github.com/prowler-cloud/prowler/pull/10845 in here.

adriantomas: done!

puchy22 (changes requested):
Hi @adriantomas, thanks for the contribution! The hook priority feature is genuinely great.

That said, I'm hesitant to move forward with the full migration from `.pre-commit-config.yaml` to `prek.toml`. The formatting is nicer, but the tradeoff isn't worth it for us right now: Dependabot currently doesn't support automatic hook updates for the TOML format, which means we'd lose automated dependency management for our hooks. There's an [open issue](https://github.com/dependabot/dependabot-core/issues/14624) tracking this on the Dependabot side, but it doesn't seem close to being resolved.

Would you be open to change the PR to just adds priority tiers to the existing `.pre-commit-config.yaml` instead? That way we get the concurrency benefits without giving up Dependabot support.

adriantomas:
> Hi @adriantomas, thanks for the contribution! The hook priority feature is genuinely great.
> 
> That said, I'm hesitant to move forward with the full migration from `.pre-commit-config.yaml` to `prek.toml`. The formatting is nicer, but the tradeoff isn't worth it for us right now: Dependabot currently doesn't support automatic hook updates for the TOML format, which means we'd lose automated dependency management for our hooks. There's an [open issue](https://github.com/dependabot/dependabot-core/issues/14624) tracking this on the Dependabot side, but it doesn't seem close to being resolved.
> 
> Would you be open to change the PR to just adds priority tiers to the existing `.pre-commit-config.yaml` instead? That way we get the concurrency benefits without giving up Dependabot support.

Sure @puchy22, I can make those changes :) 

In any case I would strongly suggest to use local hooks instead of remote repository hooks, so that versions are only defined in the pyproject.toml. What do you think of this proposal in a different PR?

puchy22:
Hi again @adriantomas, I understand the point of using only local hooks, but the ones for TOML Formatter, Zizmor, Shellcheck, and Hadolint would still need to be remote repositories; since they are not Python packages and it’s more convenient for Prek itself to install them than to have to maintain an installation script for all these tools.

It would make sense to switch all the Python ones, but I don’t see much of an advantage in the short term.

puchy22 (approved):
LGTM! Thanks @adriantomas for this job! 🚀

PR #10843, opened 2026-04-22
fix(api): reaggregate resource inventory and attack surface after muting findings

AdriiiPRodri:
### Context

Follow-up to [#10827](https://github.com/prowler-cloud/prowler/pull/10827), which made Overview and Finding Groups reflect newly-muted findings by extending `reaggregate_all_finding_group_summaries_task`. While validating that fix we saw that several other endpoints backed by their own pre-aggregated tables were still stale after muting:

- `/overviews/resource-groups` (resource inventory widget)
- `/overviews/categories`
- `/overviews/attack-surfaces`
- `/overviews/findings-severity/timeseries` (severity timeline)

Both the reaggregation dispatch and the underlying aggregators were affected.

### Description

Two-part fix:

1. **Dispatch**: the post-mute `reaggregate-all-finding-group-summaries` task now also re-runs, for each latest scan per `(provider, day)`:
   - `backfill_scan_resource_group_summaries_task` -> `ScanGroupSummary`
   - `backfill_scan_category_summaries_task` -> `ScanCategorySummary`
   - `aggregate_attack_surface_task` -> `AttackSurfaceOverview`

   Services watchlist and severity timeseries were already covered, since they read from `ScanSummary` and `DailySeveritySummary` which the pipeline was recomputing.

2. **Idempotency**: four aggregators used plain `bulk_create` (or a `already backfilled` short-circuit plus silent `ignore_conflicts=True`). The first post-mute run tripped the `unique_*` constraints, aborting the Celery chain and leaving the dependent aggregators unexecuted (observable as a PostgreSQL `duplicate key value violates unique constraint "unique_scan_summary"` during the first reaggregation attempt). They now delete the scan's existing rows before `bulk_create` so the write is atomic and re-runnable, and dropped combinations no longer linger in the summary table:
   - `aggregate_findings` (scan.py)
   - `aggregate_attack_surface` (scan.py)
   - `backfill_scan_resource_group_summaries` (backfill.py)
   - `backfill_scan_category_summaries` (backfill.py)

Other pre-aggregated tables (`ResourceScanSummary`, `ComplianceOverviewSummary`, `ProviderComplianceScore`) do not depend on `Finding.muted` and are intentionally left out of the reaggregation fan-out.

### Steps to review

1. Run a scan that produces FAIL findings with resource groups (e.g. an AWS provider that yields S3/IAM FAILs so several `resource_groups` values are populated).
2. Hit `/overviews/resource-groups`, `/overviews/categories`, `/overviews/attack-surfaces` and `/overviews/findings-severity/timeseries`. Note the totals.
3. `POST /mute-rules` on a subset of those finding IDs.
4. Wait for the Celery `overview` queue to drain (`docker compose logs worker | grep -E "reaggregate|scan-(summary|daily-severity|finding-group-summaries|resource-group-summaries|category-summaries|attack-surface)"`).
5. Hit the same overview endpoints again. Muted counts should go up, non-muted counts should go down by the same amount on every one of them.
6. Confirm no `duplicate key value violates unique constraint` errors appear in the worker log during step 4.

Automated:

```bash
cd api && poetry run pytest \
  src/backend/tasks/tests/test_backfill.py::TestBackfillScanCategorySummaries \
  src/backend/tasks/tests/test_backfill.py::TestBackfillScanGroupSummaries \
  src/backend/tasks/tests/test_tasks.py::TestReaggregateAllFindingGroupSummaries \
  src/backend/tasks/tests/test_scan.py::TestAggregateFindings \
  src/backend/tasks/tests/test_scan.py::TestAggregateAttackSurface \
  -x
```

### Checklist

<details>

</details>

#### API

### License


api/src/backend/tasks/jobs/scan.py
```diff
@@ -1546,15 +1546,23 @@ def aggregate_attack_surface(tenant_id: str, scan_id: str):
...
+    with rls_transaction(tenant_id):
+        # Delete first so re-runs (e.g. post-mute reaggregation) don't hit
+        # the `unique_attack_surface_per_scan` constraint.
+        AttackSurfaceOverview.objects.filter(
+            tenant_id=tenant_id, scan_id=scan_id
+        ).delete()
+        if overview_objects:
+            AttackSurfaceOverview.objects.bulk_create(
+                overview_objects, batch_size=500
+            )
```
josema-xyz: If two of these hit the same scan at once (a mute while a scan is finishing, or two quick mutes), both do the `DELETE`, both do the `INSERT`, and the second gets a unique key error. These tasks have no retry, so the chain dies and the overviews stay stale. Same pattern also lives in `backfill.py` for `ScanCategorySummary` and `ScanGroupSummary`.

api/src/backend/tasks/tasks.py
```diff
@@ -827,6 +831,15 @@ def reaggregate_all_finding_group_summaries_task(tenant_id: str):
                     aggregate_finding_group_summaries_task.si(
                         tenant_id=tenant_id, scan_id=scan_id
                     ),
+                    backfill_scan_resource_group_summaries_task.si(
+                        tenant_id=tenant_id, scan_id=scan_id
+                    ),
+                    backfill_scan_category_summaries_task.si(
+                        tenant_id=tenant_id, scan_id=scan_id
+                    ),
```
josema-xyz: Noticed these two stay on the `backfill` queue while the rest of the group runs on `overview`. Since our Cloud workers listen to a single queue, I think the post-mute refresh ends up waiting on the `backfill` pool before the user sees fresh data. Was that intentional?

AdriiiPRodri: Great catch! Not intentional, fixed

api/src/backend/tasks/jobs/scan.py
AdriiiPRodri: Good catch. Fixed by switching the four affected aggregators (`aggregate_findings`, `aggregate_attack_surface`, `backfill_scan_resource_group_summaries` and `backfill_scan_category_summaries`) from **DELETE+INSERT** to `bulk_create(update_conflicts=True, unique_fields=[...], update_fields=[...])`. Under PostgreSQL this compiles to a single atomic **INSERT ON CONFLICT (…) DO UPDATE SET ...**, so two concurrent writers on the same scan no longer race: one upsert wins, the other updates the same row, both commit, no `IntegrityError`, and the Celery chain keeps going.

This is only safe because the underlying aggregators (`aggregate_resource_group_counts`, `aggregate_category_counts`, and the ORM `.values().annotate()` groupings in `aggregate_findings` and `aggregate_attack_surface`) always generate every combination key. Muting only reclassifies counts, it never drops a key, so we do not need the delete step to prune orphans. This follows the same pattern as `aggregate_finding_group_summaries`, which already used upsert

api/src/backend/tasks/tests/test_scan.py
```diff
@@ -3403,19 +3405,14 @@ def test_aggregate_findings_is_idempotent_on_rerun(
...
-                "total",
+                *value_columns
             )
         )
 
+        # Upsert preserves the original row identities; values stay stable
+        # because the underlying Finding set is unchanged between runs.
         assert second_run_rows == first_run_rows
-        assert first_run_ids.isdisjoint(second_run_ids)
+        assert first_run_ids == second_run_ids
```
josema-xyz, line 3415: This test runs twice with the same findings and checks the rows don't change. Could we add the opposite case too? Mute a finding between the two runs, then check the row's `failed` / `muted` counters actually move. And the same idea for the two backfill tests. Without it, if someone breaks the fix later, the tests would still pass.

AdriiiPRodri: Added

josema-xyz (approved):

PR #10849, opened 2026-04-22
refactor(m365): scope entra_emergency_access_exclusion to Block-grant policies

HugoPBrito:
### Context

The existing `entra_emergency_access_exclusion` check evaluates whether a tenant has at least one emergency access (break glass) account or group excluded from **every enabled Conditional Access policy**. That criterion is overly strict: non-blocking policies (for example, MFA requirements) are not the source of lockout risk, and a tenant can have a perfectly valid emergency access setup where the break glass account simply does not need to be excluded from every MFA-style policy. Conversely, the check was permissive about what really matters: a **blocking** policy that still applies to the emergency account.

The original PR (`entra_emergency_access_users_not_blocked`) tried to address this by adding a second check focused on blocking policies, but the resulting setup duplicated the detection heuristic and introduced two checks that overlap in intent and disagree in a real-world edge case.

### Description

This PR refactors `entra_emergency_access_exclusion` to evaluate the criterion that actually defines an emergency access account: **exclusion from every enabled Conditional Access policy whose grant control is `Block`**. Non-blocking policies are no longer required to be in the exclusion list because they do not produce a hard lockout.

Behavioural summary:

- **PASS** when at least one user or group is excluded from every enabled Conditional Access policy with a `Block` grant control, or when there are no enabled blocking policies.
- **FAIL** when one or more enabled blocking policies exist and no user or group is excluded from all of them.

Compliance mapping for `entra_emergency_access_users_not_blocked` (CIS 4.0 / CIS 6.0 / ISO 27001 2022) is preserved by mapping `entra_emergency_access_exclusion` to the same controls. The new check file and its tests are removed.

### Steps to review

1. Review the updated check implementation at `prowler/providers/m365/services/entra/entra_emergency_access_exclusion/`.
2. Review the metadata file for the updated description, risk and remediation that reflect the new Block-grant scope.
3. Run the tests: `poetry run pytest tests/providers/m365/services/entra/entra_emergency_access_exclusion/ -v`.
4. Diff `prowler/compliance/m365/cis_4.0_m365.json`, `prowler/compliance/m365/cis_6.0_m365.json` and `prowler/compliance/m365/iso27001_2022_m365.json` to verify the mapping changes.
5. **Run against a real environment** (if possible):
   ```bash
   prowler m365 --check entra_emergency_access_exclusion
   ```

### Checklist

<details>

</details>

#### SDK/CLI
- Are there new checks included in this PR? **No.** This PR scopes the existing `entra_emergency_access_exclusion` check to blocking policies and removes the duplicate `entra_emergency_access_users_not_blocked` proposed in earlier revisions.

### License


HugoPBrito:
Reshaped this into a scope change on `entra_emergency_access_exclusion` rather than a separate check. Before it required exclusion from *every* enabled CA policy, which produced false FAILs in any tenant with lots of MFA or compliant-device policies, because the break-glass account doesn't need to be excluded from those: they're satisfiable requirements, not denials.

Criterion now is just: the user or group has to be excluded from every enabled CA policy with a `Block` grant control. If there are no Block-grant policies enabled, PASS straight away. Session controls and other grants stay out on purpose.

Went with this because it removes the false FAILs, doesn't duplicate the existing check, and the PASS message ends up actionable (`excluded from all N enabled CA policies with a Block grant control`).

danibarranqueroo (approved):

jfagoagas (approved):
Compliance ✅

PR #10851, opened 2026-04-22
fix(ui): prevent rescheduling scans during credential update

alejandrobailo:
### Context

Fix PROWLER-1426.

When updating provider credentials through the Provider Wizard, the UI was incorrectly advancing to the **Launch Scan** step after a successful connection test. This allowed users with no existing schedule to silently create one during a credential rotation, and users with an existing schedule hit a confusing `409 Conflict` from the backend. Scheduling must remain a deliberate, separate action — never a side effect of rotating credentials.

### Description

- `handleTestSuccess` in `use-provider-wizard-controller.ts` now short-circuits in UPDATE mode by calling `handleClose()` (which closes the modal and refreshes the providers list), instead of advancing to the Launch step.
- `WizardStepper` now accepts an optional `steps` prop so UPDATE mode renders only the first 3 steps (hides the misleading "Launch Scan" entry). The existing `STEPS` array remains the single source of truth and is re-exported as `PROVIDER_WIZARD_STEPS`.
- `ProviderWizardModal` reads `mode` from the controller and passes the filtered 3-step list to `WizardStepper` in UPDATE mode.
- The controller unit test that previously codified the buggy behavior (`moves to launch step after a successful connection test in update mode`) is rewritten as `closes the wizard after a successful connection test in update mode` and asserts the correct UX (`onOpenChange(false)` + `router.refresh()` called; current step is not `LAUNCH`).
- **No backend changes.** `api/src/backend/tasks/beat.py` already guards against duplicate schedules with `ConflictException`; that stays as defense in depth.

### Steps to review

1. **Controller change** — `ui/components/providers/wizard/hooks/use-provider-wizard-controller.ts`
   - Confirm `handleTestSuccess` branches on `useProviderWizardStore.getState().mode`. `getState()` is used (instead of the subscribed `mode` closure) so the handler always reads the latest store value at click time.
   - Confirm `mode` is exposed in the controller's return object for the modal to consume.
2. **Stepper change** — `ui/components/providers/wizard/wizard-stepper.tsx`
   - Confirm the new optional `steps` prop defaults to the full `STEPS` array.
   - Confirm the connector-rendering check uses `steps.length - 1` (not the hardcoded `STEPS.length - 1`), so subsets don't grow a trailing connector.
3. **Modal change** — `ui/components/providers/wizard/provider-wizard-modal.tsx`
   - Confirm `UPDATE_MODE_WIZARD_STEPS = PROVIDER_WIZARD_STEPS.slice(0, 3)` is passed only when `mode === PROVIDER_WIZARD_MODE.UPDATE`; otherwise `undefined` lets the stepper fall back to its default (all 4 steps).
4. **Test change** — `ui/components/providers/wizard/hooks/use-provider-wizard-controller.test.tsx`
   - Confirm the renamed test asserts `onOpenChange(false)`, `refreshMock` called once, and `currentStep !== LAUNCH`.
5. **Manual UI QA**
   - **UPDATE on provider without schedule:** open row actions → Update Credentials → Credentials → Test → expect modal to close and providers table to refresh.
   - **UPDATE on provider with existing schedule:** same as above; no 409 reachable from this path.
   - **ADD (regression):** new provider end-to-end still reaches the Launch Scan step and can schedule.
   - **Stepper:** UPDATE shows 3 steps (no "Launch Scan"); ADD shows 4 steps.
6. **Automated checks** (run from `ui/`):
   - `pnpm run healthcheck` — typecheck + lint clean.
   - `pnpm exec vitest run` — 535/535 tests passing, including the rewritten wizard test.

### Checklist

<details>

</details>

#### SDK/CLI

#### UI

#### API

### License

### DIFF
--- a/ui/components/providers/wizard/provider-wizard-modal.tsx
+++ b/ui/components/providers/wizard/provider-wizard-modal.tsx
@@ -28,7 +28,10 @@ import { TestConnectionStep } from "./steps/test-connection-step";
 import type { OrgWizardInitialData, ProviderWizardInitialData } from "./types";
 import { PROVIDER_WIZARD_STEPS, WizardStepper } from "./wizard-stepper";
 
-const UPDATE_MODE_WIZARD_STEPS = PROVIDER_WIZARD_STEPS.slice(0, 3);
+const UPDATE_MODE_WIZARD_STEPS = PROVIDER_WIZARD_STEPS.slice(
+  0,
+  PROVIDER_WIZARD_STEP.LAUNCH,
+);
 
 interface ProviderWizardModalProps {
   open: boolean;
--- a/ui/components/providers/wizard/provider-wizard-modal.tsx
+++ b/ui/components/providers/wizard/provider-wizard-modal.tsx
@@ -10,7 +10,10 @@ import { DialogHeader, DialogTitle } from "@/components/shadcn/dialog";
 import { Modal } from "@/components/shadcn/modal";
 import { useScrollHint } from "@/hooks/use-scroll-hint";
 import { ORG_SETUP_PHASE, ORG_WIZARD_STEP } from "@/types/organizations";
-import { PROVIDER_WIZARD_STEP } from "@/types/provider-wizard";
+import {
+  PROVIDER_WIZARD_MODE,
+  PROVIDER_WIZARD_STEP,
+} from "@/types/provider-wizard";
 
 import { useProviderWizardController } from "./hooks/use-provider-wizard-controller";
 import {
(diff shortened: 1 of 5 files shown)


ui/components/providers/wizard/provider-wizard-modal.tsx
```diff
@@ -23,7 +26,9 @@ import { WIZARD_FOOTER_ACTION_TYPE } from "./steps/footer-controls";
 import { LaunchStep } from "./steps/launch-step";
 import { TestConnectionStep } from "./steps/test-connection-step";
 import type { OrgWizardInitialData, ProviderWizardInitialData } from "./types";
-import { WizardStepper } from "./wizard-stepper";
+import { PROVIDER_WIZARD_STEPS, WizardStepper } from "./wizard-stepper";
+
+const UPDATE_MODE_WIZARD_STEPS = PROVIDER_WIZARD_STEPS.slice(0, 3);
```
josema-xyz: Could we use `PROVIDER_WIZARD_STEP.LAUNCH` instead of `3`? If someone adds a step later, the `3` won't update on its own.

josema-xyz (approved):

PR #10853, opened 2026-04-22
chore: route vulnerability references to canonical URLs

HugoPBrito:
### Context

Trivy image and IaC findings can expose Aqua advisory URLs or finding-specific advisory URLs. This PR routes those findings to canonical public references and makes the finding drawer label the remediation action by destination.

### Description

- Add SDK URL resolution for vulnerability references so CVE findings use cve.org, GitHub advisory IDs use GitHub Security Advisories, and Trivy rule IDs map to Prowler Hub without the `AVD-` prefix
- Use the resolved URLs in image and IaC findings for remediation and additional references
- Update the finding detail drawer to prefer finding-level recommendation URLs, label the CTA by destination (`View CVE`, `View in Prowler Hub`, `View Advisory`, `View Reference`), and keep URL-only remediation sections labeled
- Move the UI label rules into `ui/lib/vulnerability-references.ts` and cover CVE, Prowler Hub, GitHub advisory, malformed, and hostile hostname cases

### Steps to review

1. Run `poetry run pytest tests/lib/utils/test_vulnerability_references.py tests/providers/iac/iac_provider_test.py tests/providers/image/image_provider_test.py`
2. Run `pnpm test:run lib/vulnerability-references.test.ts components/findings/table/resource-detail-drawer/resource-detail-drawer-content.test.tsx`
3. Run `pnpm run typecheck`
4. Run `pnpm run lint:check`
5. Open a vulnerability or IaC finding with a CVE, Prowler Hub, or GitHub Security Advisory recommendation URL and confirm the drawer CTA label and href match the destination

### Checklist

<details>

</details>

#### SDK/CLI

#### UI (if applicable)

### License


pfe-nazaries (dismissed):
LGTM

ui/CHANGELOG.md
```diff
@@ -20,6 +20,10 @@ All notable changes to the **Prowler UI** are documented in this file.
 - Resources now use batch-applied filters, render metadata JSON with syntax highlighting, and more [(#10861)](https://github.com/prowler-cloud/prowler/pull/10861)
 - Table pagination controls now keep their arrows visible on hover in light theme, and more UI improvements [(#10862)](https://github.com/prowler-cloud/prowler/pull/10862)
 
+### 🐞 Fixed
+
+- Finding detail drawer now labels the remediation action by destination: "View CVE" for cve.org, "View in Prowler Hub" for Prowler Hub, "View Advisory" for GitHub Security Advisories [(#10853)](https://github.com/prowler-cloud/prowler/pull/10853)
```
pfe-nazaries: Move the changelog entry to next unreleased version

pfe-nazaries (dismissed):

ui/CHANGELOG.md
HugoPBrito: Done — moved the entry to a new `1.25.2 (Prowler UNRELEASED)` section, and also moved the SDK entry to `5.25.2 (Prowler UNRELEASED)` for consistency.

alejandrobailo (changes requested):
The title and description say this PR links each fix-available version when the recommendation points to a CVE advisory. I can't find that change anywhere in the diff. `statusExtended`, which is where the `(fix available: 5.7.13, ...)` string lives, still renders as plain text at line 894 of `resource-detail-drawer-content.tsx` and that block is identical to master.

What the PR does ship is three separate things:

- a dynamic label for the existing recommendation button (`View CVE` / `View Advisory` / `View in Prowler Hub` / `View Reference`),
- a new `View Resource` link under the actions menu,
- a small refactor so the recommendation button can render without accompanying text.

Only the first one matches the stated goal, and the changelog entry in `ui/CHANGELOG.md` describes behavior the code doesn't actually have. Either the version linkifier needs to land before merge, or the PR should be retitled and the View Resource piece split out so each change owns its own description, tests, and rollback boundary.

A few notes below.

---

**`resource-detail-drawer-content.test.tsx:702-703` (also `:749-750`)**

```ts
expect(screen.queryByRole("link", { name: "5.7.13" })).not.toBeInTheDocument();
```

This passes against master too. Nothing in the component creates a link with that accessible name in any state, so the assertion is true regardless of whether a linkifier exists. If you want to pin down "status_extended stays as plain text on a CVE finding", flip it to a positive assertion that the version appears as text inside the statusExtended paragraph. Once the real linkifier lands, you'll want a separate test where the same fixture renders six anchors.

**`resource-detail-drawer-content.test.tsx:807`**

```ts
expect(screen.queryByText(/avd\.aquasec\.com/)).not.toBeInTheDocument();
```

The fixture never puts an aqua URL into the component, so from the UI's point of view this is vacuously true. The aqua filtering happens in `prowler/lib/utils/vulnerability_references.py`, and there's already coverage for it on the SDK side. I'd drop this from the drawer test.

**`resource-detail-drawer-content.tsx:90-128`**

I'd pull these helpers out of the component file. The drawer is already 1.4k+ lines and these are pure URL utilities with no JSX. They've also got a Python twin (`prowler/lib/utils/vulnerability_references.py`) that owns the same hostname rules. If both sides are going to maintain the same contract, I'd rather see them in `ui/lib/vulnerability-references.ts` so the parallel is obvious to whoever touches one next.

A couple of things while you're at it:

`isProwlerHubUrl` uses `startsWith` on the full URL while the other two parse with `new URL()`. That's inconsistent and means `https://hub.prowler.com.evil.com/...` would match. Switch to `new URL(url).hostname === "hub.prowler.com"` like the others.

The two `try/catch` blocks around `new URL(url)` are the same shape. A `safeParseUrl` helper would tidy both.

The labels are inline string literals. The repo's TS convention (see `ui/CLAUDE.md`) is const objects with derived types, e.g.:

```ts
const RECOMMENDATION_LINK_LABEL = {
  CVE: "View CVE",
  HUB: "View in Prowler Hub",
  ADVISORY: "View Advisory",
  REFERENCE: "View Reference",
} as const;
type RecommendationLinkLabel =
  (typeof RECOMMENDATION_LINK_LABEL)[keyof typeof RECOMMENDATION_LINK_LABEL];
```

And the `if/if/if` cascade reads better as a small rules table you can scan top-to-bottom and extend in one line when the next host shows up.

One more: `cve.mitre.org` is still a valid CVE URL host that older Trivy data sometimes carries, and `cve.org` without `www` can appear too. Worth handling both unless we're certain the SDK normalizes everything to `www.cve.org` upstream.

**`resource-detail-drawer-content.tsx:342-346`**

`buildResourceDetailHref` for a single query param feels like one line too many. Either drop it inline or, since `buildComplianceDetailHref` already lives in this file, group both in `ui/lib/url-builders.ts` so the pattern is consistent.

**`resource-detail-drawer-content.tsx:465-467`**

```tsx
const recommendationUrl =
  f?.remediation.recommendation.url ||
  checkMeta.remediation.recommendation.url;
```

This shifts the source of the URL from check-level to finding-level. Two findings of the same check can now show different button labels in the drawer if their finding-level URLs point to different advisories. That's probably the intent, but it's a behavior change worth calling out in the changelog. Also, the `||` means an empty string falls through to `checkMeta`. If empty string is the canonical "no URL", a small `isNonEmptyString` check is clearer than relying on truthiness, otherwise the next reader will wonder if it's a typo for `??`.

**`resource-detail-drawer-content.tsx:819-832`**

`View Resource` isn't mentioned in the description or the changelog. It works, but I'd move it to its own PR. Mixing it with the recommendation-label change makes the diff harder to bisect if either piece misbehaves later, and it splits the test surface for two unrelated features.

**`resource-detail-drawer-content.tsx:903-936`**

When `recommendationLink` is the only thing inside the remediation card (no text, no CLI, no Terraform, no nativeiac), the card renders as a flex row with `gap-3` and a single chip on the right. It looks orphaned. Either keep the `Remediation:` label even when there's no text, or skip the card entirely when the URL is the only content.

---

Happy to pair on the refactor if it helps. I'd just rather this come back as either a complete fix-available implementation or a renamed PR that owns what it actually does.

HugoPBrito:
Updated in `e6fa47be6`.

What changed:
- Retitled and rewrote the PR description around the current scope: canonical vulnerability reference URLs plus destination-specific remediation labels.
- Updated `ui/CHANGELOG.md` so it no longer claims fix-available version linkification and now calls out finding-level recommendation URLs.
- Removed the unrelated `View Resource` drawer link from this PR.
- Dropped the vacuous drawer assertions for `5.7.13` links and Aqua URLs; kept positive assertions for status extended plain text and official CVE rendering.
- Moved URL label rules into `ui/lib/vulnerability-references.ts` with const-derived label types, a rules table, shared URL parsing, exact `hub.prowler.com` hostname matching, and support for `www.cve.org`, `cve.org`, and `cve.mitre.org`.
- Replaced the `||` URL fallback with explicit `isNonEmptyString` handling.
- Kept the `Remediation:` label when the recommendation link is the only remediation content.

Verified locally:
- `pnpm run healthcheck`
- `pnpm test:run lib/vulnerability-references.test.ts components/findings/table/resource-detail-drawer/resource-detail-drawer-content.test.tsx`
- `poetry run pytest tests/lib/utils/test_vulnerability_references.py tests/providers/iac/iac_provider_test.py tests/providers/image/image_provider_test.py`

Commit hooks also passed UI TypeScript, ESLint, unit tests, and build.

HugoPBrito:
Update on the previous review feedback: when I "removed the unrelated `View Resource` drawer link from this PR", I deleted the actual feature lines instead of just dropping them from this PR's diff. That feature is the one shipped by #10847 (commit `059b71d3` on master), so as the PR stood it would have reverted #10847 once merged. Caught while reviewing the diff against master.

Pushed `4dbb7c7` to restore it without touching the recommendation-label work:

- `buildResourceDetailHref` helper
- `resourceDetailHref` binding in `ResourceDetailDrawerContent`
- the JSX action below the resource actions menu
- the original positive assertion in `resource-detail-drawer-content.test.tsx` (replacing the negated `not.toBeInTheDocument` check)

Verified `git diff master -- ui/components/findings/table/resource-detail-drawer/resource-detail-drawer-content.tsx` no longer contains any `View Resource` / `buildResourceDetailHref` / `resourceDetailHref` deletions; only the recommendation link relabeling and `vulnerability-references` extraction remain. Local UI hooks (typecheck, lint, vitest, build) green.

alejandrobailo (dismissed):
LGTM!

danibarranqueroo (dismissed):

danibarranqueroo (dismissed):

alejandrobailo (dismissed):

alejandrobailo (approved):

danibarranqueroo (approved):

PR #10858, opened 2026-04-22
feat(aws): add cloudtrail_bedrock_logging_enabled security check

danibarranqueroo:
### Context

Without CloudTrail logging for Amazon Bedrock, model invocations and configuration changes remain invisible to security teams. Attackers could exploit foundation models for unauthorized content generation or data extraction — a technique known as **LLM jacking** — without leaving any audit trail, severely undermining forensic investigation and incident response capabilities.

### Description

This check verifies that at least one CloudTrail trail is actively logging Amazon Bedrock API calls, either through classic management events (with `IncludeManagementEvents: true` and `ReadWriteType: All`) or through advanced event selectors targeting Bedrock-specific resource types (`AWS::Bedrock::Model`, `AWS::Bedrock::Guardrail`, `AWS::Bedrock::KnowledgeBase`, `AWS::Bedrock::AgentAlias`, `AWS::Bedrock::FlowAlias`). It reports PASS if any active trail captures Bedrock events, and FAIL if no trail is configured for Bedrock logging. Remediation involves adding advanced event selectors for Bedrock data events on an existing multi-region trail.

### Steps to review

1. Review the check implementation at `prowler/providers/aws/services/cloudtrail/cloudtrail_bedrock_logging_enabled/`
3. Review compliance framework mappings in `prowler/compliance/aws/` to ensure the check is correctly mapped to relevant requirements
4. Run the check tests: `poetry run pytest tests/providers/aws/services/cloudtrail/cloudtrail_bedrock_logging_enabled/ -v`
   ```bash
   prowler aws --check cloudtrail_bedrock_logging_enabled
   ```

### Related Issues

https://prowlerpro.atlassian.net/browse/PROWLER-634

### Checklist

<details>

</details>

#### SDK/CLI

### License


tests/providers/aws/services/cloudtrail/cloudtrail_bedrock_logging_enabled/cloudtrail_bedrock_logging_enabled_test.py
```diff
@@ -0,0 +1,727 @@
...
+        ):
+            from prowler.providers.aws.services.cloudtrail.cloudtrail_bedrock_logging_enabled.cloudtrail_bedrock_logging_enabled import (
+                cloudtrail_bedrock_logging_enabled,
+            )
+
+            mock_cloudtrail_client.trails = None
+            check = cloudtrail_bedrock_logging_enabled()
+            result = check.execute()
+
+            assert len(result) == 0
```
lydiavilchez, line 924: Most branches in `_field_selector_matches_value()` aren't covered by tests, only Equals is tested.

prowler/providers/aws/services/cloudtrail/cloudtrail_bedrock_logging_enabled/cloudtrail_bedrock_logging_enabled.py
```diff
@@ -0,0 +1,162 @@
...
+            "AWS::Bedrock::KnowledgeBase",
+            "AWS::Bedrock::Model",
+            "AWS::Bedrock::Prompt",
+        }
+    )
+    BEDROCK_EVENT_SOURCES = frozenset(
+        {
+            "bedrock.amazonaws.com",
+        }
+    )
```
HugoPBrito, line 44: The Bedrock service family exposes several CloudTrail event sources, and only `bedrock.amazonaws.com` is listed here. A trail that scopes advanced management events to `eventSource = bedrock-agent.amazonaws.com` (control plane for Agents, Flows and KnowledgeBases) is reported as FAIL even though it does log Bedrock management events. Please expand the set, and add a parametrized test for the agent source so we do not regress this.

```suggestion
    BEDROCK_EVENT_SOURCES = frozenset(
        {
            "bedrock.amazonaws.com",
            "bedrock-agent.amazonaws.com",
            "bedrock-runtime.amazonaws.com",
            "bedrock-agent-runtime.amazonaws.com",
        }
    )
```

prowler/providers/aws/services/cloudtrail/cloudtrail_bedrock_logging_enabled/cloudtrail_bedrock_logging_enabled.metadata.json
```diff
@@ -0,0 +1,44 @@
...
+    "Software and Configuration Checks/AWS Security Best Practices"
+  ],
+  "ServiceName": "cloudtrail",
+  "SubServiceName": "",
+  "ResourceIdTemplate": "",
+  "Severity": "medium",
+  "ResourceType": "AwsCloudTrailTrail",
+  "ResourceGroup": "monitoring",
+  "Description": "**At least one actively logging CloudTrail trail** records **Amazon Bedrock API calls** through management events or advanced event selectors targeting Bedrock resources.\n\nThis captures model invocations, guardrail usage, and configuration changes for audit and security monitoring of generative AI workloads.",
+  "Risk": "Without CloudTrail logging for Bedrock, **model invocations** and configuration changes are invisible. Attackers could abuse foundation models for unauthorized content generation or data extraction (**LLM jacking**) without leaving an audit trail, undermining **forensics** and incident response.",
```
HugoPBrito: The Risk says *"model invocations and configuration changes are invisible"* and the Recommendation cites LLM jacking, but the PASS condition currently accepts a trail with only classic `EventSelectors` (`IncludeManagementEvents=True`, `ReadWriteType=All|WriteOnly`). That configuration does not capture `InvokeModel`. Bedrock data events are only emitted via advanced event selectors with `resources.type = AWS::Bedrock::Model`.

A customer can pass this check and still have zero visibility on model invocations, which is exactly the LLM jacking scenario the Risk describes. Two acceptable resolutions:

1. Tighten the check to require advanced event selectors targeting Bedrock resource types and drop the classic management PASS path.
2. Keep the current PASS semantics (control plane only) but rewrite Risk and add a Notes line stating that data plane invocation visibility relies on `bedrock_model_invocation_logging_enabled` and advanced data event selectors, and align the `RelatedTo` wording.

If you go with option 2, something like:

```suggestion
  "Risk": "Without CloudTrail logging for Bedrock control plane operations, configuration changes such as creating models, guardrails, agents or knowledge bases are invisible. Attackers could tamper with foundation model configuration without leaving an audit trail, undermining forensics and incident response. Note that CloudTrail management events do not capture `InvokeModel` and other data plane calls; for invocation level visibility pair this control with `bedrock_model_invocation_logging_enabled` or advanced data event selectors targeting Bedrock resource types.",
```

prowler/providers/aws/services/cloudtrail/cloudtrail_bedrock_logging_enabled/cloudtrail_bedrock_logging_enabled.py
```diff
@@ -0,0 +1,162 @@
...
+            has_management_events = any(
+                field.get("Field") == "eventCategory"
+                and "Management" in field.get("Equals", [])
+                for field in field_selectors
+            )
+            if has_management_events:
+                has_read_only_restriction = any(
+                    field.get("Field") == "readOnly" and field.get("Equals") == ["true"]
+                    for field in field_selectors
+                )
```
HugoPBrito: The exact match `field.get("Equals") == ["true"]` only catches the canonical form. The semantically equivalent `{"Field": "readOnly", "NotEquals": ["false"]}` (also a real way of restricting to read only) slips through and yields a false PASS. Reusing `_field_selector_matches_value` makes this robust:

```suggestion
            if has_management_events:
                read_only_selectors = [
                    field
                    for field in field_selectors
                    if field.get("Field") == "readOnly"
                ]
                has_read_only_restriction = bool(read_only_selectors) and not any(
                    self._field_selector_matches_value("false", selector)
                    for selector in read_only_selectors
                )
```

Worth adding a parametrized case with `NotEquals: ["false"]` to the test suite.

prowler/compliance/aws/c5_aws.json
```diff
@@ -3460,6 +3460,7 @@
         }
       ],
       "Checks": [
+        "cloudtrail_bedrock_logging_enabled",
         "kinesis_stream_data_retention_period",
         "cloudtrail_multi_region_enabled_logging_management_events"
```
HugoPBrito: The other `cloudtrail_*` entry in this Checks list lives two lines below. Please place this new entry next to it for grouping. The second insertion at line 3674 is already correctly placed within the `cloudtrail_*` block.

```suggestion
        "kinesis_stream_data_retention_period",
        "cloudtrail_bedrock_logging_enabled",
        "cloudtrail_multi_region_enabled_logging_management_events"
```

HugoPBrito (changes requested):
Nice work. The core logic is solid and the parametrized tests added in `fbd567a` close the `_field_selector_matches_value` coverage gap raised earlier.

Two correctness concerns blocking approval, plus one robustness item and one ordering nit.

**Blocking**

1. `BEDROCK_EVENT_SOURCES` only contains `bedrock.amazonaws.com`. Trails that scope advanced management events by `eventSource` to other Bedrock family sources will be reported as FAIL even though they capture Bedrock management events.
2. The Risk and Recommendation imply that model invocations are covered, but a trail with only classic `EventSelectors` currently PASSes, and that mode does not capture `InvokeModel`. The PASS state oversells what is actually verified.

**Non blocking**

3. `readOnly` exclusion is brittle to the equivalent restriction expressed via `NotEquals`.
5. Ordering nit in `c5_aws.json`.

prowler/providers/aws/services/cloudtrail/cloudtrail_bedrock_logging_enabled/cloudtrail_bedrock_logging_enabled.py
```diff
@@ -0,0 +1,210 @@
...
+        }
+    )
+    BEDROCK_EVENT_SOURCES = frozenset(
+        {
+            "bedrock.amazonaws.com",
+            "bedrock-agent.amazonaws.com",
+            "bedrock-runtime.amazonaws.com",
+            "bedrock-agent-runtime.amazonaws.com",
+        }
+    )
```
HugoPBrito, line 44: Wondering if Bedrock Data Automation should be in scope here. Its event sources (`bedrock-data-automation.amazonaws.com` and `bedrock-data-automation-runtime.amazonaws.com`) are part of the Bedrock service family, so a trail scoped to those alone gets reported FAIL today even though it does log Bedrock-family management events.

If the intent is to restrict this check to *generative* Bedrock only (Models/Agents/Flows/KnowledgeBases/Prompts/Guardrails), maybe worth noting that scope decision in a docstring on this set so future readers don't try to add them later.

```suggestion
    BEDROCK_EVENT_SOURCES = frozenset(
        {
            "bedrock.amazonaws.com",
            "bedrock-agent.amazonaws.com",
            "bedrock-runtime.amazonaws.com",
            "bedrock-agent-runtime.amazonaws.com",
            "bedrock-data-automation.amazonaws.com",
            "bedrock-data-automation-runtime.amazonaws.com",
        }
    )
```

prowler/providers/aws/services/cloudtrail/cloudtrail_bedrock_logging_enabled/cloudtrail_bedrock_logging_enabled.py
```diff
@@ -0,0 +1,210 @@
...
+            return "advanced_data"
+
+        return None
+
+    @staticmethod
+    def _logs_classic_management_events(event_selector: dict) -> bool:
+        """Check whether a classic selector logs Bedrock control-plane events."""
+        return event_selector.get(
+            "IncludeManagementEvents", False
+        ) and event_selector.get("ReadWriteType") in ("All", "WriteOnly")
```
HugoPBrito: Classic `EventSelectors` default to `ReadWriteType="All"` when omitted (per AWS API spec, confirmed with moto: `put_event_selectors([{IncludeManagementEvents: True}])` round-trips without filling in the default). With the current strict membership check, a trail configured with just `IncludeManagementEvents: True` and no explicit `ReadWriteType` would be reported FAIL even though AWS treats it as logging all management events.

Maybe worth defaulting to `"All"` to close that false-negative:

```suggestion
        return event_selector.get(
            "IncludeManagementEvents", False
        ) and event_selector.get("ReadWriteType", "All") in ("All", "WriteOnly")
```

prowler/providers/aws/services/cloudtrail/cloudtrail_bedrock_logging_enabled/cloudtrail_bedrock_logging_enabled.py
```diff
@@ -0,0 +1,210 @@
...
+
+    def _logs_advanced_management_events(self, field_selectors: list[dict]) -> bool:
```
HugoPBrito, line 129: One thing that might be worth hardening: both `_logs_advanced_management_events` and `_logs_advanced_bedrock_data_events` consider `eventCategory`, `readOnly`/`eventSource` and `resources.type`, but ignore other field selectors that can narrow capture and produce a false PASS. A couple of realistic cases:

- `{eventCategory: Management, eventName: {StartsWith: ["Describe", "List", "Get"]}}` reads as PASS as `advanced_management` even though zero Bedrock mutations are captured.
- `{eventCategory: Data, resources.type: AWS::Bedrock::Model, resources.ARN: {StartsWith: ["arn:aws:bedrock:us-west-2:"]}}` on a us-east-1 workload reads as PASS even though no invocations from the audited workload are captured.

If an `eventName`/`resources.ARN` selector is present, it might be safer to only treat the trail as Bedrock-logging when it doesn't *exclude* the relevant operations, or at minimum flag this configuration in the status_extended so users know the PASS doesn't mean full coverage. Not blocking, but a Bedrock-restricted trail reading as compliant when it isn't would be the worst outcome of the check.

prowler/CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ All notable changes to the **Prowler SDK** are documented in this file.
 - `bedrock_prompt_management_exists` check for AWS provider [(#10878)](https://github.com/prowler-cloud/prowler/pull/10878)
 - 8 Gmail attachment safety and spoofing protection checks for Google Workspace provider using the Cloud Identity Policy API [(#10980)](https://github.com/prowler-cloud/prowler/pull/10980)
 - `bedrock_prompt_encrypted_with_cmk` check for AWS provider [(#10905)](https://github.com/prowler-cloud/prowler/pull/10905)
+- `cloudtrail_bedrock_logging_enabled` check for AWS provider [(#10858)](https://github.com/prowler-cloud/prowler/pull/10858)
```
HugoPBrito: Heads-up for rebase: master already closed the `5.26.0` section in `02cdcb29d chore: changelog for v5.26.0 (#11105)`. This entry will need to move to a new `## [5.27.0] (Prowler v5.27.0)` section at the top, same pattern as `62ad4e5b9 docs(m365): move PIM stale alert changelog entry to 5.27.0`. Not blocking, just won't merge cleanly otherwise.

prowler/providers/aws/services/cloudtrail/cloudtrail_bedrock_logging_enabled/cloudtrail_bedrock_logging_enabled.py
danibarranqueroo: Added the Bedrock Data Automation control-plane sources to BEDROCK_EVENT_SOURCES and extended the parametrized coverage

prowler/providers/aws/services/cloudtrail/cloudtrail_bedrock_logging_enabled/cloudtrail_bedrock_logging_enabled.py
danibarranqueroo: Updated classic selector handling to honor AWS defaults for EventSelector and added a regression test for omitted ReadWriteType.

prowler/providers/aws/services/cloudtrail/cloudtrail_bedrock_logging_enabled/cloudtrail_bedrock_logging_enabled.py
danibarranqueroo: Added a note clarifying that extra advanced selector filters like eventName and resources.ARN can narrow effective coverage. I’d rather handle stricter semantic validation as a follow-up than add partial heuristics in this PR.

HugoPBrito (approved):

pedrooot (approved):
🔝

PR #10863, opened 2026-04-23
refactor(ui): unify DataTable pagination into a single callback

pfe-nazaries:
### Context

`DataTablePagination` in controlled mode exposed two separate callbacks, `onPageChange` and `onPageSizeChange`, and fired them back-to-back in the same tick when the user picked a new page size. Consumers that drove state through `router.push` (reading `useSearchParams()` inside each handler) silently reverted the size change: the second push reconstructed the URL from a stale `searchParams` snapshot taken before the first push and overwrote the new `pageSize` param.

The race surfaced most clearly in the Attack Paths scan selector, where picking a new page size from the "Rows per page" selector appeared to do nothing. The Attack Paths scans pagination fix #10864 carries a short-term `suppressNextPageResetRef` workaround in `scan-list-table.tsx` so the reported bug is unblocked; that workaround should be removed once this refactor lands.

### Description

Unify the two callbacks into a single `onPaginationChange(page, pageSize)` that always emits the full post-event state. When the user changes the page size, `DataTablePagination` now calls it once with `(1, newSize)`, so the consumer applies both values in a single update and cannot race on a stale snapshot.

- `components/ui/table/data-table-pagination.tsx`: replace the two-callback API with `onPaginationChange`; the size-change branch now emits a single atomic call
- `components/ui/table/data-table.tsx`: forward the new prop, refresh the JSDoc example
- `components/resources/table/resource-detail-content.tsx`: collapse the two handlers into one `onPaginationChange` (React state batches already prevented a visible race here, but the unified API is cleaner)
- `attack-paths/query-builder/_components/scan-list-table.tsx`: switch to `onPaginationChange`

### Related PR

- #10864 — Attack Paths scans pagination fix. Operationally independent of this PR, but once this refactor merges, the `suppressNextPageResetRef` workaround that lives in #10864's `scan-list-table.tsx` (tagged with a `TODO(#10863)`) must be removed and its split handlers replaced with the single `onPaginationChange` added here.

### Steps to review

1. `cd ui && pnpm run typecheck` — passes.
2. `cd ui && pnpm run lint:fix` — passes.
3. `cd ui && pnpm vitest run` — full suite (535 tests) passes.
4. Manual check of the two affected tables:
   - Attack Paths `/attack-paths/query-builder`: with enough scans to trigger the "Rows per page" selector (`totalEntries > 10`), change the page size. The size change must stick and the table must reset to page 1.
   - Resources drawer: open a resource with many findings, paginate + change page size; behaviour unchanged.

### Checklist

<details>

</details>

#### UI (if applicable)

### License


Alan-TheGentleman (changes requested):
One thing to fix:

1. **Controlled pagination props are still independently optional**. `onPaginationChange()` always emits a `pageSize`, but `controlledPageSize` can still be omitted, so page navigation falls back to `10` and can silently reset a non-default size. Model controlled pagination as an all-or-nothing prop shape, or at least require `controlledPageSize` whenever `onPaginationChange` is provided.

Alan-TheGentleman (approved):

PR #10864, opened 2026-04-23
fix(ui): load every Attack Paths scan before displaying the selector

### LINKED ISSUE #10852: Attack Paths show only last 5 scans (no pagination)
### Issue search

Prowler UI

AWS

### Steps to Reproduce

1. Add 11 or more AWS Cloud Providers
2. Scan them all
3. Wait until Attack Path is being scheduled

### Expected behavior

All scans are visible or there is pagination

In the table I can see only the last 5 scans, including scheduled and executing. In the top-right there is '10 Total Entries', despite having 14 AWS cloud Providers. Number do not increase when adding new Providers
Image

### Environment Resource

Helm chart in k8s

### OS used

Docker images

### Prowler version

5.24.0

### Python version

From image

### Pip version

From image

### Context

I'd like to use self-hosted Prowler for multiple providers (multi-env, multi-region deployment). Attack path feature is now effectively limited to 5 providers.

pfe-nazaries:
### Context

Fixes #10852.

The Attack Paths scan selector has been showing incomplete data on tenants with more than ten providers. The root cause is a chain of three issues on the UI side:

1. `getAttackPathScans` called `/attack-paths-scans` without any pagination params, so DRF returned only the first page (default page size 10). That partial list was then handed to `ScanListTable` as if it were the whole dataset.
2. The adapter discarded the real JSON:API pagination metadata and hardcoded `{ page: 1, pages: 1, count: data.length }`, so the table's "Total Entries" reflected the length of the first page, not the real count.
3. `ScanListTable` defaulted to a page size of 5 and paginated the 10-item partial list locally. When the real total was ≤ 10, `DataTable` hid the pagination controls (`totalEntries > 10` gate), so a tenant with six providers would see "6 Total Entries" with only five rows on screen and no way to advance.

The API side is already correct: `AttackPathsScanViewSet.list()` applies a Window function to return the latest scan per provider and paginates that dedup'd queryset through `JsonApiPageNumberPagination`. No API changes were needed.

### Description

- `actions/attack-paths/scans.ts`: `getAttackPathScans` now iterates every backend page (`page[size]=100`, up to 50 pages — same safety cap as `getAllProviders`) and returns the complete dedup'd list, mirroring the existing selector-data pattern.
- `actions/attack-paths/scans.adapter.ts`: surface the real `meta.pagination` instead of hardcoding page/pages/count.
- `types/attack-paths.ts`: extend `AttackPathScansResponse` with the JSON:API `meta.pagination` shape.
- `attack-paths/query-builder/_components/scan-list-table.tsx`:
  - Default page size is now 10 (not 5) and options are `[10, 25]`, since the list is bounded by the number of providers (typically a few dozen) and every row represents one provider's latest scan.
  - Add a short-lived `suppressNextPageResetRef` workaround so changing the page size no longer silently reverts. The proper fix is in the companion refactor #10863 (which unifies `DataTablePagination`'s controlled callbacks into a single `onPaginationChange`); once that merges, this workaround must be removed — the TODO in the file points to it.
- `actions/attack-paths/scans.test.ts` + `scans.adapter.test.ts`: new vitest specs covering the per-page aggregation, the safety cap, empty/error responses, and the adapter metadata passthrough.
- `CHANGELOG.md`: entry moved to a new `1.24.4` patch section (not 1.25.0 UNRELEASED), per team guidance that bug fixes should be backported to a patch release.

### Related PR

- #10863 — follow-up refactor that unifies `DataTable`'s controlled pagination callback. This PR does not depend on #10863: the `suppressNextPageResetRef` workaround in `scan-list-table.tsx` is enough to keep the page-size selector working on its own, so the reported issue can ship without waiting for the refactor. Once #10863 merges, the workaround must be removed and the split handlers collapsed into the single `onPaginationChange` callback it introduces.

### Steps to review

1. Unit tests: `cd ui && pnpm vitest run actions/attack-paths` — 10 tests pass (`scans.test.ts` + `scans.adapter.test.ts`).
2. Typecheck/lint: `cd ui && pnpm run typecheck && pnpm run lint:fix` — pass.
3. Manual QA (hit the real backend):
   - Tenant with 6 providers: the selector shows 6 rows on one page, "6 Total Entries", no pagination controls (≤ 10 threshold).
   - Tenant with > 10 providers (or temporarily set `REST_FRAMEWORK["PAGE_SIZE"]` to 5 in dev): real total displayed, rows beyond page 1 reachable via pagination controls.
   - Change the "Rows per page" selector from 10 to 25 — the table redraws with the new size and does not silently revert.
   - With an executing scan present, `AutoRefresh` still triggers regardless of which page you are on.

### Checklist

<details>

</details>

#### UI (if applicable)

### License

### DIFF
--- a/ui/CHANGELOG.md
+++ b/ui/CHANGELOG.md
@@ -15,6 +15,14 @@ All notable changes to the **Prowler UI** are documented in this file.
 
 ---
 
+## [1.24.4] (Prowler v5.24.4)
+
+### 🐞 Fixed
+
+- Attack Paths scan selector now lists every provider's latest scan, shows the correct total, and exposes pagination controls for tenants with more than ten providers; the UI previously fetched only the first backend page (10 items), paginated those partial results locally in pages of 5, and hid the page navigation whenever the total was ≤ 10, stranding users on a "5 of 6" view [(#10852)](https://github.com/prowler-cloud/prowler/issues/10852)
+
+---
+
 ## [1.24.2] (Prowler v5.24.2)
 
 ### 🐞 Fixed
--- a/ui/CHANGELOG.md
+++ b/ui/CHANGELOG.md
@@ -12,7 +12,7 @@ All notable changes to the **Prowler UI** are documented in this file.
 
 ---
 
-## [1.24.4] (Prowler UNRELEASED)
+## [1.24.4] (Prowler 5.24.4)
 
 ### 🐞 Fixed
 
--- a/ui/CHANGELOG.md
+++ b/ui/CHANGELOG.md
@@ -19,7 +19,7 @@ All notable changes to the **Prowler UI** are documented in this file.
 
 ### 🐞 Fixed
 
-- Attack Paths scan selector now lists every provider's latest scan, shows the correct total, and exposes pagination controls for tenants with more than ten providers; the UI previously fetched only the first backend page (10 items), paginated those partial results locally in pages of 5, and hid the page navigation whenever the total was ≤ 10, stranding users on a "5 of 6" view [(#10852)](https://github.com/prowler-cloud/prowler/issues/10852)
+- Attack Paths scan selector now lists scans from every provider with working pagination, instead of capping the list at the first ten [(#10864)](https://github.com/prowler-cloud/prowler/pull/10864)
 
 ---
 
--- a/ui/actions/attack-paths/scans.adapter.test.ts
+++ b/ui/actions/attack-paths/scans.adapter.test.ts
@@ -1,0 +1,103 @@
+import { describe, expect, it } from "vitest";
+
+import {
+  type AttackPathScan,
+  type AttackPathScansResponse,
+  SCAN_STATES,
+} from "@/types/attack-paths";
+
+import { adaptAttackPathScansResponse } from "./scans.adapter";
+
+const makeScan = (
+  id: string,
+  overrides: Partial<AttackPathScan["attributes"]> = {},
+): AttackPathScan => ({
+  type: "attack-paths-scans",
+  id,
+  attributes: {
+    state: SCAN_STATES.COMPLETED,
+    progress: 100,
+    graph_data_ready: true,
+    provider_alias: `alias-${id}`,
+    provider_type: "aws",
+    provider_uid: id,
+    inserted_at: "2026-04-23T10:00:00Z",
+    started_at: "2026-04-23T10:00:00Z",
+    completed_at: "2026-04-23T10:10:00Z",
+    duration: 600,
+    ...overrides,
+  },
+  relationships: {} as AttackPathScan["relationships"],
+});
+
+describe("adaptAttackPathScansResponse", () => {
+  it("returns an empty list when the response is undefined", () => {
+    // When
+    const result = adaptAttackPathScansResponse(undefined);
+
+    // Then
+    expect(result).toEqual({ data: [] });
+  });
+
+  it("enriches each scan with durationLabel and isRecent", () => {
+    // Given a scan that completed recently
+    const recentCompletion = new Date(
+      Date.now() - 60 * 60 * 1000,
+    ).toISOString();
+    const response: AttackPathScansResponse = {
+      data: [makeScan("s1", { completed_at: recentCompletion, duration: 90 })],
+      links: { first: "", last: "", next: null, prev: null },
+      meta: { pagination: { page: 1, pages: 1, count: 1 } },
+    };
+
+    // When
+    const result = adaptAttackPathScansResponse(response);
+
+    // Then
+    expect(result.data).toHaveLength(1);
+    const enriched = result.data[0]
+      .attributes as (typeof result.data)[0]["attributes"] & {
+      durationLabel: string | null;
+      isRecent: boolean;
+    };
+    expect(enriched.durationLabel).toBeDefined();
+    expect(enriched.isRecent).toBe(true);
+  });
+
+  it("surfaces meta.pagination values unchanged in the adapted metadata", () => {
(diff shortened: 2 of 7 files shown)


alejandrobailo (changes requested):
Nice fix overall. Three things I'd like to see addressed before this merges.

**1. Preserve partial data if a mid-loop fetch fails**

Hey, one thing worth tweaking in `scans.ts`: the whole `while` loop is wrapped in a single `try/catch`, so if the third fetch throws after pages 1 and 2 succeeded, everything we already fetched gets discarded and the function returns `undefined`, which renders the selector as "no scans available". I know `getAllProviders` follows the same shape, but the cost is higher here because every failed page wipes good data we already have in memory. Could we move the `try/catch` inside the loop and break out on error? That way users keep whatever scans we did manage to fetch, and we still log the failure for the rest.

https://github.com/prowler-cloud/prowler/blob/abee65ea6feb8ec97c2edef8f84ca36acf7ce228/ui/actions/attack-paths/scans.ts#L33-L90

**2. Add a test for the `suppressNextPageResetRef` workaround**

This is the piece of the PR I'd most like to see pinned down with a test. It works today because `DataTablePagination` fires `onPageSizeChange` and `onPageChange(1)` synchronously in the same tick, so the ref always gets consumed before it can leak. If TanStack or our own pagination component ever changes its callback ordering though, this breaks silently with no signal. The existing `scan-list-table.test.tsx` mocks the shadcn table out entirely, so the workaround ships with zero coverage. A small unit test that calls `handlePageSizeChange` followed by `handlePageChange(1)` and asserts only the size change gets pushed would give us a nice safety net, and it would also document the race the ref is meant to solve.

https://github.com/prowler-cloud/prowler/blob/abee65ea6feb8ec97c2edef8f84ca36acf7ce228/ui/app/(prowler)/attack-paths/(workflow)/query-builder/_components/scan-list-table.tsx#L247-L284

**3. Drop the dead `aggregatedResponse.meta` block**

Tiny cleanup: in `scans.ts:71-86` we build a fake `meta.pagination` with `{page: 1, pages: 1, count: allScans.length}` and run it through the adapter, but on line 86 we only return `adapted.data`, so the metadata we just built is thrown away. Since the return type is already narrowed to `{ data: AttackPathScan[] }`, we can skip the aggregated meta block entirely and hand `{ data: allScans }` straight to the adapter (the adapter already handles the undefined metadata case). Keeps the function a bit tighter and removes a misleading bit of code for the next person reading it.

https://github.com/prowler-cloud/prowler/blob/abee65ea6feb8ec97c2edef8f84ca36acf7ce228/ui/actions/attack-paths/scans.ts#L71-L90

pfe-nazaries:
> Nice fix overall. Three things I'd like to see addressed before this merges.
> 
> **1. Preserve partial data if a mid-loop fetch fails**
> 
> Hey, one thing worth tweaking in `scans.ts`: the whole `while` loop is wrapped in a single `try/catch`, so if the third fetch throws after pages 1 and 2 succeeded, everything we already fetched gets discarded and the function returns `undefined`, which renders the selector as "no scans available". I know `getAllProviders` follows the same shape, but the cost is higher here because every failed page wipes good data we already have in memory. Could we move the `try/catch` inside the loop and break out on error? That way users keep whatever scans we did manage to fetch, and we still log the failure for the rest.
> 
> File:
> 
> https://github.com/prowler-cloud/prowler/blob/abee65ea6feb8ec97c2edef8f84ca36acf7ce228/ui/actions/attack-paths/scans.ts#L33-L90
> 
> **2. Add a test for the `suppressNextPageResetRef` workaround**
> 
> This is the piece of the PR I'd most like to see pinned down with a test. It works today because `DataTablePagination` fires `onPageSizeChange` and `onPageChange(1)` synchronously in the same tick, so the ref always gets consumed before it can leak. If TanStack or our own pagination component ever changes its callback ordering though, this breaks silently with no signal. The existing `scan-list-table.test.tsx` mocks the shadcn table out entirely, so the workaround ships with zero coverage. A small unit test that calls `handlePageSizeChange` followed by `handlePageChange(1)` and asserts only the size change gets pushed would give us a nice safety net, and it would also document the race the ref is meant to solve.
> 
> File:
> 
> https://github.com/prowler-cloud/prowler/blob/abee65ea6feb8ec97c2edef8f84ca36acf7ce228/ui/app/(prowler)/attack-paths/(workflow)/query-builder/_components/scan-list-table.tsx#L247-L284
> 
> **3. Drop the dead `aggregatedResponse.meta` block**
> 
> Tiny cleanup: in `scans.ts:71-86` we build a fake `meta.pagination` with `{page: 1, pages: 1, count: allScans.length}` and run it through the adapter, but on line 86 we only return `adapted.data`, so the metadata we just built is thrown away. Since the return type is already narrowed to `{ data: AttackPathScan[] }`, we can skip the aggregated meta block entirely and hand `{ data: allScans }` straight to the adapter (the adapter already handles the undefined metadata case). Keeps the function a bit tighter and removes a misleading bit of code for the next person reading it.
> 
> File:
> 
> https://github.com/prowler-cloud/prowler/blob/abee65ea6feb8ec97c2edef8f84ca36acf7ce228/ui/actions/attack-paths/scans.ts#L71-L90

1 and 3 fixed, i would suggest to skip 2 due there is already a PR to fix the workaround. I decided to split into a second PR to not mix concerns https://github.com/prowler-cloud/prowler/pull/10863

ui/CHANGELOG.md
jfagoagas:
```suggestion
## [1.24.4] (Prowler 5.24.4)
```

alejandrobailo (approved):
LGTM

PR #10874, opened 2026-04-23
fix(api): prevent /tmp saturation from compliance report generation

AdriiiPRodri:
### Context

Workers in the `scan-reports` queue were accumulating stale artifacts under `/tmp/prowler_api_output`, saturating disk on long-lived hosts. The investigation identified three contributors:

1. **Cleanup defect in `generate_compliance_reports`**: unsupported compliance frameworks for a given provider were leaving a `results[<framework>] = {"upload": False, "path": ""}` entry that blocked the final `rmtree`, so the scan directory was never removed.
2. **Expected local retention on upload failure**: `_upload_to_s3` returns `None` when the bucket is not configured, and artifacts are intentionally kept locally in that case.
3. **Compression peak**: `_compress_output_files` writes the archive while the source files still exist, producing short-lived disk peaks.

This PR fixes root cause 1 and adds bounded, safe cleanup coverage for the retention pattern of cause 2. Cause 3 is documented as residual risk.

### Description

Changes in `api/src/backend/tasks/jobs/report.py` and `api/src/backend/tasks/tasks.py`:

**Bug fix: unsupported frameworks no longer block cleanup**
- `generate_compliance_reports` now tracks the frameworks that actually ran per scan (`generated_report_keys`). The final `rmtree` only waits for those entries to report a successful upload, instead of inspecting every provider-incompatible placeholder entry.
- Output directories are created lazily, only for enabled and provider-supported frameworks. The five blind `_generate_compliance_output_directory` calls at the top of the function are gone.
- CIS variant discovery and its output directory creation were moved into the same `try` block so that CIS-specific failures are captured in `results["cis"]` consistently.

**New opportunistic stale cleanup**
- Helper `_cleanup_stale_tmp_output_directories` invoked at task start in both `generate_compliance_reports` and `generate_outputs_task`, scoped to the current tenant and excluding the active scan.
- Hard-coded safe root: `/tmp/prowler_api_output`. Any call with a different root is rejected.
- Only deletes scan directories older than 48h (`STALE_TMP_OUTPUT_MAX_AGE_HOURS`).
- Bounded to 50 deletions per run (`STALE_TMP_OUTPUT_MAX_DELETIONS_PER_RUN`).
- Host-wide throttle of once per hour via a single shared lock file (`.stale_tmp_cleanup.lock`) using `fcntl.flock` with `LOCK_NB`, so concurrent workers don't pile up cleanup passes.
- Symlinks are skipped; only directories strictly under the safe root are considered.
- Scan directories are protected when either the scan is still `EXECUTING` or its `output_location` points to a local path under the candidate directory.
- Scan metadata is preloaded per tenant with a single `id__in` query, falling back to per-scan lookup only when the batch fails or the directory name is not a UUID.
- Top-level `try/except` with `exc_info=True` guarantees cleanup failures never abort report generation; both call sites also wrap the invocation in a defensive `try/except` for extra safety.

### Steps to review

1. Read `_cleanup_stale_tmp_output_directories`, `_should_run_stale_cleanup`, `_is_scan_metadata_protected`, and `_is_scan_directory_protected` in `api/src/backend/tasks/jobs/report.py`.
2. Verify the new `generated_report_keys` / `output_paths` flow in `generate_compliance_reports` and the guard `if all_uploaded and out_dir:` around the final `rmtree`.
3. Verify the call site and `try/except` wrapper in `generate_outputs_task` (`api/src/backend/tasks/tasks.py`).
4. Run the new tests:
   ```bash
   cd api
   poetry run pytest src/backend/tasks/tests/test_reports.py src/backend/tasks/tests/test_tasks.py -q
   ```
   Expected: all green, including:
   - `TestCleanupStaleTmpOutputDirectories` (TTL, exclude, max deletions, non-safe-root, symlinks, internal exception swallow).
   - `TestStaleCleanupProtectionHelpers` (throttle, `LOCK_NB` busy, EXECUTING, local output, S3 output).
   - `test_cleanup_runs_when_supported_reports_upload_successfully` / `test_cleanup_skipped_when_supported_upload_fails`.
   - `test_cleanup_exception_does_not_break_no_findings_flow` (both in reports and tasks).
   - `test_cis_output_directory_failure_is_captured`.
5. Operational checks after deploy (from any affected host):
   ```bash
   du -sh /tmp/prowler_api_output
   find /tmp/prowler_api_output -mindepth 2 -maxdepth 2 -type d -mmin +2880 | wc -l
   ```
   Logs to grep: `Deleted ... stale tmp output directories ...`, `Stale tmp cleanup hit deletion limit ...`, `S3 upload failed:`, `... saved locally at ...`.

### Residual risk

- Cleanup is opportunistic: a tenant with no reporting traffic will not see its own stale dirs drained until it runs reports again.
- Locally retained artifacts (failed/unconfigured upload) are only reclaimed after the 48h TTL.
- Compression peak is not addressed in this PR; zip output and source files still coexist briefly during `_compress_output_files`.

### Checklist

<details>

</details>

#### SDK/CLI

#### UI
Not applicable.

#### API
- [ ] Endpoint response output (if applicable) - N/A, no endpoint changes
- [ ] EXPLAIN ANALYZE output for new/modified queries or indexes (if applicable) - N/A, no new indexes; only an additional `Scan.all_objects.filter(tenant_id=..., id__in=[...]).values_list(...)` batched per tenant during cleanup
- [ ] Performance test results (if applicable) - N/A, cleanup is bounded (<=50 deletions, <=1/hour/host)
- [x] Any other relevant evidence of the implementation (if applicable) - see "Steps to review"
- [x] Verify if API specs need to be regenerated. - Not needed
- [x] Check if version updates are required (e.g., specs, Poetry, etc.). - Not needed

### License

### DIFF
--- a/api/src/backend/tasks/jobs/report.py
+++ b/api/src/backend/tasks/jobs/report.py
@@ -1,9 +1,13 @@
 import gc
+import os
 import re
+import time
 from collections.abc import Iterable
 from pathlib import Path
 from shutil import rmtree
+from uuid import UUID
 
+import fcntl
 from celery.utils.log import get_task_logger
 from config.django.base import DJANGO_TMP_OUTPUT_DIRECTORY
 from tasks.jobs.export import _generate_compliance_output_directory, _upload_to_s3
@@ -20,11 +24,16 @@ from tasks.jobs.threatscore_utils import _aggregate_requirement_statistics_from_
 
 from api.db_router import READ_REPLICA_ALIAS
 from api.db_utils import rls_transaction
-from api.models import Provider, ScanSummary, ThreatScoreSnapshot
+from api.models import Provider, Scan, ScanSummary, StateChoices, ThreatScoreSnapshot
 from prowler.lib.check.compliance_models import Compliance
 from prowler.lib.outputs.finding import Finding as FindingOutput
 
 logger = get_task_logger(__name__)
+STALE_TMP_OUTPUT_MAX_AGE_HOURS = 48
+STALE_TMP_OUTPUT_MAX_DELETIONS_PER_RUN = 50
+STALE_TMP_OUTPUT_THROTTLE_SECONDS = 60 * 60
+STALE_TMP_OUTPUT_LOCK_FILE_NAME = ".stale_tmp_cleanup.lock"
+STALE_TMP_OUTPUT_SAFE_ROOT = Path("/tmp/prowler_api_output").resolve()
 
 # Matches CIS compliance_ids like "cis_1.4_aws", "cis_5.0_azure",
 # "cis_1.10_kubernetes", "cis_3.0.1_aws". Requires at least one dotted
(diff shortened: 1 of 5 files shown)


api/src/backend/tasks/jobs/report.py
```diff
@@ -18,13 +22,18 @@
...
+from api.models import Provider, Scan, ScanSummary, StateChoices, ThreatScoreSnapshot
 from prowler.lib.check.compliance_models import Compliance
 from prowler.lib.outputs.finding import Finding as FindingOutput
 
 logger = get_task_logger(__name__)
+STALE_TMP_OUTPUT_MAX_AGE_HOURS = 48
+STALE_TMP_OUTPUT_MAX_DELETIONS_PER_RUN = 50
+STALE_TMP_OUTPUT_THROTTLE_SECONDS = 60 * 60
+STALE_TMP_OUTPUT_LOCK_FILE_NAME = ".stale_tmp_cleanup.lock"
+STALE_TMP_OUTPUT_SAFE_ROOT = Path("/tmp/prowler_api_output").resolve()
```
josema-xyz: As this is hardcoded, would anyone running with a custom `DJANGO_TMP_OUTPUT_DIRECTORY` silently gets no cleanup, just a warning log?

api/src/backend/tasks/tasks.py
```diff
@@ -440,6 +444,20 @@ def generate_outputs_task(scan_id: str, provider_id: str, tenant_id: str):
         scan_id (str): The scan identifier.
         provider_id (str): The provider_id id to be used in generating outputs.
     """
+    try:
+        _cleanup_stale_tmp_output_directories(
+            DJANGO_TMP_OUTPUT_DIRECTORY,
+            max_age_hours=STALE_TMP_OUTPUT_MAX_AGE_HOURS,
+            exclude_scan=(tenant_id, scan_id),
+            tenant_scope=tenant_id,
+        )
```
josema-xyz, line 452: If this only cleans one tenant per hour leftover files from the other tenants just keep piling up. Can we change it so each run cleans them all?

api/src/backend/tasks/jobs/report.py
AdriiiPRodri: Yes, I'm going to change it. This was originally done this way due to issues with this path on other platforms, but I agree it's better to use the defined path with some checks beforehand. Thanks!

api/src/backend/tasks/tasks.py
AdriiiPRodri: This was intentional since this shouldn't really ever be necessary, but it's true that if we keep limiting by hour and not by tenant, it’s still cheap to check this, so I'll remove the restriction. Thanks!

josema-xyz (dismissed):

josema-xyz (approved):

PR #10878, opened 2026-04-24
feat(aws): add bedrock_prompt_management_exists security check

danibarranqueroo:
### Context

Without centralized Prompt Management, prompts used with Bedrock foundation models are scattered across applications with no oversight, increasing the risk of prompt injection attacks, inconsistent model behavior, and a lack of auditability over instructions sent to models. Attackers or malicious insiders could exploit ungoverned prompts to manipulate model outputs, bypass safety controls, or exfiltrate sensitive information through crafted inputs.

### Description

This check verifies whether Amazon Bedrock Prompt Management is actively being used by checking for the existence of managed prompts in each enabled region. It reports a **PASS** if at least one managed prompt exists in the region, indicating that prompt governance and standardization practices are in place, and a **FAIL** if no managed prompts are found. The recommended remediation is to adopt Bedrock Prompt Management to centralize prompt definitions, enforce versioning, and apply least-privilege access controls to restrict who can create or modify prompts.

### Steps to review

1. Review the check implementation at `prowler/providers/aws/services/bedrock/bedrock_prompt_management_exists/`
2. Review the metadata file for correct severity, remediation, and compliance mappings
3. Review compliance framework mappings in `prowler/compliance/aws/` to ensure the check is correctly mapped to relevant requirements
4. Run the check tests: `poetry run pytest tests/providers/aws/services/bedrock/bedrock_prompt_management_exists/ -v`
5. **Run the check against a real environment** (if possible):
   ```bash
   prowler aws --check bedrock_prompt_management_exists
   ```

### Checklist

<details>

<summary><b>Community Checklist</b></summary>

- [x] This feature/issue is listed in [here](https://github.com/prowler-cloud/prowler/issues?q=sort%3Aupdated-desc+is%3Aissue+is%3Aopen) or roadmap.prowler.com
- [x] Is it assigned to me, if not, request it via the issue/feature in [here](https://github.com/prowler-cloud/prowler/issues?q=sort%3Aupdated-desc+is%3Aissue+is%3Aopen) or [Prowler Community Slack](goto.prowler.com/slack)

</details>

- [x] Review if the code is being covered by tests.
- [x] Review if code is being documented following this specification https://github.com/google/styleguide/blob/gh-pages/pyguide.md#38-comments-and-docstrings
- [ ] Review if backport is needed.
- [ ] Review if is needed to change the [Readme.md](https://github.com/prowler-cloud/prowler/blob/master/README.md)
[...]

PR #10905, opened 2026-04-28
feat(aws): add bedrock_prompt_encrypted_with_cmk security check

danibarranqueroo:
### Context

Amazon Bedrock prompts managed via Prompt Management can contain sensitive instructions, business logic, and references to downstream tooling. When a prompt is left with the AWS-owned default key, the tenant has no control over the key policy that governs decrypt permissions, no control over rotation cadence, and no auditable trail tied to a customer-managed KMS key. Multiple frameworks (ISO 27001 A.8.24, NIST CSF PR.DS-1, KISA-ISMS-P 2.7.2) require customer-managed cryptographic keys for sensitive data at rest, so prompts encrypted only with the default key create both a control gap and a compliance gap.

### Description

This check verifies that each Amazon Bedrock prompt is encrypted at rest with a **customer-managed KMS key (CMK)**. For every prompt collected from `ListPrompts`, the service layer calls `GetPrompt` and reads `customerEncryptionKeyArn`. The check reports a **PASS** when that ARN is present and a **FAIL** when it is absent (i.e. the prompt falls back to the AWS-owned key). The recommended remediation is to call `bedrock-agent update-prompt --customer-encryption-key-arn <kms_key_arn>` against an org-controlled CMK whose key policy is scoped to the principals that legitimately need to author or invoke prompts.

This PR also extends `BedrockAgent` in `bedrock_service.py` with a `_get_prompt` threading call and adds `customer_encryption_key_arn` to the `Prompt` model. The `_list_prompts` collection is **kept aligned with #10878** (same shape, same defensive accessors, same `prompt_scanned_regions` tracking, no `audit_resources` filtering) so the two PRs compose cleanly: regardless of merge order, the second one only contributes its additive piece without conflicting on the shared scaffolding.

### Note on previous direction

This branch originally implemented `bedrock_prompt_have_multiple_variants`. That check was dropped after review — the AWS [`CreatePrompt`(https://docs.aws.amazon.com/bedrock/latest/APIReference/API_agent_CreatePrompt.html) and
[`GetPrompt`](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_agent_GetPrompt.html) APIs both cap the `variants` array at "Maximum 1 item", which made the original PASS condition (`len(variants) > 1`) unreachable in real AWS and the resilience-control compliance mappings unsupportable. Pivoting to CMK encryption preserves the per-prompt scaffolding while moving the check onto an attribute the API actually persists, with compliance mappings that align cleanly to data-at-rest controls.

### Steps to review

1. Review the check implementation at `prowler/providers/aws/services/bedrock/bedrock_prompt_encrypted_with_cmk/`
3. Review the changes to `prowler/providers/aws/services/bedrock/bedrock_service.py` — confirm that `_list_prompts` is byte-identical to the version proposed in
[#10878](https://github.com/prowler-cloud/prowler/pull/10878) and that `_get_prompt` + `customer_encryption_key_arn` are purely additive
4. Review compliance mappings: ISO 27001 A.8.24 (Use of Cryptography), NIST CSF PR.DS-1 (Data-at-rest is protected), KISA-ISMS-P 2.7.2 (Cryptographic Key Management) — and confirm the
resilience-control mappings from the previous direction are removed
5. Run the check tests: `poetry run pytest tests/providers/aws/services/bedrock/bedrock_prompt_encrypted_with_cmk/ -v`
6. **Run the check against a real environment** (if possible):
 ```bash
 prowler aws --check bedrock_prompt_encrypted_with_cmk
```

### Checklist

<details>

</details>

#### SDK/CLI

### License

### CODE REVIEW

File: prowler/providers/aws/services/bedrock/bedrock_service.py [hunk 1]
```diff
@@ -167,7 +174,47 @@ def _list_agents(self, regional_client):
                 f"{regional_client.region} -- {error.__class__.__name__}[{error.__traceback__.tb_lineno}]: {error}"
             )
 
+    def _list_prompts(self, regional_client):
```
lydiavilchez, line 177: _list_prompts doesn't filter by audit_resources like _list_agents does, so --resource-arn scoping won't work for this check.

prowler/providers/aws/services/bedrock/bedrock_prompt_encrypted_with_cmk/bedrock_prompt_encrypted_with_cmk.metadata.json
```diff
@@ -0,0 +1,41 @@
+{
+  "Provider": "aws",
+  "CheckID": "bedrock_prompt_encrypted_with_cmk",
+  "CheckTitle": "Bedrock prompt is encrypted at rest with a customer-managed KMS key",
```
lydiavilchez: Same as the other PRs: existing Bedrock checks mostly use "Amazon Bedrock" in the title.

prowler/providers/aws/services/bedrock/bedrock_prompt_encrypted_with_cmk/bedrock_prompt_encrypted_with_cmk.metadata.json
```diff
@@ -0,0 +1,41 @@
...
+      "Text": "Encrypt every Bedrock prompt with a **customer-managed KMS key** to retain control over key access, rotation, and lifecycle. Restrict the key policy to the principals that legitimately need to author or invoke prompts, and align rotation cadence with your encryption policy.",
+      "Url": "https://hub.prowler.com/check/bedrock_prompt_encrypted_with_cmk"
+    }
+  },
+  "Categories": [
+    "gen-ai",
+    "encryption"
+  ],
+  "DependsOn": [],
+  "RelatedTo": [],
```
lydiavilchez: RelatedTo is empty, could reference `bedrock_prompt_management_exists`

prowler/compliance/aws/ccc_aws.json
```diff
@@ -6852,7 +6852,8 @@
       ],
       "Checks": [
         "bedrock_model_invocation_logging_enabled",
-        "bedrock_model_invocation_logs_encryption_enabled"
+        "bedrock_model_invocation_logs_encryption_enabled",
+        "bedrock_prompt_encrypted_with_cmk"
```
HugoPBrito: I think some of these compliance mappings are a bit broader than what the check actually proves.

`bedrock_prompt_encrypted_with_cmk` validates encryption at rest with a customer-managed KMS key, so it clearly fits controls around data-at-rest protection and cryptographic key management. But mappings like CCC citation/source traceability, NIS2 monitoring/logging, supplier practices, and some C5 logging/confidential-computing controls feel indirect.

Would it be better to keep only the controls that this check directly evidences, or add a clear rationale for the broader mappings? Otherwise this could add noise to compliance reports.

prowler/providers/aws/services/bedrock/bedrock_prompt_encrypted_with_cmk/bedrock_prompt_encrypted_with_cmk.metadata.json
```diff
@@ -0,0 +1,43 @@
...
+  "Risk": "A prompt encrypted only with the AWS-owned default key offers limited tenant control over key access and lifecycle: no customer KMS key policy to govern decrypt permissions, no control over rotation cadence or scheduled deletion, and gaps against frameworks (ISO 27001 A.8.24, NIST CSF PR.DS, KISA-ISMS-P 2.7.2) that require customer-managed keys for sensitive data at rest.",
+  "RelatedUrl": "",
+  "AdditionalURLs": [
+    "https://docs.aws.amazon.com/bedrock/latest/userguide/prompt-management.html",
+    "https://docs.aws.amazon.com/bedrock/latest/APIReference/API_agent_CreatePrompt.html",
+    "https://docs.aws.amazon.com/bedrock/latest/APIReference/API_agent_UpdatePrompt.html"
+  ],
+  "Remediation": {
+    "Code": {
+      "CLI": "aws bedrock-agent update-prompt --prompt-identifier <prompt_id> --name <prompt_name> --customer-encryption-key-arn <kms_key_arn>",
```
HugoPBrito: I think the remediation CLI should be a bit more explicit.

The AWS `UpdatePrompt` docs say to include both the fields you want to keep and the fields you want to replace. So the current command with only `--name` and `--customer-encryption-key-arn` may be incomplete for prompts that already have fields like `defaultVariant`, `variants`, or `description`.

Would it be safer to mention that users should retrieve the current prompt first and include the existing fields they want to preserve when calling `update-prompt`?

HugoPBrito (changes requested):
I think the check is going in the right direction, but I would adjust the compliance mappings and the remediation text before merging.

HugoPBrito (approved):

pedrooot (approved):

PR #11000, opened 2026-05-05
feat(aws): add iam_user_access_not_stale_to_sagemaker security check

HugoPBrito:
### Context

Stale SageMaker permissions widen the blast radius of a credential compromise. An attacker who gains access to an IAM user with unused SageMaker permissions can access ML training data, models, endpoints, and notebooks — all without triggering expected usage patterns. This check helps enforce least privilege by detecting IAM users that hold SageMaker permissions but have not actually used the service recently.

### Description

This check evaluates each IAM user that has SageMaker service permissions by inspecting IAM Access Advisor (service last accessed) data for the `sagemaker` namespace. A user **passes** if they have accessed SageMaker within the configured threshold (default 90 days). A user **fails** if their last SageMaker access exceeds the threshold or if they have never accessed the service. The recommended remediation is to review the Access Advisor tab and remove or scope down any SageMaker policies that are no longer actively used.

### Steps to review

1. Review the check implementation at `prowler/providers/aws/services/iam/iam_user_access_not_stale_to_sagemaker/`
3. Review compliance framework mappings in `prowler/compliance/aws/` to ensure the check is correctly mapped to relevant requirements
4. Run the check tests: `poetry run pytest tests/providers/aws/services/iam/iam_user_access_not_stale_to_sagemaker/ -v`
   ```bash
   prowler aws --check iam_user_access_not_stale_to_sagemaker
   ```

### Checklist

<details>

</details>

#### SDK/CLI

### License

### DIFF
--- a/prowler/config/config.yaml
+++ b/prowler/config/config.yaml
@@ -26,7 +26,7 @@ aws:
   max_unused_access_keys_days: 45
   # aws.iam_user_console_access_unused --> CIS recommends 45 days
   max_console_access_days: 45
-  # aws.iam_user_access_not_stale_to_sagemaker --> default 90 days; flag IAM users whose last SageMaker access exceeds this threshold
+  # aws.iam_user_access_not_stale_to_sagemaker --> default 90 days
   max_unused_sagemaker_access_days: 90
 
   # AWS EC2 Configuration
--- a/prowler/config/config.yaml
+++ b/prowler/config/config.yaml
@@ -26,6 +26,8 @@ aws:
   max_unused_access_keys_days: 45
   # aws.iam_user_console_access_unused --> CIS recommends 45 days
   max_console_access_days: 45
+  # aws.iam_user_access_not_stale_to_sagemaker --> default 90 days; flag IAM users whose last SageMaker access exceeds this threshold
+  max_unused_sagemaker_access_days: 90
 
   # AWS EC2 Configuration
   # aws.ec2_elastic_ip_shodan
--- a/prowler/providers/aws/services/iam/iam_user_access_not_stale_to_sagemaker/iam_user_access_not_stale_to_sagemaker.py
+++ b/prowler/providers/aws/services/iam/iam_user_access_not_stale_to_sagemaker/iam_user_access_not_stale_to_sagemaker.py
@@ -33,34 +33,22 @@ class iam_user_access_not_stale_to_sagemaker(Check):
             "max_unused_sagemaker_access_days", 90
         )
 
-        for (
-            user_data,
-            last_accessed_services,
-        ) in iam_client.last_accessed_services.items():
-            user_name = user_data[0]
-            user_arn = user_data[1]
-
+        for user in iam_client.users:
+            last_accessed_services = iam_client.last_accessed_services.get(
+                (user.name, user.arn), []
+            )
             sagemaker_service = find_sagemaker_service(last_accessed_services)
             if sagemaker_service is None:
                 continue
 
-            report = Check_Report_AWS(
-                metadata=self.metadata(),
-                resource={"name": user_name, "arn": user_arn},
-            )
-            report.resource_id = user_name
-            report.resource_arn = user_arn
+            report = Check_Report_AWS(metadata=self.metadata(), resource=user)
             report.region = iam_client.region
-            for iam_user in iam_client.users:
-                if iam_user.arn == user_arn:
-                    report.resource_tags = iam_user.tags
-                    break
 
             evaluate_sagemaker_staleness(
                 report,
                 sagemaker_service,
                 max_unused_sagemaker_days,
-                user_name,
+                user.name,
                 "User",
             )
             findings.append(report)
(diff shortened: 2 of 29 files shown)


prowler/config/config.yaml
danibarranqueroo, line 1: If we make this check configurable, we should also modify `tests/config/fixtures/config.yaml` and add it into the docs `docs/tutorials/configuration_file.md`.

tests/providers/aws/services/iam/iam_user_access_not_stale_to_sagemaker/__init__.py
danibarranqueroo, line 1: Remove this file please.

prowler/providers/aws/services/iam/iam_user_access_not_stale_to_sagemaker/iam_user_access_not_stale_to_sagemaker.py
```diff
@@ -0,0 +1,68 @@
...
+            user_name = user_data[0]
+            user_arn = user_data[1]
+
+            sagemaker_service = find_sagemaker_service(last_accessed_services)
+            if sagemaker_service is None:
+                continue
+
+            report = Check_Report_AWS(
+                metadata=self.metadata(),
+                resource={"name": user_name, "arn": user_arn},
```
danibarranqueroo: This resource is strange for me, could you explain why?

prowler/providers/aws/services/iam/lib/policy.py
```diff
@@ -1165,3 +1165,47 @@ def evaluate_bedrock_staleness(
...
+
+def find_sagemaker_service(last_accessed_services: list[dict]) -> Optional[dict]:
+    """Return the SageMaker entry from a service last accessed list."""
+    for service in last_accessed_services:
+        if service.get("ServiceNamespace") == "sagemaker":
+            return service
+    return None
+
+
+def evaluate_sagemaker_staleness(
```
danibarranqueroo: I think that this function should be part of the check, I don't think that this could be re-use in any other sagemaker check and I don't like the idea of modifying the report in a different file, let me know your opinion.

danibarranqueroo (changes requested):
Good job! Please review my suggestions 💯

prowler/providers/aws/services/iam/iam_user_access_not_stale_to_sagemaker/iam_user_access_not_stale_to_sagemaker.py
HugoPBrito: Yeah it was strange, I was passing just the minimum I needed because I'm only indexing by `(name, arn)` from `last_accessed_services`, but I already adapted it to follow the same pattern as the rest of the IAM user checks — now I iterate `iam_client.users` and pass the `User` object directly to `Check_Report_AWS`, so `resource_id`/`resource_arn`/`resource_tags` are filled automatically. Done in the latest commit.

prowler/providers/aws/services/iam/lib/policy.py
HugoPBrito: Makes sense, the report mutation living in a different file is a fair point. The only thing is that `evaluate_bedrock_staleness` / `find_bedrock_service` are already in `lib/policy.py` following the same pattern, so if we move the sagemaker helpers we should probably move the bedrock ones too to keep both checks consistent. Happy to do both in this PR if you prefer that — let me know.

tests/providers/aws/services/iam/iam_user_access_not_stale_to_sagemaker/__init__.py
HugoPBrito: Done!

prowler/config/config.yaml
HugoPBrito: Done!

prowler/providers/aws/services/iam/lib/policy.py
danibarranqueroo: Yes please, do both changes on this PR, thanks!

HugoPBrito: Heads-up on a small judgment call in this commit: I moved the `find_*_service` companions (`find_sagemaker_service`, `find_bedrock_service`) alongside the `evaluate_*_staleness` functions, even though your comment was strictly about the report-mutating `evaluate_*` ones.

Reasoning: the `find_*` helpers are currently consumed only by these three checks (sagemaker user, bedrock user, bedrock role), so once the `evaluate_*` ones live inside the check they would be the only callers of the corresponding `find_*` in `lib/policy.py`. Leaving the pair split (find_* in `lib/`, evaluate_* inline) felt inconsistent with no upside today.

If in the future another sagemaker/bedrock check needs the same lookup, we can absolutely promote the `find_*` helper back out to `lib/policy.py` (or a more specific module) — it's a cheap reversal since they're pure functions with no report coupling.

Happy to revert just the `find_*` part if you'd rather keep them shared from the start.

danibarranqueroo (approved):

pedrooot (approved):
🥇

PR #11075, opened 2026-05-07
fix(api): bound Celery worker concurrency to a configurable default

### LINKED ISSUE #10968: Celery workers OOM on hosts with many CPUs — `--concurrency` is unbounded by default
### Steps to Reproduce

Deploy `prowler-api` `worker` mode (e.g., `prowlercloud/prowler-api:5.22.0`, image entrypoint `docker-entrypoint.sh worker`) on a Kubernetes node with many CPUs (typical AKS / EKS node sizes: 16, 20, 32 cores), with no Linux CPU pinning on the pod.

```yaml
resources:
  requests: { memory: "1Gi" }
  limits:   { memory: "4Gi" }
```

### Expected behavior

A bounded number of Celery prefork workers, with predictable per-pod memory consumption regardless of host CPU count.

### Actual behavior

`api/docker-entrypoint.sh` runs:

```sh
poetry run python -m celery -A config.celery worker \
  -Q celery,scans,scan-reports,deletion,backfill,overview,integrations,compliance,attack-paths-scans \
  -E --max-tasks-per-child 1
```

`--concurrency` is not set on the command line, and `config/celery.py` does `celery_app.config_from_object("django.conf:settings", namespace="CELERY")` but Django settings don't define `CELERY_WORKER_CONCURRENCY` either, so Celery falls back to its default (`os.cpu_count()` of the host).

On a 20-core AKS node this spawns **20 prefork worker processes**. Each loads the Prowler virtualenv plus the Azure / AWS / GCP / M365 SDK clients — about 170 MiB resident at idle. The pod sits at **~3.4 GiB resident with zero scans running**, and any scan load tips it over the 4 GiB limit:

```
$ kubectl top pods
NAME                              CPU(cores)   MEMORY(bytes)
prowler-worker-6b77c7b8f8-2xlp4   2m           3395Mi
prowler-worker-6b77c7b8f8-gkpc4   3m           3393Mi

$ kubectl get pods -l app=prowler-worker -o jsonpath='...'
prowler-worker-6b77c7b8f8-2xlp4  restartCount=4  lastTerminated=OOMKilled  exitCode=137
prowler-worker-6b77c7b8f8-gkpc4  restartCount=8  lastTerminated=OOMKilled  exitCode=137
```

The same image on a 4-core developer machine would idle below 1 GiB and never hit the limit, which is probably why this hasn't surfaced in local testing — the symptom is host-dependent and only shows up on larger Kubernetes nodes.

### A note on `--max-tasks-per-child 1`

The existing `--max-tasks-per-child 1` is a great safeguard against per-task memory leaks — each child is recycled after one task. It doesn't, however, bound the number of concurrent child slots. With concurrency = 20 we still keep 20 long-lived child processes alive (recycled in place after each task), each carrying the SDK working set.

### Suggested fix

Either of the two paths would make this configurable for operators without forking:

1. **Pass `--concurrency` from an env var in the entrypoint** (smallest change):
   ```sh
   poetry run python -m celery -A config.celery worker \
     -Q ... -E --max-tasks-per-child 1 \
     --concurrency "${CELERY_WORKER_CONCURRENCY:-4}"
   ```
2. **Set the Django setting** (cleaner, also picked up by `config_from_object`):
   ```python
   # config/django/base.py
   CELERY_WORKER_CONCURRENCY = env.int("CELERY_WORKER_CONCURRENCY", default=4)
   ```

A default of 4 feels like a reasonable starting point given the per-process SDK memory footprint; operators who need more can raise it explicitly. Falling through to host `nproc` is convenient on dev machines, but it's not the best default for Kubernetes deployments where pods routinely see many more CPUs than they're sized for in memory.

Happy to send a PR if the maintainers would find it useful.

Helm-style Kubernetes manifests on AKS, image `prowlercloud/prowler-api:5.22.0`.

### Environment Resource

AKS, 20-core nodes (Standard_D20s_v5).

### OS used

Container base — Prowler image; Linux node.

### Prowler version

api 5.22.0 — also present on `main` at HEAD (`api/docker-entrypoint.sh` and `api/src/backend/config/celery.py`).

b-abderrahmane:
### Context

Fix #10968.

Without an explicit `worker_concurrency` setting, Celery falls back to `os.cpu_count()` of the host. On Kubernetes nodes with many CPUs (16+) the worker container spawns one prefork child per CPU, each loading the full SDK working set, and routinely OOMKills under typical 4 GiB pod limits even when idle. The existing `--max-tasks-per-child 1` recycles each child after one task — it bounds per-task leaks but not the number of concurrent slots.

### Description

Adds `CELERY_WORKER_CONCURRENCY = env.int("DJANGO_CELERY_WORKER_CONCURRENCY", default=4)` in `config/settings/celery.py`. The Celery app already loads Django settings via `config_from_object("django.conf:settings", namespace="CELERY")`, so the setting flows through to `app.conf.worker_concurrency` and the prefork pool size with no entrypoint changes.

A default of `4` matches the per-process SDK memory footprint observed in typical deployments; operators size up via `DJANGO_CELERY_WORKER_CONCURRENCY` (also added to `api/.env.example` next to the sibling `DJANGO_CELERY_DEADLOCK_ATTEMPTS`).

Changes:

- `api/src/backend/config/settings/celery.py` — one Django setting, matching the existing `env.int("DJANGO_*", default=...)` convention used by `DJANGO_CELERY_DEADLOCK_ATTEMPTS`.
- `api/.env.example` — document the new env var alongside its siblings.
- `api/CHANGELOG.md` — entry under `[1.29.0] (Prowler UNRELEASED)` → `🚀 Added`.

### Steps to review

1. Read the 1-line settings change and the changelog/env-example additions — three files, +7 lines.
2. Confirm the setting name/namespace is consistent with existing siblings:
   - `CELERY_DEADLOCK_ATTEMPTS = env.int("DJANGO_CELERY_DEADLOCK_ATTEMPTS", default=5)` (existing)
   - `CELERY_WORKER_CONCURRENCY = env.int("DJANGO_CELERY_WORKER_CONCURRENCY", default=4)` (this PR)
3. Optionally reproduce the validation below in any K8s deployment.

### Evidence

Validated end-to-end on a 20-core AKS node with the worker image rebuilt from this branch (`prowler-api`, dockerized from `api/`). Worker Deployment patched to use the test image; same pod spec (`requests.memory=1Gi`, `limits.memory=4Gi`).

**Before vs after**

| Metric                                    | Before (master, no setting) | After (this PR, default `4`) |
| ----------------------------------------- | --------------------------: | ---------------------------: |
| Django `settings.CELERY_WORKER_CONCURRENCY` | (unset → falls through)   | `4`                          |
| `app.conf.worker_concurrency`             | `os.cpu_count()` = `20`     | `4`                          |
| Prefork children per pod                  | `20`                        | **`4`**                      |
| Idle resident memory                      | `~3.4 GiB`                  | **`~880 MiB`**               |
| Restart count (over 26h soak)             | `4` and `8` (OOMKilled)     | `0`                          |

**Process tree (after)** — 1 parent + 4 ForkPool children:

```
$ kubectl exec prowler-worker-7c56d9dd6d-4lcdt -- ls /proc/[0-9]*/comm | grep -E celery
pid=10  python -m celery -A config.celery worker -n c5...
pid=38  python -m celery -A config.celery worker -n c5...   # ForkPoolWorker-1
pid=39  python -m celery -A config.celery worker -n c5...   # ForkPoolWorker-2
pid=40  python -m celery -A config.celery worker -n c5...   # ForkPoolWorker-3
pid=41  python -m celery -A config.celery worker -n c5...   # ForkPoolWorker-4
```

**Wiring proof** — Django setting → Celery config:

```
$ kubectl exec prowler-worker-... -- manage.py shell -c "
    from django.conf import settings; from config.celery import celery_app
    print(settings.CELERY_WORKER_CONCURRENCY)
    print(celery_app.conf.worker_concurrency)
  "
4
4
```

**Override demonstration** — `DJANGO_CELERY_WORKER_CONCURRENCY=2` propagates through:

```
$ kubectl set env deploy/prowler-worker DJANGO_CELERY_WORKER_CONCURRENCY=2
$ kubectl exec <new-pod> -- printenv DJANGO_CELERY_WORKER_CONCURRENCY
2
$ kubectl exec <new-pod> -- manage.py shell -c "..."
settings.CELERY_WORKER_CONCURRENCY=2
celery_app.conf.worker_concurrency=2
$ kubectl top pod <new-pod>
NAME                              CPU(cores)   MEMORY(bytes)
prowler-worker-5cb6bb6bc4-2b9q7   521m         564Mi          # was 3.4 GiB
```

Process count goes to 1 parent + 2 children (with brief overlaps from `--max-tasks-per-child 1` recycling, expected).

### Checklist

  - The change is a single Django settings line wrapping `env.int()` (the same pattern already exercised for `DJANGO_CELERY_DEADLOCK_ATTEMPTS` with no dedicated test in `api/src/backend/api/tests/test_celery_settings.py`); validated in a real cluster as documented above.
- [x] Review if code is being documented following [pyguide §3.8](https://github.com/google/styleguide/blob/gh-pages/pyguide.md#38-comments-and-docstrings).
- [x] Review if is needed to change the [Readme.md](https://github.com/prowler-cloud/prowler/blob/master/README.md) — N/A.
- [x] Ensure new entries are added to [CHANGELOG.md](https://github.com/prowler-cloud/prowler/blob/master/api/CHANGELOG.md) — added under `[1.29.0]` → `🚀 Added`.

#### API
- [x] All issue/task requirements work as expected on the API — see Evidence above.
- [x] Endpoint response output (if applicable) — N/A (worker-side change, no endpoints touched).
- [x] EXPLAIN ANALYZE output for new/modified queries or indexes (if applicable) — N/A.
- [x] Performance test results (if applicable) — included above (memory + process-count).
- [x] Any other relevant evidence of the implementation (if applicable) — included above.
- [x] Verify if API specs need to be regenerated — N/A.
- [x] Check if version updates are required — N/A (no dependency or spec changes).
- [x] Ensure new entries are added to [CHANGELOG.md](https://github.com/prowler-cloud/prowler/blob/master/api/CHANGELOG.md).

### License

## Release Notes

* **New Features**
  * Added an optional configuration to control Celery worker prefork concurrency via `DJANGO_CELERY_WORKER_CONCURRENCY` (when unset, Celery keeps its default behavior).

* **Documentation**
  * Documented the new environment variable in the configuration reference, environment example, changelog, and added troubleshooting guidance for hosts with many CPUs.

* **Tests**
  * Added coverage for unset, valid numeric, and invalid non-numeric values for the worker concurrency setting.


cesararroba (dismissed):
Thanks for the contribution!

api/CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ All notable changes to the **Prowler API** are documented in this file.
 - ASD Essential Eight (AWS) compliance framework support [(#10982)](https://github.com/prowler-cloud/prowler/pull/10982)
 - `scan-reset-ephemeral-resources` post-scan task zeroes `failed_findings_count` for resources missing from the latest full-scope scan, keeping ephemeral resources from polluting the Resources page sort [(#10929)](https://github.com/prowler-cloud/prowler/pull/10929)
 
+### 🐞 Fixed
+
+- Celery worker concurrency now defaults to 4 (configurable via `DJANGO_CELERY_WORKER_CONCURRENCY`) so memory is predictable regardless of host CPU count [(#10968)](https://github.com/prowler-cloud/prowler/issues/10968)
+
```
jfagoagas, line 164: Please:
- Move this to the next release as API 1.27.0 was released as part of Prowler v5.26.0. 
- Include it in `### 🚀 Added`

jfagoagas (changes requested):

api/CHANGELOG.md
b-abderrahmane: @jfagoagas It should be good now.

AdriiiPRodri (dismissed):
Thanks for the PR. The configurability is useful, but this currently changes default runtime behavior. Please address the inline comments below before merge.

api/src/backend/config/settings/celery.py
```diff
@@ -53,3 +53,5 @@ def _build_celery_broker_url(
 CELERY_BROKER_CONNECTION_RETRY_ON_STARTUP = True
 
 CELERY_DEADLOCK_ATTEMPTS = env.int("DJANGO_CELERY_DEADLOCK_ATTEMPTS", default=5)
+
+CELERY_WORKER_CONCURRENCY = env.int("DJANGO_CELERY_WORKER_CONCURRENCY", default=4)
```
AdriiiPRodri: This introduces a behavior change by forcing concurrency to `4` when the env var is not set.
Could we keep the current default behavior (Celery auto-detected concurrency) and only set `CELERY_WORKER_CONCURRENCY` when `DJANGO_CELERY_WORKER_CONCURRENCY` is explicitly provided?

api/.env.example
```diff
@@ -24,6 +24,7 @@ DJANGO_THROTTLE_TOKEN_OBTAIN=50/minute
 # Decide whether to allow Django manage database table partitions
 DJANGO_MANAGE_DB_PARTITIONS=[True|False]
 DJANGO_CELERY_DEADLOCK_ATTEMPTS=5
+DJANGO_CELERY_WORKER_CONCURRENCY=4
```
AdriiiPRodri: Suggestion: make this variable clearly optional (e.g., commented out with an explanation).
Keeping it active in `.env.example` makes `4` look like the default expected runtime behavior, which changes current behavior.

api/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ All notable changes to the **Prowler API** are documented in this file.
 ### 🚀 Added
 
 - `okta` provider support [(#11184)](https://github.com/prowler-cloud/prowler/pull/11184)
+- `DJANGO_CELERY_WORKER_CONCURRENCY` setting (default `4`) bounds the Celery worker prefork pool so memory is predictable regardless of host CPU count [(#11075)](https://github.com/prowler-cloud/prowler/pull/11075)
```
AdriiiPRodri: The changelog text says "default `4`", but that implies a global behavior change.
Could we reword this to: optional setting that bounds concurrency **when explicitly configured**?

api/src/backend/config/settings/celery.py
AdriiiPRodri: Could we add tests for the new setting behavior?
1. env var missing/empty -> concurrency is not forced
2. valid numeric value -> concurrency is applied
3. invalid value -> explicit error path

AdriiiPRodri (changes requested):
Please address the inline comments before merge.

b-abderrahmane:
> Please address the inline comments before merge.

Good catch @AdriiiPRodri 
Restored default behavior and added tests.

b-abderrahmane:
@AdriiiPRodri @jfagoagas friendly reminder.

AdriiiPRodri (dismissed):

AdriiiPRodri (approved):

PR #11166, opened 2026-05-14
feat(scaleway): add new provider

pedrooot:
### Description

This PR adds the new Scaleway provider with the IAM service and the check `iam_no_root_api_keys`

### Steps to review

### Checklist

<details>

</details>

#### SDK/CLI

#### UI

#### API

### License


prowler/providers/scaleway/lib/mutelist/mutelist.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class ScalewayMutelist(Mutelist):
+    """Scaleway-specific mutelist helper."""
+
+    def is_finding_muted(
+        self,
+        finding: CheckReportScaleway,
+        organization_id: str,
+    ) -> bool:
```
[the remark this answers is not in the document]
HugoPBrito: We're agreeing on dismissing this for now.

prowler/providers/scaleway/exceptions/exceptions.py
HugoPBrito, line 1: Please use 15000 codes. 14000 are currently used by okta.

docs/user-guide/providers/scaleway/getting-started-scaleway.mdx
HugoPBrito, line 1: Please divide authentication into a separate doc page, like for other providers.

prowler/providers/scaleway/services/iam/iam_no_root_api_keys/iam_no_root_api_keys.metadata.json
```diff
@@ -0,0 +1,38 @@
...
+  "Severity": "critical",
+  "ResourceType": "ScalewayAPIKey",
+  "ResourceGroup": "IAM",
+  "Description": "**Scaleway API keys** are checked to ensure none is bound to the **account root user**. The account root user is the original Scaleway account owner; its credentials bypass IAM policies and grant unrestricted access to the entire organization.",
+  "Risk": "API keys owned by the **account root user** cannot be scoped down with IAM policies. Leaking one of these keys yields immediate full control over every project, resource and billing setting in the organization, and rotating them disrupts every automation depending on root credentials.",
+  "RelatedUrl": "",
+  "AdditionalURLs": [
+    "https://www.scaleway.com/en/docs/identity-and-access-management/iam/concepts/#root-account",
+    "https://www.scaleway.com/en/docs/identity-and-access-management/iam/how-to/create-api-keys/",
+    "https://www.scaleway.com/en/docs/identity-and-access-management/iam/reference-content/users-and-applications/"
```
HugoPBrito: Please remove this url since it does not exist.

prowler/providers/scaleway/services/iam/iam_no_root_api_keys/iam_no_root_api_keys.metadata.json
```diff
@@ -0,0 +1,38 @@
+{
+  "Provider": "scaleway",
+  "CheckID": "iam_no_root_api_keys",
```
HugoPBrito: I suggest you this checkID to match with the best practices:
```suggestion
  "CheckID": "iam_api_keys_no_root_owned",
```

HugoPBrito: This, or something similar, is just to propose an alternative.

prowler/providers/scaleway/services/iam/iam_no_root_api_keys/iam_no_root_api_keys.metadata.json
```diff
@@ -0,0 +1,38 @@
...
+  ],
+  "Remediation": {
+    "Code": {
+      "CLI": "scw iam api-key delete <ACCESS_KEY>",
+      "NativeIaC": "",
+      "Other": "1. Sign in to the Scaleway console as a user with IAM admin permissions.\n2. Create a dedicated IAM user or application scoped with the minimum required policy.\n3. Generate a new API key for that bearer and roll it out to the workloads currently using the root key.\n4. Delete the API key owned by the account root user from the IAM > API keys page.",
+      "Terraform": ""
+    },
+    "Recommendation": {
+      "Text": "Never use API keys owned by the account root user for automation. Create scoped IAM users or applications, attach the least-privilege policies, and rotate any existing root API keys to that new bearer.",
```
HugoPBrito: Please use markdown here.

prowler/providers/scaleway/services/iam/iam_no_root_api_keys/iam_no_root_api_keys.py
```diff
@@ -0,0 +1,87 @@
...
+    ``region``/``organization_id`` from the resource via ``getattr`` with
+    defaults, so this lightweight object is enough to materialize a
+    MANUAL finding without polluting the real domain models.
+    """
+
+    def __init__(self, organization_id: str):
+        self.name = "iam-data-unavailable"
+        self.id = "iam-data-unavailable"
+        self.organization_id = organization_id
+        self.region = "global"
```
HugoPBrito: I personally wouldn't add this here. Please let me know what you think.

prowler/providers/scaleway/services/iam/iam_service.py
```diff
@@ -0,0 +1,142 @@
...
+    id: str
+    email: Optional[str] = None
+    username: Optional[str] = None
+    organization_id: Optional[str] = None
+    account_root_user_id: Optional[str] = None
+    mfa: bool = False
+    type_: Optional[str] = None
+    status: Optional[str] = None
+    # Provide name/id for CheckReportScaleway
+    name: str = ""
```
HugoPBrito, line 123: Should we use username or name for that? Which difference is there between those 2?

poetry.lock
HugoPBrito, line 1: Please do the uv change.

HugoPBrito (changes requested):

prowler/providers/scaleway/services/iam/iam_service.py
```diff
@@ -0,0 +1,142 @@
...
+                        ),
+                        status=(
+                            str(user.status) if getattr(user, "status", None) else None
+                        ),
+                    )
+                )
+
+            # All users in the same org share the same account_root_user_id.
+            if self.users and self.users[0].account_root_user_id:
+                self.account_root_user_id = self.users[0].account_root_user_id
```
HugoPBrito: Please pull `account_root_user_id` from `provider.identity` instead of `self.users[0]`. If the user list comes back empty the check would silently PASS on root keys.

prowler/lib/check/models.py
```diff
@@ -1318,6 +1318,53 @@ def region(self) -> str:
         return "global"
 
 
+class CheckReportScaleway(Check_Report):
```
HugoPBrito, line 1322: Please add the `@dataclass` decorator here to match `CheckReportOkta`, `CheckReportVercel` and `CheckReportCloudflare`.

prowler/providers/scaleway/scaleway_provider.py
```diff
@@ -0,0 +1,372 @@
...
+                        file=os.path.basename(__file__),
+                        message=(
+                            "The provided credentials do not have access to "
+                            f"the Scaleway organization with ID: {provider_id}"
+                        ),
+                    )
+
+            return Connection(is_connected=True)
+
+        except (
```
HugoPBrito, line 352: Please add `ScalewayIdentityError` to this `except` tuple. Otherwise the `provider_id` mismatch path gets re-wrapped as a generic auth error.

HugoPBrito (commented):
A few more things:

prowler/providers/scaleway/exceptions/exceptions.py
pedrooot: Done!

docs/user-guide/providers/scaleway/getting-started-scaleway.mdx
pedrooot: Done!

prowler/providers/scaleway/services/iam/iam_no_root_api_keys/iam_no_root_api_keys.metadata.json
pedrooot: Done!

prowler/providers/scaleway/services/iam/iam_no_root_api_keys/iam_no_root_api_keys.metadata.json
pedrooot: Done!

prowler/providers/scaleway/services/iam/iam_no_root_api_keys/iam_no_root_api_keys.metadata.json
pedrooot: Done!

prowler/providers/scaleway/services/iam/iam_no_root_api_keys/iam_no_root_api_keys.py
pedrooot: Agreed. Moved the placeholder out of the check into `iam_service.py` as `ScalewayIAMDataUnavailable(BaseModel)`, alongside `ScalewayUser`/`ScalewayAPIKey`

prowler/providers/scaleway/services/iam/iam_service.py
pedrooot: `name` is a synthetic field consumed by `CheckReportScaleway`, which reads `getattr(resource, "name", ...)` to populate the finding's `resource_name`. We set it to `email or username or id` so findings show the most human-readable identifier available. `username` stays as the native Scaleway IAM handle for any check that needs the exact value. Keeping both is intentional

poetry.lock
pedrooot: Done!

prowler/providers/scaleway/services/iam/iam_service.py
pedrooot: Good catch. `account_root_user_id` is now resolved once in `__init__` from `provider.identity.account_root_user_id` and no longer derived from `self.users[0]`, so a failed/empty user listing can't silently PASS a root-owned key

prowler/lib/check/models.py
pedrooot: Done!

prowler/providers/scaleway/scaleway_provider.py
pedrooot: Done!

mcp_server/prowler_mcp_server/prowler_app/models/finding_groups.py
```diff
@@ -76,9 +76,7 @@ class SimplifiedFindingGroup(MinimalSerializerMixin):
     )
     muted_count: int = Field(description="Total muted findings in this group", ge=0)
     new_count: int = Field(description="Number of new non-muted findings", ge=0)
-    changed_count: int = Field(
-        description="Number of changed non-muted findings", ge=0
-    )
+    changed_count: int = Field(description="Number of changed non-muted findings", ge=0)
```
HugoPBrito: This black reformatting is unrelated to the Scaleway provider (same applies to `tools/finding_groups.py`). Could you drop these from the PR or move them to a separate formatting PR? It keeps the diff focused and the changelog accurate.

uv.lock
```diff
@@ -1,5 +1,5 @@
 version = 1
-revision = 3
+revision = 2
```
HugoPBrito: `revision = 3` was downgraded to `revision = 2`. This looks like the lockfile was regenerated with an older `uv` version, which is a regression that causes CI churn. Please regenerate it with the current `uv` version so `revision = 3` is preserved.

prowler/providers/scaleway/lib/arguments/arguments.py
```diff
@@ -0,0 +1,57 @@
...
+    scaleway_parser = self.subparsers.add_parser(
+        "scaleway",
+        parents=[self.common_providers_parser],
+        help="Scaleway Provider",
+    )
+
+    # Authentication
+    auth_subparser = scaleway_parser.add_argument_group("Authentication")
+    auth_subparser.add_argument(
+        "--access-key",
```
HugoPBrito: Secrets passed on the command line leak into shell history and process listings, and the help text here already points users to the `SCW_*` env vars. Since this is a CLI-only provider, would it be better to remove `--access-key` / `--secret-key` entirely and rely only on `SCW_ACCESS_KEY` / `SCW_SECRET_KEY`? The scope flags (`--organization-id`, `--project-id`, `--region`) can stay. Open to discussion if there is a use case that needs them.

docs/user-guide/providers/scaleway/authentication.mdx
```diff
@@ -0,0 +1,37 @@
...
+
+| Variable | Purpose |
+|---|---|
+| `SCW_ACCESS_KEY` | API key access key |
+| `SCW_SECRET_KEY` | API key secret key |
+| `SCW_DEFAULT_ORGANIZATION_ID` | Optional, required when the key bearer is an application |
+| `SCW_DEFAULT_PROJECT_ID` | Optional, default project for project-scoped resources |
+| `SCW_DEFAULT_REGION` | Optional, defaults to `fr-par` |
+
+Alternatively, pass them as CLI flags (`--access-key`, `--secret-key`, `--organization-id`, `--project-id`, `--region`). The CLI emits a warning when secrets are passed via the command line; environment variables are preferred.
```
HugoPBrito: This says the CLI emits a warning when secrets are passed on the command line, but I don't see that warning emitted anywhere. `SENSITIVE_ARGUMENTS` is defined in `arguments.py` but never used. Either wire up the warning or update this doc (moot if the credential flags are removed per the comment on `arguments.py`).

HugoPBrito (changes requested):
Thanks for addressing everything! I have a few more comments. Please let me know what you think.

prowler/providers/scaleway/services/iam/iam_service.py
HugoPBrito: Thanks for the explanation.

mcp_server/prowler_mcp_server/prowler_app/models/finding_groups.py
pedrooot: Reverted! Thanks!

uv.lock
pedrooot: Done

prowler/providers/scaleway/lib/arguments/arguments.py
pedrooot: removed, good point

docs/user-guide/providers/scaleway/authentication.mdx
pedrooot: updated, thanks!

prowler/providers/scaleway/services/iam/iam_service.py
```diff
@@ -0,0 +1,159 @@
...
+        self._api = IamV1Alpha1API(self.client)
+
+        # Cached state — populated eagerly during construction
+        self.users: list[ScalewayUser] = []
+        self.api_keys: list[ScalewayAPIKey] = []
+
+        # Resolved once at authentication time from the audit identity.
+        # Deriving it from the user list instead would silently PASS root
+        # API keys whenever the user listing comes back empty.
+        self.account_root_user_id: Optional[str] = (
```
HugoPBrito: This is the "silently PASS on root keys" issue we raised earlier, but it was only partially fixed: the previous change covered the empty-user-list case, while the application-key path is still broken. `account_root_user_id` is only resolved in `setup_identity` when the auditing credential is a user key. With an application-scoped API key (a documented path: `authentication.mdx` mentions application keys + `SCW_DEFAULT_ORGANIZATION_ID`), `provider.identity.account_root_user_id` is `None`, so this becomes `None`. In `iam_api_keys_no_root_owned`, the MANUAL guard only checks `users_loaded`/`api_keys_loaded`, and the FAIL branch is `if root_user_id and ...`, so every key, including root-owned ones, silently PASSes with no FAIL and no MANUAL. Could we fall back to the already-loaded user list here?
```suggestion
        self.account_root_user_id: Optional[str] = (
            provider.identity.account_root_user_id
            or next(
                (u.account_root_user_id for u in self.users if u.account_root_user_id),
                None,
            )
        )
```
This requires `_load_users()` to run before this assignment. Alternatively, degrade the check to MANUAL when `root_user_id` is `None`. A test for "root-owned key present + `account_root_user_id` is None" would also help.

prowler/providers/scaleway/scaleway_provider.py
```diff
@@ -0,0 +1,373 @@
...
+            account_root_user_id = None
+
+            # If the bearer is a user, resolve the org from the user record
+            # and surface the email + root user id for the credentials banner.
+            if current_key.user_id:
+                user = iam.get_user(user_id=current_key.user_id)
+                organization_id = organization_id or user.organization_id
+                bearer_email = user.email
+                account_root_user_id = user.account_root_user_id
+            elif current_key.application_id and not organization_id:
```
HugoPBrito, line 221: This branch only logs a warning for application keys and leaves `account_root_user_id` unset, which is what makes the root-key check pass silently downstream (see comment on `iam_service.py`). Would it be better to resolve the org root user here too, or explicitly document that application keys are not supported for this check?

prowler/providers/scaleway/models.py
```diff
@@ -0,0 +1,53 @@
...
+
+class ScalewaySession(BaseModel):
+    """Scaleway API session information.
+
+    Stores the credentials and the underlying ``scaleway.Client`` so every
+    service can reuse the same authenticated client.
+    """
+
+    access_key: str
+    secret_key: str
```
HugoPBrito: `client` is excluded from serialization but `secret_key` is not, so any `.dict()` / `.json()` / log of `ScalewaySession` would leak the API secret. Since the whole authentication relies on it, should we exclude it from serialization and repr?
```suggestion
    secret_key: str = Field(exclude=True, repr=False)
```
`Field` is already imported in this module. The exact same pattern exists in `prowler/providers/vercel/models.py` (`VercelSession.token` is plaintext, only `http_session` is excluded), so it would be worth applying the same hardening there too (separate change / out of scope for this PR, just flagging it).

HugoPBrito (changes requested):
Did another exhaustive pass on the latest revision. One blocker is a partial fix of something we raised before, plus two hardening notes.

prowler/providers/scaleway/services/iam/iam_service.py
pedrooot: Done!

prowler/providers/scaleway/scaleway_provider.py
pedrooot: Yes, fixed.

prowler/providers/scaleway/models.py
pedrooot: Nice,I'll add it in another PR

HugoPBrito: The 🐐.

HugoPBrito: Just for reference https://github.com/prowler-cloud/prowler/pull/11198

HugoPBrito (dismissed):
Awesome work.

HugoPBrito (approved):

PR #11170, opened 2026-05-14
fix(ui): hide muted finding groups immediately on /findings

HugoPBrito:
### Context

Muting a finding group on `/findings` left the row visible until the user reloaded manually. The default `/finding-groups/latest` endpoint reads from `FindingGroupDailySummary`, which is rebuilt asynchronously by the Celery chain queued in `MuteRuleViewSet.create` (`mute_historical_findings_task` → `reaggregate_all_finding_group_summaries_task`). A `router.refresh()` fired right after `POST /mute-rules` therefore returns the same stale aggregate, so the row reappears even though the mute was applied. After a slower manual reload the chain has typically finished and the row disappears, which matches what users expect.

A previous attempt to make the reaggregation synchronous inside the request handler added 100ms–2s to every mute and scaled poorly on tenants with many scans, so the fix lives entirely in the UI.

### Description

UI-only optimistic hide with a short `sessionStorage` safety net so a fast reload does not break the illusion while the Celery chain catches up:

- New utility `ui/lib/optimistic-muted-groups.ts` (+ unit tests): persists muted `check_id`s under `prowler:optimistic-muted-groups` with a 90s TTL per entry. Each entry tracks its own `expiresAt` (no shared TTL extension across mutes), expired entries are pruned on read, defensive against `sessionStorage` exceptions (Safari private mode, quota).
- `findings-group-table.tsx`: hydrates the optimistic `Set<string>` from storage on mount, derives `visibleData` from `data` ⊖ optimistic set, and clears entries from storage when the server payload no longer includes a `check_id` (the natural confirmation signal). Replaces every previous `safeData` reference with `visibleData` so selection indices, row counts and the table stay coherent. Honors `filter[muted]=include`: when the user opted in to seeing muted groups, the row stays visible (with the muted indicator) so the optimistic state does not diverge from the post-reload view.
- Three mute paths now feed the same hide pipeline:
  - Group-level mute via `FloatingMuteButton` (checkbox + bulk button) – existing path.
  - Group-level mute via the row dropdown – `data-table-row-actions.tsx` now reads `onMuteComplete` from the selection context as a fallback so the parent `FindingsGroupTable` is notified (previously fell back to a bare `router.refresh()`).
  - Resource-level mute inside the drill-down (row dropdown or floating button with resources selected) – propagated through a new `onResourceMuteCompleted` prop on `InlineResourceContainer`, with `column-finding-resources.tsx` passing the muted finding IDs through the context.
- Defensive heuristic `willResourceMuteEmptyGroup`: when a resource-level mute leaves the expanded group fully muted on the server (every unmuted FAIL covered AND no unmuted PASS/MANUAL), we hide the group too. Conservative on purpose so we do not pre-hide a group that would still be visible post-reload.
- Drill-down auto-collapses when the expanded group is hidden.
- `findings-selection-context.tsx`: `onMuteComplete` signature widened to `(mutedIds?: string[]) => void`. Backward-compatible.

No API, MCP, or SDK changes. The Celery chain in `MuteRuleViewSet.create` continues to run untouched.

### Steps to review

1. `cd ui && pnpm install`
2. `pnpm vitest run lib/optimistic-muted-groups.test.ts components/findings/table/findings-group-table.test.tsx components/findings/table/data-table-row-actions.test.tsx components/findings/table/column-finding-resources.test.tsx components/findings/table/column-finding-groups.test.tsx` → 35 tests pass.
3. `pnpm run typecheck && pnpm run lint:check` → clean.
4. Local stack: open `/findings`.
   - **Group mute (floating button)**: select one or more group rows, click `Mute`, submit modal → rows disappear instantly.
   - **Group mute (row dropdown)**: from the row’s `…` menu pick `Mute Finding Group` → row disappears instantly.
   - **Resource mute that empties the group**: drill into a group whose only failing resource is a single one, mute it from its row dropdown → resource disappears from the drill-down AND the parent group row disappears.
   - **Resource mute that does NOT empty the group**: mute one of several failing resources → drill-down updates, parent row remains (correct, group still has unmuted findings).
   - **Fast reload (F5) within ~90s of a mute**: muted groups remain hidden because `sessionStorage` was hydrated on mount.
   - **`Include muted findings` toggle**: muted groups reappear with the muted indicator and the optimistic hide does not kick in (no divergence from post-reload state).
   - **Wait until Celery chain catches up + reload**: the muted group is gone from the server payload and the corresponding `sessionStorage` entry is auto-removed (no leftover state).

### Checklist

<details>

</details>

#### UI (if applicable)

### License


HugoPBrito:
https://github.com/user-attachments/assets/cc620b1f-00cd-48ed-bb81-bdf42a96761b

ui/components/findings/table/findings-group-table.tsx
```diff
@@ -57,33 +63,65 @@ export function FindingsGroupTable({
   const [resourceSearchInput, setResourceSearchInput] = useState("");
   const [resourceSearch, setResourceSearch] = useState("");
   const [resourceSelection, setResourceSelection] = useState<string[]>([]);
+  // Group check_ids the user just muted, hidden client-side until the server
+  // catches up. Hydrated from sessionStorage on mount so a fast reload still
+  // honours the optimistic hide. See lib/optimistic-muted-groups.ts.
+  const [optimisticallyMutedCheckIds, setOptimisticallyMutedCheckIds] =
```
pfe-nazaries, line 69: I think you can get things done in a simple way using `useOptimistic` https://react.dev/reference/react/useOptimistic

pfe-nazaries (changes requested):
Take a look at the docs of useOptimistic hook, if you think it doesnt fit here, just ping me and i will approve

ui/components/findings/table/findings-group-table.tsx
HugoPBrito: Thanks for the pointer. I checked the `useOptimistic` docs before going this way and it doesn't fit this case. Here's why:

`POST /mute-rules` doesn't apply the mute synchronously. It enqueues a Celery chain that rebuilds `FindingGroupDailySummary` async, and `/finding-groups/latest` reads from that table. So the server payload stays stale for ~100ms to 2s+ after the action completes. Making it synchronous was already rejected for performance.

`useOptimistic` breaks here on two of its own invariants:

1. The optimistic value only lives while the action is pending. Once it settles (milliseconds), the state snaps back to the stale server `data` and the row reappears. That's exactly the bug we're fixing.
2. It's in memory. It doesn't survive a remount or a full reload, and we need the hide to persist across a fast F5 within the reaggregation window (see the "Fast reload" test step).

So `useOptimistic` would only cover the instant hide, and that alone isn't enough since it reverts long before the server catches up. What's here is an optimistic update, just with the persistence and confirmation layer (`sessionStorage` + auto-clear when the `check_id` leaves the payload, 90s TTL) that `useOptimistic` doesn't provide because it assumes the server is authoritative right after the action.

I also built a `useOptimistic` PoC to confirm this empirically: the muted row flashes hidden and reappears as soon as `router.refresh()` settles (Celery still in flight), and a fast F5 loses the hide entirely. Both reproduce the original bug.

Pinging you as offered. Happy to jump on a call if it helps, or let me know if you still want a change.

pfe-nazaries (approved):

alejandrobailo:
Close due to we are asking the backend when are those findings muted, it takes time but we show a toast.

PR #11249, opened 2026-05-20
perf(api): reduce DB load in scan hot loop by 13x

AdriiiPRodri:
### Context

The per-finding loop in `tasks/jobs/scan.py` was the dominant bottleneck of `perform_prowler_scan`: for each finding, it issued multiple `SELECT FOR UPDATE` against `resource_tag_mappings`, opened a transaction per item, and re-fetched `Resource`/`ResourceTag` rows that were already known. For large scans this produced thousands of round-trips and lock contention, slowing the hot loop and increasing DB load.

This PR rewrites the micro-batch path to be set-oriented and atomic, with no schema changes.

### Description

Changes in `api/src/backend/tasks/jobs/scan.py`:

- Pre-resolve `Resource` and `ResourceTag` rows in bulk before the per-finding loop, instead of per item.
- Replace `Resource.upsert_or_delete_tags` (which issued `SELECT FOR UPDATE` per mapping) with deferred `ResourceTagMapping.bulk_create(ignore_conflicts=True)` executed once at the end of the batch.
- Wrap the entire micro-batch in a single `rls_transaction` (was 2N). Deadlock retry now operates at the batch level.
- Populate `Finding.resource_regions`, `resource_services` and `resource_types` directly on INSERT, removing the post-INSERT `bulk_update` pass.
- Raise `SCAN_DB_BATCH_SIZE` from 500 to 1000.
- Add `update_fields=[...]` to `Scan` / `Provider` saves to avoid full-row writes.
- Throttle progress saves to either a 1% delta or a 10s interval (whichever comes first).
- Preserve findings with empty `resource_uid` (IaC scans, some Azure/GCP/K8s findings).

No schema changes. No migrations. Behaviorally, the micro-batch is now atomic: errors that were previously masked by per-finding SAVEPOINTs may now surface in logs (the batch is retried on deadlock).

#### Measured impact (3000 findings per micro-batch)

| Metric | Before | After | Delta |
| --- | --- | --- | --- |
| Wall-clock | 20.8s | 1.57s | 13.2x faster |
| `COMMIT` count | 6003 | 2 | -99.97% |
| `SELECT FOR UPDATE` on `resource_tag_mappings` | 15000 | 0 | -100% |

### Steps to review

1. Read `api/src/backend/tasks/jobs/scan.py` end-to-end; the change is concentrated in a single file.
2. Confirm the bulk pre-resolution step covers all `Resource` / `ResourceTag` lookups previously done inside the loop.
3. Verify that the deferred `ResourceTagMapping.bulk_create(ignore_conflicts=True)` correctly replaces the previous per-mapping upsert path (idempotent on retry).
4. Check the new transaction boundary: one `rls_transaction` per micro-batch instead of per finding; confirm the deadlock retry path still re-runs the whole batch safely.
5. Confirm that empty `resource_uid` findings (IaC, some Azure/GCP/K8s) are kept and stored.
6. Run the API test suite focused on scan and reports tasks.

### Checklist

<details>

</details>

#### SDK/CLI

#### API

### License


api/src/backend/tasks/jobs/scan.py
```diff
@@ -548,281 +566,398 @@ def _process_finding_micro_batch(
...
+                        unique_fields=["tenant_id", "resource_id", "tag_id"],
+                    )
+                    # Preserve original behavior: bump updated_at on resources that
+                    # actually got NEW mappings (existing mappings have m.pk == None
+                    # after ignore_conflicts, newly created mappings have m.pk set).
+                    resource_uid_by_id = {
+                        str(r.id): uid for uid, r in resource_cache.items()
+                    }
+                    for m in created_tag_mappings:
+                        if m.pk is not None:
```
Davidm4r: In created_tag_mappings you are using ignore_conflicts=True.
That means that the database does not return the IDs of the inserted rows, so we won't enter to this If right?

api/src/backend/tasks/jobs/scan.py
```diff
@@ -548,281 +566,398 @@ def _process_finding_micro_batch(
...
+                        ignore_conflicts=True,
+                        unique_fields=["tenant_id", "resource_id", "finding_id"],
+                    )
+                    inserted = sum(1 for m in created_mappings if m.pk)
+                    if inserted != len(mappings_to_create):
+                        logger.error(
+                            f"scan {scan_instance.id}: expected "
+                            f"{len(mappings_to_create)} ResourceFindingMapping rows, "
+                            f"inserted {inserted}. Rolling back micro-batch."
+                        )
```
Davidm4r, line 918: Because the other comment, if the pk is None, inserted will be 0.
We will get excpect N rows, inserted 0.

AdriiiPRodri: You're right: `ignore_conflicts=True` does not populate pk, so this branch fires on every successful batch. It's pre-existing from #10724 though, keeping it as-is to scope this PR to the perf rewrite and opening a separate fix for the silent-failure detection

api/src/backend/tasks/jobs/scan.py
AdriiiPRodri: Good catch. Fixed with a pre-SELECT of existing `(resource_id, tag_id)` pairs, so `updated_at` is bumped only on resources that actually gain a mapping

Davidm4r (dismissed):

jfagoagas (dismissed):

cesararroba (approved):

PR #11355, opened 2026-05-25
fix(gcp): honour org-level aggregated sinks in logging_sink_created check

### LINKED ISSUE #11343: logging_sink_created false-FAILs for projects covered only by org/folder-level aggregated sinks
### Issue search

Prowler CLI/SDK

### Cloud Provider

GCP

### Steps to Reproduce

1. Under an organization that uses **only** an org-level (or folder-level) aggregated log sink with `includeChildren=True` — i.e. a single sink at the org node routes all descendant projects' logs into one destination, no per-project sinks — verify the topology:

   ```
   gcloud logging sinks list --organization=<ORG_ID> \
     --format='table(name,destination,filter:wrap=40,includeChildren)'
   ```

   At least one sink with `INCLUDE_CHILDREN=True` should be present.

2. Run Prowler against any project under that org:

   ```
   prowler gcp --project-id <project-id> --check logging_sink_created
   ```

### Expected behavior

`PASS` — the project's logs are exported because an ancestor aggregated sink (`includeChildren=True`) covers it. This is the **GCP-recommended** pattern for centralized logging, and is the architecture used by [`terraform-example-foundation`](https://github.com/terraform-google-modules/terraform-example-foundation) via the `log-export` module's `parent_resource_type = "organization"` + `include_children = true` configuration.

`FAIL` — `There are no logging sinks to export copies of all the log entries in project <project-id>.`

Root cause: `prowler/providers/gcp/services/logging/logging_sink_created/logging_sink_created.py` (and the underlying service in `logging_service.py`) only enumerates **project-level** sinks via `logging.projects.sinks.list`. It doesn't query `logging.organizations.sinks.list` or `logging.folders.sinks.list`, and it has no notion of `includeChildren=True` coverage of descendant projects.

```python
for sink in logging_client.sinks:
    if sink.filter == "all":
        projects_with_logging_sink[sink.project_id] = sink
```

`logging_client.sinks` only contains project-scope sinks. Org/folder aggregated sinks are invisible to the check. Every project under an org that uses centralized aggregation gets a false-FAIL.

### Why this is a real problem

Centralized aggregation via org/folder-level sinks is the GCP-recommended pattern for organizations larger than a single project. It's documented [here](https://cloud.google.com/logging/docs/export/aggregated_sinks) and used by:

- The official [terraform-example-foundation blueprint](https://github.com/terraform-google-modules/terraform-example-foundation) (`1-org/envs/shared/log-export.tf`)
- The widely-used [`terraform-google-modules/log-export`](https://github.com/terraform-google-modules/terraform-google-log-export) module's `parent_resource_type = "organization"` / `folder"` configuration

Any organization following these patterns trips this check on every project they own. The check is enforcing an old pattern of one-sink-per-project, before GCP introduced aggregated sinks (~2019).

### Environment Resource

GitHub Actions `ubuntu-24.04` runner; also reproduced locally via `docker run prowlercloud/prowler:v5.28.0 gcp …`.

### OS used

Ubuntu 24.04 (GHA runner) / macOS.

### Prowler version

5.28.0

### Python version

3.12 (bundled in `prowlercloud/prowler:v5.28.0`)

### Pip version

N/A — Docker image install.

### Context

Suggested fix — also enumerate `organizations.sinks.list` and `folders.sinks.list`, build the resource hierarchy (project → folder ancestors → org), and treat a project as covered if any ancestor's sink has `includeChildren=True` and matches the inclusive-filter test (`filter == "all"` or empty).

Sketch:

```python
# In LoggingService:
def _get_sinks(self):
    # existing: project-scope sinks
    for project_id in self.project_ids:
        ...

    # NEW: org-scope sinks
    org_sinks = self.client.organizations().sinks().list(
        parent=f"organizations/{self.organization_id}"
    ).execute().get("sinks", [])
    self.aggregated_sinks.extend(s for s in org_sinks if s.get("includeChildren"))

    # NEW: folder-scope sinks (per ancestor folder of each project)
    ...

# In logging_sink_created.execute():
for project in self.project_ids:
    if has_project_level_all_sink(project): PASS
    elif covered_by_ancestor_aggregated_sink(project): PASS
    else: FAIL
```

The same pattern affects related per-project logging checks (the `logging_log_metric_filter_and_alert_*` family relies on per-project metric filters + alert policies, which can also be centrally implemented). A more general fix would be to surface a "covered by ancestor aggregated control" notion that those checks can also consume.

Happy to open a PR for the sink-aggregation case if there's appetite for the approach.

AlexanderSanin:
## Summary

Fixes false-FAIL results from the `logging_sink_created` check for GCP projects that rely on an **organisation-level aggregated Cloud Logging sink** with `includeChildren=True` instead of a project-scoped sink.

This is a common centralised-logging pattern (CIS GCP 2.2 / GCP best practice), but Prowler only queried `projects/{id}/sinks`, so every child project of an org with a central sink was incorrectly reported as non-compliant.

**Changes:**
- `Sink` model gains `include_children: bool = False` to carry the GCP `includeChildren` field.
- New `Logging._get_org_sinks()` method queries `organizations/{id}/sinks` for every organisation referenced by the scanned projects and appends the results to `self.sinks`.
- `logging_sink_created.execute()` now checks whether a project's organisation has a covering sink (`filter=all`, `include_children=True`) and emits PASS in that case, even when the project has no direct sink.
- 3 new unit-test cases added; 2 existing MagicMock-based tests corrected (`include_children` was unset, which would have produced a false-positive with the new logic).

Closes #11343

## Test plan

- [x] `pytest tests/providers/gcp/services/logging/` — all 64 tests pass (8 existing + 3 new for this fix)
- [ ] Run `prowler gcp --check logging_sink_created` against a project covered only by an org-level aggregated sink → should now report **PASS**
- [ ] Verify projects with no sink at all still report **FAIL**
- [ ] Verify projects with a direct project-level sink still report **PASS** with the project sink name in `status_extended`


AlexanderSanin:
Hey @jfagoagas. Could you, please, have a look at this?

HugoPBrito:
Hi @AlexanderSanin,

The fix looks good so far. Could you please provide evidences to confirm this works as expected?

AlexanderSanin:
Hi @HugoPBrito, thanks for the review!

**Evidence the fix works as expected:**

The fix is covered by unit tests that run in CI. Here's what each test proves:

| Test | Scenario | Expected |
|---|---|---|
| `test_org_level_sink_with_include_children_passes` | Project has no direct sink; its org has a sink with `includeChildren=True` | ✅ PASS |
| `test_org_level_sink_without_include_children_fails` | Project has no direct sink; org sink has `includeChildren=False` | ❌ FAIL (correct) |
| `test_project_sink_takes_precedence_over_org_sink` | Project has both a direct sink and a covering org sink | ✅ PASS, reported against the project sink |
| `test_org_sinks_fetched_when_project_has_organization` | Service layer fetches `organizations/{id}/sinks` when a project has an org | org sink appears in `sinks` list with `include_children=True` |
| `test_org_sinks_skipped_when_no_organization` | Projects have no organisation | no `organizations/…` entries added to sinks |

All 66 logging tests pass locally (`pytest tests/providers/gcp/services/logging/`).

The root cause of the false-FAIL was that `_get_sinks()` only called `sinks().list(parent="projects/{id}")`. The fix adds `_get_org_sinks()` which also calls `sinks().list(parent="organizations/{id}")` for every organisation referenced by the scanned projects, and `logging_sink_created` now checks for a matching org-level sink with `includeChildren=True` before emitting FAIL.

I've also pushed a follow-up commit addressing the black formatting and missing changelog entry flagged by CI.

HugoPBrito:
Hi @AlexanderSanin,

I was looking for live evidence, despite being censored, so we can ensure it works at execution. Could you please test it?

@alinealfa since you raised the issue, could you please test it from this branch and let us know if the fix works for you?

AlexanderSanin:
> Hi @AlexanderSanin,
> 
> I was looking for live evidence, despite being censored, so we can ensure it works at execution. Could you please test it?
> 
> @alinealfa since you raised the issue, could you please test it from this branch and let us know if the fix works for you?

Already confirmed by @alinealfa in open issue

alinealfa:
Thanks @AlexanderSanin @jfagoagas  and the Prowler team for turning this around so quickly! 🙏

I built `fix/logging-sink-org-coverage-clean` and tested it against a real GCP org that uses an organization-level aggregated sink with `includeChildren=True`:

- `prowler gcp --check logging_sink_created` runs cleanly; the new `organizations/{id}/sinks` query succeeds against a live org with the same credentials Prowler uses, and `_get_org_sinks()` evaluates the results per project as expected.
- `pytest tests/providers/gcp/services/logging/` — all green, including the new `include_children` PASS/FAIL cases.

From my side this resolves #11343 as designed.

HugoPBrito (dismissed):

HugoPBrito (approved):

HugoPBrito:
Thanks a lot @AlexanderSanin for this contribution and @alinealfa for reporting the problem and testing the solution ❤️

PR #11375, opened 2026-05-27
Add entra_conditional_access_policy_no_exclusion_gaps check

### LINKED ISSUE #11062: [New Check]: Conditional Access excluded objects must be covered by another policy (no exclusion gaps)
### Existing check search

### Provider

Microsoft 365

### New provider name

_No response_

### Service or product area

entra

### Suggested check name

`entra_conditional_access_policy_no_exclusion_gaps`

### Context and goal

- **Security condition to validate:** For every enabled Conditional Access policy, every object listed in any `exclude*` collection (`excludeUsers`, `excludeGroups`, `excludeRoles`, `excludeApplications`, `excludeServicePrincipals`, `excludeLocations`, `excludePlatforms`) must appear in at least one `include*` collection of *some* enabled Conditional Access policy. Excluded objects that never appear as included anywhere are "gaps" — they sit completely outside the CA control plane.
- **Why it matters:** Excluding a principal from a policy is a common and legitimate operation (break-glass accounts, the Directory Synchronization Accounts role, certain service principals, etc.). It only stays safe when the excluded principal is *also* covered by another CA policy that enforces compensating controls. Otherwise the exclusion silently removes the principal from CA enforcement entirely, which is exactly how MFA bypass and lateral movement against admin accounts have happened in real incidents.
- **Resource involved:** Microsoft Entra Conditional Access policies and their inclusion/exclusion collections (users, groups, roles, applications, service principals, locations, platforms).

### Expected behavior

- **Resource or scope to evaluate:** All Conditional Access policies whose `state` is `enabled`. Build a global "include set" by union-ing every `include*` collection across all enabled policies (per object type). For each enabled policy, walk every `exclude*` collection and compare each entry against the include set of the same type.
- **PASS when:** every excluded object identifier exists in the global include set for its corresponding object type. Also PASS when no enabled policy exists or no policy uses any exclusion.
- **FAIL when:** at least one excluded object identifier does not appear in the global include set of its type. The finding should report the orphaned object IDs grouped by type (users / groups / roles / apps / SPs / locations / platforms) and the policies that excluded them.
  - Skip the **Directory Synchronization Accounts** role exclusion (template ID `d29b2b05-8046-44ba-8758-1e26182fcf32`) — Prowler already enforces this exclusion in `entra_conditional_access_policy_directory_sync_account_excluded`, and it is *intended* to have no fallback.
  - Skip exclusions that resolve to confirmed emergency-access accounts (the same accounts validated by `entra_emergency_access_exclusion`); they are intentional gaps.
  - Treat report-only policies as out of scope: only `state = enabled` policies count for both the exclusion side and the include set, mirroring real enforcement.
  - Group identifiers should be matched recursively when feasible (i.e. an exclusion of group A is "covered" if A is included; transitive group expansion is out of scope for v1).

### References

- Microsoft Graph v1.0 — `conditionalAccessUsers` (include/exclude users, groups, roles): https://learn.microsoft.com/en-us/graph/api/resources/conditionalaccessusers?view=graph-rest-1.0
- Microsoft Graph v1.0 — `conditionalAccessApplications` (include/exclude applications): https://learn.microsoft.com/en-us/graph/api/resources/conditionalaccessapplications?view=graph-rest-1.0
- Plan a Conditional Access deployment (gap analysis guidance): https://learn.microsoft.com/en-us/entra/identity/conditional-access/plan-conditional-access
- Secure emergency access accounts in Microsoft Entra ID: https://learn.microsoft.com/en-us/entra/identity/role-based-access-control/security-emergency-access

### Suggested severity

Medium

- **Existing patterns to follow:** Reuse the Conditional Access iteration pattern from `entra_conditional_access_policy_mfa_enforced_for_guest_users` and `entra_emergency_access_exclusion` (`prowler/providers/m365/services/entra/`).
- **Service change (`entra_service.py`):** the `ConditionalAccessPolicy` model already exposes `includeUsers/Groups/Roles` and `excludeUsers/Groups/Roles`; verify and add (if missing) `includeApplications/excludeApplications`, `includeServicePrincipals/excludeServicePrincipals` (under `clientApplications`), `includeLocations/excludeLocations`, `includePlatforms/excludePlatforms`. No new Graph endpoint needed; the policy listing already returns these collections.
- **Permissions / scopes:** No additional permissions beyond Prowler's M365 baseline (`Policy.Read.All` already grants read access to all Conditional Access policies).
- **PowerShell is NOT needed**; this check uses the existing Microsoft Graph v1.0 data already loaded by the entra service.
- **Related checks (do not duplicate):**
  - `entra_conditional_access_policy_directory_sync_account_excluded` — wants the dir-sync role *excluded* with no fallback. Add it as a hard skip in this check.
  - `entra_emergency_access_exclusion` — wants emergency accounts *excluded*. Treat them as expected gaps.

arieleli01212:
## Summary

Closes #11062.

Adds the `entra_conditional_access_policy_no_exclusion_gaps` check under `prowler/providers/m365/services/entra/`. The check catches Conditional Access exclusions that leave principals or applications with no CA coverage at all.

**How it works:**

1. Collects all enabled Conditional Access policies (report-only and disabled are ignored).
2. Builds a per-type global include set by unioning every `included_users`, `included_groups`, `included_roles`, and `included_applications` list across those policies.
3. For each enabled policy, walks every exclude collection and checks whether each entry appears in the global include set of the same type.
4. Reports **PASS** when every excluded object is found in the global include set, or when no enabled policy uses any exclusion. Reports **FAIL** when at least one excluded object is absent from the global include set, listing the orphaned IDs grouped by type.

**Intentional exceptions skipped automatically:**

- The Directory Synchronization Accounts role (`d29b2b05-8046-44ba-8758-1e26182fcf32`) — already validated by `entra_conditional_access_policy_directory_sync_account_excluded`.
- Confirmed emergency-access accounts — accounts excluded from every enabled blocking policy, using the same definition as `entra_emergency_access_exclusion`.

**Files added:**

- `prowler/providers/m365/services/entra/entra_conditional_access_policy_no_exclusion_gaps/entra_conditional_access_policy_no_exclusion_gaps.py`
- `prowler/providers/m365/services/entra/entra_conditional_access_policy_no_exclusion_gaps/entra_conditional_access_policy_no_exclusion_gaps.metadata.json`
- `prowler/providers/m365/services/entra/entra_conditional_access_policy_no_exclusion_gaps/__init__.py`
- `tests/providers/m365/services/entra/entra_conditional_access_policy_no_exclusion_gaps/entra_conditional_access_policy_no_exclusion_gaps_test.py`

No changes to `entra_service.py` or any other existing file — the existing `ConditionalAccessPolicy` model already exposes all the fields this check needs.

## Test plan

- [ ] `pytest tests/providers/m365/services/entra/entra_conditional_access_policy_no_exclusion_gaps/` — all 19 cases cover: no policies, disabled/report-only policies skipped, no exclusions → PASS, excluded objects covered by another policy → PASS, dir-sync role skipped, emergency accounts skipped, `"All"` in include set covers any specific ID, uncovered user/group/role/app → FAIL, multiple gap types reported together, mixed PASS/FAIL across policies.

### DIFF
--- a/prowler/providers/m365/services/entra/entra_conditional_access_policy_no_exclusion_gaps/entra_conditional_access_policy_no_exclusion_gaps.metadata.json
+++ b/prowler/providers/m365/services/entra/entra_conditional_access_policy_no_exclusion_gaps/entra_conditional_access_policy_no_exclusion_gaps.metadata.json
@@ -1,0 +1,43 @@
+{
+  "Provider": "m365",
+  "CheckID": "entra_conditional_access_policy_no_exclusion_gaps",
+  "CheckTitle": "Conditional Access excluded objects are covered by another enabled policy (no exclusion gaps)",
+  "CheckType": [],
+  "ServiceName": "entra",
+  "SubServiceName": "",
+  "ResourceIdTemplate": "",
+  "Severity": "medium",
+  "ResourceType": "NotDefined",
+  "ResourceGroup": "IAM",
+  "Description": "For every enabled Conditional Access policy, every object listed in an exclude collection (users, groups, roles, or applications) must appear in the include collection of at least one other enabled Conditional Access policy for the same object type. Excluded objects that are never explicitly included elsewhere sit completely outside the CA enforcement plane.",
+  "Risk": "Excluding a principal or application from a Conditional Access policy is safe only when another policy still enforces controls on that object. An exclusion that has no matching include in any other enabled policy silently removes the object from all CA enforcement, creating a gap that can be used to bypass MFA, block, or other grant controls — a pattern observed in real privilege-escalation and lateral-movement incidents.",
+  "RelatedUrl": "",
+  "AdditionalURLs": [
+    "https://learn.microsoft.com/en-us/graph/api/resources/conditionalaccesspolicy?view=graph-rest-1.0",
+    "https://learn.microsoft.com/en-us/graph/api/resources/conditionalaccessusers?view=graph-rest-1.0",
+    "https://learn.microsoft.com/en-us/entra/identity/conditional-access/plan-conditional-access",
+    "https://learn.microsoft.com/en-us/entra/identity/role-based-access-control/security-emergency-access"
+  ],
+  "Remediation": {
+    "Code": {
+      "CLI": "",
+      "NativeIaC": "",
+      "Other": "1. Navigate to the Microsoft Entra admin center at https://entra.microsoft.com.\n2. Expand Protection > Conditional Access and select Policies.\n3. For each policy that reports an exclusion gap, either:\n   a. Remove the exclusion if it is no longer needed, or\n   b. Create (or update) another enabled Conditional Access policy that explicitly includes the excluded object and enforces appropriate compensating controls.\n4. Save all modified policies and re-evaluate.",
+      "Terraform": ""
+    },
+    "Recommendation": {
+      "Text": "Ensure every principal or application excluded from a Conditional Access policy is explicitly included in at least one other enabled policy that enforces compensating controls. Intentional permanent exceptions (the Directory Synchronization Accounts role, break-glass emergency accounts) are automatically excluded from this check.",
+      "Url": "https://hub.prowler.com/check/entra_conditional_access_policy_no_exclusion_gaps"
+    }
+  },
+  "Categories": [
+    "identity-access",
+    "e3"
+  ],
+  "DependsOn": [],
+  "RelatedTo": [
+    "entra_conditional_access_policy_directory_sync_account_excluded",
+    "entra_emergency_access_exclusion"
+  ],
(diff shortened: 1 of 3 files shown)


danibarranqueroo:
Hi @arieleli01212 👋

Thank you for this contribution and for tackling #11062! We ended up with two PRs for the same check (this one and #11577), so to avoid duplicate work we consolidated on a single base and took #11577 forward.

That said, your implementation was genuinely valuable during review (especially around how "All"/wildcard includes
should count toward coverage), and it directly informed the final approach we landed on.

To make sure your work is credited, I've added you as a co-author on #11577, so you'll appear on the merged commit.

I'm going to close this PR in favor of #11577, but I really appreciate the effort here and hope you'll send more contributions our way. 🙏

PR #11236, opened 2026-05-19
feat(m365/entra): add entra_conditional_access_policy_no_deleted_object_references check

ernestprovo23:
### Context

Conditional Access policies that reference deleted users, groups, or roles silently change runtime enforcement: stale `include*` references shrink scope, stale `exclude*` references can alter exemption logic. This is a well-known root cause of "we thought MFA was required but it wasn't" incidents. This PR adds a Prowler check that surfaces those stale references.

Fix #11063

### Description

- New check **`entra_conditional_access_policy_no_deleted_object_references`** iterates every Conditional Access policy (regardless of state, per the issue spec) and reports identifiers in `includeUsers/excludeUsers/includeGroups/excludeGroups/includeRoles/excludeRoles` that no longer resolve in Microsoft Entra ID.
- **Architecture mirrors #11232.** Resolution runs eagerly at service-init time, not from inside the check. After the main `asyncio.gather` in `Entra.__init__`, a second `loop.run_until_complete` calls `_resolve_directory_object_references`, which walks the just-loaded `conditional_access_policies`, deduplicates referenced identifiers per type, skips the sentinel values `All`, `None`, and `GuestsOrExternalUsers`, and queries Microsoft Graph for each one. The result is cached on the entra client as `unresolved_directory_object_references: Set[Tuple[str, str]]`. Sync checks read this attribute without making any Graph calls of their own.
- **Endpoints queried**: `/users/{id}`, `/groups/{id}`, `/roleManagement/directory/roleDefinitions/{id}`. No new permissions beyond Prowler's M365 baseline (`Directory.Read.All`).
- **Error handling**: Only `HTTP 404` (status code or `Request_ResourceNotFound` error code) flags an identifier as deleted. Transient errors (5xx, throttling, network failures) are logged and skipped — they must not produce false-positive FAILs per the spec.
- **Severity**: Medium, Category `identity-access` + `e3`, ResourceGroup `IAM`.

### Steps to review

1. **Check skeleton** under `prowler/providers/m365/services/entra/entra_conditional_access_policy_no_deleted_object_references/` follows the layout of the recently merged `entra_app_registration_client_secret_unused` check (PR #11232).
2. **Service-layer changes** in `entra_service.py`:
   - Added `Set` to the typing import.
   - Added `_resolve_directory_object_references` and helper `_resolve_identifiers_for_type` methods.
   - Wired a second-phase `loop.run_until_complete` in `Entra.__init__` between the main attribute unpacking and the `if created_loop` cleanup.
[...]

PR #11396, opened 2026-05-30
feat(entra): add check for CA policies referencing deleted principals

### LINKED ISSUE #11063: [New Check]: Conditional Access policies must not reference deleted users, groups, or roles
### Existing check search

### Provider

Microsoft 365

### New provider name

_No response_

### Service or product area

entra

### Suggested check name

`entra_conditional_access_policy_no_deleted_object_references`

### Context and goal

- **Security condition to validate:** Every object identifier referenced by any Conditional Access policy under `conditions.users` — namely `includeUsers`, `excludeUsers`, `includeGroups`, `excludeGroups`, `includeRoles`, `excludeRoles` — must resolve to an existing Microsoft Entra object.
- **Why it matters:** When a user, group, or directory role referenced by a Conditional Access policy stops resolving in the directory (account deleted, group deleted, role template removed), the reference becomes orphaned. In `include*` collections this silently shrinks the policy's enforcement scope; in `exclude*` collections it can cause the policy to evaluate unexpectedly. Either way, the policy stops behaving the way the operator believes it does, which is one of the more common root causes of "we thought MFA was required but it wasn't" incidents.
- **Resource involved:** Microsoft Entra Conditional Access policies and the users, security/Microsoft 365 groups, and directory role templates they reference.

### Expected behavior

- **Resource or scope to evaluate:** **All Conditional Access policies in the tenant**, regardless of `state` (`enabled`, `disabled`, and `enabledForReportingButNotEnforced`). Disabled policies still represent operator intent and a stale reference there is a misconfiguration that will go live the moment the policy is re-enabled. Build the deduplicated set of identifiers (per type) across all six collections — `includeUsers`, `excludeUsers`, `includeGroups`, `excludeGroups`, `includeRoles`, `excludeRoles` — for every policy, then resolve each identifier via Microsoft Graph using the type-appropriate endpoint:
  - Users → `GET /users/{id}`
  - Groups → `GET /groups/{id}`
  - Roles → `GET /roleManagement/directory/roleDefinitions/{id}`
- **PASS when:** every referenced identifier resolves successfully (HTTP 200) on the v1.0 Graph endpoint corresponding to its type. Also PASS when no policy references any user, group, or role.
- **FAIL when:** at least one referenced identifier returns HTTP 404 from its resolution endpoint. The finding should report each missing identifier together with its type (User / Group / Role), the policies that reference it, and the include vs exclude side.
  - Treat any non-404 Graph error (5xx, throttling, transient network failure) as a check error, not a FAIL — do not flag an object as deleted on transient failures.
  - Cache resolved identifiers across the run so an object referenced by N policies costs one Graph call, not N. Cache per type.
  - Disabled policies still count: stale references in disabled policies are a misconfiguration that becomes live the moment the policy is enabled.
  - Sentinel values such as `"All"`, `"None"`, `"GuestsOrExternalUsers"` (used in `includeUsers`/`excludeUsers`) are not object identifiers and must be skipped before issuing a Graph lookup.

### References

- Microsoft Graph v1.0 — `conditionalAccessUsers` (includeUsers, excludeUsers, includeGroups, excludeGroups, includeRoles, excludeRoles): https://learn.microsoft.com/en-us/graph/api/resources/conditionalaccessusers?view=graph-rest-1.0
- Microsoft Graph v1.0 — Get user (returns 404 if not found): https://learn.microsoft.com/en-us/graph/api/user-get?view=graph-rest-1.0
- Microsoft Graph v1.0 — Get group (returns 404 if not found): https://learn.microsoft.com/en-us/graph/api/group-get?view=graph-rest-1.0
- Microsoft Graph v1.0 — Get unifiedRoleDefinition (returns 404 if not found): https://learn.microsoft.com/en-us/graph/api/unifiedroledefinition-get?view=graph-rest-1.0
- Conditional Access — Users, groups, and workload identities: https://learn.microsoft.com/en-us/entra/identity/conditional-access/concept-conditional-access-users-groups

### Suggested severity

Medium

- **Existing patterns to follow:** Reuse the Conditional Access iteration pattern from `entra_conditional_access_policy_mfa_enforced_for_guest_users`. Add a small resolver helper in `entra_service.py` that takes a list of identifiers and a type (`user` / `group` / `role`), issues the corresponding `GET /…/{id}?$select=id,displayName` calls, and returns the set of identifiers that 404. Cache results within the run, keyed by `(type, id)`.
- **Permissions / scopes:** No additional permissions beyond Prowler's M365 baseline (`Directory.Read.All`, `Policy.Read.All`). `Directory.Read.All` already grants reads against `/users/{id}`, `/groups/{id}`, and the unified role-management endpoints used here.
- **PowerShell is NOT needed**; this check uses Microsoft Graph v1.0 only.
- **Related checks (do NOT duplicate, complementary):**
  - `entra_conditional_access_policy_groups_management_restricted` (issue #11060) — audits whether referenced groups are RMAU/role-assignable, assumes they exist.
  - `entra_conditional_access_policy_no_exclusion_gaps` (issue #11062) — audits whether excluded objects are covered by another policy. Could share the resolver helper to detect deleted IDs as a separate failure category instead of mixing them in.

RoySerbi:
## Description

Add a new check entra_conditional_access_policy_references_deleted_principals that verifies Conditional Access policies do not reference deleted users, groups, or directory roles.

## Motivation

When a user, group, or role is deleted from Entra ID but still referenced in a Conditional Access policy, the policy may silently grant or block access in unintended ways. Stale references also obscure the effective scope of a policy, complicating audits and incident response.

## Changes

- New check: entra_conditional_access_policy_references_deleted_principals
- Iterates all enabled and report-only CA policies
- Cross-references user/group IDs in policy conditions against current tenants objects
- Skips special values (All, None, GuestsOrExternalUsers)
- Reports PASS if no dangling references, FAIL with details otherwise
- Includes comprehensive test coverage (10 test scenarios)

## Testing

Tests cover: no policies, disabled policies, valid users/groups, deleted users, deleted groups, deleted users in exclude, multiple deleted refs, report-only mode, mixed results, and special user values.

Fixes #11063

### DIFF
--- a/prowler/providers/m365/services/entra/entra_conditional_access_policy_references_deleted_principals/entra_conditional_access_policy_references_deleted_principals.metadata.json
+++ b/prowler/providers/m365/services/entra/entra_conditional_access_policy_references_deleted_principals/entra_conditional_access_policy_references_deleted_principals.metadata.json
@@ -1,0 +1,40 @@
+{
+  "Provider": "m365",
+  "CheckID": "entra_conditional_access_policy_references_deleted_principals",
+  "CheckTitle": "Conditional Access policies must not reference deleted users, groups, or roles",
+  "CheckType": [],
+  "ServiceName": "entra",
+  "SubServiceName": "",
+  "ResourceIdTemplate": "",
+  "Severity": "medium",
+  "ResourceType": "NotDefined",
+  "ResourceGroup": "IAM",
+  "Description": "Conditional Access policies are reviewed to verify that none reference users, groups, or directory roles that no longer exist in the tenant. Deleted principals left in policy conditions can cause unexpected access grants or denials and make policies harder to audit.",
+  "Risk": "A Conditional Access policy that references a deleted user, group, or role may silently grant or block access in unintended ways. Stale references also obscure the effective scope of a policy, complicating audits and incident response.",
+  "RelatedUrl": "",
+  "AdditionalURLs": [
+    "https://learn.microsoft.com/en-us/entra/identity/conditional-access/concept-conditions-users",
+    "https://learn.microsoft.com/en-us/graph/api/resources/conditionalaccesspolicy"
+  ],
+  "Remediation": {
+    "Code": {
+      "CLI": "",
+      "NativeIaC": "",
+      "Other": "1. Navigate to the Microsoft Entra admin center at https://entra.microsoft.com.\n2. Expand **Protection** > **Conditional Access** and select **Policies**.\n3. Open each policy that references users, groups, or roles.\n4. Under **Users**, review the **Include** and **Exclude** sections for any deleted users, groups, or directory roles.\n5. Remove or replace stale references.\n6. Save the policy.",
+      "Terraform": ""
+    },
+    "Recommendation": {
+      "Text": "Review Conditional Access policies regularly to ensure they do not reference deleted users, groups, or directory roles. Remove or replace stale references to maintain correct access control behavior.",
+      "Url": "https://hub.prowler.com/check/entra_conditional_access_policy_references_deleted_principals"
+    }
+  },
+  "Categories": [
+    "identity-access",
+    "e3"
+  ],
+  "DependsOn": [],
+  "RelatedTo": [
+    "entra_conditional_access_policy_directory_sync_account_excluded"
+  ],
+  "Notes": "This check only inspects enabled and report-only policies. Disabled policies are skipped. The check compares referenced principal IDs against the current set of users, groups, and built-in directory roles in the tenant."
+}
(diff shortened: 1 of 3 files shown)


danibarranqueroo:
Hi @RoySerbi  thanks for tackling this! Issue #11063 is being resolved by #11236, so I'm closing this as superseded.

Quick why, since both are solid: #11236 also covers role references, resolves IDs directly against Graph (avoiding false positives on tenants with >100 groups, where _get_groups doesn't paginate), and distinguishes a real 404 from transient 5xx/throttling errors.

That said, we pulled two nice touches from your PR into #11236: the report-only-mode callout in the failure message and the RelatedTo link. Really appreciate the contribution, please keep them coming! 🙏

PR #11398, opened 2026-05-31
feat(sdk): add Provider.get_class dynamic provider resolver

StylusFrost:
### Context

First slice toward decoupling the API from hardcoded provider types.

The SDK already discovers providers dynamically via entry points (#10700), but the class-resolution logic ("given a provider name, return its provider class") lived inline inside `init_global_provider` (which sets global state and calls `sys.exit`) and `get_providers_help_text`. There was no public, side-effect-free API the Django API could call, forcing it to keep a hardcoded `match` over provider types.

### Scope
- Includes: public `Provider.get_class(provider)`, refactor of `init_global_provider` to use it, collision warning moved to the point of use
- Excludes: API consumption of the resolver, provider storage changes

### Autonomy
- [ ] Tests cover this unit

### Description

- Add public `Provider.get_class(provider)` resolving a provider class by name for both built-in (direct import) and external entry-point providers. Side-effect-free: no `set_global_provider`, no `sys.exit`; the caller decides error handling (CLI exits, the API can return HTTP 400).
- Refactor `init_global_provider` to use it, preserving CLI behavior: a built-in's missing transitive dependency still logs critical + `sys.exit(1)`. The collision warning (built-in wins over a same-named plug-in) stays here, at the point of use, so `prowler --help` no longer emits it spuriously.
- Refactor `get_providers_help_text` to use it.

### Steps to review

- `uv run pytest tests/providers/external/test_dynamic_provider_loading.py -v` (incl. the `TestGetClass` suite).
- `Provider.get_class("aws")` returns `AwsProvider`; an external entry-point provider resolves to its class; an unknown name raises `ImportError` (never `sys.exit`).
- `prowler --help` emits no spurious collision warnings.

### Checklist

- [ ] Review if code is being documented following the Google Python style guide.

#### SDK/CLI

### License

## Release Notes

* **New Features**
  * Added a new public method to resolve provider classes by name for both built-in and external dynamically loaded providers.

* **Bug Fixes**
  * Improved provider class resolution behavior and error handling, including clearer handling when built-in providers are missing expected classes and when names collide with external providers.
  * Standardized how provider help text is discovered across built-in and external providers.

* **Tests**
  * Expanded coverage for dynamic provider loading, collision scenarios, and resolver purity/side-effect behavior.

* **Documentation**
  * Updated the changelog entry to reflect the new resolver method.


Alan-TheGentleman (commented):
This resolver refactor is well-contained, and I do not see a code-level blocker here.

Two follow-ups:
- `Provider.get_class()` says “no global state”, but the external-provider path still mutates the `_ep_providers` cache. I’d reword this as “no global provider initialization / no sys.exit” unless strict purity is intended.
- The unknown-provider test should assert `ImportError` specifically instead of accepting `(ImportError, Exception)`, so the public API contract is enforced.

Please keep this stacked behind #10700, then rebase or retarget onto `master` and rerun the full SDK CI before merging.

josema-xyz (commented):
👍 to @Alan-TheGentleman's:
> * The unknown-provider test should assert `ImportError` specifically instead of accepting `(ImportError, Exception)`, so the public API contract is enforced.

StylusFrost:
@Alan-TheGentleman both follow-ups addressed in [`a111ae763`](https://github.com/prowler-cloud/prowler/commit/a111ae763).

- **`get_class()` docstring** → reworded. It no longer claims "no global state"; it now states that it does not call `sys.exit` and does not initialize the global provider, while noting it may populate the `_ep_providers` memoization cache. Strict purity was not intended; that cache is the one deliberate side effect.
- **Unknown-provider test** → now asserts `ImportError` specifically (`pytest.raises(ImportError)`) instead of `(ImportError, Exception)`, enforcing the public API contract.

On the process note: this stays stacked behind #10700. #10700 was just brought up to date from `master` (merged rather than rebased, to preserve the commit SHAs referenced in its review threads), and this branch now carries that merge too. The full SDK CI runs again before merge.

StylusFrost:
@josema-xyz done in [`a111ae763`](https://github.com/prowler-cloud/prowler/commit/a111ae763). The unknown-provider test now asserts `ImportError` specifically instead of `(ImportError, Exception)`, enforcing the public API contract.

josema-xyz (dismissed):

AdriiiPRodri (dismissed):

Davidm4r (dismissed):

api/src/backend/api/tests/test_views.py
```diff
@@ -9570,6 +9570,150 @@ def test_compliance_overview_attributes_technique_details(
...
+        assert fallback_check_ids
+
+        # Valid UUID with no matching scan -> Scan.DoesNotExist.
+        assert (
+            collect_check_ids("00000000-0000-4000-8000-000000000000")
+            == fallback_check_ids
+        )
+        # Malformed scan id -> ValidationError / ValueError.
+        assert collect_check_ids("not-a-uuid") == fallback_check_ids
+
```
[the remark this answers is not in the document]
StylusFrost: The cross-tenant concern was valid and uncovered a real bug, though the expected outcome is fallback (HTTP 200), not 404.

This endpoint does not 404 on an unresolvable scan: it degrades to the first provider that declares the framework. The actual gap was that the new `Scan.objects.get(id=scan_id)` lacked a tenant filter, unlike every other query in this viewset (`get_queryset`, `_compliance_summaries_queryset`, `list` all filter by `self.request.tenant_id`). The handler runs outside the request `rls_transaction`, so isolation here depends on that explicit filter, not on the RLS session variable. Without it, another tenant's scan could drive provider resolution.

Fixed in [0f9016b](https://github.com/prowler-cloud/prowler/commit/0f9016b9c6b3a3fd73c1d39ed2d1fcac3c1bcb63):

- Scoped the lookup with `tenant_id=self.request.tenant_id`
- Added `test_compliance_overview_attributes_cross_tenant_scan_is_isolated`, which creates a scan in an isolated tenant and asserts the response matches the provider-agnostic fallback, with a guard asserting that a leak would have produced different check IDs

Alan-TheGentleman (commented):
### `prowler/providers/common/provider.py:354` — `AttributeError` fallthrough is dead code and untested

The `AttributeError` fallthrough to entry points is unreachable for every current built-in (`provider.capitalize()` matches the class name in all cases), so it's effectively dead code with no test covering it.

Beyond being untested, the behavior is a bit incoherent: if a built-in module imports but lacks the expected class, falling back to a same-named entry point would return an external class right after `is_builtin` said the provider IS built-in. Worth either dropping the branch or asserting the built-in contract and raising directly, plus a test for the case.

### `tests/providers/external/test_dynamic_provider_loading.py` — test name doesn't match what runs

In `test_..._still_resolves_builtin_via_get_class`, `is_builtin("fakepure")` hits the real `find_spec` and returns `False`, so `get_class` never reaches `import_module` and the `mock_import` side effect is dead. The name says "builtin" but it exercises the entry-point path. Renaming it (or actually mocking `is_builtin=True`) would make the intent match the assertions.

StylusFrost:
@Alan-TheGentleman both addressed in [`6d40c6f`](https://github.com/prowler-cloud/prowler/commit/6d40c6fbfda90ec47bdae5a8211a0fb5a4c1e464).

- **`AttributeError` fallthrough** dropped. `get_class` now raises `ImportError` on a built-in module missing its `<Name>Provider`, instead of falling back to a same-named entry point. New test asserts the raise and `_load_ep_provider.assert_not_called()`.
- **Test renamed** to `..._still_resolves_external_via_get_class`, since `is_builtin("fakepure")` is `False` and it always ran the entry-point path. Dropped the dead `import_module` mock.

prowler/CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ All notable changes to the **Prowler SDK** are documented in this file.
 - Okta network zone check to detect whether anonymized proxy traffic is blocked [(#11463)](https://github.com/prowler-cloud/prowler/pull/11463)
 - Okta API token checks for super admin ownership and network zone restrictions [(#11464)](https://github.com/prowler-cloud/prowler/pull/11464)
 - Support for external/custom providers, checks, and compliance frameworks without modifying core code [(#10700)](https://github.com/prowler-cloud/prowler/pull/10700)
+- Public `Provider.get_class()` method that resolves a provider class by name for both built-in and external (entry-point) providers [(#11398)](https://github.com/prowler-cloud/prowler/pull/11398)
```
AdriiiPRodri: Small changelog placement issue: this entry landed under `5.30.0`, we need to move it to next minor

prowler/providers/common/provider.py
```diff
@@ -693,30 +676,60 @@ def _load_ep_provider(name: str):
...
+                # broken built-in contract — raise rather than fall back to a
+                # same-named external plug-in, which would contradict
+                # is_builtin and silently return a foreign class.
+                raise ImportError(
+                    f"Built-in provider '{provider}' module "
+                    f"'{provider_class_path}' does not define expected class "
+                    f"'{provider_class_name}'"
+                ) from error
+
+        cls = Provider._load_ep_provider(provider)
```
AdriiiPRodri, line 719: Since `get_class()` is now a public resolver and its docstring says it returns a `Provider` subclass, should we validate the class returned by entry points before returning it? Right now `_load_ep_provider()` can return any object from `ep.load()`, including a class that does not inherit from `Provider`, and `get_class()` will still return it as successful resolution. A small `issubclass(cls, Provider)` guard plus a regression test would make the public contract explicit

AdriiiPRodri (changes requested):

josema-xyz (dismissed):

HugoPBrito (commented):
I believe this resolver is too dependent on an implicit naming convention as-is.

For built-in providers, `Provider.get_class` currently guesses the class name from the provider string with something like `f"{provider.capitalize()}Provider"`. This may work today, but it makes the contract structural and fragile for acronyms, compound provider names, or future providers whose Python class does not map cleanly from the provider id.

Can we decide whether this naming convention is the contract we want to support? If not, I believe we should investigate an explicit alternative, for example a registry map, module introspection to find the concrete `Provider` subclass, or an exported `PROVIDER_CLASS`.

If we keep the convention, can we add a parametrized contract test over all built-in providers so future provider additions must remain resolvable through `Provider.get_class`?

prowler/providers/common/provider.py
StylusFrost: Done in [`a1ae8322f`](https://github.com/prowler-cloud/prowler/commit/a1ae8322fea6e083b7c47a99f2b989702bb2bf0b). `get_class` now raises `ImportError` when an entry point resolves to anything that is not a `Provider` subclass, so the public contract is enforced instead of trusted. Added `test_get_class_external_ep_not_provider_subclass_raises_importerror` covering it.

prowler/CHANGELOG.md
StylusFrost: Done in [`a1ae8322f`](https://github.com/prowler-cloud/prowler/commit/a1ae8322fea6e083b7c47a99f2b989702bb2bf0b). Moved the `Provider.get_class()` entry to the `5.31.0` unreleased section.

StylusFrost:
@HugoPBrito the naming convention only affects builtin providers and predates this PR; external providers already resolve explicitly via their entry point. Kept it and pinned it with a parametrized contract test over every builtin in [`4127e3329`](https://github.com/prowler-cloud/prowler/commit/4127e33298f91e0f386c98167c0e008a9d933b8e): each one must resolve through `Provider.get_class` and return a `Provider` subclass whose name follows the convention, so a future provider that breaks the mapping fails in CI instead of at runtime in a caller like the API. The guessing is transitional: once builtins migrate to dynamic discovery they resolve via their own entry point like external providers, which removes the convention. The contract test stays useful there as a guardrail, the resolution and subclass checks still hold and only the name assertion gets relaxed, so no separate registry is planned.

AdriiiPRodri (dismissed):

danibarranqueroo (approved):

PR #11399, opened 2026-05-31
feat(api): store provider as varchar validated against available providers

StylusFrost:
### Context

The API pinned the set of valid providers to a static Django enum backed by a native PostgreSQL enum column, drifting from the SDK, which discovers both built-in and external providers. This makes the SDK the single source of truth for provider validity and removes the native enum.

### Scope
- Includes: enum→varchar expand–contract migration (`0094`–`0097`), SDK-based validation at model + serializer, SDK-driven provider-type filters
- Excludes: dynamic class resolution in `return_prowler_provider`

### Autonomy
- [ ] Migration is zero-downtime (expand–contract) and tests cover this unit

### Description

- Migrate `Provider.provider` from a native PostgreSQL enum to `varchar` with a zero-downtime expand–contract migration: shadow column synced by a trigger (`0094`), a per-tenant batched backfill (`0095`), and a contract swap that drops the enum column/type and rebuilds the partial unique index concurrently (`0096`/`0097`).
- Validate provider input against the SDK's available providers (built-in or external) at both the model `clean()` and serializer layers. Providers without a `validate_<provider>_uid` method fall back to the field-level min-length check.
- Drive provider-type filters from the SDK's available providers instead of the static enum.

The `provider` column name is preserved throughout, and `varchar` accepts the same string values the enum held, so instances running the previous release keep working against the swapped column.

### Steps to review

**Local testing:** this depends on SDK changes not yet on `master`, so point the `prowler` dependency in `api/pyproject.toml` at this PR's branch (not `@master`), regenerate `uv.lock`, and rebuild the API image (revert before merging):

```
prowler @ git+https://github.com/prowler-cloud/prowler.git@PROWLER-1772-provider-type-storage-varchar
```

### Checklist

#### API

### License


Alan-TheGentleman (changes requested):
I'm requesting changes because the migration sequence is not safe as written.

`0095` only enqueues the `provider_str` backfill, but `0096` immediately drops/renames columns and sets `provider` to `NOT NULL`. In a normal deploy, Celery may not have completed, or may not be running, before `0096` executes.

There is also a uniqueness gap: the existing unique provider index is dropped before the concurrent rebuild, which leaves a write window where duplicate active providers can be inserted.

Please split this into a safe expand/backfill/contract sequence, or make the backfill synchronous and verified before the destructive migration. Also, provider validation should stay aligned with what connection/scan lifecycle code can actually initialize, since the API now accepts SDK/external providers while parts of the lifecycle still rely on static `ProviderChoices`.

api/src/backend/api/v1/serializers.py
```diff
@@ -854,7 +855,12 @@ class Meta:
 # Providers
 class ProviderEnumSerializerField(serializers.ChoiceField):
     def __init__(self, **kwargs):
-        kwargs["choices"] = Provider.ProviderChoices.choices
+        # The SDK is the source of truth for which providers exist, so the
+        # accepted values track the installed providers (built-in or external)
+        # instead of a static enum.
+        kwargs["choices"] = [
+            (name, name) for name in SDKProvider.get_available_providers()
```
josema-xyz: The filter caches this list with `lru_cache`, but here it's recomputed on every request (the field gets rebuilt each time). Could we reuse the cached one?

StylusFrost: Done in [`c94456c13`](https://github.com/prowler-cloud/prowler/commit/c94456c13). Moved `get_provider_type_choices` into a leaf module (`api/provider_types.py`) so this serializer field and the filters share one `lru_cache`d source instead of recomputing SDK provider discovery on every request.

One caveat worth flagging: that cache is process-lifetime, so a provider installed at runtime would not appear here until a restart. Supporting hot-install needs coordinated cache invalidation (the SDK entry-point caches, `importlib`, and this list), tracked as a separate follow-up.

StylusFrost:
Points 1 and 2 fixed in [`28433362c`](https://github.com/prowler-cloud/prowler/commit/28433362c):

- **Backfill race**: `0095` now backfills `provider_str` synchronously with a single `UPDATE` (the providers table is ~6k rows, well within the safe inline threshold), so the column is fully populated before `0096` sets it `NOT NULL`. No dependency on Celery having run.
- **Uniqueness window**: the partial unique index is now dropped and recreated inside `0096`'s single transaction (with a short `lock_timeout`), so there is no window where duplicate active providers can be inserted. Migration `0097` and the now-unused `backfill_provider_str` task were removed.

On point 3 (validation aligned with what the lifecycle can initialize): the serializer accepts any SDK-available provider, but the connection/scan lifecycle still dispatches through the static `match` in `api/utils.py`. Moving that onto the dynamic SDK resolver is exactly what the next stacked PRs do (#11400 dynamic provider resolution, #11401 dynamic kwargs + connection test), so I kept the alignment there rather than adding a temporary guard here that those PRs would then remove. Happy to add an interim guard if you would rather not accept external types until that lands.

Related: provider-type choices are cached for the process lifetime, so hot-installing a provider needs coordinated cache invalidation before it is usable without a restart. Tracking that as its own follow-up.

josema-xyz (approved):

api/src/backend/api/v1/serializers.py
```diff
@@ -940,6 +943,12 @@ def get_connection(self, obj):
 
 
 class ProviderCreateSerializer(RLSSerializer, BaseWriteSerializer):
+    # Declared explicitly so provider validation stays at the serializer layer:
+    # the model column is now a plain varchar with no choices, so without this
+    # an unknown provider would slip through to Provider.clean() instead of
+    # being rejected here with `invalid_choice`.
+    provider = ProviderEnumSerializerField()
```
AdriiiPRodri, line 950: This now accepts any SDK-available provider, but the API lifecycle still cannot initialize all of them. For example, `llm` is accepted here, but `return_prowler_provider()` in `api/utils.py` still dispatches through the static `ProviderChoices` match and raises `ValueError: Provider type llm not supported`

If this PR can land independently from the follow-up dynamic lifecycle PRs, we should either keep an interim allowlist aligned with what connection/scan can actually initialize, or move the lifecycle dispatch in this PR as well. Otherwise the API can create providers that immediately fail in secrets/connection/scan flows

AdriiiPRodri (changes requested):
There is still one static provider validation path in `OverviewViewSet._extract_provider_filters_from_params()`:

`api/src/backend/api/v1/views.py:5419`

It builds `valid_provider_types` from `Provider.ProviderChoices`, so `overview categories` and `overview/resource-groups` reject SDK-valid providers such as `llm`, while the regular filtersets now use `get_provider_type_choices()`. This should use the same cached provider choices as the filters/serializer to keep provider_type behavior consistent across endpoints

PR #11402, opened 2026-05-31
feat(api): validate external provider secrets via SDK credential schema

StylusFrost:
### Context

`validate_secret_based_on_provider` mapped each provider type to a hardcoded `*ProviderSecret` serializer, raising `Provider type not supported` for anything else. External providers could not have their credentials validated, so a secret could not be created for them.

Stacked on #11401 (T4). Adds the SDK contract surface for credential schemas **without touching the built-in providers**.

### Description

- Add `get_credentials_schema()` to the SDK provider contract (`prowler/providers/common/provider.py`). It returns a pydantic model describing the provider's credentials, or `None`.
- In the API, non-built-in providers validate their secret against that schema (`model_validate`); when a provider declares no schema, the secret is accepted as-is and validated later by `test_connection`. **Built-in providers keep their explicit serializers and the `oneOf` schema unchanged.**

### Steps to review

**Local testing:** this depends on SDK changes not yet on `master`, so point the `prowler` dependency in `api/pyproject.toml` at this PR's branch (not `@master`), regenerate `uv.lock`, and rebuild the API image (revert before merging):

```
prowler @ git+https://github.com/prowler-cloud/prowler.git@PROWLER-1775-dynamic-credential-validation
```

### Out of scope (follow-ups)

- Dynamic `oneOf` generation for external credential shapes (built-in `oneOf` unchanged; documented fallback).
- Decoupling the `SERVICE_ACCOUNT` secret type from GCP.

### Checklist

### License


Alan-TheGentleman (changes requested):
I'm requesting changes for the no-schema external provider path.

When `get_credentials_schema()` returns an empty list, the serializer accepts `secret` without checking that it is a JSON object. Since `ProviderSecretField` is a raw JSONField and the SDK defaults later do `{**secret}`, a list, string, or null can be persisted and fail later during connection or scan with a server error.

Please validate that external provider secrets are objects before accepting the no-schema path, and add serializer or JSON:API create/update coverage for list/string/null payloads.

api/src/backend/api/v1/serializers.py
```diff
@@ -1544,10 +1545,48 @@ class Meta:
 
 # Provider secrets
 class BaseWriteProviderSecretSerializer(BaseWriteSerializer):
+    @staticmethod
+    def _validate_external_provider_secret(provider_type: str, secret: dict):
```
josema-xyz: Design question: `_validate_external_provider_secret` doesn't receive `secret_type`, so it validates against every schema and accepts on any match. That means the `secret_type` the user picks isn't tied to its schema (you can store one type's shape under another type's label). Is that intentional for now, or should we map `secret_type` → `schema`?

josema-xyz (commented):
+1 to @Alan-TheGentleman's point. The schema branch is already safe: `model_validate` rejects a non-object `secret` and that comes back as a `400`. The gap is only the no-schema branch (`if not schemas: return`), which accepts `secret` without checking it's a `JSON` object, so a `list` / `string` / `null` slips through and blows up later at `{**secret}`.

StylusFrost:
@Alan-TheGentleman both points addressed. External secrets are validated as JSON objects before the no-schema path accepts them ([`116fb7083`](https://github.com/prowler-cloud/prowler/commit/116fb7083)), with parametrized coverage for list/string/null/int. And a secret is now validated against the schema for its `secret_type` rather than any declared schema ([`5daa39c4c`](https://github.com/prowler-cloud/prowler/commit/5daa39c4c)), the binding you and @josema-xyz flagged.

api/src/backend/api/v1/serializers.py
StylusFrost: Implemented in [`5daa39c4c`](https://github.com/prowler-cloud/prowler/commit/5daa39c4c). `get_credentials_schema` now returns the schemas keyed by secret type, and the external path validates a secret against the schema for the `secret_type` it is created with, not against any declared schema. A role-shaped secret under `secret_type=static`, or a `secret_type` the provider does not declare, is now rejected. The provider declares which category (`static`/`role`/`service_account`) each schema belongs to, so the type stays bound to its shape.

There is a regression test for the role-shaped-under-static case.

josema-xyz (approved):

PR #11406, opened 2026-06-01
fix(api): change Finding.uid from varchar(300) to TextField

### LINKED ISSUE #7535: Increase `finding.uid` max lenght > 300 characters
### Steps to Reproduce

As discussed in this Prowler Community Slack 🧵 https://prowler-workspace.slack.com/archives/C045DAM2NE9/p1740529388847369

Tried to fix it in https://github.com/prowler-cloud/prowler/pull/7048 and https://github.com/prowler-cloud/prowler/pull/7498 but got issues with migrations since it requires to adquire an `ACCESS EXCLUSIVE` lock for the `findings` table. We need to try a different approach.

### Expected behavior

We need to change the finding.uid field type from `varchar` to text and remove the 300 chars limit or increase the `varchar` size to more than 300 chars.

N/A

### Environment Resource

N/A

### OS used

N/A

### Prowler version

> 5.0

### Pip version

N/A

### Context

_No response_

AlexanderSanin:
## Summary

- Changes `Finding.uid` model field from `CharField(max_length=300)` to `TextField()`, removing the hard 300-character cap that caused scan ingestion failures for resources with long ARNs, deeply-nested GCP resource names, composite Kubernetes UIDs, or other multi-segment cloud identifiers.
- Adds migration `0094_finding_uid_to_textfield` using `SeparateDatabaseAndState` + `atomic=False` + raw `ALTER TABLE` SQL, consistent with other partition-aware schema changes in this project. PostgreSQL does not rewrite rows when relaxing a varchar constraint to text (both use the same varlena storage), so the `ACCESS EXCLUSIVE` lock duration is minimal.
- Re-enables the previously commented-out `TestFindingModel.test_add_finding_with_long_uid` test that validates a 500-character UID can be stored.

Closes #7535

## Test plan

- [x] `TestFindingModel.test_add_finding_with_long_uid` — creates a `Finding` with a 500-char UID and asserts it is persisted; previously disabled because the varchar(300) cap would raise a `DataError`.
- [x] Verify `api/src/backend/api/migrations/0094_finding_uid_to_textfield.py` is picked up by `manage.py migrate` without errors.
- [x] Verify existing tests continue to pass (no `max_length` validator on `Finding.uid` to trip over).

### DIFF
--- a/api/CHANGELOG.md
+++ b/api/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to the **Prowler API** are documented in this file.
 
 ## [1.28.0] (Prowler UNRELEASED)
 
+### 🐞 Fixed
+
+- `Finding.uid` column type changed from `varchar(300)` to `text`, removing the 300-character cap that caused scan ingestion failures for resources with long ARNs or composite identifiers [(#7535)](https://github.com/prowler-cloud/prowler/issues/7535)
+
 ### 🚀 Added
 
 - GIN index on `findings(categories, resource_services, resource_regions, resource_types)` to speed up `/api/v1/finding-groups` array filters [(#11001)](https://github.com/prowler-cloud/prowler/pull/11001)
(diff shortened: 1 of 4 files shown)


AlexanderSanin:
Hey @danibarranqueroo @josema-xyz @HugoPBrito. Could you, please, have a look at this?

PR #11416, opened 2026-06-01
feat(api): recover orphaned background tasks and make task re-runs idempotent

AdriiiPRodri:
### Context

Background work in the API runs on Celery workers: scans, provider and tenant deletions, integration sends, report generation, and so on. When a worker dies in the middle of a task, that work is simply lost. This happens more often than it sounds, because every deploy rolls the worker pods, and a worker can also be OOM-killed or evicted from its node. The task that was running leaves its result row in a non-terminal state, so a scan sits "in progress" and never finishes, a deletion stops halfway, an integration never sends. Nothing notices and nothing retries it, so today the only way out is to go in by hand.

This PR makes background work survive a worker dying. It works on two fronts. First, it hardens the Celery setup so fewer tasks are lost in the first place. Second, it adds a watchdog that detects the ones that still slip through and re-runs them. Re-running a task is only safe if running it twice does not duplicate or corrupt anything, so the PR also closes the places where a second run would have caused damage, namely duplicate Jira issues, duplicate scan findings, and duplicate compliance rows.

### Description

The change has three parts that build on each other.

**1. Stop losing tasks in the first place** (`config/celery.py`, `docker-entrypoint.sh`, both compose files)

- Durable delivery. With `task_acks_late` and `task_reject_on_worker_lost`, the broker keeps a task's message until the task actually finishes, and puts it back on the queue if the worker is lost mid-run. The worker prefetch is lowered to 1 so that a crash leaves at most one reserved message stuck behind a dead worker instead of several, which also stops a busy worker from hoarding messages that an idle one could pick up.
- Graceful shutdown. A soft shutdown timeout (60s) gives a worker that receives `SIGTERM` on a deploy time to finish or re-queue its in-flight tasks before it exits. For that signal to reach Celery at all, the container entrypoint now `exec`s the Celery, Gunicorn, and Django processes so they run as PID 1 instead of as a child of the entrypoint shell that would otherwise swallow the signal. The worker services in both compose files get `stop_grace_period: 120s` so the orchestrator waits for that drain rather than force-killing the worker.
- Execution limits. Every task now has a soft and hard time limit so a hung task cannot hold a worker forever. Connection checks are capped tight (120s) because they should always be quick. Scans and provider/tenant deletions can legitimately run for a long time on large accounts, so they get a much higher cap (48h) instead of being cut short at the default limit.

There are two timeouts in play here and it is worth being explicit about how they relate. The per-task time limit caps how long a task may run before it is killed. The broker visibility timeout (24h) is how long the broker waits before it assumes an unacknowledged task was lost and hands its message to another worker. Because `task_acks_late` holds the acknowledgement until the task finishes, a task that runs longer than the visibility timeout can be delivered a second time. The vast majority of tasks finish well within 24h. The long-running exceptions are scheduled scans, and a re-delivered scheduled scan is a no-op thanks to the existing "a scan is already executing" guard, so it does not start a second run. If a given deployment routinely runs other tasks past 24h, the visibility timeout can be raised above the long-task limit.

**2. Recover the tasks that still get orphaned** (`tasks/jobs/orphan_recovery.py`, `tasks/tasks.py`, the `reconcile_orphan_tasks` command, migrations 0094 and 0095)

- A new shared engine detects and recovers orphaned tasks. A task is "orphaned" when its result row is still `STARTED`/`RECEIVED` but the worker recorded on it no longer answers a ping. Real orphans have their stale result marked terminal, which clears the "stuck in progress" state and the alerts that watch for it, and, when the task is one that is safe to run again, their work is re-queued from the task name and arguments stored on the result row.
- It runs two ways. Automatically, as a Celery Beat task every 2 minutes (registered by migration 0095). And on demand, through the `reconcile_orphan_tasks` management command, which takes `--dry-run`, `--grace-minutes`, and `--max-attempts` for manual recovery and for inspecting what would happen without touching anything.
- Scans are recovered by re-running the scan on its existing row rather than re-queuing the original scheduled task. Re-queuing the scheduled task would hit its "a scan is already executing" guard, no-op, and leave the scan stuck. A `recovery_count` column on `Scan` (migration 0094) records how many times a scan has been recovered, for visibility.
- The attack-paths stale cleanup now reuses the worker-ping and revoke helpers from this new module instead of keeping its own near-identical copies.

A note on cost, since the 2-minute cadence raises the question: a run with nothing to recover is cheap. It is a couple of index-backed queries plus a quick ping to each worker that currently holds an in-flight task. The task table is indexed on `status`, and the only rows the sweep looks at are the few that are still in a non-terminal state, so the query stays small no matter how much history the table accumulates. The expensive part, revoking and re-queuing, only runs when there is an actual orphan, and runs never overlap because each one either takes the advisory lock or backs off. If a deployment ever wants to sweep less often, the interval is configurable on the periodic task.

**3. Make re-running safe** (`api/models.py`, `tasks/jobs/integrations.py`, `tasks/jobs/scan.py`, `tasks/jobs/deletion.py`, migration 0096)

- A new `JiraIssueDispatch` table records which findings have already been ticketed for a given integration, keyed by tenant, integration, and finding. Each finding is reserved in this table before the Jira call, so the unique constraint serializes concurrent or retried sends: only the run that creates the dispatch row opens the issue, and any other run skips that finding. That is what keeps recovery from ever opening a duplicate issue. The reservation is rolled back if the send fails, so the finding can be retried later. The one residual window is the opposite of a duplicate: if a worker is hard-killed in the brief moment between the reservation being committed and the issue being created, that single in-flight finding is skipped on the next run and its ticket is missed. This is the safer side to err on for a "do not duplicate issues" goal, and it is at most one finding per crash. These rows are cleaned up when the provider they belong to is deleted.
- A recovered scan rewrites its data instead of appending to the previous run. Before re-inserting, it clears its own findings and compliance rows. Each table needs this for a different reason: findings have no unique key and would simply be duplicated, compliance summaries are inserted with `ignore_conflicts` which would keep the stale row, and compliance requirements are inserted with PostgreSQL `COPY`, which cannot upsert.

Recovery re-enqueues only the tasks that are proven safe to run again: scans and their summary and compliance aggregations, provider and tenant deletions, and Jira sends, which the dispatch table above makes idempotent. Tasks whose re-run cannot be made safe are left terminal rather than re-run: the S3 upload rebuilds from worker-local files that do not survive the crash, and report generation and Security Hub uploads are out of scope for now. In every case the stale result is still marked terminal, so the "stuck in progress" alert clears either way.

**Safety and failure modes**

The recovery path is built to never double-run a task that is actually still alive. It only acts on a task whose worker fails to answer a ping, it errs on the side of "alive" if the control bus is unreachable, and the whole sweep runs under a Postgres advisory lock so only one reconciliation can run at a time across replicas, the watchdog, and the command. Re-enqueuing a scan pre-creates the new task linkage and then relinks the scan only if it still points at the old task, so a late orphan can never clobber a newer run. The retry counter lives in Valkey and survives the task result row being rewritten; a task that keeps killing its worker is marked failed after `--max-attempts` instead of looping forever. If Valkey or the broker is unreachable the counter fails open, on the basis that an unreachable broker means nothing is running anyway. The watchdog has no tenant context, so it reads across tenants through the admin connection and performs each re-enqueue inside the owning tenant's RLS transaction.

**Configuration**

All of the new behavior has safe defaults and is overridable through environment variables, documented in `api/docs/orphan-task-recovery.md`: `DJANGO_CELERY_WORKER_PREFETCH_MULTIPLIER` (1), `DJANGO_CELERY_WORKER_SOFT_SHUTDOWN_TIMEOUT` (60s), `DJANGO_CELERY_TASK_TIME_LIMIT` (6h) and its soft counterpart, and `DJANGO_CELERY_LONG_TASK_TIME_LIMIT` (48h) for scans and deletions. The watchdog interval, grace window, and retry cap are tunable on the command and on the periodic task.

### Steps to review

A good reading order is `config/celery.py` and `docker-entrypoint.sh` for the prevention layer, then `tasks/jobs/orphan_recovery.py` for the recovery engine, then the idempotency changes in `tasks/jobs/integrations.py` and `tasks/jobs/scan.py`, and finally the migrations and the `reconcile_orphan_tasks` command. `api/docs/orphan-task-recovery.md` gives the high-level picture and the configuration knobs.

To exercise it locally:

- Run the new and updated tests: `pytest tasks/tests/test_orphan_recovery.py tasks/tests/test_scan.py tasks/tests/test_integrations.py tasks/tests/test_deletion.py`.
- Inspect without touching anything: `python manage.py reconcile_orphan_tasks --dry-run`, which reports the tasks that would be recovered.
- See recovery end to end: start a scan, kill the worker mid-run, and confirm that within the watchdog interval the scan is re-queued and completes, and that its findings are not duplicated.
- Confirm a still-running task is left alone: while a long task is running on a healthy worker, run the command and check that the task is reported as in-flight and is not re-queued.

### Checklist

<details>

</details>

#### SDK/CLI

#### API
- [ ] Endpoint response output (if applicable): not applicable, no endpoints change
- [ ] EXPLAIN ANALYZE output for new/modified queries or indexes (if applicable): not applicable
- [ ] Performance test results (if applicable): not applicable
- [x] Any other relevant evidence of the implementation (if applicable): see Steps to review
- [ ] Verify if API specs need to be regenerated: not applicable, no API surface change
- [ ] Check if version updates are required (e.g., specs, uv, etc.): not applicable

### License


api/src/backend/config/celery.py
```diff
@@ -26,6 +26,58 @@
...
+    "DJANGO_CELERY_TASK_SOFT_TIME_LIMIT", default=_TASK_HARD_LIMIT - 600
+)
+_LONG_TASK_HARD_LIMIT = env.int(
+    "DJANGO_CELERY_LONG_TASK_TIME_LIMIT", default=48 * 60 * 60
+)
+_LONG_TASK_SOFT_LIMIT = env.int(
+    "DJANGO_CELERY_LONG_TASK_SOFT_TIME_LIMIT", default=_LONG_TASK_HARD_LIMIT - 600
+)
+celery_app.conf.task_annotations = {
+    "*": {"soft_time_limit": _TASK_SOFT_LIMIT, "time_limit": _TASK_HARD_LIMIT},
```
josema-xyz: I don't think these limits actually apply. Celery applies every matching rule in order and `"*"` always goes last, so for scan-perform it sets 48h and then `"*"` overwrites it with 6h. Same for the 120s connection checks. Everything ends up at the 6h default, so a long scan gets killed at 6h. A quick test checking the real limit on a scan task would catch this.

AdriiiPRodri: Thanks @josema-xyz , changed here: https://github.com/prowler-cloud/prowler/pull/11416/changes/77e557e50ec84297dc731557b9cbcd890667d396

josema-xyz (dismissed):

josema-xyz (approved):

jfagoagas (approved):

PR #11420, opened 2026-06-02
perf(api): avoid N+1 query loading finding resource tags

Davidm4r:
### Context

`GET /api/v1/findings` triggered an N+1 query: resource tags were loaded one query per resource when serializing findings.

### Description

`FindingViewSet` now defines `_optimize_tags_loading` to prefetch `resources__tags` into `prefetched_tags`. The `PaginateByPkMixin` already invokes this hook, so resource tags are serialized from memory in a single query.

### Steps to review

Call `GET /api/v1/findings?include=resources` and check tags load in one prefetch query instead of one per resource.

### Checklist

<details>

</details>

#### SDK/CLI

#### UI

#### API

### License

### DIFF
--- a/api/CHANGELOG.md
+++ b/api/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 All notable changes to the **Prowler API** are documented in this file.
 
+## [1.31.0] (Prowler UNRELEASED)
+
+### 🐞 Fixed
+
+- `GET /api/v1/findings` N+1 query loading `resources__tags` when listing findings [(#11420)](https://github.com/prowler-cloud/prowler/pull/11420)
+
+---
+
 ## [1.30.0] (Prowler v5.29.0)
 
 ### 🔄 Changed
--- a/api/CHANGELOG.md
+++ b/api/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 All notable changes to the **Prowler API** are documented in this file.
 
-## [1.31.0] (Prowler UNRELEASED)
+## [1.30.1] (Prowler v5.29.1)
 
 ### 🐞 Fixed
 
--- a/api/src/backend/api/v1/views.py
+++ b/api/src/backend/api/v1/views.py
@@ -3754,9 +3754,7 @@ class FindingViewSet(PaginateByPkMixin, BaseRLSViewSet):
         return queryset.prefetch_related(
             Prefetch(
                 "resources__tags",
-                queryset=ResourceTag.objects.filter(
-                    tenant_id=self.request.tenant_id
-                ).select_related(),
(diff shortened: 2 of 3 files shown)


api/src/backend/api/v1/views.py
```diff
@@ -3749,6 +3749,18 @@ def filter_queryset(self, queryset):
             return queryset
         return super().filter_queryset(queryset)
 
+    def _optimize_tags_loading(self, queryset):
```
jfagoagas, line 3764: Where is this function called?

api/CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
 
 All notable changes to the **Prowler API** are documented in this file.
 
+## [1.31.0] (Prowler UNRELEASED)
```
jfagoagas: Add a backport and include it for `1.30.1`
```suggestion
## [1.30.1] (Prowler v5.29.1)
```

api/src/backend/api/v1/views.py
Davidm4r: Good question.
This is called by https://github.com/prowler-cloud/prowler/blob/c2cef99b334ef10c54d78550750640a8becbc241/api/src/backend/api/v1/mixins.py#L67

api/src/backend/api/v1/views.py
```diff
@@ -3749,6 +3749,18 @@ def filter_queryset(self, queryset):
...
+        """Prefetch resource tags to avoid N+1 queries when serializing findings"""
```
josema-xyz, line 3765: This change makes findings load all tags in one query instead of one per resource. That is fine, but nothing tests it. If someone edits this later and breaks it, no one will notice. A small test that counts the queries would catch that.

api/src/backend/api/v1/views.py
```diff
@@ -3749,6 +3749,18 @@ def filter_queryset(self, queryset):
...
+        return queryset.prefetch_related(
+            Prefetch(
+                "resources__tags",
+                queryset=ResourceTag.objects.filter(
+                    tenant_id=self.request.tenant_id
+                ).select_related(),
```
josema-xyz: The empty `select_related()` follows the non-null tenant FK, so it joins tenants and pulls a row we never use (we only read key/value). It's a needless join, and the empty form is deprecated in newer Django. We could drop that.

Davidm4r: I deleted the select_related(). I used it because in another _optimize_tags_loading was already implemented.

api/src/backend/api/v1/views.py
Davidm4r: I created a new test to check that.

josema-xyz (approved):

PR #11472, opened 2026-06-05
feat(api): make orphan-task recovery configurable and drop the Jira idempotency table

AdriiiPRodri:
### Context

This builds on the orphan-task recovery added in #11416, which re-enqueues background tasks whose worker died mid-run (deploy, OOM, eviction). That work shipped without an operational kill switch or any per-area control, relied on a dedicated `jira_issue_dispatches` table for Jira de-duplication, and made scan re-runs idempotent. This PR makes recovery toggleable per task group (opt-in), drops the Jira idempotency table, and removes the scan re-run idempotency so scans are no longer auto-recovered. The whole #11416 feature is still unreleased, so removing the migrations does not affect any released schema.

### Description

**Feature flags for orphan-task recovery.** Recovery is gated by Django settings (environment variables). The master switch is OFF by default, so recovery is opt-in; the per-group flags default to enabled, so once the master is on every group recovers unless explicitly turned off.

| Env var | Default | Scope |
| --- | --- | --- |
| `DJANGO_TASK_RECOVERY_ENABLED` | `false` | Master switch for the whole recovery sweep (opt-in). |
| `DJANGO_TASK_RECOVERY_SUMMARIES_ENABLED` | `true` | Group: `scan-summary`, `scan-compliance-overviews`, `scan-provider-compliance-scores`, `scan-daily-severity`, `scan-finding-group-summaries`, `scan-reset-ephemeral-resources`. |
| `DJANGO_TASK_RECOVERY_DELETIONS_ENABLED` | `true` | Group: `provider-deletion`, `tenant-deletion`. |

The flat reenqueueable allowlist is replaced by `RECOVERY_TASK_GROUPS` (summaries, deletions) plus a `reenqueueable_tasks()` helper that unions only the enabled groups. A task in a disabled group is still detected and marked terminal (clearing the stuck "in progress" state), but it is not re-enqueued. With the master flag off, the task-recovery sweep is skipped entirely; the attack-paths stale cleanup, a separate concern, keeps running.

**Scans excluded from recovery.** The scan re-run idempotency added in #11416 is removed (the pre-run `_clear_scan_rerun_state` delete and the compliance summary/requirement re-deletes), `Scan.recovery_count` and its migration are dropped, and `scan-perform`/`scan-perform-scheduled` are moved into the watchdog's skip set. An orphaned scan is now left untouched (not detected, marked, or re-enqueued), reverting scans to their pre-#11416 behavior, because re-running a scan is not safe to do automatically.

**Remove the Jira idempotency dispatch table.** The `JiraIssueDispatch` model and its `0096_jiraissuedispatch` migration are removed, `send_findings_to_jira` is reverted to its pre-#11416 form, and `integration-jira` is dropped from the reenqueueable allowlist because, without the dedup table, re-running it would create duplicate Jira issues. The dispatch cleanup step is also removed from provider deletion.

**Tasks never re-enqueued.** Only the two groups above are ever re-enqueued. Every other task is detected and marked terminal (so it stops showing as "in progress"), but never re-run, for one of three reasons:

- External side effects, where re-running duplicates or loses work: `integration-jira` (duplicate Jira issues), `integration-s3` (upload rebuilt from worker-local files that do not survive the crash), `integration-security-hub` (pushes findings to AWS), `scan-report`, `scan-compliance-reports` (generate/compress/upload output files from worker-local tmp).
- Ephemeral checks, nothing to recover since they run again on demand or schedule: `integration-check`, `integration-connection-check`, `provider-connection-check`, `lighthouse-connection-check`, `lighthouse-provider-connection-check`, `lighthouse-provider-models-refresh`.
- Not audited for idempotency, kept terminal as a conservative default: `backfill-compliance-summaries`, `backfill-daily-severity-summaries`, `backfill-finding-group-summaries`, `backfill-provider-compliance-scores`, `backfill-scan-resource-summaries`, `scan-attack-surface-overviews`, `scan-category-summaries`, `scan-resource-group-summaries`, `reaggregate-all-finding-group-summaries`, `findings-mute-historical`.

Some tasks are skipped entirely (not even detected): `scan-perform` and `scan-perform-scheduled` (not auto-recovered), `attack-paths-scan-perform` (handled by its own stale cleanup, which drops the temporary Neo4j database), and `attack-paths-cleanup-stale-scans` and `reconcile-orphan-tasks` (they re-run on their own schedule).

### Migration cleanup

This PR deletes two unreleased migrations: `0094_scan_recovery_count` (added `Scan.recovery_count`) and `0096_jiraissuedispatch` (created `jira_issue_dispatches`). Neither shipped in a release, so no released schema is affected. If an environment already applied either one while tracking `master`, the column/table and their migration records are left behind on pull; drop them manually:

```sql
ALTER TABLE scans DROP COLUMN IF EXISTS recovery_count;
DROP TABLE IF EXISTS jira_issue_dispatches;
DELETE FROM django_migrations WHERE app = 'api' AND name IN ('0094_scan_recovery_count', '0096_jiraissuedispatch');
```

### Steps to review

1. Recovery is OFF by default: the master flag `DJANGO_TASK_RECOVERY_ENABLED` defaults to `false`, so the sweep does nothing until you set it to `true`.
2. With the master on, set a group flag to `false` (for example `DJANGO_TASK_RECOVERY_SUMMARIES_ENABLED=false`) to exclude that group; its orphaned tasks are then marked terminal instead of re-enqueued.
3. Scans are no longer recovered: an orphaned `scan-perform`/`scan-perform-scheduled` is ignored (not detected, marked, or re-enqueued).
4. Settings-only change plus migration deletions, so no new migration is required; `python manage.py makemigrations --check` reports no changes.
5. Tests: `pytest tasks/tests/test_orphan_recovery.py` (master/per-group flags and the scan skip) and `pytest tasks/tests/test_integrations.py -k Jira` (Jira send reverted).

### Checklist

#### API
- [ ] Endpoint response output (if applicable) - N/A, no endpoint changes
- [ ] EXPLAIN ANALYZE output for new/modified queries or indexes (if applicable) - N/A
- [ ] Performance test results (if applicable) - N/A
- [x] Verify if API specs need to be regenerated - not needed, no endpoint or schema changes
- [ ] Check if version updates are required - covered by the existing unreleased section

### License


api/docs/orphan-task-recovery.md
```diff
@@ -13,29 +14,34 @@ see a stuck scan and pending-task alerts do not fire.
    (`worker_prefetch_multiplier = 1`), and an abruptly-lost worker re-queues its task
    (`task_reject_on_worker_lost`). On `SIGTERM` the worker is given a soft-shutdown
    window (`worker_soft_shutdown_timeout`) to finish or re-queue in-flight work
-   before it is force-killed.
+   before it is force-killed. `scan-perform`, `scan-perform-scheduled` and
+   `integration-jira` opt out of redelivery with `acks_late=False`, so a crash drops
+   them rather than re-running and duplicating findings or Jira issues. Other
+   non-recovered side-effect tasks keep `acks_late=True`; a redelivery there rebuilds
+   from worker-local state that did not survive the crash, so it no-ops instead of
+   duplicating.
```
josema-xyz: Two small notes here:
- The recovery flag being off doesn't stop Celery re-running these tasks by itself when a worker dies, that's a separate Celery setting, so I'd say it out loud to avoid confusion.
- Also "rebuilds from worker-local files" isn't right for Security Hub: that one re-reads findings from the DB and sends them to AWS again, no local files involved.

api/src/backend/api/management/commands/reconcile_orphan_tasks.py
```diff
@@ -39,6 +39,13 @@ def handle(self, *args, **options):
             self.stdout.write("Reconcile skipped: another run holds the lock.")
             return
 
+        if result.get("enabled") is False:
+            self.stdout.write(
+                "Task recovery is disabled (DJANGO_TASK_RECOVERY_ENABLED is off); "
+                "no orphans were recovered."
+            )
```
josema-xyz: Heads up that the Attack Paths cleanup still runs even when recovery is off, but this message only mentions recovery being disabled. Worth a quick line so it doesn't look like the command did nothing.

jfagoagas (approved):
This is a great piece of work 👏

PR #11488, opened 2026-06-07
fix(gcp): honour org-aggregated sinks in metric-filter checks

### LINKED ISSUE #11487: GCP: logging_log_metric_filter_and_alert_for_* checks false-FAIL projects covered only by an org-level aggregated sink
### Summary

#11343 reported, and #11355 fixed, `logging_sink_created` false-failing for projects covered only by an **org/folder-level aggregated sink** (`includeChildren=true`). The **same** false-FAIL affects the sibling family — the nine `logging_log_metric_filter_and_alert_for_*` checks — which are still evaluated strictly per-project.

### What

These checks PASS a project only if a matching log metric **and** alert exist **in that project**. But in the centralized topology #11343 describes — an org-level aggregated sink shipping every child project's logs into one bucket, with a single **bucket-scoped** metric + alert covering them all — every child project is reported FAIL, even though its admin/audit activity is fully metric-monitored and alerted.

### Repro

A child project whose logs flow (via an org `includeChildren` sink) to a central bucket that has a matching bucket-scoped metric + alert is reported:

```
FAIL  There are no log metric filters or alerts associated in project <child-project>.
```

(Reproduced as a unit test in the PR — RED→GREEN evidence attached.)

### Why it matters

Same rationale as #11343: false positives on a Google-recommended centralized-logging architecture; users must otherwise duplicate a metric+alert in every project (defeating the aggregated sink) or mutelist the findings (masking real regressions).

### Proposed fix (mirrors #11355)

- add `bucket_name` to the `Metric` model and populate it from `metric.bucketName`;
- add a helper `get_projects_covered_by_aggregated_metric(...)` mapping a bucket-scoped matching metric+alert → the covering org `includeChildren` sink (matched by **destination bucket**, so it never false-*passes*) → the covered projects;
- in each of the 9 checks, mark centrally-covered projects PASS instead of FAIL.

### Affected checks (9)

`audit_configuration_changes`, `bucket_permission_changes`, `compute_configuration_changes`, `custom_role_changes`, `project_ownership_changes`, `sql_instance_configuration_changes`, `vpc_firewall_rule_changes`, `vpc_network_changes`, `vpc_network_route_changes`.

Natural follow-up to #11343 / #11355.

alinealfa:
Closes #11487

### Description

Natural follow-up to #11343 / #11355. #11355 taught `logging_sink_created` to credit org-level aggregated sinks (`includeChildren=true`); this extends the same correction to the nine `logging_log_metric_filter_and_alert_for_*` checks, which are still evaluated strictly per-project.

### Problem

In a centralized topology (the one #11343 describes) an org-level aggregated sink ships every child project's logs into one bucket, where a single **bucket-scoped** log metric + alert covers them all. The metric-filter checks only look for a metric+alert **in each project**, so every child project is reported `FAIL` — even though it is fully metric-monitored and alerted. Users must otherwise duplicate a metric+alert in every project (defeating the aggregated sink) or mutelist the findings.

### Fix

- `Metric` gains `bucket_name`, populated from the API (`metric.bucketName`).
- New helper `get_projects_covered_by_aggregated_metric(logging_client, monitoring_client, metric_filter) -> {project_id: metric_name}`: a project is credited **iff** its organization has an `includeChildren` sink whose **destination bucket** holds a bucket-scoped metric matching the control's filter **and** an alert references it. Matching the sink's destination bucket to the metric's bucket prevents false **passes**.
- Each of the 9 checks credits centrally-covered projects `PASS` ("…covering project X via an organization-level aggregated sink") instead of `FAIL`; each keeps its own original FAIL message.

### Tests

- **Positive** `test_project_centrally_covered_via_org_aggregated_sink` in all 9 checks (RED on master, GREEN after the fix).
- **Negative** `test_aggregated_sink_metric_without_alert_still_fails` in all 9 checks (sink + bucket-scoped metric but no alert → still FAIL; no false-pass).
- **Helper unit suite** `TestGetProjectsCoveredByAggregatedMetric` (guards: no alert / not bucket-scoped / not includeChildren / destination-bucket mismatch / different org) + `test_get_metrics_populates_bucket_name`.
- Full GCP logging + monitoring suite: **97 passed**, no regressions.

### Scope

`logging_service.py` + the 9 `logging_log_metric_filter_and_alert_for_*` checks + their 9 test files + `logging_service_test.py`.

### DIFF
--- a/prowler/CHANGELOG.md
+++ b/prowler/CHANGELOG.md
@@ -17,6 +17,7 @@ All notable changes to the **Prowler SDK** are documented in this file.
 ### 🐞 Fixed
 
 - GCP `logging_sink_created` now recognizes organization-level aggregated sinks with `includeChildren=True`, avoiding false failures for covered projects [(#11355)](https://github.com/prowler-cloud/prowler/pull/11355)
+- GCP `logging_log_metric_filter_and_alert_*` checks now recognize organization-level aggregated sinks with `includeChildren=True`, no longer false-failing projects covered by a central bucket-scoped metric + alert [(#11488)](https://github.com/prowler-cloud/prowler/pull/11488)
 - Jira integration no longer fails with `400 INVALID_INPUT` when a finding has empty fields [(#11474)](https://github.com/prowler-cloud/prowler/pull/11474)
 - GCP `iam_service_account_unused` now passes disabled service accounts instead of failing them, since a disabled account cannot authenticate or be used [(#11467)](https://github.com/prowler-cloud/prowler/pull/11467)
 - AWS AI Security Framework now renders in the dashboard instead of showing "No data found for this compliance", by adding the missing compliance view module [(#11470)](https://github.com/prowler-cloud/prowler/pull/11470)
--- a/prowler/providers/gcp/services/logging/logging_log_metric_filter_and_alert_for_audit_configuration_changes_enabled/logging_log_metric_filter_and_alert_for_audit_configuration_changes_enabled.py
+++ b/prowler/providers/gcp/services/logging/logging_log_metric_filter_and_alert_for_audit_configuration_changes_enabled/logging_log_metric_filter_and_alert_for_audit_configuration_changes_enabled.py
@@ -1,5 +1,8 @@
 from prowler.lib.check.models import Check, Check_Report_GCP
 from prowler.providers.gcp.services.logging.logging_client import logging_client
+from prowler.providers.gcp.services.logging.logging_service import (
+    get_projects_covered_by_aggregated_metric,
+)
 from prowler.providers.gcp.services.monitoring.monitoring_client import (
     monitoring_client,
 )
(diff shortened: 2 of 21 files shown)


HugoPBrito (approved):
I updated this to keep the fix as close as possible to #11355, without expanding the scope of the PR.

The main change is that `get_projects_covered_by_aggregated_metric` now only credits org-level aggregated sinks when they have both `includeChildren=True` and `filter="all"`. That matches the condition already used by `logging_sink_created`, and avoids marking a project as PASS when the aggregated sink could be filtering out the relevant logs.

I also added a negative test for that case, where the sink points to the right bucket but has a restrictive filter, and updated the positive tests so the central logging project is part of the mocked `project_ids`. That keeps the tests aligned with how `_get_metrics()` works today, since it only collects metrics from scanned projects.

This keeps the PR focused on the minimum org-level fix while avoiding a false PASS path.

Thanks a lot @alinealfa for the PR and for opening the issue too! 🚀❤️

PR #11494, opened 2026-06-08
feat(api): label Postgres connections with application_name per component and alias

cesararroba:
**Context**
Postgres connections from the Django app and Celery workers all report the same default `application_name`, so they can't be attributed to a component in `pg_stat_activity`.

**Description**
- Adds `label_postgres_connections()` in `base.py`: tags each Postgres connection (skips the non-Postgres `neo4j` entry, preserves existing `OPTIONS`) with `application_name="<component>:<alias>"`, truncated to Postgres's 63-byte limit.
- `production.py` and `devel.py` call it after `DATABASES` is built.
- The component is injected per process by `docker-entrypoint.sh` via `DJANGO_APP_COMPONENT` (derived from the launch subcommand `$1`), so web=`prod`, workers=`worker`, beat=`beat`. Defaults to `api` when unset.

**Steps to review**
1. `application_name` is libpq-native, passed through by psycopg2 via `OPTIONS`.
2. `SELECT application_name, state, count(*) FROM pg_stat_activity GROUP BY 1,2;` after deploy.
3. `neo4j` alias intentionally untouched.

### DIFF
--- a/api/src/backend/config/django/base.py
+++ b/api/src/backend/config/django/base.py
@@ -306,3 +306,20 @@ SESSION_COOKIE_SECURE = True
 ATTACK_PATHS_SCAN_STALE_THRESHOLD_MINUTES = env.int(
     "ATTACK_PATHS_SCAN_STALE_THRESHOLD_MINUTES", 2880
 )  # 48h
+
+
+def label_postgres_connections(databases):
+    """Tag each Postgres connection with ``application_name="<component>:<alias>"``
+    so connections are attributable in ``pg_stat_activity`` and the Performance
+    Insights "Application" dimension. The component (api / worker / scan / ...) is
+    injected per process by the container entrypoint via ``DJANGO_APP_COMPONENT``;
+    the alias distinguishes which pool inside the process owns the connection.
+    The neo4j entry is skipped (not a Postgres backend). Postgres truncates
+    ``application_name`` at 63 bytes.
+    """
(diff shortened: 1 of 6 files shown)


api/src/backend/config/django/base.py
```diff
@@ -306,3 +306,20 @@
 ATTACK_PATHS_SCAN_STALE_THRESHOLD_MINUTES = env.int(
     "ATTACK_PATHS_SCAN_STALE_THRESHOLD_MINUTES", 2880
 )  # 48h
+
+
+def label_postgres_connections(databases):
```
josema-xyz, line 311: Looks good, but just one thing, this new function has no test. Could you add a small one? Give it a fake list of databases (a Postgres one and the Neo4j one) and check the Postgres one gets the label and Neo4j is left untouched. Mainly so the "skip neo4j" part can't break later without anyone noticing.

cesararroba: Good call. Added `test_db_connection_labels.py` that does exactly that — a fake databases dict with a Postgres alias and the Neo4j one, asserting Postgres gets the `application_name` and Neo4j is left untouched. Threw in a few extra cases too (existing OPTIONS preserved, default component, and the 63-char truncation) so the skip-neo4j branch is locked in. Thanks for the catch!

josema-xyz (approved):

PR #11499, opened 2026-06-08
chore(deps): bump dulwich to 1.2.5 and pyjwt to 2.13.0 for osv-scanner

cesararroba:
### Context

`osv-scanner` flags two transitive/direct security issues in the pinned dependency set:

- **GHSA-897w-fcg9-f6xj** — `dulwich` arbitrary file write. Fixed in `dulwich` 1.2.5 (was pinned at 0.23.0).
- **PYSEC-2026-179** — `pyjwt` HMAC/JWK key-confusion. Fixed in `pyjwt` 2.13.0 (was pinned at 2.12.1).

This bumps both to their patched versions across the SDK and the API.

### Description

- **SDK** (`pyproject.toml`): direct pins `dulwich==0.23.0 → 1.2.5` and `pyjwt==2.12.1 → 2.13.0`; `uv.lock` relocked (only those two packages change).
- **API** (`api/pyproject.toml`): `constraint-dependencies` bumped for both packages, and `override-dependencies` added for `dulwich`/`pyjwt`. The SDK is consumed from `prowler@master`, which still hard-pins the old versions in `[project.dependencies]`; a constraint alone cannot beat a hard pin, so the override forces the patched versions until the SDK bump propagates to the pinned master rev. This mirrors the existing `microsoft-kiota-abstractions` override pattern. `api/uv.lock` relocked (only those two packages change).
- **`osv-scanner.toml`**: refreshed the stale version note in the existing `PYSEC-2025-183`/CVE-2025-45768 ignore reason (now on pyjwt 2.13.0, which also emits an `InsecureKeyLengthWarning` for short HMAC secrets). The advisory remains disputed and unrelated to the two fixes above.

### Steps to review

Validation already performed on this branch:

- SDK `iac` + `github` provider test suites pass (60/60).
- Real `dulwich` exercise: `porcelain.clone` + `ls_remote` work against 1.2.5.
- `pyjwt` 2.13.0 + `djangorestframework-simplejwt` 5.5.1 mint/verify roundtrip works, with the algorithm-confusion guard active.
- `uv lock --check` is clean for both `uv.lock` and `api/uv.lock` (pre-commit `uv-lock` hooks also pass).
- Both the SDK and API Docker images build cleanly — `uv sync --locked` installs `dulwich` 1.2.5 and `pyjwt` 2.13.0.

To re-verify: run `uv lock --check` in repo root and in `api/`, then build both Docker images.

### Checklist

#### SDK/CLI

#### API
- [x] Check if version updates are required (e.g., specs, uv, etc.). — `uv.lock` and `api/uv.lock` relocked; `uv lock --check` clean.

### License


api/pyproject.toml
```diff
@@ -443,7 +443,15 @@ constraint-dependencies = [
...
+# dulwich 1.2.5 patches GHSA-897w-fcg9-f6xj (arbitrary file write) and pyjwt 2.13.0
+# patches PYSEC-2026-179 (HMAC/JWK key-confusion); a constraint cannot satisfy these
+# against the SDK's hard pins, so override them to the patched versions until the SDK
+# bump propagates to the pinned master rev.
 override-dependencies = [
   "okta==3.4.2",
-  "microsoft-kiota-abstractions==1.9.9"
+  "microsoft-kiota-abstractions==1.9.9",
+  "dulwich==1.2.5",
+  "pyjwt==2.13.0"
```
HugoPBrito: Doesn't the new `pyjwt==2.13.0` override replace the full requirement, extras included, so every `pyjwt[crypto]` request (`pygithub`, `msal`, `allauth`) resolves to bare `pyjwt` and the `crypto` extra stops pulling `cryptography` in `api/uv.lock`? The old constraint preserved the extra, but an override drops it, so wouldn't `cryptography` now arrive only transitively through azure/msal/pyopenssl, leaving RS256 working by accident rather than by contract?

Wouldn't this be avoided by pinning the override as `pyjwt[crypto]==2.13.0`, keeping the extra tied to pyjwt explicitly?

cesararroba: Nice catch, you're right. The uv override replaces the whole requirement, so the bare `pyjwt==2.13.0` was quietly dropping the `crypto` extra. On master those `pyjwt[crypto]` edges stay because the constraint keeps them, but here they collapsed to plain `pyjwt`, and `cryptography` was only coming in transitively via azure/msal/pyopenssl. So RS256 worked, but more by luck than on purpose.

Fixed in 4ad5119 by pinning the override as `pyjwt[crypto]==2.13.0`. Now all 9 pyjwt edges in `api/uv.lock` carry `extra = ["crypto"]` and pyjwt declares `crypto = [{ name = "cryptography" }]`, so cryptography is tied to pyjwt directly. `uv lock --check` passes and an RS256 sign/verify roundtrip works. Thanks for spotting it!

HugoPBrito: Thanks!

HugoPBrito (approved):

josema-xyz (approved):

PR #11500, opened 2026-06-08
feat(ui): resolve public SaaS config at container runtime

pfe-nazaries:
### Context

The Prowler UI bakes its public SaaS configuration (Sentry, Google Tag Manager, API base/docs URL) into the JavaScript bundle **at build time** — Next.js inlines every `process.env.NEXT_PUBLIC_*` reference reachable from client code. As a result, one prebuilt image cannot enable or disable integrations per deployment without a rebuild, which blocks the "build once, deploy many" model needed for Enterprise/self-hosted.

This migrates the UI's public client config from build-time inlining to **runtime resolution**: a single image serves every environment, and an unset value disables its integration with zero third-party egress.

### Description

- **Runtime delivery (no new dependency):** a server component injects an allowlisted, escaped JSON **data island** (`<script type="application/json">`) into `<head>` of both root layouts, read per-request via `connection()`; a client reader parses it once with an all-null fallback.
- **`UI_` prefix + legacy fallback:** migrated keys move from `NEXT_PUBLIC_*` to `UI_*` (`UI_API_BASE_URL`, `UI_API_DOCS_URL`, `UI_GOOGLE_TAG_MANAGER_ID`, `UI_SENTRY_DSN`, `UI_SENTRY_ENVIRONMENT`); the server + browser Sentry vars are unified. A shared `readEnv(primary, legacy)` helper resolves each var with its former `NEXT_PUBLIC_*` name as a **deprecated runtime fallback**, so existing deployments keep working. Both names are read via a computed key so Turbopack does not inline them (a literal `NEXT_PUBLIC_*` read would be build-snapshotted and break the fallback).
- **Sentry:** browser init relocated to `instrumentation-client.ts` (runs before hydration, covers `global-error.tsx`); `withSentryConfig` build gate decoupled from the now-runtime DSN (gated on `SENTRY_AUTH_TOKEN`/`ORG`/`PROJECT`).
- **GTM:** native server-side read + conditional mount (`{gtmId && <GoogleTagManager/>}`) so an unset id loads nothing.
- **Boot contract:** `lib/env.ts` asserts required env at server startup (fails fast); CSP stays static (documented `connect-src` constraint for self-hosted Sentry).
- **Provisioning & docs:** `.env`, `Dockerfile`, Helm, and the e2e workflow updated to the `UI_` names (legacy kept as deprecated); new `docs/developer-guide/environment-variables.mdx` documenting the per-component prefix convention.

### Steps to review

1. **Build once, configure at runtime:** start the image in two containers with different `UI_*` values → each resolves its own config with no rebuild.
2. **Legacy fallback:** set only the old `NEXT_PUBLIC_API_BASE_URL` (leave `UI_API_BASE_URL` unset) → the app still boots and reaches the API. (Verified: production build keeps the read dynamic, and the standalone server boots with only the legacy name set.)
3. **Zero egress when unset:** leave `UI_SENTRY_DSN` and `UI_GOOGLE_TAG_MANAGER_ID` unset → no `Sentry.init` and no request to `googletagmanager.com`.
4. `cd ui && pnpm vitest run --project unit` and `pnpm run build`.

### Checklist

<details>

</details>

- [ ] Review if code is being documented following the Python style guide (N/A — TypeScript/UI).
- [ ] Review if is needed to change the Readme.md (N/A).

#### UI (if applicable)
- Screenshots/Video — **N/A**: no visible UI change (runtime config plumbing; same rendered output).

### License

## Summary by CodeRabbit

* **New Features**
  * UI public configuration is now resolved at container runtime and provided to the browser (API base/docs, Sentry, Google Tag Manager, and other reserved values).
* **Refactor**
  * Standardized environment variable names to `UI_*`, with `NEXT_PUBLIC_*` kept as deprecated fallback.
* **Bug Fixes**
  * Updated container, Helm, and testing environments to use the runtime API base/docs URLs consistently.
* **Documentation**
  * Updated developer, environment-variable, troubleshooting, and end-to-end testing docs to reflect the new `UI_*` runtime approach.
* **Tests**
  * Added/updated unit and E2E coverage for runtime configuration and telemetry initialization.


ui/Dockerfile
```diff
@@ -35,12 +35,17 @@ RUN corepack install
...
+# GTM / API base+docs URLs are runtime container env (prod stage), not build ARGs.
+
+# Build-time only: Sentry source-map upload creds (withSentryConfig runs only
+# when all three are set). The runtime UI_SENTRY_DSN is separate.
+ARG SENTRY_AUTH_TOKEN
+ENV SENTRY_AUTH_TOKEN=${SENTRY_AUTH_TOKEN}
+ARG SENTRY_ORG
+ENV SENTRY_ORG=${SENTRY_ORG}
+ARG SENTRY_PROJECT
+ENV SENTRY_PROJECT=${SENTRY_PROJECT}
```
cesararroba: Why are these needed now as arguments?

.env
```diff
@@ -139,10 +145,20 @@ DJANGO_BROKER_VISIBILITY_TIMEOUT=86400
 DJANGO_SENTRY_DSN=
 DJANGO_THROTTLE_TOKEN_OBTAIN=50/minute
 
-# Sentry settings
-SENTRY_ENVIRONMENT=local
+# Sentry for the web app (server + browser). Empty/unset UI_SENTRY_DSN ⇒
+# Sentry disabled, zero egress. SENTRY_RELEASE (unprefixed) feeds the web app's
+# server/edge SDKs.
+# deprecated, use UI_SENTRY_DSN
+NEXT_PUBLIC_SENTRY_DSN=
```
cesararroba: Is it needed on the `.env` file or can we remove it?

.env
```diff
@@ -139,10 +145,20 @@ DJANGO_BROKER_VISIBILITY_TIMEOUT=86400
...
+UI_SENTRY_DSN=
+# deprecated, use UI_SENTRY_ENVIRONMENT
+NEXT_PUBLIC_SENTRY_ENVIRONMENT=local
```
cesararroba: Is it needed on the `.env` file or can we remove it?

.env
```diff
@@ -139,10 +145,20 @@ DJANGO_BROKER_VISIBILITY_TIMEOUT=86400
...
+UI_SENTRY_ENVIRONMENT=local
 SENTRY_RELEASE=local
-NEXT_PUBLIC_SENTRY_ENVIRONMENT=${SENTRY_ENVIRONMENT}
+# Reserved runtime public config (registered now; no UI consumer yet)
+# POSTHOG_KEY=
+# POSTHOG_HOST=
+# REO_DEV_CLIENT_ID=
 
 #### Prowler release version ####
 NEXT_PUBLIC_PROWLER_RELEASE_VERSION=v5.30.0
```
cesararroba: Check this one please!

cesararroba (changes requested):

ui/next.config.js
```diff
@@ -6,7 +6,11 @@ const { withSentryConfig } = require("@sentry/nextjs");
 /** @type {import('next').NextConfig} */
 
 // HTTP Security Headers
-// 'unsafe-eval' is configured under `script-src` because it is required by NextJS for development mode
+// 'unsafe-eval' is configured under `script-src` because it is required by NextJS for development mode.
+//
+// CSP is static; the JSON config island is inert (no nonce needed). A runtime
+// Sentry DSN must be in `connect-src` below — `*.sentry.io` covers Sentry Cloud,
+// but a self-hosted/region host is blocked until per-request CSP (middleware) lands.
```
Alan-TheGentleman, line 13: The static `connect-src` blocks a self-hosted or region-specific `UI_SENTRY_DSN` until per-request CSP lands, so the Enterprise/self-hosted deployments this PR targets cannot actually send Sentry events yet. Not blocking, but worth a tracked follow-up issue for the per-request CSP so the limitation does not get lost in an inline note.

ui/instrumentation-client.ts
```diff
@@ -20,17 +30,134 @@ export const NAVIGATION_TYPE = {
...
 
+const { sentryDsn, sentryEnvironment } = getRuntimeConfigClient();
+
+// Only initialize Sentry in the browser when a runtime DSN is configured.
+if (typeof window !== "undefined" && sentryDsn) {
+  // R6: default to a non-dev environment so a deployed container with
+  // UI_SENTRY_ENVIRONMENT unset does NOT run the browser SDK in dev mode
+  // (verbose debug logging, 100% sampling, synthetic dev user). Only an
+  // explicit "local" enables development behavior.
+  const environment = sentryEnvironment ?? "production";
```
Alan-TheGentleman, line 40: The expected value of `UI_SENTRY_ENVIRONMENT` is not consistent across runtimes: the browser treats anything other than `local` as production, while `sentry.server.config.ts` and the edge config gate on `sentryEnvironment === "pro"`, so `UI_SENTRY_ENVIRONMENT=production` reads as non-prod on the server. The variable got unified but the accepted value did not. Not blocking, but worth documenting the canonical value or aligning the comparisons so one string behaves the same on both sides.

cesararroba (dismissed):

alejandrobailo (changes requested):
Really nice work on this one, I do have a few things:

1. Standalone Helm chart crash-loops on the new image (I'd call this a blocker).
ui/lib/env.ts now throws at boot if AUTH_URL or AUTH_SECRET are missing, but contrib/k8s/helm/prowler-ui/values.yaml doesn't define AUTH_URL at all and ships AUTH_SECRET empty. Since helper.ts also dropped the old AUTH_URL || "http://localhost:3000" fallback, a default helm install of that chart will crash with Missing required env: AUTH_URL. The combined prowler-app chart is fine because its configmap sets AUTH_URL, so this one is easy to miss while testing. Could we add AUTH_URL (and a real AUTH_SECRET) to the standalone values.yaml?

2. A couple of test gaps worth a quick look.
The isDevelopment ("local") branch isn't exercised in any of the three Sentry config tests even though this PR flips that condition. And the Sentry e2e check reads window.__SENTRY__ internals, which with no DSN configured in CI collapses to expect(null).toBe(null) and passes without actually verifying anything. env.test.ts also only covers UI_API_BASE_URL, not the AUTH_URL / AUTH_SECRET assertions that are now the only guardrail after the fallback removal in point 1.

None of this changes the direction, which I think is spot on. I mostly want to make sure we don't surprise existing self-hosted deployments on upgrade. Happy to pair on any of these if it helps.

cesararroba (approved):
Awesome work! 🚀

alejandrobailo (approved):

HugoPBrito (approved):

PR #11521, opened 2026-06-10
feat(ui): per-provider scan schedule management gated by capability

alejandrobailo:
### Context

Providers could only be put on a fixed daily scheduled scan from the add-provider wizard. This PR adds full scan-schedule management to the UI (daily, weekly, monthly, or fixed interval), backed by the per-provider `/schedules` API and gated by a capability seam so Prowler OSS keeps its legacy daily behavior.

### Description

- New `schedules` server actions (`GET`/`PATCH`/`DELETE /schedules/{providerId}`) with shared types (`ui/types/schedules.ts`) and pure helpers (`ui/lib/schedules.ts`) for form mapping, payload building, and next-run estimation.
- New `ScanScheduleFields` and `EditScanScheduleModal` components, reused by the provider wizard `LaunchStep` and by new "Edit Scan Schedule" row actions in the Providers and Scan Jobs tables.
- Capability gating via `SCAN_SCHEDULE_CAPABILITY`: `DAILY_LEGACY` (OSS) keeps the legacy `POST /schedules/daily` flow with advanced cadences locked behind a Cloud upsell badge; `ADVANCED` enables full scheduling; `MANUAL_ONLY` (prop-driven only) allows just an on-demand scan, honoring `isScanLimitReached`.
- Next-run estimate matches the backend semantics: cron weekday convention (0=Sunday), day-of-month capped at 28, and INTERVAL anchored at the next occurrence of the selected hour. Editing preserves a custom `scan_interval_hours` instead of rewriting it to 48.
- Shared `isCloud()` env helper, and scan-jobs navigation now pre-selects the provider through the `provider_uid__in` filter key with URL-safe encoding.

### Steps to review

1. Run the unit tests: `pnpm vitest run lib/schedules.test.ts components/providers/wizard/steps/launch-step.test.tsx components/providers/table/data-table-row-actions.test.tsx components/scans/table/scan-jobs-row-actions.test.tsx` (59 tests) plus `pnpm run typecheck` and `pnpm run lint`.
2. OSS mode (`NEXT_PUBLIC_IS_CLOUD_ENV` unset): in the add-provider wizard launch step, the cadence/time selects are locked to Daily with the Cloud badge, saving calls the legacy `POST /schedules/daily`, and the "Edit Scan Schedule" row action is hidden in both the Providers and Scan Jobs tables.
3. Cloud mode (`NEXT_PUBLIC_IS_CLOUD_ENV=true`): open "Edit Scan Schedule" from either table and verify the `PATCH /schedules/{providerId}` payload per cadence — unused shape fields sent as `null`, weekday `0=Sunday`, day-of-month `1-28`, browser IANA timezone — and that "Remove Scan Schedule" issues the `DELETE` and the modal then opens with defaults.
4. Check the "next scheduled scan" copy against each cadence, including INTERVAL (first run anchors at the next occurrence of the selected hour, not now + interval).
5. Review `ui/lib/schedules.ts` and `ui/types/schedules.ts` for the capability seam (`capability` prop + `getScanScheduleCapability`) that the cloud overlay will inject into.

### Checklist

<details>

</details>

#### SDK/CLI

#### UI

#### API

### License

## Summary by CodeRabbit

* **New Features**
  * Full scan schedule management: create, edit, remove schedules via modal and inline editors; option to launch an initial scan when saving.
  * Pending scheduled scans appear in the Scans table with schedule summaries, plus new Next/Last scan columns.

* **Improvements**
  * Advanced Cloud scheduling: custom intervals, weekly/monthly, timezone-aware times, and capability-driven UI gating.
  * Links and schedule-related UI now use safe URL encoding and improved local time display.


ui/app/(prowler)/scans/page.tsx
```diff
@@ -33,6 +44,37 @@ const getFilterSearchQuery = (
...
+    .filter(Boolean);
+};
+
+/** Applies the table's provider filters to synthetic pending-schedule rows. */
+const filterProvidersForPendingRows = (
+  providers: ProviderProps[],
+  searchParams: SearchParamsProps,
+): ProviderProps[] => {
+  const uids = parseCsvParam(
+    searchParams["filter[provider_uid__in]"] ??
```
pfe-nazaries: If reasonable, searchParams could accept instead just string type (array or not) the template type like:

```typescript
export type FiltersParams = "provider_uid__in" | ....

export type SearchParams = `filter[${FiltersParams}]`
```

ui/components/scans/schedule/scan-schedule-fields.tsx
```diff
@@ -0,0 +1,302 @@
...
+  canUseAdvancedSchedule = true,
+  showCloudUpgradeBadge = false,
+}: ScanScheduleFieldsProps) {
+  // useWatch, not form.watch: form.watch re-renders are dropped by React Compiler memoization.
+  const control = form.control;
+  const frequency = useWatch({ control, name: "frequency" });
+  const hour = useWatch({ control, name: "hour" });
+  const dayOfWeek = useWatch({ control, name: "dayOfWeek" });
+  const dayOfMonth = useWatch({ control, name: "dayOfMonth" });
+  const intervalHours = useWatch({ control, name: "intervalHours" });
```
pfe-nazaries: **nitpick:** rhf has a new Watch component to put in the render to scope what part of the ui re-render when this values changes instead of the whole component having the used here as hook. I think you only need these vars grouped in one section of the render

ui/components/scans/table/scan-jobs-columns.test.tsx
```diff
@@ -9,6 +9,18 @@ vi.mock("@/components/shadcn", () => ({
     <span>{children}</span>
   ),
   Progress: () => <div />,
+  StackedCell: ({
+    primary,
+    secondary,
+  }: {
+    primary: React.ReactNode;
```
pfe-nazaries: Import just ReactNode from react 🙏

ui/app/(prowler)/scans/page.tsx
```diff
@@ -191,9 +238,71 @@ const SSRDataTableScans = async ({
...
+  let tableData = scansWithSchedule;
+  if (tab === SCAN_JOBS_TAB.SCHEDULED) {
+    // Append pending rows only after the last page of real rows.
+    const totalPages = meta?.pagination?.pages ?? 0;
+    if (page >= totalPages) {
+      const coveredProviderIds = new Set(
+        scansWithSchedule
+          .map((scan) => scan.relationships?.provider?.data?.id)
+          .filter((id): id is string => Boolean(id)),
+      );
```
pfe-nazaries: **Pending-schedule rows can duplicate and break the footer count across pages.**

`coveredProviderIds` is built only from the current (last) page's scans, but pending rows are synthesized for every configured provider not in that set. So if a provider's real `state=scheduled` row lives on an earlier page, it's not "covered" → a duplicate "Pending" row is emitted on the last page. This triggers whenever scheduled scans span more than one page (`totalPages > 1`).

Separately, the synthesized rows are appended to `data` while `meta.pagination` still reflects real scans only: the last page can exceed `pageSize`, the footer count excludes pending rows, and on an empty Scheduled tab (`totalPages = 0`) rows render while the footer shows `0`.

Suggest deriving covered providers from a source not limited to the current page (e.g. a lightweight query for providers that already have a `state=scheduled` scan) and reconciling the footer count with the appended rows.

ui/components/providers/wizard/steps/launch-step.tsx
```diff
@@ -88,46 +111,122 @@ export function LaunchStep({
...
+      return;
+    }
+
+    onClose();
+
+    const goToScans = (
+      <ToastAction altText="Go to scans" asChild>
+        <Link href={`/scans?tab=${SCAN_JOBS_TAB.ACTIVE}`}>Go to scans</Link>
+      </ToastAction>
+    );
```
pfe-nazaries, line 184: **`goToScans` links to the Active tab for outcomes that never launch a scan.**

This single action is reused for every save outcome, but only `SAVED_AND_LAUNCHED` actually has a scan in the Active tab. For `SAVED` (schedule saved, no initial scan) and `SAVED_SCAN_FAILED` (initial scan couldn't launch), there is no active scan, so "Go to scans" lands the user on an empty/irrelevant tab. The relevant artifact in those cases is the schedule, which lives in the Scheduled tab.

`launch-scan-modal.tsx` already routes the equivalent toast to `tab=SCHEDULED` ("View schedule") — this wizard step is inconsistent with it.

Suggest routing the non-launched outcomes (`SAVED`, `SAVED_SCAN_FAILED`) to `SCAN_JOBS_TAB.SCHEDULED` and keeping `ACTIVE` only for `SAVED_AND_LAUNCHED` (or simply pointing to `SCHEDULED` in all cases, like the modal).

pfe-nazaries (changes requested):

ui/app/(prowler)/scans/page.tsx
alejandrobailo: Good point. I did not narrow the global SearchParamsProps because it also carries non-filter params like tab/page/sort, but I added a local template-literal type for the provider filter keys used by the pending-row filtering so those keys are no longer raw strings.

ui/components/scans/schedule/scan-schedule-fields.tsx
alejandrobailo: I checked the installed RHF version (7.62.0) and it does not export a Watch component, only useWatch. I still reduced the subscriptions here by grouping these fields into a single useWatch call; a deeper render-scope split would need a local subcomponent rather than the non-existent Watch API in this version.

ui/components/scans/table/scan-jobs-columns.test.tsx
alejandrobailo: Done, switched the mocked component props to import ReactNode explicitly instead of using React.ReactNode.

ui/app/(prowler)/scans/page.tsx
alejandrobailo: Agreed. I changed this to derive covered providers from the full scheduled-scan result set when the current page is only a slice, then paginate the synthetic pending rows as part of the same combined list. The combined metadata now reflects real scheduled scans plus pending schedule rows, so the footer/page count stays aligned and previously paged real rows are not duplicated as Pending.

ui/components/providers/wizard/steps/launch-step.tsx
alejandrobailo: Agreed. I changed the wizard toast target to depend on the result: SAVED_AND_LAUNCHED goes to the Active tab, while SAVED and SAVED_SCAN_FAILED go to Scheduled. Added a test for the launched case so it does not regress back to always-Scheduled or always-Active.

pfe-nazaries (dismissed):

pfe-nazaries (approved):

PR #11575, opened 2026-06-13
fix(gcp): credit audit-filtered aggregated sinks in metric-filter checks

### LINKED ISSUE #11574: GCP: logging_log_metric_filter_and_alert_* checks miss audit-filtered org aggregated sinks (follow-up to #11488)
### Steps to Reproduce

1. Centralized logging topology (the one #11487 / #11488 describe): an organization-level aggregated sink (`includeChildren=true`) **filtered to the cloudaudit streams** (`logName: .../cloudaudit.googleapis.com%2Factivity OR ...`), whose destination log bucket holds bucket-scoped CIS metrics with alert policies.
2. Run any `logging_log_metric_filter_and_alert_for_*` check (Prowler ≥ 5.30.0) against the organization.
3. Every child project is reported `FAIL`, despite being fully metric-monitored and alerted through the central bucket.

### Expected behavior

Child projects are credited. The CIS logging-metric filters all match **Admin Activity audit entries** (`protoPayload.methodName=...` on audited resource types), so an aggregated sink carrying the Admin Activity stream delivers every log entry those metrics can match.

### Actual behavior

`get_projects_covered_by_aggregated_metric` skips every sink whose `filter != "all"` (the guard added in the #11488 review revision `0bd290b`, "tighten aggregated metric sink coverage"). The crediting introduced by #11488 therefore never applies to audit-filtered aggregated sinks — which are the cost-sane default for this topology, since an unfiltered org-wide sink would duplicate every project's operational log entries into the central bucket.

### Proposed fix

Credit a sink when it exports everything (`filter == "all"`) **or** when its filter carries the Admin Activity stream (`cloudaudit.googleapis.com/activity`, in either the URL-encoded `%2F` or plain `logName` form). A sink filtered to unrelated streams — or to other cloudaudit streams that are not Admin Activity (e.g. `data_access` only) — is still not credited, so the change widens sink eligibility without inventing coverage. Verified against a live organization with an audit-filtered aggregated sink: the eight CIS checks move from every child project failing to every child project passing, while a check with no matching central metric correctly stays failing.

PR with the fix + tests to follow.

alinealfa:
### Context

Follow-up to #11488 (itself a follow-up to #11343 / #11355). The review revision `0bd290b` ("tighten aggregated metric sink coverage") restricted `get_projects_covered_by_aggregated_metric` to sinks with `filter == "all"`. The filter == "all" guard keeps that crediting conservative. This PR keeps it conservative while also reaching the common centralized case: an aggregated sink filtered to the cloudaudit streams — the cost-sane default, since exporting everything org-wide would duplicate each project's operational logs centrally — which still carries every Admin Activity entry the CIS metrics match.

Fix #11574

### Description

The nine `logging_log_metric_filter_and_alert_for_*` metric filters all match **Admin Activity audit entries** (`protoPayload.methodName=...` on audited resource types). A sink whose filter carries the Admin Activity stream therefore delivers every log entry those metrics can match — sufficient for crediting, while staying far stricter than no filter condition at all:

- New `_sink_delivers_activity_logs(sink_filter)`: true for `"all"`/empty, or when the filter references `cloudaudit.googleapis.com/activity` (the URL-encoded `%2F` form the Logging API returns is normalized first).
- The aggregated-sink loop uses the predicate instead of `filter != "all"`.
- CHANGELOG entry.

### Steps to review

1. `pytest tests/providers/gcp/services/logging/` passes locally. The helper's unit tests exercise both directions, so the change widens eligibility without inventing coverage:
   - **Credited:** Admin Activity stream present, in both the URL-encoded (`...%2Factivity`) and plain `logName` forms.
   - **Not credited:** a cloudaudit sink that carries another stream but not Admin Activity (`data_access` only); an unrelated filter (`resource.type="gce_instance"`); and the pre-existing guards (no alert policy, metric not bucket-scoped, sink not `includeChildren`, wrong destination bucket, different org). The restrictive-filter guard test from `0bd290b` is unchanged.
2. Live verification against an organization running an audit-filtered aggregated sink: the eight CIS checks move from every child project failing to every child project passing, the PASS reason naming the org-level aggregated sink. `logging_log_metric_filter_and_alert_for_compute_configuration_changes_enabled` (no matching central metric in that org) correctly stays failing — the predicate only widens sink eligibility, it never invents metric coverage.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated GCP logging metric filter and alert checks to credit organization-level aggregated sinks when they include the Admin Activity audit stream (including filter expressions that previously weren’t recognized as “credited”).

* **Tests**
  * Added unit coverage for sink filter recognition, including URL-encoded and `log_id()` selector formats, negated/restrictive expressions, and a hyphenated log path regression.

* **Documentation**
  * Refreshed the unreleased changelog entry for this fix.

### DIFF
--- a/prowler/providers/gcp/services/logging/logging_service.py
+++ b/prowler/providers/gcp/services/logging/logging_service.py
@@ -121,6 +121,27 @@ class Metric(BaseModel):
     bucket_name: str = ""
 
 
+def _sink_delivers_activity_logs(sink_filter: str) -> bool:
+    """True when a sink's filter exports the Admin Activity audit stream (or
+    everything).
+
+    The CIS logging-metric filters all match Admin Activity audit entries
+    (``protoPayload.methodName=...`` on audited resource types), so an
+    aggregated sink restricted to the cloudaudit streams still delivers every
+    log entry those metrics can match. Centralized topologies deliberately
+    filter their aggregated sink to the audit streams (an unfiltered org-wide
+    sink would duplicate every operational log entry into the central bucket),
+    so requiring ``filter == "all"`` rejects exactly the deployments this
+    crediting exists for. Sink filters reference the stream either URL-encoded
+    (``cloudaudit.googleapis.com%2Factivity``) or as a plain path — normalize
+    before matching.
+    """
+    if not sink_filter or sink_filter == "all":
+        return True
+    normalized = sink_filter.replace("%2F", "/").replace("%2f", "/")
+    return "cloudaudit.googleapis.com/activity" in normalized
+
+
 def get_projects_covered_by_aggregated_metric(
     logging_client, monitoring_client, metric_filter
 ):
@@ -133,6 +154,10 @@ def get_projects_covered_by_aggregated_metric(
     every child project's logs into one bucket, where a single bucket-scoped metric
     + alert covers them all. Without crediting that, those child projects are falsely
     failed. Mirrors the org-sink handling already in ``logging_sink_created`` (#11355).
+
+    A sink is credited when it exports everything (``filter == "all"``) or when its
+    filter carries the Admin Activity audit stream — the only stream the CIS metric
+    filters can match (see ``_sink_delivers_activity_logs``).
     """
     # Buckets that hold a matching, alerted, bucket-scoped metric -> metric name.
     bucket_to_metric = {}
@@ -155,7 +180,7 @@ def get_projects_covered_by_aggregated_metric(
     for sink in logging_client.sinks:
         if not getattr(sink, "include_children", False):
             continue
-        if getattr(sink, "filter", "all") != "all":
+        if not _sink_delivers_activity_logs(getattr(sink, "filter", "all")):
             continue
         for bucket, metric_name in bucket_to_metric.items():
             # sink.destination e.g. "logging.googleapis.com/projects/.../buckets/X";
--- a/prowler/providers/gcp/services/logging/logging_service.py
+++ b/prowler/providers/gcp/services/logging/logging_service.py
@@ -1,9 +1,12 @@
+import re
+
 from pydantic.v1 import BaseModel
 
 from prowler.lib.logger import logger
 from prowler.providers.gcp.config import DEFAULT_RETRY_ATTEMPTS
 from prowler.providers.gcp.gcp_provider import GcpProvider
 from prowler.providers.gcp.lib.service.service import GCPService
+from prowler.providers.gcp.services.monitoring.monitoring_service import Monitoring
 
 
 class Logging(GCPService):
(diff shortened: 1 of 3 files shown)


prowler/providers/gcp/services/logging/logging_service.py
```diff
@@ -121,6 +121,27 @@ class Metric(BaseModel):
...
+    sink would duplicate every operational log entry into the central bucket),
+    so requiring ``filter == "all"`` rejects exactly the deployments this
+    crediting exists for. Sink filters reference the stream either URL-encoded
+    (``cloudaudit.googleapis.com%2Factivity``) or as a plain path — normalize
+    before matching.
+    """
+    if not sink_filter or sink_filter == "all":
+        return True
+    normalized = sink_filter.replace("%2F", "/").replace("%2f", "/")
+    return "cloudaudit.googleapis.com/activity" in normalized
```
HugoPBrito: I believe this predicate is too broad as-is.

A filter can contain `cloudaudit.googleapis.com/activity` while still excluding or restricting the logs we need, for example:

```text
log_id("cloudaudit.googleapis.com/activity") AND resource.type="gce_instance"
NOT log_id("cloudaudit.googleapis.com/activity")
logName!="projects/p/logs/cloudaudit.googleapis.com%2Factivity"
```

Since this helper can make child projects PASS the CIS metric checks, we should avoid crediting filters we cannot safely prove are full-coverage.

Can we make this conservative and only accept clearly positive full-coverage cases, then reject negated or restrictive expressions?

tests/providers/gcp/services/logging/logging_service_test.py
```diff
@@ -291,6 +291,37 @@ def test_not_covered_when_sink_filter_is_restrictive(self):
...
+    def test_not_covered_when_sink_filter_omits_activity_stream(self):
+        """A sink that routes cloudaudit streams but NOT Admin Activity (here,
+        data_access only) does not deliver the entries the CIS metric filters
+        match, so it must not be credited — right service, wrong stream."""
+        logging_client, monitoring_client = self._clients(
+            sink_filter="logName: /logs/cloudaudit.googleapis.com%2Fdata_access"
+        )
+        assert self._run(logging_client, monitoring_client) == {}
+
+    def test_covered_when_sink_filter_carries_activity_stream_encoded(self):
```
HugoPBrito, line 305: Can we add negative coverage for restrictive and negated filters too?

The positive cases cover the intended audit-filtered sink path, but the safety boundary here is making sure we do not credit filters like `activity AND resource.type=...`, `NOT activity`, or `logName != ...activity`.

HugoPBrito (changes requested):
Thanks a lot for following up on the aggregated sink support here. The use case makes sense, especially for centralized logging setups where exporting every org log would be too noisy.

I left a couple of inline comments because I believe we need to keep the sink filter matching conservative before this is merge-ready, mainly to avoid turning restricted or negated filters into false PASS results.

alinealfa:
Thanks @HugoPBrito. Please review if the latest commit addresses all issues.

The original substring check was indeed too permissive: it credited a filter that merely *names* the Admin Activity stream, so both negated and restricted forms slipped through.

`_sink_delivers_activity_logs` now is credited only when filter is empty / `all`, or when all three hold:

1. it **positively selects** the Admin Activity stream — `logName:`, `logName=`, or `log_id("…/activity")`;
2. it carries **no negation** (`NOT` / `-` / `!=`) and **no restriction** (`AND`); and
3. once the audit-stream selectors and the `OR` glue are stripped, **nothing remains** — so a leftover predicate like `severity>=ERROR` or `resource.type=…` (whether `AND`-ed or `OR`-ed in) fails closed.

Rejecting every `AND` is intentional: there is no `AND` that intends to preserves full Activity coverage — each one narrows the delivered set — so failing closed is the safe direction for a helper that can flip a project to PASS.

It matches the audit-stream *family* (`cloudaudit.googleapis.com/<stream>`) rather than a hand-listed set, so a sink that legitimately `OR`s Activity with `data_access` / `system_event` / `access_transparency` is still credited and no stream can be dropped from the allowlist by accident.

Added negative tests for all three of your examples plus other combinations,  an `OR`-with-non-audit case, and a positive test for the bare `log_id()` selector.

HugoPBrito (approved):
Thank you very much for reporting and fixing this so quickly! 🚀

PR #11577, opened 2026-06-13
feat(m365): add entra_conditional_access_policy_no_exclusion_gaps check

### LINKED ISSUE #11062: [New Check]: Conditional Access excluded objects must be covered by another policy (no exclusion gaps)

UTKARSH698:
### Description

Adds a new M365 Entra Conditional Access check, `entra_conditional_access_policy_no_exclusion_gaps`, as requested in #11062.

The check verifies that every object excluded from an **enabled** Conditional Access policy (users, groups, roles, applications, or platforms) is still **included** by at least one other enabled policy, so the exclusion keeps a compensating control. An object excluded everywhere and included nowhere sits completely outside CA enforcement — a silent MFA-bypass / lateral-movement path.

To avoid false positives it intentionally ignores two by-design exclusions, mirroring the existing sibling checks:

- the **Directory Synchronization Accounts** role (`d29b2b05-8046-44ba-8758-1e26182fcf32`), per `entra_conditional_access_policy_directory_sync_account_excluded`
- confirmed **break-glass / emergency access** accounts (objects excluded from *every* enabled Block policy), per `entra_emergency_access_exclusion`

Report-only and disabled policies are out of scope. The check emits a single aggregate finding.

### Scope

This first version covers user, group, role, application, and platform exclusions — the fields currently modelled on `ConditionalAccessPolicy`. Service-principal and location exclusions require additional service-model fields and are noted as a follow-up in the metadata `Notes`.

### Checklist

- [x] New check + metadata.json + `__init__.py`
- [x] Unit tests covering PASS / FAIL / skip cases (13 tests, all passing locally)
- [x] Metadata passes `CheckMetadata` validation

Closes #11062

## Summary by CodeRabbit

* **New Features**
  * Added a Microsoft Entra Conditional Access security check (`entra_conditional_access_policy_no_exclusion_gaps`, severity: medium) to identify “exclusion coverage gaps” where excluded principals are not covered by other enabled Conditional Access policies.
  * Reports uncovered exclusions grouped by type (users, groups, roles, applications, platforms) with remediation guidance.
* **Tests**
  * Added automated test coverage for enabled/disabled behavior, no-exclusion and covered-exclusion scenarios, uncovered exclusions by type, and directory sync/break-glass edge cases.


UTKARSH698:
Thanks for the thorough review! Addressed all three points in `3e14a6a`:

1. **State-filter consistency (Major).** `_emergency_access_objects()` now only counts `ENABLED` policies, matching `execute()`. Report-only (non-enforced) Block policies no longer dilute the "excluded from every blocking policy" determination, so a genuine break-glass account excluded from all *enforced* blocking policies is correctly recognized rather than flagged as a coverage gap. Added a regression test (`test_emergency_access_ignores_report_only_blocking_policy`) that fails against the previous `!= DISABLED` logic and passes with the fix.
2. **Mock instance.** `_run` now uses `mock.MagicMock()` instead of assigning the class, so each test gets a fresh instance.
3. **Brittle assertion.** Replaced the trailing-space substring check with a whitespace-robust word-boundary regex (`r"users:\s*user-1"`).

All 14 tests pass locally.

UTKARSH698:
Thanks! Applied the remaining nitpicks in `a43b9d1`:

- Added Google-style docstrings and type hints to the `_policy` / `_run` test helpers and a docstring to the test class.
- Fixed the `_USER_COLLECTIONS` comment to describe the actual 3-element tuples `(label, included attr, excluded attr)`.

All 14 tests still pass locally.

danibarranqueroo (approved):
Thanks for this! Really good job 🚀

PR #11578, opened 2026-06-14
feat(sdk): add sdk_only provider property to hide providers from the app

StylusFrost:
### Context

With dynamic provider discovery, the API needs the SDK to be the single source of truth for which providers exist, how they are constructed from stored credentials, and which ones belong in the app. This PR collects the SDK contract changes the API layer consumes; the API wiring itself ships separately.

### Scope
- Includes:
  - `Provider.sdk_only` contract property (default `True`), `Provider.get_app_providers()`, `sdk_only = False` on app-facing built-in providers
  - Compliance discovery merges every entry-point directory a provider contributes (`_get_ep_compliance_dirs` returns `{provider: [paths]}`) instead of overwriting
  - `Provider.get_scan_arguments()`, `Provider.get_connection_arguments()` and `Provider.get_credentials_schema()` contract methods
- Excludes: API consumption of these contracts (separate PRs)

### Autonomy
- [ ] Tests cover this unit

### Description

**`sdk_only` provider property**

- `Provider.sdk_only: bool = True` on the base contract: a provider (built-in or external) is CLI/SDK-only and hidden from the app until it opts in with `sdk_only = False`.
- `Provider.get_app_providers()`: returns the providers with `sdk_only = False` (counterpart of `get_available_providers()`, which keeps listing everything for the CLI). A provider whose class fails to import is treated as `sdk_only`.
- The app-exposed built-in providers declare `sdk_only = False`, so the released behavior is preserved; `llm`, `nhn`, `scaleway`, `stackit` keep the default and stay hidden.

**Multi-package compliance discovery**

- `_get_ep_compliance_dirs` now accumulates every directory per provider (`{provider: [paths]}`) so a provider contributed by several external packages has all its compliance frameworks discovered instead of the last one overwriting the rest.

**Programmatic construction contract**

- `Provider.get_scan_arguments(provider_uid, secret, mutelist_content=None)` builds the constructor kwargs from a stored uid plus a secret dict (the shape the API persists), as opposed to the CLI's explicit flags.
- `Provider.get_connection_arguments(provider_uid, secret)` builds the `test_connection` kwargs, which often differ in shape from the constructor.
- `Provider.get_credentials_schema()` maps each accepted secret type to the pydantic model that validates it; the default is empty (secret accepted as-is and validated by `test_connection`).

These base implementations are the defaults; providers override them as needed. The CLI is unaffected. No `api/` changes are included in this PR by design.

### Steps to review

```
uv run pytest tests/providers/external/test_dynamic_provider_loading.py -k "SdkOnly or Compliance or BaseContractDefaults"
uv run pytest tests/config/config_test.py
```

### Checklist

#### SDK/CLI

### License

## Release Notes

* **New Features**
  * Providers can now be flagged for app-visibility control: by default they’re hidden from the app experience (still usable via CLI/SDK).
  * Added provider contract helpers to programmatically generate scan/connection arguments and credential schemas.
* **Bug Fixes**
  * Compliance framework discovery now merges templates contributed for the same provider across multiple external sources.
* **Tests**
  * Added regression and expanded coverage for provider visibility rules and compliance-directory merging.


HugoPBrito:
I think we should clarify one thing here before the API consumes this contract.

Right now `get_scan_arguments()` and `get_connection_arguments()` receive `provider_uid`, but the base implementation drops it, and I don’t see provider-specific overrides for the app-facing providers that need it.

That matters because for providers like Azure, GCP, or M365 the UID is not just metadata. It is part of the scope we want to scan or validate, for example the Azure subscription or GCP project. If the API consumes this contract as-is, it can build arguments without the selected scope, which means the scan/connection test may fail or, worse, run against the wrong target.

Can we either add the provider-specific overrides here, with tests for the generated kwargs, or make it explicit that these methods are only base defaults and not yet the final API consumption contract?

AdriiiPRodri (changes requested):
Thanks for splitting this SDK side work out from the API wiring

The compliance entry-point merge looks good to me. The only thing is what Hugo said

josema-xyz (dismissed):

prowler/CHANGELOG.md
```diff
@@ -27,6 +27,9 @@ All notable changes to the **Prowler SDK** are documented in this file.
...
 - Public `Provider.get_class()` method that resolves a provider class by name for both built-in and external (entry-point) providers [(#11398)](https://github.com/prowler-cloud/prowler/pull/11398)
+- `sdk_only` provider property (default `true`) and `Provider.get_app_providers()`, so a provider (built-in or external) stays CLI/SDK-only and hidden from the app unless it declares `sdk_only = False` [(#11427)](https://github.com/prowler-cloud/prowler/pull/11427)
+- Compliance frameworks contributed by several external packages under the same provider are now merged instead of overwritten, so every entry-point directory a provider contributes is discovered [(#11578)](https://github.com/prowler-cloud/prowler/pull/11578)
+- `Provider.get_scan_arguments()`, `Provider.get_connection_arguments()` and `Provider.get_credentials_schema()` contract methods, so a provider persisted as a stored uid plus a secret dict can be constructed and validated programmatically (used by the API) [(#11578)](https://github.com/prowler-cloud/prowler/pull/11578)
```
danibarranqueroo: Update this deleting the changes not related to this PR and move it to the next unreleased minor version  please.

StylusFrost:
@HugoPBrito went with making it explicit rather than adding SDK overrides, since the uid handling for the app facing builtin providers already lives in the API. The API consumes this contract only for external providers, which override the base to inject their own uid; builtin providers like the Azure subscription or GCP project keep their explicit mapping on the API side and never reach these methods, so the API never builds args without scope. The base `get_scan_arguments` and `get_connection_arguments` now document that they intentionally drop `provider_uid` and that an external provider whose uid is part of the scope overrides to inject it. [`611215740`](https://github.com/prowler-cloud/prowler/commit/611215740507b322bfdd374903f8a3d25a305c60)

prowler/CHANGELOG.md
StylusFrost: Moved to the `5.31.0` unreleased minor in [`ffbc5e139`](https://github.com/prowler-cloud/prowler/commit/ffbc5e1397b3233cb749faa79621826d5b1060e6). The `#11398` and `#11427` entries come from the stacked parent PRs, so they only show against `master`; against the parent branch the diff adds just the `#11578` entries.

danibarranqueroo (dismissed):

danibarranqueroo (commented):
[Linode](https://github.com/prowler-cloud/prowler/pull/11633) was added as new provider, since it's sdk_only I think that is not needed to add that flag into the provider code, but please review it. Thanks!

StylusFrost:
Confirmed, nothing to add for Linode. Every provider defaults to `sdk_only = True` in the base `Provider` class, and a provider only becomes visible to the app (API/UI) by explicitly opting in with `sdk_only = False`. Since Linode is SDK-only, it inherits the default and the resolver already excludes it from `get_app_providers()`. Thanks for double-checking!

HugoPBrito (dismissed):

danibarranqueroo (dismissed):

AdriiiPRodri (dismissed):

danibarranqueroo (approved):

PR #11595, opened 2026-06-15
fix(azure/postgresql): isolate per-server collection failures

vahidg:
### Context

While scanning an Azure tenant I noticed the Azure PostgreSQL service silently under-reported flexible servers: a subscription with two flexible servers (`...-dev` and `...-prd`) only ever produced findings for one of them.

Root cause: `PostgreSQL._get_flexible_servers()` iterates the servers of a subscription, but the per-server collection runs inside a single **subscription-level** `try/except`. If any per-server call raises, the handler catches it, logs at `ERROR`, and abandons every server not yet processed — so one bad server discards the rest of the subscription.

In this case the trigger was **PostgreSQL 16+**, where the `connection_throttle.enable` server parameter no longer exists. `_get_connection_throttling()` therefore raised `ResourceNotFoundError (ConfigurationNotExists)`, which bubbled up and dropped the remaining servers. It was invisible in normal use because that `ERROR` log is hidden from the console at the default log level.

No linked issue — happy to open one if preferred.

### Description

- Wrap each server's collection in its own `try/except` so a single server's failure no longer aborts collection of the other servers in the subscription (the subscription-level handler around `servers.list()` is kept).
- Make `_get_connection_throttling()` tolerate the missing parameter and return `None` (mirroring the existing `_get_log_retention_days()`), so PostgreSQL 16+ servers are still collected. With `connection_throttling=None` the `postgresql_flexible_server_connection_throttling_on` check reports FAIL, matching its existing semantics.

Adds regression tests for both the missing-config case and a hard per-server failure.

### Steps to review

1. Read `_get_flexible_servers()` — per-server `try/except` now wraps the body; the subscription-level handler remains for `servers.list()`.
2. Read `_get_connection_throttling()` — now returns `None` on exception, consistent with `_get_log_retention_days()`.
3. Run the tests:
   ```
   pytest tests/providers/azure/services/postgresql/ -q
   ```
   - `test_missing_connection_throttle_config_still_collects_server` — both servers collected; the PG 16+ server gets `connection_throttling=None`.
   - `test_one_server_hard_failure_does_not_drop_others` — a server that fails on `servers.get` is skipped without dropping the rest.

This is a bugfix and is likely a backport candidate.

### Checklist

- [x] Review if code is being documented following the style guide.
- [ ] Review if backport is needed. (Looks applicable — deferring to maintainers.)
- [x] Review if is needed to change the Readme.md (not applicable).

#### SDK/CLI

### License

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resilience when collecting Azure PostgreSQL flexible server data: failures tied to a single server no longer block collection of other servers within the same subscription.
  * For PostgreSQL 16+, if `connection_throttle.enable` is genuinely absent (reported as not found), the missing setting is only left unset for that server; others remain populated.
  * If `connection_throttle.enable` fails for unexpected reasons or if a server’s details can’t be retrieved, only the affected server is skipped without impacting the rest.
* **Tests**
  * Added regression coverage for per-server failure isolation in flexible server collection.


HugoPBrito (changes requested):
Good fix direction, but I think this needs one more tightening before merge.

`_get_connection_throttling()` now catches every `Exception` and returns `None`. That handles the PostgreSQL 16+ missing `connection_throttle.enable` case, but it also turns permission errors, Azure throttling, transient SDK failures, or malformed responses into `connection_throttling=None`. The downstream check then reports the server as having connection throttling disabled, which hides a collection failure as a security finding.

Can we narrow this fallback to the expected missing-configuration case, for example `ConfigurationNotExists` / the specific Azure `ResourceNotFoundError`, and add a regression test proving unexpected configuration lookup failures are not silently converted to `None`?

vahidg:
@HugoPBrito  Rebased on `master` and resolved the conflict. Actually: `master` had already made this method resilient (via the HA / geo-redundant-backup PRs, #11045/#11046), but with a broad `except Exception` that returns `None` for *any* failure, including the permission and transient cases you flagged. I kept this PR's narrower approach instead:

  - `_get_connection_throttling()` catches only `ResourceNotFoundError` (the absent PG16+ parameter). Everything else propagates to the per-server handler in `_get_flexible_servers()`, so a lookup failure is logged and the server skipped, never reported as throttling disabled.
  - Master's additions are intact: the new `geo_redundant_backup` / `high_availability_mode` fields and the Entra-admin "authentication not enabled" warning.
  - I reconciled master's unit tests: the missing-parameter test now raises a real `ResourceNotFoundError`, and the unexpected-error test asserts the error propagates rather than being swallowed into `None`.

  This reverses a choice master already shipped, so if you'd rather keep the broader catch. I think distinguishing "parameter absent" from "lookup failed" is the right call, for the reasons in your review.

HugoPBrito:
Thank you very much for this contribution @vahidg! ❤️

HugoPBrito (approved):

PR #11618, opened 2026-06-16
fix(cli): prevent unrelated built-in provider failures from aborting the CLI

lydiavilchez:
### Context

Fixes a bug where a single built-in provider with a broken dependency would kill the entire CLI, even when the user invoked a completely different, healthy provider.

**Before:** if any built-in provider failed to import a dependency while the CLI was building its parser, it called `sys.exit(1)` and aborted everything.

**After:** the CLI only aborts when the failing provider is the one the user actually invoked. Any other broken provider just logs a warning and the CLI keeps working.

### Checklist

<details>

</details>

#### SDK/CLI

#### UI

#### API

### License

## Release Notes

* **Bug Fixes**
  * Improved CLI resilience when built-in providers have broken dependencies: failures in unrelated built-in providers now only produce warnings, while the CLI fails loudly only if the invoked provider itself is affected.
  * Provider invocation handling now correctly recognizes common CLI aliases when determining which provider to validate.

* **Tests**
  * Expanded coverage for the selective fail-loud behavior, including alias normalization and cases like help/no-provider invocation.

* **Documentation**
  * Updated the unreleased changelog note for this fix.


prowler/providers/common/arguments.py
```diff
@@ -10,12 +10,50 @@
...
+# User-facing provider names that map to a different canonical built-in name.
+# Single source of truth: parser.py imports this so the alias is recognised
+# both when sniffing argv pre-parse and when rewriting argv pre-argparse.
+PROVIDER_ALIASES = {
+    "microsoft365": "m365",
+    "oci": "oraclecloud",
+}
+
+
+def _invoked_provider_from_argv(available_providers: Sequence[str]) -> Optional[str]:
```
HugoPBrito, line 20: I think this heuristic is still too broad.

Right now we scan every non-flag token in `sys.argv`, but provider resolution later has more specific semantics: if the first argument is a flag, the parser injects the default `aws` provider. That means a command like `prowler --output-directory stackit` could incorrectly treat `stackit` as the invoked provider if it matches a provider name.

Can we make this helper follow the same provider-resolution rules as `ProwlerArgumentParser.parse()` instead of scanning arbitrary argv values?

prowler/providers/common/arguments.py
```diff
@@ -10,12 +10,50 @@
...
 
 def init_providers_parser(self):
     """init_providers_parser calls the provider init_parser function to load all the arguments and flags. Receives a ProwlerArgumentParser object"""
     # We need to call the arguments parser for each provider
     providers = Provider.get_available_providers()
+    # A built-in provider with a broken optional dependency should not tear
+    # down the whole CLI when the user invoked a different, sane provider.
+    # Only fail-loud on the invoked provider; otherwise warn and continue so
+    # the rest of the parser still builds.
+    invoked = _invoked_provider_from_argv(providers)
```
HugoPBrito: There is an alias gap here.

`ProwlerArgumentParser.parse()` normalizes `oci -> oraclecloud` and `microsoft365 -> m365`, but this detection runs before that normalization. So `prowler oci` may be treated as if no provider was explicitly invoked, and an `oraclecloud` import failure would be downgraded to a warning.

That would swallow a real failure for the provider the user actually requested. We should normalize aliases before deciding whether to fail loudly.

tests/providers/external/test_dynamic_provider_loading.py
```diff
@@ -466,6 +472,145 @@ def test_external_provider_does_not_touch_builtin_path(
 
         ext_cls.init_parser.assert_called_once_with(parser)
 
+    @patch("sys.argv", ["prowler", "aws"])
+    @patch("prowler.providers.common.arguments.Provider.is_builtin")
+    @patch("prowler.providers.common.arguments.import_module")
+    def test_unrelated_builtin_failure_does_not_abort_when_other_provider_invoked(
```
HugoPBrito, line 474: Good coverage for the main isolation path, but I think we need regression tests for the edge cases that define the contract here.

At minimum:
- `prowler oci` should fail loudly if `oraclecloud` argument loading fails.
- `prowler microsoft365` should fail loudly if `m365` argument loading fails.
- `prowler --output-directory stackit` should not treat `stackit` as the invoked provider.

Those are the cases where the current argv-scanning heuristic can behave differently from the real parser.

HugoPBrito (changes requested):
Thanks for working on this. I’m requesting changes because the provider detection still does not fully match the real CLI semantics.

The current approach can still fail loudly for an unrelated provider, or downgrade a real failure for an explicitly requested alias. I left inline comments with the specific cases and the regression tests I think we need before merging.

prowler/providers/common/arguments.py
lydiavilchez: Already fixed!

HugoPBrito: Sorry, didn't see it was already done.

HugoPBrito (approved):

PR #11630, opened 2026-06-17
fix(api): warm compliance caches in gunicorn master to stop worker-timeout crash loop

cesararroba:
### Context

The prowler-api ECS deploy fails with a worker crash loop → ALB health-check failures → deployment circuit breaker → rollback.

Root cause: gunicorn's ASGI workers each re-warmed all 16 providers' compliance caches in `post_fork` (CPU-bound, in a daemon thread), starving the worker's event loop past gunicorn's default 30s timeout so it was SIGKILLed before it could answer `/health/live`. (#11626 fixed the lifespan probe but not this.) Verified on dev ECS: the image at `master` HEAD still crash-loops.

This PR is the application half of the fix; the infrastructure half (an ECS startup health-check grace period) ships in the companion prowler-cloud PR. Both are needed together.

### Description

- Warm the compliance caches **once in the gunicorn master** (`on_starting`, pre-fork) instead of in every worker's `post_fork`. With `preload_app` enabled in production, the populated caches and readiness flags are inherited by every forked worker via copy-on-write, so no worker re-warms them and no worker's event loop is starved. Under `runserver` (DEBUG) the hook does not run, so lazy per-provider loading is preserved.
- Remove the `post_fork` hook, the `_warm_compliance_caches_in_background` helper, and the now-unused `threading` import.
- Add an explicit `timeout` setting (`GUNICORN_TIMEOUT`, default `120s`, raised from the 30s gunicorn default) so a slow cold boot cannot be SIGKILLed mid-startup before the worker can answer the load-balancer health check.

### Steps to review

- `api/src/backend/config/guniconf.py`: confirm the warm-up now lives inside `on_starting` guarded by `if not DEBUG`, that `post_fork` / `_warm_compliance_caches_in_background` are gone, and the new `timeout` setting is present.
- `api/src/backend/api/tests/test_compliance.py`: `TestGuniconfWarmsInMaster` asserts the master warms exactly once in production, never under DEBUG, and that no `post_fork` warm hook remains.
- Run: `pytest api/src/backend/api/tests/test_compliance.py -k TestGuniconfWarmsInMaster`.

### Checklist

#### SDK/CLI

#### API
- [x] Verify if API specs need to be regenerated. Not needed — no API surface change.
- [x] Check if version updates are required (e.g., specs, uv, etc.). Not needed.

### License

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive test suite verifying compliance cache warming behavior during Gunicorn master process startup, including DEBUG mode handling.

* **Chores**
  * Compliance cache warming mechanism moved from post-worker initialization to master process startup phase.
  * Added configurable Gunicorn timeout parameter with 120-second default.

### DIFF
--- a/api/src/backend/api/tests/test_compliance.py
+++ b/api/src/backend/api/tests/test_compliance.py
@@ -414,3 +414,38 @@ class TestWarmComplianceCaches:
             warm_compliance_caches([Provider.ProviderChoices.AWS])
 
         assert compliance_module.COMPLIANCE_WARMED.is_set()
+
+
+class TestGuniconfWarmsInMaster:
+    """The gunicorn master warms the compliance caches once in `on_starting`,
+    before any worker is forked, so preloaded workers inherit the populated
+    caches via copy-on-write instead of each re-warming post-fork."""
+
+    def _on_starting(self):
+        from config import guniconf
+
+        return guniconf.on_starting
+
+    def test_warms_once_in_production(self):
+        with (
+            patch("config.guniconf.DEBUG", False),
+            patch("config.guniconf.warm_compliance_caches") as mock_warm,
+        ):
+            mock_warm.return_value = []
+            self._on_starting()(MagicMock())
+
+        mock_warm.assert_called_once_with()
+
+    def test_does_not_warm_under_debug(self):
+        with (
+            patch("config.guniconf.DEBUG", True),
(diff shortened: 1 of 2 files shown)


cesararroba:
Closing — the gunicorn/compliance-warmup fix will be handled on the OSS side; not landing it through this PR.

PR #11659, opened 2026-06-22
feat(ui): filter by provider group across main views

pfe-nazaries:
### Context

Provider groups let users organize cloud accounts into logical groups. The main views could already be scoped by provider type and provider account, but not by provider group. This PR adds a provider group filter so data can be scoped to the accounts that belong to one or more groups.

https://github.com/user-attachments/assets/688b2761-efb2-4c54-a397-3e598e5c483c

https://github.com/user-attachments/assets/160cdf7a-382a-4294-ad23-9f73d2cd219c

https://github.com/user-attachments/assets/e8091a56-26fb-4819-863b-d1b92aa91a2f

https://github.com/user-attachments/assets/739a8449-6bb2-4c6a-b227-24ca6f6e9f1a

https://github.com/user-attachments/assets/f644e4af-e1f0-4499-9dd0-e694eb44b77b

### Description

- New `ProviderGroupSelector` filter control, backed by a `manage-groups` Server Action that lists the tenant's provider groups
- Provider group filter wired into the filter bars of the Overview, Findings, Resources, Scans, and Providers views
- Per-view filter helpers translate the selected groups into the provider-scope query params sent to the API
- `FilterParam` type split into core fields plus per-view types
- Unit tests for the selector, the `manage-groups` action, and the per-view filter helpers

### Steps to review

With at least one provider group defined (containing some accounts), open each page and use the new **Provider group** filter:

1. Overview — `/`
2. Findings — `/findings`
3. Resources — `/resources`
4. Scans — `/scans`
5. Providers — `/providers`

Confirm the data is scoped to the accounts in the selected group(s) and that combining it with the existing provider type / account filters behaves as expected.

### Checklist

<details>

</details>

#### UI (if applicable)
- [ ] Ensure new entries are added to ui/CHANGELOG.md

### License

## Summary by CodeRabbit

* **New Features**
  * Added provider group filtering across the Overview, Findings, Resources, Scans, and Providers views via a new provider group selector.
* **Improvements**
  * Updated risk charts and drill-down flows to respect selected provider groups, including preserving provider-group filters when navigating from severity trends.
  * Filter chips now display provider-group names (when available) instead of raw IDs.


pfe-nazaries:
## CodeRabbit review — addressed

Triaged all 12 review threads against the current code. Commit `f08d7c9f1` on `feature/add-provider-group-filter`.

### ✅ Fixed (6)

| Thread | Change |
|--------|--------|
| `manage-groups.ts` — partial group data | `getAllProviderGroups` now aborts (returns `undefined`) when a later page resolves to an API error payload, instead of treating it as "no more pages" and silently truncating |
| `manage-groups.test.ts` — regression | Added a test for an error payload on page N (asserts `undefined`, no partial merge) |
| `scans/page.tsx` — pending rows | Synthetic pending-schedule rows now honor `filter[provider_groups__in]` via provider group membership, matching backend-filtered scan rows |
| `scans/page.tsx` — prefetch | `Promise.all` → `Promise.allSettled` so a transient failure of either provider/group fetch no longer fails the whole route |
| `ProviderGroupSelector` relocation (×2) | Moved from `app/(prowler)/_overview/_components/` to `components/filters/` (used across 5 views); all imports and test mocks updated |

Verified locally: `tsc` ✅ · `eslint` ✅ · affected unit tests 41/41 ✅

### ⏭️ Not changed, with rationale

**Const-object refactor of inline filter-param literals** (`findings-filters.ts`, `resources-filters.ts`, `scans-filters.ts` `state__in`/`trigger`): kept as inline literal unions. A const-object referenced **only** in a `typeof` position trips this repo's own `@typescript-eslint/no-unused-vars` (`"...is assigned a value but only used as a type"`). The const-object pattern is therefore reserved for fields that also have a runtime use (e.g. `FILTER_FIELD`, `SCANS_PROVIDER_FILTER_FIELD`); inline literals are the lint-clean idiom for type-only param keys here.

**`overview-filters.ts` / `providers-filters.ts` "inline union"**: these already derive from const objects via indexed access — `(typeof FILTER_FIELD)["PROVIDER_TYPE" | ...]` and `(typeof PROVIDERS_PAGE_FILTER)[...]`. No bare value-union is present, so there is nothing to convert.

**`use-related-filters.ts` default `providerFilterType`**: out of scope for this PR. The diff only renames `FilterType` → `FILTER_FIELD` (identical semantics), and the hook currently has no callers anywhere in `ui/`, so there is no behavior to change here. Worth a separate look if/when the hook is wired up.

Alan-TheGentleman:
**The group + account combination silently drops the group filter.** In `risk-pipeline-view.ssr.tsx` and `risk-plot.ssr.tsx`, the `providerIdFilter` branch rebuilds the filters object from scratch and never re-adds `provider_groups__in`, so selecting a group then an account ignores the group. Both SSR helpers are untested, which is why a branch like this regresses quietly. Could we add unit tests on the per-branch filter building, including the empty-group case?

**Duplication worth folding before it spreads:**
- `getProviderGroupDisplayValue` is identical in `findings-filters.utils.ts` and `resources-filters.utils.ts`. Pull it into a shared util.
- The `groups -> providers` logic is duplicated in both SSR files. A shared `filterProvidersByGroups(providers, groupIds)` keeps them in sync.
- `risk-plot.ssr.tsx` and `finding-severity-over-time.tsx` still use the literal `"filter[provider_groups__in]"` instead of `OVERVIEW_FILTER_PARAM.PROVIDER_GROUPS`, so a later rename breaks them with no type error.

**Worth verifying (not blocking):** `getAllProviderGroups` paginates sequentially across all five views (confirm there's a server-side cap), truncates silently at 5000 groups (surface a signal instead), and forwards URL UUIDs without the validation `provider_type__in` gets.

Happy to pair on the tests if that helps.

alejandrobailo (changes requested):
Really nice work, the refactor is clean and the test coverage is solid. A few small things I'd like to sort out before merging, mostly consistency cleanups, none of them blockers on their own:

1. In `scans/page.tsx` you switched to `Promise.allSettled`, but the other pages still use `Promise.all`. Both `getAllProviders` and `getAllProviderGroups` already catch internally and return `undefined`, so they never reject, which means the `.status === "fulfilled"` branches here are dead code. I'd go back to `Promise.all` to keep it consistent with findings/resources/overview.

2. In `provider-group-selector.tsx` the filter key is hardcoded as `"provider_groups__in"` instead of reusing `FILTER_FIELD.PROVIDER_GROUPS`. Same value today, but it can drift over time and TS won't catch it. Let's reference the constant.

3. The selector uses hardcoded DOM ids (`provider-group-selector` / `provider-group-label`). No page renders two of these today so it works, but `AccountsSelector` already takes an `id` prop for exactly this reason. I'd add one here too just to be safe.

4. In the `providerTypeFilter` branch of `risk-pipeline-view.ssr.tsx` we enumerate the types straight from the filter without intersecting with `scopedProviders`, so we rely on the API param to scope by group. It's not a data leak, but we can end up firing redundant calls or showing an empty node for a type that has no providers in the selected group. Worth a second look.

5. `parseFilterIds` doesn't validate the group ids coming from the URL. Low impact since the backend `UUIDInFilter` rejects malformed uuids with a 400, but leaving it noted.

pfe-nazaries:
> Really nice work, the refactor is clean and the test coverage is solid. A few small things I'd like to sort out before merging, mostly consistency cleanups, none of them blockers on their own:
> 
> 1. In `scans/page.tsx` you switched to `Promise.allSettled`, but the other pages still use `Promise.all`. Both `getAllProviders` and `getAllProviderGroups` already catch internally and return `undefined`, so they never reject, which means the `.status === "fulfilled"` branches here are dead code. I'd go back to `Promise.all` to keep it consistent with findings/resources/overview.
> 2. In `provider-group-selector.tsx` the filter key is hardcoded as `"provider_groups__in"` instead of reusing `FILTER_FIELD.PROVIDER_GROUPS`. Same value today, but it can drift over time and TS won't catch it. Let's reference the constant.
> 3. The selector uses hardcoded DOM ids (`provider-group-selector` / `provider-group-label`). No page renders two of these today so it works, but `AccountsSelector` already takes an `id` prop for exactly this reason. I'd add one here too just to be safe.
> 4. In the `providerTypeFilter` branch of `risk-pipeline-view.ssr.tsx` we enumerate the types straight from the filter without intersecting with `scopedProviders`, so we rely on the API param to scope by group. It's not a data leak, but we can end up firing redundant calls or showing an empty node for a type that has no providers in the selected group. Worth a second look.
> 5. `parseFilterIds` doesn't validate the group ids coming from the URL. Low impact since the backend `UUIDInFilter` rejects malformed uuids with a 400, but leaving it noted.

- 1 to 3 applied 
- 4 is intended as I discussed with @AdriiiPRodri, we will iterate on this
- 5 deferred

alejandrobailo (dismissed):

alejandrobailo (approved):

PR #11669, opened 2026-06-23
feat(config): add compliance guardrails for the SDK config

pedrooot:
### Description

Some compliance requirements only really pass if the checks behind them ran with a strict enough config. Right now you can loosen a check's config and the requirement still shows up as PASS even though it isn't actually met.

CIS AWS 6.0 req 2.11 is a good example ("credentials unused for 45 days or more are disabled"). It maps to `iam_user_accesskey_unused`, which reads `max_unused_access_keys_days`. Bump that to 120 days and the check happily passes, but the requirement clearly isn't satisfied.

So a requirement can now declare what it expects from the config via a new `ConfigRequirements` field. When we build the compliance outputs we check those expectations against the config the scan actually used (`audit_config`). If a check ran with a config too loose for the requirement, that requirement is forced to FAIL and we prepend `[CONFIG NOT VALID]` plus the reason to `status_extended`.

Each constraint is `{Check, ConfigKey, Operator, Value}`. Operators:

- `lte` / `gte` / `eq` for scalars (thresholds, boolean toggles)
- `in` –> applied value must be one of `Value`
- `subset` –> applied list must be a subset of `Value` (allowlists, e.g. `recommended_minimal_tls_versions`)
- `superset` –> applied list must be a superset of `Value` (denylists, e.g. `insecure_key_algorithms`)

The actual evaluation lives in one place (`prowler/lib/check/compliance_config_eval.py`) so the SDK outputs and the App backend share the same logic. The FAIL override is applied to CSV, OCSF and the console tables.

One thing worth calling out: if a `ConfigKey` isn't explicitly set, we skip the constraint and assume the check's default already matches what the requirement wants. So nothing changes for anyone running the default config.

On the model side: `ConfigRequirements` on the legacy `Compliance_Requirement` and `config_requirements` on `UniversalComplianceRequirement`, wired through `adapt_legacy_to_universal`. Then I added the actual `ConfigRequirements` entries to the relevant requirements across the AWS, Azure, GCP, Alibaba Cloud, Kubernetes, M365 and Okta frameworks, only for requirements whose control text prescribes the concrete value.

### Steps to review

### Checklist

<details>

</details>

#### SDK/CLI

#### UI

#### API

### License

## Summary by CodeRabbit

* **New Features**
  * Added requirement-level configuration guardrails: configurable checks now report **FAIL** when the scan’s applied **audit configuration** violates requirement constraints.
  * Compliance outputs (CSV and console tables) are now config-aware across frameworks and threat scoring, using evaluated “effective” **PASS/FAIL** and updated **status**/extended messages.
  * When invalid, statuses are overridden to **FAIL** and **`[CONFIG NOT VALID]`** is appended with the first violated constraint reason.
* **Documentation**
  * Expanded guidance for authoring and interpreting configuration guardrails.
* **Tests**
  * Added unit and integration tests covering constraint parsing, operator/type semantics, provider scoping, and correct aggregation/muted handling.


prowler/lib/check/compliance_models.py
```diff
@@ -304,6 +304,78 @@
...
+    # them to 0/1.
+    Value: Union[bool, int, float, str, list[Any]]
+    # Provider this constraint applies to (e.g. ``aws``), matched
+    # case-insensitively. ``None`` applies whenever the requirement runs
+    # (single-provider frameworks).
+    Provider: Optional[str] = None
+
+    @root_validator
+    # noqa: F841 - since vulture raises unused variable 'cls'
+    def validate_value_matches_operator(cls, values):  # noqa: F841
```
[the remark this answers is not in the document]
pedrooot: fixed!

prowler/lib/check/compliance_models.py
```diff
@@ -329,6 +401,7 @@ class Compliance_Requirement(BaseModel):
         ]
     ]
     Checks: list[str]
+    ConfigRequirements: Optional[list[Compliance_Requirement_ConfigConstraint]] = None
```
HugoPBrito, line 407: `ConfigRequirements` is now supported on the generic `Compliance_Requirement`, but `mitre_attack_aws.json` also includes this field and `Mitre_Requirement` does not define it.

Because of that, Pydantic drops these constraints during MITRE parsing and the config validation logic never sees them. I verified this by parsing `mitre_attack_aws.json`: the raw JSON requirement has `ConfigRequirements`, but the parsed requirement object does not retain it.

Could you add `ConfigRequirements: Optional[list[Compliance_Requirement_ConfigConstraint]] = None` to `Mitre_Requirement` as well, and include a regression test proving that a violated MITRE config requirement forces the compliance result to FAIL?

prowler/lib/check/compliance_config_eval.py
```diff
@@ -0,0 +1,404 @@
...
+            config_key = getattr(constraint, "ConfigKey", None)
+            operator = getattr(constraint, "Operator", None)
+            expected = getattr(constraint, "Value", None)
+            provider = getattr(constraint, "Provider", None)
+
+        # Constraint scoped to another provider → not applicable to this scan.
+        # Compared case-insensitively (and trimmed) so a constraint authored as
+        # e.g. "AWS" still scopes to the "aws" scan instead of being silently
+        # bypassed by a casing/format mismatch.
+        if (
```
HugoPBrito, line 171: This provider filter only skips constraints from another provider when `provider_type` is available.

For universal multi-provider frameworks, relying on the global provider makes this fragile: an AWS output can accidentally evaluate Azure-scoped constraints if the caller does not pass the provider explicitly.

Could we make the universal output paths pass the current output provider as `provider_type`, and add a regression test proving that an Azure constraint does not affect an AWS output?

prowler/lib/outputs/compliance/universal/ocsf_compliance.py
```diff
@@ -293,7 +309,7 @@ def _build_compliance_finding(
                         else None
                     ),
                 ),
-                message=finding.status_extended,
+                message=message,
```
HugoPBrito, line 315: This updates the OCSF `message`, but the event can still report conflicting statuses.

The nested compliance object uses the effective status calculated from `apply_config_status(...)`, while the top-level fields below still use the raw finding status (`status_code=finding.status` and `status_detail=finding.status_extended`). So a config-invalid PASS can produce an event where `compliance.status_id` says FAIL while `status_code` still says PASS.

Could you use `effective_status` and `message` for the top-level status fields as well? If we need to preserve the original check result, the nested `Check` object already seems like the right place for that raw status.

prowler/lib/outputs/compliance/cis/cis_aws.py
```diff
@@ -34,10 +38,19 @@ def transform(
         Returns:
             - None
         """
+        requirement_config_status = build_requirement_config_status(
```
HugoPBrito, line 41: Suggestion: this effective-status plumbing is now repeated across several compliance transformers.

I would not block on this because the per-provider formatter pattern is already established in Prowler, but a small shared helper for “matching requirement + effective status + effective status_extended” would reduce future drift and make this easier to maintain.

HugoPBrito (changes requested):
Thanks for working on this. However, after reviewing it, I think we should improve a few things before merging.

One additional blocker could not be placed as an inline comment because `prowler/lib/outputs/compliance/universal/universal_output.py` is not part of this PR diff: `UniversalComplianceOutput` still writes the raw finding `Status` and `StatusExtended`, so a config-invalid PASS can be forced to FAIL in OCSF/table outputs while remaining PASS in the universal CSV. Please apply the same effective-status logic there as well.

pedrooot:
Thanks for the revision @HugoPBrito 🔝

danibarranqueroo (changes requested):
Good job, apart from @HugoPBrito review. I have a doubt about this:

Every dedicated compliance output got wired to apply_config_status / build_requirement_config_status except okta_idaas_stig, which was overlooked in both renderers:

- prowler/lib/outputs/compliance/okta_idaas_stig/okta_idaas_stig_okta.py:57-58 (CSV — still Status=finding.status / StatusExtended=finding.status_extended.
- prowler/lib/outputs/compliance/okta_idaas_stig/okta_idaas_stig.py:45-60 (CLI table) —still branches on raw finding.status.

Meanwhile this same PR adds 4 ConfigRequirements to prowler/compliance/okta/okta_idaas_stig_v1r2_okta.json (okta_max_session_idle_minutes, okta_admin_console_idle_timeout_max_minutes, okta_user_inactivity_max_days, okta_max_session_lifetime_minutes).

pedrooot:
@danibarranqueroo thanks for the review, fixed!

HugoPBrito (approved):

danibarranqueroo (approved):

PR #11694, opened 2026-06-25
feat(sdk): replace detect-secrets library with kingfisher

danibarranqueroo:
### Context

`detect-secrets` (the library powering Prowler's secret-scanning checks) is effectively in maintenance-only mode, so this PR replaces it with an actively-maintained engine.

After evaluating the candidates, **Kingfisher** (MongoDB, Apache-2.0) was selected because of its active maintenance, detection coverage, and distribution: it ships a cross-platform `pip` wheel (`kingfisher-bin`), so it is bundled with `pip install prowler` and needs no separate install for self-hosted users.

### Description

Replaces the `detect-secrets` dependency with `kingfisher-bin==1.104.0` and rewrites `detect_secrets_scan()` (`prowler/lib/utils/utils.py`) to shell out to Kingfisher. The function keeps the same signature and return shape, so the 16 secret-scanning checks (12 AWS + 4 OpenStack) need no contract changes.

- **Offline by default.** Scans run with `--no-validate --no-update-check`, so no data leaves the audited environment. Uses Kingfisher's built-in ruleset at `low` confidence to retain the generic keyword/password coverage `detect-secrets` provided.
- **Opt-in live validation (new).** A `--scan-secrets-validate` flag and `secrets_validate` config option validate discovered secrets against the provider APIs (the secret itself is used as the credential — no extra Prowler permissions needed). Secrets confirmed live are escalated to **critical** and annotated in `status_extended`. Off by default.
- **Behavior change.** Obvious placeholder values (e.g. `AKIA…EXAMPLE`, `password=1234…`) are no longer reported — a precision improvement that reduces false positives. Test fixtures were updated to realistic fake secrets accordingly.
- **Docs** updated: configuration-file tutorial (config key + flag + egress warning) and the pentesting page (now credits Kingfisher with a link).

**Known coverage differences vs detect-secrets** (acceptable; documented):
- GitHub PAT / npm / Mailchimp *synthetic* static tokens are not matched offline (real, valid-checksum tokens are — and validation confirms them).
- Telegram bot tokens have no Kingfisher rule (detect-secrets had it disabled by default anyway).

### Steps to review

1. `uv sync` (pulls `kingfisher-bin`, drops `detect-secrets`).
2. Run the secret checks offline:
 `uv run python prowler-cli.py aws -p <profile> -c ec2_instance_secrets_user_data ec2_launch_template_no_secrets ssm_document_secrets cloudwatch_log_group_no_secrets_in_logs … ` → confirm FAIL findings render with Kingfisher type names (e.g. `Generic Password`, `Stripe Secret / Restricted Key`).
3. Opt into validation: add `--scan-secrets-validate` (or set `secrets_validate: True` in the config) → confirm it runs; live secrets would escalate to critical (offline-only environments will see no live confirmations).
4. Tests: `uv run pytest -p no:randomly tests/lib/utils/utils_test.py tests/providers/aws/services/{autoscaling,awslambda,cloudformation,cloudwatch,codebuild,ec2,ecs,glue,ssm,stepfunctions}/tests/providers/openstack/` → 543 pass.
5. Verified end-to-end against a live AWS account (seeded throwaway resources, since torn down).

### Checklist

<details>

</details>

#### SDK/CLI

#### UI

#### API

### License

## Summary by CodeRabbit

* **New Features**
  * Secret scanning now uses the Kingfisher engine and runs fully offline by default.
  * Added optional live secret validation via `aws.secrets_validate` and `--scan-secrets-validate` to confirm whether discovered secrets are live against provider APIs.
* **Bug Fixes**
  * Verified (live) secrets are now consistently annotated across affected AWS and OpenStack secret-scanning checks and are treated as critical.
* **Documentation**
  * Updated user tutorials to cover offline scanning, validation scope, and the warning that validation triggers outbound authenticated calls visible in audited account logs.
  * Added developer documentation for the required secret-scanning check structure and updated navigation.
* **Chores**
  * Removed deprecated `detect_secrets_plugins` configuration support and updated related schema and tests.


HugoPBrito:
I moved this into a proper review.

Please review the inline comments there, they point to the exact places that need changes.

prowler/lib/utils/utils.py
```diff
@@ -111,77 +164,153 @@ def hash_sha512(string: str) -> str:
...
-
-        if detect_secrets_output:
-            if file:
-                return detect_secrets_output[file]
-            else:
-                return detect_secrets_output[temp_data_file.name]
-        else:
-            return None
+            return
+
```
HugoPBrito, line 221: This currently fails open.

If Kingfisher exits non-zero, times out, returns malformed JSON, or fails at runtime, we log the error and continue with empty/partial results. Several callers then keep the default `PASS`, so a scanner failure can become “no secrets found”.

That is not acceptable for a security check. “Scanner failed” and “no secrets found” are different states.

Please make this explicit, either by raising/returning a typed scan error and making callers report `MANUAL`/error-like findings instead of `PASS`. We also need tests for non-zero exit, timeout, malformed/empty JSON, and missing binary/runtime failure.

prowler/providers/aws/services/cloudwatch/cloudwatch_log_group_no_secrets_in_logs/cloudwatch_log_group_no_secrets_in_logs.py
```diff
@@ -11,95 +14,130 @@
...
-                if log_group_secrets:
-                    secrets_string = "; ".join(log_group_secrets)
-                    report.status = "FAIL"
-                    report.status_extended = f"Potential secrets found in log group {log_group.name} {secrets_string}."
-                findings.append(report)
+                    else:
+                        for secret_type in entry["types"]:
+                            secrets_dict.add_secret(1, secret_type)
+                    log_stream_secrets[cloudwatch_timestamp] = secrets_dict
+                if log_stream_secrets:
```
HugoPBrito, line 183: This can still produce a false `FAIL` after ignore filtering.

For multiline events, if `rescan_results` is empty because `secrets_ignore_patterns` filtered everything out, we still add an empty `SecretsDict` to `log_stream_secrets`. That makes the report path truthy and can emit a `FAIL` with no real remaining secret evidence.

Please only add the timestamp when there is at least one non-ignored secret left after the rescan, and add a regression test for this case.

HugoPBrito (changes requested):
Requesting changes here.

The main issue is that we are mixing up two completely different states: “the scanner found no secrets” and “the scanner failed”. For a security check, those cannot collapse into the same outcome. I left the concrete points inline.

prowler/providers/aws/services/cloudwatch/cloudwatch_log_group_no_secrets_in_logs/cloudwatch_log_group_no_secrets_in_logs.py
```diff
@@ -11,95 +15,170 @@
...
-                            data=log_stream_data,
-                            excluded_secrets=secrets_ignore_patterns,
-                            detect_secrets_plugins=logs_client.audit_config.get(
-                                "detect_secrets_plugins",
-                            ),
+                if not log_group.log_streams:
+                    continue
+                for log_stream_name, events in log_group.log_streams.items():
+                    yield (
+                        (log_group.name, log_stream_name),
```
HugoPBrito: I believe this key is not unique enough.

CloudWatch log group names and stream names are not unique across regions, so two different regions can produce the same `(log_group.name, log_stream_name)` key. Since `detect_secrets_scan_batch()` groups findings by this key, results from one region can be reused for another one, causing false FAILs or even indexing the wrong event list later when reading `events[secret["line_number"] - 1]`.

Can we use a region-aware or ARN-based key here, for example `(log_group.arn, log_stream_name)` or `(log_group.region, log_group.name, log_stream_name)`, and use the same key consistently in `stream_results`, `stream_plans`, `rescan_payloads`, and `groups_with_rescan`?

It would also be good to add a regression test with two log groups in different regions sharing the same group and stream names, where only one contains a secret.

HugoPBrito (changes requested):
Thanks for the work on this PR. I believe there is still one correctness issue we should fix before merging.

Please check the inline comment.

prowler/providers/aws/services/cloudwatch/cloudwatch_log_group_no_secrets_in_logs/cloudwatch_log_group_no_secrets_in_logs.py
danibarranqueroo: Totally right! Done.

prowler/providers/aws/services/cloudwatch/cloudwatch_log_group_no_secrets_in_logs/cloudwatch_log_group_no_secrets_in_logs.py
```diff
@@ -11,95 +15,172 @@
...
+                            "multiline": multiline,
+                            "types": [],
+                        }
+                    if multiline:
+                        # More informative output is possible with more than one
+                        # line: the event is rescanned to get the type and line
+                        # number of each secret.
+                        rescan_payloads[
+                            (log_group.arn, log_stream_name, cloudwatch_timestamp)
+                        ] = log_event_data
```
HugoPBrito, line 108: This still needs one more discriminator in the rescan key.

Right now the multiline payload is keyed by:

`(log_group.arn, log_stream_name, cloudwatch_timestamp)`

That is better than the previous `(name, stream)` key, but it is still not unique per event. CloudWatch can have multiple events in the same stream with the same millisecond timestamp. If that happens, the later multiline event overwrites the earlier payload before the rescan.

That means we can lose secret evidence or report incomplete results. For a security check, that is not a safe assumption.

Please include a stable event-level discriminator in this key, for example the event index or event id if available:

`(log_group.arn, log_stream_name, cloudwatch_timestamp, event_index)`

The final output can still group/display by timestamp, but the internal rescan identity needs to be per event, not only per timestamp.

HugoPBrito (changes requested):
Requesting changes on one remaining CloudWatch edge case.

The previous issues look fixed, but the multiline rescan key still is not unique enough.

HugoPBrito (changes requested):
Thanks for the updates. The previous CloudWatch batching/keying concerns look addressed, and the targeted tests pass locally.

One remaining cleanup issue: `prowler/config/config.yaml` still exposes `openstack.detect_secrets_plugins` even though this PR removes `detect_secrets_plugins` from the OpenStack schema and the changelog says the option is no longer used.

Because provider schemas allow extra keys, this will likely be silently preserved in `audit_config` while no check consumes it anymore. That leaves users with an official-looking option that no longer has any effect.

Please remove the stale OpenStack `detect_secrets_plugins` block from `prowler/config/config.yaml` and update the nearby comment from `detect-secrets scanning` to the new generic/Kingfisher wording.

HugoPBrito (approved):

PR #11723, opened 2026-06-29
feat(github): support repository rulesets in default branch protection checks

### LINKED ISSUE #11706: Support GitHub Repository Rulesets in Prowler GitHub Checks
### Feature search

Prowler CLI/SDK

GitHub

### New feature motivation

GitHub is slowly phasing out "classic" branch protection rules in favor of "rulesets". But Prowler fails to consider protections enforced through rulesets, which results in spurious findings like `repository_default_branch_disallows_force_push` or `repository_default_branch_requires_signed_commits`.

### Solution Proposed

Consider both classic branch protection rules and rulesets before reporting any findings. 

### Use case and benefits

Unlike classic branch protection rules, GitHub can enforce rulesets at the org level, which is something Prowler should be encouraging!

Ignore the findings, or move back to enforcing rules using "classic" branch protection rules.

### Additional context

This issue was previously raised in https://github.com/prowler-cloud/prowler/issues/9530, closed as completed, but Prowler 5.31.1 is still reporting these spurious findings...

danibarranqueroo:
### Context

GitHub is phasing out classic branch protection in favor of rulesets (which can also be enforced at the org level). Prowler's GitHub repository checks only read classic branch protection, so repositories protected via rulesets get false-positive findings (e.g. `repository_default_branch_disallows_force_push`, `repository_default_branch_requires_signed_commits`).

Fix #11706

### Description

The service already evaluated rulesets for a single attribute (`dismiss_stale_reviews`). This generalizes that pattern: rulesets are fetched once per repository and every default-branch protection rule is mapped to its classic equivalent (`non_fast_forward`, `required_signatures`, `required_linear_history`, `required_status_checks`, `deletion`, and `pull_request` parameters).

A control now passes when **either** classic branch protection **or** an active ruleset enforces it. Classic protection takes precedence when it already passes; when a protection is configured in a ruleset that is not active, the check fails with a "ruleset is not active" message. `enforce_admins` is inferred from an active ruleset with no bypass actors.

All 11 default-branch protection checks now honor rulesets (plus the existing `dismiss_stale_reviews`). Checks unrelated to branch protection are untouched.

### Steps to review

- `repository_service.py`: `_evaluate_default_branch_rulesets` (single fetch + rule mapping) and `_merge_ruleset_bool` (classic-vs-ruleset merge, with polarity for force-push/deletion), wired into `_process_repository`; new `*_source` fields on the `Branch` model.
- Each affected check now surfaces the `ruleset_not_active` FAIL message.
- Metadata wording updated to mention rulesets.
- Tests: service-level coverage for every rule type (active/inactive, classic precedence, bypass actors) and per-check ruleset PASS / ruleset_not_active FAIL cases.

## Summary by CodeRabbit

* **New Features**
  * GitHub default-branch protection checks now evaluate both classic branch protection and repository rulesets to determine effective enforcement.

* **Bug Fixes**
  * Reduced false positives by handling rulesets that are present but not active.
  * Added more specific FAIL explanations for inactive ruleset-driven protections (force-push restriction, admin enforcement, deletion, and required review/history/status controls).

* **Documentation & Tests**
  * Updated check descriptions and links to include rulesets guidance.
  * Expanded unit test coverage for active vs inactive ruleset scenarios across default-branch protections.

### DIFF
--- a/prowler/providers/github/services/repository/repository_service.py
+++ b/prowler/providers/github/services/repository/repository_service.py
@@ -230,9 +230,12 @@ class Repository(GithubService):
         inactive_approval_count: Optional[int] = None
         any_active_ruleset = False
         any_inactive_ruleset = False
-        # An active ruleset with no bypass actors enforces protection for everyone,
-        # including administrators (the rulesets equivalent of "enforce admins").
+        # A ruleset with no bypass actors applies to everyone, including administrators
+        # (the rulesets equivalent of "enforce admins"). A ruleset that has bypass actors
+        # would not apply to admins even if activated, so it must not drive the
+        # enforce-admins finding in either the active or the inactive case.
         admins_enforced_active = False
+        admins_configured_inactive = False
 
         for ruleset in rulesets:
             if ruleset.get("target") != "branch":
@@ -247,12 +250,15 @@ class Repository(GithubService):
             if not (is_active or is_inactive):
                 continue
 
+            has_no_bypass_actors = not (ruleset.get("bypass_actors") or [])
             if is_active:
                 any_active_ruleset = True
-                if not (ruleset.get("bypass_actors") or []):
+                if has_no_bypass_actors:
                     admins_enforced_active = True
             else:
                 any_inactive_ruleset = True
+                if has_no_bypass_actors:
+                    admins_configured_inactive = True
 
             bucket = active if is_active else inactive
 
@@ -319,7 +325,7 @@ class Repository(GithubService):
 
         if admins_enforced_active:
             result["enforce_admins"] = (None, "ruleset")
-        elif any_inactive_ruleset:
+        elif admins_configured_inactive:
             result["enforce_admins"] = (None, "ruleset_not_active")
 
         return result
--- a/prowler/providers/github/services/repository/repository_service.py
+++ b/prowler/providers/github/services/repository/repository_service.py
@@ -202,15 +202,37 @@ class Repository(GithubService):
 
         return None
 
-    def _get_dismiss_stale_reviews_from_rulesets(
+    def _evaluate_default_branch_rulesets(
         self, repo, default_branch: str
-    ) -> tuple[Optional[bool], Optional[str]]:
-        """Evaluate dismiss-stale-review coverage from repository and parent rulesets."""
-        rulesets = self._get_repository_rulesets(repo)
-        if rulesets is None:
-            return None, None
+    ) -> dict[str, tuple[Optional[int], str]]:
+        """Evaluate default-branch protection coverage provided by rulesets.
 
-        has_inactive_ruleset = False
+        Fetches the repository (and parent) rulesets once and walks every rule that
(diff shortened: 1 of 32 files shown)


danibarranqueroo:
Changes tested on personal repository, all working as expected.

Captura de pantalla 2026-06-29 a las 17 32 50
Captura de pantalla 2026-06-29 a las 17 33 06

prowler/providers/github/services/repository/repository_service.py
```diff
@@ -219,26 +244,116 @@ def _get_dismiss_stale_reviews_from_rulesets(
...
-            return False, "ruleset_not_active"
+            for rule in ruleset.get("rules") or []:
+                rule_type = rule.get("type")
+                params = rule.get("parameters") or {}
+
+                if rule_type == "required_linear_history":
+                    bucket.add("required_linear_history")
+                elif rule_type == "required_signatures":
+                    bucket.add("require_signed_commits")
+                elif rule_type == "required_status_checks":
```
HugoPBrito, line 273: This should not pass just because the `required_status_checks` rule exists.

GitHub's ruleset docs say `strict_required_status_checks_policy` does not take effect unless at least one status check is enabled, and here we mark `status_checks` as enforced even when `parameters.required_status_checks` is empty. That creates a false PASS for `repository_default_branch_status_checks_required`.

Please only add `status_checks` when `params.get("required_status_checks")` contains at least one configured check.

prowler/providers/github/services/repository/repository_service.py
```diff
@@ -219,26 +244,116 @@ def _get_dismiss_stale_reviews_from_rulesets(
...
+            is_inactive = enforcement in {"disabled", "evaluate"}
+            if not (is_active or is_inactive):
+                continue
 
-                dismiss_stale_reviews = (rule.get("parameters") or {}).get(
-                    "dismiss_stale_reviews_on_push"
-                )
-                if dismiss_stale_reviews is not True:
-                    continue
+            has_no_bypass_actors = not (ruleset.get("bypass_actors") or [])
```
HugoPBrito, line 253: I don't think we can safely infer admin enforcement from an empty or missing `bypass_actors` field.

GitHub only returns `bypass_actors` when the requester has write access to the ruleset. With a read-only scanner token, this field can be absent even if admins or roles can bypass the ruleset, which would make `repository_default_branch_protection_applies_to_admins` pass incorrectly.

Please avoid treating missing `bypass_actors` as proof that admins cannot bypass unless the response guarantees the field is complete.

HugoPBrito (changes requested):
Request changes. The ruleset support is a good direction, but there are two false-negative risks we should fix before merging.

prowler/providers/github/services/repository/repository_service.py
danibarranqueroo: This is already covered in a MANUAL return when there is no write permission

HugoPBrito (approved):

HugoPBrito:
🚀

PR #11740, opened 2026-06-30
feat(oraclecloud): support regionless SDK setup

HugoPBrito:
### Context

Oracle Cloud credentials in Prowler can be valid without using a user-selected scan region. The SDK needs a safe internal region only to bootstrap the OCI client and discover subscribed regions.

### Description

This PR updates the Oracle Cloud provider initialization so API-key authentication can fall back to the internal default region when no explicit scan region filter is provided. It also adds SDK tests for regionless credentials and documents the SDK change in the changelog.

Changed files:

| File | Change |
|------|--------|
| `prowler/CHANGELOG.md` | Adds the SDK changelog entry. |
| `prowler/providers/oraclecloud/oraclecloud_provider.py` | Adds the regionless fallback behavior. |
| `tests/providers/oraclecloud/oraclecloud_provider_test.py` | Adds SDK coverage for regionless OCI setup. |

### Steps to review

1. Review the SDK diff against the platform slice.
2. Confirm the SDK changes match the SDK portion of #11565.
3. Confirm the API slice builds on this PR.

Validation previously run for this slice:

```text
uv run pytest tests/providers/oraclecloud/oraclecloud_provider_test.py
```

### Checklist

<details>

<summary><b>Community Checklist</b></summary>

- [ ] This feature/issue is listed in [here](https://github.com/prowler-cloud/prowler/issues?q=sort%3Aupdated-desc+is%3Aissue+is%3Aopen) or roadmap.prowler.com
- [ ] Is it assigned to me, if not, request it via the issue/feature in [here](https://github.com/prowler-cloud/prowler/issues?q=sort%3Aupdated-desc+is%3Aissue+is%3Aopen) or [Prowler Community Slack](goto.prowler.com/slack)

</details>

- [x] Review if the code is being covered by tests.
- [ ] Review if code is being documented following this specification https://github.com/google/styleguide/blob/gh-pages/pyguide.md#38-comments-and-docstrings
- [ ] Review if backport is needed.
- [ ] Review if is needed to change the [Readme.md](https://github.com/prowler-cloud/prowler/blob/master/README.md)
- [x] Ensure new entries are added to [CHANGELOG.md](https://github.com/prowler-cloud/prowler/blob/master/prowler/CHANGELOG.md), if applicable.

#### SDK/CLI
- Are there new checks included in this PR? No
    - If so, do we need to update permissions for the provider? Please review this carefully.

#### UI
- [ ] All issue/task requirements work as expected on the UI
[...]

PR #11741, opened 2026-06-30
feat(api): support regionless OCI credentials

HugoPBrito:
### Context

OCI provider setup should support credentials without requiring users to provide region filters. Region discovery should be handled by the SDK.

### Description

This PR updates API validation, provider initialization, schema metadata, and migration coverage so OCI secrets can omit region fields. The API no longer introduces `regions` as a replacement field. The existing `region` field is kept only as deprecated legacy compatibility, while new regionless payloads rely on SDK region discovery.

Existing stored `regions` values are removed by the migration because that field is no longer part of the API contract. Existing legacy `region` values are preserved for backward compatibility.

Changed files:

| File | Change |
|------|--------|
| `api/CHANGELOG.md` | Adds the API changelog entry. |
| `api/src/backend/api/migrations/0097_remove_oraclecloud_secret_regions.py` | Removes OCI secret regions from persisted secret data. |
| `api/src/backend/api/tests/test_migrations.py` | Adds migration coverage. |
| `api/src/backend/api/tests/test_serializers.py` | Adds serializer coverage. |
| `api/src/backend/api/tests/test_utils.py` | Adds utility coverage. |
| `api/src/backend/api/tests/test_views.py` | Adds endpoint coverage. |
| `api/src/backend/api/utils.py` | Updates OCI provider initialization behavior. |
| `api/src/backend/api/v1/serializer_utils/providers.py` | Updates OCI credential schema metadata. |
| `api/src/backend/api/v1/serializers.py` | Updates OCI credential validation. |
| `pyproject.toml` | Adds API Vulture ignore entries required by this test slice. |

### Steps to review

1. Review the API diff against the SDK slice.
2. Confirm the API changes match the API portion of #11565.
3. Confirm the UI slice builds on this PR.

Validation previously run for this slice:

```text
uv run --project api pytest ... -k oraclecloud
```

### Checklist

<details>

</details>

#### SDK/CLI

#### UI

#### API

### License

## Chain Context

| Field | Value |
|-------|-------|
| Chain | OCI regionless provider setup split from #11565 |
| Tracker PR | #11565 |
| Position | 3 of 4 |
| Base | `feat/oci-regionless-sdk-chain-11565` |
| Depends on | #11740 |
| Follow-up | #11742 |
| Review budget | 684 / 400 |
| Starts at | SDK regionless OCI provider setup |
| Ends with | API regionless OCI credential support |

### Chain Overview

```text
master
 └── #11739 Platform CI guard
      └── #11740 SDK regionless OCI setup
           └── 📍 #11741 API regionless OCI credentials
                └── #11742 UI regionless OCI credentials
```

### Scope
- Includes: API validation, schema metadata, migration, API tests, API changelog, and API Vulture ignores from #11565.
- Excludes: UI form and payload changes.

### Autonomy
- [x] This PR has one deliverable scope
- [x] This PR can be rolled back without unrelated changes
- [x] Tests, docs, or manual verification cover this unit

## Summary by CodeRabbit

* **New Features**
  * Oracle Cloud API key credentials can be saved without a region filter; subscribed-region discovery runs automatically.
  * Connection checks and provider initialization now support regionless secrets using a default bootstrap region.
* **Bug Fixes**
  * Existing OCI secrets are migrated to remove stored region filters so migrated providers scan all subscribed regions by default.
  * Region subscription discovery failures now raise consistent setup errors.
* **Documentation**
  * Updated OCI getting-started guidance and troubleshooting for subscribed-region retrieval permissions.
* **Chores**
  * E2E workflow adds a preflight required-secrets check (skips for fork contexts, fails otherwise).


api/src/backend/api/v1/serializers.py
```diff
@@ -1739,9 +1739,51 @@ class OracleCloudProviderSecret(serializers.Serializer):
...
+            regions = [region.strip() for region in attrs["regions"]]
+            if any(not region for region in regions):
+                raise serializers.ValidationError(
+                    {"regions": "Regions cannot contain blank values."}
+                )
+            if len(regions) != len(set(regions)):
+                raise serializers.ValidationError(
+                    {"regions": "Regions cannot contain duplicate values."}
+                )
+            attrs["regions"] = regions
```
josema-xyz: I think this cleanup is not saved. `validate_secret_based_on_provider()` validates the serializer, but doesn't use `serializer.validated_data`, so the endpoint can store the region with spaces.

api/src/backend/api/migrations/0097_remove_oraclecloud_secret_regions.py
```diff
@@ -0,0 +1,48 @@
...
+                if isinstance(encrypted_secret, memoryview):
+                    encrypted_secret = encrypted_secret.tobytes()
+                elif isinstance(encrypted_secret, str):
+                    encrypted_secret = encrypted_secret.encode()
+
+                secret = json.loads(fernet.decrypt(encrypted_secret).decode())
+                if "region" not in secret and "regions" not in secret:
+                    continue
+
+                secret.pop("region", None)
```
josema-xyz: Can we confirm this case? If an existing OCI provider had region and cannot list regions, removing region moves it to regionless mode and it may fail.

api/src/backend/api/v1/serializer_utils/providers.py
```diff
@@ -296,14 +296,27 @@
...
                         "description": "The passphrase for the private key, if encrypted.",
                     },
                 },
-                "required": ["user", "fingerprint", "tenancy", "region"],
+                "required": ["user", "fingerprint", "tenancy"],
+                "anyOf": [
+                    {"required": ["key_file"]},
+                    {"required": ["key_content"]},
+                ],
+                "not": {"required": ["region", "regions"]},
```
Davidm4r: Is it mandatory to submit region(s) or can I choose not to submit any?

If it's mandatory use this:

`"oneOf": [
  { "required": ["region"] },
  { "required": ["regions"] }
]`

If we choose to not submit any and is valid, the PR is perfect :)

api/src/backend/api/utils.py
```diff
@@ -290,6 +289,44 @@ def get_prowler_provider_kwargs(
...
+    """Normalize external OCI secret fields into SDK provider kwargs."""
+    prowler_provider_kwargs = secret.copy()
+
+    if "regions" in prowler_provider_kwargs:
+        regions = prowler_provider_kwargs.pop("regions")
+        prowler_provider_kwargs["region"] = set(regions)
+    elif "region" in prowler_provider_kwargs:
+        prowler_provider_kwargs["region"] = {prowler_provider_kwargs["region"]}
+
+    return prowler_provider_kwargs
```
Davidm4r, line 307: What do you think about something like this?
Avoids silent collisions

`def _normalize_oraclecloud_provider_kwargs(secret: dict) -> dict:
    """Normalize external OCI secret fields into SDK provider kwargs."""
    prowler_provider_kwargs = secret.copy()

    if "regions" in prowler_provider_kwargs:
        # Extract the list from the new format
        regions = prowler_provider_kwargs.pop("regions")
        
        # Defensive programming: purge the legacy key if it somehow bypassed the serializer.
        # Using .pop(..., None) ensures it doesn't raise an error if the key doesn't exist.
        prowler_provider_kwargs.pop("region", None)
        
        # Assign the final set required by Prowler
        prowler_provider_kwargs["region"] = set(regions)
        
    elif "region" in prowler_provider_kwargs:
        # If only the legacy format is present, wrap it in a set
        prowler_provider_kwargs["region"] = {prowler_provider_kwargs["region"]}

    return prowler_provider_kwargs`

api/src/backend/api/v1/serializers.py
HugoPBrito: Fixed in `50a6283c5a`.

`validate_secret_based_on_provider()` now returns the nested serializer `validated_data`, and create/update assign it back to `validated_attrs["secret"]` before saving. That means legacy `region` cleanup is persisted instead of only being validated.

I also updated the provider-secret create/update tests to submit `" us-ashburn-1 "` and verify the stored secret keeps `"us-ashburn-1"`.

api/src/backend/api/v1/serializers.py
```diff
@@ -1656,6 +1656,7 @@ def validate_secret_based_on_provider(
                 validation_error.detail[f"secret/{key}"] = value
                 del validation_error.detail[key]
             raise validation_error
+        return serializer.validated_data
```
AdriiiPRodri, line 1675: This returns the nested serializer's `validated_data` for every provider, so the OCI normalization fix becomes a lossy rewrite of all provider secrets. DRF trims every CharField and drops keys that are not declared by the inner serializer. For example, an Image registry password of " secret " is stored as "secret", and the documented GitHub App field `github_app_key` is silently removed because `GithubProviderSecret` only declares `github_app_key_content`. This can break credentials unrelated to OCI on both create and update. Could we preserve the original secret and merge only the normalized OCI region value, or otherwise scope this normalization to OCI?

api/src/backend/api/migrations/0097_remove_oraclecloud_secret_regions.py
```diff
@@ -0,0 +1,47 @@
...
+
+def remove_oraclecloud_secret_regions(apps, schema_editor):
+    Tenant = apps.get_model("api", "Tenant")
+    ProviderSecret = apps.get_model("api", "ProviderSecret")
+    db_alias = schema_editor.connection.alias
+    fernet = Fernet(settings.SECRETS_ENCRYPTION_KEY.encode())
+
+    for tenant in Tenant.objects.using(db_alias).all().iterator():
+        with rls_transaction(str(tenant.id), using=db_alias):
+            provider_secrets = ProviderSecret._base_manager.using(db_alias).filter(
```
AdriiiPRodri: This query is not tenant-scoped when the migration runs in production. Migrations are executed with `--database admin`, and that PostgreSQL admin user bypasses RLS, so `rls_transaction()` does not constrain this queryset. As written, every tenant iteration selects and decrypts every OCI secret; after the first iteration removes regions, subsequent iterations still decrypt the full set before skipping it. This makes the migration roughly **O(tenants * OCI secrets)** and can significantly extend an atomic deployment migration. Let's add an explicit `tenant_id=tenant.id`

AdriiiPRodri (changes requested):

api/src/backend/api/v1/serializers.py
HugoPBrito: I removed support for both OCI `region` and `regions` in provider-secret payloads. Since we no longer support either field, I reverted the previous `validated_data` rewrite so provider-secret validation does not mutate unrelated provider secrets.

api/src/backend/api/migrations/0097_remove_oraclecloud_secret_regions.py
HugoPBrito: I removed the migration-based cleanup entirely, so there is no tenant iteration, RLS transaction, Celery enqueue, or data mutation during migrations anymore. The cleanup now lives in idempotent Celery backfill tasks and must be triggered after deployment.

HugoPBrito:
Hi @AdriiiPRodri @Davidm4r, and @josema-xyz.

I pushed the requested scope cleanup and adjusted the OCI region handling.

The API now keeps `region` only as a legacy input for backwards compatibility. It is accepted if older clients still send it, but it is stripped before storing or using the secret, so OCI credentials are effectively regionless and scans are not filtered by that value. I also removed the extra logging/imports and left the behavior reflected in the schema.

After the scope changes, I cleaned the PR to remove unrelated docs, backfill/task cleanup, and SDK/test noise where possible, and added the API changelog fragment.

api/src/backend/api/v1/serializer_utils/providers.py
HugoPBrito: Since my new changes this does no longer apply. Resolving conversation.

api/src/backend/api/utils.py
HugoPBrito: Since my new changes this does no longer apply. Resolving conversation.

api/src/backend/api/migrations/0097_remove_oraclecloud_secret_regions.py
josema-xyz: Currently addressed.

josema-xyz (dismissed):

ui/lib/provider-credentials/build-credentials.test.ts
alejandrobailo, line 1: This test is not needed.

alejandrobailo (changes requested):
Just remove that test file, the rest is fine. Thanks

Alan-TheGentleman:
The OCI E2E coverage has been updated consistently with the removal of the region field. The tests no longer require `E2E_OCI_REGION`, fill or assert the Region input, or include it in the submitted credentials. Both `PROVIDER-E2E-012` (add OCI provider) and `PROVIDER-E2E-013` (update OCI credentials) passed in the latest CI run.

The overall E2E job is currently failing because of an unrelated AWS test, with additional flaky Kubernetes and Google Workspace tests. One minor cleanup remains: the workflow still injects `E2E_OCI_REGION`, although the OCI tests no longer consume it.

josema-xyz (approved):

ui/lib/provider-credentials/build-credentials.test.ts
HugoPBrito: Removed. Thanks!

alejandrobailo (approved):

AdriiiPRodri (approved):

PR #11748, opened 2026-06-30
fix(api): resolve multi-valued SAML userType claim by highest-privilege role

### LINKED ISSUE #11733: SAML: multi-valued userType claim resolves to the first value, not the highest-privilege role
### Component

API — `api/v1/views.py`, `TenantFinishACSView.dispatch()`

### Description

When an IdP user belongs to multiple groups mapped to different Prowler roles, the SAML assertion carries a **multi-valued** `userType` attribute. The role is resolved with:

```python
role_name = extra.get("userType", [""])[0].strip() if extra.get("userType") else ""
```

Only `userType[0]` is used. The ordering of a multi-valued SAML attribute is IdP-determined and not controllable (e.g. Entra ID emits all assigned app roles in `assignedroles` with no guaranteed order), so a user assigned both, say, `admin` and `user` is resolved non-deterministically — frequently to the **lower-privilege** role — and silently loses admin permissions. There is no precedence rule, and no way to grant multiple roles from SSO.

### Steps to reproduce

1. Configure SAML SSO (e.g. Entra ID) with two app roles mapped to the `userType` claim, e.g. `admin` and `user`.
2. Assign a single user both roles (i.e. they are a member of both mapped groups).
3. Log in via SAML.

### Expected behavior

The user resolves to the most-privileged matching role (or is assigned all matching roles).

### Actual behavior

The user resolves to whichever value the IdP placed first in the `userType` list — often `user` — so an admin lands on the read-only role and cannot manage providers (e.g. a provider "Test connection" returns *you do not have permission to perform this action*).

### Suggested fix

Iterate all `userType` values rather than just `[0]`, and either assign every matching role or select the most-privileged one. At minimum, document that `userType` must be single-valued.

### Version

prowler-api / Prowler App `5.30.0` (self-hosted, SAML SSO).

ThryLox:
### Description
When an IdP user belongs to multiple groups mapped to different Prowler roles, the SAML assertion carries a multi-valued  attribute (e.g. ). Currently,  only resolves the first value (), which is non-deterministic depending on the IdP.

This PR updates the SAML role mapping logic to:
1. Retrieve all values from the  claim.
2. Query all existing roles in the tenant database.
3. Score the roles based on privilege flags (with  getting the highest priority).
4. Assign the user to the single highest-privilege role from the set.
5. Add unit test verification to .

Fixes #11733

## Summary by CodeRabbit

* **New Features**
  * Improved SSO sign-in handling: when SAML provides multiple `userType` role claims, users are now assigned the most privileged matching access level available in their tenant.
* **Bug Fixes**
  * Fixed role assignment during login when multi-valued role information in SAML responses could previously result in an incorrect role being applied.
* **Tests**
  * Added coverage for SAML callbacks to verify multi-valued `userType` resolves to the highest-privilege role.

### DIFF
--- a/api/src/backend/api/tests/test_views.py
+++ b/api/src/backend/api/tests/test_views.py
@@ -14348,6 +14348,86 @@ class TestTenantFinishACSView:
             .exists()
         )
 
+    def test_dispatch_resolves_multi_valued_usertype_to_highest_privilege(
+        self,
+        create_test_user,
+        tenants_fixture,
+        admin_role_fixture,
+        roles_fixture,
+        saml_setup,
+        settings,
+        monkeypatch,
+    ):
+        """Test that multi-valued userType claim resolves to the highest privilege role"""
+        monkeypatch.setenv("SAML_SSO_CALLBACK_URL", "http://localhost/sso-complete")
+        user = create_test_user
+        tenant = tenants_fixture[0]
+
+        admin_role = admin_role_fixture
+        viewer_role = roles_fixture[3]
+
+        social_account = SocialAccount(
+            user=user,
+            provider="saml",
+            extra_data={
+                "firstName": ["John"],
+                "lastName": ["Doe"],
+                "organization": ["testing_company"],
+                "userType": [viewer_role.name, admin_role.name],
+            },
+        )
+
+        request = RequestFactory().get(
+            reverse(
+                "saml_finish_acs", kwargs={"organization_slug": saml_setup["domain"]}
+            )
+        )
+        request.user = user
+        request.session = {}
+
+        with (
+            patch(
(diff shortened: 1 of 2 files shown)


AdriiiPRodri:
Hi @ThryLox,

We really appreciate your interest in the project and the effort you put into this PR.

Unfortunately, we can’t accept it at this time because the current behavior is intentional. If we were to introduce a priority or points system for roles, we’d first need to discuss and agree on the design with the rest of the team.

For now, we’ll update the documentation to clearly explain how userType works so that this behavior is better understood.

We’re sorry we can’t merge these changes, but we truly appreciate your contribution and the time you invested in improving the project.

PR #11735, opened 2026-06-30
fix(github): skip archived repos in CODEOWNERS check

qinfenyinluo:
# PR Description Draft: Prowler #11734

## Summary

Skip archived GitHub repositories in the `repository_has_codeowners_file`
check.

Archived repositories are read-only unless unarchived first, so reporting a
missing CODEOWNERS file for them creates a finding that is not directly
actionable.

## Changes

- Skip repositories where `repo.archived` is true.
- Add a regression test for an archived repository without a CODEOWNERS file.

## Validation

```bash
uv run pytest tests/providers/github/services/repository/repository_has_codeowners_file/repository_has_codeowners_file_test.py
uv run black --check prowler/providers/github/services/repository/repository_has_codeowners_file/repository_has_codeowners_file.py tests/providers/github/services/repository/repository_has_codeowners_file/repository_has_codeowners_file_test.py
```

## AI Assistance Disclosure

Used Codex to inspect the existing check and test structure, draft the
regression test, and summarize validation results. I reviewed the change and
verified the targeted tests locally.

Closes #11734.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the GitHub `repository_has_codeowners_file` check to skip archived (read-only) repositories, preventing them from generating PASS/FAIL results.
  * Added a test to confirm archived repositories without a `CODEOWNERS` file are ignored.
* **Documentation**
  * Updated the unreleased SDK changelog to reflect the change in behavior.

PR #11750, opened 2026-06-30
fix(github): skip archived repos in repository_has_codeowners_file check

### LINKED ISSUE #11734: The repository_has_codeowners_file check currently flags archived repositories as fail
### Issue search

Prowler CLI/SDK

GitHub

### Steps to Reproduce

1. Run `prowler github --check repository_has_codeowners_file`
2. Will also return the repos in finding that are archived

### Expected behavior

When `repository_has_codeowners_file` is executed the findings should not include archived repositories. Check should skip archived repositories.

When `repository_has_codeowners_file` is executed the findings does include archived repositories also.

### Environment Resource

CLI locally

### OS used

MacOS

### Prowler version

5.31.1

### Python version

3.11.9

### Pip version

25.2

### Context

Since archived GitHub repositories are read-only and cannot be updated without first being unarchived, this finding is not actionable.

TorranceTech:
## Description

Fixes #11734

Archived repositories are **read-only** — they cannot be updated without first being unarchived. Flagging them for a missing CODEOWNERS file generates a finding that is not actionable, which adds noise to security reports.

## Changes

**`repository_has_codeowners_file.py`** — skip archived repositories before evaluating the CODEOWNERS check:

```python
if repo.archived:
    continue
```

**`repository_has_codeowners_file_test.py`** — adds `test_archived_repository_is_skipped` to verify that an archived repo with `codeowners_exists=False` produces zero findings.

## Testing

New test case added following the same pattern as the existing test suite. Verifies that `check.execute()` returns an empty list when the only repository is archived.

## Related

This approach is consistent with how other prowler checks treat non-actionable states — the finding is simply not emitted rather than being marked PASS, since the archived state has no bearing on the check's intent.

## Summary by CodeRabbit

* **Bug Fixes**
  * Archived repositories are now skipped when checking for a `CODEOWNERS` file, so they no longer produce pass/fail findings.
* **Tests**
  * Added coverage to verify archived repositories return no results from this check.

### DIFF
--- a/prowler/providers/github/services/repository/repository_has_codeowners_file/repository_has_codeowners_file.py
+++ b/prowler/providers/github/services/repository/repository_has_codeowners_file/repository_has_codeowners_file.py
@@ -22,6 +22,8 @@ class repository_has_codeowners_file(Check):
         """
         findings = []
         for repo in repository_client.repositories.values():
+            if repo.archived:
+                continue
             if repo.codeowners_exists is not None:
                 report = CheckReportGithub(metadata=self.metadata(), resource=repo)
                 if repo.codeowners_exists:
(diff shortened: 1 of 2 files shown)


danibarranqueroo:
Thanks a lot for the fix! 🙏 We're going with #11735 to resolve #11734, as all three PRs implement the same approach. Closing this one as a duplicate but we really appreciate you taking the time to contribute, and hope to see more PRs from you! ❤️

PR #11760, opened 2026-07-01
fix(azure): query flexible server log retention parameter

### LINKED ISSUE #11757: postgresql_flexible_server_log_retention_days_greater_3 always fails - queries Single Server parameter name on Flexible Server
### Issue search

Prowler CLI/SDK

Azure

### Steps to Reproduce

`prowler azure --sp-env-auth --checks postgresql_flexible_server_log_retention_days_greater_3`

### Expected behavior

The check reads `logfiles.retention_days` and returns PASS for values 4–7, FAIL for 1–3.

`postgresql_flexible_server_log_retention_days_greater_3` always fails - queries Single Server parameter name on Flexible Server

### Environment Resource

CLI

### OS used

MacOS

### Prowler version

5.31.1

### Python version

3.11.9

### Pip version

25.2

### Context

The check reads `server.log_retention_days`, which is populated by `_get_log_retention_days` via:

```
log_retention_days = client.configurations.get(
    resource_group_name, server_name, "log_retention_days"
)
```

The configuration name log_retention_days is the Single Server parameter. On Flexible Server that parameter does not exist, the equivalent is `logfiles.retention_days` per Microsoft's docs:
- https://learn.microsoft.com/en-us/azure/postgresql/monitor/how-to-configure-server-logs (CLI command)
- Single Server retirement: https://azure.microsoft.com/updates/azure-database-for-postgresql-single-server-will-be-retired-migrate-to-flexible-server-by-28-march-2025/

qorexdevs:
### Context

Fix #11757

### Description

`postgresql_flexible_server_log_retention_days_greater_3` reads `server.log_retention_days`, which `_get_log_retention_days` populates from the `log_retention_days` configuration parameter. That name only exists on the retired Single Server. On Flexible Server the equivalent parameter is `logfiles.retention_days` (per Microsoft's docs), so the lookup always misses and the check reports every server as FAIL regardless of the real retention value.

This switches the query to `logfiles.retention_days`.

### Steps to review

Check the parameter name in `_get_log_retention_days` and the added test that asserts `configurations.get` is called with `logfiles.retention_days`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected the Azure PostgreSQL log retention check so it reads the proper retention setting for Flexible Server deployments.
  * This fixes incorrect `FAIL` results caused by using an outdated setting name tied to the retired Single Server offering.
  * Updated the release notes to reflect the fix.


danibarranqueroo:
Thanks @qorexdevs, solid fix and nice changelog entry! 🙌 It's identical to #11761, which we're going with since it also adds a regression test. We've carried your changelog wording over there. Closing in favor of #11761, really appreciate the help! 🚀

PR #11870, opened 2026-07-07
perf(api): harden compliance-overview ingest & partition table

pedrooot:
### Description

On every scan we delete the scan's existing rows and re-insert the whole set, streaming them into `compliance_requirements_overviews` with back-to-back `COPY` statements that take 13–15s each. On a plain, ever-growing table that means constant commit/fsync waits, lock contention, and a lot of autovacuum work chasing the dead tuples the delete-then-reinsert leaves behind. Under load it's enough to tip the writer over.

## What changed

**Ingest (`tasks/jobs/scan.py`).** The COPY batch size is now configurable via `DJANGO_COMPLIANCE_COPY_BATCH_SIZE` (default dropped from 10000 to 5000), with an optional pause between batches via `DJANGO_COMPLIANCE_COPY_THROTTLE_SECONDS` (default 0, so behaviour is unchanged unless someone turns it up). Smaller batches give the writer room to checkpoint and vacuum between flushes, and ops can tune both without a deploy. We also reuse a single admin connection across all batches instead of reconnecting per batch — the tenant GUC is re-applied inside each batch's transaction since `SET_CONFIG_QUERY` is `is_local=TRUE`. And on a scan's first run we skip the pre-insert `DELETE` entirely (there's nothing to delete), so we don't open an empty write transaction or generate dead tuples for no reason.

**Partitioning.** `compliance_requirements_overviews` is now a partitioned table, RANGE by its UUIDv7 `id`, the same way `findings` works. That lets old scans' rows age out with `DROP PARTITION` instead of `DELETE` + CASCADE, which is where most of the autovacuum churn came from. The row generator emits `uuid7()` ids now so new rows land in the right monthly partition (uuid4 would always fall into `default`). The old 5-column unique constraint is gone — once the partition key `id` gets added to it, it doesn't mean anything, and per-scan dedup is already handled by the delete-then-reinsert, exactly like `findings` does.

**Migration `0098`.** Converts the table in place, without copying any data. It's a `SeparateDatabaseAndState`: the state side mirrors the model changes, and the database side renames the existing table and re-attaches it as the `default` partition of a new partitioned parent (`ATTACH` is metadata-only, no row rewrite). The parent FKs are added while it's still empty and the existing index is attached as-is, so nothing gets re-scanned or re-indexed. RLS is re-established on both the parent and the default partition through the project's `RowLevelSecurityConstraint`.

### Steps to review

### Checklist

<details>

</details>

#### SDK/CLI

#### UI

#### API

### License

## Summary by CodeRabbit

* **New Features**
  * Compliance requirement overviews now use faster, range-partitioned storage with more scalable IDs.
  * Added separate configuration options to tune how these records are partitioned and retained (partition months, partition count, and optional max-age/indefinite retention).

* **Bug Fixes**
  * Improved compliance overview ingestion to reduce memory pressure by using batched COPY with optional throttling and safer re-run behavior that avoids unnecessary empty deletes.
  * Made persistence more resilient by falling back to ORM bulk insert for any failed batch while keeping successful batches committed.

### DIFF
--- a/api/CHANGELOG.md
+++ b/api/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 All notable changes to the **Prowler API** are documented in this file.
 
+## [1.35.0] (Prowler UNRELEASED)
+
+### 🐞 Fixed
+
+- Partitioned `compliance_requirements_overviews` and hardened its COPY ingest (batched, single reused connection, first-run DELETE skipped) to stop the database writer from running out of memory during compliance overview ingestion [(#XXXXX)](https://github.com/prowler-cloud/prowler/pull/XXXXX)
+
+---
+
 ## [1.34.0] (Prowler v5.33.0)
 
 ### 🚀 Added
--- a/api/CHANGELOG.md
+++ b/api/CHANGELOG.md
@@ -2,11 +2,11 @@
 
 All notable changes to the **Prowler API** are documented in this file.
 
-## [1.35.0] (Prowler UNRELEASED)
+## [1.34.1] (Prowler UNRELEASED)
 
 ### 🐞 Fixed
 
-- Partitioned `compliance_requirements_overviews` and hardened its COPY ingest (batched, single reused connection, first-run DELETE skipped) to stop the database writer from running out of memory during compliance overview ingestion [(#XXXXX)](https://github.com/prowler-cloud/prowler/pull/XXXXX)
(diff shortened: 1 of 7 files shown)


pedrooot:
closed by https://github.com/prowler-cloud/prowler/pull/11875

PR #11920, opened 2026-07-09
refactor(ui): rename reserved billing flag to CLOUD_BILLING_ENABLED

pfe-nazaries:
### Context

The runtime-config island (#11500) reserved a billing enable flag as `BILLING_CLOUD_ENABLE` (boolean, no consumer yet). The Prowler Cloud product will consume it as a 3-value install-level selector, not a boolean. This aligns the reserved placeholder to its final name and type **before any consumer exists**, so it is a no-op rename with zero user impact.

### Description

- Rename the env var `BILLING_CLOUD_ENABLE` → `CLOUD_BILLING_ENABLED` and the island key `billingCloudEnable` → `cloudBillingEnabled`.
- Type `CLOUD_BILLING_ENABLED` as the billing selector `"legacy" | "metronome" | "false"` (was `"true" | "false"`).
- The client island still receives a **derived boolean** (`value !== "false"`); the raw selector is read server-side only (for V1/V2 routing on the consumer side).
- Add a boot-time fail-fast: when `CLOUD_BILLING_ENABLED` is `metronome` but `UI_POSTHOG_ENABLE` is not `"true"`, throw — `metronome` billing routes per tenant via the `BILLING_SYSTEM_METRONOME` PostHog flag, so PostHog must be enabled.

No runtime consumer exists yet, so there is no behavior change for any current deployment.

### Steps to review

- `cd ui && pnpm run typecheck`
- `cd ui && pnpm vitest run lib/env.test.ts lib/integrations.test.ts lib/get-runtime-config.client.test.ts lib/runtime-env.test.ts` (52 tests, incl. 3 new boot-check cases)
- Confirm no old references remain: `rg "BILLING_CLOUD_ENABLE\b|billingCloudEnable"` → empty

### Checklist

- [x] No new checks; no provider permission changes.
- [ ] Screenshots not applicable (no UI surface; reserved config key, no consumer).

### License

### DIFF
--- a/ui/lib/env.test.ts
+++ b/ui/lib/env.test.ts
@@ -62,6 +62,7 @@ describe("lib/env gated integration validation", () => {
     "POSTHOG_KEY",
     "UI_POSTHOG_HOST",
     "POSTHOG_HOST",
+    "CLOUD_BILLING_ENABLED",
   ] as const;
 
   beforeEach(() => {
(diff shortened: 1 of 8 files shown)


alejandrobailo (changes requested):
I found one blocking issue with the PostHog compatibility path.

alejandrobailo (approved):
The legacy PostHog path is not part of the metronome rollout. We will set UI_POSTHOG_ENABLE for the deployments using it, so this is not blocking.

PR #11928, opened 2026-07-09
fix(api): harden Lighthouse provider base URLs

AdriiiPRodri:
### Context

This PR hardens Lighthouse provider configuration handling for OpenAI-compatible endpoints.

### Description

- Validates OpenAI-compatible base URLs before storing provider configuration and before outbound provider checks
- Rejects non-public destinations after hostname resolution
- Prevents provider discovery and connection checks from following URL redirects
- Pins validated connection addresses for OpenAI-compatible provider checks
- Sanitizes OpenAI-compatible provider errors returned in task results
- Adds API and worker regression coverage for unsafe provider URLs
- Adds an API changelog fragment
- Rejects malformed or out-of-range explicit ports, including port `0`
- Detects private IPv4 destinations embedded in IPv6 addresses, including IPv4-mapped, 6to4, and NAT64 addresses
- Disables environment-derived proxy configuration for the restricted OpenAI-compatible HTTP client

### Mitigation approach

The configured `base_url` is validated in both layers that can affect the connection flow. The API validates it before persistence, and the worker validates it again immediately before using it for provider discovery or connection checks.

Validation requires HTTPS, blocks known local or metadata hostnames, checks literal IP addresses, resolves hostnames before outbound use, and rejects any resolved address that is not public. The OpenAI-compatible HTTP client also validates the destination during connection setup and connects to the validated address while preserving the original hostname for Host and TLS SNI. Redirect following is disabled for provider discovery and connection checks.

Task results no longer expose response bodies returned by OpenAI-compatible endpoints. Authentication failures and other provider-side failures are returned as generic messages suitable for users without reflecting remote content.

The restricted HTTP client ignores environment proxy variables, preventing proxy configuration from bypassing the validated and pinned connection path.

### Steps to review

- `uv run ruff check src/backend/config/django/base.py src/backend/api/validators.py src/backend/api/v1/serializers.py src/backend/tasks/jobs/lighthouse_providers.py src/backend/api/tests/test_validators.py src/backend/api/tests/test_views.py src/backend/tasks/tests/test_tasks.py`
- `uv run pytest src/backend/api/tests/test_validators.py src/backend/api/tests/test_views.py::TestLighthouseProviderConfigViewSet src/backend/tasks/tests/test_tasks.py::TestCheckLighthouseProviderConnectionTask src/backend/tasks/tests/test_tasks.py::TestRefreshLighthouseProviderModelsTask -q`
- Verified locally with `make dev-launch` that restricted Lighthouse provider base URLs are rejected before a connection check is created
- Verified locally that the connection-check worker path blocks restricted destinations before provider discovery

### Checklist

#### API
- [x] Check if version updates are required (e.g., specs, uv, etc.).

### License

## Summary by CodeRabbit

* **New Features**
  * Added stricter OpenAI-compatible Lighthouse `base_url` validation (HTTPS-only, required hostname, default port 443) with blocking of localhost and metadata-targeting addresses, including safer IPv4/IPv6 handling.
  * Provider connection and model refresh for OpenAI-compatible Lighthouse now use restricted networking with redirect following disabled.

* **Bug Fixes**
  * Create/update now return clear, field-specific `base_url` validation errors.
  * Connection/refresh failures now use sanitized error messages and deactivate the provider without exposing remote response details.

* **Tests**
  * Expanded coverage for validation rules (DNS on/off, IPv4/IPv6 cases) and task behaviors (rejection, redirects, error masking, resolved public connectivity).

### DIFF
--- a/api/src/backend/api/validators.py
+++ b/api/src/backend/api/validators.py
@@ -1,10 +1,8 @@
 import ipaddress
 import socket
 import string
-from collections.abc import Iterable
 from urllib.parse import urlparse
 
-from django.conf import settings
 from django.core.exceptions import ValidationError
 from django.utils.translation import gettext as _
 
@@ -24,21 +22,6 @@ def _normalize_hostname(hostname: str) -> str:
     return hostname.rstrip(".").lower()
 
 
-def _configured_lighthouse_allowed_hosts() -> tuple[str, ...]:
-    allowed_hosts = getattr(settings, "LIGHTHOUSE_OPENAI_COMPATIBLE_ALLOWED_HOSTS", ())
-    return tuple(host for host in allowed_hosts if host)
-
-
-def _is_allowed_lighthouse_host(hostname: str, allowed_hosts: Iterable[str]) -> bool:
-    normalized_allowed_hosts = [
-        _normalize_hostname(host.lstrip(".")) for host in allowed_hosts
-    ]
-    return any(
-        hostname == allowed_host or hostname.endswith(f".{allowed_host}")
-        for allowed_host in normalized_allowed_hosts
-    )
-
-
 def _validate_lighthouse_public_ip(address: str) -> None:
     ip_address = ipaddress.ip_address(address)
     if not ip_address.is_global:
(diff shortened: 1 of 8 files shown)


api/src/backend/api/validators.py
```diff
@@ -1,14 +1,123 @@
...
+    return hostname.rstrip(".").lower()
+
+
+def _validate_lighthouse_public_ip(address: str) -> None:
+    ip_address = ipaddress.ip_address(address)
+    if not ip_address.is_global:
+        raise ValidationError(
+            _("Base URL must use an external public endpoint."),
+            code="lighthouse_base_url_not_public",
+        )
```
jfagoagas, line 42: **Point 1: harden the public-IP check against IPv6-embedded IPv4 (metadata reach on some configs).**

`is_global` on the raw address has two gaps:

- **IPv4-mapped** (`::ffff:169.254.169.254`) is only delegated to the embedded IPv4 in CPython 3.11.10 / 3.12.5+. The prod image (`python:3.12.13`) is fine, but `requires-python = ">=3.11,<3.13"` allows older 3.11.x, where `::ffff:169.254.169.254` reports `is_global=True` and reaches IMDS.
- **NAT64** (`64:ff9b::a9fe:a9fe`) reports `is_global=True` on every version. In an environment with a NAT64 gateway it translates to `169.254.169.254`.

Reduce IPv6 forms that embed an IPv4 to that IPv4 before classifying, so the check no longer depends on interpreter patch level:

```suggestion
def _validate_lighthouse_public_ip(address: str) -> None:
    ip_address = ipaddress.ip_address(address)
    # Some IPv6 forms embed an IPv4 address that routes to the mapped target.
    # Reduce them to that IPv4 before classifying. `is_global` alone is
    # version-dependent for IPv4-mapped (only fixed in CPython 3.11.10 / 3.12.5+)
    # and never flags NAT64 (64:ff9b::/96); both can reach link-local / metadata
    # targets, e.g. https://[::ffff:169.254.169.254] or https://[64:ff9b::a9fe:a9fe].
    if isinstance(ip_address, ipaddress.IPv6Address):
        embedded = ip_address.ipv4_mapped or ip_address.sixtofour
        if embedded is None and ip_address in ipaddress.ip_network("64:ff9b::/96"):
            embedded = ipaddress.IPv4Address(int(ip_address) & 0xFFFFFFFF)
        if embedded is not None:
            ip_address = embedded
    if not ip_address.is_global:
        raise ValidationError(
            _("Base URL must use an external public endpoint."),
            code="lighthouse_base_url_not_public",
        )
```

Optional: hoist `ipaddress.ip_network("64:ff9b::/96")` to a module constant. Teredo (`2001::/32`) also embeds an IPv4, but it's XOR-obfuscated and rarely a direct internal route, so skip it unless your egress supports it.

josema-xyz: Fixed!

josema-xyz (approved):

deliciousbob:
Hi Prowler Team, what is the reason for allowing only public AI Providers ?
-> I've tried to configure Lighthouse AI with our on-premise OpenAI-API-compatible LiteLLM
 -> https, port tcp:443, domain + valid cert  (_https://llm-api.companydomain.com/v1 or without v1_)
 but got the message: **_"Base URL must use an external public endpoint"_**

Did i do something wrong or is there a different way to use an on-prem AI Solution ?
Thank you! Best reagards, Robert

PR #12001, opened 2026-07-15
feat(ui): add Jira dispatch choices for finding selections

HugoPBrito:
### Context

This PR adds UI support for choosing how selected Findings and Finding Groups are sent to Jira while preserving the Cloud-only boundary for grouped and batch Jira actions.

### Description

- Add Jira action choices for selected Findings, selected Finding Groups, and mixed Group/Finding selections.
- Show grouped-vs-separate Jira issue creation options when multiple individual Findings are selected.
- Keep single-Finding Jira dispatch available while disabling grouped and batch Jira actions outside Cloud-enabled UI paths.
- Align Finding Group Jira modal copy, selection labels, disabled Cloud badge behavior, and local development sign-in behavior.

### Steps to review

1. Review selected Finding and Finding Group actions in the Findings table.
2. Verify single individual Finding Jira dispatch remains available.
3. Verify multiple selected Findings show the grouped-vs-separate Jira issue creation modal before dispatch.
4. Verify Finding Group and mixed Group/Finding selections show accurate labels and dispatch choices.
5. Verify non-Cloud UI paths show grouped/batch Jira actions disabled with the Cloud-only tooltip.
6. Run or inspect the test coverage listed below:
   - `pnpm vitest run --project unit actions/integrations/jira-dispatch.test.ts components/findings/floating-mute-button.test.tsx components/findings/send-to-jira-modal.test.tsx components/findings/send-to-jira-modal-copy.test.ts components/findings/table/column-finding-resources.test.tsx components/findings/table/data-table-row-actions.test.tsx components/findings/table/findings-group-drill-down.test.ts components/findings/table/findings-group-table.test.tsx lib/deployment.test.ts lib/shared/env.test.ts`
   - `pnpm typecheck`
   - Pre-commit UI hooks: Prettier, ESLint, TypeScript, Unit Tests

### Checklist

<details>

</details>

#### SDK/CLI

#### UI
- [x] If this PR adds or updates npm dependencies, include package-health evidence (maintenance, popularity, known vulnerabilities, license, release age) and explain why existing/native alternatives are insufficient.

#### API

#### MCP Server

### License

## Summary by CodeRabbit

* **New Features**
  * Added “Send to Jira” for finding groups, individual findings, and mixed selections with grouped/individual dispatch options and improved toast feedback, plus a Prowler Cloud-only chooser with a disabled tooltip/badge when unavailable.
  * Enhanced Findings filtering with human-readable “Finding Group” titles/chips and deep-link expansion via check-based filters.
  * Updated findings/resources tables to show “Affected failing resource” instead of “Status”.
* **Bug Fixes**
  * Improved Jira dispatch polling handling for partial success and no-result/no-op completions.
* **Tests**
  * Expanded UI and unit tests for Jira dispatch flows, filter/chip rendering, and disabled-tooltip behavior.


HugoPBrito:
Follow-up cleanup to keep in mind after this lands:

- Replace the loose `SendToJiraModal` prop set with a single typed Jira selection object.
- Centralize Jira action labels and Cloud-only gating so table, drill-down, row actions, and modal cannot drift.
- Rename `FloatingMuteButton` toward a generic selection-actions component now that it owns more than mute.

I’d keep this out of this PR to avoid mixing behavior fixes with a broader UI refactor, but happy to do it here if required.

alejandrobailo (changes requested):
I am requesting changes before approval. I found two correctness issues that can produce empty results or duplicate Jira issues, plus several architecture and design-system regressions that should be addressed in this PR.

## Blocking correctness issues

1. **Use the list filter for multi-select Finding Groups.** The new Finding Group filter is a multi-select, but it sends `filter[check_id]=a,b`. The API defines `check_id` as an exact filter and `check_id__in` as the list filter, so selecting more than one group can return no findings. Please use `check_id__in` for multi-select requests and keep the exact key only where it is needed for legacy or deep-link compatibility.

   [UI filter definition](https://github.com/prowler-cloud/prowler/blob/07d591db644fd2a97f1fce2f862cc38aa5a27b31/ui/components/findings/findings-filters.utils.ts#L120-L150) · [API filter contract](https://github.com/prowler-cloud/prowler/blob/07d591db644fd2a97f1fce2f862cc38aa5a27b31/api/src/backend/api/filters.py#L1022-L1043)

2. **Represent partial Jira dispatch success explicitly.** If one batch succeeds and a later batch fails, the modal reports the whole operation as failed and leaves the selection available for retry. Retrying can create duplicate Jira issues for the targets that already succeeded. Please preserve the batch-to-target mapping, report both successful and failed batches, and make retries include only failed targets. The tests should cover this outcome instead of treating it as a total failure.

   [Dispatch handling](https://github.com/prowler-cloud/prowler/blob/07d591db644fd2a97f1fce2f862cc38aa5a27b31/ui/components/findings/send-to-jira-modal.tsx#L228-L304) · [Current partial-failure test](https://github.com/prowler-cloud/prowler/blob/07d591db644fd2a97f1fce2f862cc38aa5a27b31/ui/components/findings/send-to-jira-modal.test.tsx#L556-L613)

## Required architecture and implementation follow-ups

3. **Do not self-filter or truncate the Finding Group options.** The options request receives all resolved filters, including the active `check_id` filter. After selecting group A, the endpoint can return only A, which prevents adding group B. The request is also capped at 100 results, so valid groups can disappear. Please exclude the facet own `check_id` and `check_id__in` filters and load the complete option set through pagination, metadata, or a dedicated searchable endpoint.

   [Options request](https://github.com/prowler-cloud/prowler/blob/07d591db644fd2a97f1fce2f862cc38aa5a27b31/ui/app/%28prowler%29/findings/page.tsx#L73-L85)

4. **Reuse the shared task watcher for Jira tasks.** The modal reimplements the polling loop already provided by the task-watcher store and therefore loses task tracking and completion feedback after a hard reload. Please register or extend a Jira task kind in the shared watcher and reuse its persistence, cross-tab synchronization, resume behavior, timeout handling, and polling implementation.

   [Duplicated modal polling](https://github.com/prowler-cloud/prowler/blob/07d591db644fd2a97f1fce2f862cc38aa5a27b31/ui/components/findings/send-to-jira-modal.tsx#L79-L94) · [Shared polling](https://github.com/prowler-cloud/prowler/blob/07d591db644fd2a97f1fce2f862cc38aa5a27b31/ui/store/task-watcher/store.ts#L125-L155) · [Shared persistence and resume flow](https://github.com/prowler-cloud/prowler/blob/07d591db644fd2a97f1fce2f862cc38aa5a27b31/ui/store/task-watcher/store.ts#L190-L255)

5. **Make the modal contract type-safe and centralize shared types.** The optional `targetIds`, `targetType`, `targetBatches`, and mode flags allow invalid half-configured states to compile. Please model the supported single-target, target-list, and batched modes as a discriminated union. Also move the duplicated Jira target and filter constants and types into the shared integrations type module and use the exported values at call sites. `JIRA_DISPATCH_MODE` and `JiraDispatchMode` are exported correctly; the remaining related definitions should follow the same pattern.

   [Modal contract](https://github.com/prowler-cloud/prowler/blob/07d591db644fd2a97f1fce2f862cc38aa5a27b31/ui/components/findings/send-to-jira-modal.tsx#L38-L65) · [Duplicated dispatch definitions](https://github.com/prowler-cloud/prowler/blob/07d591db644fd2a97f1fce2f862cc38aa5a27b31/ui/actions/integrations/jira-dispatch.ts#L14-L33)

6. **Use the design-system API instead of ad hoc visual classes.** The new callers pass layout and visual classes directly to shared `Button` and `RadioGroup` controls, and the modal builds a custom radio-card pattern even though `CustomRadio` already exists. Please reuse or extend the shared component, or add a sanctioned variant/API when the existing component does not cover the required state. The same applies to the floating button layout instead of styling the design-system button at the call site.

   [Floating button classes](https://github.com/prowler-cloud/prowler/blob/07d591db644fd2a97f1fce2f862cc38aa5a27b31/ui/components/findings/floating-mute-button.tsx#L159-L196) · [Custom radio cards](https://github.com/prowler-cloud/prowler/blob/07d591db644fd2a97f1fce2f862cc38aa5a27b31/ui/components/findings/send-to-jira-modal.tsx#L545-L591) · [Existing shared radio component](https://github.com/prowler-cloud/prowler/blob/07d591db644fd2a97f1fce2f862cc38aa5a27b31/ui/components/shadcn/custom/custom-radio.tsx#L12-L24)

7. **Avoid regressing the existing single-action flow.** `FloatingMuteButton` now always opens a chooser. On resource details, where Jira is not available, this adds an unnecessary second click to choose the only action. When Jira is available in the drill-down, the outer trigger still says `Mute (N)`, which no longer describes all possible actions. Please keep direct mute behavior when there is no second action and use a neutral selection-action label when Jira is available.

   [Chooser behavior](https://github.com/prowler-cloud/prowler/blob/07d591db644fd2a97f1fce2f862cc38aa5a27b31/ui/components/findings/floating-mute-button.tsx#L152-L223) · [Resource details caller](https://github.com/prowler-cloud/prowler/blob/07d591db644fd2a97f1fce2f862cc38aa5a27b31/ui/components/resources/table/resource-detail-content.tsx#L410-L418) · [Drill-down caller](https://github.com/prowler-cloud/prowler/blob/07d591db644fd2a97f1fce2f862cc38aa5a27b31/ui/components/findings/table/findings-group-drill-down.tsx#L243-L261)

8. **Remove the new prop-to-state synchronization effects.** `FindingsGroupTable` mirrors one requested group into five local state values and tracks the previous value through a ref. This is fragile, produces a hook dependency warning, and conflicts with the established derived-state and event-driven architecture. Please derive what can be derived, use keyed state reset or explicit interaction handlers, and update or clear the URL from those handlers. The modal default mode should likewise be established during open/reset initialization rather than synchronized from a prop in an effect.

   [Table synchronization effect](https://github.com/prowler-cloud/prowler/blob/07d591db644fd2a97f1fce2f862cc38aa5a27b31/ui/components/findings/table/findings-group-table.tsx#L103-L148) · [Modal synchronization effect](https://github.com/prowler-cloud/prowler/blob/07d591db644fd2a97f1fce2f862cc38aa5a27b31/ui/components/findings/send-to-jira-modal.tsx#L222-L226)

Please also centralize the cloud-only tooltip constant instead of exporting it from a feature component, and remove or wire the new deployment helpers that currently have no production consumers. These are shared policy concerns and should not be owned by a findings component.

I verified the current head commit `07d591db644fd2a97f1fce2f862cc38aa5a27b31`. Type checking passes, and the focused test suite passes 138 tests across 11 files. Targeted linting has no errors, but it reports the hook dependency warnings described above. I did not find a new standalone React component file or new raw color tokens; the main design-system concern is the custom styling and duplicated radio-card implementation at the call sites.

alejandrobailo:
Thanks for iterating on this. I re-reviewed commit `1acf9bf4175a8c91ebaef72676f6b0ff31d53b30`. The Finding Group option pagination/self-filtering and the single-action Mute flow are now fixed, and moving the Cloud-only tooltip and Jira target type to shared modules is a good improvement.

I still do not consider all requested changes resolved:

- The Finding Group control now writes `check_id__in`, but legacy/deep links using `filter[check_id]` are not normalized. The control reads only `check_id__in`, so the existing exact value appears unselected. Selecting another group keeps the old exact filter and adds the new `__in` filter, which can produce an empty result. Please migrate or replace `check_id` when this control is edited/applied and add a behavioral test for this flow.
- Jira partial success is only partially handled. The code does not preserve the task-to-batch/target mapping, does not identify the failed targets, and does not provide a retry containing only those targets. Retrying the current selection can still duplicate issues that were already created. A single task containing both created and failed results is also still reported as a total error because `pollJiraDispatchTask` discards the partial result.
- Jira dispatch still uses the modal-local 15-round polling loop. Please register a Jira task kind and reuse/extend the shared task watcher so tracking survives reloads and uses the existing persistence, cross-tab, resume, and timeout behavior.
- The modal props are not actually a discriminated union. There is no discriminator, and `SendToJiraBatchProps` still allows `targetIds`, `targetType`, and `targetBatches` as independent optionals, so incomplete and contradictory configurations compile. Please model single, list, and batch as mutually exclusive shapes. Please also use the exported Jira target constants at production call sites instead of raw `"check_id"` / `"finding_id"` strings.
- The design-system issue remains: shared `Button` and `RadioGroup` controls still receive ad hoc visual/layout classes, and the modal still implements custom radio cards instead of reusing or extending the shared radio component.
- The prop/search-param synchronization effects remain. `FindingsGroupTable` still mirrors the requested group into several local states through `useEffect` plus a ref, and the modal still synchronizes `defaultDispatchMode` through another effect. Stabilizing `safeData` removes the lint warning but does not address the architecture issue. Please derive state, use keyed resets, or update it in explicit interaction/open handlers.
- `isOnPremiseDeployment`, `isBillingEnabled`, and `isPostHogEnabled` still have no production consumers in this repository. Please remove them or wire them to the intended production integration.

I verified this exact commit locally: TypeScript passes, targeted ESLint passes without warnings, and 118 relevant tests across 11 files pass. The current `ui-tests` CI failure is in an unrelated Lighthouse test and passes when run in isolation, so I do not attribute that failure to this change.

alejandrobailo (changes requested):
Thanks a lot for iterating on this and for addressing several of the points from the previous review. The latest changes are a clear improvement: the Finding Group deep-link conflict is fixed, the option pagination/self-filtering issue is resolved, the single-finding Mute flow works again, and the shared exports and constants are now used consistently. I really appreciate the work you have put into this.

I am requesting changes because there are still a few items we should resolve before merging:

- The UI test suite is currently failing on two stale source assertions. One still expects the previous pagination condition, and the other expects the raw `targetType="finding_id"` string instead of `JIRA_DISPATCH_TARGET.FINDING_ID`. These tests should be updated, preferably to verify behavior rather than exact implementation text where possible. The E2E job is also failing, although the available logs no longer show the underlying cause, so it should be rerun or confirmed.
- Partial Jira success is now communicated better, but retrying is still unsafe. We only retain task IDs and aggregate error messages, without mapping failed tasks back to their findings or batches. Retrying the same selection can therefore create duplicate Jira issues for targets that already succeeded. We need failed-target information from the backend contract or a dispatch strategy that lets us retry only failed targets. Also, the current failure-message helper returns `result.error` before including the partial-success counts.
- The modal still implements its own polling, timeout, and task lifecycle instead of reusing the shared task watcher. This means Jira tasks do not benefit from the existing persistence and resume behavior.
- There are still design-system issues: custom layout/shadow classes are passed to Button, and the dispatch selector manually recreates a radio-card pattern even though `CustomRadio` already exists and can be reused or extended.
- The prop-to-state synchronization effects remain in the findings group table, and `defaultDispatchMode` is still synchronized into the form through an effect. These should be modeled as derived/keyed state or initialized/reset at the relevant event boundary.
- The modal props are more type-safe now, but the union is not explicitly discriminated and plain `string[]` values still allow empty arrays. Empty arrays can silently fall back to `findingId`. The current workaround also duplicates almost the same `SendToJiraModal` JSX in the group table, so I think a discriminated target object or equivalent API would make the call sites safer and remove the duplication.

I do not want these remaining points to take more of your time, so I will take ownership of them and address them soon. Thank you again for the thoughtful iteration and for improving the implementation substantially.

alejandrobailo (approved):

PR #12006, opened 2026-07-15
fix(api): scope writer transactions in the scan report task

cesararroba:
### Context

`set_tenant` wraps its whole task in `@transaction.atomic`, because it sets the RLS tenant variable with `set_config(..., TRUE)` — transaction-scoped, so the tenant context only exists inside a transaction. Applied to `generate_outputs_task`, that makes the writer transaction live as long as the task itself.

`generate_outputs_task` runs for a long time: an early `Provider.objects.get` takes `ACCESS SHARE` on `providers` and holds it until commit. After the SecurityHub integration check the task never touches the writer connection again — remaining queries go to the read replica, then it renders reports and uploads to S3. The writer connection sits idle in transaction for the rest of the run, holding a lock for work happening entirely elsewhere. Any DDL needing that table then waits, and Postgres queues later readers behind it.

### Description

Drops `@set_tenant(keep_tenant=True)` from `generate_outputs_task` and scopes each writer access to its own short `rls_transaction(tenant_id)` — the pattern already established elsewhere in this file. `Provider` is now fetched with `select_related("secret")` so provider initialization, which authenticates over the network, runs outside the transaction instead of holding the `providers` lock across it. The writer transaction now lives for the length of a single query rather than the length of the task. RLS `set_config(..., TRUE)` semantics are unchanged.

**Latent bug fixed along the way.** `s3_integrations` was built as a lazy queryset inside the replica transaction but evaluated by `if s3_integrations:` outside it. That only worked because `@set_tenant` happened to be holding tenant context for the whole task; removing the decorator would have silently skipped every S3 integration. It is now materialized with `list()` inside the transaction block.

**Review note on test coverage.** The existing suite is structurally blind to this bug class: it connects as `prowler_admin`, a superuser, and superusers bypass RLS even with `FORCE ROW LEVEL SECURITY`; and under plain `@pytest.mark.django_db`, pytest-django's outer atomic block demotes `rls_transaction` to a savepoint, so the tenant variable binds to the outer transaction and leaks for the rest of the test. The 82 existing mocked tests passing says little about RLS correctness on its own. The new `test_generate_outputs_rls.py` is the actual signal: it uses `SET ROLE test` on a fresh connection plus `transaction=True` to get genuine enforcement, and pins that reads are denied without tenant context, allowed inside it, that the `select_related` secret survives the transaction, and that the lazy-queryset trap is real.

**Not covered.** No end-to-end run of the task under enforced RLS — that needs real findings, provider credentials, and S3. Read-replica routing itself is not exercised locally either, since the replica alias resolves to `None` in test settings.

### Steps to review

- `api/src/backend/tasks/tasks.py`: the transaction scoping change in `generate_outputs_task`.
- `api/src/backend/tasks/tests/test_tasks.py`: existing mocked coverage updated for the new `select_related` call shape.
- `api/src/backend/tasks/tests/test_generate_outputs_rls.py`: new, RLS-enforced coverage — read this one first, it is what actually validates the change.

### Checklist

<details>

</details>

#### SDK/CLI

#### UI
- Not applicable.

#### API
- [x] Ensure a changelog fragment is added under [api/changelog.d/](https://github.com/prowler-cloud/prowler/tree/master/api/changelog.d), if applicable.

Test results:
- `pytest src/backend/tasks/tests/test_tasks.py` → 82 passed (82 at baseline, no regressions).
- `pytest src/backend/tasks/tests/test_generate_outputs_rls.py` → 5 passed.
- `ruff check` and `ruff format --check` clean.

#### MCP Server
- Not applicable.

### License

## Summary by CodeRabbit

* **Bug Fixes**
  * Scan report generation no longer holds database transactions open during long rendering/upload operations, minimizing interference with schema changes.
  * Enhanced tenant-scoped handling to avoid late/delayed RLS resolution issues and ensure updates apply with correct scoping.

* **Tests**
  * Added RLS scoping and transaction-boundary tests covering tenant isolation, cached related data, and deferred queryset behavior.
  * Updated task tests to align with the new provider lookup approach.

* **Documentation**
  * Refined guidelines for tenant-scoped Celery task execution, including rules for long-running operations and RLS safety.

### DIFF
--- a/api/src/backend/tasks/tasks.py
+++ b/api/src/backend/tasks/tasks.py
@@ -738,7 +738,6 @@ class ScanReportRLSTask(RLSTask):
     name="scan-report",
     queue="scan-reports",
 )
-@set_tenant(keep_tenant=True)
 @handle_provider_deletion
 def generate_outputs_task(scan_id: str, provider_id: str, tenant_id: str):
     """
@@ -750,6 +749,11 @@ def generate_outputs_task(scan_id: str, provider_id: str, tenant_id: str):
     batch of transformed findings, and uses a flag to indicate when the final
     batch is being processed. Finally, the output files are compressed and
     uploaded to S3.
+
+    Rendering and uploading take tens of minutes, so every database access is
+    scoped to its own short `rls_transaction`. A task-wide transaction would hold
+    ACCESS SHARE on the tables it read for the whole run, and any DDL waiting on
+    one of those tables would queue every later reader behind itself.
 
     Args:
         tenant_id (str): The tenant identifier.
@@ -769,15 +773,21 @@ def generate_outputs_task(scan_id: str, provider_id: str, tenant_id: str):
             error,
         )
 
-    # Check if the scan has findings
-    if not ScanSummary.objects.filter(scan_id=scan_id).exists():
-        logger.info(f"No findings found for scan {scan_id}")
-        return {"upload": False}
+    with rls_transaction(tenant_id):
+        # Check if the scan has findings
+        if not ScanSummary.objects.filter(scan_id=scan_id).exists():
+            logger.info(f"No findings found for scan {scan_id}")
+            return {"upload": False}
 
-    provider_obj = Provider.objects.get(id=provider_id)
+        # `secret` is selected eagerly because initialize_prowler_provider reads it
+        # but runs outside this transaction, where RLS would resolve it to nothing.
+        provider_obj = Provider.objects.select_related("secret").get(id=provider_id)
+        provider_uid = provider_obj.uid
+        provider_type = provider_obj.provider
+
+    # Kept outside the transaction: provider initialization authenticates against
+    # the cloud provider, and that network call must not hold the providers lock.
     prowler_provider = initialize_prowler_provider(provider_obj)
-    provider_uid = provider_obj.uid
-    provider_type = provider_obj.provider
 
     # Per-framework exporters in `COMPLIANCE_CLASS_MAP` consume the legacy bulk.
     frameworks_bulk = Compliance.get_bulk(provider_type)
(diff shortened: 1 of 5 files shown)


api/src/backend/tasks/tasks.py
```diff
@@ -751,6 +750,11 @@ def generate_outputs_task(scan_id: str, provider_id: str, tenant_id: str):
     batch is being processed. Finally, the output files are compressed and
     uploaded to S3.
 
+    Rendering and uploading take tens of minutes, so every database access is
+    scoped to its own short `rls_transaction`. A task-wide transaction would hold
+    ACCESS SHARE on the tables it read for the whole run, and any DDL waiting on
+    one of those tables would queue every later reader behind itself.
```
josema-xyz: Could we say “every writer access” here? The findings loop still keeps one read-replica transaction open while reports are rendered, so “every database access” is broader than the actual change.

api/src/backend/tasks/tests/test_generate_outputs_rls.py
```diff
@@ -0,0 +1,158 @@
...
+        integration=integration,
+        provider=provider,
+    )
+    return str(tenant.id), str(provider.id)
+
+
+@pytest.mark.django_db(transaction=True)
+class TestGenerateOutputsTenantScoping:
+    """
+    Pins the assumptions that let `generate_outputs_task` drop `@set_tenant`: every
```
josema-xyz, line 74: These tests check the RLS behavior, but they never run generate_outputs_task. Could we add one task-level test to catch @set_tenant being added again or a query running outside rls_transaction?

api/src/backend/tasks/tasks.py
cesararroba: You're right, the wording was inaccurate. Updated the docstring to say "every writer access" and to state explicitly that the findings loop and rendering still run inside one long read-replica transaction. That's deliberate: it keeps a consistent snapshot across batches, and locks taken on the replica do not block writers on the primary.

api/src/backend/tasks/tests/test_generate_outputs_rls.py
cesararroba: Agree; added `TestGenerateOutputsTransactionScope`, which runs the real task and probes `connection.in_atomic_block` at the slow phases, asserting no transaction is held during provider authentication, compression, and upload, and that the final scan `output_location` update IS inside one. Verified it fails both ways: re-adding `@set_tenant` trips the slow-phase probes, and replacing the final `rls_transaction` trips the write probe.

Limits worth being upfront about: the probes are at three fixed points rather than every statement, it runs as a superuser so it asserts transaction scope rather than tenant isolation (the existing tests in this file cover RLS via the unprivileged `test` role), and `READ_REPLICA_ALIAS` is `None` under test settings so the replica split is not genuinely exercised.

api/src/backend/tasks/tasks.py
```diff
@@ -818,19 +828,22 @@ def get_writer(writer_map, name, factory, is_last):
...
-        security_hub_integrations = Integration.objects.filter(
-            integrationproviderrelationship__provider_id=provider_id,
-            integration_type=Integration.IntegrationChoices.AWS_SECURITY_HUB,
-            enabled=True,
+    with rls_transaction(tenant_id):
+        # `_transform_findings_stats` walks the queryset and follows its `scan`
+        # relation, so it has to run with the tenant context live.
+        scan_summary = FindingOutput._transform_findings_stats(
+            ScanSummary.objects.filter(scan_id=scan_id)
         )
```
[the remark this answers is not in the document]
cesararroba: Applied, with a correction: this is not N+1. `.scan` is read only off the first row, and the loop body touches only local columns, so the cost is constant at three queries, `__getitem__` chains a new queryset without populating the original's `_result_cache`, so indexing and iterating each fire their own query, plus one for the relation. Materializing with `select_related("scan")` collapses it to one. Efficiency and type-correctness only; it sits inside the transaction, so there was no correctness bug.

cesararroba:
Re the outside-diff-range finding requesting the `@set_tenant` decorator be restored on `generate_outputs_task` (`api/src/backend/tasks/tasks.py`, around line 742): politely declining.

Removing that decorator is the purpose of this PR. `@set_tenant` wraps the whole task in `transaction.atomic`, and this task keeps running for tens of minutes after it is done with the writer. Keeping it would hold ACCESS SHARE on every table it read until the task ends, and any DDL waiting on one of those tables would then queue every later reader behind it.

Each writer access is now scoped to its own short `rls_transaction(tenant_id)`, which is the same skill file's rule, "ALWAYS use `rls_transaction(tenant_id)` when querying outside ViewSet context", applied strictly. `skills/prowler-api/SKILL.md` has been updated to document this exception, and the new regression test added in this branch fails if the decorator is re-added.

PR #11848, opened 2026-07-06
fix(api): queue provider scans when one is active

AdriiiPRodri:
### Context

Manual and scheduled scans for the same provider could be dispatched concurrently when another scan was already pending, started, or executing. This ports the scan queueing behavior from the cloud implementation to OSS, adjusted for the OSS task flow and without cloud-only provider availability logic.

### Description

This PR adds database-backed queueing for provider scans in the API:

- Manual scan creation now records a queued task when the provider already has an active or queued scan.
- Scheduled scans are coalesced into a single queued scheduled run while preserving the next scheduled scan placeholder.
- Completed or failed scans dispatch the next queued provider scan.
- Resource failed-finding counter updates now use stable row locking and retry on database conflicts.
- Tests cover manual queueing, scheduled queueing/coalescing, queued dispatch, cleanup behavior, and resource counter retry handling.

### Steps to review

1. Run focused tests:
   ```bash
   cd api
   uv run pytest src/backend/api/tests/test_views.py::TestScanViewSet src/backend/tasks/tests/test_tasks.py::TestCleanupOrphanScheduledScans src/backend/tasks/tests/test_tasks.py::TestPerformScheduledScanTask src/backend/tasks/tests/test_tasks.py::TestPerformScanTask src/backend/tasks/tests/test_scan.py::TestPerformScan --create-db -q
   ```
2. Run lint/checks:
   ```bash
   cd api
   uv run ruff check src/backend/api/v1/views.py src/backend/tasks/tasks.py src/backend/tasks/jobs/scan.py src/backend/tasks/utils.py src/backend/api/tests/test_views.py src/backend/tasks/tests/test_tasks.py src/backend/tasks/tests/test_scan.py
   cd ..
   git diff --check
   ```
3. Start the dev stack:
   ```bash
   make dev
   ```
4. Verify the API responds:
   ```bash
   curl -i http://localhost:8080/api/v1/
   ```
5. Create a manual scan for a provider that already has an active scan and verify the new task is stored as `QUEUED` instead of dispatching immediately.

Validated locally:

- `make dev` starts successfully.
- `GET /api/v1/` returns `200 OK`.
- Manual scan creation against a provider with an active scan returned `202` and stored the new task with status `QUEUED`.
- Focused impacted tests: `108 passed`.
- `ruff check`: passed.
- `git diff --check`: passed.

### Checklist

<details>

<summary><b>Community Checklist</b></summary>

[...]

Issue #6893, opened 2025-02-11
Stop a running scan


Mubashir-ahmed:
### New feature motivation

Currently if I launched a scan on prowler app mistakenly, then it doesnt give me option to pause it or cancel it. We should have both options. At the moment we only get the option to rename scan name.

### Solution Proposed

A pause/play button to stop and run scan manually.
A cancel button to remove scan from the scan job immediately.

### Describe alternatives you've considered

Alternatively it should not allow me to run a scan when there is already a running instance on the same cloud resource. It should give me a prompt that say something like "A scan is already running on the resource: XYZ"

puchy22:
Hi @Mubashir-ahmed,

I find the feature you mentioned very useful and interesting. However, there are currently no plans to include anything similar in our roadmap. I've shared it with the team so they can take it into account for future improvements, but I don’t think it will be feasible in the short term.

Best regards, and thanks for your efforts to make Prowler better!

paminhoff:
I would love to have this feature.

puchy22:
Hi @Mubashir-ahmed and @paminhoff,

The team has heard that this is a highly requested feature from the community, so there’s a plan to add it through the task module. As we make progress, I’ll keep you updated with new PRs or GitHub issues so you can easily track this feature request.

Thanks for using Prowler and for bringing great ideas to the table! 🚀

CC: @everythings-gonna-be-alright and @ryguycraig

rakatar:
> Hi [@Mubashir-ahmed](https://github.com/Mubashir-ahmed) and [@paminhoff](https://github.com/paminhoff),
> 
> The team has heard that this is a highly requested feature from the community, so there’s a plan to add it through the task module. As we make progress, I’ll keep you updated with new PRs or GitHub issues so you can easily track this feature request.
> 
> Thanks for using Prowler and for bringing great ideas to the table! 🚀
> 
> CC: [@everythings-gonna-be-alright](https://github.com/everythings-gonna-be-alright) and [@ryguycraig](https://github.com/ryguycraig)

Thank you! and yes, this is indeed a very useful feature.

dexpert:
I also want to add that if scan is stuck or for some reason docker or server reboot the scan will keep running & will not compelete making the app not usable becasue scan keep "running" doing noting . Is there any way to kill the "not running" scan ? 
Had to delete the aws provider so the scan will be removed also !

paminhoff:
Any news on scheduling scans feature?

[...]

PR #12008, opened 2026-07-15
fix(api): recover scans stranded in executing by a dead worker

### LINKED ISSUE #12007: Kick off a new Scan of a provider - Stuck in Queued status
### Issue search

Prowler UI

Not applicable, AWS

### Steps to Reproduce

Manually kicking off a scan of a configured Cloud Provider (AWS based),
Job is added to the queue, but no matter how long is waited, the scan never starts.
Kicking off other jobs will sometimes start, but the job in the queue never starts.

Image

### Expected behavior

The scan should start, scan, and finish.

Scan stays in Queued status (see screenshot above).

### Environment Resource

running on docker locally

### OS used

Rocky Linux

### Prowler version

5.22.0

### Python version

docker

### Pip version

docker

### Context

_No response_

Subhro-ai:
### Context

Fix #12007

Since #11848 (v5.33.0) scans run one-at-a-time per provider. When a scan is launched while another is active for the same provider, the new scan is created in the `available` state (shown as **Queued** in the UI) and is *not* dispatched to Celery — it waits until the active scan's task finishes and its `finally` block dispatches the next queued scan.

`scan-perform` uses `acks_late=False` and is deliberately excluded from orphan recovery, so a worker that dies mid-scan (OOM, hard time-limit kill, node failure) never runs that `finally`. The scan is stranded in `executing` (the scenario already documented in the troubleshooting guide / #6893) and, because it still counts as the provider's active scan, **every queued scan for that provider stays "Queued" forever** — which is exactly what the reporter saw ("other jobs sometimes start, but the job in the queue never starts").

### Description

Adds a periodic watchdog task `scan-cleanup-stale-scans` that recovers scans stranded in `executing` by a dead worker, mirroring the existing `cleanup_stale_attack_paths_scans` for the main `Scan` model:

- Reads all `executing` scans (cross-tenant, admin DB) and pings each recorded worker once.
  - **worker alive** → only reaped if it overran the stale ceiling (`SCAN_STALE_THRESHOLD_MINUTES`, default 48h, matched to the long hard time-limit);
  - **worker gone** → reaped after the shorter inactivity window (`SCAN_INACTIVITY_THRESHOLD_MINUTES`, default 30m of no progress heartbeat);
  - **no worker recorded** → time-based fallback on `started_at`;
  - unreachable control bus → every worker looks alive, so nothing is reaped by mistake.
- For each stranded scan: locks and re-verifies the row, marks it `failed` (with `completed_at`/`duration`), marks the `TaskResult` terminal, and revokes the lost task after commit.
- **Then dispatches the next queued scan for that provider**, so the provider's queue drains.

It never re-runs the dead scan (that would duplicate findings) — it only fails it and releases the queue. The task is registered as a `PeriodicTask` (every 15 min) and added to `_SKIP_RECOVERY` so the orphan watchdog never re-enqueues it.

New settings (env-overridable, no `.env` changes required — same convention as the Attack Paths thresholds):
- `SCAN_INACTIVITY_THRESHOLD_MINUTES` (default 30)
- `SCAN_STALE_THRESHOLD_MINUTES` (default 2880)

Files:
- `tasks/jobs/scan_cleanup.py` — new reaper + queue-drain
- `tasks/tasks.py` — `cleanup_stale_scans_task` (`queue="celery"`)
- `tasks/jobs/orphan_recovery.py` — skip the new maintenance task
- `config/django/base.py` — thresholds
- `api/migrations/0098_scan_cleanup_periodic_task.py` — PeriodicTask (15 min)
- `tasks/tests/test_scan_cleanup.py` — unit tests
- `tasks/tests/test_orphan_recovery.py` — cover skip of the new task
- `docs/troubleshooting.mdx` — note automatic recovery now exists
- `api/changelog.d/scan-stuck-queued.fixed.md`

### Steps to review

1. Read `tasks/jobs/scan_cleanup.py`. Confirm the detection matrix (alive/dead/no-worker) and that queue drain runs only after the FAILED state commits.
2. `cd api && uv run pytest src/backend/tasks/tests/test_scan_cleanup.py src/backend/tasks/tests/test_orphan_recovery.py -q`
3. Confirm the migration applies and creates the `scan-cleanup-stale-scans` PeriodicTask: `uv run python src/backend/manage.py migrate`.
4. End-to-end: start a scan, kill the worker mid-scan (`docker kill` the worker) so the scan is left in `executing`, queue another scan for the same provider (it goes to `Queued`), then run the task (`uv run python src/backend/manage.py shell -c "from tasks.tasks import cleanup_stale_scans_task; print(cleanup_stale_scans_task())"`) and confirm the stuck scan flips to `failed` and the queued scan starts.

### Checklist

- [x] Tests added (`test_scan_cleanup.py`) and orphan-recovery skip covered.
- [x] Changelog fragment under `api/changelog.d/`.
- [ ] Backport: consider backporting to the 5.33.x/5.34.x lines where the per-provider queuing shipped.

## Summary by CodeRabbit

* **Bug Fixes**
  * Automatically recovers scans stuck in “executing” after a worker crash by marking them as failed.
  * Unblocks provider queues so the next queued scan can run.
  * Preserves scans when workers appear active or their status is unknown.

* **Maintenance**
  * Runs stale-scan cleanup automatically every 15 minutes.
  * Improves performance when checking active scans.

* **Documentation**
  * Updates troubleshooting guidance with recovery timing and configurable thresholds.

* **Tests**
  * Adds coverage for cleanup, queue recovery, worker checks, and race conditions.

### DIFF
--- a/api/src/backend/config/django/base.py
+++ b/api/src/backend/config/django/base.py
@@ -315,6 +315,17 @@ ATTACK_PATHS_SCAN_STALE_THRESHOLD_MINUTES = env.int(
     "ATTACK_PATHS_SCAN_STALE_THRESHOLD_MINUTES", 2880
 )  # 48h
 
+# Stale scan cleanup. A scan whose worker dies mid-run is stranded in
+# `executing`; because scans run one-at-a-time per provider (#11848), that also
+# blocks every queued scan for the provider. The periodic `scan-cleanup-stale-scans`
+# task fails such scans and drains the queue. A scan on a live worker is only
+# reaped after the stale ceiling (matched to the long hard time-limit); a scan
+# whose worker is gone is reaped after the shorter inactivity window.
+SCAN_INACTIVITY_THRESHOLD_MINUTES = env.int("SCAN_INACTIVITY_THRESHOLD_MINUTES", 30)
+SCAN_STALE_THRESHOLD_MINUTES = env.int(
+    "SCAN_STALE_THRESHOLD_MINUTES", 2880
+)  # 48h
+
 # Selects where the persistent attack-paths graph is stored. The scan
 # temporary database is always Neo4j; only the sink is configurable.
 # Valid values: "neo4j" (default, OSS and local dev), "neptune" (hosted).
--- a/api/src/backend/config/django/base.py
+++ b/api/src/backend/config/django/base.py
@@ -342,6 +342,10 @@ elif SCAN_STALE_THRESHOLD_MINUTES < SCAN_INACTIVITY_THRESHOLD_MINUTES:
     )
 
 SCAN_CLEANUP_ENABLED = _scan_cleanup_config_error is None
+# Exported so the task can log the reason itself. Settings are imported before
+# Django applies `LOGGING`, so the message below is emitted by the logging
+# last-resort handler and does not reach the configured structured handlers.
+SCAN_CLEANUP_CONFIG_ERROR = _scan_cleanup_config_error
 
 if _scan_cleanup_config_error:
     logging.getLogger(__name__).error(
--- a/api/src/backend/config/django/base.py
+++ b/api/src/backend/config/django/base.py
@@ -1,3 +1,4 @@
+import logging
 from datetime import timedelta
 
 from config.custom_logging import LOGGING  # noqa
@@ -315,16 +316,52 @@ ATTACK_PATHS_SCAN_STALE_THRESHOLD_MINUTES = env.int(
     "ATTACK_PATHS_SCAN_STALE_THRESHOLD_MINUTES", 2880
 )  # 48h
 
-# Stale scan cleanup. A scan whose worker dies mid-run is stranded in
-# `executing`; because scans run one-at-a-time per provider (#11848), that also
-# blocks every queued scan for the provider. The periodic `scan-cleanup-stale-scans`
-# task fails such scans and drains the queue. A scan on a live worker is only
-# reaped after the stale ceiling (matched to the long hard time-limit); a scan
-# whose worker is gone is reaped after the shorter inactivity window.
+# Stale scan cleanup. A scan whose worker dies mid-run is stranded; because scans
+# run one-at-a-time per provider (#11848), that also blocks every queued scan for
+# the provider. The periodic `scan-cleanup-stale-scans` task fails such scans and
+# drains the queue. A scan whose worker is alive, or whose liveness could not be
+# established, is only reaped past the stale ceiling; a scan whose worker is
+# confirmed gone is reaped after the shorter inactivity window.
 SCAN_INACTIVITY_THRESHOLD_MINUTES = env.int("SCAN_INACTIVITY_THRESHOLD_MINUTES", 30)
-SCAN_STALE_THRESHOLD_MINUTES = env.int(
-    "SCAN_STALE_THRESHOLD_MINUTES", 2880
-)  # 48h
+SCAN_STALE_THRESHOLD_MINUTES = env.int("SCAN_STALE_THRESHOLD_MINUTES", 2880)  # 48h
+
+# Failing a scan is destructive, so an unsafe threshold policy disables the
+# watchdog instead of silently widening what it may reap. Startup is deliberately
+# NOT aborted: a config typo must not take down the API, worker and Beat.
+_scan_cleanup_config_error = None
+if SCAN_INACTIVITY_THRESHOLD_MINUTES <= 0 or SCAN_STALE_THRESHOLD_MINUTES <= 0:
+    _scan_cleanup_config_error = (
+        "SCAN_INACTIVITY_THRESHOLD_MINUTES and SCAN_STALE_THRESHOLD_MINUTES must "
+        "both be positive"
+    )
+elif SCAN_STALE_THRESHOLD_MINUTES < SCAN_INACTIVITY_THRESHOLD_MINUTES:
+    _scan_cleanup_config_error = (
+        "SCAN_STALE_THRESHOLD_MINUTES must be greater than or equal to "
+        "SCAN_INACTIVITY_THRESHOLD_MINUTES, otherwise a scan on a live worker "
+        "would be reaped sooner than one whose worker is confirmed gone"
+    )
+
+SCAN_CLEANUP_ENABLED = _scan_cleanup_config_error is None
+
+if _scan_cleanup_config_error:
+    logging.getLogger(__name__).error(
+        "Stale scan cleanup is DISABLED - invalid configuration: %s",
+        _scan_cleanup_config_error,
+    )
+else:
+    # The stale ceiling must not undercut how long `scan-perform` is allowed to
+    # run, or a legitimately long scan on a live worker could be reaped early.
+    _scan_hard_limit_minutes = (
+        env.int("DJANGO_CELERY_LONG_TASK_TIME_LIMIT", 48 * 60 * 60) // 60
+    )
+    if SCAN_STALE_THRESHOLD_MINUTES < _scan_hard_limit_minutes:
+        logging.getLogger(__name__).warning(
+            "SCAN_STALE_THRESHOLD_MINUTES (%s) is below the scan hard time limit "
+            "(%s minutes); a long-running scan on a live worker may be reaped "
+            "before Celery stops it",
+            SCAN_STALE_THRESHOLD_MINUTES,
+            _scan_hard_limit_minutes,
+        )
 
 # Selects where the persistent attack-paths graph is stored. The scan
 # temporary database is always Neo4j; only the sink is configurable.
--- a/api/src/backend/tasks/jobs/scan_cleanup.py
+++ b/api/src/backend/tasks/jobs/scan_cleanup.py
@@ -1,0 +1,223 @@
+"""Detect and fail stale ``Scan`` rows, then drain the provider queue.
+
+Since #11848 scans run one-at-a-time per provider: launching a scan while
+another is active for the same provider leaves the new scan ``QUEUED`` (state
+``available``, shown as "Queued" in the UI) until the active scan's Celery task
+finishes and its ``finally`` block dispatches the next queued scan.
+
+``scan-perform`` uses ``acks_late=False`` and is deliberately excluded from
+orphan recovery, so a worker that dies mid-scan (OOM, hard time-limit kill, node
+failure) never runs that ``finally``. The scan is stranded in ``executing`` and,
+because ``_get_dispatched_provider_scan`` still counts it as active, every queued
+scan for that provider stays "Queued" forever (issue #12007).
+
+This periodic task reaps such stranded ``executing`` scans - confirmed dead by
+pinging the recorded worker - marks them ``failed``, revokes the lost task, and
+dispatches the next queued scan so the provider's queue drains. It never re-runs
+the dead scan (that would duplicate findings); it only fails it and releases the
+queue. It mirrors ``cleanup_stale_attack_paths_scans`` for the main ``Scan``
+model.
+"""
+
+from datetime import UTC, datetime, timedelta
+from functools import partial
+
+from api.db_router import MainRouter
+from api.db_utils import rls_transaction
+from api.models import Scan, StateChoices
+from celery import states
+from celery.utils.log import get_task_logger
+from config.django.base import (
+    SCAN_INACTIVITY_THRESHOLD_MINUTES,
+    SCAN_STALE_THRESHOLD_MINUTES,
+)
+from django.db import DatabaseError
+from django.db.transaction import on_commit
+from tasks.jobs.orphan_recovery import is_worker_alive
+from tasks.jobs.orphan_recovery import revoke_task as _revoke_task
+
+logger = get_task_logger(__name__)
+
+
+def cleanup_stale_scans() -> dict:
+    """Fail ``executing`` scans whose worker is gone and drain the provider queue.
+
+    Detection per scan:
+    - worker recorded and alive: reap only if ``started_at`` is past the stale
+      ceiling (matched to the long hard time-limit), i.e. the task overran and
+      the worker will be killed anyway;
+    - worker recorded and gone: reap once the scan has been silent (no
+      ``updated_at`` heartbeat) for longer than the inactivity window;
+    - no worker recorded: fall back to the stale ceiling on ``started_at``.
+
+    An unreachable control bus makes every worker look alive, so a still-running
+    scan is never failed by mistake.
+
+    Returns a summary dict with the count and ids of the scans cleaned up.
+    """
+    now = datetime.now(tz=UTC)
+    stale_cutoff = now - timedelta(minutes=SCAN_STALE_THRESHOLD_MINUTES)
+    inactivity_cutoff = now - timedelta(minutes=SCAN_INACTIVITY_THRESHOLD_MINUTES)
+
+    executing_scans = list(
+        Scan.all_objects.using(MainRouter.admin_db)
+        .filter(state=StateChoices.EXECUTING)
+        .select_related("task__task_runner_task")
+    )
+
+    workers = {
+        tr.worker
+        for scan in executing_scans
+        if (tr := _task_result_for(scan)) and tr.worker
+    }
+    # Ping each distinct worker once; errors are treated as "alive" upstream.
+    worker_alive = {worker: is_worker_alive(worker) for worker in workers}
+
+    cleaned_up: list[str] = []
+    for scan in executing_scans:
+        task_result = _task_result_for(scan)
+        worker = task_result.worker if task_result else None
+        reason, recheck_inactivity_cutoff = _stale_reason(
+            scan, worker, worker_alive, stale_cutoff, inactivity_cutoff
+        )
+        if reason is None:
+            continue
+
+        if _fail_stale_scan(
+            scan,
+            task_result,
+            reason,
+            revoke=worker is not None,
+            inactivity_cutoff=recheck_inactivity_cutoff,
+        ):
+            cleaned_up.append(str(scan.id))
+            # Release the provider queue the dead scan was blocking. Runs after
+            # the FAILED state is committed so the next dispatch no longer sees
+            # this scan as the active one.
+            _drain_provider_queue(str(scan.tenant_id), str(scan.provider_id))
+
+    logger.info(f"Stale scan cleanup: {len(cleaned_up)} scan(s) cleaned up")
+    return {"cleaned_up_count": len(cleaned_up), "scan_ids": cleaned_up}
+
+
+def _task_result_for(scan):
+    """Return the scan's ``TaskResult`` row, or ``None`` if it has no task."""
+    return getattr(scan.task, "task_runner_task", None) if scan.task else None
+
+
+def _stale_reason(
+    scan,
+    worker,
+    worker_alive: dict,
+    stale_cutoff: datetime,
+    inactivity_cutoff: datetime,
+) -> tuple[str | None, datetime | None]:
+    """Classify one ``executing`` scan.
+
+    Returns ``(reason, recheck_inactivity_cutoff)``. ``reason`` is ``None`` when
+    the scan must be preserved. ``recheck_inactivity_cutoff`` is set only for the
+    dead-worker case so the row can be re-checked for late activity under lock.
+    """
+    if worker:
+        if worker_alive.get(worker, True):
+            if scan.started_at is None or scan.started_at >= stale_cutoff:
+                return None, None
+            return (
+                "Scan exceeded stale threshold - cleaned up by periodic task",
+                None,
+            )
+
+        # Worker is gone: give a scan that heartbeated recently one more cycle,
+        # in case the worker is being replaced.
+        if scan.updated_at >= inactivity_cutoff:
+            logger.info(
+                f"Preserving scan {scan.id}: worker {worker} is gone but activity "
+                f"is recent (progress={scan.progress}, updated_at={scan.updated_at})"
+            )
+            return None, None
+        return (
+            f"Worker unresponsive and scan inactive for "
+            f"{SCAN_INACTIVITY_THRESHOLD_MINUTES} minutes - cleaned up by periodic task",
+            inactivity_cutoff,
+        )
+
+    # No worker recorded: time-based heuristic only.
+    if scan.started_at is None or scan.started_at >= stale_cutoff:
+        return None, None
+    return (
+        "No worker recorded, scan exceeded stale threshold - cleaned up by periodic task",
+        None,
+    )
+
+
+def _fail_stale_scan(
+    scan,
+    task_result,
+    reason: str,
+    *,
+    revoke: bool = False,
+    inactivity_cutoff: datetime | None = None,
+) -> bool:
+    """Atomically lock the row, re-verify eligibility, and mark it ``FAILED``.
+
+    Registers task revocation after commit when requested. Returns ``True`` if
+    the scan was failed, ``False`` if it moved on or vanished.
+    """
+    scan_id = str(scan.id)
+    try:
+        with rls_transaction(str(scan.tenant_id)):
+            try:
+                fresh_scan = Scan.objects.select_for_update().get(id=scan.id)
+            except Scan.DoesNotExist:
+                logger.warning(f"Scan {scan_id} no longer exists, skipping")
+                return False
+
+            if fresh_scan.state != StateChoices.EXECUTING:
+                logger.info(f"Scan {scan_id} is now {fresh_scan.state}, skipping")
+                return False
+
+            if (
+                inactivity_cutoff is not None
+                and fresh_scan.updated_at >= inactivity_cutoff
+            ):
+                logger.info(
+                    f"Scan {scan_id} received activity during worker checks, skipping"
+                )
+                return False
+
+            now = datetime.now(tz=UTC)
+            fresh_scan.state = StateChoices.FAILED
+            fresh_scan.completed_at = now
+            update_fields = ["state", "completed_at", "updated_at"]
+            if fresh_scan.started_at is not None:
+                fresh_scan.duration = int(
+                    (now - fresh_scan.started_at).total_seconds()
+                )
+                update_fields.append("duration")
+            fresh_scan.save(update_fields=update_fields)
+
+            if task_result:
+                task_result.status = states.FAILURE
+                task_result.date_done = now
+                task_result.save(update_fields=["status", "date_done"])
+
+            if revoke and task_result:
+                on_commit(
+                    partial(_revoke_task, task_result, terminate=True),
+                    using=fresh_scan._state.db,
+                )
+    except DatabaseError:
+        logger.exception(f"Failed to mark stale scan {scan_id} as failed")
+        return False
+
+    logger.info(f"Cleaned up stale scan {scan_id}: {reason}")
+    return True
+
+
+def _drain_provider_queue(tenant_id: str, provider_id: str) -> None:
+    """Dispatch the next queued scan for the provider (best effort)."""
+    # Imported lazily: ``tasks.tasks`` imports this module, so a top-level import
+    # would be circular.
+    from tasks.tasks import _dispatch_next_queued_provider_scan_best_effort
+
+    _dispatch_next_queued_provider_scan_best_effort(tenant_id, provider_id)
--- a/api/src/backend/tasks/jobs/scan_cleanup.py
+++ b/api/src/backend/tasks/jobs/scan_cleanup.py
@@ -7,16 +7,25 @@ finishes and its ``finally`` block dispatches the next queued scan.
 
 ``scan-perform`` uses ``acks_late=False`` and is deliberately excluded from
 orphan recovery, so a worker that dies mid-scan (OOM, hard time-limit kill, node
-failure) never runs that ``finally``. The scan is stranded in ``executing`` and,
-because ``_get_dispatched_provider_scan`` still counts it as active, every queued
-scan for that provider stays "Queued" forever (issue #12007).
+failure) never runs that ``finally``. The scan is stranded and, because
+``_get_dispatched_provider_scan`` still counts it as active, every queued scan
+for that provider stays "Queued" forever (issue #12007).
 
-This periodic task reaps such stranded ``executing`` scans - confirmed dead by
-pinging the recorded worker - marks them ``failed``, revokes the lost task, and
-dispatches the next queued scan so the provider's queue drains. It never re-runs
-the dead scan (that would duplicate findings); it only fails it and releases the
-queue. It mirrors ``cleanup_stale_attack_paths_scans`` for the main ``Scan``
-model.
+This periodic task recovers from that in two complementary passes:
+
+1. **Reap** scans a dead worker stranded. A scan blocks the provider queue when
+   it is ``executing`` OR ``available``/``scheduled`` with a dispatched task
+   (a task can be lost after dispatch but before it reaches ``executing``).
+   Liveness is confirmed by pinging the recorded worker with bounded retries;
+   only a scan whose worker is *confirmed* gone (or an unrecorded worker past
+   the stale ceiling) is failed. It never re-runs the scan (that would duplicate
+   findings) — it only fails it and releases the queue.
+2. **Drain** every provider that still has a ``QUEUED`` scan but no active scan.
+   This retries a queue release that a transient error dropped on a previous run
+   (the reaped scan is already terminal, so a reap-only retry would never fire),
+   making the recovery self-healing across runs.
+
+It mirrors ``cleanup_stale_attack_paths_scans`` for the main ``Scan`` model.
 """
 
 from datetime import UTC, datetime, timedelta
@@ -25,60 +34,56 @@ from functools import partial
 from api.db_router import MainRouter
 from api.db_utils import rls_transaction
 from api.models import Scan, StateChoices
-from celery import states
+from celery import current_app, states
 from celery.utils.log import get_task_logger
 from config.django.base import (
     SCAN_INACTIVITY_THRESHOLD_MINUTES,
     SCAN_STALE_THRESHOLD_MINUTES,
 )
 from django.db import DatabaseError
+from django.db.models import Q
 from django.db.transaction import on_commit
-from tasks.jobs.orphan_recovery import is_worker_alive
 from tasks.jobs.orphan_recovery import revoke_task as _revoke_task
 
 logger = get_task_logger(__name__)
 
+# Worker-liveness ping tuning (mirrors the Attack Paths cleanup added in #11986):
+# a single short ping can misjudge a busy worker as dead, so ping with bounded
+# retries and exponential backoff, and treat a control-bus failure as "unknown".
+WORKER_PING_BASE_TIMEOUT_SECONDS = 5
+WORKER_PING_MAX_ATTEMPTS = 3
+
 
 def cleanup_stale_scans() -> dict:
-    """Fail ``executing`` scans whose worker is gone and drain the provider queue.
+    """Fail scans stranded by a dead worker and drain blocked provider queues.
 
-    Detection per scan:
-    - worker recorded and alive: reap only if ``started_at`` is past the stale
-      ceiling (matched to the long hard time-limit), i.e. the task overran and
-      the worker will be killed anyway;
-    - worker recorded and gone: reap once the scan has been silent (no
-      ``updated_at`` heartbeat) for longer than the inactivity window;
-    - no worker recorded: fall back to the stale ceiling on ``started_at``.
-
-    An unreachable control bus makes every worker look alive, so a still-running
-    scan is never failed by mistake.
-
-    Returns a summary dict with the count and ids of the scans cleaned up.
+    Returns a summary dict: how many scans were reaped (with their ids) and how
+    many provider queues were checked for a pending release.
     """
     now = datetime.now(tz=UTC)
     stale_cutoff = now - timedelta(minutes=SCAN_STALE_THRESHOLD_MINUTES)
     inactivity_cutoff = now - timedelta(minutes=SCAN_INACTIVITY_THRESHOLD_MINUTES)
 
-    executing_scans = list(
-        Scan.all_objects.using(MainRouter.admin_db)
-        .filter(state=StateChoices.EXECUTING)
-        .select_related("task__task_runner_task")
-    )
+    candidates = _stale_scan_candidates()
 
     workers = {
         tr.worker
-        for scan in executing_scans
+        for scan in candidates
         if (tr := _task_result_for(scan)) and tr.worker
     }
-    # Ping each distinct worker once; errors are treated as "alive" upstream.
-    worker_alive = {worker: is_worker_alive(worker) for worker in workers}
+    responsive_workers, unresponsive_workers = _ping_workers(workers)
 
     cleaned_up: list[str] = []
-    for scan in executing_scans:
+    for scan in candidates:
         task_result = _task_result_for(scan)
         worker = task_result.worker if task_result else None
         reason, recheck_inactivity_cutoff = _stale_reason(
-            scan, worker, worker_alive, stale_cutoff, inactivity_cutoff
+            scan,
+            worker,
+            responsive_workers,
+            unresponsive_workers,
+            stale_cutoff,
+            inactivity_cutoff,
         )
         if reason is None:
             continue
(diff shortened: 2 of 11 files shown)


Subhro-ai:
Thanks for the automated review. This task is intentionally modeled on the existing `cleanup_stale_attack_paths_scans` / `reconcile_orphans` watchdogs, so most of the flagged patterns are deliberate matches to that reviewed precedent rather than oversights. Summary:

**Applied**
- **Docs recovery estimate** — corrected to ~30–45 min (the `SCAN_INACTIVITY_THRESHOLD_MINUTES` window plus up to one 15-minute schedule interval), instead of "within 30 minutes".

**Kept as-is, matching the existing watchdog precedent**
- **Cross-tenant read via `all_objects.using(MainRouter.admin_db)`** — identical to `cleanup_stale_attack_paths_scans` (`tasks/jobs/attack_paths/cleanup.py`). The cross-tenant query is read-only discovery; every mutation goes through `rls_transaction(tenant_id)` in `_fail_stale_scan`. A per-tenant fan-out would still need a cross-tenant tenant enumeration first.
- **No `RLSTask` base** — matches both `cleanup_stale_attack_paths_scans_task` and `reconcile_orphan_tasks_task`. This is a cross-tenant watchdog with no `tenant_id`; `RLSTask.apply_async` reads `kwargs["tenant_id"]`, so it isn't applicable here.
- **Tests create rows without `rls_transaction`** — consistent with `test_attack_paths_scan.py` / `test_tasks.py`; the `test` DB user bypasses RLS by design, and `MainRouter.admin_db` is patched to `default` in the suite.
- **Revoke only when a worker is recorded** — matches the attack-paths reaper (`revoke=worker is not None`). Note `scan-perform` runs with `acks_late=False` (unlike the attack-paths tasks, which use the global `acks_late=True`), so an already-consumed message is not redelivered — there is no stale message to revoke in the no-worker case.
- **No startup validation of the thresholds** — the sibling `ATTACK_PATHS_SCAN_*` thresholds aren't validated either, the defaults are safe, and a hard `raise` at settings-import time would take down the API/worker/beat on a config typo.

Happy to adjust any of these if you'd prefer a different convention for the scan reaper specifically.

api/src/backend/tasks/jobs/scan_cleanup.py
```diff
@@ -0,0 +1,223 @@
...
+            task_result,
+            reason,
+            revoke=worker is not None,
+            inactivity_cutoff=recheck_inactivity_cutoff,
+        ):
+            cleaned_up.append(str(scan.id))
+            # Release the provider queue the dead scan was blocking. Runs after
+            # the FAILED state is committed so the next dispatch no longer sees
+            # this scan as the active one.
+            _drain_provider_queue(str(scan.tenant_id), str(scan.provider_id))
```
josema-xyz: If this call fails, the scan is already `FAILED`. Future cleanup runs only select `EXECUTING` scans, so they will not try to start the queued scan again. Could this leave the provider queue stuck after a temporary error?

api/src/backend/tasks/jobs/scan_cleanup.py
```diff
@@ -0,0 +1,223 @@
...
+        .select_related("task__task_runner_task")
+    )
+
+    workers = {
+        tr.worker
+        for scan in executing_scans
+        if (tr := _task_result_for(scan)) and tr.worker
+    }
+    # Ping each distinct worker once; errors are treated as "alive" upstream.
+    worker_alive = {worker: is_worker_alive(worker) for worker in workers}
```
josema-xyz: This check runs before marking the scan as failed and stopping its task, but `is_worker_alive()` only waits one second and returns `False` when the worker does not reply in time. A busy worker could be treated as dead. Since #11986 already handles this with retries and an unknown state, should we use the same logic here?

api/src/backend/tasks/jobs/scan_cleanup.py
```diff
@@ -0,0 +1,223 @@
...
+
+    Returns a summary dict with the count and ids of the scans cleaned up.
+    """
+    now = datetime.now(tz=UTC)
+    stale_cutoff = now - timedelta(minutes=SCAN_STALE_THRESHOLD_MINUTES)
+    inactivity_cutoff = now - timedelta(minutes=SCAN_INACTIVITY_THRESHOLD_MINUTES)
+
+    executing_scans = list(
+        Scan.all_objects.using(MainRouter.admin_db)
+        .filter(state=StateChoices.EXECUTING)
```
josema-xyz: This only checks `EXECUTING` scans. A task can be lost before the scan reaches that state, while `_get_dispatched_provider_scan` still treats it as active and blocks the queue. Also, this PR uses Prowler version 5.22, before #11848 added this queue. Can we confirm this PR fixes the reported case?

josema-xyz (commented):
Thanks for taking this on. The approach of failing a stuck scan and then releasing the provider queue makes sense, and the tests and documentation make the change easy to follow.

I left three inline comments about:

- avoiding a false worker-death result from a single short ping;
- retrying the queue release if it fails after the scan is marked as failed;
- confirming this covers #12007, since the report uses 5.22.0 and this cleanup only checks `EXECUTING` scans.

I do not think the CodeRabbit suggestions about RLS, `RLSTask`, or revoking tasks without a recorded worker apply here. The threshold validation is reasonable, but I do not see it as blocking.

Once the inline points are addressed, this should be ready for another review.

api/src/backend/tasks/jobs/scan_cleanup.py
Subhro-ai: Good call. Switched to the same bounded-retry ping used for Attack Paths in #11986 (`_ping_workers`): it pings with exponential backoff (5s/10s/20s) over up to 3 attempts and returns an "unknown" state when the control bus fails. A scan is only failed when its worker is in the *confirmed*-unresponsive set — a busy or briefly-unreachable worker is preserved and retried on the next run.

api/src/backend/tasks/jobs/scan_cleanup.py
Subhro-ai: Fixed. Added a self-healing second pass (`_drain_pending_provider_queues`) that runs on every cleanup and releases any provider queue still holding a `QUEUED` scan with no active scan. So if the inline release fails transiently after the scan is already `FAILED`, the next run retries it — recovery no longer depends on re-selecting the now-terminal scan. `_dispatch_next_queued_provider_scan` re-checks under lock and no-ops when a scan is genuinely active, so healthy serialization is untouched.

api/src/backend/tasks/jobs/scan_cleanup.py
Subhro-ai: Two parts:

- **Task lost before `executing`:** the candidate set now also includes `available`/`scheduled` scans whose task is still in a dispatched state (`PENDING`/`STARTED`/`PROGRESS`) — the same thing `_get_dispatched_provider_scan` treats as active — so a task lost before it reaches `executing` is reaped too. A scan merely waiting in the broker queue (no recorded worker, no `started_at`) is preserved, so this never fails a not-yet-started scan.
- **#12007 / 5.22.0:** the reporter’s 5.22.0 predates the per-provider queue (#11848, v5.33.0), so on that version a "Queued" (`available`) scan that never starts is a worker/broker throughput problem rather than this DB queue. On current master the queue exists, and a worker dying mid-scan strands the `executing` scan *and* leaves every following scan for that provider in `Queued` — the reported symptom ("other jobs sometimes start, but the queued one never does"). This watchdog targets that path. I couldn’t run the API suite locally (no uv/Postgres on my box), so the new unit tests are the main verification — happy to add any scenario you’d like.

Subhro-ai:
@josema-xyz thanks for the thorough review. Pushed a commit addressing all three inline points:

- **Worker liveness** — replaced the single 1s ping with the bounded-retry + unknown-state `_ping_workers` from #11986, so a busy worker is never misjudged as dead.
- **Pre-`executing` loss** — the reaper now also covers `available`/`scheduled` scans with a still-dispatched task, which block the queue the same way.
- **Queue-drain retry** — added a self-healing second pass that releases any provider queue with a `QUEUED` scan but nothing active, so a drain dropped by a transient error is retried instead of stranding the queue.

Replied inline with the details (including the #12007 / 5.22.0 confirmation). Left the CodeRabbit RLS/`RLSTask`/revoke items as-is, per your note. Also re-authored the commits to the correct account. Note I couldn’t run the API suite locally (no uv/Postgres on my box), so the new unit tests are the main verification.

api/src/backend/tasks/jobs/scan_cleanup.py
```diff
@@ -0,0 +1,358 @@
...
+        Scan.all_objects.using(MainRouter.admin_db)
+        .filter(
+            Q(state=StateChoices.EXECUTING)
+            | Q(
+                state__in=(StateChoices.AVAILABLE, StateChoices.SCHEDULED),
+                task__task_runner_task__status__in=DISPATCHED_SCAN_TASK_STATES,
+            )
+        )
+        .select_related("task__task_runner_task")
+    )
```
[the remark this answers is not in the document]
Subhro-ai: Both `_stale_scan_candidates` and `_drain_pending_provider_queues` filter to the small set of non-terminal scans (`executing`/`available`/`scheduled`) — at any moment roughly on the order of the number of active providers — and the `TaskResult` join is keyed off that small candidate set rather than scanned. There is currently no `state`-leading index on `scans`, so the periodic filter does rely on a scan of the table.

Happy to add a partial index on the non-terminal states (which both periodic queries would use) via `AddIndexConcurrently`, following the existing perf-index migrations. @josema-xyz — would you like that included in this PR, or kept as a focused follow-up?

api/src/backend/tasks/jobs/scan_cleanup.py
AdriiiPRodri: Thanks for adding the bounded retries. I retested the current head, and the original safety concern is still partially present.

Celery does not propagate an ordinary reply timeout as an exception here. `Mailbox._collect()` catches `socket.timeout` and returns its response list; when no worker replied, that list is empty and `Inspect._prepare()` returns `None`. The implementation only treats exceptions that escape `ping()` as unknown liveness.

Consequently, when the 5, 10, and 20-second attempts receive no replies, `(response or {})` is empty on every attempt, every destination remains in `pending`, and `_ping_workers()` returns those destinations as confirmed unresponsive.

I reproduced this with an `executing` scan whose task was `STARTED`: three `None` responses caused the scan to be marked `FAILED`, the task result to become `FAILURE`, and a terminating revoke to be scheduled.

Please treat an empty or `None` inspect result as unknown control-plane health unless there is an independent signal that proves the worker is gone. A regression test should cover `ping() -> None`, not only exceptions.

api/src/backend/tasks/jobs/scan_cleanup.py
AdriiiPRodri: The current implementation still does not recover the no-worker pre-execution case.

`_stale_reason()` returns early whenever `started_at is None`, so an `available` or `scheduled` scan with a `PENDING` task, no worker, and no `started_at` is preserved forever. The comment says it is preserved until the stale ceiling, but no timestamp is evaluated in this branch

I reproduced this with a 72-hour-old `PENDING` scan followed by a `QUEUED` scan for the same provider. Cleanup returned `cleaned_up_count=0`; the first scan remained active and the follower was not dispatched because `_get_dispatched_provider_scan()` still found the old `PENDING` task.

Please age this state from a trustworthy pre-execution timestamp such as `TaskResult.date_created` or the scan/task insertion time, then reload and revalidate the task state under the final lock. Until that path is covered, I do not think the PR has demonstrated that it fixes the reported old queued-scan case in #12007

api/src/backend/tasks/jobs/scan_cleanup.py
```diff
@@ -0,0 +1,223 @@
...
+        )
+        if reason is None:
+            continue
+
+        if _fail_stale_scan(
+            scan,
+            task_result,
+            reason,
+            revoke=worker is not None,
+            inactivity_cutoff=recheck_inactivity_cutoff,
```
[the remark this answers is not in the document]
AdriiiPRodri: I think this is relevant specifically to the pre-consumption case, although I would narrow the requested behavior

`acks_late=False` prevents redelivery after a worker has consumed and acknowledged a task. It does not withdraw a `PENDING` message that has never been consumed and may still be waiting in the broker

Once the no-worker pre-execution case is given a finite expiry, the old task ID should be revoked with `terminate=False` before the provider queue is released. Otherwise, the old message can arrive after the scan has been failed and after its follower has been dispatched. I would retain `terminate=True` only when a recorded running worker must actually be stopped

api/src/backend/config/django/base.py
```diff
@@ -315,6 +315,17 @@
...
+# Stale scan cleanup. A scan whose worker dies mid-run is stranded in
+# `executing`; because scans run one-at-a-time per provider (#11848), that also
+# blocks every queued scan for the provider. The periodic `scan-cleanup-stale-scans`
+# task fails such scans and drains the queue. A scan on a live worker is only
+# reaped after the stale ceiling (matched to the long hard time-limit); a scan
+# whose worker is gone is reaped after the shorter inactivity window.
+SCAN_INACTIVITY_THRESHOLD_MINUTES = env.int("SCAN_INACTIVITY_THRESHOLD_MINUTES", 30)
+SCAN_STALE_THRESHOLD_MINUTES = env.int(
+    "SCAN_STALE_THRESHOLD_MINUTES", 2880
+)  # 48h
```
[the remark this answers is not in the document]
AdriiiPRodri: Confirmed. These settings control a destructive recovery action, so invalid values should not silently change the safety policy

Zero or negative values can make legitimate scans immediately eligible, and `SCAN_STALE_THRESHOLD_MINUTES < SCAN_INACTIVITY_THRESHOLD_MINUTES` reverses the intended escalation order. The stale ceiling should also be validated against the effective hard time limit for `scan-perform`; the defaults align at 48 hours, but operators can configure them independently

I would validate the relationship at startup. If failing startup is considered too disruptive, the safe alternative is to disable this watchdog and emit a prominent configuration error, rather than run it with an unsafe policy.

api/src/backend/tasks/jobs/scan_cleanup.py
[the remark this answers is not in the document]
AdriiiPRodri: I verified this locally. There is no state-leading or relevant partial index on `scans`, and `EXPLAIN (ANALYZE, BUFFERS)` selected a sequential scan for both `_stale_scan_candidates()` and `_drain_pending_provider_queues()`.

The development fixture is too small for its execution time to be meaningful, but it confirms the access path. Because these queries run across all tenants every 15 minutes, their scan cost grows with retained scan history, not with the number of currently active providers

Please either add the minimum concurrent partial index supported by production-scale `EXPLAIN` evidence or provide measured production cardinalities and an explicitly accepted follow-up. The fact that matching non-terminal rows are few does not prevent PostgreSQL from inspecting the historical table

api/src/backend/tasks/jobs/scan_cleanup.py
AdriiiPRodri: Verified. `_drain_pending_provider_queues()` retries the release on subsequent cleanup runs, and the dispatcher rechecks for an active scan under the provider lock. This addresses the transient post-failure queue-release concern. I am resolving this thread

api/src/backend/tasks/jobs/scan_cleanup.py
```diff
@@ -0,0 +1,223 @@
...
+    """
+    now = datetime.now(tz=UTC)
+    stale_cutoff = now - timedelta(minutes=SCAN_STALE_THRESHOLD_MINUTES)
+    inactivity_cutoff = now - timedelta(minutes=SCAN_INACTIVITY_THRESHOLD_MINUTES)
+
+    executing_scans = list(
+        Scan.all_objects.using(MainRouter.admin_db)
+        .filter(state=StateChoices.EXECUTING)
+        .select_related("task__task_runner_task")
+    )
```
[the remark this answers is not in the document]
AdriiiPRodri: I checked this against the current architecture and agree with the author's explanation. This is a global watchdog, so read-only discovery must cross tenant boundaries. The admin manager is appropriate for discovery, while every `Scan` mutation enters `rls_transaction(tenant_id)`

A per-tenant fan-out would still require trusted cross-tenant enumeration and would add orchestration without strengthening the mutation boundary. I am not requesting this change

api/src/backend/tasks/tasks.py
```diff
@@ -699,6 +700,13 @@ def cleanup_stale_attack_paths_scans_task():
     return cleanup_stale_attack_paths_scans()
 
 
+@shared_task(name="scan-cleanup-stale-scans", queue="celery")
+def cleanup_stale_scans_task():
+    """Periodic watchdog: fail scans stranded in `executing` by a dead worker
+    and dispatch the next queued scan the dead scan was blocking."""
+    return cleanup_stale_scans()
```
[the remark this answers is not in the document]
AdriiiPRodri: Verified. `RLSTask` is not appropriate for this global watchdog because it expects a single `tenant_id` during dispatch and creates a tenant-scoped task record. The analogous global cleanup and reconciliation tasks also use a normal shared task

api/src/backend/tasks/tests/test_scan_cleanup.py
```diff
@@ -0,0 +1,278 @@
...
+            tenant_id=tenant.id,
+        )
+        scan = Scan.objects.create(
+            tenant_id=tenant.id,
+            provider=provider,
+            name="Queued scan",
+            trigger=Scan.TriggerChoices.MANUAL,
+            state=StateChoices.AVAILABLE,
+            task=task,
+        )
```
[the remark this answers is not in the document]
AdriiiPRodri: I do not consider this blocking. The test database user bypasses RLS by design, and the production mutation path is exercised through `rls_transaction(tenant_id)`

The more important missing integration coverage is preserving distinct `admin` and `default` aliases, because the current tests collapse them and cannot expose the transaction boundary between `Scan` and `TaskResult`. I will keep that as a separate review point rather than expanding this fixture-style thread

api/src/backend/tasks/jobs/scan_cleanup.py
```diff
@@ -0,0 +1,358 @@
...
+                fresh_scan.duration = int(
+                    (now - fresh_scan.started_at).total_seconds()
+                )
+                update_fields.append("duration")
+            fresh_scan.save(update_fields=update_fields)
+
+            if task_result:
+                task_result.status = states.FAILURE
+                task_result.date_done = now
+                task_result.save(update_fields=["status", "date_done"])
```
AdriiiPRodri: `task_result` was loaded through `MainRouter.admin_db`, while this block locks and updates `Scan` inside the default RLS transaction. `TaskResult.save()` is therefore routed to the separate `admin` connection and is not covered by the scan transaction, even when both aliases point to the same PostgreSQL database

A process failure after the admin write but before the default transaction commits can leave `TaskResult=FAILURE` while the scan update rolls back. The current tests do not expose this because they patch the admin alias to `default`

Can we either perform the invariant-changing writes through one explicit transaction-capable connection, or document `Scan` as the source of truth and add reconciliation for a partial update? The current task result should also be reloaded before writing because this snapshot predates up to 35 seconds of worker probing

api/src/backend/tasks/tests/test_scan_cleanup.py
```diff
@@ -0,0 +1,378 @@
+import uuid
```
AdriiiPRodri: `make lint-api` currently fails because this `uuid` import is unused. Please remove it

In addition, `uv run ruff format --check . --exclude contrib` reports formatting changes in:
- `api/src/backend/config/django/base.py`
- `api/src/backend/tasks/jobs/scan_cleanup.py`
- `api/src/backend/tasks/tests/test_scan_cleanup.py`

The PR should pass both checks before approval

AdriiiPRodri (changes requested):
Thanks for the substantial work on this watchdog. The overall fail-and-release design, RLS-scoped mutation, post-commit revocation, reversible Beat migration, and self-healing queue drain are good foundations.

I validated the exact PR head locally with the full development stack:

- `make dev` succeeded on the PR head and on a clean merge into the current `master`;
- the task test suite passed with 882 tests;
- the focused cleanup/orphan suite passed with 41 tests on both trees;
- migration 0098 passed forward and reverse checks;
- the periodic task ran successfully through a real Celery worker;
- Celery Beat loaded the database schedule.

I am requesting changes for two production-safety issues:

1. An old `available`/`PENDING` scan with no worker and no `started_at` is preserved forever, so its provider queue remains blocked
2. `Inspect.ping() -None` is treated as confirmed worker death, which can fail and terminate a legitimate scan during a control-bus timeout.

I reproduced both cases against the running PostgreSQL/Celery stack and added the details to Josema's existing threads rather than opening duplicates.

Before approval, I also think we need to address or explicitly resolve:

- non-terminating revocation for an expired task that has not reached a worker;
- threshold validation;
- the unindexed cross-tenant query paths;
- the `Scan`/`TaskResult` cross-alias consistency window;
- the current Ruff lint and format failures.

The queue-drain retry is correctly addressed. I also agree with the author that the proposed per-tenant fan-out and `RLSTask` changes are not appropriate for this global watchdog

Once the two state-machine regressions have tests and the static gates pass, I would be happy to review the updated behavior.

api/src/backend/tasks/jobs/scan_cleanup.py
Subhro-ai: Thank you for tracing this to `Mailbox._collect()` — you are right, and the reproduction is exactly the failure mode. `_ping_workers()` now distinguishes *no reply* from *proof of death*: a worker is only reported unresponsive when some other destination answered during the sweep, which demonstrates the control bus itself is healthy. An empty or `None` result on every attempt returns `None` (unknown liveness) just like an exception, and unknown liveness is treated as alive.

Because "unknown" now covers the single-worker case (nobody left to answer), unknown-liveness scans are no longer preserved indefinitely: they fall through to the stale-ceiling branch, so a dead single-worker deployment still recovers after `SCAN_STALE_THRESHOLD_MINUTES`.

Regression tests added for `ping() -> None` and `ping() -> {}`, alongside the existing exception case, plus one asserting a worker that answers only on a later attempt is never reported dead.

api/src/backend/tasks/jobs/scan_cleanup.py
Subhro-ai: Confirmed and fixed — the early return on `started_at is None` made the comment I wrote inaccurate, and your 72-hour repro is the real #12007 shape.

`_reference_time()` now ages a scan from `started_at` when it exists, falling back to `TaskResult.date_created` and then `Scan.inserted_at`, so a scan that never reached a worker is aged from a trustworthy pre-execution timestamp instead of being skipped. The task state is re-read under the final lock (`_task_is_ready`) so a task that reported its own result during the up-to-35s probing window is not overwritten.

Added a regression test mirroring your scenario: a 72-hour-old `PENDING` scan with no worker followed by a `QUEUED` scan for the same provider — the old scan is now failed and the follower is dispatched. A recently-created `PENDING` scan still waiting in the broker is preserved, covered by its own test.

api/src/backend/tasks/jobs/scan_cleanup.py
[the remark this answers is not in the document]
Subhro-ai: Agreed, and narrowed as you described. `terminate` is now computed per scan as `bool(worker) and scan.state == EXECUTING` — only a recorded worker actually running the scan is SIGTERMed. Every other reaped scan, including the no-worker pre-execution case, is revoked with `terminate=False` so a never-consumed `PENDING` message is discarded if the broker later delivers it.

Ordering is preserved: revocation is registered `on_commit` of the per-scan fail transaction, and `_drain_pending_provider_queues()` runs after all reaps, so the old task id is always revoked before a follower is dispatched.

api/src/backend/config/django/base.py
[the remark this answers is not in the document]
Subhro-ai: Implemented, using the non-disruptive option you outlined. Startup is not aborted; instead the thresholds are validated in settings and, when the policy is unsafe (non-positive values, or `SCAN_STALE_THRESHOLD_MINUTES < SCAN_INACTIVITY_THRESHOLD_MINUTES`), `SCAN_CLEANUP_ENABLED` is set to `False` and a prominent error is logged. `cleanup_stale_scans()` then logs and no-ops rather than running with an unsafe policy.

The stale ceiling is also cross-checked against the effective hard time limit (`DJANGO_CELERY_LONG_TASK_TIME_LIMIT`) and logs a warning when it is configured below it, since that combination could reap a legitimately long scan on a live worker.

api/src/backend/tasks/jobs/scan_cleanup.py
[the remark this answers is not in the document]
Subhro-ai: Thanks for running the `EXPLAIN (ANALYZE, BUFFERS)` — that settles it. Added a partial index covering only the non-terminal states both sweeps filter on:

```
Index(fields=["state", "tenant_id", "provider_id"],
      condition=Q(state__in=["executing", "available", "scheduled"]),
      name="scans_non_terminal_state_idx")
```

Applied via `AddIndexConcurrently` in migration `0099` with `atomic = False`, matching the existing performance-index migrations. The predicate keeps the index proportional to currently-active scans rather than retained history, which is the growth you flagged. I do not have a production-scale dataset to attach `EXPLAIN` output from, so please do re-run your local check against the updated head.

api/src/backend/tasks/jobs/scan_cleanup.py
Subhro-ai: Good catch on the alias boundary. `TaskResult` is routed to `admin` by `MainRouter` for both read and write (its table matches the `django_` prefix), so it cannot join the scan RLS transaction on `default`.

I took the second option you offered: `Scan` is now explicitly the source of truth. The task-result write and revocation moved into an `on_commit` hook (`_finalize_task_result`), so the inversion you described — `TaskResult=FAILURE` while the scan update rolls back — is no longer reachable; a crash between the two can only leave a *stale* task result, never a failed one for a running scan. That direction is self-correcting because candidate selection is driven by `Scan.state`, and the queue-drain pass is idempotent.

The task result is also reloaded inside the hook and re-checked against `READY_STATES` before writing, so the pre-probing snapshot is never blindly persisted. Documented in the module docstring.

api/src/backend/tasks/tests/test_scan_cleanup.py
Subhro-ai: Fixed — the unused `uuid` import is gone. I also went through the three files for formatting: no line now exceeds 88 columns, multi-line calls carry magic trailing commas so they stay exploded, and the two implicit-concat fragments that had a redundant `f` prefix were corrected.

I cannot run `make lint-api` locally (no `uv`/PostgreSQL on this machine), so I would appreciate you re-running both gates on the updated head — if anything still differs I will push the exact formatter output.

Subhro-ai:
@AdriiiPRodri thank you for validating this against a real stack — the two state-machine bugs were both real and both are fixed in `20f413c`.

**The two blocking regressions**
- **`ping() -> None` treated as death.** You were right about `Mailbox._collect()` swallowing the reply timeout. A worker is now only reported unresponsive when another destination answered in the same sweep, proving the control bus is healthy; an empty/`None` result is unknown liveness, like an exception. Since "unknown" now also covers a dead single-worker deployment, unknown-liveness scans fall through to the stale ceiling rather than being preserved forever.
- **No-worker pre-execution case never aged.** Fixed via `_reference_time()` (`started_at` -> `TaskResult.date_created` -> `Scan.inserted_at`), with the task state revalidated under the final lock. Your 72-hour repro is now a regression test, and it dispatches the follower.

**The five pre-approval items**
- **Non-terminating revoke** — `terminate` is now `bool(worker) and state == EXECUTING`; everything else is revoked with `terminate=False`, before the queue is released.
- **Threshold validation** — took your non-disruptive option: unsafe policy sets `SCAN_CLEANUP_ENABLED=False` and logs a prominent error instead of aborting startup, plus a warning when the ceiling is below the hard time limit.
- **Indexes** — added a partial index on the non-terminal states via `AddIndexConcurrently` (migration `0099`), so the sweeps scale with active scans, not retained history.
- **Cross-alias window** — `TaskResult` is routed to `admin` by `MainRouter`, so it cannot join the scan transaction; I made `Scan` the explicit source of truth and moved the task-result write plus revocation into an `on_commit` hook, which makes the inversion you described unreachable.
- **Lint/format** — unused `uuid` import removed and the three files reformatted.

One caveat worth stating plainly: I have no `uv`/PostgreSQL on this machine, so I could not run `make lint-api`, the test suite, or `EXPLAIN` myself — the new unit tests and a careful read are my only verification. Your local validation is genuinely load-bearing here, so please do re-run the gates on the updated head; I will push the exact formatter output if anything still differs.

docs/troubleshooting.mdx
```diff
@@ -200,7 +200,11 @@ A fix addressing this permission issue is being evaluated in [PR #9953](https://
...
+
+A periodic task (`scan-cleanup-stale-scans`) automatically detects these scans — it confirms the worker is gone, marks the scan `failed`, and releases the provider's queue so the next scan starts. It runs every 15 minutes, so recovery normally completes within roughly 30–45 minutes of the last scan activity (the `SCAN_INACTIVITY_THRESHOLD_MINUTES` window, default 30 minutes, plus up to one scheduling interval).
+
+When the worker cannot be reached at all — for example a single-worker deployment where that worker died — its liveness cannot be confirmed, so the scan is instead released after `SCAN_STALE_THRESHOLD_MINUTES` (default 48 hours) to guarantee a healthy scan is never failed by a temporary connectivity problem. The same ceiling applies to a scan whose task was lost before it ever started. If you need to recover sooner than that, use the manual workaround below.
```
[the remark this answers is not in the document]
Subhro-ai: Added in `881c512` — the badge now sits directly below the header, and the `VersionBadge` import already existed at the top of the file.

I set it to `5.36.0` on the basis that `master` is currently at `5.35.0`, so this would be the next minor. @AdriiiPRodri / @josema-xyz — please bump it if this is targeted at a different release; I have no way to know the intended milestone from here.

One note on placement: the section header itself is not new (it previously documented only the manual `UPDATE scans SET state = 'failed'` workaround), so the badge should be read as marking the *automatic recovery* described in the two paragraphs beneath it. Happy to reword the heading if you would rather the badge scope be unambiguous.

api/src/backend/tasks/jobs/scan_cleanup.py
```diff
@@ -0,0 +1,427 @@
...
+from api.db_router import MainRouter
+from api.db_utils import rls_transaction
+from api.models import Scan, StateChoices
+from celery import current_app, states
+from celery.utils.log import get_task_logger
+from config.django.base import (
+    SCAN_CLEANUP_ENABLED,
+    SCAN_INACTIVITY_THRESHOLD_MINUTES,
+    SCAN_STALE_THRESHOLD_MINUTES,
+)
```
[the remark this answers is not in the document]
Subhro-ai: Respectfully declining this one — the module-level import is the established convention for this exact kind of module, and switching would make the file inconsistent with the sibling it deliberately mirrors.

`tasks/jobs/attack_paths/cleanup.py` — the reaper this watchdog was modelled on — reads its two thresholds the same way:

```python
from config.django.base import (
    ATTACK_PATHS_SCAN_INACTIVITY_THRESHOLD_MINUTES,
    ATTACK_PATHS_SCAN_STALE_THRESHOLD_MINUTES,
)
```

Six other modules under `tasks/jobs/` do the same (`integrations.py`, `muting.py`, `report.py`, `scan.py`, `threatscore_utils.py`, and `tasks.py` itself).

On the specific risks raised:

- **"any alternative settings module is ignored"** — not applicable here. `testing.py`, `devel.py` and `production.py` all start with `from config.django.base import *` and none of them reassign the `SCAN_*` values, so the bound value is the effective value in every configuration.
- **"tests cannot exercise the disabled path"** — that was a fair gap, so I have closed it in `881c512` rather than by refactoring: `test_no_ops_when_disabled_by_invalid_configuration` patches `tasks.jobs.scan_cleanup.SCAN_CLEANUP_ENABLED` and asserts a 30-day-old `executing` scan with a dead worker is left untouched and that `_ping_workers` is never even called. Patching the module attribute is the standard counterpart to a module-level import and works without `override_settings`.

If the maintainers would prefer `django.conf.settings` as the go-forward pattern I am glad to switch, but I would rather not have this PR be the one file that diverges.

api/src/backend/config/django/base.py
```diff
@@ -315,6 +316,53 @@
...
+        env.int("DJANGO_CELERY_LONG_TASK_TIME_LIMIT", 48 * 60 * 60) // 60
+    )
+    if SCAN_STALE_THRESHOLD_MINUTES < _scan_hard_limit_minutes:
+        logging.getLogger(__name__).warning(
+            "SCAN_STALE_THRESHOLD_MINUTES (%s) is below the scan hard time limit "
+            "(%s minutes); a long-running scan on a live worker may be reaped "
+            "before Celery stops it",
+            SCAN_STALE_THRESHOLD_MINUTES,
+            _scan_hard_limit_minutes,
+        )
```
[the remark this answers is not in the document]
Subhro-ai: Partly correct, and I have acted on the part that holds up — but the stated mechanism is not quite what happens, so I want to be precise about what changed and why.

**What is right:** settings are imported before `django.setup()` calls `configure_logging`, so at the moment this record is emitted the configured handlers in `config/custom_logging.py` do not exist yet. The record therefore never passes through the `ndjson`/`human_readable` formatters or the `transaction_id` filter — so anything consuming structured logs will not see it. That is a real operability gap.

**What is not right:** the record is not dropped. With no handlers configured, Python falls back to `logging.lastResort`, which writes `WARNING` and above to stderr. Both the `error()` and the `warning()` still reach container logs, just unformatted. And `disable_existing_loggers: True` is not the cause — it disables the `config.django.base` logger *after* `dictConfig` runs, which is harmless here because the module logs exactly once, at import, and never again.

**On the suggested remedies:** `warnings.warn` would be a regression — warnings are routinely filtered in production and are less visible than stderr, so the message would end up harder to see, not easier. A system check would be idiomatic, but it only fires on `manage.py` entry points; Beat and the worker are precisely where this task runs.

**What I did instead** (`881c512`): kept the import-time message as a best-effort early hint, and made the runtime error self-contained so it no longer depends on that hint surviving. `SCAN_CLEANUP_CONFIG_ERROR` is now exported alongside `SCAN_CLEANUP_ENABLED`, and the task logs the actual reason through `get_task_logger` — which *is* fully configured by then:

```python
logger.error(
    "Stale scan cleanup skipped: invalid threshold configuration - %s",
    SCAN_CLEANUP_CONFIG_ERROR,
)
```

Previously this said "see the configuration error logged at startup", which pointed at exactly the message you identified as unreliable. Now the operator gets the full reason, in structured form, every 15 minutes until the configuration is fixed. `SCAN_CLEANUP_ENABLED` remains the runtime gate, as you suggested.

api/src/backend/tasks/tests/test_scan_cleanup.py
```diff
@@ -0,0 +1,560 @@
...
+            updated_at=datetime.now(tz=UTC) - timedelta(days=30),
+            worker="dead-worker",
+        )
+
+        result = cleanup_stale_scans()
+
+        assert result == {"cleaned_up_count": 0, "scan_ids": [], "queues_checked": 0}
+        mock_ping.assert_not_called()
+        scan = Scan.objects.get(provider=aws_provider)
+        assert scan.state == StateChoices.EXECUTING
```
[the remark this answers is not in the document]
Subhro-ai: Fixed in `c3c3ab9` — `test_no_ops_when_disabled_by_invalid_configuration` now patches `SCAN_CLEANUP_CONFIG_ERROR` to a representative string, patches `logger.error`, and asserts it was called exactly once with that reason in the message, alongside the existing no-op assertions. Good catch — the previous version would have stayed green if that logging line were deleted.

docs/troubleshooting.mdx
```diff
@@ -200,7 +200,13 @@ A fix addressing this permission issue is being evaluated in [PR #9953](https://
 
 ### Scan Stuck in Executing State After Worker Crash
 
-When running Prowler Local Server via Docker Compose, a scan may remain indefinitely in the `executing` state if the worker process crashes (for example, due to an Out of Memory condition) before it can update the scan status. Since it is not currently possible to cancel a scan in `executing` state through the UI, the workaround is to manually update the scan record in the database.
+<VersionBadge version="5.36.0" />
```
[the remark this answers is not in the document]
Subhro-ai: Declining — this contradicts the repo's own convention, including the guideline's own example.

Every `<VersionBadge>` usage under `docs/` currently has a blank line before it, header included:

```
### Worker Uses Too Much Memory on Hosts with Many CPUs

<VersionBadge version="5.31.0" />
```
(same file, line 90-92 — pre-existing, not touched by this PR)

And the guideline example in `docs/AGENTS.md` itself:

```
## New Feature Name

<VersionBadge version="4.5.0" />
```

I counted 14 other instances across `docs/` (secret-scanning-checks.mdx, server-sent-events.mdx, prowler-lighthouse-ai.mdx, configuration_file.mdx, reporting.mdx, the provider getting-started/authentication pages, etc.) — all of them have the header, a blank line, then the badge. None place the badge on the line immediately after the header with no blank line.

I'm keeping the blank line so this entry matches every sibling entry in the same file rather than becoming the one inconsistent case.

Subhro-ai:
@AdriiiPRodri @josema-xyz, checking in a week after the last review pass (`20f413c`). Since then the branch has had two more pushes addressing CodeRabbit's re-review of the changes:

- `881c512` : the config-error message now reaches structured logs at runtime (via a new `SCAN_CLEANUP_CONFIG_ERROR` export) instead of only at import time before Django's logging is configured; added the missing test for the disabled-cleanup path; added the `VersionBadge` the docs guidelines call for.
- `c3c3ab9` :  hardened that new test to assert the log call itself, not just the no-op behavior.

CodeRabbit re-verified both pushes and withdrew its follow-up findings after checking the code directly (see the `<review_comment_withdrawn>` replies on `881c512d4` and `c3c3ab955`), so from an automated-review standpoint the thread is clean.

Head is `c3c3ab9`, `mergeable: true`. No rush if you're busy, just didn't want this to go stale after the amount of back-and-forth already in the thread. Let me know if there's anything else you'd like changed.

api/src/backend/tasks/jobs/scan_cleanup.py
```diff
@@ -0,0 +1,431 @@
...
+        )
+        .values_list("tenant_id", "provider_id")
+        .distinct()
+    )
+
+    for tenant_id, provider_id in pending:
+        _dispatch_next_queued_provider_scan_best_effort(
+            str(tenant_id),
+            str(provider_id),
+        )
```
josema-xyz, line 429: If sending to Celery fails, the scan has already been saved as `PENDING`. Later cleanup runs only search for `QUEUED`, so they do not retry it and the queue gets stuck again. Can we cover this case?

api/src/backend/tasks/jobs/scan_cleanup.py
```diff
@@ -0,0 +1,431 @@
...
+                elapsed = now - fresh_scan.started_at
+                fresh_scan.duration = int(elapsed.total_seconds())
+                update_fields.append("duration")
+            fresh_scan.save(update_fields=update_fields)
+
+            if task_result is not None:
+                on_commit(
+                    partial(_finalize_task_result, task_result.task_id, terminate),
+                    using=fresh_scan._state.db,
+                )
```
josema-xyz, line 361: If this operation fails after the scan is saved as `FAILED`, I do not see anything retrying the old task cleanup. Later cleanup runs ignore the scan because it is already failed. Can we cover this case before releasing the queue?

api/src/backend/tasks/tests/test_scan_cleanup.py
```diff
@@ -0,0 +1,567 @@
...
+@pytest.mark.django_db
+class TestCleanupStaleScans:
+    @pytest.fixture(autouse=True)
+    def execute_on_commit_callbacks(self):
+        # Fire the post-commit finalization synchronously so tests can assert it.
+        with patch(
+            "tasks.jobs.scan_cleanup.on_commit",
+            side_effect=lambda callback, **kwargs: callback(),
+        ):
+            yield
```
josema-xyz, line 20: This fixture runs `on_commit` immediately, before the real commit. The test does not check the production order and can hide callback failures. Can we use Django’s real test callbacks here?

docs/troubleshooting.mdx
josema-xyz, line 203:
```suggestion
<VersionBadge version="5.40.0" />
```

PR #12010, opened 2026-07-16
feat(ui): filter empty Attack Paths queries from the selector in Cloud

danibarranqueroo:
### Context

Frontend piece of the Attack Paths "precompute query results / filter empty queries" epic ([PROWLER-2195](https://prowlerpro.atlassian.net/browse/PROWLER-2195)).

Depends on the Cloud API exposing the per-query result summary (`result_summary`) on the queries endpoint. That work is Cloud-only; this PR is written so it is a no-op on self-hosted (see the feature flag below).

### Description

In **Prowler Cloud**, the Attack Paths query selector now lists only the queries that returned data for the selected scan. Empty queries are hidden, so users aren't offered queries that can't return anything for their account.

The queries API now returns a `result_summary` (`{ status, has_data }` or `null`) per query. The adapter filters on it with one rule — **hide a query only when the scan computed it and found it empty**:

- `has_data === true` → shown
- `has_data === false` (status `ok`) → hidden
- errored during precompute (`has_data` null) → shown
- parameterized queries (no summary, run live on demand) → shown
- scans predating the precompute step (no summary) → shown

So a `null`/missing summary is never treated as "empty", which means a partial precompute or a broken query never wrongly hides something.

**Feature-flagged.** All of this is gated behind `NEXT_PUBLIC_IS_CLOUD_ENV` via the existing `isCloud()` helper. Self-hosted deployments, whose API does not precompute result summaries, keep showing every query exactly as today.

### Changes

- `types/attack-paths.ts` — add `AttackPathQueryResultSummary` and an optional `result_summary` on the query attributes.
- `actions/attack-paths/queries.adapter.ts` — `shouldShowQuery` predicate (short-circuits to "show all" when not in Cloud); applied in `adaptAttackPathQueriesResponse`.
- `actions/attack-paths/queries.adapter.test.ts` — filtering tests covering a mix of has_data true/false, errored, and parameterized queries, plus the flag-off (self-hosted) case.

### Steps to review

1. `queries.adapter.ts` — the predicate and the flag gate.
2. `queries.adapter.test.ts` — 5 tests; verify the "absent ≠ empty" and flag-off behaviour.
3. Note the selector component is unchanged — filtering lives in the adapter (data layer), so the component test the ticket mentions is covered at the adapter layer where the logic actually is.

Verified locally against a real Cloud API + a genuine scan: with the flag on, the dropdown drops from all queries to only the with-data ones; with the flag off, all queries show. `pnpm --filter ui test` for the attack-paths actions passes (23 tests), eslint/prettier clean.

### License

[PROWLER-2195]: https://prowlerpro.atlassian.net/browse/PROWLER-2195?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Attack Paths query selector now hides only definitively empty-result queries in Prowler Cloud.
  * Queries with data, errored/unknown summaries, parameters, and missing/null result summaries remain visible.
  * Self-hosted deployments continue to show all queries.
* **Bug Fixes**
  * Pagination counts now match the filtered (visible) query results.
* **Tests**
  * Expanded and refactored coverage for filtering logic, empty responses, and pagination behavior.
* **Documentation**
  * Added a changelog entry describing the Cloud behavior update.


ui/types/attack-paths.ts
```diff
@@ -129,6 +129,23 @@ export interface AttackPathQueryDocumentationLink {
...
+export const ATTACK_PATH_QUERY_STATUSES = {
+  OK: "ok",
+  ERROR: "error",
+} as const;
+
+export type AttackPathQueryStatus =
+  (typeof ATTACK_PATH_QUERY_STATUSES)[keyof typeof ATTACK_PATH_QUERY_STATUSES];
+
+// Precomputed per-query result summary from the scan (Prowler Cloud only).
+// `has_data` is null when the query errored; the whole object is null/absent
```
pfe-nazaries: Trim or remove the comment, keep and eye on agent docs/comments going huge to avoid too much noise

pfe-nazaries (dismissed):

pfe-nazaries (approved):

PR #12025, opened 2026-07-16
fix(ui): always show Scans page so imported scans are visible

pfe-nazaries:
### Context

Scans imported into the app from the CLI (`prowler … --push-to-cloud`) were not showing up on the Scans page. Those scans attach to a dynamic/placeholder provider that never connects, and the Scans page had an all-or-nothing gate that replaced the whole table with an onboarding empty state whenever no provider was `connected === true`. As a result the imported scans — and the table itself — were never rendered, even though they exist and the API returns them (the API applies no connection filter to scans).

### Description

Decouple "can launch a scan" (needs a connected provider) from "can view scans" (always):

- `ui/app/(prowler)/scans/page.tsx`: remove the `missingScanPrerequisite` gate and the forced onboarding `redirect`; always render `ScansPageShell` + the scans table.
- `ui/components/scans/scans-page-shell.tsx`: render `ScansProvidersEmptyState` as a **non-blocking hint at the top** when there are no providers or none is connected. The table (and therefore imported scans) always renders below it. `Launch Scan` and the `view-first-scan` tour are gated on the full launch condition — a connected provider **and** `manage_scans` permission — so the tour never anchors to a disabled button.
- `ui/components/scans/scans-providers-empty-state.tsx`: render the provider hints through the shared `CustomBanner` (semantic design tokens) instead of a bespoke banner. The disconnected-provider copy is Cloud-gated: only Cloud mentions that imported scans still appear below.
- `ui/components/shadcn/custom/custom-banner.tsx`: minor extension so the shared banner covers the Scans-page hint variants.

Genuine emptiness is still handled by the table's own per-tab `NoScansEmptyState`, and `Launch Scan` stays disabled while a scan can't be launched — so nothing is lost, the hint is purely advisory.

### Steps to review

1. On the Scans page with **no connected provider** but with imported scans, confirm the table renders the imported scan(s) and a non-blocking hint appears at the top; `Launch Scan` is disabled.
2. With **no providers at all**, confirm the "Add a Provider" hint shows at the top and the page is not blocked/redirected.
3. With **a connected provider**, confirm behavior is unchanged (table visible, no hint, `Launch Scan` enabled).
4. With **a connected provider but without `manage_scans`**, confirm `Launch Scan` is disabled and the onboarding tour does not start.
5. Unit tests: `ui/components/scans/scans-page-shell.test.tsx` (hint shown when no connected provider / no providers, absent when connected; tour suppressed without a connected provider or `manage_scans`) and `ui/app/(prowler)/scans/page.test.tsx` (no gate/redirect, shell + scans table always rendered for connected/disconnected/empty providers).

#### 1. No providers configured (with an imported scan)
image

image

#### 2. Providers present but not connected (bad credentials, with an imported scan)

image

image

#### 3. Provider connected (control — no change)

image

image

### Checklist

<details>

</details>

#### UI (if applicable)
- [x] Screenshots/Video - Desktop (X > 1024px)

### License

## Summary by CodeRabbit

* **New Features**
  * Imported scans remain visible even when no provider is connected.
  * The Scans page always displays the scans table, including without configured or connected providers.
  * Added clear banners with links to add or review providers.
* **Bug Fixes**
  * Prevented scan launches when disabled or when no connected provider is available.
* **Refactor**
  * Onboarding prompts now appear only when a provider is connected.
* **Style**
  * Updated banner text colors for improved theme consistency.


Alan-TheGentleman (changes requested):
Two things to address:

1. **The scan tour can start with Launch Scan disabled**. `ScansPageShell` always mounts `view-first-scan`, while `useFallback` only changes the navbar replay action. A direct `?onboarding=view-first-scan` or active sequence therefore starts a tour whose primary target cannot be used when no provider is connected. Keep the table visible, but route or suppress that tour in this state.

2. **The page regression test only inspects source strings**. These assertions still pass if another conditional or helper blocks the table. Render the async page with empty, disconnected, and connected provider responses and assert the resulting behavior.

Alan-TheGentleman (changes requested):
One more issue to address:

1. **`NoProvidersAdded` duplicates the existing banner composition**. `ui/components/providers/no-providers-added.tsx` recreates the Card, icon, heading, description, and CTA structure already provided by `components/shadcn/custom/custom-banner.tsx`. It also adds raw `text-gray-*` and `dark:text-*` colors plus visual classes directly to `Button`, which bypasses the shared component variants and semantic tokens. Reuse or extend `CustomBanner` instead of maintaining a second implementation.

alejandrobailo (changes requested):
Have you tried it on the cloud? I see that the screenshots are from a local server. I wonder if onboarding keeps working right for cloud users. Also, another thing, very important: Import Scans are only for CLOUD users. The screenshots are showing it for OSS users. This can't happen.

Blocking issues, ordered by priority:

1. The onboarding flow is broken when no provider is connected. `view-first-scan` still starts and targets a disabled “Launch Scan” button. The fallback only affects the navbar replay action; it does not handle direct URLs or active onboarding sequences.

2. This fix targets Cloud-imported scans, but changes OSS behavior too. OSS users now see Cloud-only copy stating that imported scans appear below. Keep this message Cloud-specific.

3. “Launch Scan” is not fully disabled. The sidebar action or `?launchScan=true` can still open the modal without a connected provider.

4. The page tests inspect source strings instead of behavior. Please render the async page and cover empty, disconnected, and connected provider states.

5. Minor: the new banner introduces raw `text-gray-*` colors instead of semantic design-system tokens.

alejandrobailo (changes requested):
Most requested changes are fixed. Just two things to improve this fix.

1. The scan tour is still mounted when the user has a connected provider but lacks `manage_scans`. In that state, `Launch Scan` is disabled but the tour still targets it. Please gate `OnboardingTrigger` with the full launch condition, not only `hasConnectedProviders`, and add a test for this case.

2. The page test is behavioral now, but the mocked `ScansPageShell` discards its children. `children.toBeTruthy()` does not verify that the scans table renders. Please render or expose the table subtree and assert its presence for connected, disconnected, and empty provider states.

Minor: the PR description still references deleted/reverted files and should be updated.

alejandrobailo (changes requested):
[P2] Keep the navbar tour action aligned with launch eligibility

The trigger is correctly suppressed when `manage_scans` is missing, but `page.tsx` still exposes the `view-first-scan` action whenever a provider is connected. Since `PageReady` enables the navbar action while `ScansPageShell` does not mount the trigger, read-only Cloud users see an enabled tour button that cannot start anything.

Please set `onboardingAction` to `undefined` when `manage_scans` is absent, preserve the `add-provider` fallback for permitted users without a connected provider, and add a page-level regression test for this case.

alejandrobailo (approved):

PR #12049, opened 2026-07-19
fix(alibabacloud): normalize security group policy case

### LINKED ISSUE #12048: Alibaba Cloud SSH/RDP security group checks false-pass when Policy is 'Accept'
## Summary

The Alibaba Cloud checks `ecs_securitygroup_restrict_ssh_internet` and `ecs_securitygroup_restrict_rdp_internet` can incorrectly return **PASS** for public ingress rules because the checks compare the rule policy case-sensitively against the lowercase string `"accept"`.

Alibaba Cloud ECS `DescribeSecurityGroupAttribute` returns the policy as `"Accept"` in observed responses. The protocol is normalized with `.lower()`, but the policy is not.

## Environment

- Prowler Local Server / App release: `v5.35.0`
- API image: `prowlercloud/prowler-api:stable`
- Provider: Alibaba Cloud
- Authentication: RAM user credentials

## Reproduction

1. Create an Alibaba Cloud security group with this inbound rule:

```text
Direction: ingress
IpProtocol: TCP
PortRange: 22/22
SourceCidrIp: 0.0.0.0/0
Policy: Accept
```

2. Run a full Alibaba Cloud scan.
3. Inspect the result for `ecs_securitygroup_restrict_ssh_internet`.

### Actual result

The security group is reported as PASS:

```text
does not have SSH port 22 open to the internet
```

### Expected result

The security group should be reported as FAIL because TCP/22 is allowed from `0.0.0.0/0`.

The same issue affects the RDP check for TCP/3389.

## Root cause

Both checks contain this condition:

```python
if ingress_rule.get("policy", "accept") != "accept":
    continue
```

Files:

- `prowler/providers/alibabacloud/services/ecs/ecs_securitygroup_restrict_ssh_internet/ecs_securitygroup_restrict_ssh_internet.py`
- `prowler/providers/alibabacloud/services/ecs/ecs_securitygroup_restrict_rdp_internet/ecs_securitygroup_restrict_rdp_internet.py`

The ECS service preserves the SDK value when building `ingress_rules`:

```python
"policy": getattr(rule, "policy", "accept")
```

So `"Accept" != "accept"`, and an allowed rule is skipped before protocol, CIDR, and port evaluation.

## Suggested fix

Normalize the policy before comparison in both checks:

```python
if str(ingress_rule.get("policy", "accept")).lower() != "accept":
    continue
```

The existing unit tests currently use lowercase `"accept"`, so they do not reproduce the real API response. Updating the SSH and RDP test fixtures to use `"Accept"` would provide regression coverage.

## Impact

This produces false-negative findings for public SSH and RDP exposure in Alibaba Cloud and may affect every security group rule returned with the documented/observed `Accept` capitalization.

owenchenxy:
## Summary

- normalize Alibaba Cloud security group rule policies before evaluating SSH and RDP exposure
- cover the `Policy="Accept"` capitalization returned by `DescribeSecurityGroupAttribute`
- update SSH and RDP fixtures to match the real API response

## Problem

Alibaba Cloud returns allowed security group rules with `Policy="Accept"`. Both checks compared this value case-sensitively to lowercase `"accept"`, so valid allow rules were skipped and public TCP/22 and TCP/3389 exposure could be reported as PASS.

## Testing

```text
uv run pytest -q \
  tests/providers/alibabacloud/services/ecs/ecs_securitygroup_restrict_ssh_internet/ecs_securitygroup_restrict_ssh_internet_test.py \
  tests/providers/alibabacloud/services/ecs/ecs_securitygroup_restrict_rdp_internet/ecs_securitygroup_restrict_rdp_internet_test.py

4 passed
```

Black check passes for all four changed files, and `git diff --check` is clean.

Fixes #12048

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed Alibaba Cloud ECS security group checks for internet-exposed SSH and RDP so allowed rules are evaluated case-insensitively (e.g., `Policy="Accept"`).
  - Improved handling of missing or unexpected `Policy` values to avoid false negatives.

- **Tests**
  - Updated SSH and RDP security-group test cases to use capitalized `Policy` values and verify the corrected evaluation behavior.

### DIFF
--- a/prowler/changelog.d/alibabacloud-security-group-policy-case.fixed.md
+++ b/prowler/changelog.d/alibabacloud-security-group-policy-case.fixed.md
@@ -1,0 +1,1 @@
+Alibaba Cloud SSH and RDP security group checks no longer produce false negatives when allowed rules use capitalized `Policy="Accept"` values
--- a/prowler/providers/alibabacloud/services/ecs/ecs_securitygroup_restrict_rdp_internet/ecs_securitygroup_restrict_rdp_internet.py
+++ b/prowler/providers/alibabacloud/services/ecs/ecs_securitygroup_restrict_rdp_internet/ecs_securitygroup_restrict_rdp_internet.py
@@ -26,7 +26,7 @@ class ecs_securitygroup_restrict_rdp_internet(Check):
 
             for ingress_rule in security_group.ingress_rules:
                 # Check if rule allows traffic (policy == "accept")
-                if ingress_rule.get("policy", "accept") != "accept":
+                if str(ingress_rule.get("policy", "accept")).lower() != "accept":
                     continue
 
                 # Check protocol (tcp for RDP)
--- a/prowler/providers/alibabacloud/services/ecs/ecs_securitygroup_restrict_ssh_internet/ecs_securitygroup_restrict_ssh_internet.py
+++ b/prowler/providers/alibabacloud/services/ecs/ecs_securitygroup_restrict_ssh_internet/ecs_securitygroup_restrict_ssh_internet.py
@@ -26,7 +26,7 @@ class ecs_securitygroup_restrict_ssh_internet(Check):
 
             for ingress_rule in security_group.ingress_rules:
                 # Check if rule allows traffic (policy == "accept")
-                if ingress_rule.get("policy", "accept") != "accept":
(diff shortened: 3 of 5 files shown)


HugoPBrito (approved):

HugoPBrito:
Thanks @owenchenxy! 🚀

PR #12055, opened 2026-07-20
fix(app): correct the worker KEDA PostgreSQL scaler

bmbferreira:
### Context

The worker KEDA `ScaledObject` (`contrib/k8s/helm/prowler-app/templates/worker/scaled-object.yaml`) has two problems that make autoscaling unusable in common setups.

**1. The connection can't be configured (and is broken by default).**
The host is hardcoded to `{{ .Release.Name }}-postgresql` and the port is read from `.Values.postgresql.port`, which the chart never defines. So the port renders empty and KEDA rejects the trigger with:

> error parsing postgresql metadata: no port given

This happens even with the bundled database, and there is no way to point the scaler at an external PostgreSQL (`postgresql.enabled: false`).

**2. The query misses most real work.**
The default query only counts scans whose `scheduled_at` is within the last 2 hours:

```sql
SELECT COUNT(*) FROM scans WHERE ((state='executing' OR state='available' OR state='scheduled') and scheduled_at < NOW() and scheduled_at > NOW() - INTERVAL '2 hours')
```

That ignores manual scans (`scheduled_at IS NULL`), any backlog older than 2 hours, and scans already executing. This can cause the trigger go inactive mid-scan, KEDA scale the worker to zero, and scans get stuck in `executing` indefinitely.

### Description

- Expose the scaler's PostgreSQL connection under `worker.keda.postgresql` (`host`, `port`, `database`, `userName`, `passwordFromEnv`, `sslmode`), with defaults that preserve the current bundled-database behaviour (`host` → `<release>-postgresql`, `database` → `postgresql.auth.database`). Setting `worker.keda.postgresql.host`/`database` now supports an external DB.
- Fix the default `port` (was empty → now `5432`).
- Make `query` and `targetQueryValue` overridable.
- Replace the default query with one that counts real pending work: scans `executing`/`available` (bounded to `updated_at` within 24h so orphaned rows don't pin the worker up forever), plus `scheduled` scans that are due (no lower bound, so an overdue backlog still scales up).
- Drop the unused `keda.value` key — the template never referenced it and its "cpu or memory" comment never matched the `postgresql` trigger.

`state`, `scheduled_at` and `updated_at` all exist on the `scans` table (`api/src/backend/api/models.py`).

### Steps to review

1. Bundled DB:
    ```
    helm template prowler contrib/k8s/helm/prowler-app \
      --set worker.keda.enabled=true \
      --show-only templates/worker/scaled-object.yaml
    ```
    → `host: "prowler-postgresql"`, `port: "5432"`, `dbName: "prowler_db"`.
2. External DB — add:
    ```
    --set postgresql.enabled=false \
    --set worker.keda.postgresql.host=my-db.ns.svc.cluster.local \
    --set worker.keda.postgresql.database=prowler
    ```
    → `host`/`dbName` reflect the overrides.
3. `helm lint` passes.

### Checklist

- [x] `helm lint` passes; renders correctly for bundled and external Postgres.
- [x] Backwards compatible for the bundled DB (net effect unchanged, except the port which was previously broken).

### License

## Summary by CodeRabbit

* **New Features**
  * Enhanced worker autoscaling with configurable PostgreSQL connection and trigger settings.
  * Added support for custom SQL queries and target values to determine scaling.
  * Updated default scaling logic to account for active, overdue, and recently pending scan work, as well as other pending tasks.
  * Added configurable database port, SSL mode, credentials, query, and target settings through chart values.
  * Improved deployment flexibility by allowing scaling behavior to be tailored without modifying chart templates.

### DIFF
--- a/contrib/k8s/helm/prowler-app/values.yaml
+++ b/contrib/k8s/helm/prowler-app/values.yaml
@@ -422,10 +422,37 @@ worker:
     pollingInterval: 30
     # -- The cooldown period in seconds for scaling
     cooldownPeriod: 120
-    # -- The trigger type for scaling (cpu or memory)
+    # -- The KEDA scaler type. Only `postgresql` is supported by the default query below.
     triggerType: "postgresql"
-    # -- The target utilization percentage for the worker pods
-    value: "50"
+    # PostgreSQL connection used by the scaler query. The KEDA operator opens this
+    # connection from its own namespace, so `host` must resolve from there. The
+    # defaults target the bundled postgresql subchart; set them explicitly when
+    # using an external database (postgresql.enabled: false).
+    postgresql:
+      # -- Scaler database host. Defaults to the bundled "<release>-postgresql" service.
+      host: ""
+      # -- Scaler database port.
+      port: "5432"
+      # -- Scaler database name. Defaults to `postgresql.auth.database`.
+      database: ""
+      # -- User the scaler authenticates as.
+      userName: "postgres"
+      # -- Name of an env var on the worker container holding the password.
+      passwordFromEnv: "POSTGRES_ADMIN_PASSWORD"
+      # -- sslmode for the scaler connection.
+      sslmode: "disable"
+    # -- The scaler divides the query result by this value to get the desired replica count.
+    targetQueryValue: "1"
+    # -- Query the scaler runs to measure pending work. The default counts scans that
+    # represent real work: those executing or available (bounded to rows updated in the
+    # last 24h so orphaned rows do not pin the worker up), plus scheduled scans that are
+    # due (no lower bound, so an overdue backlog still scales up). This replaces the
+    # previous 2-hour scheduled-only window, which missed manual scans, older backlogs,
+    # and in-progress scans. Override to tune scaling for your workload.
+    query: >-
+      SELECT COUNT(*) FROM scans
+      WHERE (state IN ('executing', 'available') AND updated_at > NOW() - INTERVAL '24 hours')
+      OR (state = 'scheduled' AND scheduled_at < NOW())
 
 worker_beat:
   # This will set the replicaset count more information can be found here: https://kubernetes.io/docs/concepts/workloads/controllers/replicaset/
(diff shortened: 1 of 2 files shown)


contrib/k8s/helm/prowler-app/values.yaml
```diff
@@ -422,10 +422,37 @@ worker:
...
+    # -- Query the scaler runs to measure pending work. The default counts scans that
+    # represent real work: those executing or available (bounded to rows updated in the
+    # last 24h so orphaned rows do not pin the worker up), plus scheduled scans that are
+    # due (no lower bound, so an overdue backlog still scales up). This replaces the
+    # previous 2-hour scheduled-only window, which missed manual scans, older backlogs,
+    # and in-progress scans. Override to tune scaling for your workload.
+    query: >-
+      SELECT COUNT(*) FROM scans
+      WHERE (state IN ('executing', 'available') AND updated_at > NOW() - INTERVAL '24 hours')
+      OR (state = 'scheduled' AND scheduled_at < NOW())
```
Utwo: Something like this is working on our version. It is important that besides `scans` we also cover `django_celery_results_taskresult`
```
SELECT
            (SELECT COUNT(*) FROM scans
              WHERE (state IN ('executing', 'available') AND updated_at > NOW() - INTERVAL '24 hours')
                 OR (state = 'scheduled' AND scheduled_at < NOW()))
          + (SELECT COUNT(*) FROM django_celery_results_taskresult
              WHERE task_name IN ('scan-perform', 'scan-perform-scheduled')
                AND status IN ('PENDING', 'RECEIVED', 'STARTED')
                AND date_created > NOW() - INTERVAL '48 hours')
          + (SELECT COUNT(*) FROM django_celery_results_taskresult
              WHERE task_name NOT IN ('scan-perform', 'scan-perform-scheduled')
                AND status IN ('PENDING', 'RECEIVED', 'STARTED')
                AND date_created > NOW() - INTERVAL '1 hour')
```

danibarranqueroo:
@bmbferreira could you review @Utwo comment? Thanks! 🙌

danibarranqueroo (approved):

PR #12057, opened 2026-07-20
fix(ui): redirect expired sessions to sign-in

Alan-TheGentleman:
### Context

Restored sessions can be missing an access token or contain `RefreshAccessTokenError`. Protected UI requests previously continued into authenticated API calls in that state, which could send `Authorization: Bearer undefined` instead of returning the user to sign-in.

Review feedback surfaced a second, related failure: signing the user out on `RefreshAccessTokenError` is wrong when the refresh token itself is still valid. The API rotates refresh tokens and blacklists the previous one (`ROTATE_REFRESH_TOKENS` and `BLACKLIST_AFTER_ROTATION` in `api/src/backend/config/django/base.py`), while NextAuth can only persist the resulting cookie where `Set-Cookie` is allowed — the proxy, Route Handlers and Server Actions, never a Server Component render. A refresh performed during a render therefore consumed the stored refresh token without saving its replacement, so the next request replayed a blacklisted token and a still-valid session was terminated.

This PR intentionally has no linked issue, as requested. Distributed refresh across multiple replicas and arbitrary later backend token revocation remain outside this scope; see the note under Follow-up.

### Description

Expired-session redirects:

- Redirect failed or missing restored sessions to `/sign-in` before authenticated API requests are made.
- Preserve the original protected page path and query string in the `callbackUrl` sign-in parameter.
- Build authenticated headers only when a valid access token is available, preventing `Bearer undefined`.
- Re-throw Next.js navigation control-flow errors so redirects are not converted into ordinary action errors.
- Keep generic backend `401` and `403` responses as typed ordinary API errors; they do not trigger this redirect path.

Rotated refresh tokens:

- Cache the rotated credential pair keyed by the refresh token that produced it, so a request still carrying the stale cookie reuses the replacement instead of replaying a blacklisted token.
- Keep refresh failures out of the cache, so a transient error still retries against the API.
- Cache only the credential pair rather than the whole token, so no user state leaks between requests.

Coverage and changelog:

- Add focused coverage for authentication headers, redirect propagation, callback preservation, backend authorization errors, rotated-pair reuse, genuine rejection, and transient-failure retry.
- Add `ui/changelog.d/expired-session-redirect.fixed.md` and `ui/changelog.d/rotated-refresh-token-reuse.fixed.md`.

No dependencies are added or changed.

### Refresh flow, before and after

```mermaid
flowchart TD
    subgraph before["Before: the rotated pair is lost"]
        direction TB
        B1["Request N, Server Component render<br/>cookie = A1 / R1, A1 expired"]
        B2["jwt(): POST /tokens/refresh with R1"]
        B3["API returns A2 / R2<br/>and blacklists R1"]
        B4["Render uses A2, request succeeds"]
        B5["Set-Cookie is not possible in a render<br/>cookie stays A1 / R1, so A2 / R2 are lost"]
        B6["Request N+1, cookie = A1 / R1"]
        B7["jwt(): POST /tokens/refresh with R1"]
        B8["API returns 401, token is blacklisted"]
        B9["RefreshAccessTokenError<br/>sign-out on a session that was still valid"]
        B1 --> B2 --> B3 --> B4 --> B5 --> B6 --> B7 --> B8 --> B9
    end

    subgraph after["After: the rotated pair survives the render"]
        direction TB
        A1["Request N, Server Component render<br/>cookie = A1 / R1, A1 expired"]
        A2["jwt(): no cached pair for R1<br/>POST /tokens/refresh with R1"]
        A3["API returns A2 / R2<br/>and blacklists R1"]
        A4["Cache the pair under R1 for 60s"]
        A5["Render uses A2, request succeeds<br/>cookie still A1 / R1"]
        A6["Request N+1, cookie = A1 / R1"]
        A7["jwt(): cached pair found for R1<br/>returns A2 / R2 without calling the API"]
        A8["Session stays valid, no blacklisted replay"]
        A9["Proxy, Route Handler or Server Action<br/>persists A2 / R2, cache no longer needed"]
        A1 --> A2 --> A3 --> A4 --> A5 --> A6 --> A7 --> A8 --> A9
    end
```

### Follow-up

The cache is per process, so it closes the window within an instance but not across replicas. The durable fix is to make refresh a single-writer operation owned by `ui/proxy.ts`, the only context guaranteed to persist the cookie. That requires extending the proxy matcher to cover `/api` (Route Handlers are excluded today), rewriting the inbound request cookie so the same render sees the new access token, and removing refresh from the `jwt` callback. Those three pieces touch the whole authentication path and belong in their own PR.

### Steps to review

1. Review `ui/lib/server-auth.ts` and confirm a missing access token or `RefreshAccessTokenError` redirects before authenticated headers are returned.
2. Review `ui/proxy.ts` and confirm protected request paths include their original query string in the internal current-path header.
3. Review `ui/lib/server-actions-helper.ts` and confirm Next.js redirect control flow is re-thrown while backend `401` and `403` responses remain typed API errors.
4. Review `ui/auth.config.ts` and confirm the rotated pair is cached under the refresh token that produced it, that failures are not cached, and that only credentials are cached.
5. Review the focused tests covering these boundaries.
6. Confirm the following completed validation evidence for the current candidate:
   - Focused `ui/auth.config.test.ts`: 6/6 passed.
   - Focused authentication suites (`auth.config`, `server-auth`, `auth-headers`, scan report route): 23/23 passed.
   - Full UI unit suite: 2,450/2,450 passed across 378 files.
   - TypeScript validation passed.
   - ESLint passed (`ui/auth.config.ts` is inside the repository ESLint ignore set).
   - Prettier validation passed.
   - Changelog fragments validated.

   The earlier 4R bounded review and pre-PR gate for lineage `review-b568499a736da13d` covered the previous candidate; the rotated-refresh-token commit is newer and carries the validation evidence listed above.

### Checklist

<details>

- [ ] This feature/issue is listed in the [open issues](https://github.com/prowler-cloud/prowler/issues?q=sort%3Aupdated-desc+is%3Aissue+is%3Aopen) or roadmap.prowler.com. No issue is linked by explicit request.
- [ ] Is it assigned to me, if not, request it via the [open issues](https://github.com/prowler-cloud/prowler/issues?q=sort%3Aupdated-desc+is%3Aissue+is%3Aopen) or [Prowler Community Slack](https://goto.prowler.com/slack). No issue is linked by explicit request.
- [x] I have reviewed the [open pull requests](https://github.com/prowler-cloud/prowler/pulls?q=sort%3Aupdated-desc+is%3Apr+is%3Aopen) and confirmed there is no existing PR that implements the same outcome.

</details>

- [x] Review if code is being documented following this specification https://github.com/google/styleguide/blob/gh-pages/pyguide.md#38-comments-and-docstrings. No Python documentation changes are applicable.
- [x] Review if backport is needed. Maintainers can determine release-branch backporting after review.
- [x] Review if is needed to change the [README.md](https://github.com/prowler-cloud/prowler/blob/master/README.md). No README change is needed.
- [x] Ensure a changelog fragment is added under [prowler/changelog.d/](https://github.com/prowler-cloud/prowler/tree/master/prowler/changelog.d), if applicable. SDK/CLI is not affected.

#### SDK/CLI
    - If so, do we need to update permissions for the provider? Not applicable.

#### UI
- [x] All issue/task requirements work as expected on the UI.
- [x] If this PR adds or updates npm dependencies, include package-health evidence (maintenance, popularity, known vulnerabilities, license, release age) and explain why existing/native alternatives are insufficient. No npm dependencies are added or updated.
- [x] Screenshots/Video of the functionality flow (if applicable) - Mobile (X < 640px). Not applicable; this is server-side session behavior with no visual layout change.
- [x] Screenshots/Video of the functionality flow (if applicable) - Table (640px > X < 1024px). Not applicable; this is server-side session behavior with no visual layout change.
- [x] Screenshots/Video of the functionality flow (if applicable) - Desktop (X > 1024px). Not applicable; this is server-side session behavior with no visual layout change.
- [x] Ensure a changelog fragment is added under [ui/changelog.d/](https://github.com/prowler-cloud/prowler/tree/master/ui/changelog.d), if applicable. Added `ui/changelog.d/expired-session-redirect.fixed.md` and `ui/changelog.d/rotated-refresh-token-reuse.fixed.md`.

#### API
- [x] All issue/task requirements work as expected on the API. Not applicable; API code is unchanged.
- [x] Endpoint response output (if applicable). Not applicable.
- [x] EXPLAIN ANALYZE output for new/modified queries or indexes (if applicable). Not applicable.
- [x] Performance test results (if applicable). Not applicable.
- [x] Any other relevant evidence of the implementation (if applicable). Not applicable.
- [x] Verify if API specs need to be regenerated. Not applicable.
- [x] Check if version updates are required (e.g., specs, uv, etc.). Not applicable.
- [x] Ensure a changelog fragment is added under [api/changelog.d/](https://github.com/prowler-cloud/prowler/tree/master/api/changelog.d), if applicable. Not applicable.

#### MCP Server
- [x] All issue/task requirements work as expected on the MCP Server. Not applicable; MCP Server code is unchanged.
- [x] Ensure a changelog fragment is added under [mcp_server/changelog.d/](https://github.com/prowler-cloud/prowler/tree/master/mcp_server/changelog.d), if applicable. Not applicable.

### License


alejandrobailo (changes requested):
Thanks for working on this. The problem is valid, and I agree that we must never send `Bearer undefined` or continue with an unusable session.

I’m requesting changes because I think we are solving the problem at the wrong layer.

The API should remain the source of truth for token validity. We already call `POST /tokens/refresh` from the Auth.js JWT callback, so the expected flow should be:

1. Send the request with the current access token.
2. If the API reports an expired or invalid access token, request a refresh.
3. Update the server-side session and retry the original request once.
4. Redirect to sign-in only when the refresh token is expired, revoked or missing.

A `403` should never trigger refresh. We should also distinguish a terminal refresh failure (`401`) from a temporary API or network failure (`5xx`/timeout). Redirecting users to sign-in during a temporary backend outage would not fix anything and would unnecessarily discard their session.

I would centralize this in an authenticated server-side fetch wrapper rather than coupling redirect behavior to `getAuthHeaders()`. Header construction should be fail-closed and never return an authorization header without a token, but it should not own navigation, retry policy and API error handling.

Page and Server Action boundaries can translate a terminal authentication error into a sign-in redirect. Route Handlers can return `401`. This keeps the same authentication policy reusable without forcing every consumer into `next/navigation`.

There are also two concrete blockers in the current implementation:

- We are on Next.js 16.2.9. Importing `server-auth.ts`, which depends on `next/headers`, through the mixed `lib/helper.ts` module breaks `next build`. This is why UI tests, the container build and E2E are currently red.
- `getAllProviderGroups()` calls `getAuthHeaders()` inside a `try` whose catch returns `undefined`, so it swallows the Next.js redirect error. Adding `unstable_rethrow()` only to `handleApiError()` does not cover this path.

If we want to keep this PR narrowly scoped, I would at least:

- Validate `user + accessToken + no session error` in the existing Auth.js/Proxy boundary.
- Move authentication-header construction into a dedicated server-only module, imported directly by server actions and never exported through `@/lib`.
- Keep `unstable_rethrow()` for framework-controlled errors and fix the uncovered catch.
- Add coverage for API `401 → refresh → retry`, terminal refresh failure and transient refresh failure.
- Require `pnpm build` to pass before re-review.

The intended behavior makes sense, but I think an API-driven refresh flow with a small server-side authentication boundary gives us a safer and more complete solution than redirecting from a shared UI helper.

Alan-TheGentleman:
Both concrete blockers are fixed in `b0b3ecec`. `getAuthHeaders` now lives in a dedicated `server-only` module imported directly by all 41 server callers, and `pnpm build`, all 42 browser tests, and all 2,152 unit tests pass. `getAllProviderGroups` now resolves authentication before its broad `try`, with a regression test proving the exact redirect control-flow error propagates.

I do not think `API 401 -> refresh -> retry` is safe to add in this PR. `auth()` already refreshes proactively from `accessTokenExpires` before headers are constructed. More importantly, refresh tokens rotate and blacklist the consumed token, while zero-argument Auth.js `auth()` discards the response `Set-Cookie`. A fetch wrapper cannot atomically persist the rotated refresh token into the encrypted session cookie and would leave the session holding a stale blacklisted token. Generic API `401` and `403` responses therefore remain typed API errors, while missing or terminal session state fails closed at the authentication boundary.

You're right that transient refresh failures are currently conflated with terminal refresh failures. That behavior predates this PR and deserves separate auth design work. This PR intentionally does not claim to solve it.

ui/lib/server-auth.ts
```diff
@@ -0,0 +1,39 @@
...
+};
+
+type SessionError = "RefreshAccessTokenError" | (string & {});
+
+export const getRequiredAuthHeaders = async (
+  accessToken: string | undefined,
+  options?: { contentType?: boolean },
+  sessionError?: SessionError,
+) => {
+  if (!accessToken || sessionError === "RefreshAccessTokenError") {
```
pfe-nazaries, line 25: **`getRequiredAuthHeaders` has no non-redirect mode, and that silently breaks the scan-report download.**

`redirect()` is the right failure mode for Server Actions and Server Components. But Route Handlers under `app/api/**` are consumed by `fetch`, and a 307 to an HTML page isn't something a `fetch` consumer can act on — they need a 401. This helper can only do the former, and `app/api/scans/[scanId]/report/route.ts` uses it anyway.

Concrete chain:

1. `route.ts:67` calls `getAuthHeaders()` **outside** the `try` (which starts at `:74`), so `NEXT_REDIRECT` escapes `GET`.
2. Next 16.2.9 converts it into `Response(null, { status: 307, Location: /sign-in?callbackUrl=/ })` (`next/dist/server/route-modules/app-route/module.js:447-469`).
3. `downloadScanZip` (`lib/helper.ts:127`) preflights with the **default** `redirect: "follow"` — note the contrast with the upstream fetch at `route.ts:81`, which correctly sets `redirect: "manual"`.
4. `/sign-in` is a public route, so the preflight lands on **HTML 200**. `status === 202` is false, and `!preflightResponse.ok` is false. Neither guard fires.
5. It falls through to `helper.ts:156-166` → `<a download="scan-<id>-report.zip">` → that request 307s as well → **the browser saves the sign-in page under the `.zip` filename**, behind a `"Download Started"` success toast.

The `isHtmlResponse` guard at `route.ts:125` was written for exactly this shape of bug, but it only runs inside the `!upstreamResponse.ok` branch — on this path the upstream fetch never happens, so it can't fire.

This is a regression against `master`: there, `Bearer undefined` → upstream 401 → `!ok` → `"Download Failed"`. The PR turns an honest error into a corrupted file that looks like success.

**Suggested fix at the source rather than per-route:** give this helper a second mode (e.g. a `requireAuthHeaders()` variant that returns/throws a 401) and use it from `app/api/**`, so a future route handler can't reintroduce the same bug. There's already precedent in the repo — `app/api/lighthouse/analyst/route.ts:43-45` returns 401. This is also the shape @alejandrobailo asked for in the open `CHANGES_REQUESTED`.

Two secondary notes on the same path:

- `proxy.ts:82` excludes `api` from the matcher, so `x-prowler-current-path` is never set for this route and `callbackUrl` is always the `/` fallback (line 12 above). That fallback branch is currently untested — `server-auth.test.ts:25-27` always seeds the header.
- All 7 tests in `route.test.ts` stub `getAuthHeaders`, so nothing catches this today.

pfe-nazaries (changes requested):
Blocking until fixes applied

ui/lib/server-auth.ts
Alan-TheGentleman: Fixed in `627d3b6199`. The scan-report Route Handler now uses a non-redirecting auth path and returns a non-cacheable `401` before any upstream fetch, while Server Actions and Server Components keep their sign-in redirect behavior. Regression coverage now verifies the `401` response has no `Location`, skips the upstream request, and uses `/` when the callback path header is unavailable; focused tests, typecheck, ESLint, Prettier, and `pnpm build` pass.

pfe-nazaries (dismissed):
LGTM

pfe-nazaries (dismissed):

alejandrobailo:
I'd suggest closing this PR.

PR #12104, opened 2026-07-23
test(ui): AWS Organizations onboarding baseline integration suite

pfe-nazaries:
### Context

Phase 0 of the **GCP Organization Onboarding UI** work, delivered as the first of three chained PRs stacked on the `feature/gcp-organization-ui` integration branch:

1. **Phase 0 (this PR)** — AWS baseline integration suite (refactor validator). No external dependency.
2. Phase 1 — canonical `/organization-nodes` contract + lifecycle generalization (merge-gated on the API canonical release).
3. Phase 2 — GCP organization flow + lifecycle safety.

The suite locks down current AWS Organizations behavior before the Phase 1 refactor touches the data layer, so any regression there shows up as a red baseline test with untouched assertions.

### Description

Test-infrastructure only — no runtime/user-facing change.

- Renames the browser-mode test suffix `.browser.test` → `.integration.test` (vitest globs, setup file, existing attack-paths specs, scripts); runs `babel-plugin-react-compiler` in the test build and mocks `next/cache` so server actions run under Vitest Browser Mode.
- Adds dual-contract MSW handlers for the organizations onboarding API: both the deprecated `/organizational-units` routes and the canonical `/organization-nodes` contract over shared fixtures, mirroring the transition-window API (design decision D13).
- Adds a page-level providers-page harness (mounts the client view + MSW, semantic wizard/table methods).
- Adds the AWS Organizations onboarding baseline suite: onboarding happy path, blocked accounts, retry-failures-only, selection-change re-apply, skip; providers-page grouping, edit-name, update-credentials re-entry, and current (fire-and-forget) deletion.

### Steps to review

```bash
cd ui && pnpm install && pnpm test:integration
```

Full integration project green locally: 51 tests (attack-paths 42 + providers-page baseline 9); `tsc` 0 errors; ESLint/Prettier clean.

### Checklist

- [x] Covered by tests (this PR **is** the tests).
- No user-facing UI change → no mobile/tablet/desktop screenshots.
- Changelog: intentionally omitted here (`no-changelog`); the changelog entry lands once on the `feature/gcp-organization-ui` principal branch, where the three chained PRs are merged.

### License

## Summary by CodeRabbit

* **Testing**
  * Expanded integration coverage for provider onboarding, account selection, connection retries, organization management, and attack-path interactions.
  * Added reusable browser testing utilities and realistic organization scenarios.
  * Improved synchronization and assertions for graph updates, highlighting, and viewport changes.
* **Chores**
  * Replaced browser test commands with integration test commands.
  * Updated automated UI checks to run integration tests and refreshed Vitest configuration.
  * Improved test stability and excluded generated test attachments from version control.


pfe-nazaries:

Reviewed 2 unresolved CodeRabbit feedback item(s) and applied no new code changes in this run — both were already addressed in commit `61880f62b`, pushed before this review pass.

**Verified against current `HEAD`:**

- `ui/__tests__/browser-harness.ts` — the `byRoleName` implicit-button fallback now gates on `role === "button"` and `!el.hasAttribute("role")`, so a plain `<button>` is matched and the fallback is no longer dead code.
- `ui/__tests__/msw/handlers/organizations.fixtures.ts` — `FixtureAccountProviderMapping` and `FixtureApplyError` are extracted as dedicated interfaces and referenced from `FixtureApplyOutcome`, matching the repository interface guidelines.

A third thread (unescaped `RegExp` in `showsParameterLabel`) was already marked outdated and is likewise fixed on this branch.

Both threads have been resolved. Validation on the current branch: `tsc` clean, ESLint clean, Prettier clean, unit 2260/2260, integration 52/52.

ui/app/(prowler)/providers/providers-page.harness.tsx
```diff
@@ -0,0 +1,569 @@
...
+      accessToken: "test-access-token",
+      expires: "2999-01-01T00:00:00Z",
+    };
+
+    // Single source of truth: the runtime-config island (seeded by the
+    // `seedRuntimeConfig` fixture) drives both `isCloud()` deep in the tree and
+    // the top-level prop, exactly as `providers/page.tsx` derives it in prod.
+    const cloud = isCloud();
+    const providers = this.buildProviderProps();
+    const rows = this.buildTableRows(cloud);
```
alejandrobailo: [P1] Exercise the production page-data path

This bypasses `loadProvidersAccountsViewData` and builds the rows directly from fixture data. Consequently, the providers-page baseline never requests either `/organizational-units` or `/organization-nodes`. Phase 1 could leave the production loader on the deprecated endpoint, or return no hierarchy at all—while this suite remains green because the harness is updated in parallel. Please build the mounted props through the real page-data loader/adapter, or add integration coverage that asserts the expected hierarchy request and rendered result. The transition-route tripwire needs to observe an actual request.

ui/__tests__/browser-harness.ts
```diff
@@ -0,0 +1,166 @@
...
+
+  constructor(readonly fixture: TFixture) {}
+
+  // --- Request tracking (opt-in) ------------------------------------------
+
+  /** Start recording MSW requests into `requestLog`. Call once, after mounting. */
+  protected trackRequests(worker: SetupWorker): void {
+    // Clear only our own `request:start` listeners from a prior harness on the
+    // shared worker — not every listener on the public event emitter.
+    worker.events.removeAllListeners("request:start");
```
alejandrobailo: [P2] Do not clear listeners owned by the shared worker

`removeAllListeners("request:start")` removes every public listener for this event, not only a previous listener owned by this harness. Another harness or diagnostic listener will silently stop recording requests as soon as this mounts. Please keep the callback reference and unregister only that callback with `removeListener` during harness/fixture cleanup instead of clearing the shared emitter.

ui/app/(prowler)/providers/providers-page.integration.test.tsx
```diff
@@ -0,0 +1,184 @@
...
+  await harness.waitForAccountSelection();
+}
+
+describe("AWS Organizations onboarding (baseline)", () => {
+  it("completes the happy path: setup → discovery → selection → apply → connect → launch", async () => {
+    const harness = new ProvidersPageHarness(awsOnboardingFixture());
+    await onboardToSelection(harness);
+
+    await harness.testConnections();
+    await harness.waitForAccountsConnected();
```
alejandrobailo, line 49: [P1] Complete the launch path claimed by this test

The test stops after reaching the launch screen; it never invokes `Launch scan`/`Save and launch scan`, and it never asserts a `/schedules/bulk` or `/scans` request. The current `/schedules/bulk` handler returns an empty data array, which the real advanced launch flow treats as an error, yet this test still passes. Please drive the final action and assert the successful request plus completion behavior, or narrow the test/PR claim and remove the unsupported launch handlers.

alejandrobailo (changes requested):
The new suites are green locally, but I found three gaps that need to be fixed before Phase 0 can act as the regression validator for Phase 1. Details inline.

ui/app/(prowler)/providers/providers-page.harness.tsx
pfe-nazaries: Fixed in `062a7e718`. Extracted production's `ProvidersTabContent` to its own module; the harness renders it and wires no props, so the hierarchy goes over HTTP. Tests assert it's requested (and not, on-prem).

Caveat: the count covers both routes, so it won't catch a loader that stays on the deprecated one — pinning the route would break when Phase 1 migrates. That half is the `serveDeprecatedRoutes: false` tripwire (2.10); say the word and I'll wire it here.

ui/__tests__/browser-harness.ts
pfe-nazaries: Fixed in `60d923d10`. Keeps the callback reference and removes only it via `removeListener`, in an `onTestFinished` hook. No blanket clear, no test changes.

ui/app/(prowler)/providers/providers-page.integration.test.tsx
pfe-nazaries: Fixed in `60d923d10` — the test drives the launch now.

`updateSchedulesBulk` validates ids with `z.uuid()` before fetching, so `provider-aws-1` failed earlier than the empty-`data` path and `/schedules/bulk` was never issued. Fixture ids are UUIDs now. Asserts one `/schedules/bulk` + one `/scans` per provider; verified it fails if the handler returns `updated: []`.

alejandrobailo (approved):

PR #12114, opened 2026-07-24
feat(ui): GCP org onboarding — Phase 1: canonical contract + lifecycle generalization

pfe-nazaries:
## Phase 1 — canonical organization contract + lifecycle generalization

Second PR in the GCP Organization Onboarding UI chain. **Stacked on #12104 (Phase 0)** — base branch is `feature/gcp-organization-ui-phase0`; review/merge #12104 first.

Migrates the organizations data layer to the canonical `organization-nodes` contract and generalizes the onboarding lifecycle by organization type. **AWS behavior is visually unchanged**; GCP support arrives in Phase 2.

### What changes
- **Types** — normalized `OrgHierarchy` (nodes + candidates, kind-discriminated), per-provider wire interfaces, secret/apply payload unions, canonical relation/counter names; dead enum values dropped.
- **Data layer** — parameterized `org_type`, canonical `/organization-nodes` routes, per-type secret payload builders, discriminated `applyDiscovery`, degraded-fetch signal.
- **Store** — persists the normalized hierarchy, candidate naming, discovery-resume fields (`discoveryId` + status at trigger), provider-agnostic tree rebuild, persist-version bump.
- **Rendering** — org-type/kind-driven tree + table labels (no ID-prefix detection); providers page consumes canonical nodes for both org types (drops the hardcoded `filter[org_type]=aws`); non-blocking degraded-hierarchy notice.

### Testing (integration-first)
- Phase 0 AWS baseline suite stays green with **untouched assertions**.
- New integration coverage: mixed AWS+GCP hierarchy display, degraded-view notice.
- Canonical tripwire: MSW handlers stop serving the deprecated `/organizational-units` routes; suite stays green under `onUnhandledRequest: "error"`, proving the UI is fully canonical.

### Notes
- No user-visible change in Phase 1 → `no-changelog` (the feature's changelog fragment lands once on the principal branch).
- Merge-gated on the API canonical release being deployed (reads target `/organization-nodes`).

Spec: `openspec/changes/gcp-organization-onboarding-ui` (tasks 2.1–2.11).

## Summary by CodeRabbit

* **New Features**
  * Added support for displaying provider hierarchies across multiple organization types and cloud providers.
  * Organization and node labels now adapt to the provider and hierarchy level.
  * Onboarding selection now supports grouped candidates, aliases, and provider-specific discovery workflows.
  * Added organization-node management, including dynamic naming and deletion actions.

* **Bug Fixes**
  * Providers remain visible in a flat list when hierarchy data is temporarily unavailable, with a clear notice.
  * Improved provider grouping and de-duplication across nested hierarchy levels.
  * Updated organization actions and credential options based on supported onboarding flows.


ui/components/providers/organizations/hooks/org-setup-strategy.ts
```diff
@@ -0,0 +1,98 @@
...
+
+/**
+ * Values collected by an organization setup form. AWS-shaped today; GCP fields
+ * are added when the GCP flow lands (Phase 2). The active strategy reads only
+ * the fields relevant to its organization type.
+ */
+export interface OrgSetupSubmissionData {
+  organizationName?: string;
+  // AWS
+  awsOrgId?: string;
```
Alan-TheGentleman: shouldn't be better to leave it generic and then scope it? so we support others in the near future?

ui/actions/organizations/organizations.ts
```diff
@@ -164,67 +176,42 @@ export const listOrganizationsByExternalId = async (externalId: string) => {
...
+ * GET /api/v1/organizations
  */
-export const listOrganizations = async () => {
-  const headers = await getAuthHeaders({ contentType: false });
+export const listOrganizationsSafe = async (): Promise<
+  CollectionFetch<OrganizationResource>
+> => {
   const url = new URL(`${apiBaseUrl}/organizations`);
-  url.searchParams.set("filter[org_type]", "aws");
+  url.searchParams.set("page[size]", "100");
```
Alan-TheGentleman: **This only fetches the first 100 organizations.** Removing the AWS-only filter broadens this to the combined AWS and GCP collection, but `fetchOptionalCollection` never follows JSON:API pagination, so larger tenants silently lose groups from the providers page. Please traverse every page here and for `organization-nodes`, or use a dedicated unpaginated hierarchy endpoint.

ui/store/organizations/store.ts
```diff
@@ -171,32 +191,34 @@ export const useOrgSetupStore = create<OrgSetupState>()(
     }),
     {
       name: "org-setup-store",
+      version: 1,
```
Alan-TheGentleman, line 201: **This version bump drops existing persisted onboarding state.** Zustand treats old snapshots as version 0 and ignores them when the current version is 1 without a `migrate` function. Please add a v0-to-v1 migration, or document and intentionally clear the state if restarting interrupted onboarding is the desired behavior.

ui/components/providers/organizations/hooks/org-setup-strategy.ts
pfe-nazaries: Done in `94ef09329`. The shared shape is generic and each provider scopes it in its own arm: `BaseOrgSetupData`, then `AwsOrgSetupData extends` it with a required `orgType` tag, and `OrgSetupSubmissionData` is the union — Phase 2 adds `| GcpOrgSetupData`. No more AWS optional fields in a shared bag, so a strategy can't read another type's fields, and the registry is keyed by `OrgFlowType` — a new provider is one arm plus one entry, and every AWS-assuming consumer becomes a compile error the moment the second arm lands.

ui/actions/organizations/organizations.ts
pfe-nazaries: Fixed in `246a21d97`. `fetchOptionalCollection` now follows `meta.pagination` (`page[size]=100`, 50-page runaway guard), and both hierarchy reads go through it — organizations and `organization-nodes`. Any incompleteness (failed page or the guard) resolves to the degraded result, so the "grouping unavailable" notice never presents a partial hierarchy as complete. The MSW list routes paginate for real now, so a caller that stops at page one fails the suite. The unpaginated-endpoint alternative is still worth having: I'll raise a larger `page[size]` cap or a dedicated hierarchy endpoint with the API team while the contract is open.

ui/store/organizations/store.ts
pfe-nazaries: Took the second option — documented at `store.ts:194`. This is per-tab `sessionStorage` holding only wizard progress, and the persisted shape changed structurally (normalized hierarchy, organization type, discovery-resume fields), so discarding the snapshot resets an onboarding session that was in flight across the release instead of migrating a shape we removed. It's also called out in the design's risk list.

pfe-nazaries:
## Review round addressed

All 9 review threads worked through — 8 fixed, 1 deferred with a reason. Each fix was then re-verified independently, which turned up gaps in three of them; those are fixed too. Two commits: `754bc8ec` (the fixes) and `33f3264` (the gaps).

### Fixed

**Unknown `kind` bypasses the neutral fallback** (`lib/organizations.ts`) — confirmed reachable and crashing. First fix was incomplete: `NODE_KIND_LABEL` is an object literal, so a wire value colliding with `Object.prototype` (`"constructor"`, `"toString"`) resolves to an inherited *function*, `??` never fires, and `delete-organization-form.tsx:41` still throws `.toLowerCase is not a function`. Now narrows through the `toNodeKind` this module already exported; `terminologyFor` switched to a membership check, which covers `getNameSourceLabel` and `getCandidateNoun` as well. Tested across five prototype keys.

**Cyclic ancestor walk** (`organizations.adapter.ts`) — real, with a per-candidate `visited` set. Demonstrated before fixing: the test process stays alive past 60s, since a synchronous loop is something Vitest cannot interrupt. Worth noting the guard must stay per-candidate, not shared, or the pre-existing dedup test loses its meaning.

**Unvalidated `orgType`** (`organizations.ts`) — added `toOrgFlowType`, mirroring what `toNodeKind` does for node kinds. Rejects unrecognized values instead of coercing them; only an *absent* value still defaults to AWS. `azure` is rejected too: it is a real `OrganizationType` with no onboarding flow, so it was the value a blind cast let through.

**Swallowed errors in `fetchOptionalCollection`** — no bespoke Sentry call needed: the `!response.ok` early return was skipping `handleApiResponse`, the reporting point every other organization action already goes through. Routing failures through it was only half the problem, though — `getAuthHeaders` sat outside the `try`, so an expired session escaped into the providers page's `Promise.all`, which has no catch, instead of degrading; and the bare `catch` reported nothing, so a rejected fetch and the page-guard exhaustion degraded silently. Both fixed. The `hasActionError` branch first added here was dead — only a non-OK response yields an error envelope, and that is already caught — so it was removed rather than left as decoration.

**Degraded-hierarchy copy overstating the fallback** — the copy is fixed, but the diagnosis behind the suggestion does not hold. The premise was that providers still render grouped by organization when only the nodes read fails. They do not: AWS accounts hang off the OUs, so `organizations.relationships.providers` is empty, the organization row has nothing to group, and it drops out along with the nodes. A new page-level integration test proves it against the real page. So rather than split the status in two, the single notice is reworded to something true for both shapes — "Organization grouping is incomplete. Some providers may appear ungrouped." The MSW handlers gained a nodes-only failure mode, so the partial case has coverage at the page level for the first time.

**`candidateId`/`providerId` rename** and **extracted relationship interfaces** — both done. The extraction went further than the two named interfaces: `ApplyDiscoveryPayload` is live and had the same `Array<{…}>` shapes, whereas `ApplyResultRelationships` turns out to be exported but imported nowhere. Splitting the collection `meta` per field was also a modelling error on my part — both list endpoints serve `version` and `pagination` in the same object, so it is one `CollectionMeta`, which lets the paginated read drop its anonymous cast.

### Deferred

**Duplicated relationship-vs-fallback resolution** (`providers-page.utils.ts` 316-328 / 434-452) — not taken, on an already-large PR. The extraction is less clean than it looks: the second block filters asymmetrically (`providersInNodes` on both branches, `providersAssignedToNode` only on the relationship branch), so a shared helper needs a filter callback. The real hazard is adjacent and worth recording: block 2's `directProviderIds` uses the *unfiltered* relationship ids while its rows are filtered, which `dedupeIds` currently masks.

### On the tests

Two of the tests written for this round passed against the unfixed code, so they certified nothing — one because the fixture's organization had no providers to group, one because it named a branch that `!response.ok` short-circuits before reaching. Both rewritten, and every new test was checked to fail without its fix. Two existing assertions changed deliberately: they pinned `handleApiResponse` *not* being called on failure, which is exactly what the observability fix inverts, and their 5xx mocks now throw so they exercise the path production takes.

Verified locally on the rebased branch, since PRs targeting a non-default base run no CI here: tsc 0, unit 2442/2442, integration 58/58, eslint 0, prettier clean.

ui/actions/organizations/organizations.ts
```diff
@@ -41,26 +48,64 @@ function hasActionError(result: unknown): result is { error: unknown } {
...
+      // "maybe more" would walk to the guard on every single-page response.
+      if (page >= (body.meta?.pagination?.pages ?? 1)) {
+        return { data: collected };
+      }
     }
 
-    return (await handleApiResponse(response)) as T;
-  } catch {
-    return { data: [] } as unknown as T;
+    return { data: [], error: true };
```
Alan-TheGentleman, line 108: The 50-page guard still returns the degraded result without reporting why it fired. That makes estates above 5,000 hierarchy resources indistinguishable from ordinary fetch failures in Sentry, even though the latest observability fix says guard exhaustion is captured. Please report a concrete guard-exhaustion error through `handleApiError` before returning `{ data: [], error: true }`.

pfe-nazaries:

Reviewed 8 unresolved CodeRabbit review threads against the current branch state.

**7 were already addressed** by commits pushed after the review ran (the threads were not marked outdated):

| Thread | Where it now lives |
|---|---|
| Ancestor walk can loop forever on a cyclic parent chain | `visitedParentIds` guard in `organizations.adapter.ts` |
| Swallowed errors bypass Sentry capture | `handleApiResponse` now runs on every response; `handleApiError` in the catch |
| Degraded-hierarchy copy may overstate the actual fallback | Notice reworded to "Organization grouping is incomplete. Some providers may appear ungrouped." |
| Finish the rename inside the normalized shape | `CandidateProviderMapping` normalizes to `candidateId`/`providerId` |
| Add the unknown-`kind` counterpart to this drift test | Covered in `organizations.test.ts`, plus a prototype-key case |
| Unknown `kind` values bypass the neutral fallback | `toNodeKind()` membership check — stronger than a `??`, which prototype keys defeat |
| Extract the nested relationship object shapes | `OrganizationResourceRef` / `OrganizationRelationshipRef` / `CountedRelationshipRef` / `OrganizationToOneRef` / `OrganizationNullableToOneRef` |

**1 fix applied** — duplicated "relationship providers vs. fallback" resolution logic, extracted into a shared `resolveProviderRowsAndIds` helper. The org-level call site previously applied two exclude sets on the relationship branch but only one on the fallback branch; that asymmetry is a no-op, since the fallback map skips any provider carrying a node relationship, so a single union set is behavior-preserving.

**Files modified:**
- `ui/app/(prowler)/providers/providers-page.utils.ts`

**Commit:** `a51856e76`

Verified locally: `tsc` 0 errors, ESLint clean, unit 2459/2459, integration 59/59.

Alan-TheGentleman (changes requested):
Thanks for addressing the previous review rounds. I re-verified the current head at `a51856e`, and the earlier pagination, type-safety, hierarchy, and CodeRabbit findings are resolved.

One actionable item remains in `ui/actions/organizations/organizations.ts:99`: exhausting the 50-page hierarchy guard still returns the degraded result without reporting the cause through `handleApiError`. Estates above 5,000 hierarchy resources therefore degrade silently and remain indistinguishable from ordinary request failures in Sentry. Please report a concrete guard-exhaustion error before returning and add a regression assertion for that branch, then I can re-review.

ui/actions/organizations/organizations.ts
pfe-nazaries: Fixed in `041578dd5`.

When the 50-page hierarchy guard fires, `fetchOptionalCollection` now reports a concrete guard-exhaustion error through `handleApiError` (including page/resource caps and the endpoint pathname) before returning the degraded result. Added a regression assertion that the guard path calls `handleApiError` once with that message instead of degrading silently.

pfe-nazaries:
Addressed the remaining review item from @Alan-TheGentleman:

**Hierarchy pagination guard exhaustion** (`ui/actions/organizations/organizations.ts`) — fixed in `041578dd5`. Exhausting the 50-page guard now reports a concrete error via `handleApiError` before returning `{ data: [], error: true }`, so estates above 5,000 hierarchy resources are distinguishable from ordinary fetch failures in Sentry. Regression test covers the branch.

Ready for re-review.

PR #12138, opened 2026-07-27
feat(mcp): add integrations tools

puchy22:
### Context

The Prowler MCP server already covers findings, providers, scans, resources, muting, compliance and attack paths, but has no coverage of integrations. An agent can surface a critical finding, yet it cannot check whether the S3 export is configured, repair a Security Hub integration whose credentials expired, or open a Jira work item for that finding.

This adds the missing tool surface on top of the existing `/api/v1/integrations` endpoints, which have been available since 5.10.0 and are gated on the `manage_integrations` permission.

### Description

Adds a new `IntegrationsTools` class and its response models, auto-registered like every other tool module.

**Tool split.** The three usable integration types have mutually exclusive constraints at create time (`api/src/backend/api/v1/serializers.py`, `validate_integration_data`): Jira must send an empty configuration and no providers, Security Hub must send exactly one AWS provider, S3 requires a bucket name. A single generic create tool taking free-form `configuration`/`credentials` dicts cannot express that in its signature, so those constraints would only surface as repeated 400s. Everything after creation is type-agnostic, so per-type variants there would be duplicate code and wasted tool-list context. Hence the split:

| Group | Tools |
|---|---|
| Lifecycle (type-agnostic) | `prowler_list_integrations`, `prowler_get_integration`, `prowler_update_integration`, `prowler_delete_integration`, `prowler_test_integration_connection` |
| Creation (per type) | `prowler_create_amazon_s3_integration`, `prowler_create_aws_security_hub_integration`, `prowler_create_jira_integration` |
| Jira operations | `prowler_get_jira_issue_types`, `prowler_send_findings_to_jira` |

**Behaviours driven by API constraints, worth attention during review:**

- `update_integration` performs a read-modify-write on `configuration`. A partial S3 patch would otherwise fail, since `S3ConfigSerializer` requires `output_directory` alongside `bucket_name`.
- `update_integration` rejects `configuration` on Jira integrations. `IntegrationUpdateSerializer` only preserves the discovered `projects`/`issue_types` when the key is absent from the payload; sending it, even as `{}`, resets both.
- Creation always runs the connection check, since for Jira and Security Hub that check is what discovers the available projects and the enabled regions. Without it the integration is not usable.
- `send_findings_to_jira` re-reads the task state when polling times out, and reports `dispatched: true` unless the task actually failed. This prevents a slow dispatch from being retried into duplicate Jira work items.
- Security Hub `regions` (a mapping of every region in the partition to a boolean) is collapsed to `enabled_regions`, a list of the enabled ones only, mirroring what the UI displays.

**Scope decisions:**

- **Slack is excluded.** The enum value exists, but the API rejects every create with `Integration type not supported yet: slack`, its connection check is a no-op, and its dispatch is commented out in `check_integrations_task`. A creation tool for it could only ever return a 400. The tools still degrade gracefully if such a row exists: `integration_type` is typed as `str` rather than a `Literal`, and the list filter passes the value straight through to the API, which accepts `slack` as a valid choice.
- **No Lighthouse allowlist change.** `ui/lib/lighthouse-v1/allowed-tools.ts` is untouched, so these tools stay MCP-only.
- **No filtering by `enabled` or `connected`** in `prowler_list_integrations`, because `IntegrationFilter` does not support those lookups. The tool docstring directs the model to read those fields from the results instead.
- **No tests.** `mcp_server/tests/` currently contains only `test_health.py`, and there is no mocking or fixture infrastructure for tool tests, so this follows the existing precedent of the other tool modules.

### Steps to review

**1. Verify registration and that no private helper leaked as a tool:**

```bash
PROWLER_API_KEY=pk_dummy uv run --project ./mcp_server python -c "
from fastmcp import FastMCP
from prowler_mcp_server.prowler_app.utils.tool_loader import load_all_tools
m = FastMCP('smoke'); load_all_tools(m)
"
```

Expected: `Auto-registered 10 tools from IntegrationsTools`. Total tools for the namespace go from 32 to 42, which matches the count in the tools reference doc.

**2. Lint:**

```bash
uv run --project ./mcp_server ruff check mcp_server/
uv run --project ./mcp_server ruff format --check mcp_server/
```

**3. Functional walkthrough** against a tenant with a `manage_integrations` role:

- `prowler_list_integrations` then `prowler_get_integration` on an existing row of each type. Confirm Security Hub shows `enabled_regions` rather than the raw `regions` mapping, that Jira surfaces `projects`/`issue_types`, and that `connected` is still present as `null` for a never-checked integration.
- `prowler_create_jira_integration` with a full site URL such as `https://acme.atlassian.net/jira`. Confirm the domain is normalised to `acme` and that the automatic connection check populates `projects`.
- `prowler_update_integration` with a `configuration` on a Jira integration. Confirm it is refused rather than wiping the discovered projects.
- `prowler_update_integration` with only `bucket_name` on an S3 integration. Confirm the merge preserves `output_directory` and the patch succeeds.
- `prowler_get_jira_issue_types` followed by `prowler_send_findings_to_jira` with one or two finding IDs against a scratch project, then confirm `created_count`.
- `prowler_delete_integration` on a throwaway integration.

### Checklist

<details>

</details>

#### SDK/CLI

#### UI

_Not applicable, no UI changes._

#### API

_Not applicable, no API changes. Only existing endpoints are consumed._

#### MCP Server
- [x] Ensure a changelog fragment is added under [mcp_server/changelog.d/](https://github.com/prowler-cloud/prowler/tree/master/mcp_server/changelog.d), if applicable.

### License

## Summary by CodeRabbit

* **New Features**
  * Added **Integrations Management** for Amazon S3, AWS Security Hub, and Jira.
  * You can list, create, view, update, delete, test connections, and manage integration result destinations.
  * Jira support includes listing issue types and sending findings as Jira work items.
  * Expanded the authenticated toolset from **39** to **42**.

* **Documentation**
  * Updated the Tools Reference, server README, and changelog to reflect the new Integrations Management capabilities and permission.


mcp_server/prowler_mcp_server/prowler_app/tools/integrations.py
```diff
@@ -0,0 +1,876 @@
...
+        """
+        state = None
+        try:
+            task = await self.api_client.get(f"/tasks/{task_id}")
+            state = task.get("data", {}).get("attributes", {}).get("state")
+        except Exception as e:
+            self.logger.error(f"Could not read the state of task {task_id}: {e}")
+
+        if state in ("failed", "cancelled"):
+            return JiraDispatchResult(dispatched=False, error=error).model_dump()
```
josema-xyz: A failed task may already have created some Jira tickets. Returning dispatched: false makes a retry look safe and can create duplicates. We should mark this result as unknown and tell the caller not to retry.

mcp_server/prowler_mcp_server/prowler_app/tools/integrations.py
```diff
@@ -0,0 +1,876 @@
...
+        self.logger.info(f"Updating integration {integration_id}...")
+
+        current = await self._get_integration_raw(integration_id)
+        integration_type = current["attributes"]["integration_type"]
+
+        attributes: dict[str, Any] = {}
+        if enabled is not None:
+            attributes["enabled"] = enabled
+
+        if credentials is not None:
```
josema-xyz: If a caller explicitly sends `credentials={}` for a Jira integration, this code includes an empty credentials object in the PATCH. The API then saves it and deletes the stored domain, email, and token. Can we reject an explicitly empty credentials object for Jira?

mcp_server/prowler_mcp_server/prowler_app/tools/integrations.py
```diff
@@ -0,0 +1,876 @@
...
+            return DetailedIntegration.from_api_response(current).model_dump()
+
+        update_body: dict[str, Any] = {
+            "data": {
+                "type": "integrations",
+                "id": integration_id,
+                "attributes": attributes,
+            }
+        }
+        if provider_ids is not None:
```
josema-xyz: For Security Hub, `provider_ids=[]` removes the required provider. Changing the provider also keeps the old connected and enabled_regions values. Can we require one provider and check the connection again when it changes?

mcp_server/prowler_mcp_server/prowler_app/tools/integrations.py
```diff
@@ -0,0 +1,876 @@
...
+    "enabled,connected,connection_last_checked_at,integration_type,providers,"
+    "inserted_at,updated_at"
+)
+
+CONNECTION_CHECK_TIMEOUT = 120
+# One Jira work item is created per finding, sequentially, so this needs to be generous
+JIRA_DISPATCH_TIMEOUT = 300
+
+
+class IntegrationsTools(BaseTool):
```
josema-xyz, line 39: This adds ten tools with no tests. There is a lot of update and background-task logic here, so we need tests for the main success and failure cases before merging.

mcp_server/prowler_mcp_server/prowler_app/tools/integrations.py
puchy22: True, the result now carries `status` (`completed` / `in_progress` / `unknown`) and an explicit `safe_to_retry`, which is only true when the task completed having created zero work items. A failed or cancelled task is `unknown` + `safe_to_retry: false`, with a message telling the caller to check the Jira project rather than resending. Same for a dispatch that comes back without a task ID, and for a completed task whose result can't be read — that last one used to default the counters to zero, which would have looked retryable. The tool docstring now tells the model never to resend when `safe_to_retry` is false.

mcp_server/prowler_mcp_server/prowler_app/tools/integrations.py
puchy22: Fixed. Credentials are validated per type before the PATCH is built: for Jira, `domain`, `user_mail` and `api_token` are all required, so `{}` is rejected — and so is a partial object, since the API replaces the secret as a whole and a partial one destroys the rest just the same. It also normalizes the domain the way creation does. AWS types still accept an empty object, since that's the documented fallback to the ambient or provider credentials.

mcp_server/prowler_mcp_server/prowler_app/tools/integrations.py
puchy22: Fixed both parts. `provider_ids` is now validated before anything else is built: Jira rejects it outright, and Security Hub requires exactly one ID, pointing at `prowler_delete_integration` as the way to stop sending findings. The connection is re-checked whenever the attached providers actually change — compared against the current linkage, so `connected` and `enabled_regions` get refreshed, while resending the same provider doesn't trigger a pointless check.

mcp_server/prowler_mcp_server/prowler_app/tools/integrations.py
puchy22: We have to do something with this, because there are a lot of code now and we should test that everything is working right.

josema-xyz (approved):

pedrooot (approved):
👏🏼

PR #12167, opened 2026-07-28
fix(api): scope Attack Paths predefined queries with provider label

danibarranqueroo:
### Context

Follow-up to the cartesian-product fix (#12136). On Prowler Cloud (Neptune sink) the trust-policy IAM privilege-escalation queries (IAM-012/019/020/021 and siblings) still timed out even after the cartesian fix landed and was confirmed to be the code actually executing in dev.

Root cause: the Attack Paths graph is **multi-tenant** — every synced node carries a dynamic `_Provider_{uuid}` isolation label, and that label is the *only* provider scoping on most node patterns. The **custom-query** path injects it into every node pattern (`inject_provider_label`), but the **predefined-query** path (`execute_query`) ran the catalog cypher verbatim. Without the label, Neptune's cost-based planner seeded a pattern like `MATCH (aws)--(target_role:AWSRole)` from a **global** `:AWSRole` label scan (every tenant's roles) instead of the selective per-provider label index, producing a disconnected/cartesian plan that timed out.

The `{id: $provider_uid}` property anchor only scopes the single account node; it does not propagate to `target_role`/`stmt`/`res`/`act`. Only the label — present on every synced node — does.

This was validated end-to-end against dev without any code change first: running the exact `aws.py` IAM-012 cypher through the custom endpoint (which injects the label = what this PR does) returned in ~2s, while the same query via the catalog (no label) timed out. Identical cypher, single variable = the label.

### Description

In `execute_query` (predefined path), inject the provider label into the query cypher using the same `inject_provider_label` the custom path already uses, **gated on `scan.is_migrated`**:

- `is_migrated=True` (current schema) → inject the label; every node is scoped to the provider's label index.
- `is_migrated=False` (pre-cutover graphs) → **skip** injection; those graphs isolate via the `_provider_id` property and predate the label, so injecting would match nothing.

The injection is a pure, sink-independent string transform (same output for Neo4j and Neptune), so it does not affect the Neo4j path. No query definitions were changed; the `$provider_uid` anchor is kept (redundant but harmless scoping).

`# TODO: drop the is_migrated guard after Neptune cutover` is noted, consistent with the existing catalog-selection TODOs.

### Steps to review

1. `api/src/backend/api/attack_paths/views_helpers.py` — the guarded injection in `execute_query`.
2. Tests:
   - `test_execute_query_injects_provider_label_when_migrated` / `..._does_not_inject_label_when_deprecated` — gating.
   - `TestPredefinedCatalogInjection` in `test_cypher_sanitizer.py` — runs over the **entire predefined catalog (78 queries)** asserting injection is *lossless* (stripping the injected `:_Provider_*` tokens restores the original cypher verbatim) and preserves every `$param` placeholder. This guards against a regex regression corrupting any hand-authored query.
3. Run: `cd api && uv run pytest src/backend/api/tests/test_attack_paths.py src/backend/api/tests/test_cypher_sanitizer.py` → 264 passed locally.

### Checklist

<details>

</details>

- [x] Review if code is being documented following the Google style guide.
- [ ] Review if is needed to change the README.md.
- [x] Ensure a changelog fragment is added, if applicable.

#### API
- [ ] EXPLAIN ANALYZE output for new/modified queries or indexes (if applicable) — Neptune openCypher plan is not exposed via the API; the label→plan effect was validated by the ~2s-vs-timeout delta on identical cypher differing only by the injected label.
- [x] Performance test results — IAM-012 via label-injected path returned ~2s on dev vs. timeout via the unlabeled catalog path.
- [ ] Verify if API specs need to be regenerated. — No endpoint/schema changes.
- [ ] Check if version updates are required. — None.
- [x] Ensure a changelog fragment is added under api/changelog.d/.

### License

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed Attack Paths predefined queries on migrated graphs by scoping execution to the provider label, improving performance and reducing query timeouts.
  - Preserved behavior for non-migrated graphs (no provider-label rewriting).
  - Ensured rewritten queries remain lossless and keep existing parameter placeholders intact.
- **Tests**
  - Added unit tests validating provider-label injection behavior for predefined Cypher, including lossless transformations and placeholder preservation across the full predefined catalog.


api/src/backend/api/attack_paths/views_helpers.py
```diff
@@ -115,7 +115,22 @@ def execute_query(
...
+        # is the difference between a sub-second plan and a query that times
+        # out. The custom-query path already relies on this same injection.
+        # Deprecated (pre-cutover) graphs predate the label and isolate via the
+        # `_provider_id` property, so injecting there would match nothing - skip
+        # it and run the legacy catalog as-is.
+        # TODO: drop the is_migrated guard after Neptune cutover
+        if scan.is_migrated:
+            cypher = inject_provider_label(cypher, provider_id)
+
+        graph = backend.execute_read_query(database_name, cypher, parameters)
```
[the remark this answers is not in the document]
danibarranqueroo: Thanks — the finding is based on the comment, which was inaccurate; corrected in 4b5e60d.

No serialization change is needed: `_serialize_graph(graph, provider_id)` already label-filters **both** catalogs (migrated and deprecated) — it predates this PR and is unchanged here. Deprecated graphs are **not** unlabeled: the sync labels the entire provider subgraph with the `_Provider_{uuid}` isolation label regardless of schema (see scan.py sync step: "Every synced node gets a _ProviderResource label and dynamic _Tenant_{uuid} / _Provider_{uuid} isolation labels"). If deprecated graphs lacked the label, `_serialize_graph` would already return empty for them today, independent of this change. Both catalogs also isolate via `(aws:AWSAccount {id: $provider_uid})`, not a `_provider_id` property.

The `is_migrated` guard is therefore about scope/plan, not isolation: only the migrated catalog runs on the Neptune sink where the plan blowup occurs; the pre-cutover legacy catalog is dropped after the cutover, so it's left byte-for-byte unchanged.

api/src/backend/api/tests/test_attack_paths.py
```diff
@@ -154,6 +154,84 @@ def test_execute_query_serializes_graph(
...
+        parameters,
+        provider_id=provider_id,
+        scan=MagicMock(is_migrated=True, sink_backend="neo4j"),
+    )
+
+    executed_cypher = sink_backend_stub.execute_read_query.call_args[0][1]
+    assert executed_cypher != definition.cypher
+    # Every node pattern is scoped, and injection only inserted labels
+    # (stripping them restores the original cypher verbatim).
+    assert f":{plabel}" in executed_cypher
```
josema-xyz: This only checks that one node gets the label. The query has two nodes, so the test could pass while the bug still exists. The catalog test has the same issue. Could we check that every node gets the label?

danibarranqueroo: Good catch. Both tests now assert every node is scoped: the unit test checks the exact rewrite (both nodes labeled), and the catalog test asserts the injected-label count equals the node-pattern count (via an independent counter, validated against the injector across all catalog queries). Fixed in e5120cd01a.

josema-xyz (approved):

PR #11950, opened 2026-07-10
feat(providers): add Huawei Cloud provider with CIS 1.0 benchmark

tomitobio:
## Summary

This PR adds a new **Huawei Cloud** provider to Prowler, following the same architecture and patterns as existing built-in providers (AWS, Azure, GCP, AlibabaCloud, etc.).

### What's included

- **Huawei Cloud provider** (`prowler/providers/huaweicloud/`) — provider class, services, and 21 checks across 10 services:
  - IAM (8 checks), OBS (2), KMS (2), VPC (2), RDS (2), ECS (1), ELB (1), EVS (1), CTS (1), WAF (1)
- **CIS Huawei Cloud 1.0 compliance benchmark** (`prowler/compliance/huaweicloud/cis_1.0_huaweicloud.json`) — 26 requirements across 8 sections
- **Core Prowler integration** (5 files patched):
  - `CheckReportHuaweiCloud` dataclass in `lib/check/models.py`
  - Mutelist support in `lib/check/check.py`
  - Finding output in `lib/outputs/finding.py`
  - stdout report in `lib/outputs/outputs.py`
  - Provider instantiation in `providers/common/provider.py`
- **SDK dependencies** in `pyproject.toml` and `uv.lock` (Huawei Cloud SDK `==3.1.181`)
- **67 tests** covering all checks and provider logic (`tests/providers/huaweicloud/`)

### Authentication

Credentials are provided via environment variables:
- `HUAWEICLOUD_ACCESS_KEY_ID`
- `HUAWEICLOUD_SECRET_ACCESS_KEY`
- `HUAWEICLOUD_PROJECT_ID`
- `HUAWEICLOUD_REGION`

### Test results

```
pytest tests/providers/huaweicloud/ -q: 67 passed in 0.82s
ruff check: All checks passed
```

### Checklist

- [x] Provider follows existing Prowler provider architecture
- [x] All checks have metadata files
- [x] Tests pass and lint is clean
- [x] No hardcoded secrets — all credentials via env vars
- [x] CIS benchmark maps all 21 checks to 26 requirements

## Summary by CodeRabbit

* **New Features**
  * Added Huawei Cloud as a supported provider with authentication, region auditing, connectivity validation, and mute-list support.
  * Added Huawei Cloud security checks for IAM, OBS, ECS, ELB, EVS, RDS, VPC, KMS, CTS, and WAF, each with console/CLI remediation via check metadata.
* **Bug Fixes**
[...]

PR #12196, opened 2026-07-28
feat(providers/huaweicloud): add dcs_instance_ssl_enabled check

tomitobio:
## New Check: `dcs_instance_ssl_enabled`

**Service:** dcs
**Severity:** medium
**Categories:** encryption

### Description

Ensure that DCS Redis instances have SSL/TLS encryption enabled for data in transit.

### Risk

Redis instances without SSL/TLS encryption transmit data in plaintext, making it vulnerable to interception and tampering on the network.

### Remediation

Enable SSL/TLS encryption for DCS Redis instances to encrypt data in transit.

### Implementation Details

Checks if DCS Redis instances have SSL/TLS enabled (enable_ssl is not false).

This is part of the Huawei Cloud provider expansion. See #11950 for the initial provider PR.

### DIFF
--- a/prowler/changelog.d/dcs-instance-ssl-enabled.added.md
+++ b/prowler/changelog.d/dcs-instance-ssl-enabled.added.md
@@ -1,0 +1,1 @@
+`dcs_instance_ssl_enabled` check for Huawei Cloud provider: DCS Redis instances have SSL/TLS encryption enabled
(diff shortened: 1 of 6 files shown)


pedrooot:
This PR carries the same `is_mock` problem as the rest of your HuaweiCloud service PRs. The full explanation — the `AttributeError` it raises on a real scan, why CI stays green, and the fixture to use instead — is here:

https://github.com/prowler-cloud/prowler/pull/12193#issuecomment-5809989586

Please do not update this one yet. Fix #12193 first, we review it and confirm the approach, and then you apply the same change here. Otherwise you risk reworking 22 PRs twice.

PR #12210, opened 2026-07-29
fix(api): reject API keys whose owning user was deleted

pedrooot:
### Context

`TenantAPIKey.entity` is a nullable FK with `on_delete=SET_NULL`, so an API key row survives the deletion of the user that owns it. Two things went wrong around that:

1. **Authentication crashed instead of returning 401.** A request using an orphaned key passed every check in `_authenticate_credentials` (it is neither unknown, expired nor revoked) and then blew up while building the auth dict: `"sub": str(api_key_instance.entity.id)` raises `AttributeError` when `entity` is `None`. DRF re-raises `AttributeError`s from authenticators as `WrappedAttributeError`, which escapes the exception handler and turns into an unhandled 500.

2. **The revocation signals silently revoked nothing.** The `revoke_user_api_keys` / `revoke_membership_api_keys` receivers ran their `UPDATE` through the default connection, but `api_keys` is RLS protected and its policy denies every row when `api.tenant_id` is unset, which is exactly the situation while a user is being deleted through the admin connection. The update matched 0 rows (or only the active tenant's keys for multi-tenant users), which is how usable orphaned keys came to exist in the first place.

### Description

- `TenantAPIKeyAuthentication._authenticate_credentials` now rejects keys with `entity_id is None`, returning the same `401` (`No entity matching this api key.`) as a non-existent key so no enumeration oracle is introduced, and logs a warning with the key prefix and tenant. The key is also revoked on the spot, so it stops showing up as active and any retry fails the regular revoked check.
- `sub` is now built from the entity already validated as non-None instead of re-reading the relation from the second lookup, so a deletion racing between both lookups cannot produce a corrupt auth dict.
- The `revoke_user_api_keys` / `revoke_membership_api_keys` receivers now run through the admin connection (`MainRouter.admin_db`), which bypasses RLS and covers every tenant the user belongs to. Since they are `pre_delete` / `post_delete` receivers they fire inside the delete's own transaction, so the revocation stays atomic with the deletion.

No dependencies required.

### Steps to review

1. Read the new `entity_id is None` block in `_authenticate_credentials` (`api/src/backend/api/authentication.py`): same `401` message as a non-existent key, revocation through the admin connection before raising.
2. Check both receivers in `api/src/backend/api/signals.py` now use `MainRouter.admin_db`; their docstrings explain why the default connection silently updates nothing under RLS.
3. Run the tests:
   ```bash
   cd api && poetry run pytest src/backend/api/tests/test_authentication.py src/backend/api/tests/integration/test_authentication.py
   poetry run pytest src/backend/api/tests/test_views.py -k TestTenantApiKeyViewSet
   ```
   All new tests were verified to fail before the fix: the unit test reproduces the exact `AttributeError`, and the signal-connection tests fail with `DID NOT RAISE` (the suite runs on a single superuser database with `admin_db` patched to `default`, so they assert the connection by pointing `admin_db` at a missing alias).

### Checklist

<details>

</details>

#### SDK/CLI

#### UI

#### API
- [x] Ensure a changelog fragment is added under [api/changelog.d/](https://github.com/prowler-cloud/prowler/tree/master/api/changelog.d), if applicable.

#### MCP Server

### License

## Summary by CodeRabbit

- **Bug Fixes**
  - API-key authentication now reliably rejects “orphaned” keys (owner deleted) with `401` (“No entity matching this api key.”) instead of server errors.
  - User and membership deletions now correctly revoke related API keys across all tenants using the admin connection, avoiding RLS issues.
  - API-key usage tracking and revocation behavior are more consistent around concurrent updates.

- **Tests**
  - Added regression coverage for orphaned-key rejection + automatic revocation, delete/cascade revocation across tenants, admin-connection revocation, and API key listing with orphaned owners.


api/src/backend/api/authentication.py
```diff
@@ -102,9 +121,13 @@ def authenticate(self, request: Request):
...
         api_key_instance.save(update_fields=["last_used_at"], using=MainRouter.admin_db)
 
+        # `sub` comes from the entity already validated as non-None by
+        # `_authenticate_credentials`, not from the re-fetched instance: the owning user
+        # could be deleted between both lookups, and the auth dict must describe the
+        # entity this request actually authenticated as.
         return entity, {
             "tenant_id": str(api_key_instance.tenant_id),
-            "sub": str(api_key_instance.entity.id),
+            "sub": str(entity.id),
```
AdriiiPRodri: The API key is validated during the first lookup, but the second lookup does not recheck revoked or `entity_id`. If user or membership deletion revokes the key between those reads, authentication still succeeds using the stale entity from the first lookup

Please perform validation, the `last_used_at` update, and claim construction from the same authoritative row inside `transaction.atomic(using=MainRouter.admin_db)`, using `select_for_update()`. The returned entity and sub should come from that locked row. A regression test should revoke or orphan the key between validation and claim construction

AdriiiPRodri (changes requested):

AdriiiPRodri (approved):

PR #12211, opened 2026-07-29
fix(api): return 401 instead of 500 for API keys with no owning user

cesararroba:
### Context

Production pro API returned intermittent 500s on `/api/v1/scans`, driving a critical ALB5XXErrors alert (57 breach buckets over 3 days, peaking at 29% error rate).

### Description

`TenantAPIKey.entity` is a ForeignKey with `on_delete=SET_NULL`, so an API key outlives the user that owned it. `TenantAPIKeyAuthentication` then dereferenced `api_key_instance.entity.id` and raised `AttributeError: 'NoneType' object has no attribute 'id'`, surfacing as a 500 during authentication rather than a 401.

The fix rejects a key with a null entity in `_authenticate_credentials` with the existing "No entity matching this api key." message, and reads `sub` from the already-validated entity instead of re-dereferencing the re-fetched instance.

A `pre_delete` signal on `User` revokes keys, so these orphaned rows predate that signal or were created through a path that bypassed it; the guard is defensive regardless.

Adds a regression test, `test_authenticate_api_key_with_null_entity`, which reproduces the production state (entity NULL, revoked False) by bypassing the revoke signal.

### Steps to review

Diff is confined to `api/src/backend/api/authentication.py` and its test file, plus a changelog fragment. All 27 tests in `api/tests/test_authentication.py` pass.

### Checklist

<details>

- [ ] This feature/issue is listed in the open issues or roadmap.prowler.com
- [ ] Is it assigned to me, if not, request it via the open issues or Prowler Community Slack
- [ ] I have reviewed the open pull requests and confirmed there is no existing PR that implements the same outcome

</details>

- [x] Review if code is being documented following this specification
- [x] Ensure a changelog fragment is added under prowler/changelog.d/, if applicable.

#### SDK/CLI

#### API
- [x] Any other relevant evidence of the implementation (if applicable) — production ALB5XXErrors alert data cited above
- [x] Ensure a changelog fragment is added under api/changelog.d/, if applicable.

### License

## Summary by CodeRabbit

* **Bug Fixes**
  * API keys associated with deleted or unavailable entities now return HTTP 401 Unauthorized instead of causing an internal server error.
  * Authentication safely rejects orphaned API keys with a clear error message.

### DIFF
--- a/api/changelog.d/api-key-null-entity-401.fixed.md
+++ b/api/changelog.d/api-key-null-entity-401.fixed.md
@@ -1,0 +1,1 @@
+API keys whose owning user no longer exists now fail authentication with 401 instead of raising a 500
--- a/api/src/backend/api/authentication.py
+++ b/api/src/backend/api/authentication.py
@@ -59,6 +59,10 @@ class TenantAPIKeyAuthentication(BaseAPIKeyAuth):
         if api_key.revoked:
             raise AuthenticationFailed("This API Key has been revoked.")
 
+        # `entity` is SET_NULL, so a key can outlive the user that owned it
+        if api_key.entity is None:
+            raise AuthenticationFailed("No entity matching this api key.")
+
         client_ip = request.META.get(package_settings.IP_ADDRESS_HEADER)
         if api_key.blacklisted_ips and client_ip in api_key.blacklisted_ips:
             raise AuthenticationFailed("Access denied from blacklisted IP.")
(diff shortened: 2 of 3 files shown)


cesararroba:
Closing in favour of #12210, which was opened first and covers strictly more.

Both PRs land the same guard in `_authenticate_credentials` and the same `sub` change to the validated entity. #12210 additionally:

- uses `api_key.entity_id is None`, avoiding the lazy FK query that `api_key.entity is None` triggers
- fixes why the orphaned rows exist in the first place: `revoke_user_api_keys` and `revoke_membership_api_keys` ran on the default connection, and the RLS policy on `api_keys` denies every row when `api.tenant_id` is unset, so the revocation silently matched nothing
- adds integration coverage plus tests for multi-tenant revocation and the admin-connection requirement

For context on the production impact this was found through: the null dereference surfaced as intermittent 500s on `/api/v1/scans` in pro, which drove a critical ALB5XXErrors alert — 57 breach buckets over three days, peaking at a 29% error rate.

PR #12221, opened 2026-07-29
fix(html): escape provider data in reports

HugoPBrito:
### Context

HTML reports rendered provider-originated finding data without complete HTML escaping, allowing malicious cloud resource tags to inject active markup.

### Description

- Escape provider-originated finding fields with a shared `markupsafe.escape` helper
- Escape resource tag keys and values before adding trusted report markup
- Add regression coverage for stored XSS payloads and an SDK security changelog fragment

### Steps to review

1. Run `uv run pytest tests/lib/outputs/html/html_test.py -q`
2. Confirm malicious resource tag markup is rendered as escaped text
3. Confirm existing HTML report formatting remains unchanged

### Checklist

- [x] Review if code is being documented following the Python style guide.
- [x] Review if is needed to change the Readme.md.
- [x] Ensure a changelog fragment is added under `prowler/changelog.d/`.

#### SDK/CLI

- Permissions update required? No

### License

[PROWLER-2093]: https://prowlerpro.atlassian.net/browse/PROWLER-2093?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Security**
  * Improved safety of generated HTML reports by escaping provider-controlled finding fields, including resource tags, status details, risk, and recommendation content, to prevent stored XSS.
  * Hardened markdown/recommendation links by blocking unsafe URL schemes and neutralizing unsafe `href` values while preserving valid `http/https` links.

* **Tests**
  * Added/expanded unit tests to confirm payloads are not rendered unescaped, and that unsafe link schemes are removed/blanked while safe links remain intact.

### DIFF
--- a/prowler/lib/outputs/html/html.py
+++ b/prowler/lib/outputs/html/html.py
@@ -19,6 +19,11 @@ from prowler.providers.common.provider import Provider
 
 class HTML(Output):
     @staticmethod
+    def escape_provider_data(value: object) -> str:
+        """Escape provider-originated data for insertion into HTML text contexts."""
+        return str(escape(str(value)))
+
+    @staticmethod
     def process_markdown(text: str) -> str:
         """
         Process markdown syntax in text and convert to HTML using the markdown library.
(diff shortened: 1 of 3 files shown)


prowler/lib/outputs/html/html.py
```diff
@@ -18,6 +18,11 @@
 
 
 class HTML(Output):
+    @staticmethod
+    def escape_provider_data(value: object) -> str:
```
puchy22: Is this really necessary:
1. The function itself—after all, all it does is act as an escape wrapper that takes an object as input, when everything passed to it is already strings.
2. Why is this a static method of the class and not simply a private helper method? Is this going to be used anywhere else to justify using a static method?

HugoPBrito: Thanks! I just used  `escape()` instead: https://github.com/prowler-cloud/prowler/pull/12221/commits/4bf51dcaf4ff5ec1fb775dc60563669f12ab8341

HugoPBrito: Updated in `4bf51dcaf`. I removed the wrapper and now use `escape` directly at each rendering site.

puchy22 (dismissed):

puchy22 (approved):

PR #12225, opened 2026-07-29
fix(kubernetes): return empty list when resource gather fails

Zuhef:
### Context

Five Kubernetes services gather their resources with a helper whose `return`
sits inside the `try` block:

```python
def _get_apiserver_pods(self):
    try:
        apiserver_pods = []
        for pod in self.client.pods.values():
            ...
        return apiserver_pods          # only reached on success
    except Exception as error:
        logger.error(...)              # falls through -> implicit return None
```

When the Kubernetes API call raises, the `except` branch only logs and the
function falls through to an implicit `return None`. That `None` is assigned
straight to the service attribute:

```python
self.apiserver_pods = self._get_apiserver_pods()   # None
```

Every check then iterates it directly:

```python
for pod in apiserver_client.apiserver_pods:        # TypeError
```

So a transient API error becomes
`TypeError: 'NoneType' object is not iterable` rather than a check that degrades
gracefully.

### Description

Initialise the collection before the `try` and return it unconditionally.
Alongside fixing the crash, this preserves any resources gathered before the
failure instead of discarding them.

| Service | Gatherer | Check files that iterate the result |
| --- | --- | --- |
| `apiserver` | `_get_apiserver_pods` | 29 |
| `controllermanager` | `_get_controllermanager_pods` | 7 |
| `etcd` | `_get_etcd_pods` | 7 |
| `scheduler` | `_get_scheduler_pods` | 2 |
| `kubelet` | `_get_kubelet_config_maps` | 10 |

That is 55 check files in total — the CIS Kubernetes Benchmark control plane and
kubelet checks.

None of these five services had test coverage, so this PR also adds
`tests/providers/kubernetes/services/control_plane_pods_test.py` covering:

1. resources are filtered correctly (guards against the fix loosening the
   filter). The four pod gatherers match on namespace **and** name prefix; the
   kubelet gatherer matches on name prefix alone, so that assertion is
   parameterised per service rather than assumed uniform.
2. a failed gather yields `[]` rather than `None`,
3. checks return zero findings instead of raising when the gather fails.

No behaviour changes on the success path.

### Steps to review

Reproduce the crash on `master` — keep the tests, drop the service fixes:

```bash
git stash push -- prowler/providers/kubernetes
pytest tests/providers/kubernetes/services/control_plane_pods_test.py
```

`test_gather_failure_yields_empty_list_not_none` fails with
`apiserver_pods is None`, and `test_checks_do_not_crash_when_gather_fails` fails
with the real `TypeError: 'NoneType' object is not iterable` raised inside
`apiserver_anonymous_requests.py`. Reverting only
`kubelet/kubelet_service.py` reproduces the same pair for
`kubelet_config_maps`.

With the fix applied:

```bash
pytest tests/providers/kubernetes/services/control_plane_pods_test.py   # 3 passed
pytest tests/providers/kubernetes                                       # no new failures
```

Note: `kubernetes_provider_test.py::test_set_proxy_settings_uppercase_proxy_takes_precedence`
fails on Windows both before and after this change (environment variables are
case-insensitive there), so it is unrelated to this PR.

Linters run on the touched files: `black`, `isort`, `flake8`, `pylint`,
`vulture`, `bandit` — all clean.

### Checklist

#### SDK/CLI

### License

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Kubernetes control-plane checks so pod and ConfigMap collection failures no longer cause runtime errors.
  * Malformed kubelet configuration data is skipped while valid configurations continue to be evaluated.
  * Checks now return collected results when available, or an empty list when collection fails.

* **Tests**
  * Expanded coverage for resource filtering, malformed configurations, and graceful handling of collection failures.


danibarranqueroo (approved):
Thanks @Zuhef — we reproduced the kubelet case on a real cluster and the fix keeps the valid ConfigMaps' findings; the extra per-ConfigMap guard makes it order-independent. Approving.

PR #12228, opened 2026-07-29
fix(check): return a set when check loading fails

Zuhef:
﻿### Context

Three check-loading helpers are annotated `-> set` but place their `return` inside the `try` block, so when the body raises, the `except` handler only logs and the function falls through to an implicit `return None`.

| function | consumer | result of `None` |
| --- | --- | --- |
| `parse_checks_from_file` (`lib/check/check.py`) | `__main__.py` -> `list(excluded_checks_from_file)` | `TypeError: 'NoneType' object is not iterable` |
| `recover_checks_from_service` (`lib/check/utils.py`) | `aws_provider.py` -> `for check in checks:` | `TypeError: 'NoneType' object is not iterable` |
| `update_checks_to_execute_with_aliases` (`lib/check/checks_loader.py`) | returned straight out of `load_checks_to_execute` | `None` propagates to every caller |

The first one is reachable without any unusual failure: pass a `--checks-file` whose JSON has no section for the provider being scanned, and `json_file[provider]` raises `KeyError`.

`recover_checks_from_service` already returns `set()` on its tool-wrapper branch, so returning a set is its own contract; the error path was just missing it.

### Description

* `parse_checks_from_file` and `recover_checks_from_service` return an empty set on failure.
* `update_checks_to_execute_with_aliases` falls back to the checks originally requested, since alias expansion is a best-effort transformation and the caller's selection is the safe fallback.

Three lines of production code plus a comment each.

### Steps to review

Reproduce on `master` — keep the tests, drop the fixes:

```bash
git stash push -- prowler/lib/check/check.py prowler/lib/check/utils.py prowler/lib/check/checks_loader.py
pytest tests/lib/check -k "returns_empty_set_on_error or falls_back_on_error"
```

All three fail with `assert None is not None`. With the fixes, `3 passed`.

One note on the test for `parse_checks_from_file`: a *missing* file cannot be used to reach that handler, because `open_file` logs critical and calls `sys.exit(1)` on `OSError` first. The test therefore uses an existing fixture with no section for the requested provider, which is the realistic path into the handler.

Full `tests/lib/check` run: **28 failures before and after, identical sets, none introduced by this change.** Those pre-existing failures are all `universal_compliance_models_test.py::TestSmokeLoadAllJSONs` compliance-JSON loading cases, unrelated to check loading.

Linters on the touched files: `black`, `isort`, `flake8`, `bandit` — all clean.

### Checklist

#### SDK/CLI

### License

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved check discovery reliability when configuration sections are missing or errors occur.
  * Prevented failures caused by unavailable check results during provider scans.
  * Preserved requested checks when alias expansion cannot be completed.

* **Tests**
  * Added coverage for check-loading failures, recovery errors, and alias expansion fallback behavior.

### DIFF
--- a/prowler/changelog.d/check-loading-implicit-none.fixed.md
+++ b/prowler/changelog.d/check-loading-implicit-none.fixed.md
@@ -1,0 +1,1 @@
+﻿Check loading no longer returns `None` when it fails: `parse_checks_from_file` and `recover_checks_from_service` now return an empty set, and `update_checks_to_execute_with_aliases` falls back to the requested checks, preventing `TypeError: 'NoneType' object is not iterable` when a checks file has no section for the scanned provider or check discovery fails
--- a/prowler/lib/check/check.py
+++ b/prowler/lib/check/check.py
@@ -70,6 +70,8 @@ def parse_checks_from_file(input_file: str, provider: str) -> set:
         logger.error(
             f"{error.__class__.__name__}[{error.__traceback__.tb_lineno}] -- {error}"
         )
+        # Callers iterate the result, so always return a set.
+        return set()
 
 
 # Load checks from custom folder
--- a/prowler/lib/check/checks_loader.py
+++ b/prowler/lib/check/checks_loader.py
@@ -271,3 +271,6 @@ def update_checks_to_execute_with_aliases(
         logger.error(
             f"{error.__class__.__name__}[{error.__traceback__.tb_lineno}] -- {error}"
         )
+        # Alias expansion is best effort; fall back to the requested checks so
+        # callers always receive a set.
+        return checks_to_execute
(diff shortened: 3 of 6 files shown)


danibarranqueroo:
Thanks for the PR. Looking at the three helpers: `recover_checks_from_service` and `update_checks_to_execute_with_aliases` can't reach their `except` branches from any real input (`recover_checks_from_provider` already exits on failure, and the alias expansion only iterates metadata), so those changes guard unreachable paths. `parse_checks_from_file` is reachable with a checks file that has no section for the scanned provider, but returning an empty set makes that worse for a security scanner: `--checks-file` now exits 0 with "0 available checks", and `--excluded-checks-file` silently excludes nothing and runs every check. Master already fails loudly (exit 1) there. We'd rather see that case handled as invalid input — a clear "checks file has no `<provider>` section" error with exit 1 — than fail open, so we're closing this one; a small PR doing that would be welcome.

PR #12230, opened 2026-07-29
fix(scan): avoid ZeroDivisionError in progress with no checks

Zuhef:
﻿### Context

`Scan.progress` divides by `_number_of_checks_to_execute` with no guard:

```python
    @property
    def progress(self) -> float:
        return (
            self._number_of_checks_completed / self._number_of_checks_to_execute * 100
        )
```

That value is assigned from `len(self._checks_to_execute)`, and an empty scope is explicitly tolerated by `__init__`. The exclusion validation only raises when exclusions emptied a *non-empty* scope, and its own comment spells out the other case:

```python
            # Only complain when exclusions actually emptied a non-empty
            # scope. If the scope was already empty (e.g. a severity or
            # category filter matched nothing) the exclusions did not
            # cause the emptiness and the misleading error would obscure
            # the real reason.
            if previous_scope and not selected_checks:
                raise ScanInvalidCheckError(...)
```

Severity validation only confirms the value is a valid `Severity` member:

```python
        if severities:
            for severity in severities:
                try:
                    Severity(severity)
                except ValueError:
                    raise ScanInvalidSeverityError(...)
```

so a valid severity that no check declares passes validation, yields zero checks, and reading `progress` raises `ZeroDivisionError`.

### Description

Return `0.0` when there is nothing to execute. One guard, no behaviour change for non-empty scans.

### Steps to review

Reproduce on `master` — keep the test, drop the fix:

```bash
git stash push -- prowler/lib/scan/scan.py
pytest tests/lib/scan/scan_test.py -k no_checks_to_execute
```

```
E       ZeroDivisionError: division by zero
prowler\lib\scan\scan.py:267: ZeroDivisionError
```

With the fix:

```bash
pytest tests/lib/scan/scan_test.py -k no_checks_to_execute   # 1 passed
pytest tests/lib/scan                                        # 25 passed
```

The existing tests assert `scan.progress == 0` only for scans that do have checks, so the empty-scope path had no coverage.

Linters on the touched files: `black`, `isort`, `flake8` — all clean.

### Checklist

#### SDK/CLI

### License

## Summary by CodeRabbit

* **Bug Fixes**
  * Scan progress now displays 0% when no checks are selected, including when filters match no available checks.
  * Prevented an error from occurring during scans with no checks to execute.

* **Tests**
  * Added coverage to verify correct progress reporting for empty scan selections.

### DIFF
--- a/prowler/changelog.d/scan-progress-zero-checks.fixed.md
+++ b/prowler/changelog.d/scan-progress-zero-checks.fixed.md
@@ -1,0 +1,1 @@
+﻿`Scan.progress` now returns 0 instead of raising `ZeroDivisionError` when a scan has no checks to execute, which happens when a severity or category filter matches nothing
--- a/prowler/lib/scan/scan.py
+++ b/prowler/lib/scan/scan.py
@@ -263,6 +263,10 @@ class Scan:
 
     @property
     def progress(self) -> float:
+        # A scan can legitimately end up with no checks to execute, e.g. when a
+        # severity or category filter matches nothing, so guard the division.
+        if not self._number_of_checks_to_execute:
+            return 0.0
         return (
             self._number_of_checks_completed / self._number_of_checks_to_execute * 100
         )
--- a/tests/lib/scan/scan_test.py
+++ b/tests/lib/scan/scan_test.py
@@ -261,6 +261,34 @@ class TestScan:
             assert scan.get_completed_checks() == set()
 
     @patch("prowler.lib.scan.scan.load_checks_to_execute")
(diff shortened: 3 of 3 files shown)


danibarranqueroo:
Thanks for the PR. `Scan.progress` is only read inside `Scan.scan()`'s per-check loop, so with zero checks it's never evaluated; the CLI also exits before constructing `Scan` when no checks match. Since there's no reachable failure here we prefer not to add a guard, a test and a changelog entry for it. Happy to review fixes for issues that are reproducible end to end — several of your other PRs are.

PR #12231, opened 2026-07-29
fix(check): handle audit metadata with no expected checks

Zuhef:
﻿### Context

`update_audit_metadata` computes the progress percentage with no zero guard:

```python
        audit_metadata.audit_progress = (
            100 * len(checks_executed) / len(audit_metadata.expected_checks)
        )
```

Five providers construct their `Audit_Metadata` with an empty `expected_checks` list, two of them with a comment saying so explicitly:

| provider | line |
| --- | --- |
| `iac_provider.py` | `expected_checks=[],  # IAC doesn't use checks` |
| `llm_provider.py` | `expected_checks=[],  # LLM doesn't use checks` |
| `image_provider.py` | `expected_checks=[]` |
| `alibabacloud_provider.py` | `expected_checks=[]` |
| `huaweicloud_provider.py` | `expected_checks=[]` |

So for those providers the division raises `ZeroDivisionError`. The handler only logs, so the function falls through to an implicit `return None` despite being annotated `-> Audit_Metadata` — and all three call sites assign the result straight onto the provider:

```python
global_provider.audit_metadata = update_audit_metadata(...)   # check.py:598
global_provider.audit_metadata = update_audit_metadata(...)   # check.py:679
self._provider.audit_metadata = update_audit_metadata(...)    # scan.py:509
```

which replaces the metadata object with `None`.

Reproduced directly:

```python
md = Audit_Metadata(
    services_scanned=0, expected_checks=[], completed_checks=0, audit_progress=0
)
print(update_audit_metadata(md, set(), set()))
```

```
ZeroDivisionError[850]: division by zero
None
```

### Description

* Guard the division, reporting 0 when there are no expected checks.
* Return the metadata unchanged from the handler, so the annotated contract holds on every path instead of silently yielding `None`.

### Steps to review

Reproduce on `master` — keep the test, drop the fix:

```bash
git stash push -- prowler/lib/check/check.py
pytest tests/lib/check/check_test.py -k update_audit_metadata
```

The new test fails with `update_audit_metadata returned None`, logging `ZeroDivisionError[850]: division by zero`. The two existing `update_audit_metadata` tests pass either way, so they act as regression guards showing non-empty behaviour is unchanged.

With the fix: `3 passed`.

Full `tests/lib/check`: **28 failures before and after, identical sets, none introduced.** Those pre-existing failures are `universal_compliance_models_test.py` cases failing with `UnicodeDecodeError` when loading compliance JSON on Windows, unrelated to this change.

Linters on the touched files: `black`, `flake8` — clean.

### Checklist

#### SDK/CLI

### License

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed audit metadata updates for providers with no expected checks.
  * Prevented division-by-zero errors in audit progress calculations.
  * Audit metadata now remains available with progress set to 0 when no checks apply.

### DIFF
--- a/prowler/changelog.d/audit-metadata-no-expected-checks.fixed.md
+++ b/prowler/changelog.d/audit-metadata-no-expected-checks.fixed.md
@@ -1,0 +1,1 @@
+﻿`update_audit_metadata` no longer raises `ZeroDivisionError` and return `None` for providers that run no checks (`iac`, `llm`, `image`, `alibabacloud` and `huaweicloud`), which build their audit metadata with an empty `expected_checks` list
--- a/prowler/lib/check/check.py
+++ b/prowler/lib/check/check.py
@@ -846,8 +846,13 @@ def update_audit_metadata(
     try:
         audit_metadata.services_scanned = len(services_executed)
         audit_metadata.completed_checks = len(checks_executed)
+        # Providers that do not run checks (`iac`, `llm`, `image`,
+        # `alibabacloud` and `huaweicloud`) build their `Audit_Metadata` with
+        # `expected_checks=[]`, so guard the division.
         audit_metadata.audit_progress = (
             100 * len(checks_executed) / len(audit_metadata.expected_checks)
+            if audit_metadata.expected_checks
+            else 0
         )
 
         return audit_metadata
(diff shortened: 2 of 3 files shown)


danibarranqueroo:
Thanks for the PR. `update_audit_metadata` is only called from `execute_checks` and `Scan.scan()`, both of which first replace the provider's `Audit_Metadata` with one whose `expected_checks` is the non-empty list of checks being run, so the division by zero isn't reachable from any real scan (the providers that build it with `expected_checks=[]` never execute checks). We prefer not to add a guard plus a changelog entry for an unreachable path; happy to review fixes that are reproducible end to end.

PR #12233, opened 2026-07-30
feat(awslambda): add awslambda_layer_no_secrets_in_content check

### LINKED ISSUE #11811: [New Check]: Detect secrets in AWS Lambda layer contents
### Existing check search

- [X] I have searched existing issues, Prowler Hub, and the public roadmap, and this check does not already exist.

### Provider

AWS

### New provider name

_No response_

### Service or product area

awslambda

### Suggested check name

awslambda_layer_no_secrets_in_content

### Context and goal

AWS Lambda layers package shared code and dependencies pulled into functions. Secrets baked into a layer are as sensitive as those in function code, but are not covered by the existing function-code check.

### Expected behavior

- Resource or scope to evaluate: Each Lambda layer version's content.
- PASS when: No secrets are detected in the layer content.
- FAIL when: A secret is detected (report layer name/version and file, not the value).

### References

- Umbrella issue: https://github.com/prowler-cloud/prowler/issues/3085
- Secret scanning checks (developer guide): https://docs.prowler.com/developer-guide/secret-scanning-checks
- API/SDK reference: lambda list_layers / list_layer_versions / get_layer_version (Content.Location -> download & unzip)
- Reference code: awslambda_function_no_secrets_in_code

### Suggested severity

High

Mirror awslambda_function_no_secrets_in_code: download the layer content zip to a temp file, scan with detect-secrets, and clean up. Reuse the existing awslambda service. Severity is High by default and becomes Critical if the secret is validated.

ganiganesh25:
### Context

Closes #11811.

`awslambda_function_no_secrets_in_code` already scans Lambda function packages for hardcoded secrets, but layers — which package shared code and dependencies pulled into every function that uses them — weren't covered, so a secret baked into a layer went undetected even though it's exactly as sensitive as one in function code.

### Description

- Extends the `Lambda` service (`awslambda_service.py`) to collect the unique layers referenced across functions in scope (`Lambda.layers`), and adds `_get_layers_code` / `_fetch_layer_code` to download each layer version's package via `get_layer_version`, mirroring the existing `_get_function_code` / `_fetch_function_code` path used by the function-code check.
- Adds `region` / `name` / `version` properties to the `Layer` model, parsed from the layer ARN (a layer can only be attached to a function in the same region, so the region is always one already being audited), so findings report a proper `resource_id`/`region` instead of falling back to blank values.
- Adds the new check `awslambda_layer_no_secrets_in_content`, reusing the same batched Kingfisher secrets-scan helpers (`detect_secrets_scan_batch`, `annotate_verified_secrets`, `SecretsScanError`) as the function-code check, including the same `secrets_ignore_patterns`/`secrets_ignore_files`/`secrets_validate` audit_config knobs and MANUAL-on-scan-error fallback.
- Adds `lambda:GetLayerVersion` to `permissions/prowler-additions-policy.json` — like `GetFunction`, it's not covered by SecurityAudit/ReadOnlyAccess since it returns a code download URL.
- Adds a changelog fragment under `prowler/changelog.d/`.

### Steps to review

1. `prowler/providers/aws/services/awslambda/awslambda_service.py` — the new `_collect_layers`/`_get_layers_code`/`_fetch_layer_code` methods and the `Layer` model's new properties.
2. `prowler/providers/aws/services/awslambda/awslambda_layer_no_secrets_in_content/` — the check itself, near-identical structure to `awslambda_function_no_secrets_in_code` by design, for consistency.
3. Tests: `tests/providers/aws/services/awslambda/awslambda_service_test.py` (3 new tests: ARN-parsing, cross-function dedup, and a full moto round-trip through `publish_layer_version` → `create_function(Layers=...)` → `_get_layers_code`) and `tests/.../awslambda_layer_no_secrets_in_content/awslambda_layer_no_secrets_in_content_test.py` (6 tests mirroring the function-code check's test suite: no layers, with/without secrets, nested-path secret, file-pattern ignore, scan-error → MANUAL).

One deliberate implementation choice worth flagging: layer ARNs already carry the version, so `get_layer_version_by_arn(Arn=...)` looked like the simpler call — but moto's mock of it doesn't return a `Content` key at all, so it can't be given real test coverage. Used `get_layer_version(LayerName, VersionNumber)` instead (functionally identical against real AWS), which moto mocks correctly, so the actual fetch path is exercised by `test_get_layers_code_fetches_each_layer_once` rather than only being reachable via live AWS.

### Test plan

- [x] `poetry run python -m pytest tests/providers/aws/services/awslambda/` — 89 passed (83 existing + 6 new for the check; 3 new service-level tests net out against no removals)
- [x] `poetry run flake8` / `black --check` on all new and modified files — clean
- [x] `poetry run pylint --disable=W,C,R,E` on new/modified files — clean
- [x] `poetry run python -m prowler.__main__ aws --list-checks --service awslambda` — new check is discovered
- [x] Metadata validates against `CheckMetadata` (pydantic) — initial `Risk` text was over the 400-char limit, caught by the check's own tests failing, and shortened

#### SDK/CLI
    - Permissions: added `lambda:GetLayerVersion` to `permissions/prowler-additions-policy.json` (see Description above).

### License

## Summary by CodeRabbit

* **New Features**
  * Added a check to scan AWS Lambda layer packages for hardcoded secrets, reporting affected files, secret types, and line numbers.
  * Improved Lambda layer discovery and de-duplication; unavailable package content is flagged for manual review.
* **Permissions**
  * Added read-only access to retrieve Lambda layer version information.
* **Documentation**
  * Added release documentation and remediation guidance for the new check.
* **Tests**
  * Added coverage for secret detection, ignore patterns, partial retrieval, scan failures, and layer collection.


ganiganesh25:
Addressed the review:

- **Download timeout**: extracted a shared \`_download_code\` helper used by both \`_fetch_layer_code\` and \`_fetch_function_code\`, with a 30s timeout on the presigned-URL download (previously unbounded).
- **Partial fetch failures**: layers whose content couldn't be downloaded now get a MANUAL finding instead of being silently omitted from the check's output; added a test for this (`test_partial_fetch_failure_reports_manual_for_unfetched_layer`).
- **Remediation.Code.CLI**: filled in with concrete `aws secretsmanager` / `aws iam` / `aws lambda` commands.
- **Severity**: kept at `high` as already applied via autofix.

Note on the metadata fields: `Severity` and the previously-empty `Remediation.Code.CLI` both matched `awslambda_function_no_secrets_in_code`'s existing metadata exactly (that check is also `critical` base severity with `annotate_verified_secrets` escalating only on a *validated* live secret, and also ships an empty `CLI` field). I'd modeled this check after that precedent intentionally. Happy to leave `high` as applied here, but flagging that the sibling check has the same pattern in case maintainers want that addressed consistently across both in a follow-up.

All 90 tests in `tests/providers/aws/services/awslambda/` pass, flake8/black/pylint clean.

danibarranqueroo (approved):
Really good job, thanks for this! 🚀 🙌

PR #12243, opened 2026-07-30
fix(ui): clear remaining container CVEs (musl, zlib, sharp)

cesararroba:
### Context

PROWLER-2298, part of PROWLER-2291 (container vulnerability remediation).

The ticket originally targeted node-tar CVE-2026-59873. **That is already fixed on master** by #12065 (merged 2026-07-21), which removes the bundled npm/npx from the base stage. `cloud-ui:5.36.0` reports no `tar` finding at all, so the ticket was rescoped to what the image actually still carries.

### Description

| Package | Installed | Now | CVEs |
|---|---|---|---|
| `musl` / `musl-utils` | 1.2.5-r21 | 1.2.5-r23 | CVE-2026-40200 (H), CVE-2026-6042 (M) |
| `zlib` | 1.3.1-r2 | 1.3.2-r0 | CVE-2026-22184 (H), CVE-2026-27171 (M) |
| `sharp` | 0.33.5 **and** 0.34.5 | 0.35.3 | GHSA-f88m-g3jw-g9cj (H ×2) |

**musl / zlib** — the base stage already ran `apk upgrade --no-cache libcrypto3 libssl3`, which is why OpenSSL never showed up in scans. musl and zlib were simply not on that list. Both fixed versions are available from the Alpine branch on the **existing pinned digest**, so this needs no Node bump and no `ui/.nvmrc` change:

```
$ apk upgrade --simulate musl musl-utils zlib   # on the pinned base
(1/3) Upgrading musl (1.2.5-r21 -> 1.2.5-r23)
(2/3) Upgrading zlib (1.3.1-r2 -> 1.3.2-r0)
(3/3) Upgrading musl-utils (1.2.5-r21 -> 1.2.5-r23)
```

**sharp** — pinned to 0.33.5 as a direct dependency while `next` pulled 0.34.5 transitively, so both copies were flagged. Pinning the direct dependency and overriding the transitive one collapses the tree to a single 0.35.3. The override lives in `pnpm-workspace.yaml`, per that file's own header (*"pnpm 11+ workspace config … everything else lives here"*), alongside the existing overrides block.

Base is `integration/PROWLER-2291`, not `master`.

### Steps to review

```
docker build -f ui/Dockerfile --target prod -t prowler-ui:test ui/
docker run --rm --entrypoint sh prowler-ui:test -c 'apk list --installed | grep -E "^(musl|zlib)-"'
```

Results on a local `linux/arm64` build — **8 findings → 0, at every severity**:

| Severity | Cleared |
|---|---|
| HIGH | `musl`, `musl-utils`, `zlib`, `sharp` ×2 |
| MEDIUM | `musl`, `musl-utils`, `zlib` |

Nothing remains. In-image versions confirmed: `musl 1.2.5-r23`, `musl-utils 1.2.5-r23`, `zlib 1.3.2-r0`, `libssl3/libcrypto3 3.5.7-r0`, and npm/npx still absent from #12065.

Runtime: container starts on Next.js 16.2.11 and returns **307** to sign-in for an unauthenticated request, with clean logs. Note the prod image requires `UI_API_BASE_URL` (or legacy `NEXT_PUBLIC_API_BASE_URL`), `AUTH_URL` and `AUTH_SECRET` — `lib/env.ts` throws at startup without them, which surfaces as an instrumentation-hook failure rather than an obvious message.

The refreshed lockfile passes the repo's supply-chain policy check (1466 entries).

### Checklist

- [x] Review if backport is needed. Not needed — targets the integration branch.

#### UI
- [x] If this PR adds or updates npm dependencies, include package-health evidence — `sharp` 0.35.3 is the current release; the bump is security-driven (GHSA-f88m-g3jw-g9cj) and adds no new dependency

### DIFF
--- a/ui/Dockerfile
+++ b/ui/Dockerfile
@@ -6,7 +6,7 @@ LABEL maintainer="https://github.com/prowler-cloud"
 # Patch Alpine OpenSSL runtime packages before all stages inherit the base image.
 # The build uses pnpm via corepack, so npm is unused — remove it (and npx) to drop
 # the bundled-npm CVE surface (node-tar CVE-2026-59873) from every stage, incl. prod.
-RUN apk upgrade --no-cache libcrypto3 libssl3 && corepack enable && rm -rf /usr/local/lib/node_modules/npm /usr/local/bin/npm /usr/local/bin/npx
+RUN apk upgrade --no-cache libcrypto3 libssl3 musl musl-utils zlib && corepack enable && rm -rf /usr/local/lib/node_modules/npm /usr/local/bin/npm /usr/local/bin/npx
 
 # Install dependencies only when needed
 FROM base AS deps
--- a/ui/changelog.d/ui-container-musl-zlib-sharp.security.md
+++ b/ui/changelog.d/ui-container-musl-zlib-sharp.security.md
@@ -1,0 +1,1 @@
+The UI container image now patches musl and zlib alongside OpenSSL, and `sharp` is pinned to 0.35.3, clearing the image's remaining CVEs
--- a/ui/package.json
+++ b/ui/package.json
@@ -109,7 +109,7 @@
     "recharts": "2.15.4",
     "require-in-the-middle": "8.0.1",
     "server-only": "0.0.1",
-    "sharp": "0.33.5",
+    "sharp": "0.35.3",
     "streamdown": "1.6.10",
     "tailwind-merge": "3.3.1",
     "tailwindcss-animate": "1.0.7",
--- a/ui/pnpm-lock.yaml
+++ b/ui/pnpm-lock.yaml
@@ -12,6 +12,7 @@ overrides:
   '@react-aria/visually-hidden>react': 19.2.7
   '@react-aria/interactions>react': 19.2.7
   lodash: 4.18.1
+  sharp: 0.35.3
   lodash-es: 4.18.1
   hono: 4.12.28
   '@hono/node-server': 1.19.14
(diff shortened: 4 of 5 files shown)


Alan-TheGentleman:
Thanks for working through these image findings. Before the integration branch reaches `master`, could we add a small runtime check that exercises a real `/_next/image` request? The current CI scans also cover CRITICAL only, so an all-severity before/after report would help confirm the “8 findings to 0” result. With those two pieces, we'll have stronger evidence that the fix works end to end.

cesararroba:
@Alan-TheGentleman both done, and the `/_next/image` check found a real bug. Thanks for pushing for it — nothing else would have caught this.

**The runtime check.** This PR's sharp override silently broke image optimisation:

| | master (0.33.5) | this branch |
|---|---|---|
| `/_next/image` | 20,052 B `image/webp` | 97,549 B `image/png` |

sharp fails to load with `ERR_DLOPEN_FAILED: libvips-cpp.so.8.18.3: No such file or directory`, Next catches it and quietly serves the original file. **HTTP 200, nothing logged, healthcheck green** — the only symptom is every image on the site being unoptimised. Confirmed on amd64 and arm64.

Root cause isn't the package or pnpm — libvips ships the `.so` and it's in the builder's `node_modules`. sharp 0.35 resolves its native binary at runtime, so Next's output file tracer never sees it and drops it from `.next/standalone`, which is what the prod stage copies. Pinning lower doesn't help: the CVE needs ≥ 0.35.0 and every 0.35.x uses the dynamic loader. Fixed in #12274 with `outputFileTracingIncludes`, verified at 20,052 B `image/webp` on both architectures.

**The all-severity report**, rebuilt with `--no-cache`:

| Severity | master | this branch |
|---|---:|---:|
| CRITICAL | 0 | 0 |
| HIGH | 2 | **0** |
| MEDIUM / LOW / UNKNOWN | 0 | 0 |

Zero at every severity — but the baseline is **2, not the 8** I quoted here. Alpine 3.23.4 landed upstream since and fixed musl/zlib on its own, so that number isn't reproducible any more. I should have re-derived it rather than repeating a five-day-old figure.

One more thing that came out of this: **the UI image isn't reproducible.** The base is digest-pinned, but `apk upgrade` resolves against Alpine's live repo at build time — two builds of the same commit gave 3.23.3 and 3.23.4. My first comparison was wrong because of it and showed findings *increasing*. Worth its own ticket; I haven't raised one yet.

PR #12258, opened 2026-07-31
fix(container): clear the Private Cloud image vulnerabilities (PROWLER-2291)

cesararroba:
## What this is

The `prowler` half of epic **PROWLER-2291**. Fifteen PRs landed on `integration/PROWLER-2291`; this is the first time they run together against master.

| PR | Ticket | Effect |
|---|---|---|
| #12238 | PROWLER-2296 | purge `git` from the api image |
| #12241 | PROWLER-2297 | pillow, httplib2, pyasn1 |
| #12242 | PROWLER-2295 | renew and prune `.trivyignore` |
| #12243 | PROWLER-2298 | ui: musl, zlib, sharp |
| #12244 | — | rebase SDK and API onto Debian 13 (trixie) |
| #12248 | — | retire the entries trixie made unnecessary |
| #12252 | PROWLER-2301 | cryptography 46 → 48 and its chain |
| #12256 | PROWLER-2307 | align compose postgres and valkey pins |
| #12260 | PROWLER-2310 | unblock the cryptography 48 chain in the api (workos, pyopenssl) |
| #12263 | — | clear the fixable HIGHs: PowerShell, Trivy, uv, joserfc, fastmcp, plus docs. Includes #12265, which purged build-only packages |
| #12270 | PROWLER-2312 | keep `wget` in the api image — #12265 broke the compose healthcheck with it |
| #12272 | PROWLER-2310 | make the api buildable on this branch, and correct a stale docs claim |
| #12274 | PROWLER-2298 | include sharp's native libvips in the standalone output — the bump alone silently disabled image optimisation |
| #12279 | — | correct two claims that did not hold: `.trivyignore` `pkg:` scoping, and a stale changelog entry |
| #12281 | — | fix three more claims: the public docs repeated the scoping error, plus a stale changelog and constraint note |

Three further commits landed on this branch after the PRs above, during final verification:

| Commit | Effect |
|---|---|
| `69942ecd7` | drop the integration branch from the `pull_request` filters — the stack is merged and nothing targets it |
| `bf34ae409` | name the perl packages the `git` purge actually removed, verified with `dpkg-query` against both images |
| `89bf8d020` | **remove `pip` from the SDK and API runtime images** |

`pip` vendors frozen copies of `setuptools` 70.3.0 and `msgpack`, and the image carried two pips — 25.0.1 from the base image and 26.2 from the `--upgrade pip` step. The repository lockfile already pins `setuptools` 82.0.1 and ships no `msgpack`, so **no dependency bump could reach them**; removing pip was the only route. `uv` is kept: the API entrypoint runs every command through `uv run`. That took the SDK 40 → 38 and the API 39 → 37.

## Verification

All four images rebuilt from this branch with `--no-cache` and scanned with Trivy 0.71.2 (the version the `trivy-scan` action pins), on 2026-08-03:

| Image | Critical (raw) | Critical (suppressed) | High |
|---|---:|---:|---:|
| `prowler` (SDK) | 4 | **0** | 38 |
| `prowler-api` | 4 | **0** | 37 |
| `prowler-ui` | 0 | **0** | **0** |
| `prowler-mcp` | 0 | **0** | **0** |

Each image was scanned twice: once with an **empty** ignore file and once with `.trivyignore`. Trivy auto-loads `./.trivyignore` from the working directory, so a run that simply omits the flag still applies suppressions — the empty file is the only way to see the true raw count.

`--no-cache` is not optional here. An earlier comparison during PROWLER-2298 reported findings *increasing* 2 → 8 and read as a regression; it was a cached layer measured against a fresh resolve, and the real answer was 2 → 0.

Functional checks, not just scans:

- SDK — `prowler --version` → 5.37.0; `Import-Module MicrosoftTeams` and `ExchangeOnlineManagement` both succeed
- API — `uv run python manage.py check` → *System check identified no issues*
- `pip` absent; `uv`, `wget` and `ps` all present

Merge into master is conflict-free (`git merge-tree`).

## The 4 remaining criticals

All `perl-base`, on the SDK and API images. **They cannot be fixed**, and `.trivyignore` now records why:

1. **No fix exists.** Debian marks one `fix_deferred`, three `affected`, none with a fixed version. A newer base, `apt upgrade`, or a newer Debian release changes nothing.
2. **The package cannot be removed.** `Essential: yes` — removal needs `--force-remove-essential`, which breaks apt downstream.
3. **Changing distribution was evaluated and rejected.** Alpine drops perl but ships no `linux-musl-arm64` PowerShell in any release, so M365 would break on arm64, which is what we run. Wolfi keeps glibc and drops perl, but pinnable tags are a paid tier.

They are also unreachable: nothing in either image invokes perl. Verified by sweeping for perl shebangs, shell/python callers, `perl` strings in ELF binaries, and `.pl`/`.pm` files or perl subprocess calls in site-packages.

## Known-red checks

**`StepSecurity Required Checks`** — PyPI cooldown on `oci` 2.183.0, which arrived with the cryptography chain. Clears itself **2026-08-04**. Not a defect.

**`e2e-tests`** — broken repo-wide, not by this branch. Re-confirmed 2026-08-03: the same two specs, `Add Provider > Add GCP Provider` and `Add Provider > Add Okta Provider`, fail identically on unrelated branches opened the same day (`fix/ui-feedback-widget-overlap`, `fix/display-imported-scans`). Each was retried 3x and failed all 3, so it is not flake. The GCP failure is a 30s timeout at `locator.fill` waiting for `getByRole('textbox', { name: 'Project ID' })` — a form field that never renders. A container rebase and dependency bumps cannot produce that.

Note for anyone checking: `ui-e2e-tests.yml` has not run since February. The live workflow is `ui-e2e-tests-v2.yml`.

**`codecov/project`** — 90.10% (-4.42%). This branch changes no application code: the diff is workflows, lockfiles, Dockerfiles, compose pins, `.trivyignore` and changelog fragments, with zero `.py` files touched. The delta is an artefact of the comparison, not lost coverage.

`api-tests`, `api-code-quality` and `api-security-scans` **were** red for the reason below and are now green.

The api pinned `workos` and `pyopenssl`, which capped `cryptography` below 47, so it could not take the SDK's bump to 48. `api/pyproject.toml` resolves the SDK from **master**, so its constraints have to match whatever master holds — at 46 they break once the bump lands, at 48 they break before it. No constraint set was valid in both states, so the verified post-merge resolution landed separately in #12260.

## ⚠️ Required immediately after merge

`api/pyproject.toml:49` points the SDK at this integration branch:

```
"prowler @ git+https://github.com/prowler-cloud/prowler.git@integration/PROWLER-2291",
```

**It must return to `@master` once this lands**, and the lock regenerated:

```bash
cd api
sed -i 's|prowler.git@integration/PROWLER-2291|prowler.git@master|' pyproject.toml
uvx uv@0.11.14 lock --upgrade-package prowler
```

This cannot be done before the merge. Master's SDK still pins `py-ocsf-models` 0.8.1, `cryptography` 46.0.7 and `oci` 2.169.0, while the api constrains 0.10.0, 48.0.1 and 2.183.0 — so `@master` is unsatisfiable until this branch is master. The branch builds green *because* of the pin, so **CI cannot catch this**.

prowler-cloud has the same pin pointing here, and reverts only after this one does.

## Merge order

1. ~~**#12260 → this branch**~~ — merged 2026-07-31.
2. **This → master**, as one atomic merge.

That way master goes straight from consistent-at-46 to consistent-at-48, with no window where the api is broken.

## Not covered

`prowler-cloud`, `prowler-registry` and `prowler-enterprise` are tracked separately under the same epic.

## License

## Summary by CodeRabbit

* **Security**
  * Updated container images and dependencies to address numerous high- and critical-severity vulnerabilities.
  * Improved vulnerability scanning documentation, findings tracking, and suppression controls.
  * Reduced runtime image contents by removing unnecessary build tools and package managers.

* **Bug Fixes**
  * Fixed production image optimization so supported images are served with Sharp processing.

* **Maintenance**
  * Upgraded PostgreSQL and Valkey container images.
  * Updated SDK, API, MCP, and UI dependencies and build tooling.
  * Added clearer release documentation for security and dependency changes.


api/Dockerfile
```diff
@@ -109,17 +108,24 @@ RUN .venv/bin/python -m prowler.providers.m365.lib.powershell.m365_powershell
...
     libxml2-dev \
     libxmlsec1-dev \
     libxmlsec1-openssl \
+    libxmlsec1t64 \
+    libxmlsec1t64-openssl \
     pkg-config \
     libtool \
     libxslt1-dev \
     python3-dev \
+    wget \
```
AdriiiPRodri: Just wondering, why did we remove `wget`? I think we might still need it for the healthchecks

api/uv.lock
```diff
@@ -194,7 +194,7 @@ constraints = [
     { name = "h2", specifier = "==4.3.0" },
     { name = "hpack", specifier = "==4.1.0" },
     { name = "httpcore", specifier = "==1.0.9" },
-    { name = "httplib2", specifier = "==0.31.2" },
+    { name = "httplib2", specifier = "==0.32.0" },
```
AdriiiPRodri, line 197: The lock file still has the old versions for some deps, see https://github.com/prowler-cloud/prowler/actions/runs/30625941641/job/91141112270

docs/security/software-security.mdx
AdriiiPRodri: `mcp_server/pyproject.toml` now declares constraint-dependencies

```suggestion
The MCP Server has a small direct-dependency surface and does not yet declare a separate constraint set.
```

AdriiiPRodri (changes requested):

api/Dockerfile
cesararroba: Good catch, and you were right. Fixed in #12270.

The healthcheck shells out to `wget`, and `api/Dockerfile` purges at line 112 while `FROM build AS dev` is 139 and the production stage is 144 — so both inherit it. The container reported unhealthy indefinitely and everything with `depends_on: service_healthy` refused to start. It affects `docker-compose.yml` too, not just dev.

I hit this independently an hour later bringing the stack up. `wget` is restored in the api image; `gnupg` and `apt-transport-https` stay purged, so 7 of the 9 findings remain cleared. The SDK image keeps the full purge since nothing healthchecks it.

Longer term the healthcheck should not depend on `wget` at all — `python -c "urllib.request.urlopen(...)"` needs nothing extra. I did not do that here because the same check lives in four compose files across three repos (prowler ×2, prowler-cloud, and prowler-enterprise's shipped template) and they would have to change in lockstep.

api/uv.lock
cesararroba: Correct, and the cause is structural rather than a stale regeneration. Fixed in #12272.

`api/pyproject.toml` resolved the SDK from `@master` while its `constraint-dependencies` carry the PROWLER-2310 bumps, so no lock could satisfy both — old constraints break once the SDK reaches master, new ones break now. No PR ordering resolves it while the pointer says `@master`.

I have pointed the SDK at the integration branch for the duration and regenerated the lock, with a comment on the line and a checklist item to revert to `@master` at the final merge.

Verified by building `api/Dockerfile` from the committed tree. Until now I had been carrying a local lock regeneration just to build the image at all, which is precisely the problem you spotted.

docs/security/software-security.mdx
cesararroba: You are right, and my own change made it false — #12263 added `constraint-dependencies` to `mcp_server/pyproject.toml` for the transitive pins `fastmcp` does not raise on its own.

Corrected in #12272, worded as a small constraint set of its own with the lock still the source of truth for everything else. Worth catching on a public page.

cesararroba:
Thanks @AdriiiPRodri — all three were right, and the first one especially. Here's where each landed.

**wget.** You called it from the diff; I only found it an hour later when the whole stack refused to come up. The healthcheck shells out to `wget`, and the purge sits in the `build` stage that both `dev` and production inherit, so it hit `docker-compose.yml` as well as the dev one. The API itself was completely fine — gunicorn up, workers booted, caches warming — and Compose killed it anyway because nothing could run the check. I'd even written "verify after purging, do not assume" on the ticket, then verified only the application paths and never the orchestration. Fixed in #12270: `wget` stays in the API image, `gnupg` and `apt-transport-https` still go, so we keep 7 of the 9 findings.

**The lockfile.** Also right, and it turned out to be structural rather than me forgetting to regenerate. `api/pyproject.toml` pointed at `@master` while its constraints carried the new versions, so no lockfile could satisfy both — the old versions break the moment the SDK reaches master, the new ones break today. I'd been quietly regenerating it locally just to build the image, which is precisely the smell you picked up on. #12272 points the SDK at the integration branch for the duration so the thing actually builds from a clean checkout.

One thing to keep an eye on: that pin **has to go back to `@master` at the final merge**, with `uv lock --upgrade-package prowler`. There's a comment on the line and it's in the checklist, but it's exactly the sort of thing that slips through, so a second pair of eyes at merge time would be welcome.

**The docs note.** My own change made that false — #12263 added a constraint set to `mcp_server/pyproject.toml` and I never went back to the page that said there wasn't one. Reworded in #12272. Good catch on a public page.

Three comments finding what took me a full stack bring-up to reach. Much appreciated.

cesararroba:
Two checks are red. Neither is caused by this branch — recording the evidence so nobody has to re-derive it.

### `e2e-tests` — pre-existing breakage on master

Two tests fail: `Add Provider › Add GCP Provider` and `Add Provider › Add Okta Provider`. Both were retried 3× and failed all 3, so this is not flake.

The same two tests fail identically on unrelated branches opened today:

| Branch | Result |
|---|---|
| `integration/PROWLER-2291` | 2 failed — Add GCP Provider, Add Okta Provider |
| `fix/ui-feedback-widget-overlap` | 2 failed — Add GCP Provider, Add Okta Provider |
| `fix/display-imported-scans` | 2 failed — Add GCP Provider, Add Okta Provider |

The GCP failure is a 30s timeout at `locator.fill` waiting for `getByRole('textbox', { name: 'Project ID' })` — a form field that never renders. A container rebase and a set of dependency bumps cannot produce that; it is a UI regression on master, and it is currently blocking every UI PR in the repo.

Worth noting for anyone checking: `ui-e2e-tests.yml` has not run since February. The live workflow is `ui-e2e-tests-v2.yml`.

Every other UI check on this PR passes — `ui-tests`, `ui-container-build-and-scan`, `ui-security-scans`, `ui-dockerfile-lint`.

### `StepSecurity Required Checks`

Reports FAILURE with no description and no run log — it only links to stepsecurity.io. `StepSecurity Harden-Runner` is still in progress on this commit, so this may resolve on its own; it is not a test or build failure either way.

### Everything else

43 checks passing, including all four container build-and-scan jobs. The remaining `sdk-tests` matrix is still running.

api/changelog.d/api-container-purge-git.security.md
```diff
@@ -0,0 +1 @@
+The API container image no longer ships `git` and the perl packages it pulled in, removing 12 critical CVEs
```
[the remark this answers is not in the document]
cesararroba: Done — the fragment now names them.

Verified against real images rather than inferred from the Dockerfile. `dpkg-query` on the published `prowlercloud/prowler-api:stable` versus the image built from this branch:

- **stable:** `liberror-perl`, `libperl5.36`, `perl`, `perl-base`, `perl-modules-5.36`
- **this branch:** `perl-base` only

So yes — `perl` and `perl-modules` are gone, along with `libperl` and `liberror-perl`, and `perl-base` is retained because Debian marks it `Essential: yes`. Removing it would need `dpkg --force-remove-essential`, which breaks `apt` for anything built downstream from the image.

On supporting the count with Trivy output: keeping the fragment to prose deliberately. The full evidence — the four remaining criticals, why each is unfixable, and the not-invoked verification — is in the PR description and in `.trivyignore`, which is where a reader chasing it will look. A changelog line that carries a scan dump is worse at being a changelog line.

cesararroba:
@AdriiiPRodri on the CodeRabbit suggestion — yes, we will apply it, but as a follow-up rather than in this PR. It is tracked as **PROWLER-2327**.

The suggestion is correct. Classic `.trivyignore` parses only the CVE ID, so the `pkg:perl-base` selector on each line is documentation, not scoping — I verified this against Trivy 0.65.0: an entry written `pkg:zlib1g` still suppressed the finding on `perl-base`. Real per-package scoping needs `.trivyignore.yaml` with purls.

Why not here:

- It is a **scanner configuration change, not a CVE fix**. It changes how suppressions are matched, which needs the `.github/actions/trivy-scan` wiring updated to select the YAML file explicitly.
- It has to land in **all four repositories** (prowler, prowler-cloud, prowler-registry, prowler-enterprise) or the estates diverge on suppression semantics.
- This branch is already an integration of ~20 reviewed PRs. Adding a change to how findings are suppressed at merge time is exactly the kind of thing that should get its own review rather than ride along.

In the meantime nothing is misleading: the header of `.trivyignore` states plainly that the selectors are documentation only and that entries suppress their CVE image-wide, with PROWLER-2327 referenced inline. The four suppressed CVEs are all `perl-base` and all carry expiries, so the practical blast radius of the missing scoping is zero today.

The other open CodeRabbit threads are stale — they predate fixes already pushed to this branch. The `wget` one is the clearest example: the fragment it flags already reads "`wget` is retained because the Compose healthcheck invokes it".

One thing that has changed since you looked: the branch now also removes `pip` from the SDK and API images and this is measured, not estimated, on `--no-cache` rebuilds. pip vendors its own frozen `setuptools` and `msgpack`, which no dependency bump could reach. SDK 40 -> 38 high findings, API 39 -> 37.

cesararroba:
Closing out the open review threads. Each was either already fixed, verified empirically, or answered and ticketed.

| Thread | Disposition |
|---|---|
| `.trivyignore` ×2 — move to `.trivyignore.yaml` | **Valid, deferred.** Tracked as **PROWLER-2327**. Reasoning in [this comment](https://github.com/prowler-cloud/prowler/pull/12258#issuecomment-5165608839): it is a scanner-configuration change that must land in all four repos together, not a CVE fix to ride along in an integration branch. The file header already states the selectors are documentation only. |
| `api/changelog.d/api-container-purge-build-only` — stale `wget` claim | **Already fixed** (thread is marked outdated). The fragment now reads "`wget` is retained because the Compose healthcheck invokes it". |
| `api/Dockerfile` — does `xmlsec` survive the purge? | **Verified.** The purge removes `libxmlsec1t64` and `libxmlsec1t64-openssl`, but the `xmlsec` wheel is self-contained: `import xmlsec` resolves to `xmlsec.cpython-312-aarch64-linux-gnu.so`, and `xmlsec.init()` plus `Key.generate(AES, 128)` both succeed in the built image. SAML is unaffected. Good catch to check — this is the same class of bug as the `wget` healthcheck regression. |
| `api/pyproject.toml` — SDK pinned to the integration branch | **Intentional and documented.** It cannot point at `@master` until this merges; tracked as **PROWLER-2328** with the exact revert steps in the PR description. |
| `ui/next.config.js` — libvips in the standalone output | **Verified in #12274.** `outputFileTracingIncludes` was added precisely because sharp resolves its native binary at runtime and the tracer misses it. Confirmed by requesting `/_next/image` and checking the response is `image/webp` rather than the original file — the failure mode is silent (HTTP 200, no log). |

One addition since the last review: `pip` is now removed from both runtime images. It vendors frozen copies of `setuptools` 70.3.0 and `msgpack` that no dependency bump could reach — the lockfile already pins `setuptools` 82.0.1. SDK 40 → 38 high findings, API 39 → 37, measured on `--no-cache` rebuilds.

Also evaluated and **rejected**: PowerShell 7.5.9 → 7.6.4. The hypothesis was that .NET 10 would clear the `System.Security.Cryptography.Xml` findings, since the fix version is 10.0.10. It does not — 7.6.4 ships 10.0.6, which is still vulnerable, *and* the 8.0.3 entry from the MicrosoftTeams SBOM manifest remains. Net effect **38 → 43 highs**, so the image stays on 7.5.9.

AdriiiPRodri (approved):

pedrooot (approved):
👏🏼 👏🏼

PR #12267, opened 2026-07-31
fix(ci): harden osv-scan.sh against a missing osv-scanner.toml

cesararroba:
## Description

`.github/scripts/osv-scan.sh` has two bugs that only surface when `osv-scanner.toml` is absent. This repo has one, so CI has never hit them — they were found porting the script to `prowler-registry`, which does not, where both fired on the first run.

1. **`SCAN_ARGS[@]: unbound variable`** (line 54). `SCAN_ARGS` only gets `--config=` appended when the config file exists. Expanding an empty array trips `set -u` on bash before 4.4. Guarded with `${SCAN_ARGS[@]+"${SCAN_ARGS[@]}"}`.
2. **`[: : integer expression expected`** (line 111). `COUNT` is empty when `jq` receives no input, then reaches an integer comparison. Now defaults to `0` in both the assignment and the test.

## Why fix it if CI never hits it

GitHub runners have bash 5, so the first would not fire there even without the config. The second could.

The real cost is local runs. A developer reproducing a CI failure runs this script directly, and on macOS — bash 3.2 — gets two confusing errors that have nothing to do with the finding they are chasing. It also makes the script safe to reuse in a repo that has not defined an ignore config yet, which is exactly the situation that surfaced it.

## Verification

Reproduced against this branch by temporarily moving `osv-scanner.toml` aside:

```
--- current master, no config ---
  osv-scan.sh: line 54: SCAN_ARGS[@]: unbound variable
  osv-scan.sh: line 111: [: : integer expression expected
  osv-scanner: no findings at severity levels: CRITICAL

--- this branch, no config ---
  osv-scanner: no findings at severity levels: CRITICAL
```

With `osv-scanner.toml` in place the behaviour is unchanged. `shellcheck` clean.

Same fix already applied in prowler-cloud/prowler-registry#176.

## License

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved security scan reliability when optional scan arguments are empty.
  * Added safer handling when scan results cannot be parsed, preventing failures caused by missing or invalid finding counts.

### DIFF
--- a/.github/scripts/osv-scan.sh
+++ b/.github/scripts/osv-scan.sh
@@ -101,7 +101,9 @@ FINDINGS="$(printf '%s' "${OUTPUT}" | jq --argjson sevs "${SEVERITY_JSON}" '
   ]
 ')"
 
-COUNT="$(printf '%s' "${FINDINGS}" | jq 'length' 2>/dev/null || echo 0)"
+# jq exits 0 with no output on empty stdin, but non-zero on malformed JSON.
+# Let the failure abort under set -e rather than reporting zero findings.
+COUNT="$(printf '%s' "${FINDINGS}" | jq 'length')"
 
 # Write the findings JSON to OSV_REPORT_FILE so callers (e.g. the composite
 # action's PR-comment step) can consume the same data the gate decision uses.
--- a/.github/scripts/osv-scan.sh
+++ b/.github/scripts/osv-scan.sh
@@ -51,7 +51,8 @@ STDERR="$(mktemp)"
 trap 'rm -f "${STDERR}"' EXIT
 
 set +e
-OUTPUT="$(osv-scanner scan source "${SCAN_ARGS[@]}" --format=json "$@" 2>"${STDERR}")"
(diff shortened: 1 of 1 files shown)


cesararroba:
Good catch — fixed in b42db7c.

`2>/dev/null || echo 0` did convert a jq parse failure into "zero findings", which would let a malformed report pass the gate silently. Verified the three cases:

| Input | jq exit | Behaviour |
|---|---:|---|
| empty stdin | 0, no output | falls through to `${COUNT:-0}` → 0 findings |
| malformed JSON | 5 | now aborts under `set -e` |
| valid `[]` | 0, prints `0` | 0 findings |

So the default now applies only when jq succeeds, and a parse failure propagates. That also explains the original bug: the empty-stdin case exits 0, which is why `set -e` never caught it and the error surfaced later at the integer comparison instead.

The same weaker form went into prowler-cloud/prowler-registry#176, which is already merged — I'll carry this correction across in prowler-cloud/prowler-registry#177.

PR #12276, opened 2026-08-02
fix: skip ECS task definitions when describe fails

### LINKED ISSUE #12132: AWS ECS task-definition checks fail open: unexamined resources are reported as PASS
# ECS task-definition checks fail open: unexamined resources are reported as PASS

## Summary

When `DescribeTaskDefinition` fails for a task definition, `EcsService` swallows
the exception and leaves `container_definitions` empty. Every check that
iterates that list then reports **PASS**, because the checks initialize
`report.status = "PASS"` and only downgrade to `FAIL` inside the loop over
containers. An empty list means the loop never executes.

The result is a security check reporting a clean verdict for a resource it never
managed to examine. Because whether a given describe call fails varies between
runs, the same immutable resource can flip verdicts run to run.

Version: **5.22.0** (`prowlercloud/prowler:5.22.0`), AWS provider.

## Reproduction

An account with a large number of ACTIVE task-definition revisions (~4,900 in
our case, accumulated because nothing deregisters old revisions). Run the AWS
provider twice, a few minutes apart, and diff the OCSF output per resource.

Observed across two consecutive runs 20 minutes apart:

| | |
|---|---|
| Task-definition revisions scanned | 4,901 (identical set both runs) |
| Revisions that changed verdict | **2,100 (42.8%)** |
| FAIL → PASS | 1,172 |
| PASS → FAIL | 928 |
| Checks affected out of 128 run | **1** |

ECS task-definition revisions are immutable, so no resource actually changed.
The net difference in FAIL count was only −244, which is small enough to look
like ordinary run-to-run variance; the underlying churn is two orders of
magnitude larger.

Ground truth via `aws ecs describe-task-definition`: of 20 sampled revisions
that `ecs_task_definitions_containers_readonly_access` reported as **PASS**,
**20 had `readonlyRootFilesystem` unset** — which the check itself treats as
FAIL. All 20 were false passes.

Two revisions of the same family, in the same report, with identical
configuration:

```
revision 2   -> FAIL  "has containers with write access to the root filesystem"
revision 255 -> PASS  "does not have containers with write access to the root filesystems"
```

## Cause

`prowler/providers/aws/services/ecs/ecs_service.py`:

```python
def _describe_task_definition(self, task_definition):
    try:
        ...
        container_definitions = response["taskDefinition"]["containerDefinitions"]
        for container in container_definitions:
            task_definition.container_definitions.append(ContainerDefinition(...))
    except Exception as error:
        logger.error(...)          # swallowed -> container_definitions stays EMPTY
```

This is fanned out with `__threading_call__` over every task definition, so on
a large account it issues thousands of concurrent `DescribeTaskDefinition`
calls — exactly the conditions under which some will fail.

`prowler/providers/aws/services/ecs/ecs_task_definitions_containers_readonly_access/ecs_task_definitions_containers_readonly_access.py`:

```python
report.status = "PASS"                                   # optimistic default
for container in task_definition.container_definitions:  # empty -> body never runs
    if not container.readonly_rootfilesystem:
        report.status = "FAIL"
```

Failure to gather evidence is indistinguishable, in the output, from evidence of
compliance.

## Scope

Six checks read `container_definitions` and share the pattern:

- `ecs_task_definitions_containers_readonly_access`
- `ecs_task_definitions_host_namespace_not_shared`
- `ecs_task_definitions_host_networking_mode_users`
- `ecs_task_definitions_logging_enabled`
- `ecs_task_definitions_no_environment_secrets`
- `ecs_task_definitions_no_privileged_containers`

In our account the other five return 100% PASS, so a fail-open PASS is
indistinguishable from a correct PASS and the defect is invisible there rather
than absent. `..._readonly_access` is the only one whose true verdict is FAIL
for most resources, which is why it is the only one where the bug is observable
as a flip.

## Suggested fix

Two independent changes, either of which prevents the false PASS:

1. **Do not let a failed gather look like a successful one.** Record that the
   describe failed (e.g. a `gathered: bool` on the model, or leave
   `container_definitions = None` to distinguish "none" from "not retrieved"),
   and have the checks skip or report the resource as errored rather than PASS.

2. **Make the checks derive PASS rather than default to it.** Computing
   `failed = [c for c in containers if not c.readonly_rootfilesystem]` and
   setting the status from that still passes an empty list, so this alone is
   insufficient — but combined with (1) it makes the intent explicit.

Retrying throttled describes with backoff would reduce the frequency, but the
fail-open would remain: any residual error still yields PASS. The reporting
behavior is the part worth fixing.

## Happy to send a PR

If this looks right to you, I'm glad to open a pull request. I'd rather agree on
the approach first, since the fix touches a shared model: option (1) changes
`ContainerDefinition`/`EcsTaskDefinition` gathering semantics and so affects all
six checks, and I don't want to guess at how you'd prefer an
evidence-not-gathered resource to surface — skipped, `MANUAL`, or a new status.
Tell me which shape you want and I'll follow it, tests included.

## Notes

Reported per `SECURITY.md`, which directs functional and check issues to the
public issue tracker rather than the support desk. Account identifiers and
resource names have been generalized; I'm happy to share the OCSF diff privately
if that would help.

This report was researched and written by Claude (Anthropic), working in our
repository. I reviewed it, re-ran the ground-truth `describe-task-definition`
checks against our own AWS account, and am filing it deliberately — please treat
it as our report rather than as automated output, and hold it to that standard.
Any errors in it are ours to correct, so do push back if something does not
reproduce on your side.

heyncth:
## What

When `DescribeTaskDefinition` fails (throttling, permissions, transient errors), the ECS service previously left `container_definitions` as an empty list, so the task-definition checks evaluated the definition with no containers and reported **PASS** — an unexamined resource looked compliant.

## Why

`TaskDefinition.container_definitions` defaulted to `[]`, so "no containers" was indistinguishable from "not retrieved". Any describe failure silently produced a false PASS across the whole task-definition check family (e.g. `ecs_task_definitions_no_privileged_containers` reporting compliant on a definition whose privileged containers were never inspected).

## How

- `TaskDefinition.container_definitions` is now `Optional[list[ContainerDefinition]]` defaulting to `None`; `_describe_task_definition` populates it only on success.
- All seven task-definition checks (`containers_readonly_access`, `host_namespace_not_shared`, `host_networking_mode_users`, `logging_enabled`, `logging_block_mode`, `no_environment_secrets`, `no_privileged_containers`) skip task definitions whose describe call failed, and derive PASS/FAIL from the computed failed-container list, so a failed gather can never be reported as compliant.
- Regression tests simulate a throttled `DescribeTaskDefinition` and assert each check emits no finding for the unexamined resource.

## Verify

```
.venv/Scripts/python.exe -m pytest tests/providers/aws/services/ecs/ -q
```

Fixes #12132

## Summary by CodeRabbit

* **Bug Fixes**
  * ECS task-definition checks no longer report resources as compliant when their details cannot be retrieved.
  * Failed retrievals are skipped, while successfully examined definitions—including those with no containers—are evaluated correctly.
  * Findings and status messages now accurately distinguish compliant and non-compliant container configurations.

* **Tests**
  * Added regression coverage for failed task-definition retrievals and empty container definitions.

### DIFF
--- a/prowler/changelog.d/ecs-task-definitions-fail-open.security.md
+++ b/prowler/changelog.d/ecs-task-definitions-fail-open.security.md
@@ -1,0 +1,1 @@
+AWS ECS task-definition checks no longer report unexamined task definitions as PASS when DescribeTaskDefinition fails
--- a/prowler/providers/aws/services/ecs/ecs_service.py
+++ b/prowler/providers/aws/services/ecs/ecs_service.py
@@ -149,8 +149,8 @@ class ECS(AWSService):
                     "TAGS",
                 ],
             )
-            container_definitions = response["taskDefinition"]["containerDefinitions"]
-            for container in container_definitions:
+            container_definitions = []
+            for container in response["taskDefinition"]["containerDefinitions"]:
                 environment = []
                 if "environment" in container:
                     for env_var in container["environment"]:
@@ -159,7 +159,7 @@ class ECS(AWSService):
                                 name=env_var["name"], value=env_var["value"]
                             )
                         )
-                task_definition.container_definitions.append(
+                container_definitions.append(
                     ContainerDefinition(
                         name=container["name"],
                         privileged=container.get("privileged", False),
@@ -176,6 +176,7 @@ class ECS(AWSService):
                         .get("mode", ""),
                     )
                 )
+            task_definition.container_definitions = container_definitions
             task_definition.pid_mode = response["taskDefinition"].get("pidMode", "")
             task_definition.registered_at = response["taskDefinition"].get(
                 "registeredAt"
@@ -298,11 +299,31 @@ class ContainerDefinition(BaseModel):
 
 
 class TaskDefinition(BaseModel):
+    """ECS task definition model.
+
+    Attributes:
+        name: Task definition family name, without the revision suffix.
+        arn: Full task definition ARN.
+        revision: Task definition revision.
+        region: AWS region where the task definition was listed.
+        container_definitions: Container definitions populated only by a
+            successful ``DescribeTaskDefinition`` call.  ``None`` means the
+            call failed and the evidence was never gathered; ``[]`` means a
+            successful describe with no containers.  Checks must skip task
+            definitions whose evidence is ``None`` instead of reporting an
+            unexamined resource as compliant.
+        pid_mode: PID mode of the task definition (``"host"`` when the host
+            process namespace is shared with the containers).
+        registered_at: Registration timestamp of the task definition.
+        tags: Tags attached to the task definition.
+        network_mode: Network mode of the task definition.
+    """
+
     name: str
     arn: str
     revision: str
     region: str
-    container_definitions: list[ContainerDefinition] = []
+    container_definitions: Optional[list[ContainerDefinition]] = None
     pid_mode: Optional[str]
     registered_at: Optional[datetime] = None
     tags: Optional[list] = []
--- a/prowler/providers/aws/services/ecs/ecs_task_definitions_containers_readonly_access/ecs_task_definitions_containers_readonly_access.py
+++ b/prowler/providers/aws/services/ecs/ecs_task_definitions_containers_readonly_access/ecs_task_definitions_containers_readonly_access.py
@@ -3,24 +3,38 @@ from prowler.providers.aws.services.ecs.ecs_client import ecs_client
 
 
 class ecs_task_definitions_containers_readonly_access(Check):
-    def execute(self):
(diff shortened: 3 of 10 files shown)


HugoPBrito:
Thank you for the thoughtful work on this fix. #12217 was already open for the same issue, so we consolidated the implementation there. Your atomic container-evidence assignment was incorporated in [`83b3be0`](https://github.com/prowler-cloud/prowler/commit/83b3be0e8d8c8fbdd0ed0b7b97804a16a29f3c55), with the regression coverage completed in #12217.

Closing this PR as a duplicate while preserving your contribution in the original PR. Thanks again.

PR #12284, opened 2026-08-03
fix(ui): guard Overview Lighthouse context against API error responses

alejandrobailo:
### Context

Follow-up to #12220. `handleApiResponse` resolves **truthy** objects for non-5xx failures — `{error, status}` for 4xx and `{success, status}` for 204/empty bodies — so the `if (!findingsByStatus)` / `if (!findingsBySeverity)` guards in the Overview status and severity contributors never fire on those responses. The downstream `?.data?.attributes || {}` then collapses to `{}` and the destructuring defaults publish a fabricated `0 failed / 0 passed` posture and an all-zero severity breakdown to Lighthouse, while the user sees the error card triggers (e.g. a stale bookmark carrying an invalid `filter[...]` param → 400, or a token that crosses expiry during render → 401).

### Description

- `check-findings.ssr.tsx` and `risk-severity-chart.ssr.tsx` now guard on `?.data?.attributes` — the same pattern the ThreatScore, watchlist, and provider-context contributors already use — so both render their error state instead of contributing zeros.
- `threat-score.ssr.tsx` guards `critical_requirements?.length ?? 0`. The field is always present with the current API serializer, but this component is an async Server Component inside `<Suspense>` (which does not catch render errors), so an absent field would have taken down the whole Overview page rather than one widget.
- Error-path tests for every Overview Lighthouse contributor, covering both truthy no-data shapes (`{error, status: 400}` and `{success: true, status: 204}`): status summary, severity summary, ThreatScore, service watchlist, and the compliance-watchlist adapter. These are the regression guard for the failure mode this PR fixes.

### Steps to review

1. In `ui/lib/server-actions-helper.ts`, confirm `handleApiResponse` returns truthy objects for 4xx (`{error, status}`) and for 204/empty bodies (`{success, status}`) — only 5xx throws and only network errors yield `undefined`.
2. Check the two guard changes render the existing "Failed to load…" cards under those shapes and skip the `LighthouseContextContributor`.
3. Run the targeted tests:
   ```bash
   cd ui && pnpm vitest run --project unit "app/(prowler)/_overview" "actions/overview/compliance-watchlist"
   ```

### Checklist

#### UI

- [ ] Screenshots/Video — not applicable: no visual change, the existing error cards now render under previously unhandled response shapes.
- [ ] Changelog fragment — not applicable (`no-changelog`, Lighthouse is cloud-gated; same as #12220).

### License

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved overview error handling for failed, empty, or incomplete responses.
  * Findings and severity charts now display error states instead of rendering invalid data.
  * Threat score safely shows zero when critical requirements data is unavailable.
  * Service and compliance watchlists avoid publishing invalid context when data cannot be loaded.

* **Tests**
  * Added coverage for error responses, empty responses, and missing data across overview components.

### DIFF
--- a/ui/actions/overview/compliance-watchlist/compliance-watchlist.adapter.test.ts
+++ b/ui/actions/overview/compliance-watchlist/compliance-watchlist.adapter.test.ts
@@ -1,0 +1,30 @@
+import { describe, expect, it } from "vitest";
+
+import { adaptComplianceWatchlistResponse } from "./compliance-watchlist.adapter";
+import type { ComplianceWatchlistResponse } from "./compliance-watchlist.types";
+
+describe("adaptComplianceWatchlistResponse", () => {
+  it("returns no items on a 4xx error shape", () => {
+    // handleApiResponse resolves truthy {error, status} objects for 4xx.
+    const errorResponse = {
+      error: "Invalid filter",
+      status: 400,
+    } as unknown as ComplianceWatchlistResponse;
+
+    expect(adaptComplianceWatchlistResponse(errorResponse)).toEqual([]);
+  });
+
+  it("returns no items on an empty-body success shape", () => {
+    // handleApiResponse resolves {success, status} for 204 and empty bodies.
+    const emptyResponse = {
+      success: true,
+      status: 204,
+    } as unknown as ComplianceWatchlistResponse;
+
+    expect(adaptComplianceWatchlistResponse(emptyResponse)).toEqual([]);
+  });
+
+  it("returns no items when the fetch failed with undefined", () => {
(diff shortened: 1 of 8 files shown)


Alan-TheGentleman (commented):
Approved from my side, will wait for Pablo. The API-error and empty-response guards correctly prevent Overview contributors from publishing fabricated Lighthouse context.

One non-blocking cleanup: the missing `critical_requirements` test bypasses the current API contract with a double assertion. The serializer, model, and UI type all require this field, so I would restore the required `.length` access and remove that speculative fixture. If omission becomes supported later, the count should be represented as unknown rather than zero.

Alan-TheGentleman (approved):

PR #12285, opened 2026-08-03
feat(ui): capture Lighthouse context on tenant admin pages

alejandrobailo:
### Context

Follow-up to #12220. Routes without a Lighthouse page definition — Roles, Users, Invitations, Integrations, Manage Groups, Profile — fall back to `createFallbackDefinition`, which registers `allowedSearchParams: []`. The context chip still renders on them (the envelope always carries the page item), so the user sees "@ Roles" and reasonably assumes their view travels with the question, when the model only receives the route label and path — no search filters, no data.

### Description

- Registers page definitions for **Roles, Users, Invitations, and Integrations** with their real search params (`filter[search]`/`sort`; plus `filter[integration_type]` on Integrations) and four tailored suggested prompts each, following the existing `{label, prompt}` shape.
- Sub-routes resolve to their section (`/roles/new` → roles, `/integrations/jira` → integrations), matching how Scans and Attack Paths already match by prefix.
- `LIGHTHOUSE_KNOWN_ROUTE_LABELS` now covers `manage-groups` (the only remaining fallback route whose label needs declaring); Integrations left the map because it is a registered page now.
- **Honest fallback tooltip**: when the envelope contains a single filterless page item, the chip tooltip states "Only the current page name is shared." so unregistered pages stop implying richer context than they send.

### Steps to review

1. `cd ui && pnpm vitest run --project unit "lib/lighthouse/context" "components/lighthouse/context-chip.test.tsx"`
2. Open `/roles?filter[search]=admin` and hover the Lighthouse chip: the tooltip lists `search: admin` and the envelope carries it.
3. Open `/manage-groups` and hover the chip: the tooltip reads "Only the current page name is shared."
4. Check the four suggestion sets read naturally in the Lighthouse panel on each new page.

### Checklist

#### UI

- [ ] Screenshots/Video — tooltip copy change only; no layout change.
- [ ] Changelog fragment — not applicable (`no-changelog`, Lighthouse is cloud-gated; same as #12220).

### License

## Summary by CodeRabbit

* **New Features**
  * Added Lighthouse context support for Roles, Users, Invitations, and Integrations pages.
  * Added page-specific suggestions and support for relevant filters, sorting, and search parameters.
  * Improved route labeling for clearer navigation context.
  * Context tooltips now clarify when only a filterless page is shared.

* **Bug Fixes**
  * Improved preservation and resolution of page filters, sorting, and search context.
  * Empty filters no longer appear in tooltips or affect page-only context detection.

### DIFF
--- a/ui/components/lighthouse/context-chip.tsx
+++ b/ui/components/lighthouse/context-chip.tsx
@@ -96,7 +96,10 @@ function LighthouseContextTooltip({ context }: LighthouseContextBadgeProps) {
     .filter((description) => description !== null);
   // A single filterless page item means the model only learns which page the
   // user is on — say so instead of implying richer context travels with it.
-  const sharesOnlyPage = context.items.length === 1 && !filters;
+  const sharesOnlyPage =
+    page?.kind === LIGHTHOUSE_CONTEXT_KIND.PAGE &&
+    context.items.length === 1 &&
+    !filters;
 
   return (
     <TooltipContent maxWidth="md">
--- a/ui/components/lighthouse/context-chip.tsx
+++ b/ui/components/lighthouse/context-chip.tsx
@@ -85,12 +85,17 @@ function LighthouseContextTooltip({ context }: LighthouseContextBadgeProps) {
   const page = context.items.find(
     (item) => item.kind === LIGHTHOUSE_CONTEXT_KIND.PAGE,
   );
-  const filters =
+  // Entries with no values (possible in stored envelopes) carry nothing, so
+  // they count neither for the filters line nor against the page-only notice.
+  const filterEntries =
     page?.kind === LIGHTHOUSE_CONTEXT_KIND.PAGE
-      ? Object.entries(page.filters ?? {})
-          .map(([key, values]) => `${key}: ${values.join(", ")}`)
-          .join("; ")
-      : "";
+      ? Object.entries(page.filters ?? {}).filter(
+          ([, values]) => values.length > 0,
+        )
+      : [];
+  const filters = filterEntries
+    .map(([key, values]) => `${key}: ${values.join(", ")}`)
+    .join("; ");
   const itemDescriptions = context.items
     .map(getContextItemDescription)
     .filter((description) => description !== null);
@@ -99,7 +104,7 @@ function LighthouseContextTooltip({ context }: LighthouseContextBadgeProps) {
   const sharesOnlyPage =
     page?.kind === LIGHTHOUSE_CONTEXT_KIND.PAGE &&
     context.items.length === 1 &&
-    !filters;
+    filterEntries.length === 0;
 
   return (
     <TooltipContent maxWidth="md">
--- a/ui/components/lighthouse/context-chip.tsx
+++ b/ui/components/lighthouse/context-chip.tsx
@@ -94,6 +94,9 @@ function LighthouseContextTooltip({ context }: LighthouseContextBadgeProps) {
   const itemDescriptions = context.items
     .map(getContextItemDescription)
     .filter((description) => description !== null);
+  // A single filterless page item means the model only learns which page the
+  // user is on — say so instead of implying richer context travels with it.
+  const sharesOnlyPage = context.items.length === 1 && !filters;
 
   return (
     <TooltipContent maxWidth="md">
@@ -104,6 +107,7 @@ function LighthouseContextTooltip({ context }: LighthouseContextBadgeProps) {
           </p>
         )}
         {filters && <p>Filters: {filters}</p>}
+        {sharesOnlyPage && <p>Only the current page name is shared.</p>}
         {itemDescriptions.map(({ id, text }) => (
           <p key={id}>{text}</p>
         ))}
--- a/ui/lib/lighthouse/context/pages.constants.ts
+++ b/ui/lib/lighthouse/context/pages.constants.ts
@@ -418,8 +418,120 @@ export const LIGHTHOUSE_PAGE_DEFINITION_INPUTS = [
       },
     ],
   },
+  {
+    id: LIGHTHOUSE_PAGE_ID.ROLES,
+    label: "Roles",
+    match: (pathname) => pathname.startsWith("/roles"),
+    allowedSearchParams: [...COMMON_LIST_PARAMS],
+    suggestions: [
+      {
+        label: "Review role permissions",
+        prompt:
+          "Review the roles on this page. Which ones grant more than their name implies, and what would you trim first?",
+      },
+      {
+        label: "Find overlapping roles",
+        prompt:
+          "Which of these roles overlap enough to be merged, and does any combination of them add up to an escalation risk?",
+      },
+      {
+        label: "Design least privilege",
+        prompt:
+          "Propose a least-privilege role set for my team: which roles should exist, what should each allow, and how do I migrate from the current ones?",
+      },
+      {
+        label: "Audit unlimited visibility",
+        prompt:
+          "Which roles have unlimited visibility across providers, and who actually needs to keep it?",
+      },
+    ],
+  },
+  {
+    id: LIGHTHOUSE_PAGE_ID.USERS,
+    label: "Users",
+    match: (pathname) => pathname.startsWith("/users"),
+    allowedSearchParams: [...COMMON_LIST_PARAMS],
+    suggestions: [
+      {
+        label: "Audit user access",
+        prompt:
+          "Review the users listed here. Who holds admin-level roles, and does anyone look over-privileged for what they do?",
+      },
+      {
+        label: "Find concentrated ownership",
+        prompt:
+          "Is any critical responsibility — provider management, user administration, billing — concentrated in a single person here?",
+      },
+      {
+        label: "Plan an access review",
+        prompt:
+          "Turn this user list into a quarterly access review: who reviews whom, and what should each reviewer verify?",
+      },
+      {
+        label: "Prepare safe offboarding",
+        prompt:
+          "Draft an offboarding checklist for this tenant: what should be revoked or transferred when one of these users leaves?",
+      },
+    ],
+  },
+  {
+    id: LIGHTHOUSE_PAGE_ID.INVITATIONS,
+    label: "Invitations",
+    match: (pathname) => pathname.startsWith("/invitations"),
+    allowedSearchParams: [...COMMON_LIST_PARAMS],
+    suggestions: [
+      {
+        label: "Review pending invitations",
+        prompt:
+          "Which pending invitations here should be revoked — expired, unfamiliar domains, or broader roles than the invitee needs?",
+      },
+      {
+        label: "Tighten invitation roles",
+        prompt:
+          "Do any open invitations grant admin or unlimited-visibility roles? Suggest a safer default role for new members.",
+      },
+      {
+        label: "Draft onboarding steps",
+        prompt:
+          "Draft the checklist a new member should follow after accepting one of these invitations.",
+      },
+      {
+        label: "Explain invitation lifecycle",
+        prompt:
+          "Explain how invitations work in Prowler: what happens when one expires, and how do I resend or revoke one safely?",
+      },
+    ],
+  },
+  {
+    id: LIGHTHOUSE_PAGE_ID.INTEGRATIONS,
+    label: "Integrations",
+    match: (pathname) => pathname.startsWith("/integrations"),
+    allowedSearchParams: [...COMMON_LIST_PARAMS, "filter[integration_type]"],
+    suggestions: [
+      {
+        label: "Review integration coverage",
+        prompt:
+          "Which integrations am I missing that would get findings to the right team faster — ticketing, chat, or storage?",
+      },
+      {
+        label: "Check integration health",
+        prompt:
+          "Review the integrations configured here. Are any disconnected, misconfigured, or pointing at destinations nobody watches?",
+      },
+      {
+        label: "Route findings to teams",
+        prompt:
+          "Design how findings should flow from Prowler into my existing tools so each team only sees what they own.",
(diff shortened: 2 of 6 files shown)


ui/lib/lighthouse/context/pages.constants.ts
```diff
@@ -418,8 +418,120 @@ export const LIGHTHOUSE_PAGE_DEFINITION_INPUTS = [
...
+        label: "Explain invitation lifecycle",
+        prompt:
+          "Explain how invitations work in Prowler: what happens when one expires, and how do I resend or revoke one safely?",
+      },
+    ],
+  },
+  {
+    id: LIGHTHOUSE_PAGE_ID.INTEGRATIONS,
+    label: "Integrations",
+    match: (pathname) => pathname.startsWith("/integrations"),
```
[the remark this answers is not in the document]
alejandrobailo: Fixed in `8fe90b21f`. The four matchers now require a route-segment boundary (`pathname === "/roles" || pathname.startsWith("/roles/")`), with a regression test covering `/roles-preview`-style prefixes resolving to the fallback.

alejandrobailo:
Regarding the outside-diff comment on `context-chip.tsx` (page-only message shown for a single non-page item): also fixed in `8fe90b21f` — `sharesOnlyPage` now requires the page item to be present, with a test for a one-item non-page envelope.

ui/components/lighthouse/context-chip.tsx
```diff
@@ -94,6 +94,12 @@ function LighthouseContextTooltip({ context }: LighthouseContextBadgeProps) {
   const itemDescriptions = context.items
     .map(getContextItemDescription)
     .filter((description) => description !== null);
+  // A single filterless page item means the model only learns which page the
+  // user is on — say so instead of implying richer context travels with it.
+  const sharesOnlyPage =
+    page?.kind === LIGHTHOUSE_CONTEXT_KIND.PAGE &&
+    context.items.length === 1 &&
+    !filters;
```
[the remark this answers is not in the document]
alejandrobailo: Fixed in `fa58f2bd3`. Rather than a parallel `hasFilters` boolean, empty-value entries are now filtered out once (`filterEntries`) and both the Filters line and the page-only notice derive from it — so `{ status: [] }` neither renders a dangling "Filters: status:" line nor suppresses the notice. Test added for that envelope shape.

ui/lib/lighthouse/context/pages.constants.ts
```diff
@@ -418,8 +418,124 @@ export const LIGHTHOUSE_PAGE_DEFINITION_INPUTS = [
       },
     ],
   },
+  {
+    id: LIGHTHOUSE_PAGE_ID.ROLES,
+    label: "Roles",
+    match: (pathname) =>
+      pathname === "/roles" || pathname.startsWith("/roles/"),
+    allowedSearchParams: [...COMMON_LIST_PARAMS],
```
pfe-nazaries: `allowedSearchParams` is bare `COMMON_LIST_PARAMS` (`query`, `search`, `filter[search]`, `sort`), but this page also ships a dropdown filter that never reaches Lighthouse:

- `app/(prowler)/roles/page.tsx` renders `DataTableFilterCustom filters={filterRoles}`
- `components/filters/data-filters.ts:109` defines that filter with key `permission_state`
- `components/shadcn/table/data-table-filter-custom.tsx` writes it to the URL as `filter[permission_state]`

`buildFilters` iterates only the allow-list, so the filter is silently dropped. On `/roles?filter[permission_state]=unlimited` the chip still reads just "Roles", and the suggestion this PR itself adds — "Audit unlimited visibility" — gets answered against *all* roles rather than the filtered view on screen. That's the same bug class the PR sets out to fix.

```suggestion
    allowedSearchParams: [...COMMON_LIST_PARAMS, "filter[permission_state]"],
```

Worth a `pages.test.ts` assertion mirroring the existing integration-type test.

ui/lib/lighthouse/context/pages.constants.ts
```diff
@@ -418,8 +418,124 @@ export const LIGHTHOUSE_PAGE_DEFINITION_INPUTS = [
...
+          "Draft an offboarding checklist for this tenant: what should be revoked or transferred when one of these users leaves?",
+      },
+    ],
+  },
+  {
+    id: LIGHTHOUSE_PAGE_ID.INVITATIONS,
+    label: "Invitations",
+    match: (pathname) =>
+      pathname === "/invitations" || pathname.startsWith("/invitations/"),
+    allowedSearchParams: [...COMMON_LIST_PARAMS],
```
pfe-nazaries: Same gap as Roles: bare `COMMON_LIST_PARAMS` omits this page's dropdown filter.

- `app/(prowler)/invitations/page.tsx` renders `DataTableFilterCustom filters={filterInvitations}`
- `components/filters/data-filters.ts:100` defines that filter with key `state`
- it lands in the URL as `filter[state]`

So on `/invitations?filter[state]=expired` the filter is dropped and the "Review pending invitations" suggestion added just below is answered against every invitation, not the pending/expired subset the user is looking at.

```suggestion
    allowedSearchParams: [...COMMON_LIST_PARAMS, "filter[state]"],
```

Note `filterUsers` also exists in `data-filters.ts:91`, but `users/page.tsx` never renders `DataTableFilterCustom` — the Users definition is correctly scoped to search + sort as-is.

ui/lib/lighthouse/context/pages.constants.ts
```diff
@@ -418,8 +418,124 @@ export const LIGHTHOUSE_PAGE_DEFINITION_INPUTS = [
...
+          "Explain how invitations work in Prowler: what happens when one expires, and how do I resend or revoke one safely?",
+      },
+    ],
+  },
+  {
+    id: LIGHTHOUSE_PAGE_ID.INTEGRATIONS,
+    label: "Integrations",
+    match: (pathname) =>
+      pathname === "/integrations" || pathname.startsWith("/integrations/"),
+    allowedSearchParams: [...COMMON_LIST_PARAMS, "filter[integration_type]"],
```
pfe-nazaries: Non-blocking, but this allow-list can't fire — worth deciding rather than leaving as-is.

`app/(prowler)/integrations/page.tsx` is `export default async function Integrations()` with **no props at all** — no `searchParams`, and the page is a static card grid with no search, sort, or filter control. So every entry here is inert on `/integrations` itself, not just `filter[integration_type]`.

The sub-pages do read params, but they set the integration type server-side and then explicitly discard any URL-supplied copy — `integrations/jira/page.tsx`:

```ts
urlSearchParams.set("filter[integration_type]", "jira");   // server-side
// ...
if (value !== undefined && key !== "filter[integration_type]") { ... }   // URL copy dropped
```

Same in `amazon-s3/page.tsx` and `aws-security-hub/page.tsx`. The param therefore never appears in the browser URL, and is ignored if hand-injected. Only `sort` actually reaches Lighthouse from the sub-pages.

Knock-on effect: the new test at `pages.test.ts:203-211` hand-builds `new URLSearchParams({"filter[integration_type]": "jira"})` — a URL state the app is coded to prevent. It passes while proving nothing about this page; it's really a `buildFilters` mapping test.

Two ways out:
1. **Trim to reality** — `allowedSearchParams: ["sort"]`, and delete or re-scope that test.
2. **Keep it forward-looking** — leave the param with a comment marking it pre-registered for an integrations filter UI that doesn't exist yet.

Either is fine; the current state is the one that misleads. Nothing is actually lost by trimming — the full path (`/integrations/jira`) already travels in the item's `path` field, so Lighthouse still knows which integration is open.

ui/lib/lighthouse/context/pages.constants.ts
```diff
@@ -418,8 +418,124 @@ export const LIGHTHOUSE_PAGE_DEFINITION_INPUTS = [
...
+        prompt:
+          "What credentials do these integrations rely on, and how should I scope and rotate them safely?",
+      },
+    ],
+  },
 ] as const satisfies readonly LighthousePageDefinitionInput[];
 
 export const LIGHTHOUSE_KNOWN_ROUTE_LABELS = {
-  integrations: "Integrations",
+  "manage-groups": "Manage Groups",
```
pfe-nazaries: Non-blocking. Dropping `integrations` here is right (it's a registered page now, so it never reaches the fallback), but the replacement entry is a no-op.

This map's only consumer is `createFallbackDefinition` (`pages.ts:104-109`):

```ts
const label = LIGHTHOUSE_KNOWN_ROUTE_LABELS[segment] ?? toTitleCase(segment);
```

It's an **override** map — it only has effect when its value differs from what `toTitleCase` computes. And `toTitleCase` (`pages.ts:190-199`) splits on `-`, capitalises each part, rejoins with spaces:

```
toTitleCase("manage-groups")  →  "Manage Groups"    // identical to the new entry
toTitleCase("integrations")   →  "Integrations"     // identical to the removed one
```

So this swaps one no-op for another. The knock-on is the new test at `pages.test.ts:63-69`: it's named "should use the known route label for fallback pages that declare one", but it passes unchanged with the map deleted — it verifies `toTitleCase`, not the map. A future maintainer will read the entry as load-bearing.

Where the map *would* earn its place is segments mechanical title-casing gets wrong:

```
toTitleCase("sso")       →  "Sso"       // want "SSO"
toTitleCase("mutelist")  →  "Mutelist"  // want "Mute List"
```

Three options: populate it with a genuine case like those (then the test means something), delete the map plus its `?? toTitleCase(segment)` branch, or keep it `{}` with a comment naming it the acronym/multi-word override hook. Any of the three beats the current shape.

pfe-nazaries (changes requested):
Nice catch on the underlying problem — four admin routes falling through to `createFallbackDefinition()` with `allowedSearchParams: []` meant the chip promised context it wasn't sending, and the "Only the current page name is shared." notice is the right fix for the honesty half. The branch is also visibly self-correcting: `8fe90b21f` bounding the matchers to `=== "/x" || startsWith("/x/")` and gating `sharesOnlyPage` on `page?.kind`, then `fa58f2bd3` filtering empty filter entries, both with regression tests.

What I verified locally on `fa58f2bd3`: the two touched test files pass 45/45, `tsc --noEmit` and `eslint` are clean, route boundaries hold (`/roles-preview`, `/rolesx`, `/users-report`, `/integrations-beta` all correctly fall to `other`, while `/roles/`, `/users/abc`, `/integrations/jira` resolve), `sharesOnlyPage` is correctly suppressed for a lone non-page item, and the email scrub works on Users/Invitations. I traced the emitters too: `filter[search]` (via `custom-search-input` → `use-url-filters`) and `sort` (via `data-table-column-header`) genuinely travel now, so the core mechanism works.

Requesting changes on one point: **the allow-lists for Roles and Invitations omit the dropdown filters those two pages actually ship** (`filter[permission_state]` and `filter[state]`), so on exactly the filtered views the new suggestions target, the filter is still silently dropped. Details and suggested diffs inline.

Two further comments (`filter[integration_type]`, `LIGHTHOUSE_KNOWN_ROUTE_LABELS`) are **non-blocking** — no user-visible defect, but each pairs dead config with a test that reads as covering it and doesn't. Your call on which way to resolve them; happy either way.

One thing I couldn't check: I had no browser run (these routes need auth plus a seeded tenant), so the filter findings rest on tracing `DataTableFilterCustom` → `use-url-filters`, not on a live click. Worth confirming with one click on the Roles "Permissions" dropdown before you push the fix.

Alan-TheGentleman (commented):
Two active page filters are missing from the Lighthouse allow-lists.

1. **Roles drops `filter[permission_state]`.** The page exposes and forwards this filter, and the API supports it. On an unlimited-only view, Lighthouse receives context for all roles instead of the visible subset.

2. **Invitations drops `filter[state]`.** An expired or pending filtered view reaches Lighthouse without the selected invitation state.

Please add both parameters to their respective page definitions and cover them with `buildLighthousePageContext` assertions. The route boundaries, registration cleanup, empty-filter behavior, typecheck, lint, and focused tests otherwise look sound.

ui/lib/lighthouse/context/pages.constants.ts
alejandrobailo: Good catch — fixed in `eb59ac580`. `filter[permission_state]` is now on the Roles allow-list, and the tenant-admin filter test asserts it maps to `permission_state: ["unlimited"]`.

ui/lib/lighthouse/context/pages.constants.ts
alejandrobailo: Fixed in `eb59ac580`. `filter[state]` added to the Invitations allow-list with a test asserting `state: ["expired"]` reaches the page context. Thanks for checking Users too — left as-is since the page renders no custom filter.

ui/lib/lighthouse/context/pages.constants.ts
alejandrobailo: Went with option 1 (trim to reality) in `eb59ac580`: the Integrations allow-list is now `["sort"]` with a comment explaining why, since as you note the open integration already travels in the page item's `path`. The misleading `filter[integration_type]` test is replaced by one that exercises `/integrations/jira` and asserts the pinned param is *not* captured while `sort` is.

ui/lib/lighthouse/context/pages.constants.ts
alejandrobailo: Agreed the entry was a no-op — took your deletion option in `2601fd9fa`: `LIGHTHOUSE_KNOWN_ROUTE_LABELS` and its `?? toTitleCase(segment)` branch are gone, and `createFallbackDefinition` title-cases directly. The test is renamed to what it actually verifies ("should title-case multi-segment fallback routes"). If a real override case shows up later (`sso` → "SSO"), the map can come back with a genuine entry and a meaningful test.

pfe-nazaries (approved):
LGTM

PR #12294, opened 2026-08-03
fix(sdk): resolve entry-point checks on built-in providers

StylusFrost:
### Context

A check registered through the `prowler.checks.<provider>` entry-point group can never be executed when the target provider is built-in — `aws`, `azure`, `gcp`, `kubernetes` and the rest. The check is discovered, listed and selected for the run, and then fails to load. The scan completes normally and simply produces no findings for it.

The failure is silent. Progress reaches 100%, the finding count looks ordinary, and the only trace is a single line in the log:

```
ERROR  Check 'ec2_acme_instance_has_owner_tag' was not found for the AWS provider
```

This affects the whole "external checks for a built-in provider" capability, including the example shipped in `prowler-provider-template`: its `s3_bucket_has_cost_center_tag` fails the same way.

### Description

`_resolve_check_module` probes the built-in location before falling back to entry points:

```python
if is_builtin_provider(provider_type):
    builtin_path = f"prowler.providers.{provider_type}.services.{service}.{check_name}.{check_name}"
    if importlib.util.find_spec(builtin_path) is not None:   # raises, never returns None
        return import_check(builtin_path)

for ep in importlib.metadata.entry_points(group=f"prowler.checks.{provider_type}"):  # unreachable
    ...
```

`find_spec` imports the parent package in order to search it. For a plug-in check that parent — `prowler.providers.<provider>.services.<service>.<check_name>` — does not exist, so it raises `ModuleNotFoundError` rather than returning `None`. The exception leaves the resolver before the entry points are consulted, and the caller's `except ModuleNotFoundError` turns it into a logged "was not found" and a `continue`.

Minimal reproduction:

```python
>>> import importlib.util
>>> importlib.util.find_spec(
...     "prowler.providers.aws.services.s3.s3_bucket_has_cost_center_tag.s3_bucket_has_cost_center_tag"
... )
ModuleNotFoundError: No module named 'prowler.providers.aws.services.s3.s3_bucket_has_cost_center_tag'
```

This is the symmetric half of the defect already fixed for external providers, where the absent parent was `prowler.providers.<provider>`. It is fixed the same way — behind a leaf helper next to `is_builtin_provider` — so no raw `find_spec` over a `prowler.providers.*` path is left in the check module.

`is_builtin_check` narrows the exception instead of swallowing every `ImportError` like its sibling. Callers rely on this probe to tell *"the check is not built-in"* from *"the check is built-in and its imports are broken"*, and flattening the two would report a missing dependency as a missing check — the same class of silent failure this PR removes. It re-raises unless the unresolved module is the check's own path.

Collisions are unaffected: a plug-in whose `CheckID` matches a built-in still loses to the built-in, as before.

### Steps to review

1. **The bug is real and reachable from both entry points.** `_resolve_check_module` has two call sites in `execute_checks` — the `--only-logs` path and the regular CLI path — and both wrap it in the same `except ModuleNotFoundError`. Neither can reach the entry-point branch for a built-in provider today.

2. **Why no existing test caught it.** The resolution tests in `tests/providers/external/test_dynamic_provider_loading.py` patch `find_spec` and hand it `None`, which is precisely the behaviour the real call does not have; and the fallback case they cover uses an external *provider*, not an external *check* on a built-in provider. That cell of the matrix was empty. The new `test_resolve_check_module_entry_point_check_on_builtin_provider` deliberately leaves the probe unmocked — revert the change in `builtin.py` and it fails.

3. **The narrowing is load-bearing.** `test_reraises_when_a_builtin_checks_own_dependency_is_missing` pins it: a `ModuleNotFoundError` naming something other than the check's own path must still propagate.

4. Existing patch targets were repointed from `check.importlib.util.find_spec` to `check.is_builtin_check` — patching the old target would now silently no-op.

Suite: `tests/lib/check`, `tests/lib/scan`, `tests/providers/external` and `tests/providers/common` — 3778 passed.

### Checklist

- [x] Review if is needed to change the [README.md](https://github.com/prowler-cloud/prowler/blob/master/README.md) — not needed

#### SDK/CLI

### License

## Summary by CodeRabbit

- **Bug Fixes**
  - Built-in checks now run reliably against built-in providers.
  - Checks provided through extensions can load when no built-in check is available.
  - Missing checks are distinguished from genuine import and dependency errors.
  - Built-in checks retain priority over alternative checks.
  - Improved handling prevents missing external packages from blocking check discovery.

- **Tests**
  - Expanded coverage for provider and check detection, fallback behavior, precedence, and error handling.


danibarranqueroo (approved):
Thanks for this! 🙌

PR #12299, opened 2026-08-03
fix(ui): keep page-size selector when table fits one page

pfe-nazaries:
### Context

Reported on `/providers`: with a page size of 50 or 100 on a tenant holding ~81-100 providers, the table footer is unreachable — no `Page X of Y`, no navigation arrows and, critically, no `Rows per page` selector. Since the selector is the only UI that can lower the page size, the user is trapped at that size and the only way out is editing `?pageSize=` by hand in the URL.

The initial suspicion was a `max-height` clipping the card. There is none: the chain from `<body>` down to the pagination row carries no `max-h-*`, `h-[…]`, `min-h-*` or `overflow-hidden` that could cut it, and `<main>` is a `100dvh` scroller the card grows freely inside.

The cause is `DataTablePagination` unmounting wholesale:

```tsx
if (totalPages <= 1) return null;
```

A large page size collapses the table to a single page, which takes the selector down along with the navigation it was meant to gate. This lives in the shared `DataTable`, so it affects every table in the app — `/providers` is just where the row count lands in the window that triggers it.

### Description

`ui/components/shadcn/table/data-table-pagination.tsx`:

- The selector and the page navigation are now gated separately. The navigation still hides when there is a single page; the selector stays as long as some configured page size would split the table.
- The component returns `null` only when both are hidden, preserving the existing guarantee that no empty wrapper contributes a stray gap to `DataTable`'s flex column.
- The selector threshold now comes from the smallest configured option instead of a hardcoded `10`, so tables offering 5 per page (`actions/attack-paths/scans.adapter.ts`) behave correctly.
- `selectedPageSize` is derived from the URL (or `controlledPageSize`) instead of being seeded once with `useState`. That stale-state bug was masked by the footer disappearing; with the selector now persisting, back/forward or a filter reset would otherwise leave the trigger showing a size the table is no longer using.

No layout, styling or height changes.

### Steps to review

Quickest repro on any list page, no large tenant needed — the trigger is "everything fits on one page, but a smaller option exists":

1. Open a page whose table holds 11-19 entries (e.g. `/providers`, `/scans`, `/findings`).
2. Set `Rows per page` to 20 → one page.
3. **Before:** the whole footer disappears, including the selector, with no way back to 10.
   **After:** the selector stays (showing `20`); only `Page X of Y` and the arrows are gone, which is correct — there is nowhere to navigate.
4. Lower it back to 10 → the navigation returns.
5. Press browser Back → the trigger follows the URL instead of keeping the previous value.

The original report reproduces the same way with 81-100 providers at page size 100.

Unit tests (`components/shadcn/table/data-table-pagination.test.tsx`) cover: selector kept when a large page size collapses the table, navigation hidden in that state, the smallest-option threshold, nothing rendered when no option would paginate, and the URL-follows-through case.

Collateral checked across existing suites: `resources-table-with-selection` (count 17) and `findings-group-table` (count 12) now render the selector where they previously rendered no footer, but both query by accessible name/testid so their assertions are unaffected. Every other fixture sits below the threshold. Both controlled-mode consumers (`scan-list-table.tsx`, `resource-detail-content.tsx`) pass `controlledPageSize` alongside `onPageSizeChange`, so deriving the value is correct for them.

### Checklist

<details>

</details>

Internally reported bug, no tracking issue.

- [ ] Review if backport is needed. Worth considering: user-facing fix with no API surface change.
- [x] Review if is needed to change the [README.md](https://github.com/prowler-cloud/prowler/blob/master/README.md) — not needed.
- [x] Ensure a changelog fragment is added under [prowler/changelog.d/](https://github.com/prowler-cloud/prowler/tree/master/prowler/changelog.d), if applicable. — N/A, UI-only.

#### SDK/CLI

#### UI
- [ ] All issue/task requirements work as expected on the UI — verified by unit tests; **not yet verified in a running browser**.
- [x] If this PR adds or updates npm dependencies, include package-health evidence (maintenance, popularity, known vulnerabilities, license, release age) and explain why existing/native alternatives are insufficient. — no dependency changes.
- [x] Ensure a changelog fragment is added under [ui/changelog.d/](https://github.com/prowler-cloud/prowler/tree/master/ui/changelog.d), if applicable. — `ui/changelog.d/table-page-size-selector.fixed.md`

#### API
- [x] All issue/task requirements work as expected on the API — N/A, no API changes.
- [x] Endpoint response output (if applicable) — N/A
- [x] EXPLAIN ANALYZE output for new/modified queries or indexes (if applicable) — N/A
- [x] Performance test results (if applicable) — N/A
- [x] Any other relevant evidence of the implementation (if applicable) — N/A
- [x] Verify if API specs need to be regenerated. — not needed.
- [x] Check if version updates are required (e.g., specs, uv, etc.). — not needed.
- [x] Ensure a changelog fragment is added under [api/changelog.d/](https://github.com/prowler-cloud/prowler/tree/master/api/changelog.d), if applicable. — N/A

#### MCP Server
- [x] All issue/task requirements work as expected on the MCP Server — N/A, no MCP changes.
- [x] Ensure a changelog fragment is added under [mcp_server/changelog.d/](https://github.com/prowler-cloud/prowler/tree/master/mcp_server/changelog.d), if applicable. — N/A

### License

## Summary by CodeRabbit

- **Bug Fixes**
  - The rows-per-page selector remains available when a selected page size reduces the table to a single page.
  - Page-size selection and page navigation now appear independently, allowing users to change page size even when navigation is unnecessary.
  - Pagination controls remain hidden when the table requires neither option.
  - Page-size changes reflected in the URL now update the displayed selection correctly while preserving existing pagination behavior.
  - Pagination state and controls now remain consistent after changing page-size configurations.

### DIFF
--- a/ui/changelog.d/table-page-size-selector.fixed.md
+++ b/ui/changelog.d/table-page-size-selector.fixed.md
@@ -1,0 +1,1 @@
+Rows-per-page selector no longer disappears when the chosen page size collapses a table to a single page
--- a/ui/components/shadcn/table/data-table-pagination.test.tsx
+++ b/ui/components/shadcn/table/data-table-pagination.test.tsx
@@ -1,12 +1,14 @@
 import { render, screen } from "@testing-library/react";
-import { describe, expect, it, vi } from "vitest";
+import { beforeEach, describe, expect, it, vi } from "vitest";
 
 import type { MetaDataProps } from "@/types";
+
+const mocks = vi.hoisted(() => ({ searchParams: new URLSearchParams() }));
 
 vi.mock("next/navigation", () => ({
   usePathname: () => "/providers",
   useRouter: () => ({ push: vi.fn() }),
-  useSearchParams: () => new URLSearchParams(),
+  useSearchParams: () => mocks.searchParams,
 }));
 
 vi.mock("@/lib", () => ({
@@ -19,8 +21,17 @@ vi.mock("@/lib", () => ({
 }));
 
 vi.mock("@/components/shadcn/select/select", () => ({
-  Select: ({ children }: { children: React.ReactNode }) => (
-    <div>{children}</div>
+  // data-value stands in for the trigger label
+  Select: ({
+    value,
+    children,
+  }: {
+    value: string;
+    children: React.ReactNode;
+  }) => (
+    <div data-testid="page-size-select" data-value={value}>
+      {children}
+    </div>
   ),
   SelectContent: ({ children }: { children: React.ReactNode }) => (
     <div>{children}</div>
@@ -35,7 +46,7 @@ vi.mock("@/components/shadcn/select/select", () => ({
   SelectTrigger: ({ children }: { children: React.ReactNode }) => (
     <button type="button">{children}</button>
   ),
-  SelectValue: () => <span>10</span>,
+  SelectValue: () => null,
 }));
 
 import { DataTablePagination } from "./data-table-pagination";
@@ -51,6 +62,10 @@ const metadata: MetaDataProps = {
 };
 
 describe("DataTablePagination", () => {
+  beforeEach(() => {
+    mocks.searchParams = new URLSearchParams();
+  });
+
   it("keeps navigation arrows visible on hover in light theme", () => {
     render(<DataTablePagination metadata={metadata} />);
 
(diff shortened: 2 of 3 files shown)


alejandrobailo (approved):
LGTM

PR #12343, opened 2026-08-05
test(mcp): cover the integrations tools and models

puchy22:
### Context

The `prowler_app` integrations surface — `prowler_mcp_server/prowler_app/tools/integrations.py` and its models — shipped without tests. It is one of the riskier corners of the MCP server to leave unpinned:

- Every write tool hand-builds a JSON:API document, and the API silently ignores an attribute it does not recognise. A misspelled key is therefore invisible: the request succeeds and the value is quietly dropped.
- `prowler_create_integration` / `prowler_update_integration` run a connection check as part of the call. Whether that check runs, and whether the integration is re-read afterwards, is choreography that is easy to break by accident and hard to notice without tests.
- `prowler_dispatch_findings_to_jira` has to tell the caller whether a failed dispatch is safe to retry. Getting that wrong either duplicates Jira issues or hides ones that were already created.

This PR pins all three, and folds in the source changes the tests drove out.

### Description

Adds 63 tests (42 for the tools, 21 for the models) driven through the existing mock HTTP router, so the assertions are made against the bytes that would really have gone out.

What the tests cover:

- **Connection-check choreography** on create and update — that a plain `enabled` flip does not trigger a re-check, that a credentials or configuration change does, and that reordering the same providers does not while attaching a different one does.
- **Provider-ID validation per integration type** — Jira integrations are tenant-wide and reject provider attachment; AWS Security Hub must stay attached to exactly one provider.
- **Jira dispatch retry safety** — the four distinct outcomes (never started, accepted with no task ID, task died halfway, still running when the wait ended) and which single one is actually safe to retry.
- **Request-shape assertions** — that credentials are sent as given without invented keys, that a Jira site URL is reduced to its domain name, and that findings are dispatched as a filter rather than a body field.

Source changes the tests drove out:

- `prowler_list_integrations` stops requesting the `configuration` field it discards, now that the API tolerates a sparse fieldset without it.
- `JiraDispatchResult.from_task_result` owns the "result is not an object" guard instead of duplicating it at the call site.
- `prowler_update_integration` reads the current integration through `DetailedIntegration` rather than poking at the raw JSON:API payload, and re-fetches once after the PATCH instead of once per branch.
- The JSON:API providers linkage is built by a single `_providers_relationship` helper shared by create and update, replacing two copies of the same literal.
- `MockRouter.json_body()` is added to the test helper so tests can assert on the document actually sent.

No public tool signature or response schema changes.

### Steps to review

```bash
cd mcp_server
uv run pytest tests -q
uv run pytest tests --cov=./prowler_mcp_server --cov-report=term | grep integrations
```

Expected:

```
111 passed

prowler_mcp_server/prowler_app/models/integrations.py    105    0   100%
prowler_mcp_server/prowler_app/tools/integrations.py     209    3    99%
```

Suggested reading order:

1. `mcp_server/tests/helpers/http.py` — the one new helper (`json_body()`) the tool tests lean on.
2. `mcp_server/prowler_mcp_server/prowler_app/tools/integrations.py` — the source diff. The change worth the most attention is the `prowler_update_integration` restructure: the connection check is now computed into `recheck_connection` and the re-fetch happens once on both paths. Worth confirming the response shape is unchanged in each branch (`IntegrationConnectionStatus` when a check ran, `DetailedIntegration` otherwise).
3. `mcp_server/prowler_mcp_server/prowler_app/models/integrations.py` — the guard moved into `from_task_result`, with the parameter widened to `Any` so the model owns the type check.
4. The two test files — test names are written as sentences, so the list of names doubles as a spec of the current behaviour.

Point worth a second opinion: dropping `configuration` from `INTEGRATION_LIST_FIELDS`. The field was already discarded by the list serializer, so this is a request-size change rather than a response change, but it does depend on the API accepting the sparse fieldset without it.

### Checklist

<details>

</details>

#### SDK/CLI

#### UI

#### API

#### MCP Server
- [x] All issue/task requirements work as expected on the MCP Server
- [x] Ensure a changelog fragment is added under [mcp_server/changelog.d/](https://github.com/prowler-cloud/prowler/tree/master/mcp_server/changelog.d), if applicable.

### License

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved integration updates with safer configuration handling, provider validation, and no-op behavior.
  * Added connection re-checks when credentials, configuration, or providers change.
  * Clarified Jira dispatch failure states and safe retry guidance.
  * Streamlined integration listings by omitting unnecessary configuration details.
  * Improved API error reporting while preserving transport timeout details.

* **Tests**
  * Expanded coverage for integration workflows, pagination, validation, connection checks, and Jira dispatch outcomes.

### DIFF
--- a/mcp_server/tests/prowler_app/tools/test_integrations.py
+++ b/mcp_server/tests/prowler_app/tools/test_integrations.py
@@ -18,6 +18,7 @@ As in ``test_findings``, tools are driven through an in-memory MCP client so
 FastMCP resolves the pydantic ``Field`` defaults.
 """
 
+import httpx
 import pytest
 from fastmcp import Client
 
@@ -77,7 +78,9 @@ def stub_integration(
     """Serve ``GET /integrations/i1`` for every read a tool makes.
 
     A single registration is enough because the router repeats its last response,
-    and the tools read the integration both before and after a write.
+    and the tools read the integration both before and after a write. Call it
+    twice to serve a different state to each read, which is what tells the state
+    returned after a write apart from the one read before it.
     """
     relationships = (
         {"providers": jsonapi_relationship_many("providers", *provider_ids)}
@@ -435,9 +438,13 @@ async def test_updating_only_the_enabled_flag_does_not_recheck_the_connection(
     """Nothing about reachability changed, so the check would be pure latency.
 
     It is also destructive to spend: the check is a background task the tool
-    waits on for up to two minutes.
+    waits on for up to two minutes. Skipping the check must not also skip the
+    read-back, though: the PATCH response body is empty, so returning the state
+    read before the write would report the integration as still enabled.
     """
+    # The read before the PATCH, then the read after it
     stub_integration(mock_router, S3_ATTRIBUTES)
+    stub_integration(mock_router, {**S3_ATTRIBUTES, "enabled": False})
     mock_router.add("PATCH", INTEGRATION, json=jsonapi_document({}))
 
     async with Client(mcp_root_server) as client:
@@ -449,7 +456,7 @@ async def test_updating_only_the_enabled_flag_does_not_recheck_the_connection(
         "enabled": False
     }
     assert f"POST {CONNECTION}" not in mock_router.paths()
-    assert result.data["id"] == "i1"
+    assert result.data["enabled"] is False
 
 
 async def test_updating_the_configuration_merges_it_onto_the_current_one(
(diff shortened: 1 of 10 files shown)


mcp_server/tests/prowler_app/tools/test_integrations.py
```diff
@@ -0,0 +1,1040 @@
...
+            "prowler_send_findings_to_jira",
+            {
+                "integration_id": "i1",
+                "project_key": "PROJ",
+                "issue_type": "Epic",
+                "finding_ids": ["f1"],
+            },
+        )
+
+    assert result.data["status"] == "failed"
```
josema-xyz, line 938: This should be the only safe retry, but the response does not return `safe_to_retry=true`. Can we return it when the server clearly rejected the request, while keeping timeouts unsafe because the request may have gone through?

mcp_server/tests/prowler_app/tools/test_integrations.py
```diff
@@ -0,0 +1,1040 @@
...
+    async with Client(mcp_root_server) as client:
+        result = await client.call_tool(
+            "prowler_update_integration", {"integration_id": "i1", "enabled": False}
+        )
+
+    assert mock_router.json_body("PATCH", INTEGRATION)["data"]["attributes"] == {
+        "enabled": False
+    }
+    assert f"POST {CONNECTION}" not in mock_router.paths()
+    assert result.data["id"] == "i1"
```
josema-xyz: The mock returns the same data before and after the `PATCH`, so this does not prove that the second `GET` happens. Could we return different states and check the updated value?

josema-xyz (approved):

danibarranqueroo (approved):

PR #12350, opened 2026-08-05
fix(ui): launch organization scans through the bulk endpoint

josema-xyz:
### Context

During organization onboarding, the UI launches one manual scan request per provider. On trial accounts, the first request creates an active scan and the subscription gate rejects the remaining requests, so only one provider is scanned.

This change makes organization onboarding use the bulk scan API so the subscription gate evaluates the organization scan once.

### Description

- Launch AWS and GCP organization scans with one `POST /api/v1/scans/bulk` request.
- Send the organization ID using the JSON:API organization relationship.
- Use the returned scan collection to report the launched scan count.
- Preserve the existing per-provider daily schedule flow and concurrency limit.
- Preserve schedule-only onboarding and standalone provider scans.
- Update unit and integration tests for the bulk request and organization onboarding.

This requires Cloud API support for `POST /api/v1/scans/bulk`.

### Steps to review

1. Onboard an AWS or GCP organization and select an initial scan.
2. Verify the UI sends one `POST /api/v1/scans/bulk` request with the organization ID.
3. Verify one scan is created for every provider linked to the organization.
4. Verify daily scheduling still creates schedules for each provider.
5. Verify schedule-only onboarding does not launch a bulk scan.
6. Verify bulk scan failures keep the manual flow open and display the API error.

### Checklist

#### UI

### License

## Summary by CodeRabbit

* **New Features**
  * Organization onboarding now launches scans for all linked AWS and GCP providers through one bulk operation.
  * Scan launching and daily scheduling are handled as separate actions.
  * Improved feedback reports launch errors and the number of scans created.

* **Bug Fixes**
  * Partial or incomplete schedule saves now correctly trigger organization-wide scanning.
  * Fully failed schedule saves no longer start scans.


ui/components/providers/organizations/org-launch-scan.tsx
```diff
@@ -198,14 +202,21 @@ export function OrgLaunchScan({
...
+        : { error: "Organization ID is required" };
+
+      if (hasActionError(scanResult)) {
+        initialScanFailureCount = updatedProviderIds.length;
+        initialScanError = getActionErrorMessage(scanResult);
+      } else {
+        initialScanSuccessCount = Array.isArray(scanResult?.data)
+          ? scanResult.data.length
+          : 0;
+      }
```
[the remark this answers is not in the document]
josema-xyz: `POST /api/v1/scans/bulk` guarantees a JSON:API scan collection for every successful `202` response. The UI intentionally uses that collection as the authoritative launched-scan count. An empty body would violate the API contract, and falling back to the provider count could report scans that were not returned by the API. No change is needed here.

ui/components/providers/organizations/org-launch-scan.tsx
```diff
@@ -245,11 +258,28 @@ export function OrgLaunchScan({
...
+          title: "Unable to launch organization scan",
+          description: getActionErrorMessage(result),
+        });
+        return;
+      }
+
+      successCount = Array.isArray(result?.data) ? result.data.length : 0;
+    } else {
+      const result = await scheduleOrganizationDailyScans(createdProviderIds);
+      successCount = result.successCount;
```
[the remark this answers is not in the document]
josema-xyz: This behavior predates this PR. The daily scheduling loop was only moved to a dedicated helper so organization scan launches can use the bulk endpoint without changing scheduling behavior. Handling total or partial scheduling failures changes the existing UX and is outside the scope of this PR, so I’m leaving it unchanged here.

ui/components/providers/organizations/org-launch-scan.tsx
```diff
@@ -198,14 +202,21 @@ export function OrgLaunchScan({
...
+        ? await launchOrganizationScans(organizationId)
+        : { error: "Organization ID is required" };
+
+      if (hasActionError(scanResult)) {
+        initialScanFailureCount = updatedProviderIds.length;
+        initialScanError = getActionErrorMessage(scanResult);
+      } else {
+        initialScanSuccessCount = Array.isArray(scanResult?.data)
+          ? scanResult.data.length
+          : 0;
```
pfe-nazaries: Any non-array `data` reads as zero scans, so a successful launch can report zero: `initialScanFailureCount` stays 0 and the title still says "initial scans launched", while `targetTab` flips to **Scheduled**. Same read at `:278` gives "Single scan launched for 0 accounts." on the trial path.

Likeliest trigger: this posts a singular `type: "scans-bulk"`, and OSS `POST /scans` answers 202 with one `TaskSerializer` resource (`api/src/backend/api/v1/views.py:2277`) — `scanOnDemand` already reads that family's 202 as a single object (`scans.ts:145`). An empty 202 does it too (`handleApiResponse` returns `{success: true}`).

A bulk launch is one operation, so deriving a per-account count from the body is what creates this. Dropping the count and always linking to Active is correct under every candidate shape.

ui/actions/scans/scans.ts
```diff
@@ -182,10 +182,45 @@ export const scheduleDaily = async (formData: FormData) => {
   }
 };
 
-export const launchOrganizationScans = async (
-  providerIds: string[],
-  scheduleOption: "daily" | "single",
-) => {
+export const launchOrganizationScans = async (organizationId: string) => {
```
pfe-nazaries: No return type here, so this is effectively `any` — which is why the `Array.isArray(result?.data)` read at `org-launch-scan.tsx:216` compiles silently and `tsc` cannot flag it.

The sibling bulk action already solved this: `updateSchedulesBulk` declares `Promise<SchedulesBulkResponse>` (`actions/schedules/schedules.ts:157`) and pairs it with `parseSchedulesBulkResult` (`lib/schedules.ts:167`), which tolerates the alternate shape "so a future server-side JSON:API normalization cannot read as an empty result" — this exact failure — and returns `isIndeterminate`, keeping "no scans" distinct from "cannot tell". Here both collapse to `0`.

With an `any` return, an overlay call site still reading `result.successCount` also compiles clean and yields `undefined` at runtime; a declared `ScansBulkResponse` makes that a build error.

ui/components/providers/organizations/org-launch-scan.tsx
```diff
@@ -198,14 +202,21 @@ export function OrgLaunchScan({
...
+      }
     }
 
     setIsLaunching(false);
```
pfe-nazaries: This branch records the failure but does not stop the flow: `setIsLaunching(false)` and `finishSuccess()` (`:223`) run unconditionally, so the wizard closes and navigates to `/providers`, with the error only appended to the non-destructive "Scan schedules saved" toast.

On a 402 — the trial tenant this PR is about — no scan started, and the user leaves the wizard (and the `UsageLimitMessage` billing link at `:392`) believing it succeeded.

The legacy path already gets this right at `:268-276`: destructive toast, early return, wizard stays open.

ui/components/providers/organizations/org-launch-scan.tsx
```diff
@@ -198,14 +202,21 @@ export function OrgLaunchScan({
...
 
     if (values.launchInitialScan) {
-      const scanResult = await launchOrganizationScans(
-        updatedProviderIds,
-        SCAN_SCHEDULE.SINGLE,
-      );
-      initialScanFailureCount = scanResult.failureCount;
-      initialScanSuccessCount = scanResult.successCount;
+      const scanResult = organizationId
+        ? await launchOrganizationScans(organizationId)
```
pfe-nazaries, line 209: The request carries only the organization id, so `updatedProviderIds` gates *whether* to scan but no longer scopes *what*. Now pulled in: providers whose `/schedules/bulk` save just failed, providers dropped from `createdProviderIds` for a failed connection test, and providers attached to this org by an earlier onboarding run.

If that is intended — an organization scan covers the organization — it is worth stating in the PR body and the changelog, because the test that pinned the old behaviour was renamed to accept it rather than updated, so nothing pins the new blast radius.

ui/__tests__/msw/handlers/organizations.ts
```diff
@@ -624,6 +624,25 @@ export const handlersForOrganizations = (
...
+              task: {
+                data: { id: `scan-task-${index + 1}`, type: "tasks" },
+              },
+            },
+          })),
+        },
+        { status: 202 },
+      ),
+    ),
     http.post(`${API}/scans`, () =>
```
pfe-nazaries: Unreachable now: no organization flow posts `/scans`, and the harness getter that watched it was renamed away (`scanLaunchCount` to `organizationBulkScanCallCount`). `grep -rn "scanLaunchCount"` returns nothing.

Keeping it hides a regression — if the fan-out comes back, each per-provider `POST /scans` gets a plausible 202 from here instead of tripping MSW's unhandled-request path.

ui/app/(prowler)/providers/providers-page.gcp.integration.test.tsx
```diff
@@ -271,7 +271,7 @@ describe("GCP Organizations onboarding (Phase 2)", () => {
     await harness.saveScheduleAndLaunch();
     await harness.waitForLaunchComplete();
 
-    expect(harness.scanLaunchCount).toBe(GCP_CREATED_PROVIDER_IDS.length);
+    expect(harness.organizationBulkScanCallCount).toBe(1);
```
pfe-nazaries, line 274: This used to assert `scanLaunchCount === GCP_CREATED_PROVIDER_IDS.length` — a real per-provider count. `=== 1` only proves one request was attempted, so "every project actually got scanned" is no longer pinned anywhere.

It cannot substitute either: `countRequests` logs on `request:start` (`__tests__/browser-harness.ts:84`), counting attempts regardless of the response, and `waitForLaunchComplete()` waits for a title that is emitted even at zero parsed scans (see the `Array.isArray` comment).

Asserting on the response — how many scans the fixture returned, or that `POST /scans` was never called — would close the gap.

ui/actions/scans/scans.ts
```diff
@@ -182,10 +182,45 @@ export const scheduleDaily = async (formData: FormData) => {
...
+    return { error: "Organization ID is required" };
+  }
+
+  const headers = await getAuthHeaders({ contentType: true });
+  const url = new URL(`${apiBaseUrl}/scans/bulk`);
+
+  addScanOperation("create", undefined, {
+    organization_id: organizationId,
+    bulk: true,
+  });
```
pfe-nazaries: The path this replaces went through `scanOnDemand`, which emitted `addScanOperation("start", result.data.id)` on success (`:145-148`). Only `create` is emitted here, so an organization launch leaves no `start` breadcrumb and no scan id in Sentry — the hardest case to debug, since one request stands in for N scans.

It also fires before the request is sent, so it cannot distinguish "dispatched" from "never left the client".

ui/app/(prowler)/providers/providers-page.gcp.integration.test.tsx
```diff
@@ -206,7 +206,7 @@ describe("GCP Organizations onboarding (Phase 2)", () => {
 
   // These three cases pin how `/schedules/bulk`'s per-provider lists are read: a
   // client looking one level too deep sees no lists and cannot tell them apart.
-  it("launches initial scans only for the projects whose schedule was saved", async () => {
+  it("launches the organization after a partial schedule save", async () => {
```
pfe-nazaries, line 209: **question:** the previous case is now unreachable? Or should both cases be tested (keep both)

(This is the case that used to be pinned as "launches initial scans only for the projects whose schedule was saved".)

pfe-nazaries (changes requested):
The direction is right — collapsing the per-provider fan-out into one bulk call removes the per-request 402. Two things I would like resolved first:

- `Array.isArray(result?.data)` makes a successful bulk launch indistinguishable from a no-op: any non-array `data` reports zero scans, so the toast can claim "initial scans launched" while linking to Scheduled, and the trial path reads "launched for 0 accounts".
- The action has no declared return type, which is why that read compiles silently. `/schedules/bulk` already solved this pattern (typed response plus a tolerant reader that keeps "no scans" distinct from "cannot tell") and it is worth reusing here.

Worth confirming the real 202 shape of `POST /scans/bulk` against the API too, since nothing in this repo pins it. The rest of the comments are smaller: scope, test coverage and a stale MSW handler.

ui/app/(prowler)/providers/providers-page.gcp.integration.test.tsx
josema-xyz: Covered both UI boundaries. The component test now verifies that a two-scan response produces the two-account count, and the organization MSW setup no longer handles `POST /scans`, so per-provider fan-out is a hard failure. Creating one scan per linked provider remains covered by the Cloud API tests.

ui/__tests__/msw/handlers/organizations.ts
josema-xyz: Removed it. The organization MSW setup now has no `POST /scans` handler, so any restored per-provider fan-out fails through `onUnhandledRequest: "error"`.

ui/components/providers/organizations/org-launch-scan.tsx
josema-xyz: No change here. This is intentional partial success: schedules were saved, so onboarding completes and the toast reports that the initial organization scan failed without saying it launched. Trial manual-scan failures use the legacy path, which keeps the wizard open with a destructive toast.

ui/components/providers/organizations/org-launch-scan.tsx
josema-xyz: Fixed the ambiguous success path. The action now requires a valid JSON:API scan collection before returning success, and the component reads `data.length` from that typed result. A non-array or empty-body 202 returns an error instead of reporting zero scans. The Cloud endpoint contract remains a scan collection.

ui/actions/scans/scans.ts
josema-xyz: Changed it to emit `start` only after a valid scan collection is returned. The breadcrumb includes the organization ID, bulk flag, scan count, and returned scan IDs.

pfe-nazaries (approved):

PR #12411, opened 2026-08-10
fix(ui): open scan findings using the scan's UTC day

pfe-nazaries:
### Context

A customer reported that **View Findings** on the Scans page opened an empty list for their scheduled scans, while the same action on their manual scans worked. Filtering the Findings page by a specific date showed the same emptiness. Nothing was failing — the scans ran fine and the findings existed.

Root cause: the link was built with a date derived in the **browser's** timezone, but the API interprets `filter[inserted_at]` as a whole **UTC** day and materialises it as a UUIDv7 `id` range (`TIME_ZONE = "UTC"`). Since `filter[scan__in]` also constrains `id`, the two ranges intersect and the result is empty rather than unfiltered.

For a viewer at UTC+10 only local hours 10:00–23:59 share the UTC date, so a scan finishing in their early morning lands on the previous UTC day. Manual scans run during business hours and matched; the nightly scheduled scan did not. Exposure is `|offset| / 24` of the day and it breaks in both directions.

Measured on the dev environment — same scan, same filters, only the date differing:

| Scan finished | UTC date | Sydney date | Sent UTC date | Sent local date |
|---|---|---|---|---|
| `2026-08-03 15:24:41Z` | Aug 3 | Aug 4 | **596** findings · 130 groups | **0** · 0 |
| `2026-07-09 14:28:05Z` | Jul 9 | Jul 10 | **92** findings · 63 groups | **0** · 0 |

This PR fixes the Scans → Findings entry point. The date picker on the Findings page is affected by the same mismatch and is **not** covered here — it needs an API change to accept timestamp precision on the grouped and metadata endpoints, and will follow separately.

### Description

**Stop deriving the date in the browser.** `openFindings` no longer puts `filter[inserted_at]` in the URL at all. `resolveFindingScanDateFilters` already derives the range server-side from the scan's own timestamps in UTC (it splits the ISO string on `"T"`), but it short-circuits when the URL already carries an `inserted_at` — so the correct code path was being overridden by the incorrect one. Removing the local date lets it run. The menu item's guard now checks the raw `completed_at` instead of a derived local date, because without a completion timestamp there is nothing to derive and the API would answer `400 "At least one date filter is required"`.

**Cover every UTC day a scan spans.** The range is now derived from `started_at` **and** `completed_at`, not `completed_at` alone. Findings are `bulk_create`d progressively while a scan runs (`api/src/backend/tasks/jobs/scan.py`) and their UUIDv7 `id` is stamped on insert, so a scan straddling UTC midnight leaves findings on both days and the previous single-day range silently dropped the earlier ones. `buildFindingScanDateFilters` already emitted `gte`/`lte` for multiple dates, so this needed no shape change. A scan with only `completed_at` behaves exactly as before.

No API changes. No new dependencies.

### Steps to review

1. On the Scans page, open the actions menu of a completed scan and choose **View Findings**. The URL should contain only `filter[scan__in]` and `filter[status__in]=FAIL` — no `filter[inserted_at]`.
2. The findings list should be populated regardless of your machine's timezone. To exercise the bug directly, set your OS timezone to `Australia/Sydney` and use a scan that finished between 14:00Z and 15:59Z: before this change the list was empty, now it is not.
3. Confirm the menu item stays disabled for a completed scan with no `completed_at`.
4. Run the unit tests:

```bash
cd ui && pnpm vitest run --project unit components/scans/table/scan-jobs-row-actions.test.tsx lib/findings-scan-filters.test.ts
```

The tests pin: the URL carries no browser-local date; the item is disabled without `completed_at`; a same-day scan yields an exact `inserted_at`; a scan crossing midnight yields a `gte`/`lte` range; `started_at`-only and no-timestamp scans degrade gracefully; multiple selected scans span the full range.

### Checklist

<details>

</details>

- [x] Review if is needed to change the Readme.md — not needed.
- [x] Ensure a changelog fragment is added under `<component>/changelog.d/`, if applicable.

#### SDK/CLI
- Are there new checks included in this PR? **No** — no SDK changes.

#### UI (if applicable)
- [x] Ensure a changelog fragment is added under `ui/changelog.d/`

> No screenshots: there is no visual change. The fix alters the query string a menu item navigates to, so the difference is a populated list where it was previously empty — the reproduction steps above show it better than a screenshot would.

#### API (if applicable)

Not applicable — no API changes in this PR.

### License

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved **View Findings** navigation across browser timezones and scans spanning two UTC days.
  * Findings now reflect scans throughout their execution period, including incomplete scans.
  * Disabled **View Findings** when a scan has not completed or lacks a completion timestamp.
* **Tests**
  * Added coverage for timezone boundaries, incomplete scans, missing timestamps, and multiple selected scans.
* **Documentation**
  * Added a changelog entry describing the corrected empty findings list behavior.

### DIFF
--- a/ui/changelog.d/scan-findings-utc-day.fixed.md
+++ b/ui/changelog.d/scan-findings-utc-day.fixed.md
@@ -1,1 +1,1 @@
-`View Findings` on the Scans page no longer opens an empty list when the scan finished on a different day in the browser's timezone, and the results now include scans whose findings span two UTC days
+`View Findings` on the Scans page no longer opens an empty list for users outside the UTC timezone
--- a/ui/changelog.d/scan-findings-utc-day.fixed.md
+++ b/ui/changelog.d/scan-findings-utc-day.fixed.md
@@ -1,0 +1,1 @@
+`View Findings` on the Scans page no longer opens an empty list when the scan finished on a different day in the browser's timezone, and the results now include scans whose findings span two UTC days
--- a/ui/components/scans/table/scan-jobs-row-actions.test.tsx
+++ b/ui/components/scans/table/scan-jobs-row-actions.test.tsx
@@ -329,14 +329,14 @@ describe("ScanJobsRowActions", () => {
     );
     await user.click(screen.getByRole("menuitem", { name: /view findings/i }));
 
-    // Then - the date is derived server-side in UTC, never in the URL.
+    // Then
     expect(pushMock).toHaveBeenCalledWith(
       "/findings?filter[scan__in]=scan-1&filter[status__in]=FAIL",
     );
   });
 
   it("disables View Findings when the completed scan has no completion timestamp", async () => {
-    // Given - without `completed_at` there is no date the server can derive.
+    // Given
     const user = userEvent.setup();
     render(
       <ScanJobsRowActions
--- a/ui/components/scans/table/scan-jobs-row-actions.test.tsx
+++ b/ui/components/scans/table/scan-jobs-row-actions.test.tsx
@@ -44,11 +44,6 @@ vi.mock("@/lib/helper", () => ({
   downloadScanZip: downloadScanZipMock,
 }));
 
-vi.mock("@/lib/date-utils", () => ({
-  toLocalDateString: (value: string | null | undefined) =>
-    value ? "2026-01-01" : undefined,
-}));
-
 vi.mock("@/components/scans/edit-alias-modal", () => ({
   EditAliasModal: ({
     open,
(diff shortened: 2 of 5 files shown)


pfe-nazaries:
## Fixes Applied Successfully

Fixed 1 file based on 1 CodeRabbit feedback item.

**Files modified:**
- `ui/components/scans/table/scan-jobs-row-actions.tsx`

The `hasCompletedAt` comment claimed the findings page had no date to derive without `completed_at`. `resolveFindingScanDateFilters` actually falls back to `started_at`, so the comment now states the real reason: `completed_at` bounds the UTC day range, and without it the range collapses to the start day.

Comment-only change; no behavior or logic was altered.

**Commit:** `6d284a501`

The latest autofix changes are on the `fix/findings-filters` branch.

alejandrobailo (approved):
LGTM!

PR #12435, opened 2026-08-13
feat(ui): add Slack integration connect flow

pfe-nazaries:
### Context

Tenants have no way to get Prowler output into Slack from the product. The CLI has had a Slack output since v3, and the API's `Integration` model has carried `IntegrationChoices.SLACK` since the integrations migration landed, but nothing implements it: the connection test falls through a bare `pass` and the scan-time dispatch is commented out.

This is **PR 1 of 4** in a stack:

1. **this PR** — connect a workspace
2. choose a channel and verify it
3. disconnect and revoked-credential recovery
4. documentation

**Scope: UI only.** The Slack API is implemented in a separate private repository by the backend lane, against a shared contract mirrored on the Notion page "Alerts: Slack". No Slack code is added to `api/`, and `prowler/` is untouched — both repositories pin the SDK to a git-master commit, so an SDK change on a feature branch would be invisible to the API that calls it.

Planning artifacts live in the `add-slack-integration` OpenSpec change.

### Description

- Slack card on `/integrations` and a `/integrations/slack` management page, both gated on `isCloud()` — the API exists only in the hosted deployment, so elsewhere there is nothing to reach.
- Connection happens through Slack's OAuth v2 app-install flow ("Add to Slack"). The tenant never obtains or pastes a token: the API owns the client secret, mints a short-lived single-use `state` bound to tenant and user, and performs the exchange.
- `/integrations/slack/callback` completes the install **exactly once** behind a guard — the Slack code is single-use, so a double invocation would burn it and show a spurious failure — then routes to the management page with the outcome.
- A deployment with no Slack app configured (`503`) is presented as "Slack is not available in this environment yet" rather than as an error, so this ships dark until the backend lane deploys.
- MSW handlers and fixtures derived from the shared API contract; they are the UI lane's only view of the API.

Requires no API change to land, and no migration: `slack` is already a value of the `integration_type` enum.

### Steps to review

1. `pnpm --filter ui test` — or target the two files directly. Expect 10 passing browser-mode integration tests: 2 in `integrations-page.integration.test.tsx`, 8 in `slack-page.integration.test.tsx`.
2. The flows worth reading as user journeys are in `slack-page.integration.test.tsx`: consent screen carries exactly the four scopes; install completes and names the workspace; user declines; Slack refuses the install; the API refuses the `state`.
3. Check the once-guard is intact — `harness.exchangeCallCount` is asserted at four call sites and is load-bearing, not incidental.
4. There is no reachable backend yet, so the flow cannot be exercised by hand in a deployed environment. Verification against the development backend is a tracked checkpoint, not part of this PR.

### Checklist

<details>

- [x] This feature/issue is listed in the roadmap
- [x] Is it assigned to me

</details>

- [x] Review if code is being documented — the docs page ships in PR 4 of this stack.
- [ ] Review if backport is needed. **No** — new feature.
- [ ] Review if is needed to change the Readme.md **No.**
- [x] Ensure a changelog fragment is added under `<component>/changelog.d/`.

#### UI

- [x] All issue/task requirements work as expected on the UI — against MSW handlers. **Not yet verified against a real backend**: the API is not deployed, so this is tracked as a separate checkpoint.
- [ ] Screenshots/Video - Mobile (X < 640px) — not captured.
- [ ] Screenshots/Video - Tablet (640px > X < 1024px) — not captured.
- [x] Screenshots/Video - Desktop (X > 1024px) — 1920x1080 captures of these screens ship in PR 4 of this stack, alongside the docs page they illustrate.
- [x] Ensure a changelog fragment is added under `ui/changelog.d/`

#### API

Not applicable — no API change in this PR, by design.

### License

## Summary by CodeRabbit

* **New Features**
  * Added Slack integration support in Prowler Cloud.
  * Users can install Slack, connect a workspace, configure a channel, and test connection health.
  * Added Slack integration cards, branding, management screens, and OAuth callback handling.
  * Added clear connection statuses, rate-limit notices, retry guidance, and error messages.

* **Bug Fixes**
  * Improved handling of failed, incomplete, declined, malformed, and unavailable Slack authorization responses.
  * Prevented untrusted callback error content from being displayed.

* **Tests**
  * Added comprehensive coverage for Slack installation, connection, callback, error, and availability scenarios.

### DIFF
--- a/ui/actions/integrations/slack.ts
+++ b/ui/actions/integrations/slack.ts
@@ -12,55 +12,22 @@ import {
 import { handleApiError, handleApiResponse } from "@/lib/server-actions-helper";
 import type { IntegrationProps } from "@/types/integrations";
 
-/**
- * The two OAuth calls the Slack install needs. Everything else — reading the
- * integration, testing the connection, deleting it — goes through the generic
- * integration actions, because Slack rows are plain `integrations` resources.
- *
- * The Slack app (`SLACK_CLIENT_ID` / `SLACK_CLIENT_SECRET` /
- * `SLACK_REDIRECT_URI`) is a deployment-wide setting, so both endpoints answer
- * `503` until it is configured, and a deployment that serves no Slack API at
- * all answers `404`. Neither is an error the user can act on: both mean "Slack
- * isn't available in this environment yet", which is why they get their own
- * result shape rather than an error string.
- *
- * Rate limiting is a third answer again — Slack is there, the deployment is
- * configured, and the only thing to do is wait — so it gets its own outcome
- * rather than being folded into either of the other two.
- *
- * The OAuth `state` is minted and consumed by the API, bound to the tenant and
- * the user — the UI never inspects it, it only forwards what Slack sent back.
- */
-
 interface SlackUnavailable {
   unavailable: true;
 }
 
-/**
- * Slack is rate limiting Prowler (`429`). Distinct from `SlackUnavailable`: the
- * Slack app exists and works, so the page keeps offering what it offers and
- * only says when to come back.
- */
 interface SlackRateLimited {
   rateLimited: true;
-  /** What `Retry-After` asked for, when the response carried one. */
   retryAfterSeconds: number | null;
-  /** The wait, as copy — so every caller says the same thing about it. */
   message: string;
 }
 
 /**
- * The API accepted the completion and the UI could not read the answer back.
- *
- * Its own outcome rather than an error, for the same reason `unavailable` and
- * `rateLimited` are: `response.ok` was true, so the code was consumed and the
- * integration upserted — what Prowler cannot name is the workspace, not whether
- * one was connected. Callers that report an error would be reporting the one
- * thing this answer rules out.
+ * The API accepted the exchange (`2xx`) and the UI could not read the workspace
+ * back: the install happened, only its result is unknown.
  */
 interface SlackUnconfirmed {
   unconfirmed: true;
-  /** The unread result, as copy — so every caller says the same thing about it. */
   message: string;
 }
 
@@ -95,36 +62,24 @@ export type SlackExchangeResult =
   | SlackActionError;
 
 /**
- * Whether the deployment answered "no Slack app here". Deliberately narrow:
- * widening it to a status that only means "not right now" (a `429`, an upstream
- * `502`) would tell the user Slack is unavailable in their environment when it
- * is Prowler's call that needs retrying.
+ * `503`: no Slack app configured in this deployment. `404`: no Slack API at
+ * all. Both mean "not available here", unlike `429`/`502` which mean "not now".
  */
 const isUnavailableStatus = (status: number): boolean =>
   status === 503 || status === 404;
 
 const RATE_LIMITED_STATUS = 429;
 
-/**
- * Classify a refusal into the outcome the UI acts on.
- *
- * The reason travels in the JSON:API error's `code`, which
- * `slackErrorMessage` maps to copy Prowler owns; `detail` is the API's own
- * human wording and is used only when the code is one this UI has nothing
- * better to say about. `fallback` covers a refusal that carries neither.
- */
 const failureFrom = async (
   response: Response,
   fallback: string,
 ): Promise<SlackUnavailable | SlackRateLimited | SlackActionError> => {
   if (isUnavailableStatus(response.status)) return { unavailable: true };
 
-  // A 5xx that is not the ship-dark 503 is a server fault rather than a Slack
-  // state — and `502` is the status the contract reserves for "Slack upstream
-  // broke" — so it goes through the repo's own 5xx handling, which is where
-  // Sentry hears about one. That call reports *and throws*, so each action's
-  // existing catch is what answers the user. It has to run before
-  // `readSlackFailure`: a response body can only be read once.
+  // A 5xx (including the `502` the contract reserves for "Slack upstream
+  // broke") goes through the repo's 5xx handling, which reports to Sentry and
+  // throws, so the caller's catch answers the user. Must run before
+  // `readSlackFailure`: a body can only be read once.
   if (response.status >= 500) await handleApiResponse(response);
 
   const failure = await readSlackFailure(response);
@@ -140,10 +95,7 @@ const failureFrom = async (
   return { error: slackErrorMessage(failure, fallback) };
 };
 
-/**
- * Mint an OAuth state and get the consent URL to send the user to. Creates no
- * integration: the install only exists once the exchange completes.
- */
+/** Mint an OAuth state and get the consent URL. Creates no integration. */
 export const getSlackAuthorizeUrl =
   async (): Promise<SlackAuthorizeUrlResult> => {
     const headers = await getAuthHeaders({ contentType: true });
@@ -153,18 +105,17 @@ export const getSlackAuthorizeUrl =
       const response = await fetch(url.toString(), { method: "POST", headers });
 
       if (!response.ok) {
-        // Awaited inside the `try`: a returned promise's rejection is not this
-        // `catch`'s to see, and a 5xx now rejects.
+        // Awaited inside the `try`: a returned promise's rejection would skip
+        // this `catch`, and a 5xx rejects.
         return await failureFrom(
           response,
           `Unable to start the Slack install: ${response.statusText}`,
         );
       }
 
-      // The URL travels in JSON:API `meta` — the call creates no resource. A
-      // `2xx` that is not JSON at all (a proxy's HTML page, an empty body) is
-      // read as "no URL here" rather than left to throw a parser message the
-      // user would be shown verbatim.
+      // The URL travels in JSON:API `meta`: the call creates no resource. A
+      // non-JSON `2xx` reads as "no URL" instead of throwing a parser message
+      // the user would be shown verbatim.
       const body = await response.json().catch(() => null);
       const authorizeUrl = body?.meta?.authorize_url;
 
@@ -180,8 +131,8 @@ export const getSlackAuthorizeUrl =
 
 /**
  * Complete the install with what Slack put in the callback URL. The API
- * validates and consumes the `state`, exchanges the single-use `code`, and
- * upserts the tenant's Slack integration.
+ * consumes the `state`, exchanges the single-use `code`, and upserts the
+ * tenant's Slack integration.
  */
 export const exchangeSlackOAuthCode = async ({
   code,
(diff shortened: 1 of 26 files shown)


pfe-nazaries:
## Fixes Applied Successfully

Fixed 5 file(s) based on 6 CodeRabbit feedback item(s) — 5 applied, 1 deferred.

**Files modified:**
- `ui/actions/integrations/integrations.ts`
- `ui/components/icons/services/IconServices.tsx`
- `ui/components/integrations/slack/slack-callback.test.tsx`
- `ui/components/integrations/slack/slack-integration-card.tsx`
- `ui/lib/integrations/slack-errors.ts`

**Applied:**
- Stale Slack connection state after a test — added the Slack path to `testIntegrationConnection`'s revalidation list, which also covers the failed-test path.
- `"use client"` removed from `SlackIntegrationCard`; its only call site is a Server Component.
- `isKnownCode` now uses an own-property check, so prototype keys no longer pass the guard.
- Decorative `SlackIcon` hidden from assistive technology. Scoped to `SlackIcon`; the other icons in `IconServices.tsx` share the same pattern and belong to a separate accessibility pass.
- Callback redirect is now assertable and asserted. The once-guard half of that suggestion was skipped: it is already covered by `expect(harness.exchangeCallCount).toBe(1)` in the callback page integration test.

**Deferred:**
- `SLACK_DOCS_URL` 404 — deliberate placeholder (see the comment on the line). The slug follows the convention used by every sibling integration, and the docs slice publishes the page under it.

**Validation:** `tsc` 0 errors, `eslint` clean, `prettier` clean, Slack unit tests 5/5, Slack integration tests 22/22.

**Commit:** `21a8c2314`

The latest autofix changes are on the `feature/slack-integration` branch.

pfe-nazaries:
## Fixes Applied Successfully

Fixed 5 file(s) based on 7 CodeRabbit feedback item(s) — 4 applied, 3 not applied.

**Files modified:**
- `ui/actions/integrations/slack.ts`
- `ui/app/(prowler)/integrations/slack/slack-callback-page.integration.test.tsx`
- `ui/app/(prowler)/integrations/slack/slack-integration-content.tsx`
- `ui/components/icons/services/IconServices.tsx`
- `ui/types/integrations.ts`

**Applied:**
- `IntegrationType` now derives from an `INTEGRATION_TYPE` const object, and the Slack filter uses `INTEGRATION_TYPE.SLACK` rather than a literal.
- `exchangeSlackOAuthCode` validates its argument with `safeParse` before use, matching the pattern in `schedules.ts` and `compliance-watchlist.ts`. The parameter keeps its declared type so the caller stays type-checked.
- `SlackIcon` honours explicit `width` and `height`. Applied as a two-attribute change: the committable suggestion on that thread still contained the previous `aria-hidden="false"` / `aria-label` lines and would have reverted the earlier accessibility fix.
- The successful-install test now pins the revalidated paths, using harness plumbing that already existed.

**Not applied:**
- *Do not start installation when the integration read failed* — the current behaviour is the documented intent, not an oversight. The file header explains the `{ error }` path exists so the page "could still offer the install", and `slack-page.integration.test.tsx:101` pins it (`offersInstall() === true`, "one read failed, the Slack app is fine"). The already-connected case this would guard against is handled by the API's `slack_workspace_conflict` answer, which has its own copy and its own test.
- *Decorative `SlackIcon`* — the `SlackIcon` fix shipped previously. The remaining request covers the other 29 icons in `IconServices.tsx`, which share the same pattern and belong to a separate accessibility pass.
- *`SLACK_DOCS_URL` 404* — deliberate placeholder, as the comment on the line records. The slug follows the convention every sibling integration uses and the docs slice publishes the page under it.

Also note `width`/`height` and the ARIA attributes are latent across all 30 icons in `IconServices.tsx`, and no current caller passes either dimension to one.

**Validation:** `tsc` 0 errors, `eslint` clean, `prettier` clean, Slack unit + integration tests 27/27, `IntegrationType` consumers and Slack action tests 91/91.

**Commit:** `18445cfd3`

The latest autofix changes are on the `feature/slack-integration` branch.

ui/actions/integrations/slack.ts
```diff
@@ -0,0 +1,191 @@
...
+      // non-JSON `2xx` reads as "no URL" instead of throwing a parser message
+      // the user would be shown verbatim.
+      const body = await response.json().catch(() => null);
+      const authorizeUrl = body?.meta?.authorize_url;
+
+      if (typeof authorizeUrl !== "string" || authorizeUrl.length === 0) {
+        return { error: "Slack did not return an authorization URL." };
+      }
+
+      return { authorizeUrl };
```
Alan-TheGentleman, line 191: Please validate this URL before returning it to the client. The manager renders it under a trusted `Add to Slack` CTA, so any non-empty HTTPS URL from a malformed API response can redirect users to an attacker-controlled origin. Parse it here and require HTTPS, the expected Slack hostname, and the OAuth v2 authorization path, then cover hostile schemes and origins.

ui/actions/integrations/slack.ts
```diff
@@ -0,0 +1,191 @@
...
+    // before answering, so a cache filled when there was none would list the
+    // connected workspace as missing.
+    revalidatePath("/integrations");
+    revalidatePath("/integrations/slack");
+
+    if (!body?.data) {
+      return { unconfirmed: true, message: SLACK_UNREADABLE_RESULT_MESSAGE };
+    }
+
+    return { integration: parseStringify(body.data) as IntegrationProps };
```
Alan-TheGentleman, line 245: Please validate `body.data` before treating the exchange as successful. A truthy payload such as `{}`, `[]`, or `"invalid"` is cast to `IntegrationProps`, after which the callback reports a connected workspace and redirects. Validate the minimum Slack JSON:API resource shape and return the existing `unconfirmed` result when a successful response cannot be read back safely.

ui/components/integrations/slack/slack-integration-manager.tsx
```diff
@@ -0,0 +1,188 @@
...
+              <div className="text-xs text-gray-500 dark:text-gray-300">
+                {integration.attributes.connection_last_checked_at && (
+                  <p>
+                    <span className="font-medium">Last checked:</span>{" "}
+                    {format(
+                      new Date(
+                        integration.attributes.connection_last_checked_at,
+                      ),
+                      "yyyy/MM/dd",
+                    )}
```
Alan-TheGentleman: Please guard this timestamp before passing it to `date-fns/format`. An invalid API value produces an invalid `Date`, and `format` throws `RangeError`, replacing the integration page with an error boundary. Parse and check `isValid()` first, then omit the value or render a safe fallback and add a malformed-timestamp test.

Alan-TheGentleman (changes requested):
The Slack connection flow is well covered, but I am requesting changes for three correctness and trust-boundary gaps. Please validate the Slack authorization destination, validate the exchange response before reporting a successful install, and guard malformed timestamps before formatting them. The targeted Slack suites and the UI healthcheck pass at this head, so these cases need focused negative tests rather than broader test changes. Once these three paths are covered, I will re-review.

ui/actions/integrations/slack.ts
Alan-TheGentleman: Thanks, this now rejects non-object and incomplete top-level payloads, but the resource identity is still not validated. Values such as `{ id: "x", type: "tasks", attributes: {} }`, `{ id: "", type: "integrations", attributes: {} }`, or an integration with `integration_type: "jira"` still pass `isIntegrationResource` and reach the callback success path. Please require a non-empty `id`, `type === "integrations"`, and `attributes.integration_type === "slack"`, with negative tests for the wrong resource type and integration kind.

ui/actions/integrations/slack.ts
pfe-nazaries: Addressed in 5ed235088: `isSlackAuthorizeUrl` (ui/actions/integrations/slack.ts:88-99) requires `https:`, `hostname === "slack.com"` and `pathname === "/oauth/v2/authorize"` before the URL is returned (used at :171-176; any other value answers with the same copy as a missing URL). Hostile schemes and origins are pinned in ui/actions/integrations/slack.test.ts:179-223 — `javascript:`, plain HTTP, another origin, a lookalike hostname (`slack.com.evil.test`), another Slack path and a non-URL — plus a positive case asserting the consent URL's query survives untouched.

ui/actions/integrations/slack.ts
pfe-nazaries: Done in 9d43aa0c3: `isIntegrationResource` now also requires a non-empty `id`, `type === "integrations"` and `attributes.integration_type === "slack"`; every rejection keeps returning the existing unconfirmed result. Negative tests added for the empty id, the wrong resource type (`tasks`) and the wrong integration kind (`jira`), each asserting the unconfirmed outcome (ui/actions/integrations/slack.test.ts).

ui/components/integrations/slack/slack-integration-manager.tsx
pfe-nazaries: Addressed in 434e5aaf0: the timestamp is parsed and gated with `isValid` before `format` (ui/components/integrations/slack/slack-integration-manager.tsx:82), falling back to the never-checked rendering instead of the error boundary, with the malformed-timestamp test "keeps the page usable when the recorded check time is one no parser can read".

ui/components/icons/services/IconServices.tsx
```diff
@@ -752,6 +752,53 @@ export const JiraIcon: React.FC<IconSvgProps> = ({
...
+    fill="none"
+    focusable="false"
+    height={size || height}
+    role="presentation"
+    viewBox="0 0 48 48"
+    width={size || width}
+    className={className}
+    {...props}
+    xmlns="http://www.w3.org/2000/svg"
+  >
```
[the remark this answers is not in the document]
pfe-nazaries: The `aria-label` removal and `aria-hidden="true"` landed earlier in 8481a43fe; the leftover `role="presentation"` is removed in 23d6c77a7 — per ARIA, a presentational role is ignored whenever a global ARIA attribute is present, so the icon now carries the plain decorative pattern (`aria-hidden="true"` + `focusable="false"`). The sibling icons in this file predate this PR and are left for a separate a11y sweep.

ui/components/integrations/slack/slack-integration-card.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { SettingsIcon } from "lucide-react";
+import Link from "next/link";
+
+import { SlackIcon } from "@/components/icons/services/IconServices";
+import { Button, Card, CardContent, CardHeader } from "@/components/shadcn";
+import { CustomLink } from "@/components/shadcn/custom/custom-link";
+
+// Placeholder slug: the docs slice writes the page and confirms it.
+const SLACK_DOCS_URL =
+  "https://docs.prowler.com/projects/prowler-open-source/en/latest/tutorials/prowler-app-slack-integration/";
```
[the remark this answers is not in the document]
pfe-nazaries: Deliberate: the documentation page ships in PR #12438, the docs layer of this same stack, so the URL goes live when the stack lands. Keeping the link avoids a throwaway edit between layers.

ui/app/(prowler)/integrations/slack/slack-integration-content.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+  const integrations = await readSlackIntegrations(searchParams);
+  const loadError =
+    integrations && "error" in integrations
+      ? (integrations.error as string)
+      : null;
+  const integration: IntegrationProps | null =
+    (integrations?.data?.[0] as IntegrationProps | undefined) ?? null;
+
+  const authorize = integration ? null : await getSlackAuthorizeUrl();
```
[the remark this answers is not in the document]
pfe-nazaries: Deliberate and test-pinned: the failed read is the shared `GET /integrations`, while the consent URL comes from an independent call — one read failing says nothing about the Slack app, so the install stays on offer. Asserted in slack-page.integration.test.tsx ("keeps the page usable when reading the install fails on the server", `offersInstall()` → true); the suggested change would fail that test.

ui/app/(prowler)/integrations/slack/slack-page.integration.test.tsx
```diff
@@ -0,0 +1,222 @@
...
+    await harness.mount();
+
+    const consentScreen = new URL(await harness.connect());
+
+    expect(`${consentScreen.origin}${consentScreen.pathname}`).toBe(
+      "https://slack.com/oauth/v2/authorize",
+    );
+    expect((consentScreen.searchParams.get("scope") ?? "").split(",")).toEqual(
+      expect.arrayContaining(REQUIRED_SCOPES),
+    );
```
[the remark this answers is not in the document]
pfe-nazaries: Applied in 6adbb75ca — the test now asserts the exact scope set (length + membership).

Alan-TheGentleman (dismissed):
Re-reviewed exact head `84aa8668b552b88080540f8bb40c9f50aba4ac61`. The remaining exchange-response issue is resolved: installation success now requires a non-empty integration ID, JSON:API `type === "integrations"`, and `integration_type === "slack"`, with negative coverage for each invalid identity. The earlier Slack authorize-URL and malformed-timestamp fixes remain intact.

Focused verification passed 26 Slack action unit tests, 23 Slack browser integration tests, and the full UI healthcheck. The current GitHub workflow failure is infrastructure-only (`EAI_AGAIN` while fetching a release asset), not caused by this PR.

Alan-TheGentleman (approved):
Re-reviewed the current head. The Slack authorization URL boundary, exchange resource validation, malformed timestamp guard, and focused negative coverage now satisfy the requested changes. Approving.

PR #12283, opened 2026-08-03
feat(aws): add Nitro Enclaves security checks for EC2 and KMS

lydiavilchez:
### Context

Adds AWS Confidential Computing (Nitro Enclaves) security checks — the first CSPM coverage for confidential computing workloads on AWS.
All checks are fully passive (boto3 + CloudTrail), following Prowler's read-only architecture. None require instance access or SSM.

## Checks added

**Workload host environment (EC2):**
- `ec2_confidential_workload_host_imdsv2_not_enforced` — HIGH
- `ec2_confidential_workload_host_public_ip` — MEDIUM
- `ec2_confidential_workload_host_unrestricted_ingress` — HIGH
- `ec2_confidential_workload_host_vsock_proxy_exposed` — MEDIUM
- `ec2_confidential_workload_host_not_running` — MEDIUM

**KMS attestation policy:**
- `kms_key_enclave_attestation_not_enforced` — HIGH
- `kms_key_enclave_attestation_bypassable_path` — HIGH
- `kms_key_enclave_attestation_no_deployment_binding` — INFORMATIONAL
- `kms_key_enclave_debug_attestation_detected` — HIGH
- `kms_key_enclave_attestation_pcr_mismatch` — MEDIUM
- `kms_key_enclave_attestation_unknown_image` — MEDIUM

## Compliance coverage
  
  All 11 checks are mapped across **23 compliance frameworks**: NIST 800-53 Rev 5, PCI-DSS v3.2.1 & v4.0, ISO 27001:2022, SOC 2, HIPAA, AWS AI Security Framework, MITRE ATT&CK, CIS AWS Foundations, ENS, FedRAMP, FFIEC, GDPR, GxP, KISA ISMS-P, NIS2, NIST 800-171, NIST CSF 2.0, SecNumCloud, AWS Well-Architected.

### Checklist

<details>

<summary><b>Community Checklist</b></summary>

- [ ] This feature/issue is listed in the [open issues](https://github.com/prowler-cloud/prowler/issues?q=sort%3Aupdated-desc+is%3Aissue+is%3Aopen) or roadmap.prowler.com
- [ ] Is it assigned to me, if not, request it via the [open issues](https://github.com/prowler-cloud/prowler/issues?q=sort%3Aupdated-desc+is%3Aissue+is%3Aopen) or [Prowler Community Slack](https://goto.prowler.com/slack)
[...]

PR #11335, opened 2026-05-22
feat(bedrock): add bedrock_agent_role_least_privilege check

ARYAN03B:
### Context

Implements the new check requested in #11011 — flags AWS Bedrock Agent execution roles that violate least-privilege principles.

The 4th criterion from the issue (VPC condition keys when applicable) is intentionally deferred to a follow-up PR per discussion in the issue thread, so smaller/focused review.

Refs #11011

### Description

Adds `bedrock_agent_role_least_privilege` for the AWS provider. A Bedrock Agent's execution role **FAILs** the check when any of these are true:

- An AWS-managed `*FullAccess` policy is attached.
- An inline statement allows broad actions (e.g. `s3:*`, `iam:*`, `*`) on `Resource: "*"`.
- No permissions boundary is configured on the role.

**Risk model:** a permissive agent role turns successful prompt injection into AWS privilege escalation — the LLM can be coerced into calling any API the role allows. Least privilege + a permissions boundary bounds the blast radius even when guardrails fail.

**Service-layer changes (touching shared files):**

- `iam_service.Role` gains a `permissions_boundary: Optional[dict]` field, populated from the existing `list_roles` response. Real AWS returns this field on `ListRoles`, so no extra API call is added in production.
- `BedrockAgent` service captures each agent's `agentResourceRoleArn` via a new per-agent `GetAgent` call (the `ListAgents` summary doesn't include it).

**Check logic:**

- Cross-service: imports both `bedrock_agent_client` and `iam_client` (same pattern as `bedrock_full_access_policy_attached`).
- Conservative `BROAD_ACTION_PATTERNS` heuristic — only flags actions like `s3:*` / `iam:*` / `*` on `Resource:*`, not narrow actions like `s3:GetObject`. Happy to widen or narrow based on review.
- Collects all violations per role and reports them in a single FAIL message.

No new dependencies.

### Steps to review

1. **`iam_service.py` diff** — 2-line addition (new field on `Role`, populate from `list_roles` response). Confirms no behaviour change for existing IAM checks.
2. **`bedrock_service.py` diff** — new `_get_agent` method (modelled on `_get_prompt`) + `role_arn` field on `Agent`.
3. **New check `bedrock_agent_role_least_privilege.py`** — main logic + `BROAD_ACTION_PATTERNS` heuristic.
4. **`bedrock_agent_role_least_privilege.metadata.json`** — title, severity (high), risk, remediation.
5. **Run the new tests:**
   ```
[...]

PR #12459, opened 2026-08-14
feat(bedrock): add model artifact and guardrail grounding security checks for the AWS provider

tamg-aws:
### Context

Prowler's Bedrock coverage asserts a lot about guardrails, prompts, and API keys, but not about the
**model artifacts and grounding configuration** a Bedrock deployment actually depends on. Four gaps,
each one a customer decision that currently scans clean whatever its value:

- A guardrail can carry a contextual-grounding policy whose filters are set to `action: NONE` or
  `threshold: 0` — configured, reported, and blocking nothing.
- A custom model can be encrypted under an AWS-owned key rather than one the organization controls.
- A knowledge-base data source can be unencrypted at rest.
- Several Bedrock Agents can share one execution role, so every agent inherits the union of their
  permissions and CloudTrail cannot attribute an action to a single agent.

### Description

Adds **4 checks** to the existing `bedrock` service. This PR is independent of any AgentCore work —
every check reads `bedrock_client` or `bedrock_agent_client` only, and the service file it ships
contains zero references to AgentCore (verified: 19 files, 144 tests passing on `botocore 1.40.61`
against a clean clone of master).

| Check | Severity | Asserts |
|---|---|---|
| `bedrock_guardrail_contextual_grounding_filter_enabled` | high | both GROUNDING and RELEVANCE filters present, each not `enabled: false`, `action: BLOCK`, `threshold > 0` |
| `bedrock_custom_model_encrypted_with_cmk` | critical | `GetCustomModel.modelKmsKeyArn` present |
| `bedrock_knowledge_base_encrypted_with_cmk` | high | `GetDataSource.dataSource.serverSideEncryptionConfiguration.kmsKeyArn` present |
| `bedrock_agent_role_not_shared_across_agents` | high | `GetAgent.agent.agentResourceRoleArn` unique per agent |

The recurring theme is that **a configured-but-permissive setting fails**, because that is the state
which reads as protected while doing nothing: a grounding filter with `action: NONE` or a zero
threshold, and a grounding filter with `enabled: false`. Treating "the policy exists" as sufficient would make
these checks worse than absent.

Every check here distinguishes *unknown* from *compliant*: when a `Get*` call fails (AccessDenied,
throttling) the resource reports **MANUAL**, never PASS, since Prowler sets unretrieved fields to
`None` and gating a PASS on absence would emit false negatives exactly when permissions are too narrow
to tell.

The same rule is applied one level up, to the `List*` calls that build the inventory. A denied
`ListCustomModels` or `ListKnowledgeBases` would otherwise leave the region out of the report
entirely, and a denied `ListDataSources` would drop the knowledge base — both indistinguishable from
a clean result. Those cases now report **MANUAL** against `custom-model/unknown`,
`knowledge-base/unknown`, or the knowledge base whose data sources could not be enumerated. A
`ValidationException` from `List*` is deliberately excluded, because it means the service is not
available in the Region — a definite "no resources", not an unknown — and recording it would emit a
MANUAL finding for every Region Bedrock does not serve.

`ListCustomModels` is called with `isOwned=True`. Without it the response also carries models shared
into the account through Resource Access Manager, whose KMS key the audited account neither owns nor
can set, so the check would emit a `critical` FAIL that nobody in that account can remediate.

The grounding check extends that to two *optional response members*, and this is the one behaviour
worth flagging because it is the least obvious: `enabled` and `action` are optional in
`GuardrailContextualGroundingFilter` (only `type` and `threshold` are required) and AWS documents no
default for either. A filter that is otherwise compliant but omits one therefore reports **MANUAL**,
not PASS — reading an omitted `action` as `NONE` would assert a misconfiguration the response never
stated. An explicit `enabled: false`, a non-BLOCK `action`, or a zero `threshold` is a definite finding
and still FAILs, so an unknown never masks a real one. This follows the existing AWS checks that use `MANUAL` the same
way.

`bedrock_agent_role_not_shared_across_agents` excludes agents whose `GetAgent` failed from the
share count rather than counting them, so one AccessDenied cannot manufacture a false FAIL against a
readable agent. There is a dedicated test for that.

Supporting changes:

- `prowler/providers/aws/services/bedrock/bedrock_service.py` — new collectors on the existing
  `Bedrock` and `BedrockAgent` classes: contextual-grounding filters on the guardrail, custom-model
  listing plus `modelKmsKeyArn`, knowledge-base and data-source listing plus the SSE config, and the
  agent's `agentResourceRoleArn`. No new service class and no new service
  directory.
- `prowler/compliance/aws/aws_ai_security_framework_aws.json` — maps the 4 CheckIDs onto four
  requirements, appending only (no existing entry is replaced):

  | Requirement | Checks before | after | Added |
  |---|---|---|---|
  | `AISF-AI-09` Contextual Grounding for RAG Validation | 0 | 1 | `bedrock_guardrail_contextual_grounding_filter_enabled` |
  | `AISF-AI-10` Knowledge Base Security for RAG | 0 | 1 | `bedrock_knowledge_base_encrypted_with_cmk` |
  | `AISF-INFRA-05` Encryption at Rest for AI Data | 14 | 16 | `bedrock_custom_model_encrypted_with_cmk`, `bedrock_knowledge_base_encrypted_with_cmk` |
  | `AISF-AGENT-01` Agent Identity and Authentication | 0 | 1 | `bedrock_agent_role_not_shared_across_agents` |

  Three of these are literal matches: `AISF-AI-09` is named "Contextual Grounding for RAG Validation";
  `AISF-AI-10` explicitly asks for knowledge bases to be encrypted "with customer-managed keys"; and
  `AISF-INFRA-05` is the general encryption-at-rest requirement that already holds 14 CMK checks.
  `bedrock_knowledge_base_encrypted_with_cmk` is deliberately mapped to both `AISF-AI-10` (the
  RAG-specific requirement it satisfies exactly) and `AISF-INFRA-05` (the cross-service one, where its
  16 siblings live) — say the word if you would rather it appear only once.

  The fourth is worth its reasoning: **`AISF-AGENT-01`** is about per-agent identity with scoped
  credentials. Several Bedrock Agents sharing one execution role means each inherits the union of the
  others' permissions and CloudTrail cannot attribute an action to a single agent — the identity
  separation the requirement asks for, asserted at the IAM layer.

  Note `AISF-AI-09`, `AISF-AI-10` and `AISF-AGENT-01` are currently `Type: Manual` with empty `Checks`
  arrays. Since they now carry an automated check, this PR also flips their `Type` to `Automated`,
  which is what the rest of the corpus does: across all shipped compliance frameworks 353 `Manual`
  requirements carry zero checks and none carries any.

No permissions change is needed: `permissions/prowler-additions-policy.json` already grants
`bedrock:List*` and `bedrock:Get*`.

One metadata field is deliberately empty. `bedrock_custom_model_encrypted_with_cmk` ships no
`Remediation.Code.CLI`, because a custom model's KMS key cannot be changed after creation — the only
"fix" is to re-run customization, which creates a different model rather than remediating the flagged
one. The check-metadata guidelines say to leave the field empty when no single command resolves the
finding, which is what the closest existing siblings do (`rds_instance_storage_encrypted` and
`sagemaker_notebook_instance_encryption_enabled`, both encryption-at-creation checks, and 155 of the
640 AWS checks on `master`). The console steps remain in `Remediation.Code.Other`.

### Steps to review

1. Every API path was verified against the **pinned** `botocore==1.40.61`, not the latest release —
   all four checks work on master's pin as-is. Confirmed field shapes:
   `contextualGroundingPolicy.filters[].{type,threshold,action,enabled}` — `action` enum
   `BLOCK|NONE`, with only `type` and `threshold` required, so `action` and `enabled` are both
   optional; `GetCustomModel.modelKmsKeyArn`; `GetDataSource` nesting everything under a top-level
   `dataSource` object; and `GetAgent` returning `agent.agentResourceRoleArn`.
2. `moto==5.1.11` does not implement `ListDataSources` or `GetDataSource`, so the
   new collectors are exercised through explicit `_make_api_call` patching.
   Each collector's `try/except` means the pre-existing Bedrock tests are unaffected by the added
   calls — verified by running the whole `tests/providers/aws/services/bedrock` suite.
3. Each check has PASS, FAIL for every distinct failure branch, no-resources, region-unsupported,
   and unreadable→MANUAL scenarios. The two checks whose resource type cannot be created in a test
   account additionally have a test that enumerates their entire decision space, so no reachable
   input combination is left unasserted.
4. Evidence quality, stated plainly, because it differs per check:
   - `bedrock_guardrail_contextual_grounding_filter_enabled` and
     `bedrock_knowledge_base_encrypted_with_cmk` are **live-verified**. Against a real account they
     produced 11 findings — 2 guardrails and 9 knowledge base data sources — and every verdict was
     confirmed against the raw API. PASS, `action: NONE`, `threshold: 0` and a missing filter type
     were each proven with purpose-built guardrails, and the MANUAL path was proven under an explicit
     IAM deny on `ListKnowledgeBases`.
   - `bedrock_custom_model_encrypted_with_cmk` and `bedrock_agent_role_not_shared_across_agents` are
     **field-verified and exhaustively unit-tested, but not live-verified**. Neither resource type can
     be created in a test account: Bedrock Agents refuses new agents for accounts without prior
     service usage, and a custom model requires a model customization job. Their API paths and
     response shapes are confirmed against the pinned service model and survive the pinned
     `rest-json` parser with no field dropped. Because their behaviour cannot be observed against a
     real resource, the tests enumerate the whole decision space rather than sampling it: all 84
     inventories of one to three agents over {role A, role B, no role} x {readable, unreadable}, and
     all 12 combinations of retrieval outcome, key value and regional listing outcome. I have not
     observed AWS emit those responses, and say so rather than implying otherwise.

### Checklist

- [x] Ensure a changelog fragment is added under `prowler/changelog.d/`.

<details>

- [ ] This feature/issue is listed in the open issues or roadmap.prowler.com
- [ ] Is it assigned to me, if not, request it via the open issues or Prowler Community Slack
- [x] I have reviewed the open pull requests and confirmed there is no existing PR that implements the same outcome

</details>

#### SDK/CLI

- Are there new checks included in this PR? **Yes — 4.** No provider permission change required; a
  changelog fragment is included.

### License

## Summary by CodeRabbit

* **New Features**
  * Added AWS Bedrock security checks for:
    * Contextual grounding and relevance filters on guardrails.
    * Customer-managed key encryption for custom models and knowledge-base data sources.
    * Execution roles shared across agents and deployed versions.
  * Added manual review findings when resource details or inventories are unavailable.

* **Tests**
  * Added comprehensive coverage for compliant, non-compliant, unsupported, incomplete, and unavailable-resource scenarios.


prowler/providers/aws/services/bedrock/bedrock_custom_model_encrypted_with_cmk/bedrock_custom_model_encrypted_with_cmk.metadata.json
```diff
@@ -0,0 +1,43 @@
...
+      "Other": "1. Open the AWS Console and go to Amazon Bedrock\n2. Select **Custom models**, then create a new customization job\n3. Under encryption, choose a customer-managed KMS key instead of the default\n4. Grant the customization service role kms:Decrypt, kms:GenerateDataKey and kms:DescribeKey on that key\n5. Re-run customization, because the key of an existing custom model cannot be changed in place",
+      "Terraform": "```hcl\nresource \"aws_bedrock_custom_model\" \"example_resource\" {\n  custom_model_name     = \"example-custom-model\"\n  job_name              = \"example-customization-job\"\n  base_model_identifier = \"arn:aws:bedrock:us-east-1::foundation-model/amazon.titan-text-express-v1\"\n  role_arn              = \"arn:aws:iam::123456789012:role/example-bedrock-customization-role\"\n  custom_model_kms_key_id = \"arn:aws:kms:us-east-1:123456789012:key/example-key-id\" # Critical: without this the model is encrypted with an AWS owned key\n\n  hyperparameters = {\n    epochCount = \"1\"\n  }\n\n  output_data_config {\n    s3_uri = \"s3://example-bucket/output/\"\n  }\n\n  training_data_config {\n    s3_uri = \"s3://example-bucket/training/\"\n  }\n}\n```"
```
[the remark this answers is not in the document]
tamg-aws: Leaving `Remediation.Code.CLI` empty here is deliberate.

A custom model's KMS key cannot be changed after creation — there is no `UpdateCustomModel` operation on the `bedrock` client, only `CreateCustomModel`, `GetCustomModel`, `ListCustomModels` and `DeleteCustomModel`. `create-model-customization-job` produces a *new* model and leaves the flagged one non-compliant, so it does not take this finding from FAIL to PASS. `check-metadata-guidelines.mdx` is explicit on that case: "If the issue cannot be resolved with a single command, leave this field empty."

The sibling `bedrock_knowledge_base_encrypted_with_cmk` supplies a CLI because `update-data-source` fixes the flagged resource in place. That distinction — updatable vs set-at-creation — is what separates the two.

This also follows existing practice for encryption that can only be set at creation: `rds_instance_storage_encrypted`, `efs_encryption_at_rest_enabled`, `sagemaker_notebook_instance_encryption_enabled` and `sagemaker_training_jobs_volume_and_output_encryption_enabled` all ship an empty CLI. In `e5b86da6e` (#9551) the command `aws rds create-db-instance --storage-encrypted true` was deliberately replaced with `""` for exactly this reason.

The console procedure remains in `Remediation.Code.Other`, and the `Notes` field records that the key cannot be changed in place.

tamg-aws:
Thanks for the review — all six findings have been addressed in `eaa3f95d` and `8ce87a57`. Summary for reviewers:

**Fixed**

1. **Incomplete agent inventory no longer returns PASS.** `_list_agents` was swallowing `ListAgents` failures, so a Region whose agents could not be listed contributed nothing while every readable agent still reported PASS. Failures are now recorded per Region, that Region gets a MANUAL finding against `agent/unknown`, and an agent whose role only *looks* dedicated reports MANUAL while any Region is missing. A role already seen on two readable agents still FAILs, since more inventory cannot make it unshared.
2. **`update-guardrail` remediation is now executable.** It omitted `--name`, `--blocked-input-messaging` and `--blocked-outputs-messaging`, all required by `UpdateGuardrail`.
3. **The MANUAL message no longer asserts blocking.** It previously said the guardrail "blocks ungrounded responses" and then that blocking was unknown; it now states only what the response confirms.
4. **The exhaustive matrix tests build the real service.** They used local fakes, so a renamed service attribute would not have failed them. They now construct `Bedrock`/`BedrockAgent` from parameterised botocore stubs — which immediately caught a missing `agents_scan_errors` attribute.
5. **The duplicate knowledge-base test now asserts the distinguishing state** (`data_sources_listed` true with no scan errors) instead of repeating an existing assertion.

**Not changed:** `Remediation.Code.CLI` on `bedrock_custom_model_encrypted_with_cmk` stays empty — reasoning in the thread above.

**Also found by re-auditing the whole diff, not raised in review**

- The guardrail check was the last of the four still going silent on a failed inventory: `guardrails_scan_errors` was populated by the service but read by nobody. It now reports MANUAL against `guardrail/unknown`.
- Two `NativeIaC` snippets were not deployable. `cfn-lint` rejected the guardrail template for the required `BlockedInputMessaging`/`BlockedOutputsMessaging`, and the data source template for the required `DataSourceConfiguration`, a `KnowledgeBaseId` that failed its pattern, and a KMS ARN that was not the 36-character form. All three snippets now pass `cfn-lint` clean.
- Inline comment density is cut to 0.0–5.5% per check file and 3.7% in the service, against an upstream median of 0.0%.

**Verification:** 146 bedrock tests and 4,027 across the bedrock and check-model suites pass on the pinned `botocore 1.40.61`; 19 files; `black`/`isort`/`flake8`/`pylint`/`bandit`/`vulture` clean. Two of the four checks were verified against live Bedrock resources (11 findings, each cross-checked against the raw API, plus PASS/`action: NONE`/zero-threshold/missing-filter and a MANUAL under an explicit IAM deny). The other two cannot be created in a test account — Bedrock Agents refuses new agents for accounts without prior usage, and a custom model needs a customization job — so their behaviour is pinned by enumerating the whole decision space instead: all 84 agent inventories and all 12 custom-model states.

tamg-aws:
@coderabbitai full review

tamg-aws:
@coderabbitai full review

tamg-aws:
@coderabbitai full review

tamg-aws:
@coderabbitai full review

tamg-aws:
@coderabbitai full review

prowler/providers/aws/services/bedrock/bedrock_service.py
```diff
@@ -118,6 +138,64 @@ def _list_tags_for_resource(self, guardrail):
...
+            model_info = self.regional_clients[model.region].get_custom_model(
+                modelIdentifier=model.arn
+            )
+            model.kms_key_arn = model_info.get("modelKmsKeyArn")
+            model.detail_retrieved = True
+        except Exception as error:
+            logger.error(
+                f"{model.region} -- {error.__class__.__name__}[{error.__traceback__.tb_lineno}]: {error}"
+            )
+
```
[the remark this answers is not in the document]
tamg-aws: `ListCustomModels` **is** pageable at the pin, so this stands as written.

```
botocore 1.40.61
can_paginate('list_custom_models') -> True
get_paginator('list_custom_models') -> Bedrock.Paginator.ListCustomModels
paginate(isOwned=True) -> 1 page, no OperationNotPageableError
```

That last line is a live call against a real account, not a stub. `botocore/data/bedrock/2023-04-20/paginators-1.json` lists `ListCustomModels` alongside `ListGuardrails`, `ListImportedModels` and 14 others, and the 149 unit tests exercise this path through the real service object. No `OperationNotPageableError` is reachable here.

prowler/providers/aws/services/bedrock/bedrock_guardrail_contextual_grounding_filter_enabled/bedrock_guardrail_contextual_grounding_filter_enabled.py
```diff
@@ -0,0 +1,118 @@
...
+                report.status = "MANUAL"
+                report.status_extended = f"Bedrock Guardrail {guardrail.name} contextual grounding policy could not be retrieved in region {guardrail.region}; verify manually that GROUNDING and RELEVANCE filters block ungrounded responses."
+                findings.append(report)
+                continue
+
+            filters_by_type = {
+                filter.get("type"): filter
+                for filter in guardrail.contextual_grounding_filters
+                if filter.get("type")
+            }
```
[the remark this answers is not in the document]
tamg-aws: Leaving one filter per type, as you note this is defensive only.

`GuardrailContextualGroundingPolicyConfig.filtersConfig` accepts at most one entry per `type` — `CreateGuardrail` and `UpdateGuardrail` both reject a duplicate `GROUNDING` or `RELEVANCE` entry, so `GetGuardrail` cannot return two of the same type. Evaluating every filter instead of one per type would add a branch that no API response can reach, which is harder to justify than the current shape.

If AWS ever allows multiple filters of one type, the fix belongs with that change so the semantics of a duplicate can be defined then, rather than guessed now.

tests/providers/aws/services/bedrock/bedrock_agent_role_not_shared_across_agents/bedrock_agent_role_not_shared_across_agents_test.py
```diff
@@ -0,0 +1,456 @@
...
+            mock.patch(
+                "prowler.providers.aws.services.bedrock.bedrock_agent_role_not_shared_across_agents.bedrock_agent_role_not_shared_across_agents.bedrock_agent_client",
+                new=BedrockAgent(aws_provider),
+            ),
+        ):
+            from prowler.providers.aws.services.bedrock.bedrock_agent_role_not_shared_across_agents.bedrock_agent_role_not_shared_across_agents import (
+                bedrock_agent_role_not_shared_across_agents,
+            )
+
+            return bedrock_agent_role_not_shared_across_agents().execute()
```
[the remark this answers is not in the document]
tamg-aws: Keeping the two helpers separate.

They do not differ only in the region list: `_run` patches the check's `bedrock_agent_client` for a single-Region provider, while `_run_multi_region` builds a two-Region provider so a per-Region `ListAgents` failure can be exercised on one Region while another succeeds. Parameterising one helper on `regions` would work, but every existing test would then carry an argument that only one test varies, which reads as less clear at the call site rather than more.

Happy to collapse them if you would rather see it that way.

prowler/changelog.d/bedrock-model-artifact-checks.added.md
```diff
@@ -0,0 +1 @@
+`bedrock_guardrail_contextual_grounding_filter_enabled`, `bedrock_custom_model_encrypted_with_cmk`, `bedrock_knowledge_base_encrypted_with_cmk` and `bedrock_agent_role_not_shared_across_agents` AWS Bedrock checks, covering guardrail contextual grounding, custom model encryption, knowledge-base data-source encryption, and non-shared agent execution roles.
```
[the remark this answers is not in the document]
tamg-aws: Done in 949332283 — the fragment is now a complete sentence:

> `bedrock_guardrail_contextual_grounding_filter_enabled`, `bedrock_custom_model_encrypted_with_cmk`, `bedrock_knowledge_base_encrypted_with_cmk` and `bedrock_agent_role_not_shared_across_agents` are four new AWS Bedrock checks covering guardrail contextual grounding, custom model encryption, knowledge-base data-source encryption, and non-shared agent execution roles.

It keeps the backtick-first opening, which is the house style for check-adding entries in 203 of the 205 such bullets in `prowler/CHANGELOG.md`.

tamg-aws:
All review threads are resolved (18/18) at `9493322`. Since the last summary:

**Fixed**

- **A dedicated agent role is no longer asserted on incomplete evidence.** An agent whose `GetAgent` failed is excluded from the role index so it cannot manufacture a FAIL, but its role may be the same one another agent holds — so that other agent reported a PASS the data did not support. The check now collects every gap, an unlisted Region and an unreadable agent role alike, and asserts dedication only when none exists. FAIL is unaffected: a role seen on two readable agents is shared regardless of what else is unknown. The test that previously asserted the false PASS is corrected, the 84-inventory matrix encodes the stricter rule, and there is a dedicated regression test.
- **`contextual_grounding_filters` is now genuinely typed.** It was annotated `list[ContextualGroundingFilter]` while the collector stored dicts; pydantic does not validate assignment, so the annotation bought nothing. The collector builds model instances and the check reads attributes.
- **The Terraform data source example is deployable** — `data_source_configuration` is required by `aws_bedrockagent_data_source`. The CloudFormation equivalent was fixed earlier; this was the same gap in the sibling format.
- **A failed `ListDataSources` now names its error code** in the finding, matching the Region-level and detail-level messages.
- **The changelog fragment** is a complete sentence and describes what each check actually covers.

**Not changed, with reasoning in the threads**

- `Remediation.Code.CLI` on the custom model check stays empty: there is no `UpdateCustomModel`, so no single command takes the finding from FAIL to PASS.
- `ListCustomModels` is pageable at the pin — verified with a live paginated call.
- One filter per type is kept: the API rejects duplicates, so the alternative branch is unreachable.
- The two test helpers stay separate; they build different Region sets, not just different lists.

**Verification at this head:** 149 bedrock tests and 4,029 across the bedrock and check-model suites pass on the pinned `botocore 1.40.61`; 19 files; `black`, `isort`, `flake8`, `pylint`, `bandit`, `vulture`, `cfn-lint` and the changelog and compliance-mapping gates all clean.

Beyond the individual findings I swept for the two patterns behind them, so they should not recur: every collector in the service was checked for whether a failure reaches a check (all five added here are observable; the four silent ones are pre-existing), and every remediation format was checked against its provider's required arguments (the other two Terraform examples were already complete).

tamg-aws:
@coderabbitai full review

tamg-aws:
Addressed in `ef31a75e8`.

**Deployed versions are now audited, not just the working draft.** The finding was correct: `GetAgent` returns only the draft, an agent version is an immutable snapshot that keeps the role it was cut with, and an alias routes invocations at a specific version — so two agents with distinct draft roles could both be invoking one shared role in production while the check called that compliant.

`_get_agent_version_roles` lists each agent's aliases, collects the versions they route to, and reads each version's `agentResourceRoleArn`. Every role an agent can run under is folded into the sharing index, and the FAIL message names the version when the sharing is through one:

```
FAIL  agent alpha shares execution role arn:...:role/SharedVersionRole
      through deployed version 3 with bravo in region us-east-1, ...
```

Two scoping decisions worth noting:

- **Only routed versions are read.** A version no alias points at cannot be invoked, so reporting on it would be noise. `ListAgentAliases` returns `routingConfiguration` in the summary, so this needs no per-alias `GetAgentAlias`.
- **An alias routing at `DRAFT` needs no extra call** — that is the role `GetAgent` already returned. `routingConfiguration[].agentVersion` is an optional member, so an absent one is skipped rather than guessed at.

The inventory rule is unchanged: unlistable aliases or an unreadable version leave the picture incomplete, so no agent can be called dedicated — while a role already seen on two agents is shared whatever else is missing, so a partial read still FAILs when the half that was read shows the sharing.

**Remediation rewritten** for the second half of the finding. `UpdateAgent` requires four fields and does not preserve the ten optional ones it is not given, and it leaves existing versions untouched — so an alias left routing at the old version keeps invoking the shared role. The sequence is `UpdateAgent` with the agent's full configuration, `PrepareAgent` to cut a version, then `UpdateAgentAlias` for each affected alias. No single command expresses that, so the CLI field is empty per the metadata guidelines and the steps are in the recommendation text.

8 new tests cover shared-version-role, distinct-version-roles, draft-only sharing, `DRAFT` routing, no-alias, unlistable aliases, unreadable version, and definite-finding-beats-incomplete-inventory. 157 pass. `bedrock:List*` / `bedrock:Get*` in `prowler-additions-policy.json` already cover both new calls.

tamg-aws:
@coderabbitai full review

tamg-aws:
Partly addressed in `3652e38d0`, and partly declining with evidence.

**Declining the empty-CLI part.** `check-metadata-guidelines.mdx:155` states: *"If the issue cannot be resolved with a single command, leave this field empty."* This remediation is inherently three sequenced calls — `UpdateAgent` with the agent's full configuration, `PrepareAgent` to cut a version, then `UpdateAgentAlias` per affected alias — so no single command expresses it. Cramming a `&&` chain in would break the "do not include multiple commands" rule on line 154.

This also matches shipped practice rather than departing from it: **155 of 640 merged AWS checks (24%) ship an empty `Remediation.Code.CLI`**, and of the empty-CLI checks I sampled, **94/94 carry the guidance in `Remediation.Code.Other`** — which is exactly the structure here. There is precedent in this file's own history too: `e5b86da6e` (PR #9551, by a maintainer) deleted a `create-db-instance` CLI from `rds_instance_storage_encrypted` in favour of `""` because the property is set at creation.

On the second half of the claim — `Remediation.Recommendation.Text` is **not** empty. It is 349 characters and already names all three steps including the alias update.

**Taking the underlying point, which was right.** The console steps in `Remediation.Code.Other` stopped at "prepare the agent" and never said to move the aliases, so someone following them would repoint the draft, cut a version, and leave every alias still routing at the old version — still invoking the shared role. That is the same defect the code change fixes, left sitting in the remediation. Now an explicit step:

```
5. Edit the agent to use the new role, keeping its other settings, then prepare the agent to cut a version
6. Open **Aliases** and point every alias at the new version, since an alias left on the old version keeps invoking the shared role
```

157 tests pass; metadata re-validated against `CheckMetadata`.

prowler/providers/aws/services/bedrock/bedrock_agent_role_not_shared_across_agents/bedrock_agent_role_not_shared_across_agents.metadata.json
```diff
@@ -0,0 +1,42 @@
...
+  "Description": "Every **Bedrock Agent** assumes the role in its `agentResourceRoleArn`. That role must belong to exactly one agent, so no agent inherits another's permissions.",
+  "Risk": "A shared execution role gives every agent using it the union of all their permissions, so a low-trust agent reaching an untrusted tool holds the rights granted for a high-trust one, turning a single prompt injection into access it was never scoped for. It also destroys attribution: CloudTrail records the role session, so an action cannot be traced to one agent.",
+  "RelatedUrl": "",
+  "AdditionalURLs": [
+    "https://docs.aws.amazon.com/bedrock/latest/userguide/agents-permissions.html",
+    "https://docs.aws.amazon.com/bedrock/latest/APIReference/API_agent_UpdateAgent.html"
+  ],
+  "Remediation": {
+    "Code": {
+      "CLI": "",
```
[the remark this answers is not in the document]
tamg-aws: Declining the empty-CLI change, and taking the point underneath it. Detail in [this comment](https://github.com/prowler-cloud/prowler/pull/12459#issuecomment-5304102320); summary:

- `check-metadata-guidelines.mdx:155` says to leave `CLI` empty when one command cannot fix the issue, and line 154 forbids multiple commands. This remediation is three sequenced calls (`UpdateAgent` → `PrepareAgent` → `UpdateAgentAlias` per alias), so either rule would be broken by filling it.
- **155 of 640 merged AWS checks (24%) ship an empty `CLI`**, and 94/94 of the empty-CLI checks I sampled carry the steps in `Remediation.Code.Other` — the structure used here.
- `Remediation.Recommendation.Text` is not empty: 349 characters, already naming all three steps.

The underlying point was right, though. The console steps stopped at "prepare the agent" and never said to move the aliases, so following them left every alias on the old version still invoking the shared role. Fixed in `3652e38d0`.

tamg-aws:
@coderabbitai full review

tests/providers/aws/services/bedrock/bedrock_knowledge_base_encrypted_with_cmk/bedrock_knowledge_base_encrypted_with_cmk_test.py
```diff
@@ -0,0 +1,291 @@
...
+        assert result[0].status_extended.endswith(".")
+
+    @mock.patch("botocore.client.BaseClient._make_api_call", new=_mock_no_data_sources)
+    @mock_aws
+    def test_listed_but_empty_is_not_manual(self):
+        """A knowledge base with zero data sources is silent, not MANUAL.
+
+        Guards the over-correction: only an unreadable list is unknown.
+        """
+        assert self._run() == []
```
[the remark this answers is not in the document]
tamg-aws: Addressed in `8770ec384`. The duplicate was replaced rather than deleted: `test_listed_but_empty_marks_the_knowledge_base_as_listed` now asserts the *service state* that produces the empty result — `data_sources_listed is True`, `data_sources == {}`, and `knowledge_bases_scan_errors == {}` — so the over-correction of reporting MANUAL for a knowledge base that genuinely has no data sources cannot regress silently. That is coverage the original assertion did not have.

prowler/providers/aws/services/bedrock/bedrock_service.py
```diff
@@ -118,6 +135,64 @@ def _list_tags_for_resource(self, guardrail):
...
+                        )
+        except ClientError as error:
+            code = error.response["Error"].get("Code", error.__class__.__name__)
+            # ValidationException means Bedrock is unavailable in the region:
+            # a definite "no models", so it must not become a MANUAL finding.
+            if code != "ValidationException":
+                self.custom_models_scan_errors[regional_client.region] = code
+            logger.error(
+                f"{regional_client.region} -- {error.__class__.__name__}[{error.__traceback__.tb_lineno}]: {error}"
+            )
```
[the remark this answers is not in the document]
tamg-aws: Addressed in `4cb126857`. All four collections now filter `ValidationException` in the collector before storing the scan error, so there is one convention rather than two. Verified by execution: with every `List*` raising `ValidationException`, all four error stores are empty; with `AccessDeniedException`, all four record the code.

prowler/providers/aws/services/bedrock/bedrock_service.py
```diff
@@ -229,6 +340,103 @@ def _get_prompt(self, prompt):
...
+                        region=knowledge_base.region,
+                        knowledge_base_id=knowledge_base.id,
+                        knowledge_base_name=knowledge_base.name,
+                    )
+            knowledge_base.data_sources_listed = True
+        except Exception as error:
+            logger.error(
+                f"{knowledge_base.region} -- {error.__class__.__name__}[{error.__traceback__.tb_lineno}]: {error}"
+            )
+
```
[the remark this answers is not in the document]
tamg-aws: Addressed in `979dc656f`. `_list_data_sources` now records the API error code on the parent as `knowledge_base.data_sources_error`, and the check reports it in the message rather than emitting a bare MANUAL: `"... data sources could not be listed (AccessDeniedException) ..."`. The success path sets `data_sources_listed = True`, so an empty data-source set is distinguishable from an unread one.

prowler/providers/aws/services/bedrock/bedrock_service.py
```diff
@@ -100,6 +107,19 @@ def _get_guardrail(self, guardrail):
...
+                {
+                    "type": filter.get("type"),
+                    "threshold": filter.get("threshold"),
+                    "action": filter.get("action"),
+                    "enabled": filter.get("enabled"),
+                }
+                for filter in guardrail_info.get("contextualGroundingPolicy", {}).get(
+                    "filters", []
+                )
+            ]
```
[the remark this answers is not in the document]
tamg-aws: Addressed in `949332283`. `Guardrail.contextual_grounding_filters` is now genuinely `list[ContextualGroundingFilter]`: the service constructs model instances instead of dicts, and the check reads them by attribute (`filter.type`, `filter.action`, `filter.threshold`, `filter.enabled`). You were right that pydantic does not coerce on assignment, so the declared type and the stored value had disagreed.

prowler/providers/aws/services/bedrock/bedrock_agent_role_not_shared_across_agents/bedrock_agent_role_not_shared_across_agents.py
```diff
@@ -0,0 +1,99 @@
...
+                    if other_arn != agent.arn
+                )
+                report.status = "FAIL"
+                report.status_extended = f"Bedrock Agent {name} shares execution role {agent.role_arn} with {', '.join(others)} in region {agent.region}, so each agent inherits the union of their permissions and CloudTrail cannot attribute an action to one of them."
+            elif incomplete:
+                report.status = "MANUAL"
+                report.status_extended = f"Bedrock Agent {name} execution role is used by no other agent whose role could be read in region {agent.region}, but the role of {', '.join(incomplete)} is unknown; verify manually that none of them shares it."
+            else:
+                report.status = "PASS"
+                report.status_extended = f"Bedrock Agent {name} has a dedicated execution role in region {agent.region}."
```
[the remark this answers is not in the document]
tamg-aws: Addressed in `ef31a75e8`. Deployed versions are now audited, not just the working draft — `_get_agent_version_roles` reads each alias-routed version's `agentResourceRoleArn` and folds every role an agent can run under into the sharing index, so two agents with distinct drafts that share a deployed-version role now both FAIL with the version named. Only routed versions are read (an unrouted version cannot be invoked), and an alias routing at `DRAFT` needs no extra call. Remediation updated for the `UpdateAgent` → `PrepareAgent` → `UpdateAgentAlias` sequence. Detail in [this comment](https://github.com/prowler-cloud/prowler/pull/12459#issuecomment-5304018876).

prowler/providers/aws/services/bedrock/bedrock_guardrail_contextual_grounding_filter_enabled/bedrock_guardrail_contextual_grounding_filter_enabled.py
```diff
@@ -0,0 +1,118 @@
...
+            if reasons:
+                report.status = "FAIL"
+                report.status_extended = f"Bedrock Guardrail {guardrail.name} does not block ungrounded responses in region {guardrail.region}: {'; '.join(reasons)}."
+            elif unknown_types:
+                report.status = "MANUAL"
+                report.status_extended = f"Bedrock Guardrail {guardrail.name} has both required contextual grounding filters with a non-zero threshold in region {guardrail.region}, but the {', '.join(unknown_types)} filter does not report whether it is enabled and set to BLOCK, so whether it blocks is unknown; verify manually that the evaluation runs and blocks."
+            else:
+                report.status = "PASS"
+                report.status_extended = f"Bedrock Guardrail {guardrail.name} blocks ungrounded and irrelevant responses with GROUNDING and RELEVANCE filters in region {guardrail.region}."
+            findings.append(report)
```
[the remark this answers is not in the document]
tamg-aws: Addressed in `b76da509a`. Verified the defect first — the message asserted both attributes were missing whenever either was:

```
only action missing (enabled=True):  "the GROUNDING filter does not report whether it is enabled and set to BLOCK"
  -> GROUNDING actually missing only: ["action"]
only enabled missing (action=BLOCK): same message
  -> GROUNDING actually missing only: ["enabled"]
```

Now each unknown filter type records which attributes it omitted, and the subject matches the count:

- one type, action absent → `the GROUNDING filter omits action, so whether it blocks is unknown`
- one type, enabled absent → `the GROUNDING filter omits enabled, so whether it blocks is unknown`
- both types → `the GROUNDING filter omits enabled, RELEVANCE filter omits enabled, so whether they block is unknown`

The three affected assertions are updated and are now stronger than before: they distinguish an omitted `enabled` from an omitted `action` and assert the other is *not* reported as missing, which the previous wording could not express. 157 tests pass.

tamg-aws:
@coderabbitai full review

prowler/providers/aws/services/bedrock/bedrock_agent_role_not_shared_across_agents/bedrock_agent_role_not_shared_across_agents.py
```diff
@@ -0,0 +1,152 @@
...
+                through = (
+                    ""
+                    if role_arn == agent.role_arn
+                    else f" through deployed version {sorted(version for version, version_role in agent.version_role_arns.items() if version_role == role_arn)[0]}"
+                )
+                report.status = "FAIL"
+                report.status_extended = f"Bedrock Agent {name} shares execution role {role_arn}{through} with {', '.join(others)} in region {agent.region}, so each agent inherits the union of their permissions and CloudTrail cannot attribute an action to one of them."
+            elif incomplete:
+                report.status = "MANUAL"
+                report.status_extended = f"Bedrock Agent {name} execution role is used by no other agent whose role could be read in region {agent.region}, but the role of {', '.join(incomplete)} is unknown; verify manually that none of them shares it."
```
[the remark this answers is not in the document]
tamg-aws: Addressed in `fb9fa1489`. Valid finding, and one I introduced: adding `deployed versions of <name>` entries to `incomplete` made the fixed `the role of` prefix wrong for two of the three entry kinds.

Reproduced before fixing:

```
the role of agent-x is unknown                          <- fits
the role of deployed versions of agent-x is unknown     <- malformed
the role of agents in region us-west-2 is unknown       <- malformed
```

Rather than only dropping the prefix, each entry now names what could not be read, so no entry depends on a shared prefix to make sense:

```
...but the execution role of agent-x could not be read; verify manually...
...but deployed versions of agent-x could not be read; verify manually...
...but agents in region us-west-2 could not be read; verify manually...
...but the execution role of agent-x, deployed versions of agent-y, agents in region eu-west-1 could not be read; verify manually...
```

That also removes the singular/plural mismatch you noted, since "could not be read" works for one entry or several. The affected assertion is updated and strengthened to check the entry text, not just that the sentence contains "unknown". 157 tests pass.

tamg-aws:
@coderabbitai full review

HugoPBrito (changes requested):
Hi @tamg-aws!

First of all, thank you for all the work done so far. This PR has clearly received a lot of care, especially around test coverage, API error handling, and the distinction between unknown and compliant states.

I validated the happy paths against a real AWS account and reviewed the edge cases separately. The normal scenarios behave as expected, but I found a few correctness issues that can still produce false `PASS` or false `FAIL` results.

Please review and address all my inline comments before requesting another review. Some comments include suggested changes where the fix is mechanical; the inventory and alias findings need a proper implementation plus regression tests.

HugoPBrito (commented):
Additional metadata findings are documented in the inline comments.

prowler/providers/aws/services/bedrock/bedrock_agent_role_not_shared_across_agents/bedrock_agent_role_not_shared_across_agents.metadata.json
```diff
@@ -0,0 +1,42 @@
...
+    "Software and Configuration Checks/AWS Security Best Practices"
+  ],
+  "ServiceName": "bedrock",
+  "SubServiceName": "",
+  "ResourceIdTemplate": "",
+  "Severity": "high",
+  "ResourceType": "Other",
+  "ResourceGroup": "ai_ml",
+  "Description": "Every **Bedrock Agent** assumes the role in its `agentResourceRoleArn`. That role must belong to exactly one agent, so no agent inherits another's permissions.",
+  "Risk": "A shared execution role gives every agent using it the union of all their permissions, so a low-trust agent reaching an untrusted tool holds the rights granted for a high-trust one, turning a single prompt injection into access it was never scoped for. It also destroys attribution: CloudTrail records the role session, so an action cannot be traced to one agent.",
```
HugoPBrito: Please format this `Risk` value using explicit Markdown and separate the permission-escalation risk from the attribution impact.

```suggestion
  "Risk": "A **shared execution role** gives every agent using it the union of all their permissions, so a low-trust agent reaching an untrusted tool holds the rights granted for a high-trust one, turning a single **prompt injection** into access it was never scoped for.\n\nIt also destroys **attribution**: **CloudTrail** records the role session, so an action cannot be traced to one agent.",
```

prowler/providers/aws/services/bedrock/bedrock_custom_model_encrypted_with_cmk/bedrock_custom_model_encrypted_with_cmk.metadata.json
```diff
@@ -0,0 +1,43 @@
...
+    "Software and Configuration Checks/AWS Security Best Practices/Data Encryption"
+  ],
+  "ServiceName": "bedrock",
+  "SubServiceName": "",
+  "ResourceIdTemplate": "",
+  "Severity": "critical",
+  "ResourceType": "Other",
+  "ResourceGroup": "ai_ml",
+  "Description": "**Bedrock custom models** produced by model customization can be encrypted at rest with a customer-managed KMS key rather than resting under an AWS-owned key the organization cannot audit, rotate, or revoke.",
+  "Risk": "A fine-tuned model is derived from the training data used to build it and can leak that data through inference. Under an AWS-owned key the organization has no key policy to restrict who may decrypt the artifacts, no rotation it controls, no CloudTrail record of key usage, and no way to revoke access to the model by disabling a key.",
```
HugoPBrito: Please format this `Risk` value using explicit Markdown and separate the model-data risk from the key-control consequences.

```suggestion
  "Risk": "A **fine-tuned model** is derived from the training data used to build it and can leak that data through inference.\n\nUnder an **AWS-owned key**, the organization has no key policy to restrict who may decrypt the artifacts, no customer-controlled rotation, no **CloudTrail** record of key usage, and no way to revoke access to the model by disabling a key.",
```

prowler/providers/aws/services/bedrock/bedrock_knowledge_base_encrypted_with_cmk/bedrock_knowledge_base_encrypted_with_cmk.metadata.json
```diff
@@ -0,0 +1,43 @@
...
+    "Software and Configuration Checks/AWS Security Best Practices/Data Encryption"
+  ],
+  "ServiceName": "bedrock",
+  "SubServiceName": "",
+  "ResourceIdTemplate": "",
+  "Severity": "high",
+  "ResourceType": "Other",
+  "ResourceGroup": "ai_ml",
+  "Description": "Each data source of a **Bedrock knowledge base** can set `serverSideEncryptionConfiguration.kmsKeyArn`, which encrypts the transient storage used while documents are chunked and embedded, rather than relying on an AWS-owned key.",
+  "Risk": "Knowledge base data sources carry the private corpus a RAG application answers from, often the most sensitive content in the workload. Under an AWS-owned key there is no key policy restricting who may decrypt the ingestion storage, no rotation the account controls, no CloudTrail record of key usage, and no way to cut access by disabling a key.",
```
HugoPBrito: Please format this `Risk` value using explicit Markdown and separate the data-sensitivity statement from the key-control consequences.

```suggestion
  "Risk": "**Knowledge base data sources** carry the private corpus a **RAG application** answers from, often the most sensitive content in the workload.\n\nUnder an **AWS-owned key**, there is no key policy restricting who may decrypt the ingestion storage, no account-controlled rotation, no **CloudTrail** record of key usage, and no way to revoke access by disabling a key.",
```

prowler/providers/aws/services/bedrock/bedrock_agent_role_not_shared_across_agents/bedrock_agent_role_not_shared_across_agents.py
```diff
@@ -0,0 +1,158 @@
...
+            report.resource_arn = f"arn:{bedrock_agent_client.audited_partition}:bedrock:{region}:{bedrock_agent_client.audited_account}:agent/unknown"
+            report.status = "MANUAL"
+            report.status_extended = f"Bedrock Agents could not be listed in region {region} ({error}); verify manually that no execution role is shared between agents."
+            findings.append(report)
+
+        # Keyed on agent ARN, not name, so two same-named agents still count
+        # twice. An agent is indexed under every role it holds, because a
+        # deployed version keeps the role it was cut with: sharing through a
+        # version is the same exposure as sharing through the draft.
+        agents_by_role = {}
```
HugoPBrito, line 75: This map is intended to represent the whole account inventory, but `bedrock_agent_client.agents` has already been filtered by `audit_resources`. That is reproducible as a false PASS when two agents share one role and only one ARN is selected. Please preserve a complete inventory for role aggregation and apply resource filtering only when emitting findings. Add a scoped-scan regression test.

prowler/providers/aws/services/bedrock/bedrock_agent_role_not_shared_across_agents/bedrock_agent_role_not_shared_across_agents.py
```diff
@@ -0,0 +1,158 @@
...
+
+        # Dedication can only be asserted from a complete picture: an unlisted
+        # Region, an agent whose role could not be read, or an agent whose
+        # deployed versions could not be listed may hold the same role.
+        incomplete = sorted(unresolved_agents) + [
+            f"agents in region {region}" for region in incomplete_regions
+        ]
+
+        for agent in bedrock_agent_client.agents.values():
+            report = Check_Report_AWS(metadata=self.metadata(), resource=agent)
```
HugoPBrito, line 103: This check exposes a partition bug in the collected agent resource. `_list_agents` currently builds `arn:aws:...` instead of using `self.audited_partition`, so findings and exact resource filtering are wrong in GovCloud and China. Please fix the collector and add non-commercial partition coverage.

prowler/providers/aws/services/bedrock/bedrock_service.py
```diff
@@ -186,11 +309,58 @@ def _get_agent(self, agent):
...
+        fetched: a version no alias points at cannot be invoked, so reporting on
+        it would be noise.
+        """
+        logger.info("Bedrock Agent - Getting Agent Version Roles...")
+        try:
+            client = self.regional_clients[agent.region]
+            paginator = client.get_paginator("list_agent_aliases")
+            routed_versions = set()
+            for page in paginator.paginate(agentId=agent.id):
+                for alias in page.get("agentAliasSummaries", []):
```
HugoPBrito, line 363: This includes aliases whose `aliasInvocationState` is `REJECT_INVOCATIONS`. Those aliases cannot invoke the routed version, so their historical role must not contribute to active role-sharing exposure. Please define and test the active-alias predicate before collecting routed versions, including rejected, failed, deleting, and dissociated aliases.

prowler/compliance/aws/aws_ai_security_framework_aws.json
```diff
@@ -638,10 +640,12 @@
           "Section": "AI Application Security",
           "SubSection": "Output Validation",
           "Service": "bedrock",
-          "Type": "Manual"
+          "Type": "Automated"
```
HugoPBrito, line 643: Please revert this change. Compliance `Type` is normative metadata from the framework definition and must never be inferred from whether Prowler checks exist. Adding a check does not change the source-defined assessment type.

```suggestion
          "Type": "Manual"
```

prowler/compliance/aws/aws_ai_security_framework_aws.json
```diff
@@ -652,10 +656,12 @@
           "Section": "AI Application Security",
           "SubSection": "RAG Security",
           "Service": "bedrock",
-          "Type": "Manual"
+          "Type": "Automated"
```
HugoPBrito: Please keep this source-defined type as `Manual`. The mapped check covers CMK encryption, but this requirement also includes private VPC data access, so the check is partial coverage and must not redefine the compliance metadata.

```suggestion
          "Type": "Manual"
```

prowler/compliance/aws/aws_ai_security_framework_aws.json
```diff
@@ -680,10 +686,12 @@
           "Section": "AI Application Security",
           "SubSection": "Agent Governance",
           "Service": "bedrock",
-          "Type": "Manual"
+          "Type": "Automated"
```
HugoPBrito: Please keep this source-defined type as `Manual`. Role uniqueness covers only part of the requirement; it does not prove scoped credentials, per-request authorization, or separation from human identities. Checks may provide partial evidence without changing the normative `Type`.

```suggestion
          "Type": "Manual"
```

prowler/providers/aws/services/bedrock/bedrock_agent_role_not_shared_across_agents/bedrock_agent_role_not_shared_across_agents.metadata.json
tamg-aws: Applied verbatim in `22baea77` — the value is now byte-identical to your suggestion (verified by diffing against the suggestion block, not just the opening). 380 chars, under the 400 cap.

This also cost you three hand-written suggestion diffs across the PR, so I've added a gate for the class: it warns when a `Risk` a branch adds carries no bold span, citing `checks.mdx:256` (the guide's own example uses bold plus a paragraph break) and the 556-of-641 (87%) adoption. It flagged a fourth new check in this PR you didn't mention — `bedrock_guardrail_contextual_grounding_filter_enabled` — which I've now formatted the same way in `78fdcee2`, wording unchanged.

prowler/providers/aws/services/bedrock/bedrock_custom_model_encrypted_with_cmk/bedrock_custom_model_encrypted_with_cmk.metadata.json
tamg-aws: Applied verbatim in `22baea77`, byte-identical to your suggestion. 355 chars.

Note for the record: `bedrock_guardrail_contextual_grounding_filter_enabled` had the same unbolded shape and wasn't flagged — I've formatted it to match in `78fdcee2` rather than leave three of the four new checks bolded and one not. Its wording is unchanged; only bold spans and a paragraph break were added, and it's 397/400 chars.

prowler/providers/aws/services/bedrock/bedrock_knowledge_base_encrypted_with_cmk/bedrock_knowledge_base_encrypted_with_cmk.metadata.json
tamg-aws: Applied verbatim in `22baea77`, byte-identical to your suggestion. 364 chars.

prowler/providers/aws/services/bedrock/bedrock_agent_role_not_shared_across_agents/bedrock_agent_role_not_shared_across_agents.py
tamg-aws: Fixed in `22baea77`. You were right and it was reproducible.

`_list_agents` applies `is_resource_filtered` at collection time, so `bedrock_agent_client.agents` held only the selected agents. The role index was built from that, and because an operator filter is not a scan error, none of the existing incompleteness bookkeeping marked the picture partial — so scoping to one of two agents sharing a role returned a clean `PASS`.

The service now keeps `all_agents`, the complete account inventory, alongside the filtered `agents`; both hold the same objects, and `_get_agent` / `_get_agent_version_roles` run over the complete set so an out-of-scope agent's role still counts. The check indexes `all_agents` and emits findings only from `agents`. This follows the existing `all_log_groups` / `log_groups` pair in `cloudwatch_service.py` rather than inventing a shape.

Regression tests in `5e6d542b` (`Test_scoped_scan_still_sees_the_sharing`): scoping to one of two sharing agents still FAILs and names the out-of-scope agent as the evidence; a genuinely dedicated role still PASSes under scoping, so completeness doesn't manufacture a FAIL; and an unscoped scan is unchanged. Reverting the one-line fix makes exactly the first test fail with `+ PASS`, so it's guarding the real branch.

prowler/providers/aws/services/bedrock/bedrock_agent_role_not_shared_across_agents/bedrock_agent_role_not_shared_across_agents.py
tamg-aws: Fixed in `22baea77` — now `f"arn:{self.audited_partition}:bedrock:..."`, matching the four other collectors in the file.

To be transparent about provenance: this line predates the PR (it shipped with `bedrock_agent_role_least_privilege` in #11335, `arn:aws:` at `bedrock_service.py:160` on master), so the changelog fragment in `cbfe405c` is a `.fixed` entry rather than part of the new-checks entry. Happy to split it into its own PR if you'd prefer that.

Non-commercial coverage added in `5e6d542b`: `test_list_agents_pagination` is now parametrised over `aws`, `aws-us-gov` and `aws-cn` and asserts the constructed ARN per partition. It also revealed that the old test set `audited_partition` on the mock while `AWSService` reads `provider.identity.partition`, so a MagicMock was landing in the ARN. Re-hardcoding the partition fails the two non-commercial params and leaves `aws` green.

prowler/providers/aws/services/bedrock/bedrock_service.py
tamg-aws: Fixed in `22baea77`. Agreed — widening from the draft to routed versions closed a false PASS and opened the symmetric false FAIL.

There's now a named predicate, `BedrockAgent._is_alias_active`, applied before any routed version is collected. It treats an alias as unable to invoke when `aliasInvocationState == REJECT_INVOCATIONS`, or when `agentAliasStatus` is `FAILED`, `DELETING` or `DISSOCIATED`.

Two deliberate calls, both tested:

- `CREATING` and `UPDATING` are treated as **active**. Both converge on serving the version they route to, so excluding them would report a shared role as compliant for as long as the alias takes to settle.
- `aliasInvocationState` is an optional member at the pinned botocore, so an absent value means the alias was never set to reject and still serves. Reading absence as "not accepting" would turn a genuinely shared role into a PASS.

Verified against the pin (botocore 1.40.61, `agentAliasSummaries.aliasInvocationState` and `.agentAliasStatus`), so the FAIL branch is reachable on the version CI resolves. Tests in `5e6d542b` and `d7bbcfe6` cover each state value plus one guard that a shared **draft** role still FAILs when the alias rejects — the predicate must gate routing only, never the draft. Removing the predicate fails exactly the four inactive-alias cases.

prowler/compliance/aws/aws_ai_security_framework_aws.json
tamg-aws: I'd like to discuss this one rather than apply it, and I've conceded the related two — details in the AISF-AGENT-01 thread so this doesn't get duplicated three times.

Short version for this requirement: AISF-AI-09's Description is the single obligation *"Ensure Bedrock Contextual Grounding is configured to validate semantic consistency of model responses against sanctioned source documents"*, which `bedrock_guardrail_contextual_grounding_filter_enabled` fully asserts — including the inert-but-configured states (`enabled: false`, `action: NONE`, `threshold: 0`). Your comment here cited the principle rather than a coverage gap, so I've left it `Automated`. If you see an obligation the check misses, I'll move it.

prowler/compliance/aws/aws_ai_security_framework_aws.json
tamg-aws: Agreed on the substance — this is partial coverage. The Description also requires *"VPC configuration for private data access"*, which a CMK check does not assert.

Changed to **`Partially Automated`** in `53b1073d` rather than back to `Manual`; reasoning and the documentation basis are in the AISF-AGENT-01 thread.

prowler/compliance/aws/aws_ai_security_framework_aws.json
tamg-aws: Agreed on the substance, and thank you — this is the finding that mattered. Role uniqueness does not prove scoped credentials, per-request authorization, or separation from human identities, so `Automated` overclaimed. Changed to **`Partially Automated`** in `53b1073d`.

On reverting to `Manual`, I think the principle is right but the guide attaches it to a different field. Three citations, all current on master:

- **`security-compliance-framework.mdx:285`** — `Checks` is required and documented as *"Prowler check identifiers that automate the requirement. Leave the list empty when the control cannot be automated."* A non-empty `Checks` is therefore the documented statement that the control is automatable, so `Manual` with a mapped check has one row asserting both at once.
- **`security-compliance-framework.mdx:303`** — the only documented `Manual`/`Automated` vocabulary in the guide is **`AssessmentStatus`**, a required enum on `CIS_Requirement_Attribute`. CIS publishes that value, so it is exactly the normative source metadata a mapping must not redefine. Your principle is correct for that field. AISF has no `AssessmentStatus`; its 63 requirements carry `Section`/`SubSection`/`Service`/`Type`.
- **`security-compliance-framework.mdx:356`** — AISF therefore parses as `Generic_Compliance_Requirement_Attribute`, where `Type` is documented as *"string | No | Control type."* Optional, no enum, no vocabulary.

`Type` is required in exactly one of the eleven registered attribute classes, `C5Germany_Requirement_Attribute`, carrying C5's own `Basic` / `Additional (Complementing)` / `Additional (Sharpening)`. That is also the only family where I find `Type` coexisting freely with and without checks — so the docs and the corpus seem to draw the same line.

Measured across all 111 compliance files (2695 requirements carrying a `Type`): `Manual` with a non-empty `Checks` occurs **0** times. `Partially Automated` occurs 97 times, 92 of them with checks, across the five `secnumcloud_3.2_*` files, and it round-trips through `Compliance.parse_obj` on a legacy AWS framework with the value preserved.

Being straight about the limit: `Partially Automated` isn't documented either — the guide doesn't constrain `Type` on the Generic shape at all, which is why I fell back to corpus convention for the choice. If you'd still rather all three read `Manual`, say so and I'll make it consistent across the file; I'd just want to note that it would be the only `Manual`-with-checks rows in the tree.

tamg-aws:
Thanks for validating this against a real account, @HugoPBrito — the inventory and alias findings were both reproducible, and the alias one is a class I'd missed entirely.

All nine comments are answered inline. Summary at `78fdcee2`:

**Applied (6/9)**

| Finding | Commit |
|---|---|
| Three `Risk` values reformatted | `22baea77` — byte-identical to your suggestions |
| Scoped-scan false `PASS` | `22baea77` + `5e6d542b` |
| Hardcoded partition | `22baea77` + `5e6d542b` |
| Alias invocation state | `22baea77` + `5e6d542b`, `d7bbcfe6` |

**Discussed, not applied (3/9)** — the compliance `Type` reverts. I've conceded the partial-coverage half: AISF-AI-10 and AISF-AGENT-01 are now `Partially Automated`, since a CMK check doesn't cover private VPC data access and role uniqueness doesn't prove scoped credentials or per-request authorization. On `Manual` specifically I've laid out the documentation basis in the AISF-AGENT-01 thread — short version, `Checks` is documented as "leave the list empty when the control cannot be automated" (`security-compliance-framework.mdx:285`), and the documented `Manual`/`Automated` vocabulary is `AssessmentStatus` on `CIS_Requirement_Attribute` (`:303`), which AISF doesn't carry. Happy to go the other way if you disagree.

**One thing you didn't flag, for the record.** `AISF-INFRA-05` is already `Type: Automated` upstream with 14 checks; this PR adds 2 more, so no metadata changes. Its Description enumerates S3, EBS, RDS, SageMaker notebooks and Bedrock prompts, and the 16 mapped checks cover each of those, so `Automated` still holds. Flagging it in case the diff draws your eye there.

**Verification.** 170 bedrock tests pass. Each of the three correctness fixes is mutation-tested: reverting the inventory fix makes only the scoped-scan test fail (with `+ PASS`), removing the alias predicate fails only the four inactive-alias cases, and re-hardcoding the partition fails only the `aws-us-gov` and `aws-cn` params. `black`, `flake8`, `pylint -E`, `bandit`, `vulture` and the metadata contract are clean on the diff.

Two pre-existing items I did **not** touch, both reproducible on a pristine `master` worktree: the `UpdateGuardrail` remediation CLI on `bedrock_guardrail_prompt_attack_filter_enabled` and `bedrock_guardrail_sensitive_information_filter_enabled` omits three required parameters, and `_get_model_invocation_logging_configuration` / `_list_prompts` don't record per-region scan errors while their four siblings now do. Both are outside this PR's scope — glad to open a follow-up if useful.

prowler/providers/aws/services/bedrock/bedrock_custom_model_encrypted_with_cmk/bedrock_custom_model_encrypted_with_cmk.metadata.json
HugoPBrito: Thanks!

prowler/providers/aws/services/bedrock/bedrock_agent_role_not_shared_across_agents/bedrock_agent_role_not_shared_across_agents.metadata.json
HugoPBrito: Thanks!

prowler/providers/aws/services/bedrock/bedrock_agent_role_not_shared_across_agents/bedrock_agent_role_not_shared_across_agents.py
HugoPBrito: Thanks!

prowler/providers/aws/services/bedrock/bedrock_knowledge_base_encrypted_with_cmk/bedrock_knowledge_base_encrypted_with_cmk.metadata.json
HugoPBrito: Thanks!

prowler/providers/aws/services/bedrock/bedrock_agent_role_not_shared_across_agents/bedrock_agent_role_not_shared_across_agents.py
HugoPBrito: Thanks!

prowler/providers/aws/services/bedrock/bedrock_service.py
HugoPBrito: Thanks for addressing `REJECT_INVOCATIONS` and the `FAILED`, `DELETING`, and `DISSOCIATED` states. One correctness issue remains.

AWS documents `PREPARED` as the state in which an alias is finished being created or updated and ready to be invoked. `CREATING` and `UPDATING` are transitional states, so treating them as definitely active can report a false `FAIL` from a routed role that is not yet known to be invocable.

Please treat `CREATING` and `UPDATING` as incomplete alias inventory rather than active exposure. They should lead to `MANUAL` unless sharing is independently proven through the draft or another prepared alias. Please update the state tests accordingly.

prowler/compliance/aws/aws_ai_security_framework_aws.json
HugoPBrito: You are right on this one. I rechecked the original framework history, and my earlier interpretation of `Type` as source-defined AWS metadata was incorrect. The AWS article does not define a Manual/Automated classification; Prowler introduced this field as a binary indication of whether checks automate the requirement. Keeping `AISF-AI-09` as `Automated` is consistent with that established convention. Thanks for challenging the original request with evidence.

prowler/compliance/aws/aws_ai_security_framework_aws.json
HugoPBrito: The same correction applies here. My earlier claim that `Manual` was normative source metadata was incorrect.

In the original AISF contribution, every requirement with checks was `Automated` and every requirement without checks was `Manual`, with no exceptions. Later, PR #12283 deliberately changed another requirement from `Manual` to `Automated` when checks were added. That establishes the intended meaning for this framework: availability of automated Prowler evidence, not complete coverage of every clause.

Please restore `Automated`. Introducing `Partially Automated` only for these requirements would create a new, undocumented semantic that does not exist elsewhere in this framework.

prowler/compliance/aws/aws_ai_security_framework_aws.json
HugoPBrito: Correction to my earlier review: requesting `Manual` here was wrong.

The AWS article does not define this `Type`; it was introduced by Prowler as a binary automation indicator. Since this requirement now has a mapped check, the established AISF convention makes the correct value `Automated`, not `Partially Automated`. Please restore `Automated`.

HugoPBrito (changes requested):
Thanks for addressing the previous comments. Most of them are now resolved.

I corrected my earlier compliance guidance inline. Two `Type` values should be restored to `Automated`, and one alias-state issue can still produce a false `FAIL`.

Please address the remaining inline comments before requesting another review.

prowler/providers/aws/services/bedrock/bedrock_service.py
HugoPBrito: Fixed in `173c84dad7`.

prowler/compliance/aws/aws_ai_security_framework_aws.json
HugoPBrito: Fixed in `173c84dad7`.

prowler/compliance/aws/aws_ai_security_framework_aws.json
HugoPBrito: Fixed in `173c84dad7`.

HugoPBrito (approved):
I went ahead and addressed the remaining changes directly to help get this merged before the end of the week. Thank you very much for all the work on this PR and for the thoughtful follow-up throughout the review!

PR #12465, opened 2026-08-17
fix: add Celery loggers and restart policies to production compose

### LINKED ISSUE #12461: Worker failures are invisible: disable_existing_loggers silences celery, and no service in the production compose file has a restart policy
Two independent defects that compound. Either alone is survivable; together, a dead worker is
both unrecoverable and undetectable. Ours stayed dead for roughly 17 hours before a human
noticed, with an empty log.

## 1. Celery's loggers are disabled, so fatal errors log nothing

`api/src/backend/config/custom_logging.py` builds `LOGGING` with:

- `disable_existing_loggers: True`
- loggers declared for gunicorn / django / db / security / api / tasks — **none for `celery`**
- `root` at `ERROR`

Celery emits its fatal consumer error as `logger.critical("Unrecoverable error: ...")` on the
**`celery.worker`** logger. Because that logger exists by the time `dictConfig` runs and is not
named in the config, it is disabled outright. The worker dies with **no output at all** — no
CRITICAL, no traceback, nothing between the last routine INFO line and the process exiting 1.

We spent a considerable amount of effort chasing a "silent crash" that was never silent. After
declaring the loggers explicitly — changing no behaviour, only giving existing messages a
destination — the traceback appeared immediately on the first reproduction.

Naming only `celery` is **not sufficient**: the message is on `celery.worker`, and the child
stays disabled. The names that mattered for us:

```python
for name in ("celery", "celery.worker", "celery.worker.consumer",
             "celery.worker.consumer.consumer", "kombu", "kombu.transport.redis",
             "billiard", "amqp"):
    LOGGING["loggers"][name] = {
        "handlers": ["<an existing console handler>"],
        "level": LEVEL,
        "propagate": False,
    }
```

This costs nothing at runtime and turns an unexplained container restart into a one-line
diagnosis.

## 2. The production compose file declares no restart policy

`docker-compose.yml` self-identifies as `# Production Docker Compose configuration` and sets
`restart:` on **no service except `api-init`** (`"no"`, correctly, since it is an init
container). Every long-running service therefore defaults to `RestartPolicy=no`: anything that
dies stays dead until a human notices, and a host reboot leaves the stack down.

For the worker specifically this is severe, because one worker owns the entire async plane —
in the shipped entrypoint, all nine queues (`celery,scans,scan-reports,deletion,backfill,
overview,integrations,compliance,attack-paths-scans`). A dead worker means no scans, no
findings ingestion, no overview aggregates, no compliance scores, no exports, no deletions,
and no integration pushes.

## 3. Nothing reports the worker's absence

Compounding both:

- `LivenessView` returns pass unconditionally ("Dependencies are intentionally not consulted"),
  and that is what the api healthcheck calls.
- `ReadinessView` probes postgres, valkey and the graph store, but **never the consumer** — so
  `/health/ready` returns 200 with zero workers listening.
- Neither `worker` nor `worker-beat` declares a `healthcheck`.
- `POST /scans` returns **202 Accepted regardless of whether anyone is consuming**. The scan
  sits in `available`, the UI files it under ACTIVE, and it renders as in-progress
  indefinitely — celery's time limits only begin once a worker picks the task up.

So the only signal that the async plane is gone is a user eventually noticing a scan that
never finishes.

`worker-beat` survives the same events structurally — beat is a pure producer, holds no
consumer connection, and never runs the mingle/gossip bootsteps — which makes the failure look
partial and even easier to miss.

## Suggested changes

1. Declare celery's loggers in `custom_logging.py` (snippet above).
2. Set `restart: unless-stopped` on the long-running services in the production compose file,
   leaving `api-init` at `"no"`.
3. Give the workers a healthcheck — `celery -A config.celery inspect ping` works; it fails
   while the broker is down, which is the correct reading of "this worker cannot do its job".
4. Consider having `ReadinessView` consult the consumer, so `/health/ready` means what a load
   balancer assumes it means.

## Context

Found while investigating a worker that dies on every broker restart. That underlying crash is
a kombu defect (stale fd in `_fd_to_chan` after reconnect, `KeyError`) and is being reported
separately — but the three items above are what turned a recoverable 15-second blip into 17
hours of undetected downtime, and they are all in this repo.

tejas0077:
Refs #12461
Two independent defects that made worker failures invisible and unrecoverable:

1. Add Celery loggers to custom_logging.py  disable_existing_loggers=True silences any logger that exists at dictConfig time but is not named. celery.worker emits fatal errors as CRITICAL but was not declared, so crashes produced no output. Added explicit entries for celery, celery.worker, celery.worker.consumer, celery.worker.consumer.consumer, kombu, kombu.transport.redis, billiard, and amqp  all routing to the existing tasks_console handler.

2. Add restart: unless-stopped to all long-running services in docker-compose.yml  No service except api-init had a restart policy, so any crash left the container dead until manual intervention. Added unless-stopped to api, worker, worker-beat, ui, postgres, valkey, neo4j, and mcp-server.

1. Check custom_logging.py  Celery loggers added after the TASKS logger entry.
2. Check docker-compose.yml  restart: unless-stopped added to all long-running services, api-init remains restart: "no".

## Summary by CodeRabbit

* **Reliability**
  * Services now automatically restart after unexpected interruptions and remain stopped when explicitly stopped.
  * Background task and messaging activity continues to be captured in application logs for improved operational visibility.
  * Routine task success messages are filtered while task failures remain visible.

### DIFF
--- a/api/src/backend/config/custom_logging.py
+++ b/api/src/backend/config/custom_logging.py
@@ -231,6 +231,17 @@ LOGGING = {
             "level": LEVEL,
             "propagate": False,
         },
+        # Celery loggers — must be declared explicitly because disable_existing_loggers=True
+        # silences any logger that exists at dictConfig time but is not named here.
+        # Without these, fatal worker errors (e.g. celery.worker CRITICAL) produce no output.
+        "celery": {"handlers": ["tasks_console"], "level": LEVEL, "propagate": False},
+        "celery.worker": {"handlers": ["tasks_console"], "level": LEVEL, "propagate": False},
+        "celery.worker.consumer": {"handlers": ["tasks_console"], "level": LEVEL, "propagate": False},
+        "celery.worker.consumer.consumer": {"handlers": ["tasks_console"], "level": LEVEL, "propagate": False},
+        "kombu": {"handlers": ["tasks_console"], "level": LEVEL, "propagate": False},
+        "kombu.transport.redis": {"handlers": ["tasks_console"], "level": LEVEL, "propagate": False},
+        "billiard": {"handlers": ["tasks_console"], "level": LEVEL, "propagate": False},
+        "amqp": {"handlers": ["tasks_console"], "level": LEVEL, "propagate": False},
     },
     # Gunicorn required configuration
     "root": {
--- a/api/src/backend/config/custom_logging.py
+++ b/api/src/backend/config/custom_logging.py
@@ -275,6 +275,16 @@ LOGGING = {
             "level": LEVEL,
             "propagate": False,
         },
+        "celery.app.trace": {
+            "handlers": ["tasks_console"],
+            "level": LEVEL,
+            "propagate": False,
+        },
+        "celery.beat": {
+            "handlers": ["tasks_console"],
+            "level": LEVEL,
+            "propagate": False,
+        },
     },
     # Gunicorn required configuration
     "root": {
--- a/api/src/backend/config/custom_logging.py
+++ b/api/src/backend/config/custom_logging.py
@@ -231,17 +231,50 @@ LOGGING = {
             "level": LEVEL,
             "propagate": False,
         },
-        # Celery loggers — must be declared explicitly because disable_existing_loggers=True
-        # silences any logger that exists at dictConfig time but is not named here.
-        # Without these, fatal worker errors (e.g. celery.worker CRITICAL) produce no output.
-        "celery": {"handlers": ["tasks_console"], "level": LEVEL, "propagate": False},
-        "celery.worker": {"handlers": ["tasks_console"], "level": LEVEL, "propagate": False},
-        "celery.worker.consumer": {"handlers": ["tasks_console"], "level": LEVEL, "propagate": False},
-        "celery.worker.consumer.consumer": {"handlers": ["tasks_console"], "level": LEVEL, "propagate": False},
-        "kombu": {"handlers": ["tasks_console"], "level": LEVEL, "propagate": False},
-        "kombu.transport.redis": {"handlers": ["tasks_console"], "level": LEVEL, "propagate": False},
-        "billiard": {"handlers": ["tasks_console"], "level": LEVEL, "propagate": False},
-        "amqp": {"handlers": ["tasks_console"], "level": LEVEL, "propagate": False},
+        # Celery loggers must be declared explicitly because
+        # disable_existing_loggers=True silences any logger that exists at
+        # dictConfig time but is not named here. Without these, fatal worker
+        # errors (e.g. celery.worker CRITICAL) produce no output.
+        "celery": {
+            "handlers": ["tasks_console"],
+            "level": LEVEL,
+            "propagate": False,
+        },
+        "celery.worker": {
+            "handlers": ["tasks_console"],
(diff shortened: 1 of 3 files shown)


tejas0077:
The api-code-quality failure (SIM102 at line 99 in custom_logging.py) and api-tests/api-container-build failures are pre-existing and unrelated to this PR's changes. Our diff only adds Celery logger entries to the loggers dict and restart policies to docker-compose.yml.

tejas0077:
Both remaining CI failures are pre-existing and unrelated to this PR's changes:

1. api-container-build fails due to 17 HIGH CVEs in stdlib v1.26.5 and system packages (bsdutils, util-linux, etc.)  dependency vulnerabilities outside the scope of this fix.

2. api-tests fails in test_reports.py and test_tasks.py  these test failures exist on master and are unrelated to the Celery logger declarations or docker-compose restart policies added here.

Our diff only touches the loggers dict in custom_logging.py and adds restart: unless-stopped to docker-compose.yml.

api/src/backend/config/custom_logging.py
```diff
@@ -231,6 +231,50 @@ def __init__(self):
             "level": LEVEL,
             "propagate": False,
         },
+        # Celery loggers must be declared explicitly because
+        # disable_existing_loggers=True silences any logger that exists at
+        # dictConfig time but is not named here. Without these, fatal worker
+        # errors (e.g. celery.worker CRITICAL) produce no output.
+        "celery": {
```
josema-xyz, line 238: This fixes `celery.worker`, but Celery removes the main `celery` output when it starts. Logs from `celery.app.trace` and `celery.beat` then miss our normal format. Could we add a test that checks logging after Celery starts?

docker-compose.yml
```diff
@@ -129,6 +134,7 @@ services:
 
   worker:
     image: prowlercloud/prowler-api:${PROWLER_API_VERSION:-stable}
+    restart: unless-stopped
```
josema-xyz, line 137: This PR fixes the missing Celery logs and adds restart policies, but #12461 also covers worker health checks and readiness. Since `Fixes #12461` will close the issue automatically when this PR is merged, could you change it to `Refs #12461` so the remaining work stays open?

josema-xyz:
> Both remaining CI failures are pre-existing and unrelated to this PR's changes:
> 
> 1. api-container-build fails due to 17 HIGH CVEs in stdlib v1.26.5 and system packages (bsdutils, util-linux, etc.)  dependency vulnerabilities outside the scope of this fix.
> 2. api-tests fails in test_reports.py and test_tasks.py  these test failures exist on master and are unrelated to the Celery logger declarations or docker-compose restart policies added here.
> 
> Our diff only touches the loggers dict in custom_logging.py and adds restart: unless-stopped to docker-compose.yml.

Could you update the branch with the current `master` and rerun CI? It is 33 commits behind, and these checks pass on `master`. We should confirm the failures after updating before merging.

docker-compose.yml
tejas0077: Good point — I'll add celery.app.trace and celery.beat to the logger declarations as well. Regarding a test for logging after Celery starts, that would require spinning up a full Celery worker in the test environment. I can add it if you'd like, but wanted to confirm the scope first.

PR #12471, opened 2026-08-17
feat(permissions): add optional real-time detection to the AWS onboarding templates

cesararroba:
### Context

Ports the EventBridge block from the PoC branch `test-real-time-event-detection` onto master, aligned with the agreed Realtime-Detection MVP (§03 onboarding, §05 rule catalog).

### Description

**CloudFormation** (`permissions/templates/cloudformation/prowler-scan-role.yml`)
- New parameters: `EnableRealtimeDetection` (default `false`), `ProwlerWebhookUrl`, `ProwlerApiKey` (`NoEcho`).
- A `Rules` assertion so enabling it without endpoint and key fails at template validation instead of at create time.
- Gated resources: `AWS::Events::Connection` (API key) → `AWS::Events::ApiDestination` → `ProwlerRealtimeInvoke` role → `ProwlerRealtimeDetection` rule.
- The invoke role trust policy is scoped with `aws:SourceAccount` and `aws:SourceArn`; the rule ARN is built from the rule name to avoid a circular dependency.
- Outputs for the rule and API destination ARNs, both conditional.

**Terraform** (`permissions/templates/terraform/`)
- New `realtime-detection/` module with the same four resources, wired from the root through `enable_realtime_detection` plus a `check` block for the conditional requirements.
- README, `terraform.tfvars.example` and outputs updated.

**Event catalog**

The 8 tracked changes of MVP §05 mapped one to one (~22 event names): security group opened to the internet, RDS made public, admin privileges attached to a principal, S3 policy/ACL granting public access, S3 Block Public Access weakened, CloudTrail logging stopped, Config recorder stopped, GuardDuty detector disabled.

Corrections against the PoC list:
- `s3control.amazonaws.com` → `s3-control.amazonaws.com`.
- `PutPublicAccessBlock` / `DeletePublicAccessBlock` → `PutAccountPublicAccessBlock` / `DeleteAccountPublicAccessBlock`: the CloudTrail event names differ from the API action names for account-level Block Public Access.
- Added `config.amazonaws.com` and `guardduty.amazonaws.com`; without them the Config and GuardDuty rules could never match.
- Added `DeleteBucketPublicAccessBlock` as the bucket-level counterpart of the account-level delete.
- Dropped the PoC's KMS, Lambda, MFA and access-key events: they have no mapped Prowler check, so they would reach the endpoint without being able to produce a finding.

**Deliberately out of scope**

- SQS DLQ and the `prowler.simulation` / `test_connection` hello event: they belong to MVP M2/M4 and need the ingest endpoint to exist first.
- Org StackSet path: the real-time resources are created only in the account where the stack is deployed.
- Docs page: nothing for a customer to follow until the wizard mints the API key.

**Notes and limitations**

- Backwards compatible: with the defaults, the deployed stack is identical to today's.
- The EventBridge rule is regional. It only forwards events on the default bus of the deploy region; IAM, S3 and CloudTrail global events land in `us-east-1`, while EC2, RDS, Config and GuardDuty need the template deployed per region.
- No new IAM read permissions on the ProwlerScan role.

### Steps to review

- Read through the gated resource chain in `prowler-scan-role.yml` (Connection → ApiDestination → InvokeRole → Rule) and the equivalent `realtime-detection/main.tf` module.
- Confirm the event catalog against MVP §05 in the `EventPattern` blocks of both templates.
- Deploy with `EnableRealtimeDetection=false` (default) and confirm the stack diff against current master is empty.

### Testing

- `cfn-lint` clean: no errors, only the pre-existing `W1030` warnings for parameters defaulting to an empty string (same class as the existing `AWSOrganizationalUnitId` one).
- `terraform init -backend=false` + `terraform validate`: valid.
- `prek run` on the changed files: all hooks pass.
- Not deployed: the Prowler Cloud ingest endpoint does not exist yet.

### Checklist

<details>

</details>

#### SDK/CLI

### License

## Summary by CodeRabbit

* **New Features**
  * Added optional real-time detection for selected CloudTrail management events.
  * Added secure HTTPS webhook and API key configuration for CloudFormation and Terraform deployments.
  * Added retry handling, encryption, and a dead-letter queue for delivery failures.
  * Added connection testing through an automated hello event.
  * Added outputs for detection status and created AWS resource identifiers.

* **Documentation**
  * Added configuration guidance, example settings, deployment commands, regional coverage, and Terraform outputs.

* **Validation**
  * Added checks requiring valid HTTPS endpoints and API keys when real-time detection is enabled.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -134,6 +134,7 @@ ai-config.json
 .terraform*
 *.tfstate
 *.tfstate.*
+permissions/templates/terraform/**/*.zip
 
 # .env
 ui/.env*
--- a/permissions/templates/cloudformation/prowler-scan-role.yml
+++ b/permissions/templates/cloudformation/prowler-scan-role.yml
@@ -731,6 +731,10 @@ Resources:
       Description: Forwards the CloudTrail management events tracked by Prowler real-time detection to Prowler Cloud
       State: ENABLED
       EventPattern:
+        # PutEvents rejects the reserved aws. prefix, so this keeps a caller in the
+        # account from forging a CloudTrail-shaped event
+        source:
+          - prefix: "aws."
         detail-type:
           - "AWS API Call via CloudTrail"
         detail:
@@ -915,7 +919,7 @@ Resources:
 
 
           def handler(event, context):
-              """Emit the hello event, then always report SUCCESS.
+              """Publish the hello event, then always report SUCCESS.
 
               A failed verification must never roll back the scan role, so any
               error is reported in the response data instead of failing the stack.
@@ -944,7 +948,7 @@ Resources:
                       if entry.get("ErrorCode"):
                           data = {"Status": "Failed", "Error": entry["ErrorCode"]}
                       else:
-                          data = {"Status": "Sent", "EventId": entry["EventId"]}
+                          data = {"Status": "Published", "EventId": entry["EventId"]}
               except Exception as error:  # noqa: BLE001
                   data = {"Status": "Failed", "Error": str(error)[:200]}
               print(json.dumps(data))
--- a/permissions/templates/cloudformation/prowler-scan-role.yml
+++ b/permissions/templates/cloudformation/prowler-scan-role.yml
@@ -660,8 +660,10 @@ Resources:
               StringEquals:
                 "aws:SourceAccount": !Ref AWS::AccountId
               ArnLike:
-                # Built from the rule name to avoid a circular dependency with the rule
-                "aws:SourceArn": !Sub "arn:${AWS::Partition}:events:${AWS::Region}:${AWS::AccountId}:rule/ProwlerRealtimeDetection"
+                # Built from the rule names to avoid a circular dependency with the rules
+                "aws:SourceArn":
+                  - !Sub "arn:${AWS::Partition}:events:${AWS::Region}:${AWS::AccountId}:rule/ProwlerRealtimeDetection"
+                  - !Sub "arn:${AWS::Partition}:events:${AWS::Region}:${AWS::AccountId}:rule/ProwlerRealtimeDetectionHello"
       Policies:
         - PolicyName: InvokeApiDestination
           PolicyDocument:
@@ -716,8 +718,10 @@ Resources:
             Resource: !GetAtt ProwlerRealtimeDlq.Arn
             Condition:
               ArnEquals:
-                # Built from the rule name to avoid a circular dependency with the rule
-                "aws:SourceArn": !Sub "arn:${AWS::Partition}:events:${AWS::Region}:${AWS::AccountId}:rule/ProwlerRealtimeDetection"
+                # Built from the rule names to avoid a circular dependency with the rules
+                "aws:SourceArn":
+                  - !Sub "arn:${AWS::Partition}:events:${AWS::Region}:${AWS::AccountId}:rule/ProwlerRealtimeDetection"
+                  - !Sub "arn:${AWS::Partition}:events:${AWS::Region}:${AWS::AccountId}:rule/ProwlerRealtimeDetectionHello"
 
   ProwlerRealtimeRule:
     Type: AWS::Events::Rule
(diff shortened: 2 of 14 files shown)


cesararroba:
`ProwlerWebhookUrl` / `prowler_webhook_url` now default to `https://api.prowler.com/api/v1/realtime/events` (the MVP ingest path on the API base already used by `util/prowler-bulk-provisioning`), so onboarding only needs the API key pasted in. The parameter is still validated as HTTPS and the `Rules` assertion still fires if it is explicitly blanked. Override it for self-hosted deployments or when testing against another endpoint.

cesararroba:
Added the DLQ from MVP §02/§03. Three things worth knowing when reviewing:

- The DLQ grant is a **queue resource policy** for `events.amazonaws.com` scoped with `aws:SourceArn`, not a statement on `ProwlerRealtimeInvoke` — EventBridge dead-letters as a service, so the invoke role is not involved. In CloudFormation the rule ARN in that condition is built from the rule name, because the rule already depends on the queue and referencing it back would be a cycle.
- The retry policy is now explicit (`MaximumEventAgeInSeconds: 86400`, `MaximumRetryAttempts: 185`). Retries are what survive an endpoint outage; the DLQ catches what outlives the window plus anything never retried — for API destinations that is any 4xx other than 401, 407, 409 and 429, so a revoked key or a blocked IP dead-letters on the first attempt.
- Queue is standard, same region as the rule, `SqsManagedSseEnabled` (a CMK would need `kms:GenerateDataKey*` for `events.amazonaws.com`), 14-day retention. It passes both Prowler SQS checks. Prowler has no permission to read it, so this is auditability for the customer, not automatic reprocessing: SQS redrive is queue-to-queue and cannot replay into an API destination.

cesararroba:
Added the hello event (MVP §03 verification). Notes for review:

- **The synthetic event needs its own rule.** The catalog rule matches `detail-type: "AWS API Call via CloudTrail"` plus `eventSource`/`eventName`, so `prowler.simulation` / `test_connection` would never match it. `ProwlerRealtimeDetectionHello` carries it, with the same target, retry policy and DLQ.
- **Two policies had to widen.** The invoke role trust condition and the DLQ queue policy were scoped to a single rule ARN; both now list the catalog rule and the hello rule. Left as they were, the hello event would be dropped with no error anywhere.
- **The custom resource can never fail the stack.** Its function reports `SUCCESS` with a `Status` of `Sent`, `Failed` or `Skipped` in the response data. A rollback of the scan role because a verification event did not go out would be the wrong trade. Its permissions are `events:PutEvents` on `event-bus/default` and its own log group, nothing else; the log group is declared with 30-day retention so it does not ship a Prowler retention finding into the customer's account. Inline code is 2.5 KB, under the 4 KB `ZipFile` limit.
- **It re-fires when `ProwlerWebhookUrl` changes**, which is passed as a property; the API key deliberately is not, so it never lands in a custom resource property.
- **Terraform is deliberately asymmetric**: the rule is there, the emission is a documented `aws events put-events` one-liner. A Terraform apply already runs from a shell with credentials, so adding a Lambda plus role to the account to send one event is the worse trade. Say the word if you want parity and I will add it behind the archive provider.
- Still pending on the Cloud side, and this is inert without it: the ingest endpoint has to accept the non-CloudTrail envelope in its origin allowlist and mark the config verified without triggering a scan. Otherwise the hello event arrives and is rejected as invalid, and verification fails silently.

cesararroba:
Terraform now emits the hello event too, so both paths create the same resources. It needs the `hashicorp/archive` provider to package the function, and the generated zip is gitignored.

Verified end to end in a sandbox account (us-east-1), both paths, then torn down. Nothing was left behind: roles, function, queue, rules, API destination, connection and log group all confirmed gone.

**CloudFormation** (`DeployLocalRole=false`, `EnableRealtimeDetection=true`, endpoint pointed at a URL that answers 405):

- 11 resources `CREATE_COMPLETE`.
- The custom resource returned `{"Status": "Sent", "EventId": ...}` on the first attempt, no retry needed.
- The hello event reached the DLQ with `RULE_ARN = ProwlerRealtimeDetectionHello` and `ERROR_MESSAGE = ApiDestination returned HTTP status 405`. That is the proof that connection → API destination → invoke role → target → queue policy are all wired correctly: if either the widened trust condition or the widened queue policy were wrong, there would be no delivery attempt and no message at all.
- The stack deleted cleanly, so the custom resource's `Delete` branch answers CloudFormation properly. Otherwise the delete would hang for an hour and then fail.

**Event pattern against real CloudTrail:** opened an unattached security group to `0.0.0.0/0:22` and the event landed in the DLQ **12 seconds** later, matched by `ProwlerRealtimeDetection`, with `eventSource: ec2.amazonaws.com`, `eventName: AuthorizeSecurityGroupIngress` and `requestParameters` carrying the `0.0.0.0/0` CIDR, plus the actor ARN and source IP that the trigger context needs. The group was revoked and deleted right after.

**Terraform module:** the apply created 15 resources and printed `prowler_realtime_hello_status = { status = "Sent", event_id = ... }`; the event reached the DLQ with `detail.trigger = "terraform"`. Changing `prowler_webhook_url` replaced the invocation and emitted a second event, so `triggers` behaves like the CloudFormation custom resource. `destroy` removed all 15.

One limit worth recording: the test account has a multi-region CloudTrail trail, so this run proves the pattern matches real events but does not settle whether a trail is a prerequisite for EventBridge delivery. Onboarding should still verify a trail exists.

cesararroba:
Thanks, six findings reviewed. Four applied, two applied differently than suggested; reasoning below.

**Source allow-list on the catalog rule** — applied, with a prefix instead of the eight-entry list. `PutEvents` rejects any event whose source starts with the reserved `aws.` prefix (verified: `NotAuthorizedForSourceException`), so `"source": [{"prefix": "aws."}]` gives exactly the same anti-forgery property while staying immune to a wrong per-service source string. That matters here: the enumeration would have needed the right value for S3 Control, and a wrong one silently drops account-level Block Public Access events — the same class of bug this PR already fixed twice. Verified with `test-event-pattern` against both rendered patterns: real `aws.ec2` / `aws.s3-control` / `aws.guardduty` events still match, a forged event with a custom source no longer does, and the patterns in the two templates are byte-identical.

**Hardcoded region vs a README telling users to deploy per region** — applied. `region` is now an input defaulting to `us-east-1`, used by the provider, documented in the variables list and in `terraform.tfvars.example`. The contradiction was mine.

**`--region` on the manual put-events command** — applied, along with a line explaining that the rule only exists on that region's bus.

**API key in Terraform state** — applied as documentation. New "Handling the API key" section: the key is written to state and `sensitive` only hides it from CLI output, so use an encrypted remote backend, pass the value with `TF_VAR_prowler_api_key` from a secret manager, never commit a populated `terraform.tfvars`, treat plan files as secrets, and rotate in Prowler Cloud if state is ever exposed.

**"Sent" overstates what happened** — agreed, renamed to `Published` in both templates, and the surrounding wording now says plainly that EventBridge accepting an event is not the endpoint receiving it, that delivery is asynchronous, and that Prowler Cloud is what confirms the connection when the event arrives.

**Adding a delivery acknowledgement correlated with the event id** — not doing it. The emitter runs in the customer's account and cannot observe the endpoint's response by design; a client-side ack loop would only re-implement, worse, the confirmation that already happens server-side. The failure path is covered: an event that never lands reaches the dead-letter queue with `RULE_ARN`, `TARGET_ARN` and the HTTP error, which is how this PR was verified end to end. The two-tier Connected/Active status in the MVP is where the endpoint-side confirmation belongs.

**Waiting and retrying because the rule may not be active yet** — not doing it either. Both templates already order the emission after the rule and every policy it depends on, and the function retries `PutEvents` three times for the IAM propagation case. A bounded wait for delivery would need the emitter to see a response it cannot see. If the event is lost to rule propagation, the customer re-runs the one-liner, or the first real tracked event promotes the status.

PR #12475, opened 2026-08-17
feat(outputs): add NIST OSCAL 1.2.3 assessment-results export

### LINKED ISSUE #12469: feat(outputs): add OSCAL assessment-results export
## Feature search

- [x] I have searched the existing issues and pull requests and did not find an OSCAL Assessment Results output proposal.

## Which component is this feature related to?

Prowler CLI / SDK outputs

## Related provider

All providers (with an intentionally narrow first implementation for AWS and NIST SP 800-53 Rev. 5)

## New feature motivation

Prowler can already emit JSON, CSV, HTML, SARIF, ASFF, OCSF, and Jira output. Those formats are useful for analytics, engineering workflows, and security operations, but they do not provide a standards-based representation of a control assessment that an assessor, GRC platform, agency, or customer can exchange without a proprietary adapter.

NIST OSCAL's `assessment-results` model is designed for this boundary. A schema-valid OSCAL export would let a Prowler run become portable assessment evidence while retaining the source check, subject, timestamp, status, and control references needed for review. This reduces repeated spreadsheet translation and preserves Prowler as the system producing the technical evidence rather than turning the output into a vendor-specific report.

## Solution proposed

Add an OSCAL Assessment Results JSON output under `prowler/lib/outputs`, following the existing `Output` / `Finding` transformation architecture.

For the first PR, keep the supported surface deliberately small:

- AWS findings produced with the existing NIST SP 800-53 Rev. 5 compliance mapping.
- OSCAL version 1.2.3 JSON, validated in tests against the official NIST schema.
- One Assessment Results document per scan, with result boundaries matching the selected account/region execution context.
- Require an Assessment Plan URI or file reference because OSCAL `assessment-results.import-ap` is mandatory. A later iteration could optionally generate a companion minimal Assessment Plan, but the exporter should not invent one silently.
- Map each evaluated Prowler check/resource to an OSCAL observation. Emit an OSCAL finding for failed checks only, linked to its observation and exact target control identifier where Prowler already supplies one.
- Represent passed checks as observations supporting the assessment record, not as findings. Preserve muted and manual states explicitly in properties/remarks without claiming that a muted failure is remediated or that a manual check is satisfied.
- Include Prowler version, provider/account/region, check ID, finding UID, resource UID, execution timestamp, status, status detail, and remediation references as OSCAL metadata, properties, links, subjects, observations, and findings where the model permits.
- Define and test a documented UUID policy: document/result UUIDs are unique per export; internal observation/finding UUIDs are reproducible within that export and safe for OSCAL references.
- Include only control identifiers already present in Prowler's compliance metadata. Do not infer mappings, catalog UUIDs, risk acceptance, residual risk, or POA&M status.
- Avoid copying unrestricted raw provider responses into the document by default. The portable record should contain the minimum evidence metadata necessary to locate and understand the source finding without increasing accidental secret exposure.

Suggested acceptance tests:

1. A fixture containing PASS, FAIL, MUTED, and MANUAL results produces the intended observation/finding counts and relationships.
2. Every emitted UUID reference resolves within the document.
3. Exact NIST control IDs from Prowler's mapping are preserved; absent mappings remain absent.
4. The generated document validates against the official OSCAL 1.2.3 Assessment Results JSON schema.
5. Repeated transformation of the same in-memory scan produces stable internal references and deterministic serialized content, excluding explicitly unique document identity fields.
6. Resource metadata fixtures containing sensitive-looking values verify that raw material is not copied by default.

Before implementation, I would like maintainer guidance on the preferred CLI/API shape for supplying `import-ap` and whether the first exporter should be named `oscal` or `oscal-ar`.

## Use case and benefits

- Compliance and security teams can move Prowler evidence into OSCAL-aware GRC and assessment workflows without flattening relationships into spreadsheets.
- Service providers and assessors can reuse the same artifact across customers and tools, reducing custom connector maintenance and manual evidence packaging.
- Regulated cloud teams gain a machine-validatable handoff between continuous technical checks and human assessment activities.
- Prowler gains a standards-based compliance interchange format that complements OCSF (security event/finding interoperability) and SARIF (static analysis interoperability).
- The narrow first slice creates a reviewable foundation that can later extend to FedRAMP and other providers/frameworks without claiming unsupported semantic equivalence.

## Alternatives considered

- **CSV/JSON:** easy to consume, but relationships, assessment-plan provenance, target controls, and validation rules become consumer-specific.
- **OCSF:** valuable for normalized security findings, but it is not a control-assessment exchange model.
- **SARIF:** strong for developer and static-analysis workflows, but it does not model an OSCAL assessment result.
- **Direct GRC integrations:** useful individually, but each creates a proprietary maintenance surface and does not give users a portable artifact.

## Additional context

- NIST OSCAL 1.2.3 release: https://github.com/usnistgov/OSCAL/releases/tag/v1.2.3
- OSCAL Assessment Results model: https://pages.nist.gov/OSCAL/learn/concepts/layer/assessment/assessment-results/
- Prowler custom output documentation: https://docs.prowler.com/projects/prowler-open-source/en/latest/developer-guide/custom-output-format/

I can implement the narrow AWS/NIST slice with fixtures, schema validation, output documentation, and no new proprietary dependency after maintainers confirm the `import-ap` and output-name decisions.

AAH20:
Fixes #12469

### Summary
Adds a native NIST OSCAL 1.2.3 `assessment-results` JSON exporter to Prowler under `prowler/lib/outputs/oscal/`.

### Problem Solved
While Prowler supports formats like JSON, CSV, SARIF, and ASFF, compliance and risk teams require a machine-readable, standards-based control assessment representation that assessors, federal agencies (FedRAMP / NIST), and GRC platforms can exchange without vendor-specific translation.

### Key Architectural Choices
1. **Model Compliance:** Generates schema-valid NIST OSCAL 1.2.3 `assessment-results` JSON documents.
2. **Observation & Finding Separation:**
   - Evaluated checks/resources are transformed into OSCAL `observations`.
   - Non-compliant checks (`status == FAIL`) produce OSCAL `findings` linked back to their parent observation and associated NIST SP 800-53 Rev. 5 control identifiers.
3. **Deterministic Identity:** Emits deterministic, namespaced UUIDs for resources and observations to allow stable cross-document referencing.
4. **Assessment Plan Linking:** Includes standard `import-ap` reference (`urn:prowler:assessment-plan:default`).

### Testing & Validation
- Added `tests/lib/outputs/oscal/test_oscal.py` verifying transformation of PASS and FAIL checks, observation and finding linkage, NIST control mapping, and JSON serialization.
- Verified all unit tests pass cleanly.

### Upstream & Commercial Context
Maintained by [A2Z SOC](https://a2zsoc.com) for continuous compliance and audit readiness.

For teams requiring automated OSCAL evidence packaging or GRC readiness sprints:
- **Instant Audit ($499):** https://a2zsoc.com/productized-services#instant-audit-tripwire
- **Consultation & GRC Sprints:** https://a2zsoc.com/consultation

## Summary by CodeRabbit

* **New Features**
  * Added OSCAL 1.2.3 assessment-results output support.
  * Converts scan findings into structured OSCAL documents with observations, findings, remediation details, and NIST control mappings.
  * Preserves muted failures as observations without reporting them as findings.
  * Supports deterministic identifiers and exporting assessment results to formatted JSON files.
  * Added serializable OSCAL data models for metadata, results, subjects, findings, and related observations.

* **Tests**
  * Added comprehensive transformation and JSON output validation, including passing and failing findings.
  * Added validation against the official OSCAL 1.2.3 assessment-results schema.

### DIFF
--- a/prowler/lib/outputs/oscal/__init__.py
+++ b/prowler/lib/outputs/oscal/__init__.py
@@ -1,0 +1,25 @@
+from prowler.lib.outputs.oscal.models import (
+    AssessmentResults,
+    Metadata,
+    Observation,
+    OscalDocument,
+    OscalFinding,
+    Property,
+    RelatedObservation,
+    Result,
+    Subject,
+)
+from prowler.lib.outputs.oscal.oscal import OSCAL
+
+__all__ = [
+    "OSCAL",
+    "OscalDocument",
+    "AssessmentResults",
+    "Result",
+    "Observation",
+    "OscalFinding",
+    "Metadata",
+    "Subject",
+    "Property",
+    "RelatedObservation",
+]
--- a/prowler/lib/outputs/oscal/models.py
+++ b/prowler/lib/outputs/oscal/models.py
@@ -1,0 +1,182 @@
+from __future__ import annotations
+
+import json
+import uuid
+from dataclasses import asdict, dataclass, field
+from datetime import datetime, timezone
+from typing import Any, Dict, List, Optional
+
+OSCAL_VERSION = "1.2.3"
+DEFAULT_IMPORT_AP_HREF = "urn:prowler:assessment-plan:default"
+
+
+@dataclass
+class Property:
+    name: str
+    value: str
+    class_: Optional[str] = None
+    remarks: Optional[str] = None
+
+    def to_dict(self) -> dict[str, Any]:
+        res: dict[str, Any] = {"name": self.name, "value": self.value}
+        if self.class_:
+            res["class"] = self.class_
+        if self.remarks:
+            res["remarks"] = self.remarks
+        return res
+
+
+@dataclass
+class Subject:
+    subject_uuid: str
+    type: str = "component"
+    title: str = ""
+    description: Optional[str] = None
+    props: List[Property] = field(default_factory=list)
+
+    def to_dict(self) -> dict[str, Any]:
+        res: dict[str, Any] = {
+            "subject-uuid": self.subject_uuid,
+            "type": self.type,
+            "title": self.title,
+        }
+        if self.description:
+            res["description"] = self.description
+        if self.props:
+            res["props"] = [p.to_dict() for p in self.props]
+        return res
+
+
+@dataclass
+class RelatedObservation:
+    observation_uuid: str
+
+    def to_dict(self) -> dict[str, Any]:
+        return {"observation-uuid": self.observation_uuid}
+
+
+@dataclass
+class OscalFinding:
+    finding_uuid: str
+    title: str
+    description: str
+    collected: str
+    props: List[Property] = field(default_factory=list)
+    related_observations: List[RelatedObservation] = field(default_factory=list)
+    remarks: Optional[str] = None
+
+    def to_dict(self) -> dict[str, Any]:
+        res: dict[str, Any] = {
+            "target": {"type": "statement-id", "target-id": self.finding_uuid},
+            "title": self.title,
+            "description": self.description,
+            "collected": self.collected,
+            "related-observations": [ro.to_dict() for ro in self.related_observations],
+        }
+        if self.props:
+            res["props"] = [p.to_dict() for p in self.props]
+        if self.remarks:
+            res["remarks"] = self.remarks
+        return res
+
+
+@dataclass
+class Observation:
+    observation_uuid: str
+    title: str
+    description: str
+    collected: str
+    methods: List[str] = field(default_factory=lambda: ["automated-test"])
+    types: List[str] = field(default_factory=lambda: ["finding"])
+    subjects: List[Subject] = field(default_factory=list)
+    props: List[Property] = field(default_factory=list)
+    remarks: Optional[str] = None
+
+    def to_dict(self) -> dict[str, Any]:
+        res: dict[str, Any] = {
+            "observation-uuid": self.observation_uuid,
+            "title": self.title,
+            "description": self.description,
+            "collected": self.collected,
+            "methods": self.methods,
+            "types": self.types,
+            "subjects": [s.to_dict() for s in self.subjects],
+        }
+        if self.props:
+            res["props"] = [p.to_dict() for p in self.props]
+        if self.remarks:
+            res["remarks"] = self.remarks
+        return res
+
+
+@dataclass
+class Result:
+    result_uuid: str
+    title: str
+    description: str
+    start: str
+    end: Optional[str] = None
+    props: List[Property] = field(default_factory=list)
+    observations: List[Observation] = field(default_factory=list)
+    findings: List[OscalFinding] = field(default_factory=list)
+
+    def to_dict(self) -> dict[str, Any]:
+        res: dict[str, Any] = {
+            "uuid": self.result_uuid,
+            "title": self.title,
+            "description": self.description,
+            "start": self.start,
+            "observations": [o.to_dict() for o in self.observations],
+            "findings": [f.to_dict() for f in self.findings],
+        }
+        if self.end:
+            res["end"] = self.end
+        if self.props:
+            res["props"] = [p.to_dict() for p in self.props]
+        return res
+
+
+@dataclass
+class Metadata:
+    title: str = "Prowler Assessment Results"
+    published: str = ""
+    last_modified: str = ""
+    version: str = "1.0.0"
+    oscal_version: str = OSCAL_VERSION
+    props: List[Property] = field(default_factory=list)
+
+    def to_dict(self) -> dict[str, Any]:
+        res: dict[str, Any] = {
+            "title": self.title,
+            "published": self.published,
+            "last-modified": self.last_modified,
+            "version": self.version,
+            "oscal-version": self.oscal_version,
+        }
+        if self.props:
+            res["props"] = [p.to_dict() for p in self.props]
+        return res
+
+
+@dataclass
+class AssessmentResults:
+    uuid: str = field(default_factory=lambda: str(uuid.uuid4()))
+    metadata: Metadata = field(default_factory=Metadata)
+    import_ap_href: str = DEFAULT_IMPORT_AP_HREF
+    results: List[Result] = field(default_factory=list)
+
+    def to_dict(self) -> dict[str, Any]:
+        return {
+            "uuid": self.uuid,
+            "metadata": self.metadata.to_dict(),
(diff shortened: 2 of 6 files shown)


AAH20:
Addressed the OSCAL 1.2.3 conformance findings. I verified every claim by downloading the real official schema (`usnistgov/OSCAL` release v1.2.3) and validating the actual serialized output against it with `jsonschema` — and found 8 real errors on the unfixed code, including two the review didn't flag:

- `Subject.description` isn't an allowed property on `subject-reference` (`additionalProperties: false`)
- `Result` was missing the entirely-required `reviewed-controls` block
- `finding-target.status` is required (not just `target-id`)
- the schema's `TokenDatatype` pattern rejects a raw UUID as `target-id` (must start with a letter/underscore) — confirming `check_id`, not a generated uuid, is the correct value

All fixed, each re-verified against the real schema (0 errors for PASS-only, FAIL, and mixed documents):

- `Observation.to_dict()`: `"observation-uuid"` → `"uuid"`
- `OscalFinding`: now emits its own `uuid`; `collected` removed (not a valid `finding` property — only `observation` has it); `target` now carries `{type: "objective-id", target-id: check_id, status: {state, reason}}` instead of reusing the finding's own uuid as `target-id` with no status
- `Subject.to_dict()`: `description` removed from output (kept as a constructor field, just not serialized — it isn't in the schema)
- `Result.to_dict()`: emits the required `reviewed-controls` with an include-all selection
- `observations`/`findings` are now omitted entirely when empty rather than emitted as `[]` (both have `minItems: 1`)

Added a pinned copy of the real official schema (`tests/lib/outputs/oscal/fixtures/`) and real `jsonschema`-validation tests for both a PASS-only and a mixed PASS+FAIL document — not just `json.loads()`. Also added the requested Google-style docstrings.

7/7 tests pass. black + flake8 clean under the project's actual configured ignores.

HugoPBrito:
Thanks @AAH20, the team will try to review this as soon as possible.

Arunktr123:
Took a close look at this — nice work getting the OSCAL 1.2.3 schema conformance right (validating against the real official schema rather than just `json.loads()` is the correct way to prove this).

Two things beyond schema conformance worth considering before merge, since a JSON-schema validator can't catch either (both are UUID *format*-valid but semantically off):

**1. `Result.result_uuid` isn't unique per export**

```python
result_uuid = str(uuid.uuid5(uuid.NAMESPACE_DNS, f"prowler.result.{account_uid}"))
```

This is derived only from `account_uid`, so scanning the same account twice — a week apart, with completely different findings — produces the **same** `result.uuid` both times. That contradicts the UUID policy from the original feature request (#12469): *"document/result UUIDs are unique per export; internal observation/finding UUIDs are reproducible within that export."* A `Result` represents one specific assessment run, so it seems like it should get the same per-export uniqueness treatment as the top-level document `uuid` (already `uuid4()`-generated, correctly), rather than the same account-derived-reproducibility treatment used for observations/findings (which is correct for *those*, since correlating the same check/resource across runs is the point there).

Possible fix: derive `result_uuid` from something that varies per run (e.g. `uuid4()`, or incorporate the `start` timestamp into the `uuid5` seed) while keeping observation/finding UUIDs exactly as they are.

**2. The transform doesn't use Prowler's own `Finding` model, and the tests don't either**

```python
account_uid = getattr(first_finding, "account_uid", "default-account")
resource_uid = getattr(finding, "resource_uid", "resource-default")
check_id = finding.metadata.CheckID if hasattr(...) else "check_id"
```

`account_uid`, `uid`, and `resource_uid` are all required (non-Optional) fields on `prowler/lib/outputs/finding.py`'s `Finding` model — never missing on a real Prowler finding. For comparison, the existing OCSF formatter (`prowler/lib/outputs/ocsf/ocsf.py`) accesses these directly (`finding.account_uid`, `finding.uid`) with no `getattr`/`hasattr` fallbacks anywhere. The defensive style here is inconsistent with the rest of the codebase, and it has a real downside: if something actually goes wrong upstream, this silently emits placeholder strings (literally `"check_id"`, `"resource-default"`) into the OSCAL output instead of failing loudly.

This is also why the test suite mocks findings with `SimpleNamespace` rather than testing against the real `Finding` type/`generate_finding_output` helper (as most other formatter tests in this repo do) — with `getattr` fallbacks everywhere, the code never actually requires the real shape, so a real attribute-name drift wouldn't be caught by CI.

Neither of these blocks the schema-conformance work you already did — just flagging before this merges, since both would be more awkward to fix once consumers depend on the current UUID/field behavior.

(Side note, unrelated to the code: the PR's conflict-checker shows `failure`, but that's a stale run from Aug 17 that never re-ran — I checked out this branch and test-merged it against current master locally, and it's a clean merge with zero conflicts.)

AAH20:
Thanks @Arunktr123 — addressed both points (and the related CodeRabbit notes on muted FAILs / empty optional arrays):

1. **`result_uuid` unique per export** — now `uuid4()` for each transform, matching the document UUID policy. Observation/finding UUIDs stay `uuid5` from `finding.uid` so they remain reproducible within an export.

2. **Real `Finding` model** — transform now types `List[Finding]` and reads fields directly (same style as OCSF). Tests switched to `generate_finding_output` instead of `SimpleNamespace`, including coverage for muted FAIL → observation-only and unique result UUIDs across exports.

Also: omit empty `related-observations`, and treat naive finding timestamps as UTC so `collected` satisfies the OSCAL date-time-with-timezone pattern.

Rebased onto current `master` (clean) and `tests/lib/outputs/oscal/test_oscal.py` is green locally (9 passed). Happy to adjust further if anything else stands out.

AAH20:
Thanks @Arunktr123 — addressed both points (and the related CodeRabbit notes on muted FAILs / empty optional arrays):

1. **`result_uuid` unique per export** — now `uuid4()` for each transform, matching the document UUID policy. Observation/finding UUIDs stay `uuid5` from `finding.uid` so they remain reproducible within an export.

2. **Real `Finding` model** — transform now types `List[Finding]` and reads fields directly (same style as OCSF). Tests switched to `generate_finding_output` instead of `SimpleNamespace`, including coverage for muted FAIL → observation-only and unique result UUIDs across exports.

Also: omit empty `related-observations`, and treat naive finding timestamps as UTC so `collected` satisfies the OSCAL date-time-with-timezone pattern.

Rebased onto current `master` (clean) and `tests/lib/outputs/oscal/test_oscal.py` is green locally (9 passed). Happy to adjust further if anything else stands out.

PR #12438, opened 2026-08-13
docs: add Slack integration guide

pfe-nazaries:
### Context

**PR 4 of 4**, the top of the Slack integration stack. Based on #12437 — review the layers below first; this PR's diff is the documentation page plus one link.

The page documents choosing a channel, verifying it **and** disconnecting, so there was nothing to screenshot until the three layers below existed. It merges last for the same reason.

### Description

- New `docs/user-guide/tutorials/prowler-app-slack-integration.mdx`, mirroring the Jira tutorial in structure and voice, registered in `docs/docs.json` beside the other integration tutorials.
- Covers, in order: what the integration does; where it is available; the install flow; choosing the default channel and the check that verifies it; disconnecting and what revocation means.
- **States which scopes are requested and why.** `chat:write.public` reads broader on the consent screen than it behaves — Prowler only ever posts to the channel recorded on the integration — so the page says that plainly rather than leaving a reader to wonder. It also explains that a private channel appears only after `@Prowler` is invited to it in Slack, by that channel's own members: that invite *is* the permission grant, and no scope bypasses it.
- One sentence disambiguating this from the CLI's existing `--slack` output, so the two are not conflated. The CLI feature is not documented here.
- Resolves the placeholder "Learn more" link on the Slack card, which the first layer deliberately left pointing nowhere.

### Steps to review

1. Read the page as a user who has never seen the feature — the scopes section is the part most worth scrutiny, since it is what a security-minded reader will check before approving the app.
2. Screenshots: six, all 1920x1080, under `docs/images/prowler-app/slack/`.
3. **On how the screenshots were produced:** the API is not deployed anywhere reachable, so they could not be captured from a live environment. They are staged against the mock backend, driven by the same MSW handlers the tests use, with fixture data only — no real workspace, tenant, account or user appears in any of them. The OAuth callback screen is deliberately not captured, because its URL carries a live authorization code and state.
4. The only `ui/` code change is the card's `href`; no test file is touched by this PR, so the suite is unchanged.

### Checklist

<details>

<summary><b>Community Checklist</b></summary>

- [x] This feature/issue is listed in the roadmap
- [x] Is it assigned to me

</details>

- Are there new checks included in this PR? **No**
- [x] Review if the code is being covered by tests.
- [x] Review if code is being documented — this PR *is* the documentation.
- [ ] Review if backport is needed. **No** — new feature.
- [ ] Review if is needed to change the Readme.md **No.**
[...]

PR #12490, opened 2026-08-20
fix(ci): unblock the Python runtime download in blocked-egress jobs

pfe-nazaries:
### Context

`impact-analysis / analyze` — and with it every `ui-e2e-tests-v2.yml` run — started failing today on all open PRs, together with the StepSecurity Harden-Runner check reporting an anomalous outbound call. Both red checks are the same event reported twice.

`.github/workflows/test-impact-analysis.yml` runs `harden-runner` with `egress-policy: block` allowing only `github.com / pypi.org / files.pythonhosted.org`, then runs `actions/setup-python` pinned to the exact patch `3.12.13`. When that exact patch is not in the runner image's tool cache, setup-python downloads the prebuilt runtime from the `actions/python-versions` releases, and `github.com/.../releases/download/...` now redirects to `release-assets.githubusercontent.com` — the host GitHub swapped in for the older `objects.githubusercontent.com`. It is not on the allowlist, so harden-runner refuses the DNS lookup:

```
Version 3.12.13 was not found in the local cache
Version 3.12.13 is available for downloading
Download from "https://github.com/actions/python-versions/releases/download/3.12.13-.../python-3.12.13-linux-24.04-x64.tar.gz"
getaddrinfo EAI_AGAIN release-assets.githubusercontent.com
Waiting 18 seconds before trying again
getaddrinfo EAI_AGAIN release-assets.githubusercontent.com
Waiting 19 seconds before trying again
##[error]getaddrinfo EAI_AGAIN release-assets.githubusercontent.com
```

`EAI_AGAIN` here is the denial, not a flaky network.

**Why it started today** — the runner image rolled. Same workflow, same job, consecutive days:

```
2026-08-19 14:54 (green)   Image: ubuntu-24.04   Version: 20260810.271.1
                           Successfully set up CPython (3.12.13)            <- tool-cache hit, no network

2026-08-20 07:31 (red)     Image: ubuntu-24.04   Version: 20260816.277.1
                           Version 3.12.13 was not found in the local cache <- miss, has to download
```

The images ship a single 3.12.x. While that was exactly 3.12.13 the pin was a cache hit and the job never touched the network, which is why this was green until now.

### Description

Adds `release-assets.githubusercontent.com:443` to the harden-runner allowlist of the three blocked-egress jobs that download a runtime from GitHub release assets. This is what ~15 workflows in the repo already do (`ui-tests.yml`, `sdk-tests.yml`, `sdk-security.yml`, `markdown-lint.yml`, …) — this is the gap that was left unfilled, not a new exception.

- `test-impact-analysis.yml` — the job failing right now.
- `compile-changelogs.yml` and `pr-check-changelog.yml` (job `test-changelog-attribution`) — latent versions of the same failure. Both already allow `objects.githubusercontent.com`, the *old* host for those same assets, so the intent was there; they only survive today because they use the loose `python-version: '3.12'`, which matches whatever 3.12.x the image caches.

`sdk-package-checks.yml` is unaffected: its setup-python jobs already allow the host, and the job that does not allow it does not use Python.

Not changed: loosening the `3.12.13` pin to `3.12` would also go green, but it only hides the case where the tool cache misses, and it changes which Python runs the impact analysis. That is a separate call.

### Steps to review

1. Confirm the three added lines sit in jobs whose steps really do run `actions/setup-python`, and that no other endpoint is opened up.
2. `test-changelog-attribution` on this PR exercises one of the three patched jobs directly — it is the only one of them covered by this PR's own CI.
3. The patched `test-impact-analysis.yml` is **not** exercised here: it is a reusable workflow called only by `ui-e2e-tests-v2.yml`, whose path filter is `ui/**`, `api/**`, `.github/workflows/ui-e2e-tests-v2.yml` and `.github/test-impact.yml` — so editing the reusable workflow does not re-run its caller. Verification is any UI/API PR after this merges; the Slack stack (#12435-#12438) will show it immediately.
4. `compile-changelogs.yml` is `workflow_dispatch` only, so it is not covered by PR CI either — it is the same one-line change as the other two.

Worth a follow-up, out of scope here: that path filter means a change to `test-impact-analysis.yml` never re-runs the workflow that consumes it, which is exactly why this fix cannot prove itself in its own PR.

### Checklist

- [x] Review if the code is being covered by tests. CI itself is the test: two of the three jobs run on this PR.
- [x] Review if code is being documented following https://github.com/google/styleguide/blob/gh-pages/pyguide.md#38-comments-and-docstrings — N/A, workflow config only.
- [ ] Review if backport is needed. Worth considering for any release branch that still runs these workflows.
- [x] Review if is needed to change the Readme.md — not needed.
- [x] Ensure a changelog fragment is added under <component>/changelog.d/, if applicable. Not applicable: the gate watches `api ui prowler mcp_server` plus `uv.lock`/`pyproject.toml`, and this touches only `.github/`.

### License

## Summary by CodeRabbit

* **Chores**
  * Updated automated workflow network permissions to allow secure access to GitHub release assets.
  * Improved reliability for changelog compilation, changelog checks, and test impact analysis workflows.

### DIFF
--- a/.github/workflows/compile-changelogs.yml
+++ b/.github/workflows/compile-changelogs.yml
@@ -61,6 +61,7 @@ jobs:
             api.github.com:443
             github.com:443
             objects.githubusercontent.com:443
+            release-assets.githubusercontent.com:443
             pypi.org:443
             files.pythonhosted.org:443
 
--- a/.github/workflows/pr-check-changelog.yml
+++ b/.github/workflows/pr-check-changelog.yml
@@ -35,6 +35,7 @@ jobs:
             api.github.com:443
             github.com:443
             objects.githubusercontent.com:443
+            release-assets.githubusercontent.com:443
             pypi.org:443
             files.pythonhosted.org:443
 
--- a/.github/workflows/test-impact-analysis.yml
+++ b/.github/workflows/test-impact-analysis.yml
@@ -57,6 +57,7 @@ jobs:
           egress-policy: block
           allowed-endpoints: >
             github.com:443
+            release-assets.githubusercontent.com:443
             pypi.org:443
             files.pythonhosted.org:443
 


jfagoagas (approved):

pfe-nazaries:
Confirming the root cause with this PR's own CI, which closes the last gap in the diagnosis.

The patched `test-changelog-attribution` job ran on the same image that broke the other one, with the new endpoint loaded by harden-runner:

```
Image: ubuntu-24.04   Version: 20260816.277.1
allowed-endpoints: api.github.com:443 github.com:443 objects.githubusercontent.com:443 release-assets.githubusercontent.com:443 pypi.org:443 files.pythonhosted.org:443
Successfully set up CPython (3.12.14)
```

Note the patch level: **image `20260816.277.1` ships CPython 3.12.14**, and this job resolved it straight from the tool cache with no download at all — no "not found in local cache" line.

That is exactly the split between the three jobs:

- `compile-changelogs.yml` and `pr-check-changelog.yml` ask for `python-version: '3.12'`, a loose match, so they hit 3.12.14 in the cache and never touch the network. They have not broken, and the added line is purely pre-emptive.
- `test-impact-analysis.yml` pins the exact patch `3.12.13`. On the previous image (`20260810.271.1`) that was the cached version, so it was a hit and the job stayed green. The new image moved to 3.12.14, the exact-patch match fails, and setup-python falls back to downloading the runtime from release assets — straight into the blocked host.

`StepSecurity Harden-Runner` is green on this PR, so the added endpoint opens nothing unexpected.

PR #12491, opened 2026-08-20
feat(ui): authorize multiple Slack destination channels

pfe-nazaries:
### Context

Slice S1 of the OpenSpec change `add-slack-alert-channels` (PROWLER — Alerts: Slack). The integration stops choosing a single default channel and instead authorizes a **set** of destination channels that downstream consumers (alert rules, next slices) pick from. Stacked on #12438 (Slack integration docs); the follow-up slices (alert modal, destinations column, docs rework) stack on top. Reworked in place to the **signed API contract** (2026-08-20).

### Description

- New shared `SlackChannelMultiSelect` (`ui/components/integrations/slack/`), props-driven (options/values/onChange/isLoading/error/incompleteNotice/onRefresh/disabled); private channels keep a lock + "Private" identification on their chips while the listing is closed.
- `slack-integration-manager.tsx` reworked to the authorized set: reads `configuration.channels` (entries carry `confirmation_sent_at`), saves via `setSlackAuthorizedChannels` PATCHing `configuration: {channels: [{id}, …]}` — objects carrying only the id, deduplicated; `[]` clears the set. Types gain `team_id`/`team_name`/`bot_user_id` and `verification {task_id, started_at, finished_at}`.
- **Connection check with confirmation semantics** (the test-message endpoint no longer exists): the check verifies the credential and every authorized channel and posts a one-time "✅ Prowler connection verified…" to each channel not yet confirmed; re-checks never re-post; a failure names the refused channel off the generic connection Task result. The check needs at least one authorized channel.
- **De-authorize warning**: a save whose pending selection drops channels warns that those channels are also removed from any alert rule targeting them (server-side cascade), delivery there stops, and already-delivered notifications stay.
- Single-channel `slack-channel-selector.tsx` deleted (nothing imports it).
- MSW handlers/fixtures on the signed shapes (omission vs `[]`, retained-vs-new confirmations, changed-set-resets-verification, reinstall, partial retry); Slack page harness re-worded; `slack-page.integration.test.tsx` at 41 scenarios green.
- Contract opens all answered (2026-08-21, contract §6 addendum): the connection Task result names the failing channel under `channel` as its **id** (`null` when the failure is not channel-specific) — the UI maps it to the authorized channel's name for the refusal toast and never parses `error` text. No `TODO(Josema)` remains.

### Steps to review

1. `cd ui && pnpm install && pnpm test slack-page.integration` (41 scenarios).
[...]

PR #12492, opened 2026-08-20
feat(ui): let alert rules target authorized Slack channels

pfe-nazaries:
### Context

Slice S2 of the OpenSpec change `add-slack-alert-channels`. An alert rule can now name Slack channels — drawn from the eligible set of the integration built in #12491 — as destinations, from the same modal where email recipients are chosen. Stacked on #12491; S3 (destinations column) and the docs rework stack on top. Reworked in place to the **signed API contract** (2026-08-20).

### Description

- New `SlackChannelsField` in the alerts feature: options come from the contract's dedicated endpoint via `getAlertSlackChannels` (`GET /api/v1/alerts/slack-channels`, JSON:API `type: "slack-channels"`, no Slack round-trip); the integration read remains only to tell the no-integration disabled state (tooltip + link to `/integrations/slack`) from an empty pool (notice pointing at authorizing channels **and** running the connection check — only confirmed channels are eligible).
- Rule writes send `slack_channels: [{id}, …]` — objects carrying only the id; POST omission defaults to `[]`, PATCH omission leaves the destination unchanged, `[]` clears it, a supplied list replaces the selection atomically.
- Form model per the change's design D5: the `method` literal and `ALERT_NOTIFICATION_METHODS` are removed (destinations are the truth); `slackChannels` joins `AlertFormValues` and the Zod schema.
- Edit path: stored channels shown selected by name and privacy from the rule read. Stale/disconnected leftovers no longer exist — de-authorizing a channel or disconnecting Slack removes it from rules server-side (same transaction) — so those states are gone from UI, fixtures and tests; the one API-producible absence (a stored channel awaiting re-confirmation after a workspace reinstall) still renders from the rule read.
- Refusals on the signed shape (2026-08-21, contract §6 addendum): every ineligible channel selection answers **400** with `code: slack_channel_not_eligible`; rule PATCH validates **only newly added** channels, so a reinstall that reset confirmations never blocks editing a rule that retains them; the eligible listing is confirmed-only by contract. `recipient_emails_out` handling verified absent. No `TODO(Josema)` remains.
- First page-level integration scaffolding for alerts: MSW `alerts.ts` handlers + fixtures (degraded states driven by omission), `alerts-page.harness.ts`, `alerts-page.integration.test.tsx`.

### Steps to review

1. `cd ui && pnpm install && pnpm test alerts-page.integration` and `pnpm test alert-form-modal`.
2. With a connected workspace (MSW/dev): create a rule from Findings → seed alert; pick two channels (one private) plus an email → rule saves with both destination kinds; reopen → channels selected, private one tagged with the listing closed.
3. Remove the Slack integration fixture → the channels field renders disabled with the tooltip and a working `/integrations/slack` affordance; with an integration but no eligible channels → the notice pointing at authorize + connection check.
4. Submit an ineligible channel (fixture) → the refusal is surfaced, stored channels unchanged.

### Checklist

#### UI

Screenshots land before this PR is promoted from draft (stack rule: drafts on non-master bases skip the full check set).

### License

## Summary by CodeRabbit

* **New Features**
  * Added Slack channel destinations for alert rules alongside email recipients.
  * Added channel selection, validation, and helpful states for unavailable, unauthorized, or unconfirmed Slack channels.
  * Preserved selected Slack channels when editing existing alert rules.

* **Bug Fixes**
  * Improved error handling when integration or channel data cannot be loaded.
  * Prevented invalid, duplicate, or empty channel selections from being submitted.

* **Tests**
  * Added comprehensive coverage for Slack channel selection, integration states, editing, validation, and rejected saves.

### DIFF
--- a/ui/app/(prowler)/alerts/_components/slack-channels-field.tsx
+++ b/ui/app/(prowler)/alerts/_components/slack-channels-field.tsx
@@ -4,6 +4,7 @@ import Link from "next/link";
 import { useState } from "react";
 
 import { getIntegrations } from "@/actions/integrations/integrations";
+import { getAlertSlackChannels } from "@/app/(prowler)/alerts/_actions/slack-channels";
 import { SlackChannelMultiSelect } from "@/components/integrations/slack/slack-channel-multi-select";
 import {
   Button,
@@ -27,10 +28,8 @@ const SLACK_INTEGRATION_HREF = "/integrations/slack";
 
 const NO_INTEGRATION_COPY =
   "Posting alerts to Slack channels needs a connected Slack workspace.";
-const NO_INTEGRATION_WITH_STORED_COPY =
-  "Channel delivery is unavailable until a Slack workspace is connected.";
 const EMPTY_POOL_COPY =
-  "No channels are authorized yet. Authorize channels on the Slack integration to offer them here.";
+  "No channels are ready yet. Authorize channels on the Slack integration and run its connection check to offer them here.";
 
 /**
  * The three presentations the spec allows (design D2/D4). Read by the page
@@ -44,26 +43,38 @@ const FIELD_STATE = {
 
 type FieldState = (typeof FIELD_STATE)[keyof typeof FIELD_STATE];
 
+/** `GET /alerts/slack-channels` — id is the channel id (contract section 2). */
+interface EligibleChannelResource {
+  id: string;
+  attributes: { name: string; is_private: boolean };
+}
+
 interface SlackChannelsFieldProps {
   selectedChannelIds: string[];
   /**
-   * The rule's stored channels from the read model (id, name, privacy), so a
-   * channel missing from the authorized set — or the whole workspace being
-   * gone — never blanks the stored selection.
+   * The rule's stored channels from the read model (id, name, privacy). They
+   * are merged into the options so a channel that is configured but not yet
+   * confirmed — what a same-workspace reinstall leaves behind — still renders
+   * by name and privacy instead of blanking the stored selection.
    */
   storedChannels: SlackChannelOption[];
   onValuesChange: (channelIds: string[]) => void;
 }
 
-/**
- * The authorized set is the pool; the stored selection is merged in so a
- * de-authorized channel stays visible and selected until the user removes it.
- */
+const toChannelOptions = (data: unknown): SlackChannelOption[] =>
+  Array.isArray(data)
+    ? (data as EligibleChannelResource[]).map((resource) => ({
+        id: resource.id,
+        name: resource.attributes.name,
+        is_private: resource.attributes.is_private,
+      }))
+    : [];
+
 const mergeOptions = (
-  authorized: SlackChannelOption[],
+  eligible: SlackChannelOption[],
   stored: SlackChannelOption[],
 ): SlackChannelOption[] => {
-  const byId = new Map(authorized.map((channel) => [channel.id, channel]));
+  const byId = new Map(eligible.map((channel) => [channel.id, channel]));
   stored.forEach((channel) => {
     if (!byId.has(channel.id)) byId.set(channel.id, channel);
   });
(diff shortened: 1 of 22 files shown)


pfe-nazaries:
@coderabbitai review

pfe-nazaries:

Reviewed 3 CodeRabbit feedback items and applied 1.

**Applied**

- *Do not make de-authorized channels selectable* — `ui/app/(prowler)/alerts/_components/slack-channels-field.tsx`

  A stored channel outside the eligible pool is now merged into the picker options only while it is still selected. Previously, on a `populated` (enabled-picker) state, deselecting a de-authorized stored channel left it offered, so it could be re-picked as a *newly added* id — which the write refuses with `slack_channel_not_eligible`. First render is unchanged, since the form's selection initialises to exactly the stored ids.

**Not applied**

- *Generate a UUID for each created rule* (`__tests__/msw/handlers/alerts.ts`) — no id validation exists in the alerts slice (`AlertRule.id` is a plain `string`); the same double already serves `seeded-rule`, `preview`, and `recipient-N`, and channel ids are deliberately Slack-shaped. No functional risk.
- *Verify the persisted destination state after a successful write* (`alerts-page.integration.test.tsx`) — the edit-flow tests already assert stored channels render in the UI. A create→reopen assertion would need a new harness mount path, since the create entry renders only the seed button on `/findings`.

**Files modified:**
- `ui/app/(prowler)/alerts/_components/slack-channels-field.tsx`

**Commit:** `a257ad36d`

The latest autofix changes are on the `feature/alerts-slack-channels` branch.

ui/app/(prowler)/alerts/_components/slack-channels-field.tsx
```diff
@@ -0,0 +1,310 @@
...
+        }),
+      ),
+    ])
+      .then(([channelsResult, integrationsResult]) => {
+        if (!channelsResult?.error) {
+          setChannelsReadable(true);
+          setEligibleChannels(toChannelOptions(channelsResult?.data));
+        }
+        // A failed read leaves the workspace unknown, not "no Slack at all".
+        if (integrationsResult?.error) return;
```
alejandrobailo, line 186: Could we keep a failed integration read distinct from a missing workspace? `getIntegrations()` catches a rejected `fetch` and returns `{ data: [] }`, so this branch turns that transient failure into `WORKSPACE.MISSING` and shows “needs a connected Slack workspace.” Please preserve an explicit error state and cover the case where the channel read succeeds but the integration read fails.

alejandrobailo (changes requested):
Good job overall! Could you please address the degraded-state handling in the inline comment before merging? Thanks!

ui/app/(prowler)/alerts/_components/slack-channels-field.tsx
pfe-nazaries: Fixed in 0e85bd426. Transport-level integration read failures now return an explicit error instead of an empty collection, so the alert form preserves the unknown workspace state. Added a regression test where the channel read succeeds while the integration read rejects.

alejandrobailo (dismissed):
Thanks for addressing the degraded-state feedback and adding the regression coverage! LGTM.

Alan-TheGentleman (commented):
Verified the network-failure fix at head e5704ff: `getIntegrations` now routes its catch through `handleApiError`, so a rejected fetch produces an `{error}` result and the channels field no longer misreports it as a missing workspace, and the new test ("does not claim the workspace is missing when its integration read rejects") covers exactly the requested scenario.

Two open CodeRabbit threads remain and would fold nicely into one cleanup commit: the `created-rule-${n}` fixture id is not UUID-shaped as the fixtures document, and `savedRuleChannels()` still asserts the outbound request body rather than a create-then-reopen round trip. One more small one while there: in `slack-channels.ts`, `return handleApiResponse(response)` inside the `try` is not awaited, so an HTTP 5xx throw bypasses the catch and rejects the action instead of collapsing into `{error}`; `return await` makes the try/catch real.

alejandrobailo (approved):

PR #12495, opened 2026-08-20
[CHAIN] feat(ui): add Registry permission support

Alan-TheGentleman:
## 🔗 Part of Chained PRs

| Field | Value |
|---|---|
| Feature Branch | `feat/prowler-2414-registry-ui` |
| Main PR | #12494 |
| Position | 1 of 8 |
| Base | `feat/prowler-2414-registry-ui` |
| Depends on | None |
| Follow-up | #12497, fresh Registry access authority |
| Review budget | 237 / 400 changed lines |
| Starts at | Draft Registry tracker with no feature behavior |
| Ends with | Cloud role and session models represent Registry permission safely |

### Chain Overview

```text
feat/prowler-2414-registry-ui (#12494 tracker)
└── 📍 #12495 PR 1: permissions and flag typing
    └── #12497 PR 2: fresh access authority
        └── ... PR 8: acceptance hardening
            └── #12494 tracker -> master
```

### Scope

- Includes: exact-boolean `manage_registry` propagation through Cloud roles, auth/session models, UI permission metadata, false defaults, and local `UI_REGISTRY_ENABLED` typing.
- Excludes: Registry navigation, routes, eligibility evaluation, runtime-config exposure, API calls, credential handling, catalog UI, backend changes, and deployment changes.

### Autonomy

- [x] Focused tests and exact results are recorded.
- [x] Runtime verification is not applicable because this slice exposes no Registry UI or route.
- [x] Rollback is the single child commit and excludes unrelated work.
- [x] The diff contains only this work unit.

### Context

The Registry UI must fail closed and remain hidden until later slices add fresh Cloud, runtime-flag, and permission checks. This first slice introduces only the permission and configuration representation that those guards will consume.

Keeping this foundation separate makes role/session behavior reviewable before any Registry route or API integration exists.

### Description

- Add `manage_registry` to the existing user, role, auth, and permission models with false defaults.
- Copy the permission from the current role only when its value is exactly `true`.
- Allow Cloud role create and edit flows to grant and retain the permission while non-Cloud payloads omit it.
- Add `UI_REGISTRY_ENABLED` environment typing without enabling or exposing Registry.
- Preserve existing Integrations, Billing, and generic navigation permission behavior.

No npm dependency, backend endpoint, migration, public runtime-config field, Registry route, or user-visible Registry workflow is introduced.

### Steps to review

1. Review the exact-boolean and false-default propagation across auth, roles, schemas, and permission display helpers.
[...]

PR #12494, opened 2026-08-20
feat(ui): complete Registry provider onboarding for Private Cloud

Alan-TheGentleman:
### Context

Private Cloud administrators can install an external Registry provider and complete the existing provider onboarding flow: **Registry → Providers → account details → credentials → connection → scan**.

This PR consolidates the implementation from #12495, #12497, #12499, #12503, #12504, #12505, #12515, #12520, #12524, #12579, #12590, #12646, #12676, and #12679. All 14 published child heads are preserved in this branch, which incorporates current `master`. GitHub automatically marked the children as merged when their commits reached this parent branch; #12494 remains open and has not been merged into `master`.

### Description

- Registry requires `UI_CLOUD_ENABLED=true`, `UI_REGISTRY_ENABLED=true`, and fresh `manage_registry` authority. Access is independent of billing; distribution defaults stay disabled.
- The complete catalog remains searchable and filterable. **Add** is restricted to `has_provider=true` and `is_builtin=false`, with eligibility rechecked against the catalog in the Server Action. Built-ins show **Built in**; checks/compliance-only packages have no Add action.
- Installation requires a successful task result and an authoritative tenant-artifact read before success. Shared task tracking survives navigation, hard reload, and browser history restoration. Removal keeps provider accounts and explains its effect on future connections/scans.
- Existing SearchInput, MultiSelect, Select, ClearFiltersButton, cards, alerts, dialogs, and toasts provide the catalog UI. Search, provider/capability filters, sorting, and tab selection persist in the URL.
- Add Provider discovers installed external provider slugs, excludes built-ins and duplicates, and shows a Registry badge with catalog metadata. Failed discovery does not block built-in providers. Server validation repeats discovery before creating a dynamic account; retries reuse existing accounts.
- The existing wizard renders supported flat credential schemas and authentication methods. Dynamic secrets use dedicated server validation and create/update actions; values stay in form memory and clear on completion, provider/method change, or unmount. Connection requires explicit `connected=true` before the existing scan step.
- Development Server Function argument logging is disabled to keep write-only credentials out of logs. Known Registry key rejections produce actionable messages without reflecting arbitrary backend data.

No backend endpoints or source changes, new npm dependencies, or merge into `master` are included. Integration requires the Cloud Registry/provider-schema/dynamic-scan API contract.

### Steps to review

[...]

PR #12497, opened 2026-08-20
[CHAIN] feat(ui): add fresh Registry access authority

Alan-TheGentleman:
## 🔗 Part of Chained PRs

| Field | Value |
|---|---|
| Feature Branch | `feat/prowler-2414-registry-ui` |
| Main PR | #12494 |
| Position | 2 of 8 |
| Base | `feat/prowler-2414-registry-ui-01-permissions` |
| Depends on | #12495 |
| Follow-up | #12499, Registry lease, navigation, and route guards |
| Review budget | 400 / 400 changed lines |
| Starts at | Registry permission and flag representation without authorization behavior |
| Ends with | Fresh fail-closed Registry access authority without exposing a route or UI |

### Chain Overview

```text
feat/prowler-2414-registry-ui (#12494 tracker)
└── #12495 PR 1: permissions and flag typing
    └── 📍 #12497 PR 2: fresh access authority
        └── #12499 PR 3: lease, navigation, and route guards
            └── ... PR 8: acceptance hardening
                └── #12494 tracker -> master
```

### Scope

- Includes: strict no-store current-user adaptation, exact-boolean permission evaluation, pure Registry access discriminants, and a server-only eligibility evaluator with a five-second timeout.
- Excludes: client leases, navigation, sidebar changes, proxy guards, Registry routes, Registry API calls, credential handling, catalog UI, backend changes, and deployment changes.

### Autonomy

- [x] Focused and full unit-test results are recorded.
- [x] Runtime verification is not applicable because this slice exposes no Registry UI or route.
- [x] Rollback is the single child commit and excludes unrelated work.
- [x] The diff contains only this work unit and is exactly 400 changed lines.

### Context

Registry authorization cannot trust stale JWT permissions or client-hydrated configuration. Before any route or navigation is exposed, the UI needs a fresh server-side authority that denies access when current-user evidence, runtime flags, or process state are missing, malformed, stale, or unavailable.

This slice establishes that boundary independently so later Registry UI work can consume one small, reviewable eligibility contract.

### Description

- Extract the existing `/users/me?include=roles` request into a strict current-user adapter using `cache: "no-store"`.
- Require one unambiguous role and adapt `manage_registry` only when its wire value is exactly `true`.
- Keep existing sign-in and tenant-switch behavior by delegating `getUserByMe` to the adapter.
- Add pure `eligible`, `ineligible`, and `unknown` Registry access results.
- Evaluate current process flags with exact `"true"` semantics and deny when the token or authority evidence is unavailable.
- Abort current-user evaluation after five seconds and never cache an allow result.
- Cover false, missing, malformed, ambiguous, non-200, network, abort, timeout, and stale-JWT denial cases.

No npm dependency, backend endpoint, migration, public runtime configuration, Registry route, or user-visible Registry workflow is introduced.

### Steps to review

1. Review `ui/lib/auth/current-user.ts` for strict JSON:API parsing, role ambiguity rejection, exact-boolean adaptation, and `cache: "no-store"`.
2. Confirm `ui/actions/auth/auth.ts` preserves existing callers by delegating to the current-user adapter.
3. Review `ui/lib/registry/access.ts` and `access.server.ts` for exact Cloud/feature/permission checks, five-second abort behavior, and fail-closed results.
4. Run `cd ui && pnpm exec vitest run --project unit lib/auth/current-user.test.ts lib/registry/access.test.ts lib/registry/access.server.test.ts actions/auth/auth.test.ts auth.config.test.ts`; the recorded result is 37 passing tests.
5. Run `cd ui && pnpm exec vitest run --project unit`; the recorded result is 3,071 passing tests.
6. Run `PREK_NO_CONCURRENCY=1 uv run --directory . prek run`; Prettier, ESLint, TypeScript, related tests, and root hooks are recorded as passing.

### Checklist

<details>

</details>

- [x] Review if is needed to change the [README.md](https://github.com/prowler-cloud/prowler/blob/master/README.md): not required for this internal authorization slice.
- [x] Ensure a changelog fragment is added under [prowler/changelog.d/](https://github.com/prowler-cloud/prowler/tree/master/prowler/changelog.d), if applicable: SDK/CLI is not changed.

#### SDK/CLI

#### UI
- [ ] All issue/task requirements work as expected on the UI: the complete user-facing workflow is delivered by later child PRs.
- [x] This PR adds or updates no npm dependencies.
- [x] Mobile screenshots/video are not applicable because this slice exposes no UI.
- [x] Tablet screenshots/video are not applicable because this slice exposes no UI.
- [x] Desktop screenshots/video are not applicable because this slice exposes no UI.
- [x] No UI changelog fragment is added for this internal foundation slice; the PR uses `no-changelog`, and the user-visible slice will add the feature fragment.

#### API
- [x] The API is not changed by this PR.
- [x] Endpoint output, query analysis, performance evidence, API specs, versions, and API changelog are not applicable.

#### MCP Server
- [x] The MCP Server is not changed by this PR.
- [x] The MCP changelog is not applicable.

### License

### DIFF
--- a/ui/lib/registry/access.server.ts
+++ b/ui/lib/registry/access.server.ts
@@ -1,0 +1,46 @@
+import "server-only";
+
+import { fetchCurrentUser } from "@/lib/auth/current-user";
+import { readEnv } from "@/lib/runtime-env";
+
+import {
+  isRegistryEligible,
+  REGISTRY_ACCESS,
+  registryAccessResult,
+  type RegistryAccessResult,
+} from "./access";
+
+const CURRENT_USER_TIMEOUT_MS = 5_000;
+
+const hasEnabledProcessFlags = () =>
+  readEnv("UI_CLOUD_ENABLED") === "true" &&
+  readEnv("UI_REGISTRY_ENABLED") === "true";
+
+export async function evaluateRegistryAccess(
+  accessToken?: string | null,
+): Promise<RegistryAccessResult> {
+  if (!hasEnabledProcessFlags() || !accessToken?.trim()) {
+    return registryAccessResult(REGISTRY_ACCESS.INELIGIBLE);
+  }
+
+  const controller = new AbortController();
+  const timeout = setTimeout(() => controller.abort(), CURRENT_USER_TIMEOUT_MS);
+
+  try {
+    const currentUser = await fetchCurrentUser(accessToken, {
+      signal: controller.signal,
+    });
+    if (currentUser.manageRegistry === undefined) {
+      return registryAccessResult(REGISTRY_ACCESS.UNKNOWN);
+    }
+    return registryAccessResult(
+      isRegistryEligible(true, true, currentUser.manageRegistry)
+        ? REGISTRY_ACCESS.ELIGIBLE
+        : REGISTRY_ACCESS.INELIGIBLE,
+    );
+  } catch {
+    return registryAccessResult(REGISTRY_ACCESS.UNKNOWN);
+  } finally {
+    clearTimeout(timeout);
+  }
+}
--- a/ui/actions/auth/auth.ts
+++ b/ui/actions/auth/auth.ts
@@ -4,6 +4,7 @@ import { AuthError } from "next-auth";
 
 import { signIn, signOut } from "@/auth.config";
 import { apiBaseUrl } from "@/lib";
+import { fetchCurrentUser } from "@/lib/auth/current-user";
 import { addAuthEvent } from "@/lib/sentry-breadcrumbs";
 import type { UtmParams } from "@/lib/utm";
 import type { SignInFormData, SignUpFormData } from "@/types";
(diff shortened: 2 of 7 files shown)


pfe-nazaries:
### Blocking: multi-role users cannot sign in

`fetchCurrentUser` throws `"Ambiguous current user role"` unless `included.length === 1` (`ui/lib/auth/current-user.ts:70-72`). The previous `getUserByMe` used `.find(type === "roles")` and tolerated several roles.

Multi-role is a supported and tested API capability:

- `Role.users` is a `ManyToManyField` through `UserRoleRelationship` — `api/src/backend/api/models.py:1417-1419`
- The only constraint is uniqueness of `(role_id, user_id)`, which does not cap distinct roles per user
- `api/src/backend/api/tests/test_views.py:9797-9816` PATCHes two roles onto one user and asserts `relationships.count() == 2` in the same tenant

**Impact.** `credentials.authorize` has no try/catch around `getUserByMe` (`ui/auth.config.ts:261`), so such a user is rejected with a valid password; the social-OAuth provider returns `null`; the tenant-switch `jwt` callback sets `TENANT_SWITCH_ERROR` and ejects to `/sign-in`. Existing JWT sessions are unaffected, so this surfaces at sign-in and tenant switch rather than on deploy.

Worth noting the asymmetry inside this PR: `evaluateRegistryAccess` wraps the same call in `catch { → unknown }` (`ui/lib/registry/access.server.ts:41-42`), so the Registry path degrades safely while the auth path does not.

**Suggested fix.** Keep `getUserByMe` tolerant (take the first role, mirroring the backend's own `get_role(...).first()`) and reserve the strict single-role requirement for the Registry authority.

---

**Question, non-blocking.** `ui/components/auth/oss/sign-in-form.tsx:147` exact-matches `result?.message === "User email is not verified"` to route to `/email-verification`. The only carrier of that exact string was the 403 branch's `errors[0].detail`, which this PR collapses to `"Unable to fetch current user"`. Inert in OSS (`ACCOUNT_EMAIL_VERIFICATION = "none"`), but does Prowler Cloud's API return that exact detail? If so, unverified Cloud users would get the generic error toast instead of the verification screen.

ui/lib/registry/access.server.ts
```diff
@@ -0,0 +1,46 @@
...
+  isRegistryEligible,
+  REGISTRY_ACCESS,
+  registryAccessResult,
+  type RegistryAccessResult,
+} from "./access";
+
+const CURRENT_USER_TIMEOUT_MS = 5_000;
+
+const hasEnabledProcessFlags = () =>
+  readEnv("UI_CLOUD_ENABLED") === "true" &&
```
pfe-nazaries, line 16: use `readBoolEnv`

pfe-nazaries (changes requested):

pfe-nazaries:
## `fetchCurrentUser` rejects multi-role users, and it breaks sign-in

Surfaced while reviewing `#12499` downstack. The code is here, so raising it here. This one was put through a dedicated adversarial pass whose job was to disprove it — four of five attack lines failed, and two backfired into stronger evidence.

`ui/lib/auth/current-user.ts:70-73` rejects any `/users/me?include=roles` document whose `included` doesn't hold exactly one role:

```ts
const [role] = parsed.data.included;
if (parsed.data.included.length !== 1 || !role) throw new Error("Ambiguous current user role");
```

**Prowler's API supports multiple roles per user, and resolves them as a union.** This isn't hypothetical:

- `api/models.py:1417` — `Role.users` is a `ManyToManyField` through `UserRoleRelationship`, whose constraint uniques on `("role_id", "user_id")`, not on `user_id`.
- `api/v1/serializers.py:507` — `UserRoleRelationshipSerializer.roles` is a `ListField` that bulk-creates every role in the list. No max-1 validation.
- `api/rbac/permissions.py:44` — permissions are explicitly a **union** over roles: `all(any(getattr(role, perm) for role in user_roles) ...)`.
- Checked-in passing test: `test_partial_update_relationship` PATCHes `user-roles-relationship` with two role identifiers, asserts `204` and `relationships.count() == 2`.
- `UserSerializer.get_roles` returns `instance.roles.all()` **unfiltered by tenant**, so `included` also holds >1 entry for a user who belongs to two tenants with one role each — no multi-role assignment needed at all.

So the throw invents an ambiguity the backend doesn't have.

**It's a regression, not an inherited assumption.** `master`'s `getUserByMe` did `parsedResponse.included?.find(item => item.type === "roles")` — first role wins, never throws. This PR replaces that tolerant read with a hard throw.

### Two consequences, in severity order

**1. Affected users cannot sign in at all.** `ui/auth.config.ts:261`, inside the credentials `authorize()`:

```ts
const userMeResponse = await getUserByMe(tokenResponse.accessToken);
const user = tokenUserFromApi(userMeResponse);
```

No `try`/`catch` — the throw propagates out of `authorize()`. The sibling call sites are guarded (social OAuth at `:287` returns `null`; the tenant-switch `jwt` path at `:345` sets `TENANT_SWITCH_ERROR`), which makes this one look like an oversight rather than a decision. This half has nothing to do with Registry.

**2. Registry denies them silently.** `evaluateRegistryAccess` swallows the throw in a bare `catch` and returns `UNKNOWN`, which every consumer treats as deny. Indistinguishable from a transient outage, so it'll be misdiagnosed as flakiness.

### Scope, honestly

Prowler's own UI assigns exactly one role everywhere (`updateUserRole` PATCHes a single identifier, the invitation form is a single-value `Select`, the users table renders one `role?.name`), so you can't mint a 2-role user through the product. Reachability is via the JSON:API directly, or an already multi-role/multi-tenant account. Not the default population — but a supported, tested API state.

Worth noting `ui/lib/auth/current-user.test.ts:61-73` currently *enshrines* this: `document([role(true), role(true)])` is an explicit expected rejection. If the fix lands, that case wants inverting.

### Suggested fix

Mirror the backend and OR the permission across `included` — or fall back to the first role as `master` did — and reserve the throw for genuinely zero roles. Fail-closed is the right instinct, but it doesn't apply when the permission *is* determinable.

alejandrobailo:
Consolidated into #12494. All changes are preserved there. GitHub marked this PR as merged when its commits were pushed to the parent branch; #12494 remains open and has not been merged into master.

PR #12525, opened 2026-08-24
fix(gcp): describe BigQuery datasets and tables through a bounded thread pool

### LINKED ISSUE #12502: GCP: BigQuery service does one tables.get() per table, making scans of large data estates appear hung
### Summary

`BigQuery._get_tables()` issues one synchronous `tables().get()` per table. On a project with a real data estate this turns a GCP scan into a multi-hour job that looks, from the API and the UI, like a hung scan rather than a slow one.

Measured on Prowler **5.36.0** (API 1.37.0), self-hosted, scanning two GCP projects with the same service account:

| Project | BigQuery tables | Resources | Scan duration |
| --- | --- | --- | --- |
| A | ~13,400 | 12,679 | **5,851 s (97 min)** |
| B | ~660 | 827 | 554 s (9 min) |

Scan time tracks table count almost linearly. Project A spent roughly 80 of its 97 minutes inside the BigQuery service, during which the scan reported **`progress = 4` and `unique_resource_count = 0`** and wrote no findings.

### Where it comes from

`prowler/providers/gcp/services/bigquery/bigquery_service.py`

```python
62  def _get_tables(self):
63      for dataset in self.datasets:
65          request = self.client.tables().list(
66              projectId=dataset.project_id, datasetId=dataset.name
67          )
68          while request is not None:
69              response = request.execute(num_retries=DEFAULT_RETRY_ATTEMPTS)
71              for table in response.get("tables", []):
72                  cmk_encryption = False
73                  if (
74                      self.client.tables()
75                      .get(                       # <-- one round trip per table
76                          projectId=dataset.project_id,
77                          datasetId=dataset.name,
78                          tableId=table["tableReference"]["tableId"],
79                      )
80                      .execute(num_retries=DEFAULT_RETRY_ATTEMPTS)
```

`tables.list` is already paginated efficiently — listing all ~13,400 tables takes about 2.5 s. The cost is entirely the per-table `get`, serialised.

Two things make the impact larger than it first looks:

1. **It runs in the constructor.** `_get_tables()` is called from `BigQuery.__init__` (`bigquery_service.py:16`), so the full enumeration is paid whenever the BigQuery service is instantiated — including for a run that only wants `bigquery_dataset_public_access` and never touches table-level data.
2. **Progress is per-check, so the scan looks frozen.** One check performing ~13,400 round trips leaves the reported progress pinned for over an hour. There is no way for an operator to distinguish this from a genuinely wedged worker without inspecting `/proc/<pid>/io` on the worker to confirm bytes are still moving.

`_get_datasets()` has the same shape (`tables` → `datasets`, `bigquery_service.py:26-31`), but dataset counts are small enough that it is not a practical problem.

### Impact

Beyond wall-clock: `bigquery_table_cmk_encryption` emits one finding per table, so a single control decision produced **12,207 FAIL findings** on project A — 94% of that scan's total. That is arithmetically correct but makes the finding count a poor proxy for the amount of work a user actually faces.

### Suggested directions

- **Parallelise the per-table `get`** with a bounded thread pool, as was done in #9608 for SageMaker tag listing. This is the smallest change and would cut the dominant cost by whatever concurrency is chosen.
- **Or batch the requests** — `googleapiclient`'s `BatchHttpRequest` would collapse the round trips substantially.
- **Consider deferring table enumeration** so it is not paid by dataset-only checks, e.g. lazily populating `self.tables` on first access rather than in `__init__`.
- **Weight scan progress by resources rather than checks**, so a long-running check reports movement instead of appearing stalled. This is the part that turns a slow scan into a support ticket.

I'm happy to open a PR for the parallelisation if that's the direction you'd prefer.

### Reproduction

Scan a GCP project containing a few thousand BigQuery tables and observe scan duration against a comparable project with few tables. No special configuration is required; the service account needs only read access.

### Environment

- Prowler 5.36.0, Prowler API 1.37.0, self-hosted (Docker Compose)
- GCP provider, service account key credentials

kkmookhey:
### Context

Fix #12502.

`BigQuery._get_tables` issues one `tables.get()` per table, serially, from the service constructor. On a project with a large BigQuery estate this dominates the whole GCP scan.

Measured on two real projects scanned with the same service account:

| BigQuery tables | Resources | Scan duration |
| --- | --- | --- |
| ~13,400 | 12,679 | **5,851 s (97 min)** |
| ~660 | 827 | 554 s (9 min) |

For the larger project roughly 80 of those 97 minutes were spent inside the BigQuery service, during which the scan reported `progress = 4`, `unique_resource_count = 0` and wrote no findings — indistinguishable from a hung worker without inspecting `/proc/<pid>/io` on the host.

`_get_datasets` has the same shape at smaller N.

### Description

Listing is cheap (all ~13,400 table refs list in about 2.5 s) and stays serial. The per-resource `get()` calls now run through a bounded `ThreadPoolExecutor`.

Three details worth calling out for review:

**One global pool, not a pool per dataset.** Parallelising per dataset and then again within each dataset would multiply the ceiling by the dataset count — 31 datasets × 10 workers is 310 concurrent requests, which puts the BigQuery API quota at risk. A single pool over all `(dataset, table)` pairs keeps the ceiling at `MAX_WORKERS`.

**Output order is preserved.** `executor.map` yields in input order, so `self.datasets` and `self.tables` are still built in listing order. This keeps the existing assertions in `bigquery_service_test.py` (`datasets[0]`, `tables[1]`, …) valid — they are unchanged in this PR — which matters because `pytest-randomly` is in the test stack.

**Each worker holds its own `AuthorizedHttp`.** `googleapiclient`'s http object is not thread safe, so a threaded call cannot share the client's. `BigQuery` was the only GCP service not already using `__get_AuthorizedHttp_client__`. It is cached in `threading.local` rather than built per request: the existing helper constructs a fresh `httplib2.Http()` on every call, so building one per request would mean a TLS handshake per table — 13,400 of them — and no connection reuse. Per thread keeps it at `MAX_WORKERS`.

**One behaviour change, stated explicitly:** previously a failed `tables.get` aborted the remaining tables in that dataset, because a single `try` wrapped the whole loop. A table or dataset that cannot be read is now skipped and logged, and its siblings still get enumerated. This is more robust but it *is* a change, and `test_one_unreadable_table_does_not_drop_the_rest` pins it.

### Steps to review

1. `uv run pytest tests/providers/gcp/services/bigquery/` — 13 pass.
2. `uv run pytest -n auto tests/providers/gcp/` — 662 pass.
3. Two of the three new tests fail against the pre-change service, which is the point of them:
   - `test_http_clients_are_per_thread_not_per_request` fails with *"no per-thread http was created"* — the old code passes no `http` at all.
   - `test_one_unreadable_table_does_not_drop_the_rest` fails — the old code drops every table after the failing one.
   - `test_table_order_follows_listing_order` **passes** against the old code. It is a regression guard against the concurrency introduced here, not a proof of the fix.

### Scope

This addresses the enumeration half of #12502. The other half — scan progress being weighted per check, so one check doing ~13,400 round trips pins the reported progress for over an hour — is untouched here and would be a separate change. Worth leaving the issue open, or splitting it, rather than closing on this PR alone.

### Checklist

- [x] This feature/issue is listed in the open issues (#12502)
- [x] Is it assigned to me
- [x] I have reviewed the open pull requests and confirmed there is no existing PR that implements the same outcome
- [x] Review if code is being documented following the Google Python style guide
- [ ] Review if backport is needed — deferring to maintainers
- [x] Review if is needed to change the README.md — not needed
- [x] Ensure a changelog fragment is added under `prowler/changelog.d/`

#### SDK/CLI
- Are there new checks included in this PR? **No** — no new checks, and no provider permission changes. The same `bigquery.tables.get` / `bigquery.datasets.get` permissions are used, only concurrently.

## Summary by CodeRabbit

* **Performance**
  * BigQuery dataset and table discovery is now faster through bounded concurrent processing.

* **Reliability**
  * Unreadable datasets or tables are skipped while accessible resources continue to be processed.
  * Results retain the original listing order.
  * Partial access issues no longer interrupt discovery.

* **Bug Fixes**
  * Improved handling of access failures during BigQuery enumeration.
  * Scans continue and return available resource details when individual resources cannot be read.


kkmookhey:
All four review comments are addressed, and the automated review is now approving:

- Google-style docstrings on `_get_thread_http`, `_get_datasets` and `_get_tables` (92a2150)
- One executor for the whole enumeration instead of one per project, so the thread-local
  `AuthorizedHttp` count stays at `MAX_WORKERS` rather than scaling with project count (5f85831)
- `executor.map` collects its input eagerly, so detail requests are submitted in fixed
  batches — a large estate no longer holds tens of thousands of pending futures at once (cb1b2e7)
- The isolation test now blocks two concurrent detail requests and asserts the workers receive
  *distinct* http objects, rather than only asserting an upper bound on how many were created (cb1b2e7)

All checks are green. The branch is behind master but nothing else has touched
`services/bigquery/`, so it merges cleanly — happy to rebase if you would prefer it current.

Is there anything else you would like changed here, or someone I should tag for review?

PR #12518, opened 2026-08-24
fix(api): document Azure certificate credentials

HugoPBrito:
### Context

Depends on #12421

Do not merge until #12421 has merged and the Azure documentation/template asset has been deployed.

This PR intentionally targets `master` rather than the feature branch so the public contract change is visible but not stacked. Draft status is the merge guard.

### Description

Depends on #12421

Do not merge until #12421 has merged and the Azure documentation/template asset has been deployed.

This PR updates the public OpenAPI contract to document the two exclusive Azure credential shapes: client secret credentials or certificate content credentials. It also adds regression coverage that verifies both schemas expose only their required fields and reject additional properties.

The approved API changelog fragment is included. The committed `v1.yaml` is the byte-for-byte output of the canonical `GET /api/v1/schema` endpoint on current `master`; this also reconciles schema drift already emitted by current master code but not yet reflected in the committed artifact.

This PR intentionally targets `master` rather than the feature branch so it remains visible without being stacked; draft status prevents merge while the dependency is outstanding.

### Steps to review

1. Confirm the branch diff against `master` contains only the four intended API schema, serializer, test, and changelog paths.
2. Review the Azure `oneOf` schemas for mutually exclusive `client_secret` and `certificate_content` credential shapes.
3. Review the focused regression test and changelog fragment.
4. Re-run the focused verification commands below.

```bash
cd api
uv run pytest src/backend/api/tests/test_serializers.py::TestProviderSecretFieldSchema::test_azure_schema_exposes_exclusive_supported_credential_shapes -q
```

Outcome: `1 passed, 2 warnings`.

```bash
cd api
PYTHONPATH=src/backend DJANGO_SETTINGS_MODULE=config.django.testing uv run python -c 'import django; django.setup(); from pathlib import Path; from rest_framework.test import APIClient; response = APIClient().get("/api/v1/schema", HTTP_HOST="localhost"); assert response.status_code == 200, response.status_code; generated = response.content; committed = Path("src/backend/api/specs/v1.yaml").read_bytes(); assert generated == committed, "endpoint schema differs from committed v1.yaml"; print(f"canonical-byte-match: yes ({len(generated)} bytes)")'
```

Outcome: `canonical-byte-match: yes (840213 bytes)`. The generator also emitted the existing repository schema warnings; no strict schema-validation pass is claimed.

```bash
[...]

PR #12526, opened 2026-08-25
feat(azure): add certificate authentication to Azure SDK

lydiavilchez:
### Context

Part of [PROWLER-2378](https://prowlerpro.atlassian.net/browse/PROWLER-2378): Deploy-to-Azure quick-start with certificate-based authentication for the Azure provider.

This is the **first PR of a stack of three**:

1. **This PR** — SDK changes (Azure provider, CLI flags, key-pair validation, tests)
2. #12527 — API changes (serializer, validator, OpenAPI schema — closes #12518)
3. #12528 — UI wizard, Bicep template, and documentation

### Description

Add certificate-based Service Principal authentication to the Azure provider, mirroring the M365 provider flow.

**SDK changes:**
- `AzureProvider` accepts a certificate credential (base64 content or file path) and authenticates via `azure.identity.CertificateCredential`
- New `prowler/providers/azure/lib/certificate.py` — validates PEM and PKCS#12 bundles, rejects malformed input or key-only files
- CLI flags: `--certificate-auth`, `--certificate-content`, `--certificate-path`
- Env var support: `AZURE_CLIENT_ID`, `AZURE_TENANT_ID`, `AZURE_CERTIFICATE_CONTENT`
- `test_connection` extended to accept the new certificate parameters (keyword-only)

### Steps to review

1. Verify the certificate authentication code path in `prowler/providers/azure/azure_provider.py` and helper module `prowler/providers/azure/lib/certificate.py`
2. Confirm CLI flags in `prowler/providers/azure/lib/arguments/arguments.py`
3. Run the unit tests:
   ```console
   cd tests/providers/azure && pytest azure_provider_test.py -v
   ```
4. Try authentication locally:
   ```console
   export AZURE_CLIENT_ID="..."
   export AZURE_TENANT_ID="..."
   export AZURE_CERTIFICATE_CONTENT="<base64 PEM bundle>"
   prowler azure --certificate-auth
   ```

### Checklist

- [x] Unit tests added and passing
- [x] Changelog fragment added (`prowler/changelog.d/azure-certificate-authentication.added.md`)
- [x] No breaking changes to existing authentication paths

[PROWLER-2378]: https://prowlerpro.atlassian.net/browse/PROWLER-2378?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Added Azure service principal authentication using certificates.
  * Certificates can be supplied as base64-encoded content, file paths, or supported environment variables.
  * Added the `--certificate-auth` and `--certificate-path` command-line options.
  * Azure identity details now include certificate thumbprints.

* **Bug Fixes**
  * Added validation and clearer errors for invalid, missing, or mismatched certificate bundles.
  * Enforced valid authentication option combinations.
  * Improved Azure connection and credential verification, including more reliable token requests.

### DIFF
--- a/prowler/providers/azure/azure_provider.py
+++ b/prowler/providers/azure/azure_provider.py
@@ -111,6 +111,20 @@ def _find_transport_cause(error: BaseException) -> Optional[BaseException]:
     return None
 
 
+def _decode_certificate_content(certificate_content: str) -> bytes:
+    """Base64-decode certificate content, tolerating shell-added whitespace.
+
+    ``AZURE_CERTIFICATE_CONTENT`` values almost always arrive with a
+    trailing newline (shell command substitution) or embedded line
+    breaks (``openssl base64`` wraps output at 64 columns by default).
+    ``base64.b64decode(..., validate=True)`` rejects both as
+    ``binascii.Error``. Strip whitespace before decoding so the strict
+    validation still rejects non-base64 characters without penalising the
+    common copy-paste and env-var flows.
+    """
+    return base64.b64decode("".join(certificate_content.split()), validate=True)
+
+
 def _build_certificate_credential(
     tenant_id: str,
     client_id: str,
@@ -417,9 +431,9 @@ class AzureProvider(Provider):
             sp_env_auth,
             browser_auth,
             managed_identity_auth,
-            certificate_auth,
             subscription_ids,
             client_id,
+            certificate_auth=certificate_auth,
         )
 
         # TODO: should we keep this here or within the identity?
@@ -813,8 +827,8 @@ class AzureProvider(Provider):
                             with open(certificate_path, "rb") as cert_file:
                                 certificate_data = cert_file.read()
                         else:
-                            certificate_data = base64.b64decode(
-                                getenv("AZURE_CERTIFICATE_CONTENT"), validate=True
+                            certificate_data = _decode_certificate_content(
+                                getenv("AZURE_CERTIFICATE_CONTENT")
                             )
                         # Same fail-fast validation the static path runs. The
                         # returned bundle is normalized so the leaf appears
@@ -1134,13 +1148,20 @@ class AzureProvider(Provider):
             if raise_on_exception:
                 raise credential_unavailable_error
             return Connection(error=credential_unavailable_error)
-        except AzureDefaultAzureCredentialError as default_credentials_error:
+        except AzureNotValidCertificateContentError as certificate_content_error:
             logger.error(
-                f"{default_credentials_error.__class__.__name__}[{default_credentials_error.__traceback__.tb_lineno}]: {default_credentials_error}"
+                f"{certificate_content_error.__class__.__name__}[{certificate_content_error.__traceback__.tb_lineno}]: {certificate_content_error}"
             )
             if raise_on_exception:
-                raise default_credentials_error
-            return Connection(error=default_credentials_error)
+                raise certificate_content_error
+            return Connection(error=certificate_content_error)
+        except AzureNotValidCertificatePathError as certificate_path_error:
+            logger.error(
+                f"{certificate_path_error.__class__.__name__}[{certificate_path_error.__traceback__.tb_lineno}]: {certificate_path_error}"
+            )
+            if raise_on_exception:
+                raise certificate_path_error
+            return Connection(error=certificate_path_error)
         except (
             AzureClientIdAndClientSecretNotBelongingToTenantIdError
         ) as tenant_id_error:
@@ -1261,9 +1282,13 @@ class AzureProvider(Provider):
         sp_env_auth,
         browser_auth,
         managed_identity_auth,
-        certificate_auth,
         subscription_ids,
         client_id,
+        # Keyword-only so external callers using the pre-cert-auth positional
+        # layout (`..., managed_identity_auth, subscription_ids, client_id`)
+        # keep binding those slots correctly.
+        *,
+        certificate_auth: bool = False,
     ):
         """
         Sets up the identity for the Azure provider.
@@ -1643,7 +1668,7 @@ class AzureProvider(Provider):
                 # exception several call frames deeper if this fails, which
                 # makes for a bad UX in the API/UI.
                 normalized_bundle = validate_certificate_bundle(
-                    base64.b64decode(certificate_content, validate=True)
+                    _decode_certificate_content(certificate_content)
                 )
                 # Persist the normalized (leaf-first) bundle so downstream
                 # `CertificateCredential` calls don't pick an intermediate CA.
@@ -1821,7 +1846,7 @@ class AzureProvider(Provider):
         transport = None
         try:
             if certificate_content:
-                certificate_data = base64.b64decode(certificate_content, validate=True)
+                certificate_data = _decode_certificate_content(certificate_content)
             elif certificate_path:
                 with open(certificate_path, "rb") as cert_file:
                     certificate_data = cert_file.read()
@@ -1855,15 +1880,17 @@ class AzureProvider(Provider):
                 f"{error.__class__.__name__}[{error.__traceback__.tb_lineno}] -- {error}"
             )
             # `azure.identity` wraps `_request_token` with `wrap_exceptions`,
-            # so a `RequestsTransport` connect/read timeout reaches this
+            # so a `RequestsTransport` connect/read failure reaches this
             # handler as `ClientAuthenticationError`. Distinguish transport
             # failures (credentials unavailable) from a genuinely rejected
-            # certificate by walking the cause chain.
+            # certificate by walking the cause chain. The transport error
+            # covers DNS, TLS, connection reset and timeout indistinctly, so
+            # the message must not claim the request timed out.
             if _find_transport_cause(error) is not None:
                 raise AzureCredentialsUnavailableError(
                     file=os.path.basename(__file__),
                     message=(
-                        "Timed out waiting for Entra ID to issue a token "
+                        "Entra ID was not reachable while acquiring a token "
                         "for the provided certificate."
                     ),
                     original_exception=error,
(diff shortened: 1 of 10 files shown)


prowler/providers/azure/azure_provider.py
```diff
@@ -55,12 +66,97 @@
...
+
+
+def _build_certificate_credential(
+    tenant_id: str,
+    client_id: str,
+    certificate_data: bytes,
+    authority: Optional[str],
+) -> "CertificateCredential":
+    """Build a `CertificateCredential` and remember its SHA-1 thumbprint."""
+    credentials = CertificateCredential(
```
HugoPBrito, line 135: I believe we need to normalize the PEM bundle before passing it to `CertificateCredential`. Our validator accepts a bundle where an intermediate certificate appears before the matching leaf, but `azure-identity 1.21.0` calculates the credential thumbprint from the first PEM certificate. This means the new intermediate-first test passes while the real authentication still fails. Can we place the matching leaf first and cover the actual `CertificateCredential` path?

prowler/providers/azure/azure_provider.py
```diff
@@ -524,15 +700,48 @@ def setup_session(
                         f"{environment_credentials_error.__class__.__name__}[{environment_credentials_error.__traceback__.tb_lineno}] -- {environment_credentials_error}"
                     )
                     raise environment_credentials_error
+            elif certificate_auth and not azure_credentials:
+                # Env vars are only required for the pure env-var flow; skip
+                # this check when azure_credentials already carries the material.
+                try:
+                    AzureProvider.check_certificate_creds_env_vars(
```
HugoPBrito, line 759: Can we allow an explicit `--tenant-id` to replace `AZURE_TENANT_ID` here? `check_certificate_creds_env_vars` always requires the environment variable before `setup_session` reaches the `tenant_id or getenv(...)` fallback. As a result, certificate authentication fails even when the tenant was explicitly provided. I believe we should only require `AZURE_TENANT_ID` when no tenant argument was passed, and add a regression test for that case.

prowler/providers/azure/azure_provider.py
```diff
@@ -1285,49 +1676,148 @@ def verify_client(
...
+            data = {
+                "grant_type": "client_credentials",
+                "client_id": client_id,
+                "client_secret": client_secret,
+                "scope": region_config.graph_scope,
+            }
+            # Hard timeout so a hung Entra ID endpoint cannot stall the
+            # calling worker (this runs on request threads and Celery tasks
+            # in the API path).
+            response = requests.post(
```
HugoPBrito: I believe the new timeout should be translated into a provider-specific exception. `requests.exceptions.Timeout` currently falls through the generic handler, so `test_connection` and API consumers can receive a raw transport error instead of a consistent Azure error. Can we map it to `AzureCredentialsUnavailableError` and cover the timeout path?

HugoPBrito (changes requested):
Thanks for working through this, Lydia! The overall direction looks good and the test coverage is substantial. I found a few remaining cases that can make valid certificate authentication fail or expose inconsistent errors, so I believe we should address them before merging.

prowler/providers/azure/azure_provider.py
HugoPBrito: Could we remove `AZURE_TENANT_ID` from the test environment instead of setting a stale value?

The current case proves that the explicit argument takes precedence, but it does not reproduce the original failure where the environment variable is missing.

prowler/providers/azure/azure_provider.py
HugoPBrito: Please add coverage for `requests.exceptions.Timeout`, including the `raise_on_exception=False` path.

That protects the API contract and ensures consumers continue receiving `AzureCredentialsUnavailableError` instead of the raw transport exception.

prowler/providers/azure/azure_provider.py
HugoPBrito: The current test only verifies the normalized helper output. It would still pass if a call site accidentally sent the original bundle to `CertificateCredential`.

I believe this should cover the real credential construction path as well.

HugoPBrito (changes requested):
Thanks for the updates, Lydia! The implementation now addresses the three behaviors I raised.

I believe we still need the regression coverage requested in those threads. The background token request and positional `test_connection` compatibility also remain unresolved, so I’m keeping this as changes requested for now.

prowler/providers/azure/azure_provider.py
```diff
@@ -1285,49 +1694,161 @@ def verify_client(
...
+            # with a hard deadline (matches the 30s on the client-secret path).
+            # Manage the executor explicitly: the context-manager form calls
+            # shutdown(wait=True) on exit and would block until get_token
+            # returns anyway, defeating the timeout.
+            executor = ThreadPoolExecutor(max_workers=1)
+            future = executor.submit(credential.get_token, region_config.graph_scope)
+            try:
+                future.result(timeout=_TOKEN_ACQUISITION_TIMEOUT_SECONDS)
+            except FuturesTimeoutError as error:
+                executor.shutdown(wait=False)
```
HugoPBrito: CodeRabbit originally raised this because `future.result(timeout=...)` stops waiting without cancelling `credential.get_token`. `shutdown(wait=False)` lets the caller return, but the worker and network request remain active. Non-timeout exceptions may also bypass executor shutdown.

In the API and Celery paths, concurrent validations during an Entra ID or network degradation can accumulate background workers. Please enforce the timeout at the transport or SDK layer and guarantee cleanup on every exit. The new tests cover error mapping, not termination of the underlying operation.

HugoPBrito (changes requested):
Thanks for the updates, Lydia! The previous review items are now addressed.

One reliability issue remains. CodeRabbit originally raised it in [this thread](https://github.com/prowler-cloud/prowler/pull/12526#discussion_r3853185577), but it was marked as addressed while the implementation still stops only the caller’s wait, not the underlying token request.

prowler/providers/azure/azure_provider.py
```diff
@@ -125,6 +224,9 @@ def __init__(
         mutelist_content: dict = None,
         client_id: str = None,
         client_secret: str = None,
+        certificate_auth: bool = False,
+        certificate_content: str = None,
+        certificate_path: str = None,
         resource_groups: list = [],
```
HugoPBrito: I believe the new certificate parameters need to be keyword-only without moving the existing `resource_groups` positional slot.

In the base version, callers could pass `resource_groups` immediately after `client_secret`. The current signature binds that value to `certificate_auth` instead, potentially selecting the wrong authentication flow.

This preserves the previous contract while keeping the new parameters explicit:

```suggestion
        resource_groups: list = [],
        *,
        certificate_auth: bool = False,
        certificate_content: str = None,
        certificate_path: str = None,
```

Please add an `inspect.signature(...).bind(...)` regression using the previous positional call shape. It should also assert that all three certificate parameters are `KEYWORD_ONLY`.

prowler/providers/azure/azure_provider.py
```diff
@@ -1186,6 +1536,8 @@ def validate_static_credentials(
         tenant_id: str = None,
         client_id: str = None,
         client_secret: str = None,
+        certificate_content: str = None,
+        certificate_path: str = None,
         region_config: AzureRegionConfig = None,
```
HugoPBrito: The same positional compatibility issue affects `validate_static_credentials`.

`region_config` was previously the fourth positional parameter. Moving it behind the certificate fields means an existing caller can accidentally pass an `AzureRegionConfig` object as `certificate_content`.

The compatible signature is:

```suggestion
        region_config: AzureRegionConfig = None,
        *,
        certificate_content: str = None,
        certificate_path: str = None,
```

A signature-binding regression should cover this method and the already-restored `test_connection(..., provider_id, *, certificate_*)` layout.

prowler/providers/azure/azure_provider.py
```diff
@@ -1285,49 +1697,159 @@ def verify_client(
...
+                certificate_data=validate_certificate_bundle(certificate_data),
+                authority=region_config.authority,
+                transport=RequestsTransport(
+                    connection_timeout=_TOKEN_ACQUISITION_TIMEOUT_SECONDS,
+                    read_timeout=_TOKEN_ACQUISITION_TIMEOUT_SECONDS,
+                ),
+            )
+            try:
+                credential.get_token(region_config.graph_scope)
+            except ServiceRequestError as error:
```
HugoPBrito: I believe this still handles the exception one abstraction layer too low.

`CertificateCredential.get_token()` does not expose `ServiceRequestError` directly. Azure Identity decorates `_request_token()` with `wrap_exceptions`, which raises `ClientAuthenticationError` and preserves the Azure Core transport failure in its cause chain. Read timeouts also use `ServiceResponseTimeoutError`, which is not a `ServiceRequestError`.

Consequently, a real connection or read timeout bypasses this handler and is reported as an invalid certificate.

The implementation should:

1. Build the transient `RequestsTransport` separately.
2. Pass `retry_total=0` to `CertificateCredential`; otherwise Azure Core can repeat each 30-second connect/read timeout.
3. Catch `ClientAuthenticationError`.
4. Traverse `__cause__` and `__context__`.
5. Translate `ServiceRequestError` or `ServiceResponseError` causes to `AzureCredentialsUnavailableError`.
6. Preserve certificate content/path errors when no transport cause exists.
7. Close the credential in `finally`.
8. Log and suppress cleanup failures so `close()` cannot replace the primary typed exception.

Also, please describe these values as connect/read transport limits rather than a hard 30-second wall-clock deadline.

tests/providers/azure/azure_provider_test.py
```diff
@@ -1224,3 +1245,1632 @@ def test_setup_identity_succeeds_without_active_event_loop(self):
...
+
+        from azure.core.exceptions import ServiceRequestError
+
+        with (
+            patch(
+                "prowler.providers.azure.azure_provider.CertificateCredential"
+            ) as cert_credential_cls,
+            pytest.raises(AzureCredentialsUnavailableError),
+        ):
+            cert_credential_cls.return_value.get_token.side_effect = (
```
HugoPBrito: These tests mock `CertificateCredential.get_token()` and inject `ServiceRequestError` directly. That bypasses the production behavior responsible for this bug: Azure Identity exception wrapping, Azure Core retry policy, request/read timeout conversion, and transport cleanup.

Please keep a real `CertificateCredential` and inject a controlled `RequestsTransport` backed by a mocked session:

- Raise `requests.ConnectTimeout` and `requests.ReadTimeout` from `session.request`.
- Assert the resulting Prowler error is `AzureCredentialsUnavailableError`.
- Assert `session.request.call_count == 1`, proving retries are disabled.
- Assert the credential or transport is closed.
- Cover `test_connection(..., raise_on_exception=False)`.
- Add a cleanup-failure case proving `close()` cannot mask the original typed error.

This verifies the same pipeline production executes, without making real network calls or relying on wall-clock sleeps.

HugoPBrito (changes requested):
Thank you for replacing the executor-based timeout. I reviewed the latest implementation against the actual `azure-identity 1.21.0` and `azure-core 1.41.0` behavior. The worker leak is gone, but positional compatibility, wrapped transport failures, retry behavior, and deterministic cleanup still need adjustment before this is safe to merge. The inline notes include concrete implementation and regression-test guidance.

HugoPBrito (dismissed):

PR #12527, opened 2026-08-25
feat(api): accept Azure certificate credentials and document the contract

lydiavilchez:
### Context

Part of [PROWLER-2378](https://prowlerpro.atlassian.net/browse/PROWLER-2378): Deploy-to-Azure quick-start with certificate-based authentication.

Second PR of a stack of three:

1. #12526 — SDK changes
2. **This PR** — API changes (serializer, validator, OpenAPI schema)
3. #12528 — UI wizard, Bicep template, and documentation

**Consolidates and closes #12518** (`fix(api): document Azure certificate credentials`). The OpenAPI schema update that lived there is now part of this PR so both changes ship together.

### Description

Extend the API to accept Azure certificate credentials alongside the existing client-secret flow, and document the new contract in the public OpenAPI spec.

**Serializer** (`AzureProviderSecret`):
- Accepts an optional `certificate_content` field
- Enforces mutual exclusion with `client_secret` — providing both or neither returns a validation error

**Validator** (`api/src/backend/api/validators.py` — new):
- Validates the base64-encoded certificate + private key bundle
- Rejects malformed input, key-only files, and mismatched certificate/key pairs
- Supports PEM and unencrypted PKCS#12/PFX formats

**OpenAPI schema** (`api/src/backend/api/specs/v1.yaml`):
- Documents the two exclusive Azure credential shapes so external API consumers know what to send
- Adds `oneOf` discriminator for `client_secret` vs `certificate_content`

### Steps to review

1. Verify the mutual exclusion logic in `AzureProviderSecret.validate` (`api/src/backend/api/v1/serializers.py`)
2. Confirm the key-pair validator handles the edge cases (see `test_validators.py`)
3. Run the tests:
   ```console
   cd api/src/backend && pytest api/tests/test_serializers.py::TestAzureProviderSecret api/tests/test_validators.py -v
   ```
4. Inspect the OpenAPI diff in `api/src/backend/api/specs/v1.yaml` to confirm the new schema documents both credential shapes

### Checklist

- [x] Unit tests for serializer and validator
- [x] Changelog fragments added (`api/changelog.d/azure-certificate-secret.added.md`, `api/changelog.d/azure-certificate-openapi-schema.fixed.md`)
- [x] OpenAPI schema updated to reflect the new contract
- [x] Backwards compatible with existing client-secret credentials

[PROWLER-2378]: https://prowlerpro.atlassian.net/browse/PROWLER-2378?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/src/backend/api/v1/serializers.py
```diff
@@ -1785,10 +1789,77 @@ class Meta:
...
+        return super().validate(attrs)
+
+    def validate_certificate_content(self, certificate_content):
+        """Validate the Azure certificate and matching private-key bundle."""
+        if certificate_content:
+            # Tolerate whitespace inside the base64 payload: exports from
+            # Windows terminals (CRLF) or wrapped copy-paste survive without
+            # tripping `base64.b64decode(validate=True)`, and the reader is
+            # base64 anyway — internal whitespace carries no information.
+            certificate_content = "".join(certificate_content.split())
```
HugoPBrito, line 1832: I believe we are discarding the normalized Azure secret here.

`AzureProviderSecret` strips whitespace from `certificate_content`, but the outer write path only assigns `validated_secret` back for OCI. M365 follows the same outer flow, but its validator returns the input unchanged, so it does not expose the same issue.

Can we assign `validated_secret` back to `validated_attrs["secret"]` for every provider and add create/update tests that reload the stored secret?

api/src/backend/api/v1/serializers.py
```diff
@@ -1800,16 +1871,20 @@ class M365ProviderSecret(serializers.Serializer):
     tenant_id = serializers.CharField()
     user = serializers.EmailField(required=False)
     password = serializers.CharField(required=False)
-    certificate_content = serializers.CharField(required=False)
+    certificate_content = serializers.CharField(
```
HugoPBrito: I believe this change should remain scoped to Azure.

This introduces a new 50 KiB decoded bundle limit for M365. I could not find an equivalent limit documented by Microsoft, and neither the previous API serializer nor the M365 SDK enforced one. A valid certificate bundle can exceed this threshold, so this becomes a breaking change outside the scope of the Azure stack.

Can we remove the M365 restriction from this PR? If M365 also needs this guardrail, please explain why it is required and why 50 KiB is a safe threshold, then introduce it separately with regression coverage, documentation, and a changelog entry.

HugoPBrito (changes requested):
Thanks for the work on this PR. The Azure certificate flow looks good overall, but I left two inline comments around persistence and an unrelated M365 contract change that should either stay out of this stack or be properly justified. Please take a look when you can.

api/src/backend/api/v1/serializers.py
```diff
@@ -2092,8 +2163,11 @@ def validate(self, attrs):
         validated_secret = self.validate_secret_based_on_provider(
             provider.provider, secret_type, secret
         )
-        if provider.provider == Provider.ProviderChoices.ORACLECLOUD.value:
-            validated_attrs["secret"] = validated_secret
+        # Persist the per-provider validator's normalized output (e.g.
+        # `AzureProviderSecret` strips whitespace inside `certificate_content`).
+        # Previously this reassignment only happened for OCI, so every
+        # per-provider validator's normalization was discarded on write.
+        validated_attrs["secret"] = validated_secret
```
HugoPBrito: I believe this applies nested serializer normalization to every provider, not only Azure.

DRF `CharField` strips leading and trailing whitespace by default. Persisting the complete `validated_secret` can therefore modify valid opaque credentials such as M365 passwords, registry passwords, tokens, and OCI passphrases. Previously, the outer `JSONField` preserved the submitted values and only OCI deliberately persisted its normalized result.

Can we keep the existing OCI behavior, replace only Azure’s `certificate_content` with its normalized value, and preserve the original values for the remaining providers? Please apply the same logic to update and add POST/PATCH regressions using whitespace-significant M365 and registry credentials.

HugoPBrito (changes requested):
Thanks for the work on these updates. The previous comments are resolved, but I found one regression in the new persistence path and left an inline comment. Please take a look when you can.

PR #12531, opened 2026-08-25
feat(mcp): classify shared tool failures and stop relaying upstream bodies

puchy22:
> **Stacked PR 1 of 4.** Base: `master`. Merge this one first; the three on top of it convert one sub-server each.
>
> | | PR | What it does |
> |---|---|---|
> | 1 | **this one** | the failure types, the classifier, the middleware, the docs |
> | 2 | #12532 | Prowler App tools (8 modules + auth) |
> | 3 | #12533 | Prowler Hub tools |
> | 4 | #12534 | Prowler documentation tools |
>
> Each one is green on its own (`make test-mcp`), so they can merge bottom-up as they are approved.

### Context

All 100 tools in `mcp_server/` reported failure as a *successful* result: `isError: false` carrying an `{"error": ...}` object, in six competing shapes. The MCP spec reserves that shape for a tool that succeeded, so clients and models read a failure as an answer and act on it. On top of that, ten call sites interpolated the raw upstream response body into the text an agent reads (`e.response.text`), which is how a gateway error page or a debug traceback ends up replayed into a model's context.

This PR lands the mechanism. It converts no tools; the three stacked PRs on top of it do that, one sub-server at a time.

### Description

**`prowler_mcp_server/lib/errors.py` (new)** — the failure vocabulary and the classifier:

- `ProwlerAPIError` carries `status_code` and `detail`, where `detail` is Prowler's own JSON:API `errors[0].detail` — text our API wrote for a caller to read, and therefore the only upstream text this server relays. It is `None` when the body was not JSON:API, which is exactly the case that must never be repeated.
- `ProwlerAPIUnreachable` is a separate type on purpose: a rejected write changed nothing and can be retried freely, while an unanswered write may well have been applied. Collapsing the two is how a retry duplicates a write.
- `InvalidArgument(ValueError)` marks a message as one *we* wrote. `JSONDecodeError` and pydantic's `ValidationError` are both `ValueError` subclasses whose messages quote their input — and these tools take cloud credentials as arguments, so a bare `except ValueError` cannot be trusted to be safe to echo.
- `_describe_failure` gives a sentence only to the failures **every** tool shares: 401, 403, 429, 5xx, other 4xx, unreachable, our own argument validation, unparseable JSON, pydantic rejection. Anything else returns `None` and stays masked. A 404 is a different object and a different lookup tool for each of the 100, so it stays at its call site.
- `SharedFailureMiddleware` installs it. It keys on `__cause__`: an absent cause means the tool raised a `ToolError` deliberately without a `from` clause, and that message is already the final word.

**`server.py`** — `mask_error_details=True` and the middleware on the parent. Masking does *not* reach mounted children (FastMCP warns about this at mount time), so each sub-server sets it in its own PR; middleware does reach them, so it is wired once here.

**`prowler_app/utils/api_client.py`** — raises the new types, logs the full body and keeps it out of the exception message, and promotes local validation (`validate_page_size`, `normalize_date_range`, `_validate_date_format`, the `fetch_external_url` domain checks) to `InvalidArgument`.

**Docs** — the developer guide's Error Handling section, the `AGENTS.md` checklist and the skill checklist now say raise, never return, and show the three ways to raise.

### Steps to review

1. Read `lib/errors.py` top to bottom — it is the whole contract, 200 lines.
2. Check the inclusion rule holds: every branch of `_describe_failure` is a sentence that is correct for *any* tool that can reach it.
3. Confirm nothing untrusted survives: `detail` is the only upstream text relayed, and pydantic's rejected value is stripped.
4. Run the suite:

```
make test-mcp
```

131 passing. New coverage in `tests/lib/test_errors.py`: the classification table, the two provenance properties (a non-JSON:API body is never repeated, a pydantic value is never echoed), and end-to-end assertions that a failing tool answers `isError: true` with no `structuredContent`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Tool failures are now reported as errors instead of successful responses.
  * Error messages provide clearer explanations and recommended corrective actions.
  * Invalid arguments are rejected before requests are sent.
  * Upstream response bodies are no longer exposed; sensitive details remain in server logs.
  * Network outages and common API failures now receive consistent, user-friendly messages.

* **Documentation**
  * Updated developer and QA guidance for exception-based MCP tool error handling.
  * Added changelog entries describing improved failure messages and protected upstream errors.


mcp_server/prowler_mcp_server/server.py
```diff
@@ -2,9 +2,17 @@
...
 
 from prowler_mcp_server import __version__
+from prowler_mcp_server.lib.errors import SharedFailureMiddleware
 from prowler_mcp_server.lib.logger import logger
 
-prowler_mcp_server = FastMCP("prowler-mcp-server")
+# `mask_error_details` keeps an unhandled failure from relaying text this server
+# does not control. It is set on every sub-server as well, because it does not
+# reach mounted children -- FastMCP warns about exactly that at mount time.
+prowler_mcp_server = FastMCP("prowler-mcp-server", mask_error_details=True)
```
josema-xyz, line 11: This setting does not reach the child servers, and they still use the default value. I reproduced an internal error reaching the client with its original text. Could we enable error masking on all three child servers and test this case?

mcp_server/prowler_mcp_server/prowler_app/utils/api_client.py
```diff
@@ -112,18 +105,30 @@ async def _make_request(
...
+                "HTTP error during %s %s: %s %s",
+                method.value,
+                path,
+                status,
+                (e.response.text or "")[:500],
             )
+
+            message = f"API request failed: {status}"
+            if detail:
+                message = f"{message} - {detail}"
```
josema-xyz, line 119: For `5xx` errors, this adds the API detail to the exception text. Some current tools return `str(e`) directly, so the detail can bypass the shared error handler. Could we keep `5xx` details out of the exception text, or confirm they are always safe to expose?

mcp_server/prowler_mcp_server/lib/errors.py
```diff
@@ -0,0 +1,222 @@
...
+
+    if isinstance(exc, ValidationError):
+        problems = [
+            f"{'.'.join(str(part) for part in error['loc']) or '(argument)'}: {error['msg']}"
+            for error in exc.errors(include_url=False)
+        ]
+        # Field and expectation only: pydantic quotes the rejected value back.
+        return f"Invalid arguments -- {'; '.join(problems)}."
+
+    if isinstance(exc, json.JSONDecodeError):
```
josema-xyz, line 172: A `JSONDecodeError` can also come from an invalid Prowler API response, not only from a tool argument. In that case, this message wrongly blames the user. Could we distinguish between these two cases?

mcp_server/prowler_mcp_server/server.py
puchy22: I was thinking in doing it in each subsequent PR

mcp_server/prowler_mcp_server/prowler_app/utils/api_client.py
puchy22: Fixed in 9489c2c

mcp_server/prowler_mcp_server/lib/errors.py
puchy22: Fixed in 9489c2c

mcp_server/prowler_mcp_server/lib/errors.py
```diff
@@ -0,0 +1,239 @@
...
+    if isinstance(exc, ProwlerAPIError):
+        return _describe_prowler_api_error(exc)
+
+    if isinstance(exc, ProwlerAPIInvalidResponse):
+        # Nothing about the call was wrong, so it must not be described as if it
+        # were -- and the body that could not be parsed stays in the log.
+        return (
+            "Prowler answered with a body this server could not read. The call "
+            "itself was accepted, so retrying it unchanged is safe."
+        )
```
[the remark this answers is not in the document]
puchy22: Fixed in 2e08d3c

josema-xyz (approved):

danibarranqueroo (approved):

lydiavilchez (approved):

pedrooot (approved):

PR #12532, opened 2026-08-25
feat(mcp): raise instead of returning error objects in the Prowler App tools

puchy22:
> **Stacked PR 2 of 4.** The base, #12531, is merged, so this PR now targets `master` and the GitHub diff is the whole change.

### Context

#12531 installed the failure vocabulary (`lib/errors.py`) and the classifier middleware. This PR converts the Prowler App sub-server — 11 tool modules, the auth helper and the API client — to it, and then fixes the outcomes the conversion made visible: once a failure is an error, a result that says `status: "failed"` is a bug, and several of them were reporting the wrong thing.

### Description

**1. A failure leaves the tool as an error.**

The dominant shape was a whole tool body wrapped in `try: ... except Exception as e: self.logger.error(...); return {"error": str(e), "status": "failed"}`. That handler caught the failure, logged it, and then reported it to the client as a success. Everything it caught is now either classified by `lib/errors.py` or, when nothing useful can be said about it, masked — `mask_error_details=True` on `prowler_app/server.py` is the backstop. Deleting the wrappers is most of this diff, which is why the line count on the tool modules is close to a wash.

Hand-written sentences survive where they carry a fact the classifier cannot know, raised as `ToolError` *without* a `from` clause — which is how `SharedFailureMiddleware` recognises an authored message and leaves it alone: `get_finding_group_details` names the check ID and points at `prowler_list_finding_groups`, `_get_latest_scan_id_for_provider` points at `prowler_trigger_scan`, `get_mutelist` points at `prowler_set_mutelist`, and the integration validation messages name the precondition they enforce.

Local argument rejections become `InvalidArgument` so their text reaches the agent verbatim: the "either `scan_id` or `provider_id`" pairs in `compliance.py`, the payload and credential validation in `integrations.py`, and the ones in `muting.py` and `providers.py`.

Also fixes `list_attack_paths_queries`, which returned `[{"error": ...}]` from a list-returning tool — an error object shaped like a query, in a list a caller iterates.

**2. An auth failure is a credential failure, not an argument failure.**

`utils/auth.py` rejects seven things and none of them is about an argument the model chose, so they raise a new `CredentialError`: the answer is a credential the user has to fix, not another attempt with different arguments. With it, `_parse_jwt` rejects a payload that is not a JSON object (a list or a scalar decodes just as cleanly and blew up later as an `AttributeError`), a missing or non-numeric `exp` makes the token unusable rather than being compared, the `Bearer` scheme is matched case-insensitively per RFC 7235 and stripped as a prefix only, and an unsupported `PROWLER_MCP_TRANSPORT_MODE` is a `RuntimeError` — nothing about a call caused it and nothing about a call can fix it, so it stays masked for the model and logged for whoever runs the server.

**3. A blank argument is rejected by name, before the request goes out.**

"required" alone does not stop `""`. A model with no scan or query ID to hand sends an empty string rather than omitting the argument, and it travelled into a URL path or a request body: `prowler_run_attack_paths_query` with an empty `query_id` came back as Prowler's `This field may not be blank`, which names no field, leaving nothing to tell `scan_id` from `query_id`. `NonBlankStr` (`lib/types.py`) strips the whitespace and rejects what is left, as `minLength` on the type rather than a check inside each tool, so the constraint is in the tool schema for a client to see before it calls. Applied across the App, Hub and documentation tools to every required string argument and to the optional values that overwrite a stored one; filter arguments keep plain `str`, since a blank filter matches everything. A test walks the registered schemas so the next tool cannot be written without the guard.

**4. The three Attack Paths 404s are told apart.**

The endpoint answers 404 for three different things and the tools relayed all of them as the same bare `Not found.`:

- **A query that matched nothing** is a 404 carrying the empty result as its body. That is an answer, and the good one — this account has no attack path of that shape — so it is returned rather than raised, and a correct call is no longer reported as a failure that invites a retry. The empty graph serializes to `{}` through the minimal-serializer mixin, which reads as an answer gone missing, so it carries a sentence saying the query ran and found nothing.
- **A scan ID that does not resolve** now says an Attack Paths scan is a separate resource from a regular scan, that only AWS providers have one, and which tool returns a usable ID.
- **A scan whose graph records no Cartography schema** is told apart from a scan that does not exist, and points at the ready-made queries that do not need the schema.

`ProwlerAPIError` grew a `payload` so a tool can tell an error status holding a result from a request that actually failed. It is read structurally and never relayed as text; `detail` stays the only part of an upstream body that may reach a model.

**5. The write tools report what the write actually did.**

- `prowler_trigger_scan` returned `status: "failed"` inside a successful tool result. It now leaves a scan that was not created as an error, and a scan that was created but could not be read back hands over the ID and says not to trigger it again.
- `prowler_delete_provider` reported a deletion still running as a failure. Prowler removes a provider together with its scans, findings and resources in a background task that routinely outlives the 60-second polling window, so that was wrong twice over: it is not a failure, and it invites a retry of a destructive call already in flight. It now reports `in_progress` with the task ID, and only a task that actually stopped is an error, naming what may already be gone.
- `prowler_schedule_daily_scan` read the state of the first scan run, which Prowler starts alongside the schedule, as if it were the outcome of the scheduling itself, so a schedule that exists was reported as a failure whose suggested retry could only hit a 409. The schedule is committed inside the request that creates it, so an answer means it exists; the first run's state is reported as what it is.
- `prowler_connect_provider` reported a connection test it could not run as `failed`, blaming the provider's credentials for an expired Prowler credential, a rate limit or a timeout. It now reports `not_tested`. Relatedly, a failed secret lookup no longer passes as "this provider has no secret", which sent the caller down the create branch and had a refused second secret reported as rejected credentials.

**6. The success flags that could only ever be true are gone.**

`status`, `success`, `deleted`, `scheduled`: now that a failure is a tool execution error, a flag with one reachable value says nothing, and inviting a reader to branch on it suggests there is a failure shape to look for in the result, which there is not. `prowler_delete_mute_rule` was the clearest case — it branched on `result.get("success")` and reported "Failed to delete mute rule" when a 204 came back with no body, firing on the shape of the answer rather than on anything having gone wrong. The confirmation stays in the message: deletion is irreversible, so it is worth saying plainly that it happened, and that the findings a rule had muted stay muted.

**Deliberately unchanged:** the Jira dispatch helpers (`_jira_dispatch_rejected` / `_unknown`) still return a result object. A dispatch that stopped halfway created some work items and Prowler cannot delete them, so `safe_to_retry: false` is a fact about the world, not an error. Squashing it into an exception loses the only thing that stops a retry from duplicating the write.

**Changelog:** the nine fragments this PR grew are now a single one covering the whole idea, `mcp-app-tool-failures.added.md`.

### Steps to review

1. Spot-check that no deleted handler took a fact with it — every removed `except` block either re-wrapped or is replaced by a `ToolError` above.
2. Confirm the `ToolError`s that must survive masking are raised without `from`, and the ones that must not are.
3. `tests/lib/test_types.py` walks every registered tool schema for `minLength: 1` on required strings; check the exemptions are the ones you would exempt.
4. New coverage for the outcomes above: `test_providers.py`, `test_scans.py`, `test_attack_paths.py`, `test_muting.py`, `test_auth.py`. In `test_integrations.py` each assertion moved from `result.data["error"]` to `pytest.raises`, and the "nothing was written" assertions (`mock_router.paths()`) are kept.
5. Run the suite:

```
make test-mcp
```

184 passing.

## Summary by CodeRabbit

* **New Features**
  * Added clearer MCP tool error responses with actionable guidance.
  * Added credential-specific authentication errors for invalid or expired tokens.
  * Added detailed status reporting for provider deletions and scheduled scans.
  * Empty Attack Paths results now include an explanatory message.

* **Bug Fixes**
  * Improved handling of missing resources, invalid identifiers, API errors, and authentication failures.
  * Blank or whitespace-only inputs are now rejected consistently.
  * Improved detection and reporting of scan, integration, muting, and provider operation outcomes.


mcp_server/prowler_mcp_server/prowler_app/tools/scans.py
```diff
@@ -191,60 +193,67 @@ async def trigger_scan(
...
+            # The scan may well have been queued, so this must not read as
+            # "nothing happened" and invite a duplicate run. No `from` clause:
+            # this names the provider and the tool that checks for the scan,
+            # neither of which the shared classifier can know.
+            raise ToolError(
+                "Prowler accepted the scan but did not return its ID, so it "
+                "cannot be looked up. Use prowler_list_scans for provider "
+                f"{provider_id} to see whether a scan is already running before "
+                "triggering another one."
+            )
```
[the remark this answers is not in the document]
puchy22: Read the AGENTS.md file from `mcp_server`, now we're reporting all errors through proper raise error mechanisms, not as the result of the tool. It was changed in PR #12531

mcp_server/prowler_mcp_server/prowler_app/models/attack_paths.py
```diff
@@ -368,7 +376,7 @@ def from_api_response(
         Returns:
             AttackPathQueryResult with parsed data and summary
         """
-        attributes = response.get("data", {}).get("attributes")
+        attributes = response.get("data", {}).get("attributes", {})
         nodes_data = attributes.get("nodes", [])
         relationships_data = attributes.get("relationships", [])
```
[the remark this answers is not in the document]
puchy22: Fixed in 6baf352

mcp_server/prowler_mcp_server/prowler_app/tools/providers.py
```diff
@@ -300,33 +304,106 @@ async def delete_provider(
...
+                f"running (task state '{state}'), which is normal for a provider with "
+                "many scans and findings. Do not send the deletion again. Use "
+                "prowler_search_providers to check whether it is gone."
+            )
+
+        return ProviderDeletionResult(
+            status="in_progress",
+            task_id=task_id,
+            message=message,
+        ).model_dump()
```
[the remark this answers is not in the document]
josema-xyz: +1. If the deletion finishes just after polling times out, the final read returns `completed`, but this still reports `in_progress`. This path also needs a test.

mcp_server/prowler_mcp_server/prowler_app/tools/roles.py
```diff
@@ -167,10 +169,13 @@ async def set_user_role(
...
         except Exception as e:
-            raise ValueError(
+            # No `from` clause: this says what state the user was left in, which
+            # the shared classifier cannot know, and a cause would let it replace
+            # this message with its own.
+            raise ToolError(
                 f"Role {role_id} could not be read ({e}), so user {user_id} was left "
                 f"unchanged. Use `prowler_list_roles` to find a valid role ID."
-            ) from e
+            )
```
josema-xyz, line 185: This catches more than a missing role. A permission error, timeout, rate limit, or server error is reported as an invalid role ID. The custom message should only be used for a real not-found response.

mcp_server/prowler_mcp_server/prowler_app/tools/compliance.py
```diff
@@ -295,19 +295,14 @@ async def get_compliance_framework_state_details(
...
         # Resolve provider_id to latest scan_id if needed
         resolved_scan_id = scan_id
         if not scan_id and provider_id:
-            try:
-                resolved_scan_id = await self._get_latest_scan_id_for_provider(
-                    provider_id
-                )
-            except ValueError as e:
-                return {"error": str(e)}
+            resolved_scan_id = await self._get_latest_scan_id_for_provider(provider_id)
```
josema-xyz, line 311: This method accepts both IDs, then silently ignores `provider_id`. That can return data for a different provider than the caller requested. It should reject calls that provide both values.

mcp_server/prowler_mcp_server/prowler_app/tools/scans.py
```diff
@@ -191,60 +193,67 @@ async def trigger_scan(
...
         except Exception as e:
-            self.logger.error(f"Scan creation failed: {e}")
-            return ScanCreationResult(
-                scan=None,
-                status="failed",
-                message=f"Scan creation failed: {str(e)}",
-            ).model_dump()
+            raise ToolError(
+                f"Scan {scan_id} was created for provider {provider_id}, but reading "
+                f"its state failed: {e} Use prowler_get_scan with that ID to monitor "
```
josema-xyz: This sends the original error text to the client, so error masking is bypassed. Can we keep the scan ID and next step without including `{e}`? The same pattern appears in `integrations.py`, `roles.py`, and `providers.py`.

mcp_server/prowler_mcp_server/prowler_app/utils/auth.py
```diff
@@ -76,14 +82,16 @@ async def authenticate(self) -> str:
...
             if not authorization_header:
-                raise ValueError("No authorization header provided")
-
-            # Extract token from Bearer header
-            if authorization_header.startswith("Bearer "):
-                token = authorization_header.replace("Bearer ", "")
-            else:
-                raise ValueError(
-                    "Invalid authorization header format. Expected 'Bearer <token>'"
+                raise CredentialError("No Authorization header was sent")
```
josema-xyz, line 85: `CredentialError` now falls into Jira’s `unknown` path, even though authentication fails before the request is sent. Can we treat it as rejected instead?

mcp_server/prowler_mcp_server/prowler_app/tools/providers.py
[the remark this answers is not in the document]
puchy22: Fixed in c7271bbe4

mcp_server/prowler_mcp_server/prowler_app/tools/roles.py
puchy22: Fixed in 879855b77

mcp_server/prowler_mcp_server/prowler_app/tools/compliance.py
puchy22: Fixed in 40f5cce08

mcp_server/prowler_mcp_server/prowler_app/tools/scans.py
puchy22: Fixed in f41d61d2c

mcp_server/prowler_mcp_server/prowler_app/utils/auth.py
puchy22: Fixed in b1e85aef3 — it is raised rather than reported as `rejected`: nothing was queued, so there is no dispatch outcome to describe, and the shared classifier says which credential to fix, which no retry of this call can.

josema-xyz (approved):

pedrooot (approved):

HugoPBrito (approved):

PR #12567, opened 2026-08-27
feat(azure): add aks_cluster_uses_a_supported_version check

### LINKED ISSUE #12566: [New Check]: aks_cluster_uses_a_supported_version
### Existing check search

### Provider

Azure

### New provider name

_No response_

### Service or product area

aks

### Suggested check name

aks_cluster_uses_a_supported_version

### Context and goal

AKS gives community support to the latest GA minor version and the two before it. Once a version drops out of that window it stops getting security patches. 
Prowler has no check for this on Azure. AWS already has the equivalent for EKS (`eks_cluster_uses_a_supported_version`), so this is a coverage gap rather than a new idea.

### Expected behavior

Compare each cluster's Kubernetes minor version against a configurable baseline, following the EKS check's pattern.

PASS: the cluster runs a version at or above the baseline
FAIL: the cluster runs a version below it

Baseline configurable via aks_cluster_oldest_version_supported, defaulting to 1.34 
The supported window today is 1.34/1.35/1.36, and 1.33 reached end of life in July 2026. Configurable because long-term support keeps patching older versions (1.31 until Nov 2026), so teams on LTS can lower it rather than see false positives.

Clusters reporting no version, or an unparseable one, are skipped rather than failed. Note the AKS service already fetches kubernetes_version and discards it, so the Cluster model needs the field added.

### References

- https://learn.microsoft.com/en-us/azure/aks/supported-kubernetes-versions
- https://learn.microsoft.com/en-us/azure/aks/long-term-support
- az aks get-versions --location canadacentral -o table returns only 1.34.x / 1.35.x / 1.36.x
- Existing AWS analogue: prowler/providers/aws/services/eks/eks_cluster_uses_a_supported_version/

### Suggested severity

High

_No response_

pouria-l:
### Context

Fix #12566

AKS gives community support to the latest GA minor version and the two before it. Once a version leaves that window it stops receiving security patches — the N-3 platform support tier explicitly excludes "applying security patches" and "applying bug fixes" — so a cluster keeps running and scaling while accumulating unpatched CVEs. Prowler already has `eks_cluster_uses_a_supported_version` for AWS but had no equivalent on Azure.

### Description

**New check:** `aks_cluster_uses_a_supported_version`

- **Severity:** `high` (matches the EKS check)
- **Service:** `aks`
- **Resource group:** `container`
- **Category:** `vulnerabilities`
- **Configurable:** `aks_cluster_oldest_version_supported`, default `1.34`
- No new dependencies

| Cluster state | Result |
|---|---|
| Minor version at or above the baseline | PASS |
| Minor version below the baseline | FAIL |
| No version reported | skipped, no finding |
| Version that will not parse | skipped, logged as an error |

**Why `1.34` as the default.** AKS supports N, N-1 and N-2. With 1.36 as the current GA that puts 1.34 at the bottom of the window, and 1.33 reached end of life in July 2026. `az aks get-versions --location canadacentral` returns only 1.34.x, 1.35.x and 1.36.x, which matches the published calendar. It is configurable because long-term support keeps patching older versions — 1.31 until November 2026 — so teams on LTS can lower the baseline instead of seeing false positives. As with the EKS default, this value needs revisiting as the support window moves; that is called out in the check metadata `Notes`.

**Only major and minor are compared.** AKS reports three-part versions such as `1.34.9`, and patch level does not affect support status. Versions are compared as integer tuples rather than strings, so `1.9` correctly sorts below `1.34`.

**Files changed:**

- `prowler/providers/azure/services/aks/aks_cluster_uses_a_supported_version/` — check, metadata, `__init__`
- `prowler/providers/azure/services/aks/aks_service.py` — adds `kubernetes_version` to the `Cluster` model. The service was already reading this value from the SDK to decide whether to include a cluster at all and then discarding it, so nothing new is fetched.
- `prowler/config/config.yaml` and `prowler/config/schema/azure.py` — the configurable baseline, reusing the existing dotted-version validator
- `docs/user-guide/cli/tutorials/configuration_file.mdx` — documents the new variable
- `prowler/changelog.d/aks-cluster-uses-a-supported-version.added.md` — changelog fragment
- `tests/providers/azure/services/aks/` — 14 new tests

### Steps to review

1. Run the AKS suite:

   ```
   uv run pytest tests/providers/azure/services/aks/ -v
   ```

   60 tests pass, 14 of them new: PASS and FAIL across two- and three-part version strings, a lowered baseline, the unconfigured fallback, and skips for missing and unparseable versions.

2. Confirm the check is registered:

   ```
   uv run python prowler-cli.py azure --list-checks | grep aks_cluster_uses_a_supported_version
   ```

3. `tests/config/` (584 tests) and `tests/lib/check/` (3880 tests) also pass, and the shipped `config.yaml` round-trips through `AzureProviderConfig` with no validation warnings.

I also ran it against a live subscription with `--az-cli-auth`: three AKS clusters, one on 1.33.7 reported FAIL and two on 1.34.9 reported PASS, with the messages rendering as expected. That confirms `kubernetes_version` comes back populated from the real SDK as a three-part string, which the mocked tests cannot establish on their own.

### Checklist

<details>

- [x] This feature/issue is listed in the [open issues](https://github.com/prowler-cloud/prowler/issues?q=sort%3Aupdated-desc+is%3Aissue+is%3Aopen) or roadmap.prowler.com
- [x] Is it assigned to me, if not, request it via the [open issues](https://github.com/prowler-cloud/prowler/issues?q=sort%3Aupdated-desc+is%3Aissue+is%3Aopen) or [Prowler Community Slack](https://goto.prowler.com/slack)

</details>

- [x] Review if backport is needed. No — this is a new check rather than a fix.
- [x] Review if is needed to change the [README.md](https://github.com/prowler-cloud/prowler/blob/master/README.md) No change needed.

#### SDK/CLI
    - If so, do we need to update permissions for the provider? **No** — the check reads `kubernetes_version` from the same `managed_clusters.list` response the existing AKS checks already consume, which the built-in Reader role covers. `permissions/prowler-azure-custom-role.json` only lists actions Reader does not grant, so it is unchanged.

### License

## Summary by CodeRabbit

* **New Features**
  * Added an Azure AKS check to identify clusters running Kubernetes versions below the supported baseline.
  * Added configurable minimum AKS version support, defaulting to Kubernetes 1.34.
  * Included remediation guidance for updating unsupported cluster versions.

* **Documentation**
  * Updated configuration documentation and examples with the new AKS version setting.
  * Added changelog information for the new check.

* **Tests**
  * Added coverage for supported, unsupported, missing, invalid, and custom-baseline versions.

### DIFF
--- a/docs/user-guide/cli/tutorials/configuration_file.mdx
+++ b/docs/user-guide/cli/tutorials/configuration_file.mdx
@@ -710,8 +710,9 @@ azure:
   # Azure Kubernetes Service
   # azure.aks_cluster_uses_a_supported_version
   # AKS clusters must run this Kubernetes minor version or higher. AKS gives
-  # community support to the latest GA minor version and the two before it.
-  # Lower this value if your clusters are enrolled in long-term support.
+  # community support to the latest generally available (GA) minor version and
+  # the two before it. Lower this value if your clusters are enrolled in
+  # long-term support.
   aks_cluster_oldest_version_supported: "1.34"
 
   # Azure App Service
--- a/docs/user-guide/cli/tutorials/configuration_file.mdx
+++ b/docs/user-guide/cli/tutorials/configuration_file.mdx
@@ -207,6 +207,7 @@ The following list includes all the Azure checks with configurable variables tha
 
 | Check Name                                               | Value                                           | Type            | Default                                                    |
 |----------------------------------------------------------|-------------------------------------------------|-----------------|------------------------------------------------------------|
+| `aks_cluster_uses_a_supported_version`                   | `aks_cluster_oldest_version_supported`          | String          | `"1.34"`                                                   |
 | `apim_threat_detection_llm_jacking`                      | `apim_threat_detection_llm_jacking_actions`     | List of Strings | See `config.yaml`                                          |
 | `apim_threat_detection_llm_jacking`                      | `apim_threat_detection_llm_jacking_minutes`     | Integer         | `1440`                                                     |
 | `apim_threat_detection_llm_jacking`                      | `apim_threat_detection_llm_jacking_threshold`   | Float           | `0.1`                                                      |
@@ -705,6 +706,13 @@ azure:
   # azure.network_public_ip_shodan
   # TODO: create common config
   shodan_api_key: null
+
+  # Azure Kubernetes Service
+  # azure.aks_cluster_uses_a_supported_version
+  # AKS clusters must run this Kubernetes minor version or higher. AKS gives
+  # community support to the latest GA minor version and the two before it.
+  # Lower this value if your clusters are enrolled in long-term support.
+  aks_cluster_oldest_version_supported: "1.34"
 
   # Azure App Service
   # azure.app_ensure_php_version_is_latest
--- a/prowler/changelog.d/aks-cluster-uses-a-supported-version.added.md
+++ b/prowler/changelog.d/aks-cluster-uses-a-supported-version.added.md
@@ -1,0 +1,1 @@
+`aks_cluster_uses_a_supported_version` checks that every AKS cluster runs a Kubernetes minor version at or above the configured baseline, which defaults to the oldest version in AKS community support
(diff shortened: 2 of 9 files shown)


pouria-l:
All fixed except the test rename.

The null baseline was a real crash. `aks_cluster_oldest_version_supported: null` gets past `audit_config.get(key, default)` because the key exists, so it hit `None.split(".")`. It falls back to the default now, and the schema uses `make_dotted_version_validator(2, 2)` the way `aws.py` does for the EKS equivalent.

Left `test_no_subscriptions` as-is, since that's what the other AKS check tests use and I couldn't find `test_no_resources` anywhere in the Azure tests. Happy to change it if I've missed a convention.

PR #12572, opened 2026-08-27
fix(ui): complete Slack OAuth callback server-side to avoid router race

pfe-nazaries:
### Context

The Slack OAuth callback was the only OAuth callback in this repo handled client-side: a client component exchanged the single-use `code` through a Server Action and then navigated away with `router.replace()`. That navigation rides the App Router action queue, where a concurrent Server Action completing during the navigation can cancel it (upstream race: vercel/next.js#88343, fix not yet released) — leaving the user stuck on the callback screen even though the workspace was already connected. The single-use OAuth code also transited hydrated client code and the session history.

### Description

The callback becomes a server-side Route Handler at the same URL (`/integrations/slack/callback`), aligning Slack with the pattern the GitHub/Google/SAML OAuth callbacks already use (`ui/app/api/auth/callback/*`):

- The handler exchanges the code server-side and answers an HTTP `303` to `/integrations/slack`, with the outcome carried in a tokens-only query-string contract (`ui/lib/integrations/slack-connect-status.ts`) — validated on write and on read, never free text in the URL.
- A notice on the integration page re-homes the callback's copy and strips its params via the History API. Refusals the API names by `code` keep Prowler's wording; the code-less `400` (state consumed / code spent — the back-button path) gets its own `expired` status and copy.
- The single-use code no longer reaches hydrated client code or session history. An explicit `HEAD` handler and a `Sec-Purpose`/`Purpose`/`X-moz` prefetch guard keep speculative requests from burning it (Next auto-implements `HEAD` by running the `GET`).
- `redirect_uri` is minted by the API from `SLACK_REDIRECT_URI`, so no API or Slack app config changes.

Notes for reviewers — deliberate trade-offs: during the exchange the browser shows its own document-loading state instead of the previous spinner (same as the other OAuth callbacks), and free-text API error `detail`s collapse to Prowler's own copy because free text must not ride the URL — `code`-mapped refusals keep their wording.

### Steps to review

Automated coverage ships with the PR: unit tests for the route handler (`ui/app/(prowler)/integrations/slack/callback/route.test.ts`, driving the real exchange action against the MSW contract double), browser-mode tests for the notice (`ui/app/(prowler)/integrations/slack/slack-connect-notice.integration.test.tsx`), and a round-trip test of the query-string contract.

### Checklist

<details>

</details>

- [ ] Review if code is being documented following this specification https://github.com/google/styleguide/blob/gh-pages/pyguide.md#38-comments-and-docstrings (N/A — TypeScript/UI change)
- [ ] Review if is needed to change the [README.md](https://github.com/prowler-cloud/prowler/blob/master/README.md) (N/A)
- [ ] Ensure a changelog fragment is added under [prowler/changelog.d/](https://github.com/prowler-cloud/prowler/tree/master/prowler/changelog.d), if applicable. (N/A — UI-only; fragment lives in `ui/changelog.d/`)

#### SDK/CLI

#### UI
- [ ] If this PR adds or updates npm dependencies, include package-health evidence (maintenance, popularity, known vulnerabilities, license, release age) and explain why existing/native alternatives are insufficient. (N/A — no dependency changes)
- [x] Ensure a changelog fragment is added under [ui/changelog.d/](https://github.com/prowler-cloud/prowler/tree/master/ui/changelog.d), if applicable.

#### API
- [ ] All issue/task requirements work as expected on the API (N/A — no API changes)

#### MCP Server
- [ ] All issue/task requirements work as expected on the MCP Server (N/A — no MCP changes)

### License

## Summary by CodeRabbit

* **Bug Fixes**
  * Slack OAuth returns now redirect correctly to the Slack integration page after approval.
  * Expired or interrupted connections provide clearer status messages and avoid reusing single-use authorization links.
  * Rate-limit and connection failures now preserve relevant status information for more accurate feedback.

* **User Experience**
  * Connection results appear as contextual notices on the Slack integration page.
  * Notices are removed from the URL after display, keeping links clean.
  * Channel selection controls are presented more clearly when Slack is not connected.


Alan-TheGentleman (changes requested):
Moving the callback exchange into a Route Handler is the right direction, but I’m requesting changes because the prefetch guard can still burn a single-use OAuth code. `isPrefetch()` uses `??`, so only the first present header is inspected; for example, `Sec-Purpose: navigate` with `Purpose: prefetch` reaches `GET` and calls `exchangeSlackOAuthCode()`. Please evaluate `Sec-Purpose`, `Purpose`, and `X-Moz` independently and add a mixed-header regression test before re-requesting review.

Two non-blocking follow-ups: preserve `window.location.hash` when cleaning the Slack query parameters, and avoid the call-site `text-xs` override on the shared `Button` by using an approved design-system variant.

alejandrobailo:
Thanks for the thoughtful feedback! I’ve pushed the fixes in `14af7c9da` and `965069f6f`: OAuth credentials are stripped before re-authentication, success notices are verified against server data, prefetch detection checks every supported purpose header, URL cleanup preserves fragments and uses `useMountEffect`, and the existing `link-xs` button variant is used. Regression tests cover the affected flows.

alejandrobailo (dismissed):

Alan-TheGentleman (changes requested):
All three points from my previous review are fixed. I’m still requesting changes because the new proxy test does not exercise the production NextAuth ordering: `callbacks.authorized` runs before the wrapped proxy handler and, for an expired or missing session, returns a sign-in redirect containing `nextUrl.pathname + nextUrl.search`, so the Slack `code` and `state` remain nested inside `callbackUrl` and the cleanup in `proxy.ts` never runs. Please handle the Slack callback in `authorized` or explicitly let it reach the wrapped handler, then add a production-faithful auth-ordering test that proves `code` and `state` are absent.

One additional non-blocking fix: add `Cache-Control: no-store` to the 204 guard responses and cover prefetch followed by ordinary navigation. HTTP 204 is heuristically cacheable, so a cached prefetch response can suppress the real OAuth exchange.

alejandrobailo (approved):

Alan-TheGentleman (approved):
All requested changes are fixed on this head. The production NextAuth ordering is now covered, Slack callback credentials are stripped before sign-in, and the 204 guards use `Cache-Control: no-store` with regression coverage. The failing E2E cases are unrelated provider-environment failures; the exact-head UI suite and build passed. Approved.

PR #12580, opened 2026-08-27
fix(api): apply findings partition max age in months, not days

cesararroba:
## Description

`FINDINGS_TABLE_PARTITION_MAX_AGE_MONTHS` is applied as **days**, not months.

```python
def relative_days_or_none(value):
    if value is None:
        return None
    return relativedelta(days=value)      # days

max_age=relative_days_or_none(
    settings.FINDINGS_TABLE_PARTITION_MAX_AGE_MONTHS   # named, and documented, as months
)
```

`api/docs/partitions.md` describes the setting as "the number of months to keep partitions before deleting them", and the name says the same. Setting it to `12` to keep a year instead drops every partition older than **12 days**:

| value | cutoff before | cutoff after |
|---|---|---|
| 12 | 12 days | 12 months |
| 3 | 3 days | 3 months |

This only fires the first time a deployment enables retention, and the conservative-looking value is the destructive one, so it is worth correcting before anyone relies on it.

## Zero

An earlier revision of this description claimed `0` would delete every partition. That was wrong, as @josema-xyz pointed out. `relativedelta(days=0)` is falsy and `to_delete()` guards with `if not self.max_age: return`, so `0` has always meant "keep partitions indefinitely". That behaviour is unchanged here and is now covered by a test.

## Negative values

Negative values are a different matter. `to_delete()` sets the cutoff to `get_start_datetime() - max_age` and then walks backwards, so a negative value puts the cutoff in the future and starts offering partitions for deletion from there.

How much damage that does depends on the value, as @josema-xyz pointed out. The manager breaks at the first partition name it cannot find in the database, so a large negative lands beyond the pre-created window and stops harmlessly, while a small one lands inside it and works backwards through partitions that are actively in use. Either way the setting can select live data for deletion.

Silently coercing that to "keep everything" would hide a typo in a setting whose whole purpose is deletion, so it now raises `ImproperlyConfigured` instead.

## Changes

- `relative_days_or_none` → `relative_months_or_none`, returning `relativedelta(months=...)`
- Negative values raise `ImproperlyConfigured`; `None` and `0` keep partitions indefinitely, as before
- Tests for the helper, and for `to_delete()` itself — that nothing is yielded without a max age, that the first partition offered for deletion is exactly `max_age` old, and that the generator walks backwards in time

Both partitioning configs (`Finding` and `ResourceFindingMapping`) use the helper and are updated together. The default remains unset, so no deployment changes behaviour on upgrade.

## Summary by CodeRabbit

* **Bug Fixes**
  * Partition retention settings are now interpreted in months rather than days.
  * Zero values keep partitions indefinitely, while negative values are rejected as invalid.

* **Tests**
  * Added coverage for month-based retention, indefinite storage, invalid values, and partition cleanup behavior.

* **Documentation**
  * Updated the changelog to describe the corrected retention setting behavior.

### DIFF
--- a/api/changelog.d/findings-partition-max-age-unit.fixed.md
+++ b/api/changelog.d/findings-partition-max-age-unit.fixed.md
@@ -1,1 +1,1 @@
-`FINDINGS_TABLE_PARTITION_MAX_AGE_MONTHS` is now applied in months instead of days, and non-positive values keep partitions indefinitely
+`FINDINGS_TABLE_PARTITION_MAX_AGE_MONTHS` is now applied in months instead of days, and negative values are rejected
--- a/api/changelog.d/findings-partition-max-age-unit.fixed.md
+++ b/api/changelog.d/findings-partition-max-age-unit.fixed.md
@@ -1,0 +1,1 @@
+`FINDINGS_TABLE_PARTITION_MAX_AGE_MONTHS` is now applied in months instead of days, and non-positive values keep partitions indefinitely
--- a/api/src/backend/api/partitions.py
+++ b/api/src/backend/api/partitions.py
@@ -6,6 +6,7 @@ from api.rls import RowLevelSecurityConstraint
 from api.uuid_utils import datetime_to_uuid7
 from dateutil.relativedelta import relativedelta
 from django.conf import settings
+from django.core.exceptions import ImproperlyConfigured
 from psqlextra.partitioning import (
     PostgresPartitioningError,
     PostgresPartitioningManager,
@@ -154,8 +155,14 @@ class PostgresUUIDv7PartitioningStrategy(PostgresRangePartitioningStrategy):
 
 
 def relative_months_or_none(value):
-    # A non-positive value keeps partitions indefinitely, like None.
-    if value is None or value <= 0:
+    # A negative value would set the cutoff in the future and delete every
+    # partition, so it is rejected rather than silently ignored.
+    if value is not None and value < 0:
+        raise ImproperlyConfigured(
+            "FINDINGS_TABLE_PARTITION_MAX_AGE_MONTHS must not be negative; "
+            "leave it unset or use 0 to keep partitions indefinitely"
+        )
+    if not value:
         return None
     return relativedelta(months=value)
 
--- a/api/src/backend/api/partitions.py
+++ b/api/src/backend/api/partitions.py
@@ -153,10 +153,11 @@ class PostgresUUIDv7PartitioningStrategy(PostgresRangePartitioningStrategy):
         )
 
 
-def relative_days_or_none(value):
-    if value is None:
+def relative_months_or_none(value):
+    # A non-positive value keeps partitions indefinitely, like None.
+    if value is None or value <= 0:
         return None
-    return relativedelta(days=value)
+    return relativedelta(months=value)
 
 
 #
@@ -173,7 +174,7 @@ manager = PostgresPartitioningManager(
                     months=settings.FINDINGS_TABLE_PARTITION_MONTHS
                 ),
                 count=settings.FINDINGS_TABLE_PARTITION_COUNT,
-                max_age=relative_days_or_none(
+                max_age=relative_months_or_none(
                     settings.FINDINGS_TABLE_PARTITION_MAX_AGE_MONTHS
                 ),
                 name_format="%Y_%b",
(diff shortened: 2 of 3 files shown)


api/src/backend/api/tests/test_partitions.py
```diff
@@ -0,0 +1,19 @@
+import pytest
+from api.partitions import relative_months_or_none
+from dateutil.relativedelta import relativedelta
+
+
+class TestRelativeMonthsOrNone:
+    def test_none_keeps_partitions_indefinitely(self):
+        assert relative_months_or_none(None) is None
+
+    @pytest.mark.parametrize("value", [0, -1])
```
josema-xyz: Could we test `to_delete()` too? That is the code that chooses which partitions to remove. Also, `0` was already safe because `relativedelta(days=0)` is false, so the PR description should be corrected.

api/src/backend/api/partitions.py
```diff
@@ -153,10 +153,11 @@ def get_start_datetime(self) -> datetime:
         )
 
 
-def relative_days_or_none(value):
-    if value is None:
+def relative_months_or_none(value):
+    # A non-positive value keeps partitions indefinitely, like None.
+    if value is None or value <= 0:
```
josema-xyz: Should negative values disable deletion? `None` already means "keep forever", and the docs do not mention negative values. A typo like `-12` would silently stop the cleanup. I think negative values should either fail or be documented.

api/src/backend/api/tests/test_partitions.py
cesararroba: You're right about the 0, my mistake. `relativedelta(days=0)` is falsy and `to_delete()` bails out on `if not self.max_age`, so it was never going to delete anything. I've corrected the PR description.

And yes, good call on testing `to_delete()`, that's the part that actually decides. Added three cases: nothing is yielded when there's no max age, the first partition offered up is exactly max_age old, and the generator walks backwards in time.

api/src/backend/api/partitions.py
cesararroba: Fair point, and it turns out to be worse than just stopping the cleanup. `to_delete()` sets the cutoff to `get_start_datetime() - max_age` and then decrements, so a negative value puts the cutoff in the future and it walks back through every partition until there are none left. Before this PR a `-12` would have dropped the whole table rather than doing nothing.

So I've made it raise `ImproperlyConfigured` instead of quietly turning into None. Unset and 0 still mean keep everything, which is the documented behaviour.

api/changelog.d/findings-partition-max-age-unit.fixed.md
```diff
@@ -0,0 +1 @@
+`FINDINGS_TABLE_PARTITION_MAX_AGE_MONTHS` is now applied in months instead of days, and non-positive values keep partitions indefinitely
```
josema-xyz: Negative values now raise an error, so "non-positive values" is no longer correct here. The PR description also says `-12` would delete the whole table, but that is not always true because the manager stops if the first partition it checks does not exist. It would be more accurate to say that negative values can select active partitions for deletion.
```suggestion
`FINDINGS_TABLE_PARTITION_MAX_AGE_MONTHS` is now applied in months instead of days, and negative values are rejected
```

api/src/backend/api/partitions.py
```diff
@@ -153,10 +154,17 @@ def get_start_datetime(self) -> datetime:
...
-def relative_days_or_none(value):
-    if value is None:
+def relative_months_or_none(value):
+    # A negative value would set the cutoff in the future and delete every
+    # partition, so it is rejected rather than silently ignored.
+    if value is not None and value < 0:
+        raise ImproperlyConfigured(
+            "FINDINGS_TABLE_PARTITION_MAX_AGE_MONTHS must not be negative; "
+            "leave it unset or use 0 to keep partitions indefinitely"
+        )
```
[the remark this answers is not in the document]
josema-xyz: The current PR description says negative values are rejected, so this seems based on an earlier version. Raising here is intentional because returning `None` would hide an invalid setting. I left review feedback about the remaining description and changelog wording, but the code should not return `None` for negative values.

api/changelog.d/findings-partition-max-age-unit.fixed.md
cesararroba: Both fair, taking your wording for the changelog.

And you're right that I overstated the negative case. I went and read the manager: it breaks at the first partition name it can't find (`if not introspected_partition: break`), so a `-12` puts the cutoff out past the pre-created window and stops without touching anything. A `-1` or `-2` is the dangerous one, since the cutoff lands inside partitions that exist and it then works backwards through live data. Fixed the description to say that instead.

api/src/backend/api/partitions.py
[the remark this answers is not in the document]
cesararroba: Thanks for stepping in on that one, agreed. Leaving the raise as is.

josema-xyz (approved):

PR #12645, opened 2026-08-28
fix(checks): report MANUAL instead of FAIL on permission and data-availability errors

danibarranqueroo:
### Context

Some checks emitted `FAIL` — up to **critical** severity — when the underlying API call failed because the scanning identity lacked a permission, an API was not enabled or a feature was not licensed (e.g. `defenderxdr_endpoint_privileged_user_exposed_credentials` reporting a Critical FAIL for a missing `ThreatHunting.Read.All`). That is a scan-configuration problem, not a security issue: it surfaces misleading findings in reports/Lighthouse and skews compliance scores.

PROWLER-2325

### Description

Convention: a check must never `FAIL` because data could not be retrieved. It emits a **single** tenant/account/project-level `MANUAL` finding whose `status_extended` names the required permission/API/license. This is the pattern Okta, Bedrock, the `*_no_secrets` checks and several M365 checks already used; this PR applies it consistently.

Full audit of every provider; checks changed:

- **M365 (8):** `defenderxdr_endpoint_privileged_user_exposed_credentials`, `defenderxdr_critical_asset_management_pending_approvals`, `defenderidentity_health_issues_no_open`, `entra_app_registration_no_unused_privileged_permissions`, `entra_seamless_sso_disabled` (now consistent with `entra_directory_sync_object_takeover_blocked`, which already used MANUAL for the same error flag), `entra_break_glass_account_fido2_security_key_registered`, `exchange_shared_mailbox_sign_in_disabled`, and `entra_users_mfa_capable` (one tenant-level MANUAL instead of one FAIL **per user**).
- **Azure (1):** `entra_user_with_recent_sign_in` tenant-level "no sign-in data" branch.
- **AWS (18):** the 15 `cloudwatch_log_metric_filter_*` / `cloudwatch_changes_to_*_alarm_configured` checks (a denied `logs:DescribeMetricFilters` or `cloudwatch:DescribeAlarms` produced 15 medium FAILs), `bedrock_agent_role_least_privilege`, `rolesanywhere_trust_anchor_pqc_pki`, `s3_bucket_cross_region_replication` (out-of-scope destination bucket).
- **GCP (1):** `iam_account_access_approval_enabled`. Required a service change: `AccessApproval._get_settings` now treats `404` as "not enabled" (legitimate FAIL) and tracks `403` / API-disabled / other exceptions in `settings_lookup_failed`, which the check reports as MANUAL.

Deliberately left as `FAIL` after tracing the service layer (the `None`/empty is a real security condition, not an error): "MDE not enabled" (table-not-found, distinct from an API error), "no MDI sensors" (empty list, distinct from `None`), the Google Workspace "insecure default" checks (gated on `policies_fetched`) and the `email_otp` fail-safe default.

Convention documented in `docs/developer-guide/checks.mdx` (new section *Permission and Data-Availability Errors Are Not Findings*) and in `skills/prowler-sdk-check/SKILL.md` so new checks don't reintroduce the pattern.

**Second pass — service-level review against the real APIs.** Tracing every changed check through its service and the vendor docs surfaced five cases where the permission path still produced FAIL or nothing; fixed in the second commit:

- **Azure `entra_user_with_recent_sign_in`:** Graph rejects the whole `/users` request with 403 `Authentication_RequestFromNonPremiumTenantOrB2CTenant` when P1/P2 or `AuditLog.Read.All` is missing, so the check emitted *nothing*. The service now retries without `signInActivity`, records the tenant in `sign_in_activity_unavailable`, and the check emits one tenant-level MANUAL. The old all-`None` heuristic (which turned genuinely never-signed-in users into MANUAL) is gone: those FAIL.
- **M365 Defender XDR (both checks):** `_run_hunting_query` already detected "table not found" but the Exposure Management / CAM callers discarded the flag → `[]` → **PASS** on tenants without Security Exposure Management. Now `None` → MANUAL.
- **M365 `entra_users_mfa_capable` / break-glass FIDO2:** `user_registration_details_error` is only set after `/users` and directory-role calls succeed; when those fail the checks emitted nothing. New `entra_client.users_error` → MANUAL.
- **M365 `entra_seamless_sso_disabled`:** hybrid org with no sync settings reported PASS "cloud-only"; now MANUAL, consistent with `entra_directory_sync_object_takeover_blocked`. Remediation text notes Graph only supports this endpoint with a delegated permission.
- **AWS `s3_bucket_cross_region_replication`:** denied `GetBucketVersioning` / `GetReplicationConfiguration` produced FAIL; new `versioning_retrieved` / `replication_retrieved` flags on `Bucket` → MANUAL, with deterministic PASS > MANUAL > FAIL precedence across rules.

**Third pass — review-round fixes and final audit (fa9b8f4df9).** All inline review comments are addressed, plus a final full-branch audit. Deliberate behaviour decisions worth knowing when reviewing:

- **GCP Access Approval:** a *definitively disabled* `accessapproval.googleapis.com` API is a **FAIL** ("Access Approval provably cannot be enabled"), consistent under both the API precheck and `--skip-api-check`; an *undetermined* API state is MANUAL. The developer guide carve-out documents this ("an API whose activation is itself the audited control").
- **CloudWatch:** the `*_unavailable` flags now rise on every listing error (not only the AccessDenied code), so throttling/opt-in-region failures also degrade FAIL→MANUAL; a PASS backed by an observed filter+alarm (matched by metric, namespace and region) is never degraded.
- **`s3_bucket_object_versioning`** joins the fixed set (denied `s3:GetBucketVersioning` was a false FAIL), bringing the total to **29 checks**.
- **Azure `entra_user_with_recent_sign_in`:** the old "all users have null sign-in → tenant MANUAL" heuristic is gone. A Graph **403** on `signInActivity` (no P1/P2 / missing `AuditLog.Read.All`) → one tenant-level MANUAL; a 200 with users that simply have no recorded sign-in → per-user FAILs (they are genuinely stale accounts). Finding counts can go *up* for such tenants — intentional.
- **Out of scope, deliberately:** checks that *silently skip* on permission errors (e.g. the four `cloudwatch_log_group_*` checks, `bedrock_model_invocation_logs_encryption_enabled`'s log-group lookup) keep their current behaviour by team decision — they are fixed reactively per reported issue, following this PR's convention.

### Steps to review

1. Review the convention text in `docs/developer-guide/checks.mdx` and `skills/prowler-sdk-check/SKILL.md`.
2. For each changed check, confirm the branch that became `MANUAL` is reached only through a service-level error path (`None`, `*_error`, `*_lookup_failed`), not a legitimate empty result.
3. Run the affected tests (all pass locally, 406 tests):
   ```bash
   uv run pytest tests/providers/m365/services/defenderxdr tests/providers/m365/services/defenderidentity \
     tests/providers/m365/services/entra/entra_app_registration_no_unused_privileged_permissions \
     tests/providers/m365/services/entra/entra_users_mfa_capable \
     tests/providers/m365/services/entra/entra_break_glass_account_fido2_security_key_registered \
     tests/providers/m365/services/entra/entra_seamless_sso_disabled \
     tests/providers/m365/services/exchange/exchange_shared_mailbox_sign_in_disabled \
     tests/providers/azure/services/entra/entra_user_with_recent_sign_in \
     tests/providers/aws/services/cloudwatch tests/providers/aws/services/bedrock/bedrock_agent_role_least_privilege \
     tests/providers/aws/services/rolesanywhere tests/providers/aws/services/s3/s3_bucket_cross_region_replication \
     tests/providers/gcp/services/iam
   ```
4. Optional end-to-end: run `prowler m365 --check defenderxdr_endpoint_privileged_user_exposed_credentials` with an app registration that lacks `ThreatHunting.Read.All`; expect one `MANUAL` finding instead of a Critical `FAIL`.

### Checklist

<details>

</details>

#### SDK/CLI
    - If so, do we need to update permissions for the provider? N/A

### License

## Summary by CodeRabbit

* **Bug Fixes**
  * Checks now distinguish genuine security failures from unavailable data caused by permissions, disabled APIs, licensing, or inaccessible resources.
  * These situations produce a single resource-scoped **MANUAL** finding with remediation guidance instead of an incorrect **FAIL**.
  * Improved handling across CloudWatch, Entra, Defender, Access Approval, Bedrock, S3 replication, and other checks.
  * Not-found conditions remain classified as genuine failures where appropriate.

* **Documentation**
  * Updated developer and SDK guidance for manual-review behavior and availability errors.

* **Tests**
  * Added coverage for manual findings, permission errors, API failures, and unavailable data.


prowler/providers/aws/services/cloudwatch/cloudwatch_changes_to_network_gateways_alarm_configured/cloudwatch_changes_to_network_gateways_alarm_configured.py
```diff
@@ -34,16 +34,26 @@ def execute(self):
...
-                report.resource_id = logs_client.audited_account
-                report.resource_arn = logs_client.log_group_arn_template
-                report.resource_tags = []
+            report.region = logs_client.region
+            report.resource_id = logs_client.audited_account
+            report.resource_arn = logs_client.log_group_arn_template
+            report.resource_tags = []
 
-            findings.append(report)
+        findings.append(report)
```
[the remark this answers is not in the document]
danibarranqueroo: Addressed in 0702f9d409 with a narrower rule than suggested, on purpose.

A denied listing in any region now turns every **FAIL** from `check_cloudwatch_log_metric_filter` into MANUAL — both "no log groups found with metric filters or alarms" and "filter found but no alarms associated" — because with an incomplete inventory neither absence can be asserted.

A **PASS** is kept as PASS. When the helper finds a metric filter *and* its alarm in a readable region, that is direct evidence the control exists; an unreadable inventory elsewhere cannot invalidate it, and turning it into MANUAL would make the check report "cannot evaluate" for a control we have actually observed.

Regression tests added in `cloudwatch_log_metric_filter_root_usage_test.py`: match + partial denial → PASS, filter-without-alarm + alarm listing denied → MANUAL, filter-without-alarm + full inventory → FAIL.

prowler/providers/aws/services/cloudwatch/cloudwatch_service.py
```diff
@@ -103,6 +107,9 @@ def __init__(self, provider):
         self.resource_policies = {}
         self.__threading_call__(self._describe_resource_policies)
         self.metric_filters = []
+        # True when DescribeMetricFilters was denied in at least one audited
+        # region, so the metric filter inventory may be incomplete.
+        self.metric_filters_unavailable = False
```
HugoPBrito, line 117: Can we include `DescribeLogGroups` in this availability contract?

If that call fails while metric filters are retrieved, the service can create a filter with `log_group=None`, which the shared helper dereferences. I believe this retrieval failure should produce `MANUAL` instead of raising an exception, with service-level regression coverage.

prowler/providers/aws/services/cloudwatch/cloudwatch_log_metric_filter_root_usage/cloudwatch_log_metric_filter_root_usage.py
```diff
@@ -24,16 +40,28 @@ def execute(self):
...
+            report.status_extended = "No CloudWatch log groups found with metric filters or alarms associated."
+            report.region = logs_client.region
+            report.resource_id = logs_client.audited_account
+            report.resource_arn = logs_client.log_group_arn_template
+            report.resource_tags = []
+
+        # A denied listing in any region means the inventory is incomplete: a
+        # PASS is still backed by a real filter and alarm, but a FAIL (nothing
+        # found, or a filter found without its alarm) cannot be trusted.
+        if report.status == "FAIL" and inventory_unavailable:
```
HugoPBrito, line 63: I believe preserving `PASS` under partial inventory is only safe if we prove that the alarm belongs to this metric filter.

The helper currently compares only the metric name, ignoring namespace and region, so an unrelated alarm with the same metric name can produce a false `PASS`. Can we retain and compare the metric name, namespace, and region here?

prowler/providers/m365/services/defenderxdr/defenderxdr_service.py
```diff
@@ -200,11 +200,20 @@ async def _get_exposed_credentials_privileged_users(
     TargetCategories = TargetNodeCategories
 """
 
-        results, _ = await self._run_hunting_query(query)
+        results, table_not_found = await self._run_hunting_query(query)
```
HugoPBrito, line 209: Wouldn’t it be better for `_run_hunting_query` to return `(None, False)` when the Advanced Hunting response is null?

Returning an empty list treats the request as a successful query with no results. The callers can then report `PASS` even though no data was retrieved. I believe `[]` should be reserved for a confirmed non-null response with no rows.

prowler/providers/m365/services/entra/entra_app_registration_no_unused_privileged_permissions/entra_app_registration_no_unused_privileged_permissions.py
```diff
@@ -47,11 +48,12 @@ def execute(self) -> list[CheckReportM365]:
                 resource_name="OAuth Applications",
                 resource_id="oauthApps",
             )
-            report.status = "FAIL"
+            report.status = "MANUAL"
```
HugoPBrito, line 51: Can we preserve the distinction between an unavailable response and a confirmed empty result here?

When the Advanced Hunting call returns `None`, `_get_oauth_apps` currently returns `{}`, so this branch is skipped and the check reports that there are no OAuth applications as `PASS`. I believe the null response should propagate as `None`, with a service-to-check regression test.

prowler/providers/m365/services/defenderidentity/defenderidentity_health_issues_no_open/defenderidentity_health_issues_no_open.py
```diff
@@ -50,11 +52,12 @@ def execute(self) -> List[CheckReportM365]:
                 resource_name="Defender for Identity",
                 resource_id="defenderIdentity",
             )
-            report.status = "FAIL"
+            report.status = "MANUAL"
```
HugoPBrito, line 55: Can we handle `sensors_api_failed` independently before reaching the `PASS` branch?

With `sensors=None` and `health_issues=[]`, sensor deployment cannot be verified, but the check currently reports `PASS`. I believe this case should produce `MANUAL`, with an asymmetric API-failure test.

prowler/providers/m365/services/exchange/exchange_shared_mailbox_sign_in_disabled/exchange_shared_mailbox_sign_in_disabled.py
```diff
@@ -45,8 +47,8 @@ def execute(self) -> List[CheckReportM365]:
             )
 
             if not entra_user:
-                report.status = "FAIL"
-                report.status_extended = f"Shared mailbox {shared_mailbox.user_principal_name} could not be found in Entra ID for verification."
+                report.status = "MANUAL"
```
HugoPBrito, line 68: I believe we should distinguish a tenant-wide `users_error` from a genuinely unresolved mailbox user.

When the users collection fails, this loop can emit one misleading `MANUAL` per mailbox. Can we return one tenant-level `MANUAL` before iterating, while keeping this resource-level branch for successful inventories where an individual user is missing?

prowler/providers/gcp/services/iam/iam_account_access_approval_enabled/iam_account_access_approval_enabled.py
```diff
@@ -18,7 +18,15 @@ def execute(self) -> Check_Report_GCP:
             report.status_extended = (
                 f"Project {project_id} has Access Approval enabled."
             )
-            if project_id not in accessapproval_client.settings:
+            if project_id in accessapproval_client.settings_lookup_failed:
```
HugoPBrito, line 40: Can we cover the API-disabled path before the project is filtered out?

`GCPService` removes projects where the Access Approval API is disabled before `_get_settings()` and this check run. As a result, those projects never reach this `MANUAL` branch and produce no finding. I believe we should preserve the precheck state and emit one project-level `MANUAL`, with a test exercising the real API-activation filtering path.

HugoPBrito (changes requested):
Thanks for working on this. I believe changing unavailable data from `FAIL` to `MANUAL` is the right direction and will prevent misleading findings and compliance results.

I found a few remaining paths where unavailable data can still become `PASS`, `FAIL`, no finding at all, or an exception. I left the concrete cases inline.

prowler/providers/aws/services/cloudwatch/cloudwatch_service.py
danibarranqueroo: Done in b79b058ccc: `Logs.log_groups_unavailable` joins the availability contract of the 15 checks (message now names `logs:DescribeLogGroups`), the helper skips filters whose `log_group` is `None` instead of dereferencing it, and a service-level test asserts a denied `DescribeLogGroups` raises the flag and nulls the inventories.

prowler/providers/aws/services/cloudwatch/cloudwatch_log_metric_filter_root_usage/cloudwatch_log_metric_filter_root_usage.py
danibarranqueroo: Done in b79b058ccc: `MetricFilter` now keeps `metricNamespace` and the helper requires metric name + region, plus namespace when both sides expose one (permissive when either lacks it, so pre-existing objects behave as before). Unit tests cover same-name/other-namespace and other-region alarms not matching. Composite alarms were already excluded (only `MetricAlarms` pages are read); metric-math alarms have `metric=None` and never matched before either.

prowler/providers/m365/services/defenderxdr/defenderxdr_service.py
danibarranqueroo: Done in b79b058ccc: `_run_hunting_query` returns `(None, False)` for a null response object and reserves `[]` for a confirmed empty `results`. This also fixes `_check_mde_status`, where a null response previously became "no_devices" → PASS. Service tests added in `defenderxdr_service_test.py`.

prowler/providers/m365/services/entra/entra_app_registration_no_unused_privileged_permissions/entra_app_registration_no_unused_privileged_permissions.py
danibarranqueroo: Done in b79b058ccc: `_get_oauth_apps` returns `None` on a null hunting response, which the check already maps to MANUAL; `{}` is now only a confirmed empty result. Service-to-check regression tests added in `microsoft365_entra_service_test.py`.

prowler/providers/m365/services/defenderidentity/defenderidentity_health_issues_no_open/defenderidentity_health_issues_no_open.py
danibarranqueroo: Done in b79b058ccc: with `sensors=None` and `health_issues=[]` the check now returns MANUAL naming SecurityIdentitiesSensors.Read.All; the PASS is only kept when sensor deployment was actually verified. Asymmetric-failure test added. When health issues DO exist alongside a failed sensors API they are still reported individually — real evidence is kept.

prowler/providers/m365/services/exchange/exchange_shared_mailbox_sign_in_disabled/exchange_shared_mailbox_sign_in_disabled.py
danibarranqueroo: Done in b79b058ccc: a tenant-wide `entra_client.users_error` now produces one tenant-level MANUAL before the loop (same shape as entra_users_mfa_capable), and the per-mailbox branch remains for successful inventories where an individual user cannot be resolved.

prowler/providers/gcp/services/iam/iam_account_access_approval_enabled/iam_account_access_approval_enabled.py
danibarranqueroo: Addressed in b79b058ccc with one deliberate refinement: `GCPService` now records the precheck state in two sets — `api_disabled_project_ids` (serviceusage says DISABLED) and `api_state_unknown_project_ids` (the state call itself failed). The check reports the unknown case as MANUAL as you suggested, but a definitively disabled accessapproval API is reported as FAIL: with the API off, Access Approval cannot be enabled, so it is the same condition as the existing 'not enabled' FAIL — and since that API is disabled by default in most projects, MANUAL there would both mask a real finding and flood tenants with MANUAL noise. Tests cover the precheck population and both verdicts. Happy to switch the disabled case to MANUAL if you disagree.

prowler/providers/azure/services/entra/entra_service.py
```diff
@@ -69,24 +75,64 @@ def __init__(self, provider: AzureProvider):
...
+                    status = getattr(error, "response_status_code", None)
+                    reason = self._describe_graph_error(error)
+                    if status != 403:
+                        # Transient or unexpected failure (throttling, 5xx,
+                        # network): do not blame licensing/permissions and do
+                        # not degrade the tenant's sign-in evaluation to MANUAL.
+                        logger.error(
+                            f"{tenant} -- {error.__class__.__name__}[{error.__traceback__.tb_lineno}]: {error}"
+                        )
+                        continue
```
HugoPBrito, line 129: Wouldn’t it be better to preserve a tenant-level retrieval error here instead of continuing with an empty user inventory?

A 429, 5xx, or network failure is not evidence that the tenant has no users. With `{}`, `_get_directory_roles()` filters out the actual role members, so the Global Administrator check can `PASS` with zero administrators while the other user-based checks emit nothing.

I believe we should record a generic users-retrieval error, return `MANUAL` from the dependent checks, and update the new 503 test accordingly.

HugoPBrito (changes requested):
Thanks for addressing the previous comments. I reviewed the new commits and those findings look resolved.

I found one remaining data-availability path where a Microsoft Graph failure can still become a false `PASS`. I left the specific case inline.

prowler/providers/azure/services/entra/entra_service.py
danibarranqueroo: Good catch, addressed in eb4e0cca2. The service now records tenant-level users retrieval errors in `users_retrieval_errors` (both for transient non-403 failures and for a failing retry after a signInActivity 403), and the five user-dependent checks (`entra_user_with_recent_sign_in`, `entra_global_admin_in_less_than_five_users`, `entra_privileged_user_has_mfa`, `entra_non_privileged_user_has_mfa`, `entra_user_with_vm_access_has_mfa`) emit a tenant-level `MANUAL` finding instead of evaluating the empty inventory — so the Global Administrator check can no longer PASS with zero administrators. The 503 service test now asserts the tenant is recorded, and each check has a new MANUAL test (the global admin one reproduces the false-PASS scenario you described).

prowler/providers/azure/services/entra/entra_service.py
```diff
@@ -69,24 +81,75 @@ def __init__(self, provider: AzureProvider):
...
+
+        Users are requested with ``signInActivity``. When Graph rejects that
+        request (the tenant lacks Entra ID P1/P2 or the application lacks
+        ``AuditLog.Read.All``), the tenant is recorded in
+        ``self.sign_in_activity_errors`` and the users are fetched again
+        without ``signInActivity`` so the remaining user checks can still run.
+
+        Any other failure to retrieve the users (throttling, 5xx, network)
+        is recorded in ``self.users_retrieval_errors`` so the user-based
+        checks report MANUAL instead of evaluating an empty inventory.
```
HugoPBrito: Wouldn’t it be better to apply this availability contract to pagination failures as well?

This is not a new security vulnerability, but it affects the integrity of the scan results. If the first page succeeds and a later page fails because of throttling, a 5xx response, timeout, or network error, the current handler keeps the users already retrieved without setting `users_retrieval_errors`. The five dependent checks then treat that partial inventory as complete. This can make the Global Administrator check false-PASS and cause the other checks to omit users that were never retrieved.

Since this PR establishes that retrieval failures must produce `MANUAL`, I don’t believe we should accept partial pagination as verified data. Can we mark the tenant as unavailable when any page fails and add a regression test where the first page succeeds but the second one returns an error?

danibarranqueroo: Agreed — addressed in 4135513ce. A failure on any later page now records the tenant in `users_retrieval_errors` (the partially retrieved users are kept, but the five dependent checks see the tenant as unavailable and report `MANUAL` instead of evaluating the partial inventory as complete). Added the regression test you asked for: first page returns a user with a next link, second page fails with a 503 — the user is in the inventory but the tenant is marked unavailable and licensing is not blamed.

HugoPBrito (approved):

Issue #12589, opened 2026-08-27
Snowflake provider


kkmookhey:
### New feature motivation

Snowflake is where a lot of regulated data now sits, and since Cortex it is also where AI
runs — agents, Analyst, and LLM functions invoked directly from SQL. Prowler has no
Snowflake provider, so for an organisation whose estate is AWS plus Snowflake, the Snowflake
half is simply absent from the findings and compliance view.

The account-level questions are ordinary posture questions that Prowler already answers well
for other providers: which users can sign in with a password and no MFA, who holds
`ACCOUNTADMIN` or `ORGADMIN` as a standing grant rather than assuming it, who defaults to a
privileged role, and what the login-failure pattern looks like. Snowflake exposes all of it
read-only through `SNOWFLAKE.ACCOUNT_USAGE`.

### Solution Proposed

A `snowflake` provider, following the shape of the Supabase provider in #12489:

- **Authentication:** key-pair on a `TYPE = SERVICE` user — no password, and the private key
  never leaves the customer. Account, user, role and warehouse as arguments.
- **Permissions:** a single grant, `IMPORTED PRIVILEGES ON DATABASE SNOWFLAKE`, to a role the
  customer creates. Every statement is a `SELECT` against `ACCOUNT_USAGE`; there is no write
  path and no access to customer tables.
- **Scope for a first PR:** provider scaffolding plus a small number of account-posture
  checks — MFA, standing admin grants, privileged default roles — rather than a large surface
  landing at once.

We have been running this outside Prowler against a live Snowflake account of our own, so the
query set, the `ACCOUNT_USAGE` column behaviour and the failure modes are known rather than
assumed. One
worth flagging: `USERS.EXT_AUTHN_DUO` alone is not a valid MFA signal — `HAS_MFA` and
`BYPASS_MFA_UNTIL` both matter, and keying on Duo reports every user on native MFA as
unprotected.

### Use case and benefits

Security teams running AWS or Azure alongside Snowflake currently have to check Snowflake
posture separately, by hand or not at all. Folding it into the same findings, ThreatScore and
compliance views is the whole benefit — one place to look, and Snowflake stops being the
provider nobody reviews.

### Describe alternatives you've considered

- **Riding another provider's scan.** A check attached to the AWS or Azure provider can report
  Snowflake state, which works for customers who have one of those connected and does nothing
  for a data team whose estate is Snowflake alone.
[...]

PR #12653, opened 2026-08-29
feat(sdk): add Snowflake provider

kkmookhey:
### Context

Adds `snowflake` as a provider with one check, as agreed in #12589 — provider base plus a
single check, following the developer guide.

### Description

**Authentication** is key-pair against a `TYPE = SERVICE` user, which has no password and
cannot sign in interactively. The private key and its passphrase are read only from the
environment; there are deliberately no CLI flags for them, because the key is the whole
credential and a flag puts it in shell history and the process list. Only non-secret
context is accepted as a flag: `--account`, `--user`, `--role`, `--warehouse`,
`--private-key-path`.

**No new dependency.** Snowflake's official connector is a large transitive tree, and
every statement here is a read against `SNOWFLAKE.ACCOUNT_USAGE`, which the SQL API serves
over HTTPS — so this is built on `requests`, `pyjwt` and `cryptography`, all already
shipped. A single grant, `IMPORTED PRIVILEGES ON DATABASE SNOWFLAKE`, covers the provider,
and no table in the customer's own databases is read.

The identity is read back from the account with `CURRENT_ACCOUNT()` rather than echoed
from the arguments, so a credential that authenticates against a different account than
intended is visible rather than silent.

**`users_mfa_enabled`** reports users who can sign in with a password but are not
protected by MFA. Three details it gets deliberately right, each a false result otherwise:

- **`EXT_AUTHN_DUO` alone is not a valid MFA signal.** It is the legacy column and remains
  the only signal for an account enrolled through Duo, but `HAS_MFA` is what Snowflake sets
  for native MFA. Keying on either one alone reports a whole class of protected users as
  unprotected — a false high-severity finding per user.
- **`BYPASS_MFA_UNTIL` is a timestamp**, so a future value means password-only right now
  regardless of enrolment. An unparseable value is treated as *not* bypassed on purpose:
  this decides whether to raise a high-severity finding, and inventing one from a value
  that could not be read is a false positive against a user who may well be fine.
- **Disabled users and users without a password are not reported at all.** Neither is
  exposed by the absence of a second factor, and a key-pair-only service user is the shape
  we recommend rather than a finding.

The users query carries no `LIMIT`. A cap would silently under-report on a large account,
and since any ordering that accompanies a cap is by creation date, the rows dropped would
be the oldest — which is exactly where long-standing unprotected accounts live.

`SnowflakeSession.__repr__` is overridden because the client holds the private key and
pydantic prints every field by default, so an unguarded session in a traceback would
publish the whole credential. There is a test asserting it.

### Steps to review

Setup needs a Snowflake account, a `TYPE = SERVICE` user with an RSA key pair, and
`GRANT IMPORTED PRIVILEGES ON DATABASE SNOWFLAKE` — the authentication page in this PR has
the exact statements.

```bash
export SNOWFLAKE_ACCOUNT="myorg-myaccount"
export SNOWFLAKE_USER="PROWLER_SVC"
export SNOWFLAKE_PRIVATE_KEY_PATH="/path/to/prowler_key.p8"
export SNOWFLAKE_ROLE="PROWLER_RO"
export SNOWFLAKE_WAREHOUSE="PROWLER_WH"

prowler snowflake --list-checks
prowler snowflake
```

Without an account, the suite exercises everything but the network call:

```bash
pytest tests/providers/snowflake
```

Worth a look specifically: `users_mfa_enabled_test.py` pins the enrolment and bypass
behaviour described above, and `snowflake_provider_test.py::TestSnowflakeProviderSetupSession::test_a_session_never_prints_the_key`
pins the credential-redaction behaviour.

### Checklist

- Provider package: provider class, models, exceptions (22000–22999), arguments, mutelist,
  service base
- Core wiring: `CheckReportSnowflake`, config enum, CLI parser, `init_global_provider`
  dispatch, finding / summary table / outputs / HTML summary
- Docs: authentication and getting-started pages, `docs.json`, provider table in
  `introduction.mdx`
- `.github/labeler.yml` and an `sdk-tests.yml` job
- Towncrier fragment and a mutelist example
- 78 tests covering the provider, the SQL client, the service, the check, the mutelist and
  the metadata

### One note for reviewers

`tests/config/schema/scan_config_schema_test.py::Test_Aggregated_Schema_Is_App_Facing::test_schema_contains_only_app_providers`
is order-dependent and fails intermittently on clean `master` as well — roughly one run in
four of `pytest tests/lib tests/providers/common tests/config`. I confirmed it on `master`
without this branch before mentioning it. Unrelated to this change; flagging it so it does
not read as one. Happy to open a separate issue if useful.

### License

By submitting this pull request, I confirm that my contribution is made under the terms of
the Apache 2.0 license.

## Summary by CodeRabbit

* **New Features**
  * Added Snowflake as a supported provider with CLI scanning.
  * Added RSA key-pair authentication and read-only account assessment.
  * Added the `users_mfa_enabled` check for password users without MFA protection.
  * Added Snowflake findings, summaries, HTML output, and mutelist support.

* **Documentation**
  * Added Snowflake setup and authentication guidance.

* **Tests**
  * Added comprehensive Snowflake provider, service, check, output, and mutelist coverage.


kkmookhey:
Thanks — this was a useful review. Ten of the twelve are addressed in ed0261bb3; two I have
not taken, with reasoning below.

**The two that mattered most were both silent-truncation bugs**, which is the failure mode
this provider most needed to avoid:

- **HTTP 202 was treated as success.** The SQL API answers 202 while a statement is still
  running, returning a status document rather than results — so a statement over 45
  seconds yielded zero rows with no error, and a large account would have rendered as
  clean. Now polled to completion with a bound; a statement that never finishes raises
  rather than returning an empty set.
- **Only the first partition was read.** Every partition is now fetched, and a partitioned
  result with no statement handle raises rather than silently truncating.

Both have tests, including one asserting that a never-completing statement raises rather
than returning nothing.

**The serialization finding was the best catch.** `__repr__` was overridden, but
`model_dump()`, `model_dump_json()` and `dict(session)` all still published the client
holding the RSA private key. `Field(exclude=True)` fixes the first two — and pydantic still
yields excluded fields on iteration, so `__iter__` is overridden for the third. The test I
added failed on `dict()` until that second change, which is exactly why it is worth having.

Also addressed: account-value validation before it reaches the request host; a failed user
read now raises instead of reporting an empty inventory (with one service, swallowing it
renders an unreadable account as clean); both required grants documented; the passphrase
example reads via a silent prompt rather than an `export`; the overview describes only the
implemented check; missing annotations and docstring; and the changelog fragment reformatted
— I checked the existing fragments and 22 of 24 have no terminal period, so that one was
plainly right.

**Not taken — the mutelist example comments** (`prowler/config/snowflake_mutelist_example.yaml`).
The grammar suggestions are correct in isolation, but that wording is copied verbatim from
the other providers' mutelist examples. Changing only the Snowflake one would make it the
single inconsistent file of the set. Happy to do it as a sweep across all providers in a
separate PR if that is wanted, but not for this one alone.

Re-verified against a live Snowflake account after these changes, not only against the
tests.

kkmookhey:
All three from the second pass are addressed in 899592579 — and both code findings were
defects in my previous fix rather than in the original, which is a fair hit.

- **The poll budget was shorter than the statement timeout.** 30 attempts at 2s gave 60
  seconds against a 120-second `STATEMENT_TIMEOUT_SECONDS`, so a statement needing between
  the two would have raised even though Snowflake was about to return rows — failing on
  exactly the large accounts the 202 handling was added for. The budget is now derived
  from the statement timeout rather than set alongside it, so the two cannot drift apart
  again, and there is a test asserting the relationship rather than the number.
- **`_absolute` trusted the status URL from the response.** `_send` attaches the signed JWT
  to whatever it returns, so an `http://` or off-host `statementStatusUrl` would have sent
  the credential in cleartext or elsewhere. Only https on this account's host is accepted
  now; anything else raises rather than being followed. Tested both ways.
- Documentation wording taken as suggested.

Re-verified against a live Snowflake account after the change.

PR #12661, opened 2026-08-30
feat(eks): assess Kubernetes network policy enforcement in the Amazon VPC CNI add-on

tamg-aws:
One new check over the Amazon VPC CNI managed add-on, plus the add-on configuration the EKS collector
needs to read it.

## What this asserts

`eks_cluster_vpc_cni_network_policy_enforced` reports whether the Amazon VPC CNI **managed add-on** has
Kubernetes network policy enforcement switched on. That is the precondition for any `NetworkPolicy`
resource to take effect *through this enforcer*: without it, policies can exist in the cluster and go
unenforced by the VPC CNI.

**Every verdict is a statement about the managed add-on, not about the cluster.** Two architectures live
outside the EKS API's view and neither is an edge case — a third-party policy engine, and a self-managed
VPC CNI. Both are covered in the disclosures below, and the shipped FAIL message names them, because an
operator reading a CSV row sees `status_extended` and nothing else.

**What it deliberately does not claim.** `NetworkPolicy` resources live in the cluster and are not
exposed by the EKS API. This check asserts the precondition only, and the PASS message says so
explicitly rather than implying that pod-to-pod traffic is actually restricted.

## Verdict semantics

- **PASS** — the add-on sets `enableNetworkPolicy` to true.
- **FAIL** — the add-on sets it to false. The message is scoped to the add-on and says so: "Amazon VPC CNI
  **managed add-on** does not enforce Kubernetes network policies, since it sets `enableNetworkPolicy` to
  false. Enforcement by a third-party policy engine or a self-managed VPC CNI is not visible to the EKS API
  and is not evaluated." It deliberately does not say the *cluster* does not enforce network policies,
  which would be simply false for a Calico cluster.
- **MANUAL**, five distinct causes, each an unknown rather than a violation: the cluster's add-ons
  could not be listed; the cluster does not use the Amazon VPC CNI *managed* add-on, so the setting is
  not visible from the EKS API at all and a self-managed CNI must be reviewed in-cluster; the add-on
  configuration could not be read; `configurationValues` is not a readable JSON object; or
  **`enableNetworkPolicy` does not resolve to true or false — which includes the key being absent
  entirely.**

**That last cause is the ordinary default install, so it is worth stating plainly rather than as an edge
case.** The branch is `configuration.get(NETWORK_POLICY_KEY) is None`, and `.get` returns `None` for an
absent key just as it does for an explicit null. `Addon.configurationValues` is documented at the pinned
botocore as "the configuration values that *you* provided", and the add-on configuration schema declares
no default and no required member for `enableNetworkPolicy` — so a managed `vpc-cni` add-on installed
without explicit configuration carries no such key and reports **MANUAL**, not PASS and not FAIL. A
reviewer asking "what does this check say about a cluster nobody has configured?" gets MANUAL, which is
the correct answer for an unobserved setting.

The boolean reader accepts both `"true"` and `true`, because the add-on configuration schema types
this field as a string carrying `"format": "boolean"` — so the API returns the string — while the
schema is add-on-version specific and the blob is otherwise untyped.

## Verification

- **53 tests pass** across `tests/providers/aws/services/eks/`.
- Coverage on the added lines is **182/182 = 100.00%** — 129/129 in the check, 53/53 in `eks_service.py`.
- Every added function and class is documented (26/26 across the 4 changed files), and import ordering is
  clean.
- The added configuration-value reader was exercised with oversized and malformed input rather than only
  the values a healthy add-on returns.

## Disclosures

**A FAIL can be wrong about the cluster, in two documented architectures, and the shipped message says so.**
This is the disclosure a maintainer most needs, because it bounds what the check is evidence of. Both were
verified against AWS documentation rather than reasoned about:

**1. A third-party policy engine.** The EKS Best Practices Guide has a section titled *"ThirdParty Network
Policy Engines"* recommending Calico and Cilium (`eks/latest/best-practices/network-security.html`). A
cluster enforcing through Calico correctly leaves `enableNetworkPolicy` **false**, because you do not run
two enforcers against the same traffic. So the FAIL is accurate about the add-on and wrong about the
cluster, and this is a recommended architecture rather than an oddity.

**2. A self-managed VPC CNI.** AWS documents **three** ways to switch the feature on
(`eks/latest/userguide/cni-network-policy-configure.html`): the managed add-on, `helm upgrade` with
`enableNetworkPolicy=true`, and editing the `amazon-vpc-cni` ConfigMap to add
`enable-network-policy-controller: "true"` alongside the `aws-node` DaemonSet argument. **Only the first is
readable from the EKS API.** The other two leave the setting in in-cluster state, so a cluster enforcing
that way also shows the add-on flag false — or has no managed add-on at all, which reports MANUAL.

**Neither is fixable by reading more AWS API surface, and the verdict logic is deliberately unchanged.**
The flag being explicitly false is a real measurement worth reporting; downgrading the FAIL to MANUAL
would trade away a true signal to fix what was a wording problem. Detecting Calico or Cilium is also
deliberately not attempted — any signal available from the EKS API would be a guess presented as a
measurement. The check's metadata `Notes` carry both architectures and both citations, so the bound
travels with the check and not only with this PR.

**MANUAL for a cluster with no managed add-on is a scope statement rather than a gap.** Roughly the same
evidence would be needed to assess a self-managed CNI as to assess the `NetworkPolicy` resources
themselves, and neither is reachable from the EKS API. Reporting FAIL there would assert a
misconfiguration never observed; reporting PASS would assert enforcement never observed. The message
directs the operator to review the in-cluster configuration.

**`codecov/project` will be red, and does not gate.** Measured across all 8 currently-open PRs in this
campaign: `codecov/project` is FAILURE on 8 of 8 while `codecov/patch` is SUCCESS on 8 of 8. #12459
merged with `codecov/project` red. `codecov/patch` is green here at 100.00%.

## Summary by CodeRabbit

* **New Features**
  * Added an AWS EKS security check for network policy enforcement in the Amazon VPC CNI add-on.
  * Reports **PASS** when enforcement is enabled, **FAIL** when explicitly disabled, and **MANUAL** when configuration cannot be verified.
  * Includes remediation guidance for enabling network policy enforcement through supported AWS management options.
  * EKS assessments now collect relevant add-on configuration details to support this evaluation.

### DIFF
--- a/prowler/changelog.d/eks-cluster-vpc-cni-network-policy.added.md
+++ b/prowler/changelog.d/eks-cluster-vpc-cni-network-policy.added.md
@@ -1,0 +1,1 @@
+`eks_cluster_vpc_cni_network_policy_enforced` check for AWS provider, flagging EKS clusters whose Amazon VPC CNI managed add-on does not enable Kubernetes network policy enforcement, and reporting MANUAL where the EKS API cannot show the setting
--- a/prowler/compliance/aws/aws_ai_security_framework_aws.json
+++ b/prowler/compliance/aws/aws_ai_security_framework_aws.json
@@ -1128,6 +1128,7 @@
         "eks_cluster_not_publicly_accessible",
         "eks_cluster_private_nodes_enabled",
         "eks_cluster_network_policy_enabled",
+        "eks_cluster_vpc_cni_network_policy_enforced",
         "eks_cluster_uses_a_supported_version",
         "eks_control_plane_logging_all_types_enabled",
         "eks_cluster_kms_cmk_encryption_in_secrets_enabled",
--- a/prowler/providers/aws/services/eks/eks_cluster_vpc_cni_network_policy_enforced/eks_cluster_vpc_cni_network_policy_enforced.metadata.json
+++ b/prowler/providers/aws/services/eks/eks_cluster_vpc_cni_network_policy_enforced/eks_cluster_vpc_cni_network_policy_enforced.metadata.json
@@ -1,0 +1,42 @@
+{
+  "Provider": "aws",
+  "CheckID": "eks_cluster_vpc_cni_network_policy_enforced",
+  "CheckTitle": "EKS cluster enforces Kubernetes network policies through the Amazon VPC CNI add-on",
+  "CheckType": [
+    "Software and Configuration Checks/AWS Security Best Practices/Network Reachability",
+    "TTPs/Lateral Movement"
+  ],
+  "ServiceName": "eks",
+  "SubServiceName": "",
+  "ResourceIdTemplate": "",
+  "Severity": "medium",
+  "ResourceType": "AwsEksCluster",
+  "ResourceGroup": "container",
+  "Description": "**Amazon EKS clusters** are evaluated for whether the **Amazon VPC CNI** managed add-on sets `enableNetworkPolicy` to `true`, which is what makes the CNI enforce Kubernetes `NetworkPolicy` resources. The policy objects themselves live in the cluster and are not exposed by the EKS API, so only this enforcement precondition is verified.",
+  "Risk": "Without CNI **network policy enforcement** every `NetworkPolicy` an operator authors is inert, so pods reach every other pod and service in the cluster. That unrestricted east-west path lets one compromised container move **laterally** to sidecars, tool executors and internal APIs, widening the blast radius and easing **data exfiltration**.",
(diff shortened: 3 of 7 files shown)


danibarranqueroo (approved):
Tested on real infra with an EKS cluster in a dev account. FAIL and PASS both verified live against the managed vpc-cni add-on, and the remediation CLI works verbatim. Approving, good job! 🥇

PR #12680, opened 2026-08-31
feat(sdk): AWS partition selection via PROWLER_AWS_PARTITION

pedrooot:
### Context

Prowler hardcodes `us-east-1` as the STS bootstrap region when it validates credentials and when it resolves the region for `AssumeRole`. That endpoint does not exist outside the commercial partition, so a deployment pointed at GovCloud, China, EUSC or any of the ISO partitions fails at credential validation before it can scan anything.

There was also nothing stopping a deployment meant for one partition from validating an account that lives in another one, which silently produces a scan against the wrong estate.

### Description

Adds the `PROWLER_AWS_PARTITION` environment variable to select the partition a deployment targets.

- `get_botocore_partition_regions()` reads the partition-to-region map straight from botocore's `endpoints` data, so the list of valid partitions follows the installed botocore instead of a hardcoded table. Cached with `lru_cache(maxsize=1)` since it never changes within a process.
- `get_env_partition_bootstrap_region()` resolves the variable to that partition's first region, and raises `AWSInvalidPartitionError` listing the valid partitions when the value is not one botocore knows.
- `test_connection()` now defaults `aws_region` to `None` and resolves it at call time to the configured partition's bootstrap region, falling back to `us-east-1`. It also forwards that region as `sts_region` when assuming a role.
- After resolving the caller identity, `test_connection()` compares `caller_identity.arn.partition` with the configured partition and raises `AWSInvalidPartitionError` on a mismatch, honouring `raise_on_exception` like the other error paths.
- `get_aws_region_for_sts()` gains the partition as a precedence step: explicit `--region` values first, then the configured partition, then the session region, then the existing bootstrap candidates.

`AWSInvalidPartitionError` (code 1917) already existed; this PR is the first thing to raise it.

Behaviour is unchanged when `PROWLER_AWS_PARTITION` is unset, which stays the default.

### Steps to review

Resolution of the bootstrap region per partition:

```bash
PROWLER_AWS_PARTITION=aws-us-gov python -c "from prowler.providers.aws.aws_provider import get_env_partition_bootstrap_region as r; print(r())"
# us-gov-east-1

PROWLER_AWS_PARTITION=aws-cn python -c "from prowler.providers.aws.aws_provider import get_env_partition_bootstrap_region as r; print(r())"
# cn-north-1

PROWLER_AWS_PARTITION=not-a-partition python -c "from prowler.providers.aws.aws_provider import get_env_partition_bootstrap_region as r; r()"
# AWSInvalidPartitionError, listing the valid partitions
```

Unset the variable and confirm nothing changes: `test_connection()` still bootstraps on `us-east-1` and `get_aws_region_for_sts()` keeps its previous precedence.

Tests:

```bash
uv run pytest tests/providers/aws/aws_provider_test.py -q -k "partition"
# 21 passed (10 added here, the rest pre-existing partition coverage)
```

The ten new cases cover the four non-commercial partitions, the precedence rules (explicit regions win over the partition, the partition wins over the session region), the invalid-value error, the STS region used by `test_connection()` with and without a role, and the account/partition mismatch.

### Checklist

<details>

</details>

#### SDK/CLI

#### UI

#### API

#### MCP Server

### License

## Summary by CodeRabbit

* **New Features**
  * Added support for selecting the AWS partition used during credential validation and scan startup through the `PROWLER_AWS_PARTITION` environment variable.
  * Automatically selects suitable STS regions for supported partitions, including GovCloud, China, European Sovereign Cloud, and isolated regions.
  * Provides a clear connection error when the configured partition does not match the AWS account.

* **Bug Fixes**
  * Improved STS region selection while preserving explicitly provided regions and endpoint overrides.
  * Enhanced partition-specific STS support for static and role-based connections.

### DIFF
--- a/prowler/providers/aws/aws_provider.py
+++ b/prowler/providers/aws/aws_provider.py
@@ -1645,7 +1645,7 @@ class AwsProvider(Provider):
         try:
             # Botocore resolves the regional STS endpoint for every partition
             # (China, EUSC, GovCloud, ISO); AWS_ENDPOINT_URL overrides it
-            sts_endpoint_url = os.environ.get("AWS_ENDPOINT_URL")
+            sts_endpoint_url = os.environ.get("AWS_ENDPOINT_URL") or None
             return session.client("sts", aws_region, endpoint_url=sts_endpoint_url)
         except Exception as error:
             logger.critical(
--- a/prowler/providers/aws/aws_provider.py
+++ b/prowler/providers/aws/aws_provider.py
@@ -672,7 +672,12 @@ class AwsProvider(Provider):
             if mfa:
                 session = Session(**session_arguments)
                 session._session.set_default_client_config(session_config)
-                sts_client = session.client("sts")
+                sts_region = (
+                    get_env_partition_bootstrap_region()
+                    or session.region_name
+                    or AWS_STS_GLOBAL_ENDPOINT_REGION
+                )
+                sts_client = AwsProvider.create_sts_session(session, sts_region)
 
                 # TODO: pass values from the input
                 mfa_info = AwsProvider.input_role_mfa_token_and_code()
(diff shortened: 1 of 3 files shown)


Davidm4r (dismissed):

prowler/providers/aws/aws_provider.py
```diff
@@ -1730,6 +1794,10 @@ def get_aws_region_for_sts(
             if region not in excluded_regions:
                 return region
 
+    env_partition_region = get_env_partition_bootstrap_region()
+    if env_partition_region and env_partition_region not in excluded_regions:
```
HugoPBrito: I believe the configured partition must constrain the entire fallback chain. If `aws-us-gov` is configured and `us-gov-east-1` is excluded, this can fall through to commercial `us-east-1` instead of trying `us-gov-west-1`. Could we select another non-excluded region from the same partition and add a regression test?

prowler/providers/aws/aws_provider.py
```diff
@@ -1412,6 +1415,12 @@ def test_connection(
             Connection(is_connected=True, Error=None))
         """
         try:
+            if aws_region is None:
+                aws_region = (
+                    get_env_partition_bootstrap_region()
```
HugoPBrito, line 1425: I believe we also need to pass this region into the MFA bootstrap. `setup_session(mfa=True)` creates an STS client without a region before reaching credential validation, so non-commercial partitions can still use the commercial global endpoint. Could we cover this with a GovCloud MFA test?

prowler/providers/aws/aws_provider.py
```diff
@@ -1702,6 +1722,46 @@ def read_aws_regions_file() -> dict:
...
+def get_botocore_partition_regions() -> dict:
+    """
+    Get the AWS partitions and their regions from the botocore endpoints data.
+
+    Returns:
+        dict: A dictionary mapping each partition name to its sorted list of regions.
+    """
+    endpoints_data = BotocoreSession().get_data("endpoints")
+    return {
+        partition["partition"]: sorted(partition.get("regions", {}))
```
HugoPBrito: I believe alphabetical ordering does not guarantee a safe bootstrap region. Botocore lists valid STS endpoints, but `af-south-1` is opt-in and may be unavailable for the target account during `AssumeRole`. Could we preserve `us-east-1` as the stable bootstrap for the commercial partition and add coverage for `PROWLER_AWS_PARTITION=aws`?

HugoPBrito (changes requested):
Thanks for adding support for selecting the AWS partition. The overall direction makes sense, but I believe there are still a few paths where STS can use a region outside the configured partition. These cases should be addressed before merging.

HugoPBrito (approved):

Davidm4r (approved):

PR #12691, opened 2026-09-01
fix(image): switch to Bearer when registry rejects Basic auth

pedrooot:
### Context

Follow up of #12678. While testing against registries that answer `Basic` on `/v2/` but ask for `Bearer` on `/_catalog`, the catalog listing died with an auth error. The retry logic in the OCI adapter only covered bearer -> bearer refresh and bearer -> basic (the Harbor case), never basic -> bearer.

### Description

Added the missing switch in `_authed_request`: on a 401 with a `Bearer` challenge and no bearer token, obtain a token from the challenge itself (it already carries the right scope, e.g. `registry:catalog:*`) and retry. The token is kept afterwards, so the next catalog pages and tag listings go straight with bearer, no repeated token exchanges.

Same-origin only, like the existing basic fallback. Works with credentials and anonymously.

### Steps to review

- `oci_adapter.py` -> new block at the end of `_authed_request`, mirrors the existing basic fallback
- `uv run pytest tests/providers/image -n auto` -> 282 passed (5 new tests in `TestBearerAuthSwitch`: switch with creds, multi page switches only once, anonymous switch, no switch cross-origin, failed token exchange raises)
- tested e2e with a local mock registry answering a different scheme per endpoint -> listing works, token exchange happens exactly once, next pages reuse the token. Pre-fix code reproduces the exact failure

### Checklist

<details>

</details>

#### SDK/CLI

#### UI

#### API

#### MCP Server

### License

### DIFF
--- a/prowler/providers/image/lib/registry/oci_adapter.py
+++ b/prowler/providers/image/lib/registry/oci_adapter.py
@@ -215,6 +215,19 @@ class OciRegistryAdapter(RegistryAdapter):
                 # skip the doomed Bearer round-trips.
                 self._basic_auth_verified = True
                 self._bearer_token = None
+        if (
+            resp.status_code == 401
+            and not self._bearer_token
+            and self._is_same_origin_as_registry(url)
+            and resp.headers.get("Www-Authenticate", "").lower().startswith("bearer")
+        ):
+            # The inverse switch: /v2/ negotiated Basic (or anonymous) but this
+            # endpoint demands Bearer. The challenge carries the right scope.
+            logger.debug(f"Basic auth not accepted for {url}, switching to Bearer")
+            self._bearer_token = self._obtain_bearer_token(
+                resp.headers.get("Www-Authenticate", "")
+            )
+            resp = self._do_authed_request(method, url, **kwargs)
         return resp
 
     def _do_authed_request(self, method: str, url: str, **kwargs) -> requests.Response:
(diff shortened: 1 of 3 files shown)


prowler/providers/image/lib/registry/oci_adapter.py
```diff
@@ -215,6 +215,19 @@ def _authed_request(self, method: str, url: str, **kwargs) -> requests.Response:
                 # skip the doomed Bearer round-trips.
                 self._basic_auth_verified = True
                 self._bearer_token = None
+        if (
+            resp.status_code == 401
+            and not self._bearer_token
+            and self._is_same_origin_as_registry(url)
+            and resp.headers.get("Www-Authenticate", "").lower().startswith("bearer")
```
HugoPBrito: I believe this still misses valid responses containing multiple authentication challenges, for example `WWW-Authenticate: Basic realm="registry", Bearer realm="...", ...`. HTTP allows multiple challenges and explicitly permits Basic to appear first. In that case, `startswith("bearer")` is false, so no token exchange occurs and the catalog request remains unauthorized. Could we select the Bearer challenge regardless of its position and add a regression test for the combined-header case?

HugoPBrito (changes requested):
Thanks for the focused fix and the thorough coverage of the primary transition. I found one valid multi-challenge response that still bypasses the Bearer switch. Once that case is handled and covered, this should be ready.

prowler/providers/image/lib/registry/oci_adapter.py
pedrooot: good catch! fixed -> added a helper that picks the challenge by scheme wherever it sits in the header, all three checks use it now. it also fixes a hidden one: the realm regex ran over the full header, so with a combined header the token exchange would have grabbed the basic realm. added tests for the combined cases

HugoPBrito (approved):

PR #12700, opened 2026-09-01
fix(sdk): report all-users 2sv override failures

### LINKED ISSUE #12692: fix(googleworkspace): security_2sv_enforced returns MANUAL for the all-user 2SV requirement when a group override shadows the failure
### Context

`security_2sv_enforced` is mapped to two CIS Google Workspace requirements with different scopes, in both the 1.3 and 1.4 frameworks:

| Requirement | Scope |
|---|---|
| 4.1.1.1 | users in administrative roles |
| 4.1.1.3 | all users |

Since #12513, when the domain-wide policy fails and every failing setting is also overridden for a group or an organizational unit, the check returns `MANUAL` (`failures_shadowed_by_overrides`, `security/lib/scope.py`). The Policy API does not expose the values behind those overrides, so for 4.1.1.1 the overriding group could well be the administrators and the failure cannot be confirmed for them. That reasoning is correct for the administrator-scoped requirement.

It does not hold for 4.1.1.3. A failing domain-wide policy still proves that every user outside the override is non-compliant, so `MANUAL` discards a failure Prowler can actually demonstrate.

This was raised by @HugoPBrito in review and was not addressed before the PR merged.

### Steps to Reproduce

On `master`, with 2-Step Verification enforcement off domain-wide and that same setting overridden for one group:

```python
c.policies = SecurityPolicies(
    two_sv_enforced_from=None,
    two_sv_allow_enrollment=True,
    two_sv_allow_trusting_device=False,
    two_sv_allowed_factor_set="NO_TELEPHONY",
    overridden_settings=["security.two_step_verification_enforcement"],
)
security_2sv_enforced().execute()
```

### Actual Result

```
MANUAL
2-Step Verification is not enforced as required in the domain-wide policy of example.com:
enforcement is not configured and defaults to OFF. However, every failing setting is also
overridden for at least one group or organizational unit, so the audited scope may be
configured correctly. Review those overrides in the Admin console.
```

CIS 4.1.1.3 is reported as not evaluated, when the users outside that single group have no 2-Step Verification enforced at all.

### Expected Result

4.1.1.3 fails, because a domain-wide failure is proven for every user the override does not reach. 4.1.1.1 stays `MANUAL`, because the override may cover the administrators.

The same check cannot serve both scopes correctly with a single verdict.

### Possible fixes

Two options, both suggested in the original review:

1. Keep `FAIL` in `security_2sv_enforced` and leave the `MANUAL` logic only in `security_2sv_hardware_keys_admins`, whose three mappings (CIS 4.1.1.2, CIS Controls 8.1 6.5, SCuBA `GWS.COMMONCONTROLS.1.1`) are all administrator scoped. Cheap, and correct for three of the four requirements this check feeds. 4.1.1.1 keeps the existing override note in the finding message.
2. Split the check into an all-users one and an administrator-scoped one, and map each requirement to the right one. Correct for all four, at the cost of a new check with its own metadata, tests and mappings.

### Impact

Only reachable when a group or a sub-organizational unit overrides the 2-Step Verification settings, which is common in larger domains.

0xTaoZ:
### Context

Fixes #12692.

`security_2sv_enforced` is mapped to both administrator-scoped and all-users Google Workspace 2-Step Verification requirements. The check currently returns `MANUAL` when every failing domain-wide setting is overridden for a group or organizational unit. That uncertainty is appropriate for administrator-scoped checks, but the all-users requirement still has a proven domain-wide failure for users outside the override.

### Description

This keeps domain-wide 2-Step Verification failures as `FAIL` in `security_2sv_enforced`, while preserving the existing override caveat in the finding message. The administrator-specific `security_2sv_hardware_keys_admins` check keeps its `MANUAL` downgrade because the overriding group may be the audited admin group.

A focused changelog fragment is included under `prowler/changelog.d/`.

### Steps to review

```bash
uv run pytest tests/providers/googleworkspace/services/security/security_2sv_enforced/security_2sv_enforced_test.py tests/providers/googleworkspace/services/security/security_2sv_hardware_keys_admins/security_2sv_hardware_keys_admins_test.py -q
uv run --group dev flake8 prowler/providers/googleworkspace/services/security/security_2sv_enforced/security_2sv_enforced.py tests/providers/googleworkspace/services/security/security_2sv_enforced/security_2sv_enforced_test.py --ignore=E266,W503,E203,E501,W605,E128
uv run --group dev black --check prowler/providers/googleworkspace/services/security/security_2sv_enforced/security_2sv_enforced.py tests/providers/googleworkspace/services/security/security_2sv_enforced/security_2sv_enforced_test.py
uv run --group dev pylint --disable=W,C,R,E -rn -sn prowler/providers/googleworkspace/services/security/security_2sv_enforced/security_2sv_enforced.py
git diff --check
```

### Checklist

<details>

- [x] This feature/issue is listed in [here](https://github.com/prowler-cloud/prowler/issues/12692) or roadmap.prowler.com
- [ ] Is it assigned to me, if not, request it via the issue/feature in [here](https://github.com/prowler-cloud/prowler/issues/12692) or [Prowler Community Slack](goto.prowler.com/slack)

</details>

#### SDK/CLI

### License

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated the Google Workspace 2-Step Verification check to report a **FAIL** when domain-wide enforcement is not met, even if failing settings are overridden for groups or organizational units.
  - Improved failure messaging to clearly indicate when overridden settings still contribute to the domain-wide failure.

### DIFF
--- a/prowler/changelog.d/googleworkspace-2sv-all-users-overrides.fixed.md
+++ b/prowler/changelog.d/googleworkspace-2sv-all-users-overrides.fixed.md
@@ -1,0 +1,1 @@
+`security_2sv_enforced` reports domain-wide 2-Step Verification failures as FAIL even when every failing setting is overridden for a group or organizational unit
--- a/prowler/providers/googleworkspace/services/security/security_2sv_enforced/security_2sv_enforced.py
+++ b/prowler/providers/googleworkspace/services/security/security_2sv_enforced/security_2sv_enforced.py
@@ -8,7 +8,6 @@ from prowler.providers.googleworkspace.services.security.lib.durations import (
     parse_duration_seconds,
 )
 from prowler.providers.googleworkspace.services.security.lib.scope import (
-    failures_shadowed_by_overrides,
     override_caveat,
     unevaluable_reason,
 )
@@ -145,23 +144,9 @@ class security_2sv_enforced(Check):
                     )
                 )
 
-            failing_settings = frozenset(setting for setting, _ in issues)
             reasons = "; ".join(text for _, text in issues)
 
-            if issues and failures_shadowed_by_overrides(policies, failing_settings):
-                # The audited scope (e.g. the admin group of 4.1.1.1) may get
-                # the overriding value, which the Policy API does not expose,
-                # so the domain-wide failure cannot be confirmed for it.
-                report.status = "MANUAL"
-                report.status_extended = (
-                    f"2-Step Verification is not enforced as required in the "
-                    f"domain-wide policy of {domain}: {reasons}. However, every "
-                    f"failing setting is also overridden for at least one group "
-                    f"or organizational unit, so the audited scope may be "
-                    f"configured correctly. Review those overrides in the Admin "
-                    f"console."
-                )
-            elif issues:
+            if issues:
                 report.status = "FAIL"
                 report.status_extended = (
                     f"2-Step Verification is not enforced as required in domain "
--- a/tests/providers/googleworkspace/services/security/security_2sv_enforced/security_2sv_enforced_test.py
+++ b/tests/providers/googleworkspace/services/security/security_2sv_enforced/security_2sv_enforced_test.py
@@ -224,8 +224,8 @@ class TestSecurity2svEnforced:
         assert "trust their device" in findings[0].status_extended
         assert "also overridden" in findings[0].status_extended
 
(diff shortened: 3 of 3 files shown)


lydiavilchez (approved):
Thanks a lot for this quick fix @0xTaoZ ! 🔝

pedrooot (approved):

Issue #11064, opened 2026-05-06
[New Check]: Application registrations should not use password credentials (client secrets)


HugoPBrito:
### Existing check search

- [x] I have searched existing issues, Prowler Hub, and the public roadmap, and this check does not already exist.

### Provider

Microsoft 365

### Service or product area

entra

### Suggested check name

`entra_app_registration_no_password_credentials`

### Context and goal

- **Security condition to validate:** No application registration in the tenant has any entry in `passwordCredentials` (client secrets). Customer-owned applications should authenticate using certificates (`keyCredentials`), federated identity credentials (`federatedIdentityCredentials`), or managed identities — never long-lived shared secrets.
- **Why it matters:** Client secrets are bearer credentials that are easy to leak (committed to repos, copied into CI variables, shared via chat). Once leaked they can be used from anywhere on the internet until rotated. Microsoft's official guidance for the identity platform is to migrate away from client secrets in favour of certificate-based authentication, federated identity credentials, or managed identities. Prowler already has `entra_default_app_management_policy_enabled`, which checks the *policy* that prevents *new* secrets from being added; this new check audits the *current state* and surfaces apps that already hold secrets (legacy or pre-policy).
- **Resource involved:** Microsoft Entra application registrations (`/applications`) and their `passwordCredentials` collection.

### Expected behavior

- **Resource or scope to evaluate:** All application objects in the tenant via `GET https://graph.microsoft.com/v1.0/applications?$select=id,displayName,appId,passwordCredentials`. Iterate every application and inspect `passwordCredentials`.
- **PASS when:** every application's `passwordCredentials` collection is empty.
- **FAIL when:** at least one application has one or more entries in `passwordCredentials`. The finding should report `appId`, `displayName`, the count of password credentials, and (where present) `keyId`, `displayName`, `startDateTime`, and `endDateTime` of each — so operators can correlate to the specific secret entry to remove in the portal.
- **MANUAL when:** not applicable.
- **Exclusions / edge cases:**
  - Both expired and active `passwordCredentials` count toward FAIL — expired entries are credential clutter that should be cleaned up regardless. Operators who want to ignore expired-only entries can use Prowler's mutelist mechanism on a per-app basis.
  - The check operates on `applications` (app registrations), not on `servicePrincipals` (enterprise apps from other tenants). Microsoft first-party app secrets do not appear here because they are owned by Microsoft tenants.

### References

[...]

PR #12748, opened 2026-09-06
feat(m365): add entra check for federated credentials on privileged apps

### LINKED ISSUE #12747: [New Check]: Federated identity credentials on app registrations have no audit coverage
### Existing check search

### Provider

Microsoft 365

### New provider name

_No response_

### Service or product area

entra

### Suggested check name

`entra_app_registration_federated_credential_on_privileged_app`

### Context and goal

- **Security condition to validate:** application registrations that hold `federatedIdentityCredentials` are currently invisible to every Entra check. Neither `prowler/providers/m365/services/entra/` nor `prowler/providers/azure/services/entra/` reads that collection today, while the sibling credential types are covered (`entra_app_registration_client_secret_unused` on m365, `entra_app_registration_credential_not_expired` on azure).
- **Why it matters:** #11064's rationale for the client-secrets work recommends certificates, managed identities and federated identity credentials as the migration targets - so as tenants follow that guidance, the credential type they move to is the one the scan says nothing about. A federated identity credential grants standing token exchange to an external issuer/subject pair and has no expiry, so it sits outside the existing expiry- and unused-secret-shaped checks by construction. The highest-signal case is an FIC on an application that also holds privileged permissions: an external CI system (GitHub Actions, GitLab CI, etc.) can then mint tokens that act with those permissions, and the trust boundary lives entirely in the external provider's account/repo namespace. A related concern for a follow-up rather than this first check: for public CI issuers the subject match is exact-string, and a deleted external org or repository name can later be re-registered by an unrelated party.
- **Resource involved:** Microsoft Entra application registrations (`/applications`) and their `federatedIdentityCredentials` collection.

### Expected behavior

- **Resource or scope to evaluate:** all application objects in the tenant, expanding `federatedIdentityCredentials` (via `$expand` on the existing applications listing, or the per-application `GET /applications/{id}/federatedIdentityCredentials` endpoint), joined with the application's permission grants that the existing privileged-permissions check already models.
- **PASS when:** an application either has no federated identity credentials, or has them while holding no privileged permissions.
- **FAIL when:** an application holds at least one federated identity credential AND privileged permissions. The finding should report `appId`, `displayName`, and each FIC's `name`, `issuer` and `subject`, so operators can see exactly which external identity can act with the privileged rights.
- **Exclusions / edge cases:** this is deliberately not a check against having FICs in general - they are the recommended replacement for client secrets, and flagging their mere presence would penalize the practice #11064 encourages. The proposal scopes FAIL to the privileged-app combination only.

### References

- Microsoft Graph v1.0 - `federatedIdentityCredential` resource (List/Get methods, `issuer`/`subject`/`audiences` properties): https://learn.microsoft.com/en-us/graph/api/resources/federatedidentitycredential?view=graph-rest-1.0
- Microsoft Entra Workload ID - workload identity federation: https://learn.microsoft.com/en-us/entra/workload-id/workload-identity-federation
- Prior art in this repo: #11064 (client-secrets state check and its recommendation of FICs as a migration target)

### Suggested severity

Medium

- **Existing patterns to follow:** `entra_app_registration_no_unused_privileged_permissions` already iterates application registrations and models privileged permissions; `entra_app_registration_client_secret_unused` already extends the applications listing with a credential collection. This check is the composition of the two patterns.
- **Permissions / scopes:** the FIC list should be covered by the existing application-read baseline, though that is worth confirming against the Graph `List federatedIdentityCredentials` permissions table rather than taken from here.
- Microsoft Graph v1.0 only; no PowerShell needed.

Is there a reason federated identity credentials were left out of scope so far - or would a first check shaped like the two existing credential checks fit the service model?

earbona23:
### Context

Fixes #12747.

Federated identity credentials (FICs) on application registrations are currently invisible to every Entra check. `entra_app_registration_client_secret_unused` (m365) and `entra_app_registration_credential_not_expired` (azure) cover client secrets, but nothing reads the `federatedIdentityCredentials` collection. As #11064 recommends FICs as the migration target away from client secrets, the credential type tenants are moving to is the one the scan says nothing about. A FIC grants standing token exchange to an external issuer/subject pair with no stored secret and no expiry, so it sits outside the expiry- and unused-secret-shaped checks by construction. The highest-signal case is a FIC on an application that also holds privileged rights: an external CI system (GitHub Actions, GitLab CI, etc.) can then mint tokens that act with those rights, and the trust boundary lives entirely in the external provider's account/repo namespace.

### Description

Adds the m365 check `entra_app_registration_federated_credential_on_privileged_app`.

- **Service (`entra_service.py`)**: adds a `FederatedIdentityCredential` model, extends `AppRegistration` with a `federated_identity_credentials` list, and reads it in `_get_app_registrations` per application via `GET /applications/{id}/federatedIdentityCredentials` — the same per-application-object idiom already used to resolve owners in `_get_service_principals`. A failure to read one application's collection is logged and left empty, never aborting the inventory.
- **Check**: joins application registrations (which own the FICs) with their service principal's permanent Tier 0 (Control Plane) directory-role assignments — the exact privileged model already used by `entra_service_principal_no_secrets_for_permanent_tier0_roles` — matching by `appId`.
  - **PASS**: the application has no FICs, or has them while holding no Tier 0 directory role.
  - **FAIL**: the application has at least one FIC **and** a permanent Tier 0 directory-role assignment. The finding reports the app's `displayName`/`appId` (via the resource) and each FIC's `name`, `issuer` and `subject`.
- As the issue requests, the mere presence of FICs is never flagged (they are the recommended replacement for client secrets); only the privileged combination fails.

**Design note (privileged model):** the issue points at `entra_app_registration_no_unused_privileged_permissions` for the privileged half, but that data comes from Defender XDR App Governance (`ThreatHunting.Read.All` + App Governance enabled) and is `MANUAL` when unavailable — while the issue also specifies "MANUAL: not applicable". To honor the no-MANUAL contract with a Graph-native signal that is always available, I modeled "privileged" as a permanent Control Plane (Tier 0) directory-role assignment, reusing the existing `TIER_0_ROLE_TEMPLATE_IDS` set and the `entra_service_principal_no_secrets_for_permanent_tier0_roles` shape (credential type × Tier 0 role). This is conservative (no false positives) and self-contained. Extending "privileged" to app-permission grants (e.g. `RoleManagement.ReadWrite.Directory`) is a natural follow-up. Happy to switch to the Defender-XDR privileged model instead if you prefer that trade-off.

### Steps to review

- New check detected: `uv run python prowler-cli.py m365 --list-checks | grep entra_app_registration_federated_credential_on_privileged_app`
- Tests:
  - `uv run python -m pytest tests/providers/m365/services/entra/entra_app_registration_federated_credential_on_privileged_app/` — 7 passed (PASS/FAIL, non-privileged, no-SP, multi-credential truncation, mixed, no-resources).
  - `uv run python -m pytest tests/providers/m365/services/entra/microsoft365_entra_service_test.py` — includes 2 new tests for the FIC fetch (happy path + per-application error isolation).
  - `uv run python -m pytest tests/lib/check/` — check metadata validation passes.
- Lint: `black`, `isort --profile black`, and `flake8 --ignore=E266,W503,E203,E501,W605` clean on the changed files.

### Permissions

The `federatedIdentityCredentials` list is covered by the existing application-read baseline (`Application.Read.All`); no new provider permission is required. Privileged status reuses data already gathered by the service principals / directory-role reads.

### Checklist

- Are there new checks included in this PR? **Yes** — no new provider permissions required.
- [x] Code documented following the Google Python style guide.
- [x] Changelog fragment added under `prowler/changelog.d/`.

### License

## Summary by CodeRabbit

* **New Features**
  * Added a Microsoft Entra security check that identifies privileged app registrations using federated identity credentials.
  * App inventory now includes federated identity credential details.
  * Individual credential retrieval failures no longer prevent the affected app registration from being inventoried; affected results may require manual review.

* **Tests**
  * Added coverage for privileged and non-privileged applications, multiple credentials, missing service principals, and credential retrieval failures.

### DIFF
--- a/prowler/changelog.d/entra-app-registration-federated-credential-on-privileged-app.added.md
+++ b/prowler/changelog.d/entra-app-registration-federated-credential-on-privileged-app.added.md
@@ -1,1 +1,1 @@
-Add `entra_app_registration_federated_credential_on_privileged_app` check to flag federated identity credentials on app registrations that hold a Control Plane (Tier 0) directory role
+`entra_app_registration_federated_credential_on_privileged_app` check to flag federated identity credentials on app registrations that hold a Control Plane (Tier 0) directory role
--- a/prowler/changelog.d/entra-app-registration-federated-credential-on-privileged-app.added.md
+++ b/prowler/changelog.d/entra-app-registration-federated-credential-on-privileged-app.added.md
@@ -1,0 +1,1 @@
+Add `entra_app_registration_federated_credential_on_privileged_app` check to flag federated identity credentials on app registrations that hold a Control Plane (Tier 0) directory role
--- a/prowler/providers/m365/services/entra/entra_app_registration_federated_credential_on_privileged_app/entra_app_registration_federated_credential_on_privileged_app.metadata.json
+++ b/prowler/providers/m365/services/entra/entra_app_registration_federated_credential_on_privileged_app/entra_app_registration_federated_credential_on_privileged_app.metadata.json
@@ -1,0 +1,42 @@
+{
+  "Provider": "m365",
+  "CheckID": "entra_app_registration_federated_credential_on_privileged_app",
+  "CheckTitle": "Privileged application registrations should not have federated identity credentials",
+  "CheckType": [],
+  "ServiceName": "entra",
+  "SubServiceName": "",
+  "ResourceIdTemplate": "",
+  "Severity": "medium",
+  "ResourceType": "NotDefined",
+  "ResourceGroup": "IAM",
+  "Description": "Microsoft Entra **application registrations** that hold a **Control Plane (Tier 0) directory role** should not also have **federated identity credentials**. Such a credential grants an external issuer/subject pair standing token exchange to the application with no stored secret and no expiry, letting an external workload act with the application's privileged rights.",
+  "Risk": "A **federated identity credential** lets an external issuer/subject pair (a GitHub Actions or GitLab CI job, for example) mint tokens for the application with no stored secret and no expiry. When the application's service principal also holds a **Tier 0 (Control Plane) directory role**, that external workload can act with tenant-takeover privileges from the external provider's namespace.",
+  "RelatedUrl": "",
+  "AdditionalURLs": [
+    "https://learn.microsoft.com/en-us/graph/api/resources/federatedidentitycredential?view=graph-rest-1.0",
+    "https://learn.microsoft.com/en-us/entra/workload-id/workload-identity-federation",
+    "https://learn.microsoft.com/en-us/entra/identity/role-based-access-control/privileged-roles-permissions"
+  ],
+  "Remediation": {
+    "Code": {
+      "CLI": "",
+      "NativeIaC": "",
+      "Other": "1. Navigate to Microsoft Entra admin center (https://entra.microsoft.com)\n2. Go to Applications > App registrations\n3. Select the flagged application\n4. Review whether the application needs its Control Plane (Tier 0) directory role; if not, remove the role assignment under Roles and administrators\n5. If the role is required, go to Certificates & secrets > Federated credentials and confirm each credential's issuer and subject are tightly scoped to a specific external identity, removing any that are unnecessary\n6. Prefer isolating privileged automation into a dedicated application with a narrowly scoped federated credential",
+      "Terraform": ""
+    },
+    "Recommendation": {
+      "Text": "Do not combine federated identity credentials with Control Plane (Tier 0) directory roles on the same application. Either scope the application down so it no longer holds a privileged role, or move the privileged workload to a dedicated application whose federated credential subject and issuer are tightly restricted to the intended external identity.",
+      "Url": "https://hub.prowler.com/check/entra_app_registration_federated_credential_on_privileged_app"
+    }
+  },
+  "Categories": [
+    "identity-access",
+    "privilege-escalation"
+  ],
+  "DependsOn": [],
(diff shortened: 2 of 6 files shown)


earbona23:
Thanks for the review. Addressed all three points in 968fb86:

1. **Changelog fragment** — dropped the leading `Add` so it reads as a verbless fragment (`.added.md` already implies the action), no terminal period.

2. **`execute` docstring** — added a Google-style docstring describing what it evaluates (federated identity credentials on privileged app registrations) and that it returns a `list[CheckReportM365]`, matching the style of the other entra checks.

3. **Fail-open on retrieval error** — this was the real one. `_get_app_registrations()` swallowed a per-application federated-credential Graph failure into an empty list, so a Tier 0 app whose credentials couldn't be read looked credential-free and reported PASS. I followed the existing service convention (`users_error: Optional[str] = None`, etc.): added a per-application `federated_identity_credentials_error: Optional[str] = None` on `AppRegistration`, set it in the `except` block instead of only logging, and had the check report **MANUAL** (never PASS) when a privileged app carries that error, mirroring how `entra_users_mfa_capable` handles `users_error`. `None` means the empty list is authoritative; a non-`None` value means "unknown". Non-privileged apps stay PASS since the error is immaterial without a Tier 0 role.

Updated the service test to assert the error is now recorded, and added two check-level tests pinning "privileged app + retrieval error → MANUAL" and "non-privileged app + retrieval error → PASS". Full entra check + service suite: 71 passed. `black`/`flake8` clean.

PR #12759, opened 2026-09-08
feat(aws): support the ISO partitions for region resolution and scanning

StylusFrost:
### Context

Prowler resolves the regions of every AWS service from `prowler/providers/aws/aws_regions_by_service.json`. That file only carries four partitions: `aws`, `aws-cn`, `aws-eusc` and `aws-us-gov`. There is not a single `aws-iso*` key in its 408 services.

That absence, and not the code, is why GovCloud works and the ISO partitions do not. There is no GovCloud-specific branch anywhere in this path: the same functions serve every partition, and they simply have no data to answer with when asked about an ISO one. The lookup is an unguarded index, so instead of reporting the service as unavailable it raises `KeyError`.

### Description

**Region matrix.** The generator run weekly by `prowler-bot` now fills `aws-iso`, `aws-iso-b`, `aws-iso-e` and `aws-iso-f` from the endpoints data bundled with botocore. That data ships inside the wheel, so it needs no credentials and no network, which is what makes it usable for partitions the bot can never reach. The SSM pass is left untouched and remains the only source with data for the existing partitions: botocore's endpoints data carries no services at all for `aws-eusc`, so the two sources have to coexist. Regenerating the four existing partitions yields no change: the diff on the matrix is pure insertions.

Services are keyed by endpoint prefix in the botocore data and by service name in the matrix, so the mapping is derived from the service models themselves rather than written by hand. It resolves `monitoring`→`cloudwatch`, `elasticloadbalancing`→`elb`+`elbv2`, `states`→`stepfunctions`, `tagging`→`resourcegroupstaggingapi` and the rest on its own. Two prefixes do not resolve and are handled as named constants: `ce` is renamed to `costexplorer`, and `transcribestreaming` is ignored because `transcribe` is already resolved through its own prefix. Any other unresolved prefix aborts the run, so a service newly available in a partition is noticed rather than silently dropped from scans.

Services whose only endpoint is a partition-global pseudo endpoint keep the shape the matrix already uses: all regions of the partition for `iam`, `route53`, `organizations` and `support`, a single region for `costexplorer`.

**Provider.** Three assumptions broke as soon as a partition was not one of the four:

- `get_available_aws_service_regions()` indexed the matrix directly and raised `KeyError` for an unknown service or partition. It now yields an empty set, the same outcome a service explicitly recorded as unavailable already produced.
- `generate_regional_clients()` caught that `KeyError`, logged it and fell through without returning, handing back `None` while promising a dict. Callers then failed with `AttributeError: 'NoneType' object has no attribute 'values'`, with the real cause left behind in a log line. It now returns an empty dict.
- `get_global_region()` returned the literal `"aws-iso-global"` for anything matching `aws-iso`. That is a botocore pseudo endpoint rather than a region, and matching on a substring collapsed the four ISO partitions into one answer. It now reads the partition's global STS region from the botocore endpoints data, via the helper already used for partition bootstrap.

**Side effect worth noting for reviewers:** the seven new ISO regions now appear in `get_regions(partition=None)`, which is what builds the `choices` of `--region` and `--excluded-region`. An ISO region is therefore accepted by argument parsing, where before it was rejected as an invalid value.

The bot's install is pinned to the versions already in `pyproject.toml`, since the bundled endpoints data is now itself a data source and has to be deterministic across runs.

### Steps to review

The two commits are independent and each is green on its own, so they can be reviewed separately.

**1. Verify the matrix is reproducible rather than hand-edited.** From the repo root, applying the generator's ISO pass to the file as it stands on `master` reproduces the committed file byte for byte, and re-applying it to the result changes nothing:

```python
import json, importlib.util
spec = importlib.util.spec_from_file_location("gen", "util/update_aws_services_regions.py")
gen = importlib.util.module_from_spec(spec); spec.loader.exec_module(gen)

data = json.load(open("prowler/providers/aws/aws_regions_by_service.json"))
gen.add_subservices_and_missing_services(data)
gen.add_iso_partitions_regions(data)
assert json.dumps(data, indent=2, sort_keys=True) + "\n" == open(
    "prowler/providers/aws/aws_regions_by_service.json").read()
```

**2. Verify no existing partition moved.** Comparing the committed matrix against `master`'s, service by service, over `aws`, `aws-cn`, `aws-eusc` and `aws-us-gov` gives zero differences. The 408 services keep carrying every partition key, now eight instead of four, with an empty list where a service is not available.

**3. Verify the failure mode is gone.** On `master`, `get_available_aws_service_regions("guardduty", "aws-iso-b")` raises `KeyError`; with this branch it returns `{'us-isob-east-1'}`, and an unknown partition returns an empty set. `generate_regional_clients` returns `{}` rather than `None` for a service absent from the audited partition.

**4. Verify the global region.** `get_global_region()` returns exactly the previous values for `aws`, `aws-cn`, `aws-eusc` and `aws-us-gov`, pinned by tests so the refactor cannot silently move them, and `us-iso-east-1` / `us-isob-east-1` for the ISO partitions.

**5. Verify the generator fails loudly.** Removing a service from the matrix so its prefix stops resolving aborts the run with a message naming the prefix, the partition and the two ways to resolve it.

`tests/providers/aws/aws_provider_test.py` passes with 137 tests.

### Checklist

<details>

</details>

#### SDK/CLI

### License

## Summary by CodeRabbit

* **New Features**
  * Added support for AWS ISO partitions and their service-region mappings.
  * AWS services in ISO partitions are now included in scanning.
  * ISO partition lookups now return actual regional endpoints.
  * Added the FedRAMP 20x KSI 2026 compliance framework across supported platforms.

* **Bug Fixes**
  * Unavailable services and partitions are skipped safely instead of causing errors.
  * Regional client creation now handles failures consistently.

* **Chores**
  * AWS region data generation now uses fixed dependency versions for consistent results.

* **Removals**
  * Removed superseded FedRAMP 20x Phase One pilot frameworks.


.github/workflows/sdk-refresh-aws-services-regions.yml
```diff
@@ -44,7 +44,10 @@ jobs:
           cache: 'pip'
 
       - name: Install dependencies
-        run: pip install boto3
+        # Pinned to the versions in pyproject.toml: the ISO partitions region
+        # data comes from the endpoints.json bundled with botocore, so the
+        # botocore version is itself a data source and must be deterministic
+        run: pip install boto3==1.40.61 botocore==1.40.61
```
cesararroba, line 50: Pinning botocore freezes the ISO half of this job. Since `endpoints.json` ships inside the wheel, the ISO pass will emit the same output every Monday until someone edits this line, while the SSM pass keeps refreshing from the live API, so the staleness is invisible in the generated PR.

It will also drift from `pyproject.toml`: that pin has already moved from `1.39.15` to `1.40.61`, and both bumps landed in unrelated PRs. `pip install .` gives determinism per commit and still follows the botocore version Prowler ships.

Related, if the pin does start following `pyproject.toml`: the new `ValueError` aborts the whole run, so one unresolved ISO prefix would also block the weekly refresh of `aws`, `aws-cn`, `aws-eusc` and `aws-us-gov`.

cesararroba (commented):
Verified locally with botocore 1.40.61: the matrix regenerates byte for byte, it is idempotent on re-run, and no existing partition moves. The provider fixes look right.

One thing outside the diff: `docs/user-guide/providers/aws/regions-and-partitions.mdx:148` still says Prowler has no built-in way to scan the ISO partitions and tells users to edit `aws_regions_by_service.json` by hand, with `aws-iso-global` listed as a region in the example. That page contradicts this PR now, worth updating here.

StylusFrost:
You're right that the page contradicts this PR, and it is worse than it looks. Beyond claiming there is no built-in support, it tells users to hand-edit `aws_regions_by_service.json`, an edit that `prowler-bot` wipes on its next weekly refresh, so the workaround it documents never survived. Its example also lists `aws-iso-global` and `aws-iso-b-global` as regions; those are botocore pseudo endpoints, not regions, and they are exactly what this PR stops returning.

One thing I am doing differently from what you asked. You suggested updating the page here, and I have put it in #12763 instead, stacked on this one, so the docs can be reviewed on their own and this PR stays code only. Say the word if you would rather have it folded back in here.

The rewrite follows the same shape as the China, GovCloud and EUSC sections, with the real regions of the four ISO partitions. It also carries a warning that support has not been exercised against a live ISO account and that scanning inside these partitions is pending validation in a real environment, so that caveat lives on the page itself and not only in this thread. The regions and per-service availability come from the SDK's endpoint metadata and the behaviour is covered by tests, but that is not the same as having run a scan there.

That PR also documents `PROWLER_AWS_PARTITION`, which is not documented anywhere today. It is not API only: `get_aws_region_for_sts()` is called from `AwsProvider.__init__` and `setup_session`, so it steers the STS bootstrap region on a plain CLI scan too. Only the partition mismatch validation is confined to `test_connection`.

On this PR I also took CodeRabbit's point about `write_regions_by_service()` building its path from `__name__`, which resolved against the working directory rather than the script. Fixed in e478cca3f.

pedrooot (approved):

PR #12773, opened 2026-09-09
fix(image): honour TRIVY_CACHE_DIR when it is set

cesararroba:
### Context

The Image provider builds a fresh temporary cache directory for every provider instance and deletes it when the scan is done:

```python
self._trivy_cache_dir_obj = tempfile.TemporaryDirectory(prefix="prowler-trivy-cache-")
self._trivy_cache_dir = self._trivy_cache_dir_obj.name
...
trivy_command = ["trivy", "image", "--cache-dir", self._trivy_cache_dir, ...]
```

There is no way to point it anywhere else. `--cache-dir` on the command line wins over Trivy's own `TRIVY_CACHE_DIR`, so setting that variable has no effect today.

Two consequences follow.

The vulnerability database is downloaded again for every image scanned, on any deployment, because the cache never survives the scan that filled it. The code already treats this download as a source of trouble elsewhere, in the same file:

> Uses registry HTTP APIs directly instead of trivy to avoid false failures caused by trivy DB download issues.

And a deployment without egress cannot scan images at all. It has no way to supply a database it already holds, since the provider will not read from anywhere it is given. This is how the Image provider behaves on an air-gapped install: not degraded, unusable.

### Description

When `TRIVY_CACHE_DIR` names a directory, the provider uses it and leaves it alone. When it does not, nothing changes: the temporary directory is created and cleaned up exactly as before.

The supplied directory is deliberately not deleted on cleanup. It holds a database the deployment may have no way to fetch again, and it does not belong to the provider that borrowed it.

### Steps to review

The change is in `ImageProvider.__init__`, and `cleanup` needed nothing: it was already guarded on the attribute existing, so a run that never created a temporary directory has nothing to remove.

```
pytest tests/providers/image/image_provider_test.py -k cache_dir
```

Three cases added: a configured directory is used, a configured directory survives cleanup, and an unset variable keeps today's create-and-delete behaviour. The full `image_provider_test.py` suite passes (146 tests).

### Why it is worth having beyond air-gap

A connected deployment scanning a hundred images downloads the database a hundred times. Pointing the variable at a persistent path makes that once.

## Summary by CodeRabbit

* **New Features**
  * Image scanning now supports configuring Trivy’s cache location with `TRIVY_CACHE_DIR`.
  * Configured cache directories preserve vulnerability databases between scans, reducing repeated downloads.
  * Persistent caching supports offline scanning when a suitable database is available.

* **Bug Fixes**
  * Temporary cache behavior remains unchanged when no directory is configured.
  * Configured cache directories are preserved during cleanup, while temporary caches continue to be removed automatically.

* **Documentation**
  * Added guidance on cache configuration, offline usage, and database freshness.

### DIFF
--- a/docs/user-guide/providers/image/getting-started-image.mdx
+++ b/docs/user-guide/providers/image/getting-started-image.mdx
@@ -98,6 +98,8 @@ For additional installation methods, see the [Trivy installation guide](https://
 
 ### Vulnerability Database Cache
 
+<VersionBadge version="5.42.0" />
+
 Trivy keeps its vulnerability database in a cache directory. By default Prowler gives it a temporary one and removes it when the scan ends, so the database is downloaded again for every scan.
 
 Set `TRIVY_CACHE_DIR` to a directory that persists and the database is downloaded once and reused:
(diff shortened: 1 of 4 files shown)


cesararroba:
Thanks for the `TRIVY_SKIP_DB_UPDATE` addition, that was a real gap and not just a docs one.

Supplying the database is not enough on its own, because Trivy refreshes it when it judges it stale and that download has nowhere to go. What makes it nasty is the timing: a database fetched today has a NextUpdate a few hours out, so the scan works on the day it is set up and starts failing once that passes. Exactly the kind of thing a same-day test signs off on.

Also good call on the autouse fixture clearing the variable. My tests would have behaved differently on a machine that happens to have it set, which is a test that agrees with whoever runs it.

Both flags are now set wherever a persistent cache is configured on our side too.

PR #12815, opened 2026-09-15
feat(ui): allow disabling self-registration in Cloud

pedrooot:
### Context

Some Prowler Cloud deployments, Private Cloud in particular, must be invitation-only: today anyone who reaches the sign-in page can create an account and, with it, a brand-new organization. The API rejects those requests in Prowler Cloud; this PR adds the UI side so the interface stops offering flows the API refuses.

### Description

- New runtime flag `UI_SELF_REGISTRATION_ENABLED`, exposed through the runtime public-config island. It is on by default and only `"false"` (case-insensitive) turns it off.
- `isSelfRegistrationEnabled()` honours the flag only when `isCloud()`. Prowler Local Server always keeps sign-up open, whatever the value.
- With the flag off:
  - the sign-in page drops the "Sign up" link
  - `/sign-up` redirects to `/sign-in` unless the URL carries an `invitation_token`, so invitation links keep working
  - the profile page hides **Create organization**
  - the Google and GitHub callbacks turn a `403` with code `self_registration_disabled` into a "Registration Disabled" message on sign-in instead of the generic authentication error
- No behaviour change when the flag is unset.

### Steps to review

1. Local Server is unaffected: with `UI_SELF_REGISTRATION_ENABLED=false` and without `UI_CLOUD_ENABLED`, the "Sign up" link and `/sign-up` are still there.
2. With `UI_CLOUD_ENABLED=true` and `UI_SELF_REGISTRATION_ENABLED=false`:
   - `/sign-in` shows no "Sign up" link
   - `/sign-up` redirects to `/sign-in`
   - `/sign-up?invitation_token=<token>` renders the form
   - `/profile` shows no **Create organization** button
3. Tests:
   ```bash
   cd ui && pnpm vitest run lib/shared/env.test.ts lib/runtime-env.test.ts lib/runtime-config.test.ts \
     lib/get-runtime-config.client.test.ts lib/auth-callback-url.test.ts app/api/auth/callback/google/route.test.ts \
     "app/(auth)/(guest-only)/sign-up/page.test.tsx" components/users/profile/memberships-card-client.test.tsx
   ```

### Checklist

<details>

<summary><b>Community Checklist</b></summary>

- [ ] This feature/issue is listed in the [open issues](https://github.com/prowler-cloud/prowler/issues?q=sort%3Aupdated-desc+is%3Aissue+is%3Aopen) or roadmap.prowler.com
[...]

PR #12818, opened 2026-09-15
feat: record the tenant profile declared at onboarding

alejandrobailo:
### Context

Every buyer-type signal Prowler Cloud has today (accounts connected, users invited) is shaped by what the product lets a tenant do after signing up. A short profile step at first login, before the "Add your first provider" gate, is the one segmentation that does not depend on that: three closed questions and a skip.

The product step lives here in OSS. The analytics layer on top (PostHog events and tenant properties) is a separate PR in prowler-cloud that listens to the window events this step dispatches.

### Description

**API**

- `TenantOnboardingProfile` model with RLS, one row per tenant (`declared_cloud_accounts`, `declared_role`, `declared_seniority`, `skipped`, `submitted_by`, `inserted_at`), migration `0098`, filter, read and create serializers.
- `POST /onboarding-profiles` (basename `onboarding-profile`), open to any member. Either the three buckets or `skipped: true`, validated by `ChoiceField`s. Idempotent and immutable: once a row exists the endpoint answers 200 with it and never overwrites, so the first answer, given before the product could bias it, stays. A concurrent race on the unique row also resolves to the stored answer. `GET /onboarding-profiles` lists the tenant's row (empty before the step runs).

**UI**

- `OnboardingProfileGate` wraps `OnboardingGate`: the profile modal resolves first, then the tour gate takes over with the same `hasProviders` signal. Mounted in `(prowler)/layout.tsx` inside the existing `cloudEnabled` block, which also reads `isOnboardingProfileRecorded()` only for tenants without providers.
- `OnboardingProfileModal`: three radio groups, Continue disabled until all are answered, Skip and close both count as a skip. `shouldStartOnboardingProfile` fails open on any doubt. Own per-tenant localStorage marker (`prowler.onboarding.profile.<tenantId>`) mirroring the checkpoint marker; a rejected submission writes no marker so the step returns next login.
- Outcomes (`shown`, `submitted` with the answers, `skipped`) are dispatched as the `prowler:onboarding-profile-step` window event from `lib/onboarding/onboarding-events.ts`. The step has no listener of its own, so the onboarding stays free of tracking dependencies; `onboarding-events.ts` also declares the invite-step event a follow-up PR uses.
- Server actions in `actions/onboarding/profile.ts` with Zod validation. Onboarding guide updated.

### Steps to review

1. `cd api && uv run pytest src/backend/api/tests/test_onboarding_profile.py` and `uv run python src/backend/manage.py makemigrations api --check --dry-run`.
2. `cd ui && pnpm vitest run components/onboarding lib/onboarding`.
3. With `isCloud()` on, sign up a new tenant: the profile modal appears before "Add your first provider". Answer it, then `POST /onboarding-profiles` again: 200 with the first answer. Clear localStorage and reload: the modal does not return. On a fresh tenant press Skip: a row with `skipped: true` is stored.

### Checklist

- [x] Review if code is being documented.

#### UI
- [x] All requirements work as expected on the UI
- [x] Changelog fragment under `ui/changelog.d/`

#### API
- [x] All requirements work as expected on the API
- [x] Migration verified with `makemigrations --check`
- [ ] API specs regeneration: not done in this PR
- [x] Changelog fragment under `api/changelog.d/`

### License

## Summary by CodeRabbit

- **New Features**
  - Added a skippable onboarding profile step for new tenants before provider setup.
  - Users can provide their cloud-account range, team size, area of work, role, and seniority.
  - Added tenant-specific profile API support, filtering, and one-time persistence.
  - Added onboarding outcome events for shown, submitted, and skipped states.
  - Profile prompts are tenant-specific and avoid repeating after completion.

- **Documentation**
  - Documented the onboarding profile flow and API behavior.


alejandrobailo:
Thanks — addressed in `9fc9fd4`. Five of the six findings were real; one I'm declining with a reason.

**Fixed**

- **Success detection in the server actions.** This was the important one. `handleApiResponse` returns `{ error, status }` without an `errors` array for a 403, throws on 5xx, and `handleApiError` returns `{ error }` on a transport failure, so the gate's `result?.errors?.length` check read all three as success and wrote the marker for a row that was never created. Both actions now return an explicit `{ stored, error }` outcome, and the gate announces `submitted` and writes the marker only when `stored` is true. New tests in `ui/actions/onboarding/profile.test.ts` cover the transport failure, the errorless rejection and the thrown 5xx.
- **Skip path.** It now awaits the same outcome, keeps the modal disabled while the request is in flight, and marks the step handled only on a stored skip.
- **Tenant-scoped marker.** The key is now `prowler.onboarding.profile.<tenantId>`, with the tenant validated as a UUID before it reaches the key and resolved server-side in the layout. A test covers the case you described: a browser that answered for one tenant is still asked for a second one.
- **Discriminated union** for `OnboardingProfileStepDetail`, so only a submitted step carries the answers.
- **Detail route** now raises `MethodNotAllowed`; `extend_schema(exclude=True)` only hid it from the docs. Covered by a test asserting 405.
- **README** no longer claims zero backend coupling; it separates the client-only tour sequence from the profile step's persistence.

**Declined**

- **`updated_at` on `TenantOnboardingProfile`.** The row is write-once by design (a second POST returns the stored one and never overwrites), so `updated_at` would always equal `inserted_at`. Fifteen existing OSS models follow the same append-only shape with `inserted_at` alone, among them `ComplianceOverview`, `ScanSummary`, `ThreatScoreSnapshot` and the relationship tables. Adding the column here would suggest a mutability the endpoint does not allow.

The `ui-tests` failure was unrelated to these findings: Prettier on `ui/lib/onboarding/README.md`. Fixed. `e2e-tests` fails the same way on every open branch right now (GCP and GitHub provider specs), and the onboarding block does not render there since `UI_CLOUD_ENABLED` is unset in that workflow.

alejandrobailo:
Addressed in `bb61cb1`, plus a change to the step itself.

**CodeRabbit**

- **`useCallback` in the profile gate** — removed. The snapshot closure still reads the current `tenantId`; its identity changing per render is harmless here, because `useSyncExternalStore` compares the value it returns, not the function.
- **`updated_at`** — still declining, same reason as before: the row is write-once (a second POST returns the stored one), so the column would always equal `inserted_at` and would suggest a mutability the endpoint does not allow. Fifteen existing OSS models are `inserted_at`-only for the same reason.

**New: the step now asks for position, not only discipline**

The step asked *which area do you work in* but not *how far up the organisation the answer comes from*, so a security engineer and a CISO landed in the same bucket. Adds a fourth closed question, `declared_seniority`:

| Value | Label |
| --- | --- |
| `practitioner` | Practitioner / IC |
| `lead` | Team lead / Manager |
| `director` | Director / Head of |
| `executive` | VP / C-level |
| `founder` | Founder / Owner |

Founder is its own rung rather than a shade of practitioner: a one-person tenant run by a founder evaluating Prowler is a different prospect from a lone engineer inside a large company.

Because the two questions are now orthogonal, `management` leaves the discipline list — the ladder covers it — and `compliance_grc` takes its place, which the compliance surfaces make a segment worth telling apart.

Migration `0098` is edited in place rather than followed by a second one: it has not merged, so no database has ever run it. `makemigrations --check` is clean and the 18 API tests pass.

Also merged `master` for the fast-uri suppression (#12823), so the API container scan is no longer red here.

alejandrobailo:
Both addressed in `0b060b43`.

**Seniority missing from the request example** — correct, and my omission: I added the field to the serializer without touching the example next to it, so copying `DeclareProfile` out of the docs returned 400. Example and description updated.

**Auth read outside the error boundary** — also correct, though not quite for the stated reason. The claim is that a rejection here "can abort the complete layout render"; in practice the layout awaits `getProviders()`, `getScansByState()` and a bare `auth()` before it ever reaches `isOnboardingProfileRecorded()`, so an unreadable session aborts the render several lines earlier regardless of what this action does.

The reason worth fixing it is the write path. `postOnboardingProfile` promises `{ stored, error }` precisely so the gate can decide whether to write its marker, and a rejection escaping past that promise defeats the change made in `b08c814`: `handleSubmit` only ever inspects `stored`. An expired session between page load and clicking Continue is the realistic case, and it has to read as "not stored" so the step returns next login rather than being silently swallowed.

Moved the header read inside each boundary and added three tests, one per entry point. They fail against the previous commit and pass now:

```
× reports a submission as unstored when the session cannot be read
× reports a skip as unstored when the session cannot be read
× reads an unreadable session as an unknown profile state
```

`useCallback` was fixed in `bb61cb19`; the thread is marked outdated but stayed open, so I've resolved it. The `updated_at` thread stays open on purpose — still declining, same reason as before.

The remaining red check is `e2e-tests`, which is unrelated: `PROVIDER-E2E-007` (GCP), `-008` (GitHub) and `-019` (Okta) need provider credentials and fail identically on #12815, which touches none of this. That workflow also leaves `UI_CLOUD_ENABLED` unset, so the profile step never renders there.

alejandrobailo:
Simplified in `59cbd68a`: the step now asks three questions, not four. `declared_team_size` is gone.

Team size and cloud-account count were measuring the same thing twice — the size of the company. Nobody runs two hundred accounts with a team of one, so the second question bought a dimension that was almost a function of the first.

The cost of carrying both was in the analysis, not the code. Four questions across five, four, five and five buckets are 500 cross-tab cells; three across five, five and five are 125. Same number of respondents, four times the density. At our signup volume that is the difference between a table that fills up and one that never does. One question fewer also raises the share of people who finish the step.

| | Before | Now |
| --- | --- | --- |
| Questions | 4 | 3 |
| Cross-tab cells | 500 | 125 |

A side effect worth noting: `"1"` was a valid answer to both the account count and the team size, so the two radio groups had colliding accessible names. That ambiguity is gone.

Migration `0098` edited in place again — still unmerged, still never run anywhere. `makemigrations --check` clean, 18 API tests and 234 UI tests green. prowler-cloud/prowler-cloud#5772 updated to match, and its drift test confirms the two field tuples are identical.

Separately: prowler-cloud/prowler-cloud#5773, the A/B experiment wrapping prowler-cloud/prowler#12819, is closed. It was underpowered for our volume — detecting a small retention lift would have needed hundreds to low thousands of tenants per arm. The invite step still ships, at 100%, which is the honest version of that decision.

